PR #5, opened 2015-04-07
Bump to a rust commit that builds on BSD

wg:
Version 0.1.5 is missing a couple of fixes from @semarie for building on various BSD variants, which also prevents cargo from building. This PR updates the rust submodule to a version that includes the fixes.


alexcrichton:
Thanks!

Issue #15, opened 2015-07-11
Update rust to fix musl builds


aidanhs:
I'm in no particular hurry about this, I've just created the issue so I can be notified when it's closed.

https://github.com/rust-lang/rust/commit/52862e4cda1ff72e52dfaea21cf82ae93725b3be#diff-4c1827e2ef5156ad90c1cef13755f4c8

alexcrichton:
Done!

alexcrichton:
Ended up yanking 0.1.9 due to [compatibility concerns](https://github.com/rust-lang/libc/issues/16)

dhuseby:
@alexcrichton this will also need to be updated to pick up the FreeBSD i686 changes when [Rust PR #27360 lands](https://github.com/rust-lang/rust/pull/27360).

dhuseby:
I'll send you a PR.  The only question is if I need to bump the minor version or just the patch number.

agrover:
This is a blocker for me. I need flock(2) for http://github.com/agrover/melvin. I added it to Rust repo on June 24, but until libc crate is updated I can't update nix-rust, and then start using it in my project. @alexcrichton I know you say in #17 you're busy but I just wanted to describe my situation to maybe increase this issue's relative status a little.

aidanhs:
Why not just use a .cargo file? http://doc.crates.io/guide.html#overriding-dependencies

Clone the repo (with submodules), add .cargo and you're pretty much good to go.

alexcrichton:
I've opened https://github.com/rust-lang/rust/pull/27762 and if that lands then we can upgrade libc without breaking anything.

tailhook:
I can confirm that libc with rust submodule works nice for me (even with musl build). So looking forward for release soon.

tailhook:
Thanks!

PR #17, opened 2015-08-05
Add build script and LIBC_PATH environ variable

tailhook:
Well, the solution here is probably not optimal. But it's the easiest way I've found so far to build libc (as dependency) for `x86_64-unknown-linux-musl`.

The problem is that when building with musl, the target libc is not the one in default path. So I get output similar to:
```
error: could not find native static library `c`, perhaps an -L flag is missing?
Could not compile `libc`.

Caused by:
  Process didn't exit successfully: `rustc libc/rust/src/liblibc/lib.rs --crate-name libc --crate-type lib -g --cfg feature="cargo-build" --cfg feature="default" -C metadata=49eeab76b364dec9 
-C extra-filename=-49eeab76b364dec9 --out-dir /work/target/x86_64-unknown-linux-musl/debug/deps --emit=dep-info,link --target x86_64-unknown-linux-musl -L dependency=/work/target/x86_64-unknown-linux-musl/debug/deps -L dependency=/work/target/x86_64-unknown-linux-musl/debug/deps`
```

According to cargo docs the `build.rs` script is there suited to find dependencies (or sometimes build one if not found). So with this script setting env var  to `LIBC_PATH=/usr/local/musl/lib` I can now build my project.

Any other ideas of how to build the lib are appreciated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ A library for types and bindings to native C functions often found in libc or
 other common platform libraries.
 """
 include = ["Cargo.toml", "rust/src/liblibc/*"]
+build = "build.rs"
 
 [features]
 default = ["cargo-build"]
@@ -21,3 +22,5 @@ cargo-build = []
 [lib]
 name = "libc"
 path = "rust/src/liblibc/lib.rs"
+
+
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,7 @@
+use std::env;
+
+fn main() {
+    if let Ok(path) = env::var("LIBC_PATH") {
+        println!("cargo:rustc-link-search={}", path);
+    }
+}


alexcrichton:
Thanks for the PR! This was actually [fixed upstream](https://github.com/rust-lang/rust/commit/52862e4cda1ff72e52dfaea21cf82ae93725b3be) and we just need to [update libc](https://github.com/rust-lang/libc/issues/15) to pick up the change. We haven't updated yet because there was a breaking change in liblibc.

tailhook:
@alexcrichton  So how do you recommend to build libc now? It's also not clear what stops you to update now? Is it incompatibility in "nix" crate? And when do you plan to upgrade?

alexcrichton:
Lack of time to deal with this right now stops the update unfortunately.

PR #21, opened 2015-10-29
Implement RFC 1291 changes

alexcrichton:
This PR is an implementation of [RFC 1291](https://github.com/rust-lang/rfcs/blob/master/text/1291-promote-libc.md) which contains the changes before migrating this repo to the rust-lang organization.


src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+mod dox;
+
+#[repr(u8)]
+pub enum c_void {
+    // Two dummy variants so the #[repr] attribute can be used
+    #[doc(hidden)]
+    __variant1,
+    #[doc(hidden)]
+    __variant2,
+}
```
brson: Can these be private? This seems risky exposing these bogus fields.

brson: Also, why does this want to be u8? Can you comment?

brson:
I don't see a version bump in Cargo.toml. What's the version after this lands?

src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub fn strerror(n: c_int) -> *mut c_char;
+    pub fn strtok(s: *mut c_char, t: *const c_char) -> *mut c_char;
+    pub fn strxfrm(s: *mut c_char, ct: *const c_char, n: size_t) -> size_t;
+    pub fn wcslen(buf: *const wchar_t) -> size_t;
+
+    pub fn memcmp(cx: *const c_void, ct: *const c_void, n: size_t) -> c_int;
+    pub fn memchr(cx: *const c_void, c: c_int, n: size_t) -> *mut c_void;
+}
+
+// These are all inline functions on android
```
brson: What does that mean for their definition in Rust? Do we provide compatibility? Can you comment?

brson:
Stunning work.

src/lib.rs
alexcrichton: Unfortunately I don't think they can be private (no way to have a public enum with private variants), but I think we're covered in terms of "reasonable backwards compatibility". I've added a comment though.

src/lib.rs
alexcrichton: Right now this just means they're not defined in Rust at all, so they just don't show up in the docs at all, but I'll add a comment.

alexcrichton:
I plan to run some more tests and migrate a few repos to this, and then after that I'll bump to 1.0.0 and release

brson:
OK. I'm pretty concerned about the ramifications of bumping this massively complicated crate to 1.0. Before you do it you should reassure me that we can maintain semver.

Here's hoping the test coverage is amazing, since I just have to take your word all these definitions are correct.

r=me

alexcrichton:
Gonna go ahead and merge now, as discussed on IRC we can hold off on the version bump/publish for now.

PR #27, opened 2015-11-02
Add sigaltstack/siginfo bindings

alexcrichton:


alexcrichton:
Oops, already there under other names.

PR #37, opened 2015-11-04
Make fd_set fields public

### LINKED ISSUE #36: fd_set has private fields
http://rust-lang-nursery.github.io/libc/x86_64-unknown-linux-gnu/libc/struct.fd_set.html

sfackler:
Closes #36

### DIFF
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -64,7 +64,7 @@ s! {
     }
 
     pub struct fd_set {
-        fds_bits: [i32; FD_SETSIZE / 32],
+        pub fds_bits: [i32; FD_SETSIZE / 32],
     }
 }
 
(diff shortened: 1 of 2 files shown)


thibault-ml:
I had an issue when using `fd_set` because of private fields. I ended up having to use `malloc`, which has isn't great. 

However, I don't believe the field should be public as it has private meanings.
Instead, libc should provide a `new` method or a initialiser function.

alexcrichton:
I'd be fine to merge something like this, but the CI builds are red unfortunately.

Note that in the meantime you can initialize an `fd_set`  with `mem::{uninitialized, zero}` and then use the `FD_*` macros to manipulate them (as C would)

alexcrichton:
I think this needs some more work to move forward (e.g. rather than just making the fields public), so I'm going to close this for now. Feel free to resubmit if the definitions are updated though!

PR #38, opened 2015-11-04
bind additional filesystem APIs

danburkert:


alexcrichton:
Travis seems to be taking its sweet time on this one...

danburkert:
@alexcrichton I didn't see the ctest module last night, so I've been fixing some issues (mostly including extra headers) and force pushing, which typically confuses travis temporarily.  Hopefully it will settle down soon.

alexcrichton:
Oh good point! This all looks good to me, thanks!

I'll merge once CI is green

alexcrichton:
Looks like some of the defined constants for linux don't exist? I'm not entirely sure where they come from, but it could mean that either a header isn't included or the CI machines are a little old?

danburkert:
Looking into the failures.  Some of them are a little strange, like it failing on linux nightly but not on 1.0 with missing header declarations.  I'm going to keep pushing to this branch since this is really the only way I have of testing against android, musl, arm, etc.

danburkert:
OK I think this is ready to go.  I had to remove some of the APIs from android and linux mips, as well as cut back the set of Linux fallocate flags.

PR #39, opened 2015-11-04
openbsd port

semarie:
- remove multiple definition of stack_t
- make types resolv correctly
- correct target definitions
  - dragonfly is already defined as freebsdlike
  - bitrig is an openbsdlike
- imports should be public

the PR makes libc build on openbsd


alexcrichton:
Awesome, thanks!

Do you know if there's any way that we can test this on Travis? I'd love to prevent this from regressing!

PR #40, opened 2015-11-04
Fix forgotten ::size_t for freebsd

myfreeweb:
Compiles and passes tests on FreeBSD again after this change.


alexcrichton:
Thanks!

PR #50, opened 2015-11-07
sockaddr_nl for netlink sockets

polachok:


polachok:
I have no idea why windows failed, but linux requires cfg.header("linux/netlink.h"), I guess?

alexcrichton:
Looks like that header is missing on MUSL?

polachok:
Yes. I guess it wouldn't make sense to add it then, and I should go nix-rust route.

alexcrichton:
Oh if you want you can feel free to just  not add this to musl just yet, it looks like it doesn't even exist in the headers there so may be fine to not add?

polachok:
Well, my initial plan was to add it to libc and then build on it in nix-rust. Absence in musl means that adding 'use libc::sockaddr_nl' to nix will break nix portability. I guess I could wrap it in #[cfg(...)] but that would lead to code duplication.

The header in question is not actually a glibc header, but comes from the linux tree (see source of http://packages.ubuntu.com/trusty/linux-libc-dev), which means its contents is probably out of scope for this crate.

alexcrichton:
Ah ok, I do think it'd still be in scope for this crate, and you can just ignore testing it on musl in that case by modifying `libc-test/build.rs`

alexcrichton:
Looks like there's still some failing CI builds?

polachok:
Yes. `musl-gcc` doesn't append -I/usr/include by default where the header is found.
One way to work around (from musl [FAQ](http://wiki.musl-libc.org/wiki/FAQ)) is to install headers separately:
```
wget -q http://ftp.barfooze.de/pub/sabotage/tarballs/kernel-headers-3.12.6-1.tar.xz -O - | tar xvJf - && cd kernel-headers-3.12.6-1/ && make ARCH=x86_64 prefix=/ DESTDIR=/opt/musl-kernel-headers install
```
and add `-I /opt/musl-kernel-headers/include` to gcc invocation.

```
root@8bbd60885895:/# cat test.c 
#include <linux/netlink.h>

int main() {
}
root@8bbd60885895:/# musl-gcc test.c 
test.c:1:27: fatal error: linux/netlink.h: No such file or directory
 #include <linux/netlink.h>
                           ^
compilation terminated.
root@8bbd60885895:/# musl-gcc -I/opt/musl-kernel-headers/include test.c 
root@8bbd60885895:/# 
```

```musl-gcc -I/usr/include -I/usr/include/x86_64-linux-gnu  test.c``` works too, but looks like a source of future problems.

Or we can just disable the test on musl.
What do you think?

**Comment by alexcrichton** [comment]
Ah it's probably easiest to just disable the test on musl, it's a kernel header verified elsewhere and you can just leave a comment to that effect.

**Comment by polachok** [comment]
I think I need a little help here, how do i skip a struct test? I can't find skip_struct option in ctest.

**Comment by alexcrichton** [comment]
Oh oops! I thought that would have already existed... I just added it to ctest though, so you should be good to use it!

**Comment by polachok** [comment]
It seems like there's some std/rustc mismatch on travis:
```
   Compiling libc v0.2.2 (file:///home/travis/build/rust-lang-nursery/libc/libc-test)

src/lib.rs:1:1: 1:1 error: the crate `std` has been compiled with rustc 1.6.0-nightly (ad3bd1b46 2015-11-11), which is incompatible with this version of rustc [E0514]

src/lib.rs:1 // Copyright 2012-2015 The Rust Project Developers. See the COPYRIGHT
```

alexcrichton:
Ah those should in theory just be intermittent, if you rebase and re-push it should work just fine!

polachok:
Omg it finally passed!

alexcrichton:
:+1:

PR #53, opened 2015-11-10
Fix undeclared type name errors when targeting android aarch64.

romanb:
Resolves `use of undeclared type name` errors for `sighandler_t` and `sigset_t` when compiling for `aarch64-linux-android`.

### DIFF
--- a/src/unix/notbsd/android/b64.rs
+++ b/src/unix/notbsd/android/b64.rs
@@ -1,8 +1,8 @@
 s! {
     pub struct sigaction {
         pub sa_flags: ::c_uint,
-        pub sa_sigaction: sighandler_t,
-        pub sa_mask: sigset_t,
+        pub sa_sigaction: ::sighandler_t,
+        pub sa_mask: ::sigset_t,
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks @romanb! I'm trying out this new "gate master on checks passing" thing GitHub has, and apparently that means that you can only merge one PR at a time! If you rebase this CI looks like it'll pass and I'll merge.

As a side note, do you know if there's a way to run binaries for aarch64-linux-android? I'd love to set up automation here to ensure it keeps building and it's correct.

romanb:
> As a side note, do you know if there's a way to run binaries for aarch64-linux-android? I'd love to set up automation here to ensure it keeps building and it's correct.

That would indeed be great but I've only recently started experimenting with cross-compiling for `aarch64-linux-android` and haven't actually used or tested the resulting binaries yet, so I don't know of a good way for automated testing yet. I only happened to run into these compile errors when upgrading `libc` for the Android library I'm building.

alexcrichton:
Ah ok, well if something ends up working out just let me know!

PR #56, opened 2015-11-11
Fix SIGSTKSZ for DragonFly

mneumann:
TODO: Add the same for FreeBSD.


alexcrichton:
Could you move this into `freebsdlike/dragonfly.rs`? That should avoid the extra `#[cfg]` here.

mneumann:
Good idea. Here we go: https://github.com/rust-lang-nursery/libc/pull/60

PR #66, opened 2015-11-20
Link to librt on non-apple platforms

alexcrichton:


.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: rust
 sudo: required
+dist: trusty
```
sfackler: Is this an unrelated change?

alexcrichton: Nah apparently this was needed to get CI to pass, I guess Travis updated some GCE images to 12.04 instead of 14.04?  Not entirely sure why some of the builders went back in time.

nagisa:
This breaks compilation on android. We should really be using build.rs here too.

PR #67, opened 2015-11-20
Netlink constants

polachok:
Let's see what ci says


alexcrichton:
Some of the failures on nightly I think may be fixed on rebasing, otherwise looks like there may be some other legit failures.

polachok:
SOL_NETLINK doesn't seem popular in headers yet, so I removed it.

PR #73, opened 2015-11-23
Remove `#[staged_api]`

petrochenkov:
Part of https://github.com/rust-lang/rust/pull/30015

I'm not sure how versions of rustc and libc correspond to each other now, so I used `custom_attribute` instead of `stage0`, i.e. libc can be used with any compiler version supporting or not supporting `staged_api`


alexcrichton:
lgtm, will merge once CI agrees

PR #79, opened 2015-11-26
POSIX sched_* APIs

polachok:
Let's see what CI says. Will add constants later.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -13,6 +13,7 @@ fn main() {
     let apple = target.contains("apple");
     let musl = target.contains("musl");
     let freebsd = target.contains("freebsd");
+    let mips = target.contains("mips");
     let bsdlike = freebsd || apple;
     let mut cfg = ctest::TestGenerator::new();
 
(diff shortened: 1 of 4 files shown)


alexcrichton:
Looks like the definition of `sched_param` may be tweaked among various platforms, and perhaps the functions could be moved to `src/unix/notbsd/mod.rs` to auto-exclude OSX? I'm also not sure if these exist in the FreeBSD/OpenBSD-like systems.

polachok:
It [exists](https://www.freebsd.org/cgi/man.cgi?query=sched_getscheduler&sektion=2) in freebsd and dragonfly. Not OpenBSD, though.

alexcrichton:
Ah ok, perhaps the definitions could be copied between `src/unix/notbsd/mod.rs` and `src/unix/bsd/freebsdlike/mod.rs`?

polachok:
Damn it, musl, [what's wrong with you](https://fossies.org/dox/musl-1.0.5/structsched__param.html)?

alexcrichton:
Looks like MIPS is a bit funny for Linux?

polachok:
I think it's just super old glibc in CI. 2.11? I have 2.15 on my ubuntu 12.04.

alexcrichton:
Ah yeah that may be the case, unfortunately the mips compiler is super old I think.

You can add exceptions to build.rs for now to get this through if you want saying "if we're mips linux don't check this value"

polachok:
Ok, I did that. Also added sched_get_priority_*

alexcrichton:
Thanks!

PR #80, opened 2015-11-26
unbreak openbsd and bitrig due to linking to rt

semarie:
openbsd and bitrig don't have librt. So do not try to link with.

cc @alexcrichton @dhuseby


semarie:
it should corrects the build failure on rust for bitrig and openbsd (when pushed to rustc)

alexcrichton:
Thanks!

PR #83, opened 2015-11-30
Add umask (POSIX.1-2001)

bossmc:
Quite likely this won't work on Windows, but I'll let CI work out the specifics.


bossmc:
Cool, CI passed on everything but windows, I've moved the definition into the `unix` module.

alexcrichton:
Thanks!

PR #84, opened 2015-11-30
Add RTLD_NEXT and RTLD_DEFAULT

dongz9:
This change adds RTLD_NEXT and RTLD_DEFAULT to libc. These two constants are useful when calling dlsym.


alexcrichton:
Thanks!

PR #86, opened 2015-11-30
Add rumprun CI and get tests passing

alexcrichton:
I was curious to see how we could get rumprun running over Thanksgiving, and it turns out it ended up working! I ended up having to fix a ton of definitions as well, and unfortunately I haven't been able to verify OpenBSD or Bitrig, so they may have broke as a result of this as well.

cc @gandro

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ env:
 matrix:
   include:
     - os: linux
-      env: TARGET=arm-linux-androideabi
+      env: TARGET=arm-linux-androideabi DOCKER=alexcrichton/rust-slave-android:2015-10-21
       rust: nightly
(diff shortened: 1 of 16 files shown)


gandro:
This is awesome. Looking at the code, I think we could use `target_os = "netbsd"` instead of `feature = "rumprun"`, since they are basically the same from libc's perspective. I hope to provide NetBSD CI as well at some point. I was working on it this weekend, but I ran out of time.

alexcrichton:
Yeah I've been meaning to get a NetBSD VM up and running to verify that these definitions are the same on NetBSD (there's a bunch of different ones), but I'll take your word for it for now!

PR #70, opened 2015-11-22
Add SO_REUSEPORT for Android

SimonSapin:
The `net2` crate currently fails to cross-compile to Android with ``error: unresolved name `libc::SO_REUSEPORT` ``

According to http://stackoverflow.com/a/14388707/1162888, Android is essentially the same as Linux for the purpose of socket features support, and the Linux kernel has supported `SO_REUSEPORT` since version 3.9.

According to http://android.stackexchange.com/a/51656, Android has had more recent kernels than that since version 4.4 Kit Kat.

(This blocks https://github.com/servo/servo/pull/8446.)

PR #87, opened 2015-12-01
Add getxattr()/setxattr() variations

polachok:


polachok:
It seems like android has [xattr.h](https://github.com/android/platform_bionic/blob/master/libc/include/sys/xattr.h) since 2012?

alexcrichton:
It could be because we're using an older android toolchain which doesn't have these headers just yet. The CI image we're using is attempting to be upgraded in #70 but Travis apparently doesn't like the new image...

alexcrichton:
Yeah I'm fine moving these to just linux, but looks like the header doesn't exist on android either?

Also could you be sure to format the definitions to 80 characters?

polachok:
Ok, done

alexcrichton:
Thanks!

PR #88, opened 2015-12-01
Update the android CI image

alexcrichton:
It's got a new toolchain which should hopefully support some more modern
definitions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ env:
 matrix:
   include:
     - os: linux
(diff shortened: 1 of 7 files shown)


alexcrichton:
Looks like the last run on #70 actually passed but failed some CI for some definitions, so trying to get an isolated log here.

alexcrichton:
Ok, I've built Servo from scratch with this libc and got no build errors, so although some functions have changed I don't think that they will cause breakage in practice. Along those lines I'm going to merge this as this doesn't have many ABI implications where it wasn't already a problem and this should enable us to add newer constants and such from Android and auto-verify them.

PR #89, opened 2015-12-02
Add dirent DT_* definitions

alexcrichton:


alexcrichton:
Wow, I'm surprised these actually have the same values across OSes. That's... fascinating!

PR #92, opened 2015-12-04
Add Linux constants for fcntl commands

nham:
I took these values from https://github.com/lattera/glibc/blob/a2f34833b1042d5d8eeb263b4cf4caaea138c4ad/sysdeps/unix/sysv/linux/x86/bits/fcntl.h  I'm not sure if they should go in an x86-specific module instead.


alexcrichton:
Looks like there's some duplicate names that make this fail to compile, you may also want to try adding them at `src/unix/mod.rs` and see what happens with CI (they may be defined on OSX/bsds as well)

alexcrichton:
Looks like the ever-illustrious mips is causing problems again :(

alexcrichton:
Oh I also just realized that these should all be included in #93, so I'm gonna close this in favor of that (it should be close to landing in theory)

PR #93, opened 2015-12-11
Add a load of constants/definitions from nix

alexcrichton:
This is a blanket import of lots of constants and function from nix-rust

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -9,6 +9,7 @@ pub type gid_t = u32;
 pub type in_addr_t = u32;
 pub type in_port_t = u16;
 pub type sighandler_t = ::size_t;
+pub type cc_t = ::c_uchar;
 
 pub enum DIR {}
 
@@ -85,6 +86,11 @@ s! {
         pub h_length: ::c_int,
         pub h_addr_list: *mut *mut ::c_char,
     }
+
+    pub struct iovec {
+        pub iov_base: *mut ::c_void,
+        pub iov_len: ::size_t,
+    }
 }
 
 pub const WNOHANG: ::c_int = 1;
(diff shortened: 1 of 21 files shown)


alexcrichton:
cc @carllerche, although this will likely take awhile to land

src/unix/mod.rs
```diff
@@ -100,6 +106,47 @@ pub const DT_REG: u8 = 8;
...
+pub static CLONE_CHILD_CLEARTID: ::c_uint = 0;
+pub static CLONE_DETACHED: ::c_uint = 0;
+pub static CLONE_UNTRACED: ::c_uint = 0;
+pub static CLONE_CHILD_SETTID: ::c_uint = 0;
+pub static CLONE_NEWUTS: ::c_uint = 0;
+pub static CLONE_NEWIPC: ::c_uint = 0;
+pub static CLONE_NEWUSER: ::c_uint = 0;
+pub static CLONE_NEWPID: ::c_uint = 0;
+pub static CLONE_NEWNET: ::c_uint = 0;
+pub static CLONE_IO: ::c_uint = 0;
```
polachok: clone() is linux specific call

alexcrichton: d'oh, looks like the usage of `pub static` instead of `pub const` caused these to all subvert CI, causing mayhem!

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -206,6 +244,65 @@ pub const RUSAGE_CHILDREN: ::c_int = -1;
...
+pub const MSDOS_SUPER_MAGIC: ::c_long = 0x00004d44;
+pub const NCP_SUPER_MAGIC: ::c_long = 0x0000564c;
+pub const NFS_SUPER_MAGIC: ::c_long = 0x00006969;
+pub const OPENPROM_SUPER_MAGIC: ::c_long = 0x00009fa1;
+pub const PROC_SUPER_MAGIC: ::c_long = 0x00009fa0;
+pub const QNX4_SUPER_MAGIC: ::c_long = 0x0000002f;
+pub const REISERFS_SUPER_MAGIC: ::c_long = 0x52654973;
+pub const SMB_SUPER_MAGIC: ::c_long = 0x0000517b;
+pub const TMPFS_MAGIC: ::c_long = 0x01021994;
+pub const USBDEVICE_SUPER_MAGIC: ::c_long = 0x00009fa2;
```
polachok: This block of constants is repeated 3 times with same values, wouldn't it be better to move it up into notbsd?

alexcrichton: Ah that's because these constants weren't mention on MUSL, so they had to be omitted

PR #99, opened 2015-12-15
Feature/ioctl

derekdreery:
Add in the linux/unix api for ioctl.

I say linux/unix because I'm not sure how general these functions/structs are. I'm pretty sure at least some of them are more general than just linux.

Also I'd appreciate someone to look at the types I've used - they are just #defines in the corresponding c code so it's not obvious what to use. What I need to do (and haven't done yet) is look at which functions these values are used for and look at what types those functions expect.

So in summary this isn't ready for merge but hopefully it will be! :)

### DIFF
--- a/src/unix/notbsd/linux/other/mod.rs
+++ b/src/unix/notbsd/linux/other/mod.rs
@@ -634,12 +634,6 @@ macro_rules! _IOC {
     }}
 }
 
-macro_rules! _IOC_TYPECHECK {
-    ($t:ty) => {{
-        size_of::<$t>() as ::c_uint
-    }}
-}
-
 // used to create numbers
 macro_rules! _IO {
     ($t:expr, $nr:expr) => {{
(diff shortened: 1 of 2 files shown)


src/unix/notbsd/linux/other/mod.rs
```diff
@@ -32,8 +36,247 @@ s! {
         __unused4: *mut ::c_void,
         __unused5: *mut ::c_void,
     }
+
+    // (bits/ioctl-types.h)
```
alexcrichton: It's ok to omit all the comments here and there, these files aren't really meant for general development as they're just reflections of what's found on the system. Any documentation should be found in local man pages and whatnot generally

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -206,6 +449,378 @@ pub const RUSAGE_CHILDREN: ::c_int = -1;
...
+pub const _IOC_NRSHIFT: u8 = 0;
+pub const _IOC_TYPESHIFT: u8 = _IOC_NRSHIFT + _IOC_NRBITS;
+pub const _IOC_SIZESHIFT: u8 = _IOC_TYPESHIFT + _IOC_TYPEBITS;
+pub const _IOC_DIRSHIFT: u8 = _IOC_SIZESHIFT + _IOC_SIZEBITS;
+
+pub const _IOC_NONE: ::c_uint = 0;
+pub const _IOC_WRITE: ::c_uint = 1;
+pub const _IOC_READ: ::c_uint = 2;
+
+macro_rules! _IOC {
```
alexcrichton: For now we're actually avoiding these kinds of macros in the repo, it's fine to just inline all values as bit values.

alexcrichton:
Awesome, thanks! If you're feeling adventurous you could try moving these definitions higher up in the chain (e.g. to encompass platforms like MUSL, MIPS linux, Android, and maybe even OSX. The CI should tell you whenever a def'n is wrong :)

polachok:
@derekdreery 
I think you missed adding include directives to `libc-test/build.rs`

derekdreery:
Thanks @polachok - I didn't totally understand the ci process - I'll add them in, along with experimenting with how general these constants are. I'm pretty sure they are the same on at least bsd (or at least some of them)

src/unix/notbsd/linux/other/mod.rs
derekdreery: Should I remove them now they're in?

alexcrichton: Yeah it's fine to just remove the comments here and below

derekdreery:
This PR has gone a bit stale now - some of the constants have appeared elsewhere - I need to go though it again if I'm going to get it committed.

alexcrichton:
Yeah I think some of these may have been added elsewhere in the meantime, but a rebase should do the trick!

PR #101, opened 2015-12-16
Expose functionality to get peer credentials for a Unix socket on Linux

maghoff:
Caveat: I put this rather randomly in `unix/notbsd/linux/other/mod.rs`. I know nothing about mips and musl, and I have no idea if this thing also applies to Android. I would guess "yes", but I don't know how to go about testing that assumption. Superficial googling seems to indicate that the constant and struct in question are present in android linux.

Further work: There is a roughly corresponding `getsockopt` for freebsd and osx (and possibly more bsds?) called `LOCAL_PEERCRED`. It would be neat to have that supported as well.

### DIFF
--- a/src/unix/notbsd/linux/other/mod.rs
+++ b/src/unix/notbsd/linux/other/mod.rs
@@ -32,6 +32,12 @@ s! {
         __unused4: *mut ::c_void,
         __unused5: *mut ::c_void,
     }
+
+    pub struct ucred {
+        pub pid: ::pid_t,
+        pub uid: ::uid_t,
+        pub gid: ::gid_t,
+    }
 }
 
 pub const RLIMIT_RSS: ::c_int = 5;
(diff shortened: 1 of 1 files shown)


maghoff:
Usage example:

    extern crate libc;
    extern crate unix_socket;

    use std::mem;
    use std::io::{self, Write};
    use std::os::unix::io::AsRawFd;
    use unix_socket::{UnixStream, UnixListener};

    fn get_peer_credentials(stream: &UnixStream) -> io::Result<libc::ucred> {
        let fd = stream.as_raw_fd();

        unsafe {
            let mut credentials: libc::ucred = mem::zeroed();
            let mut size = mem::size_of::<libc::ucred>() as libc::socklen_t;
            match libc::getsockopt(
                    fd,
                    libc::SOL_SOCKET,
                    libc::SO_PEERCRED,
                    &mut credentials as *mut _ as *mut _,
                    &mut size as *mut _ as *mut _
            ) {
                val if val < 0 => Err(io::Error::last_os_error()),
                _ => Ok(credentials)
            }
        }
    }

    fn main() {
        let listener = UnixListener::bind("socket").unwrap();
        println!("Connect to `socket` with for example: nc -U socket");

        for maybe_stream in listener.incoming() {
            let stream = maybe_stream.unwrap();
            let ucred = get_peer_credentials(&stream).unwrap();
            println!("Connected with user {} from process {}", ucred.uid, ucred.pid);
            writeln!(&stream, "Hello {}!", ucred.uid).unwrap();
        }
    }

maghoff:
To the failing AppVeyor build, this fails similarly to how it fails on master (doesn't find kernel32.lib), which is probably, and should definitely be, unrelated to this pull request :) (But you probably already realized)

alexcrichton:
Thanks!

PR #102, opened 2015-12-16
Add binding for memrchr on Linux

fhahn:
This PR adds a binding for `memrchr` on Linux and Free, Open and NetBSD. 

I hope that's okay, `memrchr` should be included in recent versions of  [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=memrchr&sektion=3&manpath=FreeBSD+7.1-RELEASE), [OpenBSD](http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man3/memchr.3?query=memrchr&sec=3) and [NetBSD](http://man.netbsd.org/7.0/usr/share/man/html3/memrchr.htm)

### DIFF
--- a/src/unix/bsd/freebsdlike/mod.rs
+++ b/src/unix/bsd/freebsdlike/mod.rs
@@ -575,6 +575,7 @@ extern {
                            len: ::off_t) -> ::c_int;
     pub fn sched_setscheduler(pid: ::pid_t, policy: ::c_int, param: *const sched_param) -> ::c_int;
     pub fn sched_getscheduler(pid: ::pid_t) -> ::c_int;
+    pub fn memrchr(cx: *const ::c_void, c: ::c_int, n: ::size_t) -> *mut ::c_void;
 }
 
 cfg_if! {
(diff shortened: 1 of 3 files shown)


fhahn:
I just read the reame and realized  I have to update this PR. Will do it later.

alexcrichton:
Ah yeah currently this lib avoids use of `#[cfg]` as much as possible, but you should be able to just include this in the corresponding platform module.

You may also want to just try adding this to `src/unix/notbsd/mod.rs` as well as the `freebsdlike/mod.rs` and `openbsdlike/mod.rs` files, as that should cover android as well.

src/unix/notbsd/linux/mod.rs
```diff
@@ -508,6 +508,7 @@ extern {
     pub fn unshare(flags: ::c_int) -> ::c_int;
     pub fn syscall(num: ::c_long, ...) -> ::c_long;
     pub fn sethostname(name: *const ::c_char, len: ::size_t) -> ::c_int;
+    pub fn memrchr(cx: *const ::c_void, c: ::c_int, n: ::size_t) -> *mut ::c_void;
```
alexcrichton: This ends up conflicting with the definition in `notbsd/mod.rs` as well, unfortunately.

fhahn: Thanks, I've updated the commit.

alexcrichton:
I think that github is also gonna require a rebase before merging this

fhahn:
I just rebased the commit

fhahn:
The only failing build on travis seems android related and it fails with `unexpected EOF`, which does not seem related to my change (https://travis-ci.org/rust-lang-nursery/libc/jobs/97329648)

alexcrichton:
Restarting the build to make sure

PR #107, opened 2015-12-18
Add emscripten support

brson:
The triple is asmjs-unknown-emscripten, target_arch = "asmjs", target_os = "emscripten".


brson:
cc @kripken This adds emscripten to rust's C bindings. Just FYI, not much interesting here.

alexcrichton:
lgtm, although the CI may barf on trying to generate docs for the emscripten triple if the compiler doesn't recognize it (the triples to document are auto-scraped from the doc attributes in the crate at the top).

PR #108, opened 2015-12-18
Add getentropy(2)'s syscall number as a const

mmcco:
I'm not convinced that the NR_* naming scheme is ideal, but it's pretty
good.

I didn't recognize any other syscall numbers in the file, so I started a
new section.


mmcco:
The build failures are on Windows and seem unrelated.

mmcco:
cc @semarie

alexcrichton:
Thanks!

PR #109, opened 2015-12-19
openbsd: map `si_addr` inside `struct siginfo_t`

semarie:
formally, under OpenBSD, `si_addr` is part of a C union. As we don't use
others members of the union inside `struct siginfo_t`, it should be safe
to directly map the field.

cc @mmcco

### DIFF
--- a/src/unix/bsd/openbsdlike/openbsd.rs
+++ b/src/unix/bsd/openbsdlike/openbsd.rs
@@ -99,8 +99,8 @@ s! {
         pub si_signo: ::c_int,
         pub si_code: ::c_int,
         pub si_errno: ::c_int,
-        __pad1: ::c_int,
-        __pad2: [u8; 240],
+        pub si_addr: *mut ::c_void,
+        __pad: [u8; 116],
     }
 }
 


alexcrichton:
Thanks! Due to the use of a union here in C it's a bit difficult to represent in Rust, so you may want to do something similar to what's currently defined on Linux. The size of the struct should be the same as C in both cases.

Better yet, if you could run `cargo run` inside of the `libc-test` directory on OpenBSD, it'll tell you whether you're right or wrong! (and I also suspect there may be a number of other failures)

semarie:
I have added padding to match the size of the C struct.
I will try to get a running cargo to do the test. But I suspect it will take some times...

semarie:
The pad value have been computed using C code:
```
#include <sys/siginfo.h>
#include <stdio.h>
#include <stdlib.h>
int main() {
  printf("pad = %llu\n", sizeof(siginfo_t) - 3*sizeof(int) - sizeof(void *));
  return 0;
}
```

siginfo definition: http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/sys/sys/siginfo.h?rev=1.11&content-type=text/x-cvsweb-markup

mmcco:
@semarie Looks like a stray ksh error message found its way into that commit...

semarie:
@mmco thanks for pointing it.

PR #110, opened 2015-12-23
Openbsd cleanup

semarie:
The following PR is the result of running `libc-test` under OpenBSD, and corrects all problems.

I have keeped separated commits for reviewing.

Some commits touch BSD globally (`getnameinfo` for example). In these cases, I have tried to review definitions in system code source.

Please note that this version of libc will break openbsd target for rustc. There are three breaking change:
  - OpenBSD don't have `HW_AVAILCPU`: it will break `num_cpus()` inside `src/libtest/lib.rs`
  - the `NR_GETENTROPY` constant don't exist under this name (it is `SYS_getentropy`). But I also remove it and declare the function `getentropy()` which should be a more stable API. The current use is inside `src/libstd/rand/os.rs`
  - the type of `siginfo_t.si_addr` isn't compatible with the current definition of  `siginfo_si_addr()` in `src/libstd/sys/unix/stack_overflow.rs`

For all these problems, I have pending patches.


libc-test/build.rs
```diff
@@ -200,6 +213,9 @@ fn main() {
     let target2 = target.clone();
     cfg.field_name(move |struct_, field| {
         match field {
+            "d_namelen" if openbsd && struct_ == "dirent" => "d_namlen".to_string(),
```
alexcrichton: Could this struct field be updated on openbsd to match the name in C? For renamings like this Rust shouldn't differ much

alexcrichton:
Thanks! I think it's fine to have some breakage here because this isn't a "tier 1 platform" of this library currently (e.g. there's no CI set up for it), but this is a great foundation to start from now!

semarie:
@alexcrichton I don't understand the failure for `mips-unknown-linux-gnu`

the previous one (6b84742) was ok, and adding 6c8a63a would make a linkage problem with `undefined reference to 'secure_getenv'` ?

alexcrichton:
Thanks @semarie! I'll work on getting this merged (mips failures are likely unrelated)

PR #111, opened 2015-12-24
Add NR_GETRANDOM syscall numbers

mmcco:
Pulled directly from src/libstd/rand/os.rs in the main rust repo. We can
delete them from there once this commit is added in rust's liblibc
submodule.

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -391,13 +391,13 @@ pub const CLONE_IO: ::c_uint = 0x80000000;
 
 // syscall numbers
 #[cfg(target_arch = "x86_64")]
-pub const NR_GETRANDOM: ::c_long = 318;
+pub const SYS_getrandom: ::c_long = 318;
 #[cfg(target_arch = "x86")]
-pub const NR_GETRANDOM: ::c_long = 355;
+pub const SYS_getrandom: ::c_long = 355;
 #[cfg(any(target_arch = "arm", target_arch = "powerpc"))]
-pub const NR_GETRANDOM: ::c_long = 384;
+pub const SYS_getrandom: ::c_long = 384;
 #[cfg(target_arch = "aarch64")]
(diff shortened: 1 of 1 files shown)


mmcco:
@alexcrichton What am I misunderstanding? I can't tell why the Linux builds are failing with NR_GETRANDOM undefined, despite the fact that they have supported architectures.

semarie:
The testsuite searchs for `NR_GETRANDOM` definition, but under Linux the name of the define is ` __NR_getrandom`.

So you need either use the proper name (and change rustc calls after that) or define exceptions in `libc-test/build.rs`.

nagisa:
NR_GETRANDOM should be `SYS_getrandom`.

mmcco:
I'll probably commit that to make sure it passes the build. However, I think @alexcrichton should confirm that that weird mixed-case naming scheme is acceptable here. I don't see it elsewhere in liblibc, so (as someone mentioned on IRC) it may be preferable to make an alias.

mmcco:
Only one nightly build on Linux - ARM, specifically - succeeded.

And now, I remember that `getrandom(2)` is brand new and therefore probably isn't even supported by the Travis workers' kernels. My mind was still off in OpenBSD land.

So, we can either leave things the way they are and wait for the API to mature, or we can add an exception to make liblibc take these values on faith.

alexcrichton:
Thanks @mmcco! Could this also split out the definitions into appropriate files instead of using `#[cfg]`? This'll also have to use the name that's available in C as opposed to one we made up.

mmcco:
Does anyone want to adopt this? The complexity of adding the definitions to libc compared with the current solution, combined with the fact that I'm not much of a Linux person, makes me lean toward leaving things as-is.

kamalmarhubi:
@mmcco I can adopt this and pick up #121.

mmcco:
@kamalmarhubi Sounds good to me. Thanks!

alexcrichton:

PR #117, opened 2015-12-28
DragonFly has different stat, dirent, clock_t, ino_t, nlink_t, blksize_t

mneumann:
Move FreeBSD's stat into freebsd_{x86,x86_64}.rs.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -14,10 +14,11 @@ fn main() {
     let apple = target.contains("apple");
     let musl = target.contains("musl");
     let freebsd = target.contains("freebsd");
+    let dragonfly = target.contains("dragonfly");
     let mips = target.contains("mips");
     let netbsd = target.contains("netbsd");
     let rumprun = target.contains("rumprun");
-    let bsdlike = freebsd || apple || netbsd;
+    let bsdlike = freebsd || apple || netbsd || dragonfly;
     let mut cfg = ctest::TestGenerator::new();
 
     // Pull in extra goodies on linux/mingw
@@ -96,7 +97,9 @@ fn main() {
     } else if !windows {
         cfg.header("glob.h");
         cfg.header("ifaddrs.h");
-        cfg.header("sys/quota.h");
+        if !dragonfly {
+            cfg.header("sys/quota.h");
+        }
         cfg.header("sys/statvfs.h");
 
         if !musl {
@@ -161,6 +164,11 @@ fn main() {
         cfg.header("sys/ioctl_compat.h");
     }
 
+    if dragonfly {
+        cfg.header("pthread_np.h");
+        cfg.header("sys/ioctl_compat.h");
+    }
+
     cfg.type_name(move |ty, is_struct| {
         match ty {
             // Just pass all these through, no need for a "struct" prefix
@@ -220,6 +228,9 @@ fn main() {
         match ty {
             // sighandler_t is crazy across platforms
             "sighandler_t" => true,
+
+            // does not exists on DragonFly
+            "fflags_t" if dragonfly => true,
 
             _ => false
         }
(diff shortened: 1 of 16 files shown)


mneumann:
Maybe, we should really split dragonfly and freebsd completely?

alexcrichton:
Thanks! I don't mind if freebsd/dragonfly are completely separated, they just seemed to have a lot of definitions in common from the original liblibc.

Could you change the organization here to `freebsd/{mod.rs,x86.rs,x86_64.rs}` as well? Other than that looks good to me!

mneumann:
@alexcrichton : Ok, will split it from FreeBSD and will come up with another pull request

mneumann:
@alexcrichton : Ok there we go. There is currently no other architecture than x86_64 for DragonFly, that's why I kept all code in x86_64.rs, but still use the cfg_if in mod.rs to be on the safe side when new architecture arrive.

libc-test/build.rs
```diff
@@ -221,6 +229,9 @@ fn main() {
             // sighandler_t is crazy across platforms
             "sighandler_t" => true,
 
+            // does not exists on DragonFly
+            "fflags_t" if dragonfly => true,
```
alexcrichton: Perhaps this could just be removed from the dragonfly definitions? It should be fine to push the typedef down lower into each respective module.

libc-test/build.rs
```diff
@@ -286,6 +297,18 @@ fn main() {
...
+            "GRPQUOTA" |
+            "Q_SYNC" |
+            "Q_QUOTAON" |
+            "Q_QUOTAOFF" |
+            "Q_GETQUOTA" |
+            "Q_SETQUOTA" |
+            "RLIMIT_NPTS" |
+            "RLIMIT_SWAP" |
+            "RUSAGE_THREAD" |
+            "MADV_PROTECT"  if dragonfly => true,
```
alexcrichton: How come these are ignored? Are they not defined on dragonfly?

libc-test/build.rs
```diff
@@ -333,6 +356,12 @@ fn main() {
             "ptrace" |
             "sigaltstack" if rumprun => true,
 
+            // uses size_t instead of socklen_t
+            "getnameinfo" if dragonfly => true,
```
alexcrichton: Hm this seems worrisome, shouldn't this be fixed?

libc-test/build.rs
```diff
@@ -333,6 +356,12 @@ fn main() {
...
+
+            // is defined elsewhere
+            "__dfly_error" if dragonfly => true,
```
alexcrichton: Oh wow, there's no C header file that returns this?

alexcrichton: Or wait, wasn't this part of `rust_builtin` which has since been removed?

src/unix/bsd/freebsdlike/dragonfly/mod.rs
```diff
@@ -0,0 +1,8 @@
+cfg_if! {
+    if #[cfg(target_arch = "x86_64")] {
```
alexcrichton: You can feel free to inline everything in this file if there's only one definition, but either way works fo rme

alexcrichton:
Looks good to me, thanks @mneumann!

libc-test/build.rs
mneumann: The quota's exist. The others not. I will add those that exist.

libc-test/build.rs
mneumann: I have to check this. DragonFly uses socklen_t, which probably is the same as size_t.

libc-test/build.rs
mneumann: exactly :). I am working on a patch to support thread-local externs, so that we don't need that anymore.

libc-test/build.rs
alexcrichton: Ah ok, feel free to just delete any constants from the libc crate on dragonfly that aren't actually present in the dragonfly libc

libc-test/build.rs
alexcrichton: Ah ok, in that case this function should actually be removed from libc

libc-test/build.rs
semarie: I think it is already addressed by #110. OpenBSD had same problem and I splitted `getnameinfo` implementations with `size_t` and `socklen_t`

mneumann:
As this pull-request became stale, I opened up a new one (https://github.com/rust-lang/libc/pull/194) which fixes all current issues. Closing this one.

PR #119, opened 2015-12-29
nlink_t depends on pointer width on android

fhahn:
`nlink_t` on android seems to depend on the pointer width. I noticed that when enabling [`os::raw::tests::unix`](https://github.com/rust-lang/rust/pull/30458/files#diff-c4f9e907b21efdcec1a2c091ff9ee641R92) which failed because the `nlink_t` sizes did not match.

I only found a pretty old definition of `[__kernel_nlink_t](http://osxr.org:8080/android/source/bionic/libc/kernel/arch-arm/asm/posix_types.h#0024)` which used `unsigned short` which should be 2 bytes on 32 bit architectures, and 4 bytes on 64 bit.

### DIFF
--- a/src/unix/notbsd/android/mod.rs
+++ b/src/unix/notbsd/android/mod.rs
@@ -13,7 +13,6 @@ pub type blkcnt_t = u32;
 pub type blksize_t = u32;
 pub type dev_t = u32;
 pub type mode_t = u16;
-pub type nlink_t = u32;
 pub type useconds_t = u32;
 pub type socklen_t = i32;
 pub type pthread_t = c_long;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Can you move these definitions into the respective modules instead of the `cfg_if!` statement as well?

alexcrichton:
It also looks like the currently tested android toolchain disagrees on the definition of `nlink_t`

alexcrichton:
I think that this may have been updated in more recent versions of Android, so I'm going to close, but thanks for the PR!

fhahn:
Sure, no problem. Didn't have a chance to look too closely in the last to weeks, but it indeed seems like that changed. The docs I found and linked are pretty old.

PR #122, opened 2015-12-29
Definitions for i686-unknown-linux-musl, arm and asmjs musl

brson:
I've tested x86 against C locally, but not arm or asmjs.
I added the arm definitions because asmjs's C is derived from arms.

Mysteriously, my locally-built musl does not contain a
definition for _SC_2_C_VERSION, so I just removed it.

Some of these conditions, particularly in notbsd, could be pushed further down the tree.


brson:
Looking at the build failures now.

brson:
Remaining failure looks bogus.

src/unix/notbsd/linux/musl/mod.rs
```diff
@@ -351,10 +308,18 @@ pub const SIG_UNBLOCK: ::c_int = 0x01;
 pub const FALLOC_FL_KEEP_SIZE: ::c_int = 0x01;
 pub const FALLOC_FL_PUNCH_HOLE: ::c_int = 0x02;
 
-pub const _SC_2_C_VERSION: ::c_int = 96;
-
+#[cfg(any(target_arch = "x86_64"))]
 pub const __SIZEOF_PTHREAD_RWLOCK_T: usize = 56;
+#[cfg(any(target_arch = "x86",
+          target_arch = "arm",
+          target_arch = "asmjs"))]
```
alexcrichton: Could these `#[cfg]` blocks be avoided by splitting this into a `musl/$arch.rs`-style configuration like with Linux?

alexcrichton:
Yeah I think build failures are unrelated, I'm prepping a PR that should bring them all in line

brson:
I've moved some of the definitions around as requested.

src/unix/notbsd/linux/musl/mod.rs
```diff
@@ -1,67 +1,34 @@
...
-        pub st_atime: ::time_t,
-        pub st_atime_nsec: ::c_long,
-        pub st_mtime: ::time_t,
-        pub st_mtime_nsec: ::c_long,
-        pub st_ctime: ::time_t,
-        pub st_ctime_nsec: ::c_long,
-        __reserved: [::c_long; 3],
-    }
+// This definition is slightly different on aarch64
+#[cfg(not(target_arch = "aarch64"))]
```
alexcrichton: Do we actually have aarch64 support?

Stylistically I also tend to prefer copying things around instead of using `#[cfg]` and/or `#[cfg(not)]` (even though it's a little wordier), but doesn't matter too much here, it can just be cleaned up later if it becomes a problem.

src/unix/notbsd/mod.rs
```diff
@@ -422,9 +426,28 @@ pub const CBAUD: ::tcflag_t = 0o0010017;
...
+#[cfg(all(target_env = "musl",
+          any(target_arch = "x86",
+              target_arch = "arm",
+              target_arch = "asmjs")))]
+pub const F_SETLK: ::c_int = 13;
+#[cfg(all(target_env = "musl",
+          any(target_arch = "x86",
+              target_arch = "arm",
+              target_arch = "asmjs")))]
+pub const F_SETLKW: ::c_int = 14;
```
alexcrichton: Could these `#[cfg]` annotations also be pushed down into the respective modules?

src/unix/notbsd/linux/musl/mod.rs
brson: No, but I noticed in passing these definitions are weird and just added these defensively. I can remove them.

brson:
Updated for most recent reviews.

src/unix/notbsd/mod.rs
```diff
@@ -108,7 +111,8 @@ s! {
         pub dli_saddr: *mut ::c_void,
     }
 
-    #[cfg_attr(any(target_arch = "x86", target_arch = "x86_64"),
+    #[cfg_attr(any(all(target_arch = "x86", not(target_env = "musl")),
+                       target_arch = "x86_64"),
```
alexcrichton: This line should be lined up with the `a` in `all`, right?

src/unix/notbsd/mod.rs
```diff
@@ -1,6 +1,9 @@
 use dox::mem;
 
+#[cfg(not(target_env = "musl"))]
 pub type rlim_t = c_ulong;
+#[cfg(target_env = "musl")]
+pub type rlim_t = ::c_ulonglong;
```
alexcrichton: Could these also be pushed down further (to avoid the `#[cfg]`)

src/unix/notbsd/linux/musl/mod.rs
```diff
@@ -1,67 +1,28 @@
...
-        pub st_ctime_nsec: ::c_long,
-        __reserved: [::c_long; 3],
-    }
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+pub type wchar_t = i32;
 
-    pub struct pthread_attr_t {
-        __size: [u64; 7]
-    }
+#[cfg(any(target_arch = "arm", target_arch = "asmjs"))]
```
alexcrichton: I'd personally prefer to push these `#[cfg]` into architecture modules under `b32`, but they're pretty minor so I could go either way

alexcrichton:
I'm fine merging whenever (just minor nits), but CI doesn't like two of the constants for MUSL

alexcrichton:
:package:

PR #123, opened 2015-12-30
Add ioctl constants

derekdreery:
Add some constants for ioctl calls.

### DIFF
--- a/src/unix/notbsd/android/mod.rs
+++ b/src/unix/notbsd/android/mod.rs
@@ -395,10 +395,7 @@ pub const O_NDELAY: ::c_int = 0x800;
 pub const NI_MAXHOST: ::size_t = 1025;
 
 pub const NCCS: usize = 19;
-pub const TCXONC: ::c_int = 0x540A;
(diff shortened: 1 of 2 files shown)


semarie:
I don't think the constants are the same for all unix.
For example, under OpenBSD, these constants have differents values.

derekdreery:
Will the ci pick that up?

derekdreery:
The CI has picked it up :)

derekdreery:
It seems that android uses int, whereas linux uses unsigned long for the __request in ioctl - so constants need to be different types for android and not android.

derekdreery:
I don't understand the mips error

alexcrichton:
Yeah the build failures are unrelated, and the constants for other platforms can be added in time (hard to do so without CI), thanks @derekdreery!

I'll work on merging this with a few others all at once

derekdreery:
:D

alexcrichton:
To link things up, landed in https://github.com/rust-lang-nursery/libc/pull/134

Issue #113, opened 2015-12-25
MS_RMT_MASK changed in Linux 4.0


diwic:
Hi,
In Linux < 4.0, the MS_RMT_MASK is correct,
in Linux >= 4.0, the [MS_RMT_MASK was changed to 0x2800051](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=0ae45f63d4ef8d8eeec49c7d8b44a1775fff13e8) to support the lazytime option.

I needed to change that in order to make the test run here. Probably this is the more compatible way as well, but I'm not sure if I can just change it without breaking tests running on older kernels...?

jimblandy:
Support for MS_LAZYTIME was added to libc [last June][1], along with the change to MS_RMT_MASK, but it seems like the glibc on the CI systems is older than that.

The flag is available from <linux/fs.h>, but I think we shouldn't be using that, for reasons I describe in pull request #127.

[1]: https://sourceware.org/git/?p=glibc.git;a=commit;h=31545c23277cd54a1edd41c85d8255fb589158e3

alexcrichton:
Ah yeah this should be fine to change, if the CI doesn't like it then it's fine to add an exception to handle special cases like this.

nodakai:
Memo: glibc 2.22 and musl 1.1.13 or later includes `MS_LAZYTIME` as `MS_RMT_MASK`

jimblandy:
> Memo: glibc 2.22 and musl 1.1.13 or later includes MS_LAZYTIME as MS_RMT_MASK

I think you mean "includes `MS_LAZYTIME` *in* `MS_RMT_MASK`"?

Right; glibc 2.22 was the first release to include those definitions.

posborne:
Hah, missed that a fix for this was already upstream.  This can probably be closed.

alexcrichton:
Ok!

PR #114, opened 2015-12-26
[WIP] Use docker for MIPS

polachok:
This image uses the same PPA as rust buildbot.

Dockerfile:
```
FROM ubuntu:15.10

RUN apt-get update
RUN apt-get install -y --force-yes --no-install-recommends \
        software-properties-common
RUN add-apt-repository ppa:angelsl/mips-cross
RUN apt-get update
RUN apt-get install -y --force-yes --no-install-recommends \
        gcc-5-mips-linux-gnu libc6-dev-mips-cross \
        gcc-5-mipsel-linux-gnu libc6-dev-mipsel-cross
RUN apt-get install -y --force-yes --no-install-recommends \
                build-essential qemu-user
RUN for f in `ls /usr/bin/mips*-linux-gnu-*-5`; do ln -vs $f `echo $f | sed -e 's/-5$//'`; done
```

I'll fix resulting fallout in following commits, reducing difference w/other archs.

Closes #100

PR #127, opened 2016-01-02
libc-test's C program should not #include <linux/fs.h> on Linux

jimblandy:
This is a fix for issue #113.

In `libc-test/build.rs`, the `main` function includes the directive `#include <linux/fs.h>` in the generated `all.c` file, but the correct header for system calls like `mount` is `<sys/mount.h>`; the `<linux/fs.h>` is just copied over from the kernel, and does not match the API offered by libc.

In particular:
- If you happen to include `<linux/fs.h>` before `<sys/mount.h>`, you'll get compilation errors, because the former defines the same constants using preprocessor macros that the latter defines as enum values. The generated `all.c` file just happens to include them in the other order.
- The `<linux/fs.h>` header includes a number of constants that are only meant to be used internal to the kernel, like `MS_NOSEC` and `MS_BORN`.
- The `<linux/fs.h>` header defines `MS_VERBOSE`, which is 1) a perverse name, because it silences some `printk` messages, rather than enabling them, and 2) is superseded by `MS_SILENT`, according to both the man page for `mount` and comments in the `<linux/fs.h>` header itself.

This pull request removes the libc crate's definitions for `MS_VERBOSE`, `MS_NOSEC`, and `MS_BORN`, and then removes the directive to include `<linux/fs.h>`.

This fixes issue #113 because the libc crate no longer sees the inappropriate-for-userspace definition of `MS_RMT_MASK`.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -149,7 +149,6 @@ fn main() {
         if !musl {
             cfg.header("linux/netlink.h");
             cfg.header("linux/magic.h");
-            cfg.header("linux/fs.h");
 
             if !mips {
                 cfg.header("linux/quota.h");
--- a/src/unix/notbsd/android/mod.rs
+++ b/src/unix/notbsd/android/mod.rs
@@ -411,7 +411,6 @@ pub const TOSTOP: ::tcflag_t = 0x00000100;
 pub const FLUSHO: ::tcflag_t = 0x00001000;
 
 pub const MS_RMT_MASK: ::c_ulong = 0x800051;
-pub const MS_VERBOSE: ::c_ulong = 0x8000;
 
 pub const ADFS_SUPER_MAGIC: ::c_long = 0x0000adf5;
 pub const AFFS_SUPER_MAGIC: ::c_long = 0x0000adff;
--- a/src/unix/notbsd/android/mod.rs
+++ b/src/unix/notbsd/android/mod.rs
@@ -410,8 +410,6 @@ pub const IEXTEN: ::tcflag_t = 0x00008000;
 pub const TOSTOP: ::tcflag_t = 0x00000100;
 pub const FLUSHO: ::tcflag_t = 0x00001000;
 
-pub const MS_RMT_MASK: ::c_ulong = 0x800051;
-
 pub const ADFS_SUPER_MAGIC: ::c_long = 0x0000adf5;
 pub const AFFS_SUPER_MAGIC: ::c_long = 0x0000adff;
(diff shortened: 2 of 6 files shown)


jimblandy:
Oh, this pull request builds on pull request #126.

jimblandy:
The MIPS build failed; it seems like the libc we're using in the MIPS QEMU doesn't have some of the newer constants. The mount(2) man page says:

> The  definitions of MS_DIRSYNC, MS_MOVE, MS_REC, MS_RELATIME, and MS_STRICTATIME were added to glibc headers in version 2.12.

Those are among the identifiers Travis reports missing.

polachok:
@jimblandy #114 may fix that

jimblandy:
@polachok: YAY

alexcrichton:
Could this be separated from https://github.com/rust-lang-nursery/libc/pull/126? (will help landing)

Also looks like there's a number of CI failures, not all of which are spurious

jimblandy:
> Could this be separated from #126? (will help landing)

Yes. (I had them on the same branch because I was trying to get a clean libc-test run, but in retrospect that's not the most helpful way to submit them.)

> Also looks like there's a number of CI failures, not all of which are spurious

The ones I see are either caused by being built on #126, or are those I mentioned in comment 8, and should be fixed by #114.

jimblandy:
I took a look at #114; this pull request will need revision if that one lands first, and vice versa.

In the process of updating libc for the MIPS image, #114 makes a bunch of changes to the libc crate itself, in order to get a clean libc-test run. But if we stop including <linux/fs.h> as I'm advocating in this PR, some of those changes will need to come back out.

If we want both PRs, I'm happy to revise this one after #114 is merged. I don't know how often the libc crate gets published to crates.io, but we should take care not to publish when only one has been merged, to avoid churn.

alexcrichton:
@jimblandy ah ok, I've merged #114 now (to fix all the mips build failures), and if you want to rebase this it should be all green and good to go.

Perhaps we should perform an audit and remove all inclusions of linux header files?

jimblandy:
The commit 53bf018f646f03303d6784e764dbd66eb13c4545 died on MIPS because the definition of MS_RMT_MASK had become the same as all the others. I fixed that, and added a commit that commons up all the MS_RMT_MASK definitions to a single one common to all Linux-kernel based targets, which makes more sense.

jimblandy:
That's a clean bill of health from Travis and Appveyor, and also on my Fedora laptop. A first!

alexcrichton:
Thanks!

PR #130, opened 2016-01-04
Offer a c_bool

solidsnack:


jimblandy:
Pull request #128 is similar to this.

brson:
Thanks! Since #128 comes with tests, and the typedef definition that I think we'll prefer, I'm going to close this in favor of that.

solidsnack:
It seems like #128 is a way's away from doing the right thing semantically, with respect to allowed values of bool.

jimblandy:
> It seems like #128 is a way's away from doing the right thing semantically, with respect to allowed values of bool.

How so? Both C and Rust constrain `bool` values to be either zero or one.

solidsnack:
@jimblandy Maybe I'm missing something but the definition of `c_bool` in #128 is `pub type c_bool = u8;`. Is there supposed to be a conversion somewhere?

jimblandy:
@solidsnack You are not missing something, I screwed up! Man... [re-pushes]

PR #131, opened 2016-01-06
Fix char for ARM with musl

joerg-krause:
For the ARM architecture a plain char represents an unsigned char value by
default.


alexcrichton:
Thanks! I think this'll be covered in https://github.com/rust-lang-nursery/libc/pull/122 so I'm gonna close in favor of that.

PR #135, opened 2016-01-12
Correct MIPS types of SA_* flags

alexcrichton:
Missed out correction from 6ac75c16b

### DIFF
--- a/src/unix/notbsd/linux/mips.rs
+++ b/src/unix/notbsd/linux/mips.rs
@@ -327,9 +327,9 @@ pub const __SIZEOF_PTHREAD_MUTEXATTR_T: usize = 4;
 pub const FIOCLEX: ::c_ulong = 0x6601;
 pub const FIONBIO: ::c_ulong = 0x667e;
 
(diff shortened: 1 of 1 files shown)


tamird:
@alexcrichton it seems that [`sigaction.sa_flags`](https://github.com/rust-lang-nursery/libc/blob/4d9ab7ebea11c8ce08101f993bb604e825c0d09b/src/unix/notbsd/linux/mips.rs#L66) and all the `SA_*` flags are signed here, but the linux headers disagree: https://github.com/torvalds/linux/blob/master/arch/mips/include/uapi/asm/signal.h#L105

Is this deviation intentional?

alexcrichton:
@tamird it is indeed, we updated to a newer toolchain of MIPS which has change the type to `int`. As to why the kernel and libc disagree, I'm not sure.

tamird:
Hm, seems odd. Another point of disagreement https://android.googlesource.com/platform/bionic/+/master/libc/kernel/uapi/asm-mips/asm/signal.h#96

PR #136, opened 2016-01-13
Skip C/Rust type checks for eventfd.

jimblandy:
This commit causes libc-test to not check the type of `libc::eventfd` against those from `<sys/eventfd.h>`, to avoid failing on GLIBC-based systems, as explained in the comments.

This commit does not change the type of `libc::eventfd`.


jimblandy:
So great to get clean test runs. Thanks very much!

PR #137, opened 2016-01-16
Add setns api to linux targets

mrjana:
Adds setns libc api to all linux targets

Signed-off-by: Jana Radhakrishnan <mrjana@gmail.com>


alexcrichton:
Thanks!

PR #138, opened 2016-01-17
Add Illumos support

nbaksalyar:
Hello!

This is a first pull request required to add support for a family of operating systems based on [Illumos](http://illumos.org) (an OpenSolaris fork), such as Joyent's [SmartOS](http://smartos.org), OmniOS, OpenIndiana, and others.

After this commit is merged, we can proceed to add support for Illumos to the main Rust repository.
The current version can be found here: https://github.com/nbaksalyar/rust/tree/solaris. It passes all tests (`2339 passed; 0 failed; 3 ignored; 0 measured` on the `run-pass` suite).

Please let me know if something is needed to be added or changed.

Thanks!

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -99,7 +99,6 @@ s! {
     }
 }
 
-pub const WNOHANG: ::c_int = 1;
 pub const SIG_DFL: sighandler_t = 0 as sighandler_t;
 pub const SIG_IGN: sighandler_t = 1 as sighandler_t;
 pub const SIG_ERR: sighandler_t = !0 as sighandler_t;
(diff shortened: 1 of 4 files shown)


alexcrichton:
Oh holy cow, nice! This looks awesome :)

Could you organize the `sunos.rs` file like others in the repository so it looks like:

* Type definitions
* Structure definitions
* Constants
* Rust-defined functions (e.g. inlined C maros)
* FFI function headers (e.g. the `extern` block)
* Further delegation via `cfg_if!` to other modules (may not be present yet for this module)

nbaksalyar:
Thanks!

Does it look better now? :)

src/unix/mod.rs
```diff
@@ -621,6 +620,7 @@ extern {
                     addrlen: *mut socklen_t) -> ::ssize_t;
     pub fn mkfifo(path: *const c_char, mode: mode_t) -> ::c_int;
 
+    #[cfg(not(target_os = "sunos"))]
```
alexcrichton: Could this function definition move directly into `bsd/mod.rs` and `notbsd/mod.rs`? We tend to avoid `#[cfg(not)]` if possible in libc currently

alexcrichton:
Looks good to me, just one last nit and then I'll merge once CI passes

alexcrichton:
Thanks!

nbaksalyar:
Awesome, thank you! :)

PR #139, opened 2016-01-18
FreeBSD and OpenBSD CI

alexcrichton:
This was the culmination of some experiments/yak shaving I was doing this weekend. I was curious how far we could push QEMU to the end-goal of running tests for essentially any OS and all targets, and ended up quite successful! With this framework we should be able to eventually add CI/testing support for all targets that liblibc will support. Yay tests!

cc @semarie, @dhuseby, @mneumann -- some points:

* I don't currently have images for DragonFly or Bitrig setup, but this should be easy to do.
* I just installed FreeBSD 10.2 from the vanilla installers, but should I use a different one instead? Will we get ABI issues if we match our definitions to FreeBSD 10.2 and try to run on some other version?
* I also installed OpenBSD 5.8, and it's the same question as FreeBSD, is an ok version to match our definitions to?
* Along the way I [added an OpenBSD Cargo snapshot](https://github.com/rust-lang/cargo/pull/2287)
* I'm using `pkg_add rust` on OpenBSD to get Rust 1.1 and `pkg install rust` on FreeBSD to get Rust 1.5. Seems to work at least!

cc @brson, @edunham -- the CI aspects here are... interesting! Wanna make sure y'all feel good about this strategy


dhuseby:
This is great.  I'll see if I can help with the bitrig and dragonfly stuff.

@alexcrichton one question, why did you have to add any freebsd definitions?  i thought i took care of that when i set up my build slaves.

alexcrichton:
Oh I didn't actually add any freebsd definitions, I just [had to fix up a few](https://github.com/alexcrichton/libc/commit/6fce1d590ba7b5c975d85a7a1ca90919ad52990c), the "removals" are actually just pushing the definitions lower down in the hierarchy into respective modules.

semarie:
for OpenBSD, yes 5.8 is a good target for now. It is the current stable release. And as cargo snapshot is linked under 5.8, it will work (but nowhere else).

maybe it could be reconsidered at some points, as it makes libc linked to a "static" target (there are no binary changes in `-stable`). So it will not permit to reflect upcoming changes in future stable, and could make libc stucks behind. But it could be seen, when breaking changes between `-stable` and `-current` version of OpenBSD will affect rust-libc.

alexcrichton:
@semarie hm, does that mean that binaries generated on 5.8 will not be able to run on 5.9? (or whatever the next version is). I'd also be quite interested in moving Cargo back as far as possible if need be!

semarie:
@alexcrichton due to libraries dependencies, yes.

```
$ readelf -d cargo-nightly-x86_64-unknown-openbsd/cargo/bin/cargo | grep NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libssl.so.35.0]
 0x0000000000000001 (NEEDED)             Shared library: [libcrypto.so.35.0]
 0x0000000000000001 (NEEDED)             Shared library: [libz.so.5.0]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.19.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.80.1]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.9.0]
```

This binary requires all these libraries to be present to run. And under 5.9 there are not here. The current versions for these libraries under `-current` are:

```
/usr/lib/libc.so.84.2
/usr/lib/libcrypto.so.37.0
/usr/lib/libm.so.9.0
/usr/lib/libpthread.so.20.1
/usr/lib/libssl.so.38.0
/usr/lib/libz.so.5.0
```

the current bootstrap for rustc deals with that by vendoring the libraries (as I have generated it in the same way that for bootstrapper for rust package under OpenBSD).

alexcrichton:
Interesting, I guess OpenBSD is just a system where you can't transfer binaries between versions? I wonder if we can link to system libraries without versions as anything we require from libc/libm/libpthread for example likely have a stable ABI.

brson:
Neat. r=me

PR #140, opened 2016-01-19
Add IP_TRANSPARENT socket option

jaksi:
Required for transparent proxying


alexcrichton:
Hm looks like we were having some CI trouble this morning, could you try rebasing and retriggering travis? I think a new nightly should be out by morning at least.

jaksi:
Did a rebase, but the nightly-related jobs still fail for the same reason.

alexcrichton:
Reran this morning, thanks!

PR #147, opened 2016-01-21
Use same type for all CLONE_ flags.

fiveop:
The `CLONE_` flags defined in `...notbsd/mod.rs` and `...notbsd/linux/mod.rs` use different types (`c_int` and `c_uint` respectively). This commit changes the type of the latter to `c_int` as well, since the implementation of `clone` of glibc takes an `int` (see `man 2 clone`).

Is there a reason, why not all of the flags are in `...notbsd/mod.rs`?

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -384,12 +384,12 @@ pub const EFD_SEMAPHORE: ::c_int = 0x1;
 
 pub const NCCS: usize = 32;
 
-pub const CLONE_NEWUTS: ::c_uint = 0x04000000;
-pub const CLONE_NEWIPC: ::c_uint = 0x08000000;
-pub const CLONE_NEWUSER: ::c_uint = 0x10000000;
-pub const CLONE_NEWPID: ::c_uint = 0x20000000;
-pub const CLONE_NEWNET: ::c_uint = 0x40000000;
-pub const CLONE_IO: ::c_uint = 0x80000000;
+pub const CLONE_NEWUTS: ::c_int = 0x04000000;
+pub const CLONE_NEWIPC: ::c_int = 0x08000000;
+pub const CLONE_NEWUSER: ::c_int = 0x10000000;
+pub const CLONE_NEWPID: ::c_int = 0x20000000;
+pub const CLONE_NEWNET: ::c_int = 0x40000000;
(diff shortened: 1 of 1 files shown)


brson:
Looks like this will be a breaking change.

alexcrichton:
@brson would you be ok merging this despite that? I'd be very willing to say that this will break no code because:

* These types accidentally disagree with the other constants in `src/unix/notbsd/mod.rs`
* If you were to call the function with these flags, you'd get a type error

Which means that any current usage has a cast, or they're not being used. Along those lines I would rather make this change and push a new version. If it does indeed cause breakage we can yank and republish and figure out another strategy, but I'm not sure it'll be worth keeping around this "wart" for too long.

brson:
@alexcrichton I think we should decide between the libs team what to do about this and the eventfd patch. I don't want to start a habit of publishing minor breakage casually.

alexcrichton:
Perhaps. I worry that it will be at least two weeks before we can triage this at which point it may be "even more too late". If you feel strongly then we can just reject these patches and say "oh well". I feel very strongly that a major version bump isn't worth it, and PRs like this are just making it more useful.

I wish we could typecheck constants but unfortunately we can't, so this is just a vector via which we're bound to make mistakes. I can't really envision a realistic scenario in which this causes breakage, however.

fiveop:
For what it's worth, I'd rather have a correct library, than the luxury of no breakage between versions (and code like `let flags = A as c_int | B;`).

I found this inconsistency because I was trying to convert the corresponding constants in nix-rust to a `bitflags!` type. Anyone who used the [`clone`](https://github.com/rust-lang-nursery/libc/blob/aea17ed1a6e87c310e333a92a51cb93a9812cb68/src/unix/notbsd/mod.rs#L611) function (or any of its siblings) with the wrongly typed constants would have had to cast them.

I'll continue creating pull request in cases such as these and eagerly await your decision.

brson:
@alexcrichton Ok, let's merge this and https://github.com/rust-lang-nursery/libc/pull/126. I agree in practice it's not going to break anything so there's no practical reason to hold it up. I'll think about how to incorporate this into our regression tracking.

alexcrichton:
Ok, I'm going to merge this with https://github.com/rust-lang-nursery/libc/pull/149 so we don't have to rebase all these PRs.

Issue #115, opened 2015-12-28
Support SOCK_NONBLOCK and SOCK_CLOEXEC


jamperry:
These values provide convenience when created sockets via `socket` to add non blocking and closing--on-exec flags as it saves calling `fcntl(2)`. 

I am happy to write a patch for this if you accept this issue. There are some gotchas because

`SOCK_NONBLOCK` != `O_NONBLOCK`
`SOCK_CLOEXEC` != `FD_CLOEXEC`

for certain architectures. 

This is due to socket flag state would be be overridden so it uses a different enumerate value.

jamperry:
Here's the MAN page for socket to give some background information: http://man7.org/linux/man-pages/man2/socket.2.html

alexcrichton:
Ah yeah I've definitely wanted these in the past, they're more than welcome here! May just take some hunting to find the BSD definitions, but those aren't critical to land.

polachok:
http://bxr.su/search?q=&defs=SOCK_CLOEXEC&refs=&path=&project=DragonFly&project=FreeBSD&project=NetBSD&project=OpenBSD

alexcrichton:
Oh wow, that's quite a useful link, thanks @polachok!

alexcrichton:
Closing as these were added in https://github.com/rust-lang/libc/pull/150, yay!

PR #150, opened 2016-01-23
Add SOCK_NONBLOCK & SOCK_CLOEXEC

jamperry:
I've added the Linux-specific flags for #115

It helps to support the flag that is used to create a non blocking socket in one kernel call instead of two.


alexcrichton:
Thanks!

PR #159, opened 2016-01-28
unix: Implement Eq, Hash, PartialEq for sockaddr_* types

kamalmarhubi:


kamalmarhubi:
Confused by the nightly x86_64 only failure: https://travis-ci.org/rust-lang-nursery/libc/jobs/105324244#L233

alexcrichton:
Thanks for the PR! Currently, however, the libc crate is intended to be as thin a wrapper as possible around the system libc, for example it only provides the guarantee that structs are `Copy` (which also forces a `Clone` impl). 

At this time we're not looking for higher level abstractions in libc (e.g. any other form of "Rust code" like `Eq` or `Hash`). That's in general left to higher level libraries.

kamalmarhubi:
Thanks for the explanation!

PR #163, opened 2016-01-29
musl: add support for mips/mipsel

japaric:
See first commit for details about the new module hierarchy for the musl module.

This has been `ctest`ed in qemu using a mipsel OpenWRT trunk image.

The commit history contains several commits where I've been fixing `ctest`s, but I can squash the commits before merging.

r? @alexcrichton

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -135,7 +135,7 @@ cfg_if! {
     if #[cfg(not(stdbuild))] {
         // cargo build, don't pull in anything extra as the libstd  dep
         // already pulls in all libs.
-    } else if #[cfg(target_env = "musl")] {
+    } else if #[cfg(target_env = "musl", not(target_arch = "mips"))] {
         #[link(name = "c", kind = "static")]
         extern {}
     } else if #[cfg(target_os = "emscripten")] {
--- a/src/unix/notbsd/linux/musl/mod.rs
+++ b/src/unix/notbsd/linux/musl/mod.rs
@@ -19,12 +19,6 @@ s! {
         _restorer: *mut ::c_void,
     }
 
-    pub struct stack_t {
-        pub ss_sp: *mut ::c_void,
-        pub ss_flags: ::c_int,
-        pub ss_size: ::size_t
-    }
-
     pub struct siginfo_t {
         pub si_signo: ::c_int,
         pub si_errno: ::c_int,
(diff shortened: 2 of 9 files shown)


src/lib.rs
```diff
@@ -27,9 +27,12 @@
 #![cfg_attr(all(target_os = "linux", target_arch = "arm"), doc(
     html_root_url = "https://doc.rust-lang.org/libc/arm-unknown-linux-gnueabihf"
 ))]
-#![cfg_attr(all(target_os = "linux", target_arch = "mips"), doc(
+#![cfg_attr(all(target_os = "linux", target_arch = "mips", target_env = "gnu"), doc(
     html_root_url = "https://doc.rust-lang.org/libc/mips-unknown-linux-gnu"
 ))]
+#![cfg_attr(all(target_os = "linux", target_arch = "mips", target_env = "musl"), doc(
```
alexcrichton: There's a match for `target_env = "musl"` below which may need the addition of `target_arch` as well

src/unix/mod.rs
```diff
@@ -135,7 +135,7 @@ cfg_if! {
     if #[cfg(not(stdbuild))] {
         // cargo build, don't pull in anything extra as the libstd  dep
         // already pulls in all libs.
-    } else if #[cfg(target_env = "musl")] {
+    } else if #[cfg(all(target_env = "musl", target_arch = "x86_64"))] {
```
alexcrichton: Shouldn't this linkage happen for mips as well?

src/unix/notbsd/linux/musl/mod.rs
```diff
@@ -0,0 +1,131 @@
+pub type clock_t = ::c_long;
+
+pub type blkcnt_t = i64;
+pub type fsblkcnt_t = u64;
+pub type fsfilcnt_t = u64;
+pub type ino_t = u64;
+pub type off_t = i64;
+pub type stat64 = ::stat;
```
alexcrichton: Up above it looked like `struct stat` and `struct stat64` had different definitions, was that just coincidental though?

alexcrichton:
This'll likely have tons of conflicts with https://github.com/rust-lang-nursery/libc/pull/122, but it's actually got enough scaffolding in place that it'd be easy enough to add i686/arm/asmjs in next. (cc @brson though)

I'd also be quite eager to slurp up the scripts you got to get the tests working, but that can likely wait until we've landed support in the compiler itself.

Thanks!

src/lib.rs
japaric: fixed

src/unix/mod.rs
japaric: With this cfg, x86-64-musl links statically to libc, but mips-musl links dynamically to libc.

src/unix/notbsd/linux/musl/mod.rs
japaric: The differences between stat and stat64 are two fields that changed from ino_t to ino64_t and blkcnt_t to blkcnt64_t. But in musl, for any arch, ino_t is always equal to ino64_t; same thing with blkcnt_t.

src/unix/mod.rs
alexcrichton: Hm interesting, that's an unusual difference between the two musl targets. In general linkage to libc is somewhat up-in-the-air (e.g. how it happens), but I guess musl+mips is small enough for now that this should be ok.

src/unix/notbsd/linux/musl/mod.rs
alexcrichton: Hm ok, in the interest of more verification though could this struct be a redefinition using the `foo64_t` types? That way ctest will auto-check the members of the struct, size, alignment, etc. Right now it doesn't handle type aliases to structs like it does struct declarations themselves.

japaric: Sure, I just have to undo one commit.

src/unix/mod.rs
japaric: From my point of view, x86_64-musl is the outlier here because it's the only target that links to libc statically :P.

PR #166, opened 2016-02-01
Define _SC_PAGE_SIZE constants

lambda:
According to POSIX, sysconf can take _SC_PAGESIZE or _SC_PAGE_SIZE,
which may be the same value.  This constant was defined in "apple" as
equal to _SC_PAGESIZE, but not any of the other platforms.

Add definitions for _SC_PAGE_SIZE for all platforms which had
_SC_PAGESIZE defined. On all platforms but Android (Bionic), they are
the same value; on Android _SC_PAGE_SIZE has a different value.

### DIFF
--- a/src/unix/notbsd/android/mod.rs
+++ b/src/unix/notbsd/android/mod.rs
@@ -196,6 +196,7 @@ pub const _SC_XOPEN_LEGACY: ::c_int = 36;
 pub const _SC_ATEXIT_MAX: ::c_int = 37;
 pub const _SC_IOV_MAX: ::c_int = 38;
 pub const _SC_PAGESIZE: ::c_int = 39;
+pub const _SC_PAGE_SIZE: ::c_int = 40;
 pub const _SC_XOPEN_UNIX: ::c_int = 41;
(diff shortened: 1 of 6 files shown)


alexcrichton:
Gah alas I merged https://github.com/rust-lang-nursery/libc/pull/167 before I got to this one, and unfortunately unlike Rust we don't have @bors to auto-rebase everything. Could you rebase and re-push? I'll merge once the CI passes there as well.

lambda:
Rebased

src/unix/notbsd/android/mod.rs
```diff
@@ -196,6 +196,7 @@ pub const _SC_XOPEN_LEGACY: ::c_int = 36;
 pub const _SC_ATEXIT_MAX: ::c_int = 37;
 pub const _SC_IOV_MAX: ::c_int = 38;
 pub const _SC_PAGESIZE: ::c_int = 39;
+pub const _SC_PAGE_SIZE: ::c_int = 40;
```
alexcrichton: android wat r u doin

PR #167, opened 2016-02-01
Add abort

lambda:
This is defined by the C standard to indicate abnormal program
terminaiton; it is defined to call raise(SIGABRT), and to not return.
This can be useful when panics are not an option, such as signal
handlers handling stack overflow.


alexcrichton:
Thanks!

PR #168, opened 2016-02-01
Use iOS nightlies instead of a pinned version

alexcrichton:


alexcrichton:
Looks like we weren't producing nightlies like I thought we were...

alexcrichton:
I *think* we have nightlies now, let's see if this works...

PR #170, opened 2016-02-04
unix: Add sendfile to platforms that support it

kamalmarhubi:


src/unix/bsd/freebsdlike/freebsd.rs
```diff
@@ -4,7 +4,16 @@ pub type fsfilcnt_t = ::uint64_t;
...
+pub const SF_MNOWAIT: ::c_int = 0x00000002;
+pub const SF_SYNC: ::c_int = 0x00000004;
+pub const SF_NOCACHE: ::c_int =   0x00000010;
 
 extern {
     pub fn __error() -> *mut ::c_int;
 }
+
+#[inline]
+pub fn SF_FLAGS(rh: ::c_int, flags: ::c_int) -> ::c_int {
```
alexcrichton: Having never used sendfile before, much less on freebsd, is this macro necessary for the usefulness of the function? Macros are unfortunately harder to validate, and I'd probably prefer to leave it if out if possible. If it's needed for inclusion, though, should be fine!

src/unix/bsd/apple/mod.rs
```diff
@@ -250,6 +250,12 @@ s! {
         pub l_type: ::c_short,
         pub l_whence: ::c_short,
     }
+    pub struct sf_hdtr {
```
alexcrichton: Typically each struct has a blank line between it and the previous one (here and below)

src/unix/notbsd/mod.rs
```diff
@@ -616,6 +616,7 @@ extern {
     pub fn fstatfs(fd: ::c_int, buf: *mut statfs) -> ::c_int;
     pub fn memrchr(cx: *const ::c_void, c: ::c_int, n: ::size_t) -> *mut ::c_void;
     pub fn syscall(num: ::c_long, ...) -> ::c_long;
+    pub fn sendfile(out_fd: ::c_int, in_fd: ::c_int, offset: *mut off_t, count: ::size_t) -> ::ssize_t;
```
alexcrichton: Could you be sure to wrap this to 80 chars? (surrounding style)

alexcrichton:
Thanks @kamalmarhubi! Looks good to me modulo some nitpicks, otherwise just waiting on some green CI

src/unix/bsd/freebsdlike/freebsd.rs
kamalmarhubi: It's actually brand new in FreeBSD: https://svnweb.freebsd.org/base?view=revision&revision=293439

I imagine it's not absolutely needed, as the man pages document the use of the upper and lower 16 bits for `flags`. I've removed it for now.

src/unix/bsd/apple/mod.rs
kamalmarhubi: Done.

kamalmarhubi:
Fixed issues. Accidentally pushed an intermediate change, so canceling this build would speed things up: https://travis-ci.org/rust-lang-nursery/libc/builds/107141122

kamalmarhubi:
@alexcrichton regarding verifying macros, at least for 32 bit values we could exhaustively check they are the same. It'd add at most a few seconds to the CI builds, which have a variance way above that anyway. I'd have to poke around the CI more to see how everything works, but do you think it makes sense?

kamalmarhubi:
Urg more `*const` use where it isn't actually in the signature. Another couple of hours...

kamalmarhubi:
Had missed an arg from FreeBSD. Confused by the Apple error, will look more tomorrow.

alexcrichton:
Ah yeah we can always add it later if we need to, I figure for now we could just take the conservative route. 

In terms of verification I'd just want to read the headers myself, but other than that it seems fine.

kamalmarhubi:
> Ah yeah we can always add it later if we need to, I figure for now we could just take the conservative route.
:+1:

kamalmarhubi:
I think freebsd is fixed with latest, but I have yet to take a look at Apple.

kamalmarhubi:
```
all.c:8115:54: error: use of undeclared identifier 'SF_NOCACHE'
            static int __test_const_SF_NOCACHE_val = SF_NOCACHE;
```

This is because `SF_NOCACHE` was introduced very recently and hasn't made it to any releases yet. I added an exception to CI for it.

alexcrichton:
Perhaps it could be elided for now? If it's so recent and not even released it may not be much use in practice anyway

kamalmarhubi:
That makes sense, removed for now.

kamalmarhubi:
For Apple I'm currently trying to get my olde laptoppe updated enough to be able to figure out what the generated code is saying.

kamalmarhubi:
Ah, figured out the issue:

```diff
diff --git a/src/unix/bsd/apple/mod.rs b/src/unix/bsd/apple/mod.rs
index e59ae50..708b198 100644
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -252,9 +252,9 @@ s! {
     }
 
     pub struct sf_hdtr {
-        pub headers: *const ::iovec,
+        pub headers: *mut ::iovec,
         pub hdr_cnt: ::c_int,
-        pub trailers: *const ::iovec,
+        pub trailers: *mut ::iovec,
         pub trl_cnt: ::c_int,
     }
 }
```

kamalmarhubi:
@alexcrichton what's the release policy for libc? It'd be great to get a release soon with `sendfile(2)` and the `splice(2)` et al for Linux. :-)

alexcrichton:
Looks like the signature may still be wrong for FreeBSD? And yeah I can push a new release soon as well, perhaps we can get the last few PRs right now merged first?

kamalmarhubi:
Yeah one more attempt. I need to get that FreeBSD thing to run locally...

Re release, I mostly was wondering if it's days or weeks, so no worries about extra PRs.

kamalmarhubi:
Ugh that one was missing full qualification of `c_int`.

PR #172, opened 2016-02-04
linux: Add splice, tee, vmsplice

kamalmarhubi:


alexcrichton:
Looks good to me, thanks! Just waiting on CI

kamalmarhubi:
Had to change some `*const` to `*mut` to get the CI to pass. Conceptually they are constant, but I guess that's up to a higher level than libc to enforce.

kamalmarhubi:
So I'm not sure what's going on here. On my system:

```
$ uname -r
4.3.0-1-amd64
$ printf "#define _GNU_SOURCE\n#include <fcntl.h>" | cpp - | grep -A1 vmsplice
extern ssize_t vmsplice (int __fdout, const struct iovec *__iov,
    size_t __count, unsigned int __flags);
```

But the man pages:

```
       ssize_t vmsplice(int fd, const struct iovec *iov,
                        unsigned long nr_segs, unsigned int flags);
```
―http://man7.org/linux/man-pages/man2/vmsplice.2.html

kamalmarhubi:
I've pushed a version with a `size_t` instead, will see what that does. Concerned I'll have to dig into headers unknown and trace some `ifdef` stuff.

@alexcrichton what's the best thing to do here?

kamalmarhubi:
Well that seems to have passed. But even the syscall's signature says `unsigned long`:

```
asmlinkage long sys_vmsplice(int fd, const struct iovec __user *iov,
			     unsigned long nr_segs, unsigned int flags);
```
―https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/include/linux/syscalls.h#n788

alexcrichton:
Ah yeah the man pages often differ from the headers slightly, for example `eventfd` is this way as well (man page says `unsigned int` and header says `int`). In terms of API the size of `unsigned long` and `size_t` is the same everywhere now (they're all just `usize`), so ABI wise we're fine.

Can you rebase this as well? Unfortunately github with branch protection only allows merging a fully rebased PR...

kamalmarhubi:
Rebased.

kamalmarhubi:
> Ah yeah the man pages often differ from the headers slightly, for example `eventfd` is this way as well (man page says `unsigned int` and header says `int`).

I guess this is ok because they're the same size?

alexcrichton:
Yeah I think it's ok because they're the same size. or presumably glibc does something to do the translation.

Looks like the OSX failures were spurious, so I restarted the build.

alexcrichton:
Unfortunately I can't seem to coerce travis into restarting one of the failed builds, could you try amending a commit and force-pushign?

kamalmarhubi:
Done.

kamalmarhubi:
@alexcrichton could you restart the mips job? it failed to install Rust. https://travis-ci.org/rust-lang-nursery/libc/jobs/107298557

PR #177, opened 2016-02-07
Add `posix_fadvise()` and related constants

### LINKED ISSUE #148: posix_fadvise() and POSIX_FADV_*
I'm working on an application which will be reading many small files, but only once each. I'd like to be able to use `posix_fadvise(fd, 0, 0, POSIX_FADV_WILLNEED)` to tell the kernel to asynchronously prefetch the file so when it comes time to access it, it will already have been pulled into the page cache in its entirety (less than 1MB on average) and can be read with minimal blocking.

I was going to implement this myself and issue a PR, but I noticed the semantically similar `posix_madvise()` has its constants (`POSIX_MADV_*`) duplicated across the Linux and various BSD/misc *nix modules and I'm not sure why because they have the same values AFAICT.

cybergeek94:
Closes #148 

I didn't touch the `POSIX_MADV` constants, those can be consolidated in another PR.

I also wrote a quick test to make sure that the call was written correctly and that `posix_fadvise()` actually works for my use-case:

https://gist.github.com/cybergeek94/ce5e989c6ac6823acea4

As you can see, the uncached file takes much longer to read without the call to `posix_fadvise()`, which is exactly the result I wanted to see. A test file size of 8M is around what I'll be dealing with in my particular use-case (large libraries of images).

cc @alexcrichton

### DIFF
--- a/src/unix/bsd/freebsdlike/mod.rs
+++ b/src/unix/bsd/freebsdlike/mod.rs
@@ -365,6 +365,13 @@ pub const POSIX_MADV_SEQUENTIAL: ::c_int = 2;
 pub const POSIX_MADV_WILLNEED: ::c_int = 3;
 pub const POSIX_MADV_DONTNEED: ::c_int = 4;
 
+pub const POSIX_FADV_NORMAL: ::c_int = 0;
+pub const POSIX_FADV_RANDOM: ::c_int = 1;
+pub const POSIX_FADV_SEQUENTIAL: ::c_int = 2;
+pub const POSIX_FADV_WILLNEED: ::c_int = 3;
+pub const POSIX_FADV_DONTNEED: ::c_int = 4;
+pub const POSIX_FADV_NOREUSE: ::c_int = 5;
+
 pub const _SC_IOV_MAX: ::c_int = 56;
 pub const _SC_GETGR_R_SIZE_MAX: ::c_int = 70;
 pub const _SC_GETPW_R_SIZE_MAX: ::c_int = 71;
@@ -630,6 +637,10 @@ extern {
                     hdtr: *mut ::sf_hdtr,
                     sbytes: *mut ::off_t,
                     flags: ::c_int) -> ::c_int;
+
+    pub fn posix_fadvise(fd: ::c_int, offset: ::off_t, len: ::off_t, 
+                         advise: ::c_int) -> ::c_int;
+
 }
 
 cfg_if! {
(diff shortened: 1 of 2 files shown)


alexcrichton:
lgtm, will merge once CI is green

alexcrichton:
Hm not entirely sure why travis didn't trigger here, perhaps a force push will fix that?

cybergeek94:
```
$ git push --force
Everything up-to-date
```

Should I move some things around or something?

alexcrichton:
Ah I think if you do `git commit --amend` and don't actually change anything it'll basically touch a timestamp to allow you to force-push

cybergeek94:
Done. I see the build has been triggered this time.

One possible concern is that `posix_fadvise()` is supposed to have a slightly different signature on ARM, as explained under the Notes section in the [documentation](http://linux.die.net/man/2/posix_fadvise). However, it appears it just means that calling this on ARM will use a couple extra registers because the arguments have to be aligned to 64-bit.

Also, the `POSIX_FADV_DONTNEED` and `POSIX_FADV_NOREUSE` constants have slightly different values on S/390 but considering that Rust does not support S/390 nor likely ever will, I figured it wouldn't be a problem.

alexcrichton:
Looks like this may not exist on OSX or on OpenBSD? Otherwise the signature seems to be agreed upon for supported arches

cybergeek94:
Support appears to be inconsistent across BSD releases. It's in some but not in others, earlier *or* later. Should I just move it to `notbsd` then?

alexcrichton:
Ah I'd be fine either way

cybergeek94:
`notbsd` covers my use case, but I find it weird that there wasn't a similar error for `posix_madvise()`. I guess the BSDs decided to support only that one?

semarie:
yes, OpenBSD supports `posix_madvise` (see http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man2/madvise.2), but don't support `posix_fadvise`.

cybergeek94:
Seems pretty arbitrary for OpenBSD but I'm fine moving this to `notbsd` and possibly `freebsdlike` so FreeBSD gets it.

alexcrichton:
Yeah that sounds good to me, just lemme know when it's pushed up!

cybergeek94:
Moved function decl and constants to relevant files and squashed.

cybergeek94:
Cheers!

PR #184, opened 2016-02-15
unix: Add mkstemp(3) and related functions

kamalmarhubi:


alexcrichton:
Looks like Android doesn't have the non-unix ones?

semarie:
Just for infomation, `mkostemp` and `mkostemps` are also present in OpenBSD, NetBSD, FreeBSD and DragonflyBSD (see http://bxr.su/search?defs=mkostemp and http://bxr.su/search?defs=mkostemps)

kamalmarhubi:
Shuffled some things around. @semarie thanks for pointing that out. Bitrig does indeed have them:
- [mkostemp](https://github.com/bitrig/bitrig/blob/f26f17e8bd4a61121b02971dc71855a8e0538669/lib/libc/stdio/mktemp.c#L140)
- [mkostemps](https://github.com/bitrig/bitrig/blob/f26f17e8bd4a61121b02971dc71855a8e0538669/lib/libc/stdio/mktemp.c#L127)

PR #190, opened 2016-02-17
Android: add some of the same LFS functions as Linux

### LINKED ISSUE #189: Android needs LFS exports too
In some ways Android's `bionic` dodges LFS, like its structs `stat` and `dirent` that are always the 64-bit versions.  But as mentioned in its own [32-bit ABI bugs](https://android.googlesource.com/platform/bionic.git/#32_bit-ABI-bugs), `off_t` is still an issue in other places.

Of those that rust itself uses, which we're actively switching to LFS:
- Every `*stat` variant is already 64-bit safe, as noted.
- `readdir_r` is similarly fine since `dirent` is always 64-bit.
- `open` already forces `O_LARGEFILE` implicitly.
- `ftruncate` and `lseek` both take `off_t` parameters, so need explicit LFS support.

At least from gingerbread they do have `ftruncate64` and `lseek64`, and more LFS functions have been added since.  What is the baseline here for Android support?  With that info, I'm willing to work on this myself.

cuviper:
Audit the LFS functions from Linux and add the same for Android if
bionic supports it, according to libc.map from API 23 (Android 6.0).

The standard library is currently limited to API 18, so it may only use
`lseek64`, `pread64`, `pwrite64`, `ftruncate64`, and `readahead`.

Note that `stat64` and `dirent64` are actually identical to the regular
versions, because bionic has always mapped them to the LFS syscalls.
It also sets `O_LARGEFILE` at all times, so `open` matches `open64`.
Still, the explicit LFS names may be useful to match Linux.

Fixes #189.

### DIFF
--- a/src/unix/notbsd/android/mod.rs
+++ b/src/unix/notbsd/android/mod.rs
@@ -646,8 +646,6 @@ extern {
                        result: *mut *mut ::dirent64) -> ::c_int;
     pub fn getrlimit64(resource: ::c_int, rlim: *mut rlimit64) -> ::c_int;
     pub fn setrlimit64(resource: ::c_int, rlim: *const rlimit64) -> ::c_int;
-    pub fn readahead(fd: ::c_int, offset: ::off64_t,
-                     count: ::size_t) -> ::ssize_t;
 }
 
 cfg_if! {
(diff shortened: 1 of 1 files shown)


cuviper:
Hmm, seems `readahead` is present but not declared in any header, so CI failed.
I'll just nuke that one...

alexcrichton:
Thanks!

PR #191, opened 2016-02-17
fixes some netbsd compile errors

dhuseby:
cleaning up the netbsd build a bit.

### DIFF
--- a/src/unix/bsd/openbsdlike/netbsd.rs
+++ b/src/unix/bsd/openbsdlike/netbsd.rs
@@ -43,19 +43,20 @@ s! {
         pub st_gid: ::gid_t,
         pub st_rdev: ::dev_t,
         pub st_atime: ::time_t,
-        pub st_atimensec: ::c_long,
+        pub st_atime_nsec: ::c_long,
         pub st_mtime: ::time_t,
-        pub st_mtimensec: ::c_long,
+        pub st_mtime_nsec: ::c_long,
         pub st_ctime: ::time_t,
-        pub st_ctimensec: ::c_long,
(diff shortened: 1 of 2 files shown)


alexcrichton:
Looks like there are some OpenBSD compile errors? The iOS ones seem unrelated and I can fix those separately

dhuseby:
@alexcrichton i think i fixed the compile error.

alexcrichton:
Looks like the automation disagrees with the change to `st_spare`?

dhuseby:
@alexcrichton this is a generated test?  I'm not sure how to fix this.  I tried to track it down in the libc-test folder but it must be autogenerating code based on the struct's fields because there is no explicit reference to st_spare.  That suggests the autogenerating code is wrong or can't handle that special case.

alexcrichton:
Yeah these are generated tests, and it was correct beforehand, so why the change?

dhuseby:
@alexcrichton because of the function here: https://github.com/dhuseby/rust/blob/master/src/libstd/os/netbsd/fs.rs#L151  but now I think the real solution is to fix rust/src/libstd instead of changing libc.

alexcrichton:
On netbsd that function should probably just be removed or properly remove `[u32; 2]`, it's intended that the trait returns the platform-specific types

alexcrichton:
Sounds like the symbols were taken care of in https://github.com/rust-lang/libc/pull/197, and libstd probably just needs to be corrected for the stat, so closing. Feel free to reopen though if I'm missing something!

dhuseby:
@alexcrichton kthx

PR #196, opened 2016-02-22
Add ucontext struct for Linux/x86

fiveop:
I need the following structs from glibc for low level work. I pulled the information about the types from the [glibc source](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/x86/sys/ucontext.h;hb=de6e5df27ead590c7f590a63665cb87764cf6de3). There is a lot of duplication due to the structure of rust libc. If the structs for other platforms would be added (and I might do that in the future), some of that repetiton might be eliminated.
I hid some structs in a module called 'detail'. I wanted to use the `s!`-macro for some structs that I did not want to provided public access to. Only the parts specified in the linux man page for getcontext are exposed publicly.

### DIFF
--- a/src/unix/notbsd/linux/other/b32/x86.rs
+++ b/src/unix/notbsd/linux/other/b32/x86.rs
@@ -18,3 +18,27 @@ pub const SO_SNDTIMEO: ::c_int = 21;
 
 pub const FIOCLEX: ::c_ulong = 0x5451;
 pub const FIONBIO: ::c_ulong = 0x5421;
+
+s! {
+
+    pub struct mcontext_t {
+        __private: [u32; 22]
+    }
+
+    pub struct ucontext_t {
+        pub uc_flags: ::c_ulong,
+        pub uc_link: *mut ucontext_t,
+        pub uc_stack: ::stack_t,
+        pub uc_mcontext: mcontext_t,
+        pub uc_sigmask: ::sigset_t,
+        __private: [u8; 112],
+    }
+
+}
+
+extern {
+    pub fn getcontext(ucp: *mut ucontext_t) -> ::c_int;
+    pub fn setcontext(ucp: *const ucontext_t) -> ::c_int;
+    pub fn makecontext(ucp: *mut ucontext_t, func:  extern fn (), argc: ::c_int, ...);
+    pub fn swapcontext(uocp: *mut ucontext_t, ucp: *const ucontext_t) -> ::c_int;
+}
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -132,6 +132,7 @@ fn main() {
 
     if linux {
         cfg.header("mqueue.h");
+        cfg.header("ucontext.h");
         cfg.header("sys/signalfd.h");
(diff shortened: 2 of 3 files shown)


alexcrichton:
Thanks for the PR! Could this match the previous structure in terms of layout? (e.g. avoiding `mod detail`). You should be able to hide away private details for now with private fields I believe.

alexcrichton:
Ah yeah duplication is fine for now, we can always lift up later. Although it may be good to start defining `ucontext_t` on other platforms as well? Looks like OSX at least has the functions defined.

fiveop:
> You should be able to hide away private details for now with private fields I believe.

I can and do that. However, some of these fields use structs as types, that I have to define. For simplicity, I wanted to use the `s!` macro to define them. That macro only allow to create public structs. If I put them in the exported module, we would export structs that no one could use.

It might be better to change `s!` to allow for non-public structs.

alexcrichton:
Yeah for now if you want to keep the structs private (is there a reason to intentionally do so?) then I'd recommend just doing `_priv: [usize; amt]`  and not worrying about the layout

fiveop:
Debugging through Travis takes a while. (There was no other reason for hiding the structs, than them not being documented in the man pages).

alexcrichton:
Thanks! I'm actually somewhat curious what it'll take to push this all the way over the finish line to a bunch of other platforms as well, so I'm gonna play around with a PR that does that for a bit.

src/unix/notbsd/linux/other/b32/x86.rs
```diff
@@ -18,3 +18,53 @@ pub const SO_SNDTIMEO: ::c_int = 21;
...
+        cw: ::c_ulong,
+        sw: ::c_ulong,
+        tag: ::c_ulong,
+        ipoff: ::c_ulong,
+        cssel: ::c_ulong,
+        dataoff: ::c_ulong,
+        datasel: ::c_ulong,
+        _st: [_libc_fpreg; 8],
+        status: ::c_ulong,
+    }
```
alexcrichton: Since these are just private structures with all private fields, can they be left out for now? They can likely just be represented via `__data: [usize; AMT]` below somehow

src/unix/notbsd/linux/other/b32/x86.rs
```diff
@@ -18,3 +18,53 @@ pub const SO_SNDTIMEO: ::c_int = 21;
...
 pub const FIOCLEX: ::c_ulong = 0x5451;
 pub const FIONBIO: ::c_ulong = 0x5421;
+
+type greg_t = ::c_int;
+
+const NREG : usize = 19;
+
+type gregset_t = [greg_t; NREG];
+
+s! {
```
alexcrichton: Can you include these definitions and such in the same style as the rest of the file? For example typdefs are at the top, then the structs, then constants, etc.

alexcrichton:
Ah looks like these are so non-portable it's not worth it to try to add bindings for other platforms, so just a few stylistic nits and then this should be good to go.

PR #197, opened 2016-02-26
Fix deprecated NetBSD sigaction/sigaltstack symbols

gandro:
Especially `sigaction` is needed for NetBSD/amd64, as otherwise every invocation crashes with ENOSYS since the legacy system call was never available on x86_64.

This replaces rust-lang/libc#191


alexcrichton:
Weird, not sure why ctest didn't pick this up? ah well

Thanks!

PR #198, opened 2016-02-27
Add ucontext bindings

alexcrichton:


alexcrichton:
Bah looks like these aren't really that portable after all, we'll just go with the first PR

PR #199, opened 2016-02-27
added sigevent and mq_notify

MarkusJais:
I've added support (Linux) for sigevent in order to add mq_notify. I've contributed mq* support for nix-rust and will update it there to use libc. mq_notifiy was missing and more complex because it requires sigevent.

Tests in ci directory (on Ubuntu) did pass but this is my first contribution to libc and should receive careful review.

Happy to help!


src/unix/notbsd/linux/mod.rs
```diff
@@ -165,6 +165,20 @@ s! {
         _pad: [::uint8_t; 48],
     }
 
+
+    pub struct sigevent {
+        pub sigev_value: sigval_t,
+        pub sigev_signo: ::c_int,
+        pub sigev_notify: ::c_int,
+        pub sigev_notify_function: ::dox::Option<extern fn(sigev_value: sigval_t)>,
```
kamalmarhubi: looks like this should be a pointer to extern fn?

https://github.com/torvalds/linux/blob/v4.3/include/uapi/asm-generic/siginfo.h#L294

src/unix/notbsd/linux/mod.rs
```diff
@@ -165,6 +165,20 @@ s! {
...
+        pub sigev_notify_attributes: *mut pthread_attr_t,
+        _pad: [u64; 4usize],
+    }
+
+    pub struct sigval_t {
+        _data: [u64; 1usize],
```
kamalmarhubi: https://github.com/torvalds/linux/blob/v4.3/include/uapi/asm-generic/siginfo.h#L7-L10

(I have no idea how to handle unions, so can't give any usefl advice)

kamalmarhubi:
I can't tell why the musl build is failing. It can't find `sigval_t`, but I think it's in `mqueue.h` which is being included?

src/unix/notbsd/linux/mod.rs
alexcrichton: This may actually want to just be a `usize`, C has a habit of sometimes using sentinel values for function pointers that mean "invalid" or "do this other specific action"

alexcrichton:
Thanks! Looks like there's a few CI errors as well for platforms like MUSL

MarkusJais:
I'll try to have a look at the MUSL library.

Part of the bindings are basically what rust-bindgen generated but with better names. I am not sure if bindgen can properly handle unions.

I don't think sigevent is particularly useful or even a good idea (low level, probably error prone with threads and signals) but I'll try to make it work as it should be available to give Rust programmers all the options of the C API (that doesn't mean that I encourage using it for IPC, there are better options).

alexcrichton:
Yeah for now it's ok to leave it as opaque and we can fill it out over time if necessary

MarkusJais:
BTW, are there any guidelines on how to handle C unions in Rust (and in libc)? I couldn't find that much searching the web.

alexcrichton:
Unfortunately not really. One of the only unions we deal with today is `siginfo` and it's just dealt with a little ad-hocly. Without language support for unions there's not a lot we can do here unless the C side has a few extra structures for each version of the union.

Basically it's case-by-case today.

MarkusJais:
In that case, is it even possible to get this done in a safe way? If unions cannot be dealt with at compile time it might be hard to get the memory layout and size correct.
I am not compiler expert but I fear it could end up in broken memory.

kamalmarhubi:
There's [an RFC for adding unions](https://github.com/rust-lang/rfcs/pull/1444), which looks like it's going places. Until then, I think it's ad hoc code downstream from libc has to be careful to do things right. There's [an RFC for bitfields too](https://github.com/rust-lang/rfcs/pull/1449), but it has a bit less activity so far.

To help with getting downstream code right, I came across the idea of [having a small C fragment](https://users.rust-lang.org/t/c-structs-with-bit-fields-and-ffi/1429/5) to ensure correct endianness / ABI handling. I think I'll be trying that out this week.

alexcrichton:
For now it's probably best to get the size/alignment correct and not worry about the fields for now. That way it's at least a usable API and the fields can be extracted manually downstream.

MarkusJais:
Wow, that RFC has a lot of comments. Obviously not a trivial problem :-)
The size right now was generated with bindgen. I had all tests run on my machine and they were ok but I haven't yet looked in musl. Will do.

kamalmarhubi:
> The size right now was generated with bindgen. I had all tests run on my machine and they were ok but I haven't yet looked in musl. Will do.

Since musl is using the same syscall ABI, the types should have the same size and layout. I took a quick look at musl, and the problem is that the union isn't typedefed as `sigval_t`:

kamalmarhubi:
Based on this, I think the only thing that can be done is to skip the type in libc-test. Is that right @alexcrichton?

alexcrichton:
I'd rather not skip types as cases like this seems tricky, perhaps this could just be an opaque struct for now?

MarkusJais:
maybe a stupid question but what do you mean with "opaque struct"?

I agree about not skipping tests!

kamalmarhubi:
I believe it means a struct with private fields that have the right size and alignment, but not necessarily the right types.

MarkusJais:
After reading more about unions and Rust I believe it is best to wait until Rust has support for unions (see the RFC mentioned above). Otherwise it seems to be too fragile to add more unions. 
I hope the union support mentioned in that RFC will be part of Rust soon.

alexcrichton:
Closing this due to inactivity, but the good news is that we've accepted an RFC for `union` and that'll hopefully be implemented in the near future!

MarkusJais:
cool, looking forward to the union support.

PR #200, opened 2016-02-27
Various Rumprun fixes

gandro:
This moves some pthread_*_np definitions that ended up in the wrong BSD modules. Additionally, it avoids linking against libc and librt on Rumprun when building libstd. librt is not available on Rumprun, and libc is pulled in automatically by Rumpruns own linker.

Because conditional compilation for Rumprun is using the unstable `target_vendor` attribute, I'm relying on the fact that `#[cfg]` evaluation is short-circuiting regarding feature gate checks.I had to change the `cfg_if!` macro to support this. Now, if the `if #[cfg(..)]` clause is true, then the `else if #[cfg(..)]` clause will not be evaluated, allowing me to use unstable attributes for libstd builds.

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -143,6 +143,12 @@ cfg_if! {
     } else if #[cfg(target_os = "emscripten")] {
         #[link(name = "c")]
         extern {}
+    } else if #[cfg(all(target_vendor = "rumprun", target_os = "netbsd"))] {
+        // Since we don't use -nodefaultlibs on Rumprun, libc is always pulled in
+        // automatically by the linker. We avoid passing it explicitly, as it
+        // causes some versions of binutils to crash with an assertion failure.
+        #[link(name = "m")]
+        extern {}
     } else if #[cfg(any(target_os = "macos",
                         target_os = "ios",
                         target_os = "android",
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -350,10 +350,6 @@ fn main() {
             // they're implemented on rumprun yet, just let them slide for now.
             // Some of them look like they have headers but then don't have
             // corresponding actual definitions either...
-            "backtrace" |
-            "pthread_main_np" |
-            "pthread_set_name_np" |
-            "pthread_stackseg_np" |
             "shm_open" |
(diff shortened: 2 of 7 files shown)


src/unix/mod.rs
```diff
@@ -143,6 +143,9 @@ cfg_if! {
     } else if #[cfg(target_os = "emscripten")] {
         #[link(name = "c")]
         extern {}
+    } else if #[cfg(all(target_vendor = "rumprun", target_os = "netbsd"))] {
+        #[link(name = "m")]
```
alexcrichton: We pass a few flags to the linker to say that default system libraries shouldn't be included by default, so shouldn't we link both libc and libm like the block just below this one?

gandro: Do you mean `-nodefaultlibs`? That flag is actually not used in [Rust's Rumprun target](https://github.com/rust-lang/rust/blob/257fa0f479e6a0fc069710a4c2ea52e551bafe97/src/librustc_back/target/x86_64_rumprun_netbsd.rs#L23), since the Rumprun linker always pulls its own libc. So not passing `-lc` explicitly should be fine. 

The reason why I removed the explicit `-lc` is however due to a bug in binutils (as invoked by the Rumprun linker). I haven't really been able to track it down, but for some reason it crashes with an assertion failure if `-lc` is passed to `ld` twice (once explicitly by Rust, and once by Rumprun's linker). 

I can try to get a workaround in the Rumprun linker, so it removes the superfluous `-lc` passed to it, but opted for removing it here, I hope that's okay. Having a redundant `-lc` is not a problem when linking C programs, so I'm not really sure what's going on here.

alexcrichton: Whoa, weird! If we're triggering bugs I guess it seems fine to leave this here. Could you add some comments to this effect though?

gandro: Rebased and added a comment to clarify why we omit `-lc`. Note that while I only observed the assertion failure when linking Rust, I don't think the crash is Rust's fault. It probably is caused by Rumpruns use of  `ld -r` which has triggered bugs in binutils before, see rumpkernel/rumprun-packages#19

PR #201, opened 2016-02-28
Add sigwait/sigwaitinfo/sigtimedwait.

fiveop:


fiveop:
I fixed it as much as I could. I noticed that a lot of functions had special linker attributes for netbsd. How do I determine the right linker targets for a function on netbsd (and apparently i686-apple-darwin)?

alexcrichton:
Ah most of those attributes came out as a result of CI, but if the CI is green then no need for any attributes! Looks like though `sigwait` on OSX has a special name as well as `sigtimedwait` on NetBSD, so I guess this isn't the case here unfortunately.

From what I can tell, however, the names seem to be:

* OSX - `sigwait$UNIX2003` (you can follow other examples here as well)
* NetBSD - `__sigtimedwait50` (there's also probably a few other examples here as well)

alexcrichton:
Thanks!

Issue #195, opened 2016-02-20
Scope of functions in libc crate: can we add things that require linking another "common" lib?


kamalmarhubi:
How far outside of actual libc does the libc crate's scope go? Eg, I'm considering adding [`openpty(3)`](http://man7.org/linux/man-pages/man3/openpty.3.html) which requires linking with `libutil` on some platforms but not others[0]. Would this fit in libc?
<br>
---
[0] https://www.gnu.org/software/gnulib/manual/html_node/openpty.html

alexcrichton:
Thanks for the issue! The scope of libc is [currently defined in its RFC](https://github.com/rust-lang/rfcs/blob/master/text/1291-promote-libc.md#scope-of-libc), which means that if it's available on some platform through the standard set of libraries that should be fine, but on other platforms we won't be able to have the API.

kamalmarhubi:
:+1: thanks, I'll make a PR :-)

PR #202, opened 2016-02-28
apple: Add openpty(3) and forkpty(3)

kamalmarhubi:


alexcrichton:
Thanks! Right now the [scope of this library](https://github.com/rust-lang/rfcs/blob/master/text/1291-promote-libc.md#scope-of-libc) doesn't include `libutil` on Unix, but we can perhaps propose an amendment to that RFC to include it. Is this library part of the POSIX spec or something like that? Or is it just "random artifact" lying around on most Unix distributions?

kamalmarhubi:
Ah I misunderstood your reply to #195 maybe! [Linux man page](http://man7.org/linux/man-pages/man3/openpty.3.html) says

```
These are BSD functions, present in glibc.  They are not standardized
in POSIX.
```

They're available on Linux and all the bsdlikes; missing on Solaris. On non-OS X, they require linking with `libutil`. This little page from gnulib docs has a good run-down: https://www.gnu.org/software/gnulib/manual/html_node/openpty.html

alexcrichton:
Ok, in that case I think it'd be fine to add these to Linux bindings, but we should leave them out elsewhere for now as they require an extra library to get linked.

kamalmarhubi:
It's OS X that doesn't require linking anything else. For the other platforms, I think libutil is always available, but I'm fine with removing them for now, which I've done.

> we can perhaps propose an amendment to that RFC to include it

Would the amendment proposal itself need to be an RFC?

alexcrichton:
Oh sorry, but yeah it's fine to add OSX for now if it doesn't require more libs. Also looks like the failure here was spurious, so I'll restart the build.

> Would the amendment proposal itself need to be an RFC?

Not necessarily a full fledged RFC itself, but a PR to the RFCs repo for sure.

kamalmarhubi:
I guess I need to figure out the implications of `#[link]` attributes. Do they result in downstream users linking those extra libraries? In debug? In release? With lto? Will do some investigation and consider opening an on the RFC to expand scope to include libs that are always available. Eg, for libutil on Linux, it is part of libc, and included in the libc package on Debian.

alexcrichton:
The `#[link]` attribute is unconditionally propagated downwards. It means that basically all Rust programs would only succeed if `-lutil` worked.

kamalmarhubi:
That's what I thought. I'll survey a few major Linux distros if / when I move to open a PR on the RFC. Thanks for clarifying!

kamalmarhubi:
I opened a PR on the RFC: https://github.com/rust-lang/rfcs/pull/1529

kamalmarhubi:
@alexcrichton could this be merged as is, or should we wait for resolution of the RFC and add it to all platforms at once?

alexcrichton:
Let's hold off on this for now until the RFC reaches a conclusion

kamalmarhubi:
:+1:

alexcrichton:
Oh wait sorry I just realized this is only adding for OSX, oh yeah this is fine to merge once CI is green.

kamalmarhubi:
Oh yeah, should have pointed that out. :-)

PR #204, opened 2016-02-29
Add futimens

pitdicker:


alexcrichton:
Looks like this may not be present on OSX?

pitdicker:
This automation is _really_ great :)

alexcrichton:
Ah looks like Android also doesn't have this function unfortunately :(

alexcrichton:
Ah I'm rolling up some other PRs, I'll take care of it.

pitdicker:
Just to make sure, android does not seem to have  `utimes`, but it does have `utimens`

PR #209, opened 2016-03-05
Add SYS_gettid

dhylands:
I'm trying to get gettid added (either in libc or nix). We discussed things a bit over here:
https://github.com/nix-rust/nix/pull/293 so I thought I would take a stab at adding the SYS_gettid constant into libc.

Should I also add a gettid function here? Or should that go in nix?

I wasn't sure which variant of MIPS was being used, so I didn't add the constant into the mips.rs. The constants are: 4222 for O32, 5178 for N64 and 6178 for N32. Since it seems to be used for OpenWRT, I'd guess O32.

Should this also be added to musl?

### DIFF
--- a/src/unix/notbsd/android/b32.rs
+++ b/src/unix/notbsd/android/b32.rs
@@ -1,5 +1,7 @@
 pub type mode_t = u16;
 
+pub const SYS_gettid: ::c_int = 224;
+
 s! {
     pub struct sigaction {
         pub sa_sigaction: ::sighandler_t,
--- a/src/unix/notbsd/android/b64.rs
+++ b/src/unix/notbsd/android/b64.rs
@@ -1,5 +1,7 @@
 pub type mode_t = u32;
 
+pub const SYS_gettid: ::c_int = 178;
+
 s! {
     pub struct sigaction {
         pub sa_flags: ::c_uint,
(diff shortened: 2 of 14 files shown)


alexcrichton:
Ah yeah feel free to add this to MUSL (it'll pass if it's defined in MUSL), and also feel free to add the `gettid` function itself (so long as it's in glibc everywhere). Other than that looks good to me, thanks!

dhylands:
gettid isn't present in glibc, but it is available in the linux kernel, which is the motivation behind adding SYS_gettid. Then it's possible to write:
```
pub fn gettid() -> pid_t {
    unsafe { syscall(SYS_gettid) }
}
```
but that needs to go someplace after syscall is defined.

So it would need to go in https://github.com/rust-lang/libc/blob/master/src/unix/notbsd/mod.rs which implies I'd need to do this in android as well.

Does android assume arm?

alexcrichton:
Ah yeah for now we don't define functions like this (just bind what's in libc), so let's avoid adding that function for now. Can you rebase this as well? (just merged something else)

dhylands:
OK - I'll just add SYS_gettid in libc and we can add the gettid function itself in nix.

I'll rebase.

dhylands:
ok - rebased and added to remaining linux modules. The travis failures I'm seeing don't seem to have anything to do with my commit.

alexcrichton:
Thanks!

PR #210, opened 2016-03-06
i686-musl: split timespec fields into (time_t, c_long) pairs

japaric:
This is how MetadaExt expects these fields to be named.

See https://github.com/rust-lang/rust/blob/c116ae35cf49b55bd8d82e31f1ba030cf7e63867/src/libstd/os/linux/fs.rs#L107-L121

---

r? @alexcrichton 
Alternatively, we could leave these fields as their are, and, instead, add some conditional logic (#[cfg]) to the MetadataExt implementation.


alexcrichton:
Thanks! Would you be ok doing this for arm and asmjs at the same time as well?

japaric:
> Thanks! Would you be ok doing this for arm and asmjs at the same time as well?

Done

PR #211, opened 2016-03-06
unix: Include fdatasync(2) on non-Apple systems

kamalmarhubi:
The function is defined in POSIX [0], but according to Gnulib docs [1],
it is missing or not declared on at least some versions of OS X. On
Solaris, it is not a system call but is present as fdatasync(3) [2].

[0] http://www.opengroup.org/onlinepubs/9699919799/functions/fdatasync.html
[1] https://www.gnu.org/software/gnulib/manual/html_node/fdatasync.html
[2] http://docs.oracle.com/cd/E36784_01/html/E36874/fdatasync-3c.html


alexcrichton:
Looks like this may not exist on FreeBSD?

kamalmarhubi:
Urgh sorry, I was sure I'd checked.

I need to figure out how to run these tests locally!

kamalmarhubi:
I think the windows failure is spurious. The failing job has no log: https://ci.appveyor.com/project/alexcrichton/libc/build/1.0.762/job/q46jx47ua7exb5ec

PR #213, opened 2016-03-06
Fix SYS_gettid type to match first argument of syscall.

dhylands:
My apoogies - I should have picked this up on the first PR


alexcrichton:

PR #214, opened 2016-03-06
musl: opt for dynamic linking for arm arch

joerg-krause:
When building with musl for an arm target link with a shared libc (like mips already does).


alexcrichton:

PR #215, opened 2016-03-07
notbsd: Add utimensat

fpgaminer:
This pull request adds the `utimensat` function, which is related to `futimens` (already in the code).  `utimensat` allows modifying the mtime/atime using a path instead of a file descriptor, and it also allows modifying mtime/atime on a symlink (the link's inode, rather than the target inode).

I also added the two constants which can be passed to `utimensat`: `AT_FDCWD` and `AT_SYMLINK_NOFOLLOW`.  I used http://lxr.free-electrons.com/source/include/uapi/linux/fcntl.h#L56 to get the constant values.

I tested `libc::utimensat` on my local system (Ubuntu 15.10).  Not sure if I'm supposed to add any unit tests with this pull request or not.


alexcrichton:

PR #217, opened 2016-03-07
Add MAP_STACK and MAP_HUGETLB entries to MIPS so that the nix crate can be built

gabrielesvelto:
This is a minimal change to the MIPS sources that enables building the nix crate which depends on libc. Additional changes are required in the nix crate to build for mips/mipsel targets but this is the only external change that's preventing me from getting a build right now.


alexcrichton:
Thanks!

PR #219, opened 2016-03-09
linux: Add SYS_perf_event_open syscall numbers

kamalmarhubi:
There is no glibc wrapper for this system call; it must be called as
`syscall(__NR_perf_event_open, ...)`.
    
The associated `perf_event_attr` struct has several unions and one
set of bitfields. These are currently represented as padding.

### DIFF
--- a/src/unix/notbsd/linux/musl/b32/arm.rs
+++ b/src/unix/notbsd/linux/musl/b32/arm.rs
@@ -304,3 +304,4 @@ pub const FIONREAD: ::c_ulong = 0x541B;
 pub const TIOCCONS: ::c_ulong = 0x541D;
 
 pub const SYS_gettid: ::c_long = 224;
+pub const SYS_perf_event_open: ::c_long = 364;
--- a/src/unix/notbsd/linux/musl/b32/mips.rs
+++ b/src/unix/notbsd/linux/musl/b32/mips.rs
@@ -303,3 +303,4 @@ pub const FIONREAD: ::c_ulong = 0x467F;
 pub const TIOCCONS: ::c_ulong = 0x80047478;
 
 pub const SYS_gettid: ::c_long = 4222;   // Valid for O32
+pub const SYS_perf_event_open: ::c_long = 4333;  // Valid for O32
--- a/src/unix/notbsd/linux/musl/b32/x86.rs
+++ b/src/unix/notbsd/linux/musl/b32/x86.rs
@@ -305,3 +305,4 @@ pub const FIONREAD: ::c_ulong = 0x541B;
 pub const TIOCCONS: ::c_ulong = 0x541D;
 
 pub const SYS_gettid: ::c_long = 224;
+pub const SYS_perf_event_open: ::c_long = 336;
--- a/src/unix/notbsd/linux/musl/b64/aarch64.rs
+++ b/src/unix/notbsd/linux/musl/b64/aarch64.rs
@@ -1,0 +1,1 @@
+pub const SYS_perf_event_open: ::c_long = 241;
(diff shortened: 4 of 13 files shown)


src/unix/notbsd/mod.rs
```diff
@@ -124,6 +124,68 @@ s! {
         pub machine: [::c_char; 65],
         pub domainname: [::c_char; 65]
     }
+
+    pub struct perf_event_attr {
+        pub type_: u32,  // called `type`
+        pub size: u32,
+        pub config: ::c_ulonglong, // TODO: __u64
```
kamalmarhubi: @alexcrichton there is some weirdness here: they are declared as `__u64`, which is a typedef of `unsigned long long` even on x86_64. In libc, we have `type uint64_t = u64` which winds pu as `unsigned long`. This passes on x86_64, but I'm not sure what best to do to make sure it's portable.

kamalmarhubi:
There's a niggle with this in that the definition of `struct perf_event_attr` has changed a few times over the years. I think it's grown each time, but I'm not completely sure. I haven't done the archaeology to figure it out.

kamalmarhubi:
Ah I seem to have some of the syscall numbers wrong. The main discussion points are independent of that though. I'll try and fix them up tomorrow!

libc-test/build.rs
```diff
@@ -141,6 +141,7 @@ fn main() {
     }
 
     if linux || android {
+        cfg.header("linux/perf_event.h");
```
alexcrichton: In the past we've been burned by including `linux` headers (mismatches with userspace or otherwise using unstable bits), is this appropriate to define?

src/unix/notbsd/mod.rs
alexcrichton: Eh for me so long as the CI passes I'm fine with it

alexcrichton:
Hm if this is changing, is it ready for use in userspace libraries? Maybe this better belongs in a downstream crate?

libc-test/build.rs
kamalmarhubi: yeah, per [man page](http://www.man7.org/linux/man-pages/man2/perf_event_open.2.html):

```
SYNOPSIS
       #include <linux/perf_event.h>
       #include <linux/hw_breakpoint.h>

       int perf_event_open(struct perf_event_attr *attr,
                           pid_t pid, int cpu, int group_fd,
                           unsigned long flags);

       Note: There is no glibc wrapper for this system call; see NOTES.
```

kamalmarhubi:
> Hm if this is changing, is it ready for use in userspace libraries? Maybe this better belongs in a downstream crate?

I'm still taking a look but I believe the changes are that it has grown a few times. This is the purpose of the `size` field:

```
       size   The size of the perf_event_attr structure  for  forward/backward
              compatibility.  Set this using sizeof(struct perf_event_attr) to
              allow the kernel to see the struct size at the time of  compila‐
              tion.

              The  related  define  PERF_ATTR_SIZE_VER0 is set to 64; this was
              the size of the first published struct.  PERF_ATTR_SIZE_VER1  is
              72,  corresponding  to  the  addition  of  breakpoints  in Linux
              2.6.33.  PERF_ATTR_SIZE_VER2 is 80 corresponding to the addition
              of  branch sampling in Linux 3.4.  PERF_ATR_SIZE_VER3 is 96 cor‐
              responding  to  the  addition  of  sample_regs_user   and   sam‐
              ple_stack_user in Linux 3.7.
```

The struct is passed to the kernel by pointer. As such, I believe it's safe to use the most recent version available; users will fill `size` in with `mem::size_of::<perf_event_open>()`. 

Thoughts?

kamalmarhubi:
Oh, and I'm new to this perf business so I can't really answer detailed questions on its usage. But I *am* trying to wite some Rust that interfaces with it.

alexcrichton:
Hm ok given the instability of this, the lack of unions, and the lack of its presence in glibc at all, I'm tempted to say that we may not want to add this to libc now? Although this could in theory house all type definitions forever, we should likely limit it to just that which is contained in glibc itself.

alexcrichton:
Yeah I'd be fine adding syscall numbers!

kamalmarhubi:
The reasoning behind leaving the struct out makes a lot of sense to me. I've revised this to include only the (hopefully correct this time!) syscall numbers.

alexcrichton:
Thanks!

PR #222, opened 2016-03-11
Fix DragonFly: clock_gettime() expects CLOCK_* constants as ulong.

mneumann:
While there add all CLOCK_ constants as found on DragonFly.


alexcrichton:
@bors: r+ 378bb73bcc960da500dd89a431f29eaf1810ee56

Let's see if this works...

PR #225, opened 2016-03-12
Pull sockaddr_nl up a layer as it is present for Android

posborne:
The addition of sockaddr_nl seems to be a fairly recent
addition to bionic, although it would appear that its lack
of presence in previous versions is more of an oversight than
anything else.

https://github.com/android/platform_bionic/blob/831c8a52498d3c7be204a00847275f6e9163626f/libc/kernel/uapi/linux/netlink.h#L54


alexcrichton:
@bors: r+ 9e94d5b777321c61f021217e8d56f18d6619d624

Thanks!

PR #226, opened 2016-03-13
apple: Add TIOC* constants

m4rw3r:
From `sys/ioctl.h`.

Extracted on OS X 10.11.3.


alexcrichton:
Looks like some 32-bit values may have some different values? Otherwise looks good to me!

alexcrichton:
Thanks! Also feel free to ping PRs when they're updated, unfortunately github doesn't send out notifications for those events :(

m4rw3r:
@alexcrichton All windows builds fail on the `winapi` crate which seems to be unrelated.

PR #228, opened 2016-03-14
linux: add missing SOL_* constants

posborne:
SOL_SOCKET appears to be the only constant that is not the
same across architectures and it was already present in libc.

Signed-off-by: Paul Osborne <osbpau@gmail.com>


alexcrichton:
Looks like CI says some constants aren't defined? Perhaps a missing header to include?

posborne:
Yes, working on fixing up the tests.  Bionic/kernel define several that aren't defined for musl/glibc.  I think it is reasonable to define them all for Linux platforms, so I'll push an amended commit once I think I have things in place.

alexcrichton:
Ok, thanks! Feel free to ping the PR once tests are passing and I will r+

posborne:
Ok @alexcrichton, should be ready for review.  Appveyor is a bit backed up, but things are passing on travis.

libc-test/build.rs
```diff
@@ -289,6 +298,24 @@ fn main() {
...
+            "SOL_ROSE" if musl => true,
+
+            // these are not in glibc or musl yet (but they are in android)
+            "SOL_SCTP" |
+            "SOL_UDPLITE" |
+            "SOL_NETBEUI" |
+            "SOL_LLC" |
+            "SOL_DCCP" |
+            "SOL_NETLINK" |
+            "SOL_TIPC" if linux => true,
```
alexcrichton: Hm I'm a little wary of including constants that aren't in glibc or musl yet. Perhaps these could be android-only for now? Or should we update glibc to start verifying them?

libc-test/build.rs
```diff
@@ -289,6 +298,24 @@ fn main() {
...
             "ERROR_NOTHING_TO_TERMINATE" if mingw => true,
 
+            // these are not in musl
+            "SOL_NETLINK" |
+            "SOL_UDP" |
+            "SOL_AX25" |
+            "SOL_ATALK" |
+            "SOL_NETROM" |
+            "SOL_IPX" |
+            "SOL_ROSE" if musl => true,
```
alexcrichton: Could these constants be perhaps non-MUSL? I guess they're defined in the kernel so the MUSL definitions shouldn't change in theory?

posborne: Bionic seems to be the only libc implementation that takes a reasonable approach (automating) to keeping constants like this from the kernel reasonably up-to-date (it would be nice if this crate could move toward a similar approach for appropriate items).

These constants get passed straight to the kernel and I do not see any potential conflict with musl.  The kernel takes modification of constants like this very seriously, so I believe the risk of inclusion is extremely low.  Without definitions like this in libc, they must be defined by downstream crates (possibly inconsistently).

I don't think that the failure of musl/glibc to provide a complete set of constants to the operating system should prevent us from doing so.

libc-test/build.rs
posborne: Working with glibc to verify these seems like it might be a reasonable approach although they probably move forward at a snails pace for fear of breaking existing code that has already defined constants missing from its headers (I found several major projects defining `SOL_NETLINK` themselves, some without an `#ifndef` check).  That is, I think the lack of definition of these constants might have less to do with stability and more to do with the inability to have nice things when working in C.

Another angle would be to verify some definitions against the kernel headers.  I don't see a good way of mixing glibc headers and kernel headers together, but having a separate set of tests which verify against kernel sources might have some legs.

libc-test/build.rs
```diff
@@ -289,6 +298,24 @@ fn main() {
             "FILE_ATTRIBUTE_INTEGRITY_STREAM" |
             "ERROR_NOTHING_TO_TERMINATE" if mingw => true,
 
+            // these are not in musl
+            "SOL_NETLINK" |
```
posborne: This is not in musl or glibc and should be removed from this list. Will handle.

libc-test/build.rs
alexcrichton: Hm ok, I'm largely just wary to include such a large set of constants in the ignore list. This'd mean that they're never validated, and I don't necessarily trust architectures like MIPS to have all the same values for all the constants. It's ok if they're not tested on MUSL, but they should be tested *somewhere*.

posborne: Understood.  I'll see what I can come up with.  I would be fine with dropping all of these as well except `SOL_NETLINK` if it comes to that.  They are tested in Android, so there is some validation going on (but we do not validate Android on non-ARM).

alexcrichton: Ok, sounds good to me!

alexcrichton:

Issue #121, opened 2015-12-29
SYS_calls should be bound


nagisa:
Since there’s a binding to `syscall` function, `SYS_*` constants should also be bound. Perhaps in their own module, since there’s so many of them.
- [ ] Linux (???)
- [ ] [Android](https://android.googlesource.com/platform/prebuilts/gcc/linux-x86/host/x86_64-linux-glibc2.7-4.6/+/master/sysroot/usr/include/bits/syscall.h) where __NR are defined in unistd_{[32](https://android.googlesource.com/platform/prebuilts/gcc/linux-x86/host/x86_64-linux-glibc2.7-4.6/+/master/sysroot/usr/include/asm/unistd_32.h),[64](https://android.googlesource.com/platform/prebuilts/gcc/linux-x86/host/x86_64-linux-glibc2.7-4.6/+/master/sysroot/usr/include/asm/unistd_64.h)};
- [ ] [FreeBSD](http://fxr.watson.org/fxr/source/sys/syscall.h?v=FREEBSD10);
- [ ] [NetBSD](http://fxr.watson.org/fxr/source/sys/syscall.h?v=NETBSD);
- [ ] [OpenBSD](http://fxr.watson.org/fxr/source/sys/syscall.h?v=OPENBSD);
- [ ] [Apple](http://www.opensource.apple.com/source/xnu/xnu-1228.5.20/bsd/sys/syscall.h)

alexcrichton:
Sounds good to me, should be relatively easy enough to add at any time!

kamalmarhubi:
I am willing to take this on. I've been wanting to script getting constants anyway—at least on Linux— and this is a great excuse.

kamalmarhubi:
Also re Linux, I found the easiest place to get the sycall numbers is actually in from musl source, eg:
- http://git.musl-libc.org/cgit/musl/tree/arch/aarch64/bits/syscall.h
- http://git.musl-libc.org/cgit/musl/tree/arch/powerpc/bits/syscall.h

It'll be worth putting these in a joint musl / other location though. Probably `unix::notbsd::linux::syscalls`, which will be included by others.

alexcrichton:
For now I'd prefer to just stick with duplication, it's not that hard to have and it presents a much more understandable organization hierarchy

kamalmarhubi:
Sounds good.

posborne:
:+1: for scripting this and getting consistent, complete coverage.

Susurrus:
[...]

PR #233, opened 2016-03-17
linux: Add missing Linux-specific fcntls

kamalmarhubi:
Also move F_DUPFD_CLOEXEC up a level as it is available on Android.

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -315,8 +315,6 @@ pub const RTLD_NOW: ::c_int = 0x2;
 
 pub const TCP_MD5SIG: ::c_int = 14;
 
-pub const F_DUPFD_CLOEXEC: ::c_int = 1030;
-
 pub const PTHREAD_MUTEX_INITIALIZER: pthread_mutex_t = pthread_mutex_t {
     __align: [],
     size: [0; __SIZEOF_PTHREAD_MUTEX_T],
--- a/src/unix/notbsd/mod.rs
+++ b/src/unix/notbsd/mod.rs
@@ -161,6 +161,16 @@ pub const F_SETFD: ::c_int = 2;
 pub const F_GETFL: ::c_int = 3;
 pub const F_SETFL: ::c_int = 4;
 
+// Linux-specific fcntls
+pub const F_SETLEASE: ::c_int = 1024;
+pub const F_GETLEASE: ::c_int = 1025;
+pub const F_NOTIFY: ::c_int = 1026;
+pub const F_DUPFD_CLOEXEC: ::c_int = 1030;
+pub const F_SETPIPE_SZ: ::c_int = 1031;
+pub const F_GETPIPE_SZ: ::c_int = 1032;
+
+// TODO(#235): Include file sealing fcntls once we have a way to verify them.
+
 pub const SIGTRAP: ::c_int = 5;
 
 pub const PTHREAD_CREATE_JOINABLE: ::c_int = 0;


kamalmarhubi:
Failures are all lack of `F_ADD_SEALS` and `F_GET_SEALS`. These are available in 3.17+ according to the man page: http://man7.org/linux/man-pages/man2/fcntl.2.html

Reading that, I realize I should have also added the `F_SEAL_*` constants, which I've now done.

@alexcrichton should I keep these constants in and modify the test code to ignore them?

kamalmarhubi:
Well, I modified test code to ignore them in the meantime in case you choose to r+ it. At least then tests should be passing when @bors gets to it. :-)

posborne:
Related to https://github.com/rust-lang/libc/pull/228, I'm seeing if I can come up with a decent solution for verifying some constants against the kernel headers themselves.  There are few challenges there:

* Getting an appropriate version of the test headers into the test environment
* Setting up defines so that the appropriate defines are in place for the targeted architecture such that the kernel definitions we compare against are accurate.
* Ignoring the many things we probably don't want to check directly against the kernel.

I haven't done much in the way of implementation towards this goal as of yet, but my gut tells me that some modifications to ctest will be in order to only _include_ (rather than exclude as is the case presently) certain symbols/patterns in what is tested.

Even if some of these things end up being definitions we want to include in dowstream crates, the ability to verify that constants and other items match the operating systems C definitions should be very useful.

kamalmarhubi:
@posborne I was thinking something along these lines, but going to generation rather than verification. This also fits in with #121 (adding all SYS_call constants for all platforms). Opened #234 to figure out an approach.

posborne:
Thanks for opening #234.  I would be in favor of an approach using generation -- might be worth studying what is done in bionic.

alexcrichton:
I'd like to avoid adding constants to libc that are entirely untested in swaths like this, ignore a constant is basically a "massive hammer" to a solution.

If these aren't present in header files, how does C access these constants? That is, if we can't verify them, how can anyone use them?

posborne:
@alexcrichton The unfortunate answer for most C code is that you end up copying the header out of the kernel and into your application -- I have had to do this many times and it always bothered me.  If you are lucky, there is a library (for example, `libnl`) that was developed along side the kernel functionality which provides headers for you.

Starting with kernel 3.7, there has been a push to separate the headers that are suitable for inclusion from userspace (or at least from libc) into the [`include/uapi`](https://github.com/torvalds/linux/tree/master/include/uapi) and [`arch/$ARCH/include/uapi`](https://github.com/torvalds/linux/tree/master/arch/mips/include/uapi/asm) directories in the kernel.  This is the case for the constants in this PR: http://lxr.free-electrons.com/source/include/uapi/linux/fcntl.h.

The idea has [been around for awhile](https://lwn.net/Articles/507794/), but it recently got a bigger push.  Pinging @mkerrisk to see if he has any advice in this regard for libc authors (in this case, in a non-C based language).

alexcrichton:
Interesting, thanks for the info @posborne! If those headers are explicitly intended to be included by userspace, perhaps this PR could start including that? Or does the travis builder not have those headers yet?

kamalmarhubi:
I didn't know too much about this, thanks for the links @posborne.

From reading a bit more, it appears that these uapi headers end up being installed under `/usr/include/linux/`. However, including that file results in conflicts with the original `fcntl.h`, which is included via several other headers. To get this to work, we'd have to make a bunch of changes to what headers get used in verification on Linux.

For the sake of this change, and the constants I actually want right now, I've deferred the file sealing fcntls and associated bitflag constants.

I opened #235 for the greater change.

kamalmarhubi:
Testing including them in Android and musl. musl should definitely have them; not 100% sure of Android.

kamalmarhubi:
Androind and musl failed; reverted to omitting them.

PR #238, opened 2016-03-20
linux: Added SCHED_DEADLINE support

tljubej:
In kernel 3.14 a new scheduler policy was added.

A new constant SCHED_DEADLINE and a new struct sched_attr was added along with new API functions sched_setattr and sched_getattr.

I am not sure if I put everything in it's correct place or if this is even a good idea since it is a very new addition to the kernel and it will not be usable on slightly older kernels.

Apologies if I did anything terribly wrong I'm new to Rust and it's community!

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -170,6 +170,17 @@ s! {
         pad: [::c_long; 4]
     }
 
+    pub struct sched_attr {
+        pub size: ::uint32_t,
+        pub sched_policy: ::uint32_t,
+        pub sched_flags: ::uint64_t,
+        pub sched_nice: ::int32_t,
+        pub sched_priority: ::uint32_t,
+        pub sched_runtime: ::uint64_t,
+        pub sched_deadline: ::uint64_t,
+        pub sched_period: ::uint64_t
+    }
+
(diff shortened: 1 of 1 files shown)


alexcrichton:
Looks like there's some Linux errors to take care of here?

tljubej:
Yes but I don't really know what to do here. Shouldn't there be only linker errors at most?

It's also possible that your build environment has an older kernel which doesn't yet expose the new interface.

It's possible that this pull request should wait a bit, it seems that it isn't even implemented in glibc, at least I can't find it in the documentation.

alexcrichton:
We do have a number of environments, you can see them constructed in `.travis.yml`, but it looks like the header with `sched_attr` isn't being included? The compile errors look like they're indicating that struct is incomplete.

tljubej:
That's the thing, it doesn't exist anywhere in glibc, it's currently only accessible as a syscall.

I'm afraid this PR should be put on hold until it gets implemented in glibc.

alexcrichton:
Ah ok, there's been some discussion of including some specific linux kernel headers to verify against, but for now yeah unfortunately this may not be able to land.

alexcrichton:
Closing due to inactivity for now, but feel free to resubmit when tests are passing!

PR #244, opened 2016-03-30
Add lconv type, and locale support for UNIX

mrhota:
We have lconv struct, setlocale(), and localeconv(). Constants live
under notbsd for now.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -38,6 +38,7 @@ fn main() {
     cfg.header("errno.h")
        .header("fcntl.h")
        .header("limits.h")
+       .header("locale.h")
        .header("stddef.h")
        .header("stdint.h")
        .header("stdio.h")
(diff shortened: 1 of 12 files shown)


mrhota:
looks like I need to pull the `struct lconv` def down into the tree. NetBSD, Apple, and musl didn't seem to like the layout of the struct.

mrhota:
musl doesn't define the additional locale categories, `LC_NAME`, `LC_IDENTIFICATION`, `LC_ADDRESS`, `LC_MEASUREMENT`, `LC_PAPER`, and `LC_TELEPHONE`. The const defs need to be pulled down lower in the hierarchy.

alexcrichton:
Yeah this all looks good to me once the CI is passing

mrhota:
@alexcrichton I pushed the defs around, expanding the size of this addition quite a bit. I'm not sure if I should be concerned about the obvious duplication between some platforms.

alexcrichton:
@bors: r+ 5399c89f603007b12d675ea3c678ec75e6aa3789

Looks good to me, thanks @mrhota! No need to worry about duplication, that's just an inevitability of libc :)

PR #246, opened 2016-03-31
unix: Add openpty(3) and forkpty(3) for non-Apple platforms

kamalmarhubi:
The functions were added for Apple in #202. Adding them to other
platforms was pending an amendment to RFC 1291 to expand the scope of
libc to include libutil. The amendment was merged as
  https://github.com/rust-lang/rfcs/pull/1529

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -130,6 +130,12 @@ fn main() {
 
     if bsdlike {
         cfg.header("sys/event.h");
+
+        if freebsd {
+            cfg.header("libutil.h");
+        } else {
+            cfg.header("util.h");
+        }
     }
 
     if linux {
@@ -139,6 +145,7 @@ fn main() {
         cfg.header("sys/xattr.h");
         cfg.header("sys/ipc.h");
         cfg.header("sys/shm.h");
+        cfg.header("pty.h");
     }
 
     if linux || android {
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks! Can you add the link-to-libutil in the same place that we link to libc? That way we won't double-link it as the standard library will already be linking to it.

kamalmarhubi:
Is f0d652f what you meant?

I'll squash before merge.

kamalmarhubi:
> Is f0d652f what you meant?

Evidently not quite; the linkers are not happy.

kamalmarhubi:
Or maybe the issue is that CI should somehow not build with stdbuild?

alexcrichton:
Hm right, we're adding a new library so the libstd libc doesn't link these just yet. I don't believe there's any harm for linking these libraries twice, so yeah I think that we'll need to add link directives in the libc crate itself for them for now.

It's ok to not extract out the functions into a separate `extern` block, let's just attach the `#[link]` directives to a one-line `extern{}` statement for now (or on top of the already-large list of functions)

kamalmarhubi:
Fixed (I think), and squash / rebased. :-)

kamalmarhubi:
Missed the link attribute on freebsd.

Issue #235, opened 2016-03-18
Use linux/* headers for linux and android


kamalmarhubi:
This comes out of discussion on #233. Additions to Linux system APIs are included in headers like `<linux/header.h>` rather than `<header.h>`. We should use these headers for verification on Linux, in order to include constants that are not present in `<header.h>`.

@posborne provided some useful background:

> Starting with kernel 3.7, there has been a push to separate the headers that are suitable for inclusion from userspace (or at least from libc) into the include/uapi and arch/$ARCH/include/uapi directories in the kernel. This is the case for the constants in this PR: http://lxr.free-electrons.com/source/include/uapi/linux/fcntl.h.
> 
> The idea has [been around for awhile](https://lwn.net/Articles/507794/), but it recently got a bigger push.

As a specific example from #233:

```
$ printf '#define _GNU_SOURCE\n#include <fcntl.h>' | cpp -dM - | grep '#define F_
GET_SEAL'
$ printf '#define _GNU_SOURCE\n#include <linux/fcntl.h>' | cpp -dM - | grep '#def
ine F_GET_SEAL'
#define F_GET_SEALS (F_LINUX_SPECIFIC_BASE + 10)
```

kamalmarhubi:
> From reading a bit more, it appears that these uapi headers end up being installed under `/usr/include/linux/`. However, including that file results in conflicts with the original `fcntl.h`, which is included via several other headers. To get this to work, we'd have to make a bunch of changes to what headers get used in verification on Linux.
―[@kamalmarhubi on #233](https://github.com/rust-lang/libc/pull/233#issuecomment-198217306)

posborne:
> However, including that file results in conflicts with the original fcntl.h, which is included via several other headers

You Betcha (I'm feeling very Minnesotan today).  This issue will be commonplace when using these headers.  My initial thought on this is to have a separate test function with its own `ctest::TestGenerator` instance that includes the Linux UAPI headers and not ones from $libc.  The problem here is that we end up with two sets of tests that don't play particularly nicely together.

If ctest were modified to support an inclusive approach (in addition to the exclusive model that exists currently), I think it might be sane to do the following:

* Define the set of constants, structures, etc. that are to be verified using the Linux headers (possibly with some kind of pattern matching for constants to reduce duplication)
* Exclude these from the linux/musl verification against libc compatability TestGenerator
* Include these for the Linux compatibility TestGenerator

[...]

PR #247, opened 2016-03-31
Add clock_getres(3) and improve clockid_t constants.

nodakai:
This is still a draft

### DIFF
--- a/src/unix/notbsd/mod.rs
+++ b/src/unix/notbsd/mod.rs
@@ -5,6 +5,7 @@ pub type pthread_key_t = ::c_uint;
 pub type speed_t = ::c_uint;
 pub type tcflag_t = ::c_uint;
 pub type loff_t = ::c_longlong;
+pub type clockid_t = ::c_int;
 
 pub enum timezone {}
 
@@ -203,8 +204,20 @@ pub const SIGTRAP: ::c_int = 5;
 pub const PTHREAD_CREATE_JOINABLE: ::c_int = 0;
 pub const PTHREAD_CREATE_DETACHED: ::c_int = 1;
 
-pub const CLOCK_REALTIME: ::c_int = 0;
-pub const CLOCK_MONOTONIC: ::c_int = 1;
+pub const CLOCK_REALTIME: clockid_t = 0;
+pub const CLOCK_MONOTONIC: clockid_t = 1;
+pub const CLOCK_PROCESS_CPUTIME_ID: clockid_t = 2;
+pub const CLOCK_THREAD_CPUTIME_ID: clockid_t = 3;
+pub const CLOCK_MONOTONIC_RAW: clockid_t = 4;
+pub const CLOCK_REALTIME_COARSE: clockid_t = 5;
+pub const CLOCK_MONOTONIC_COARSE: clockid_t = 6;
+pub const CLOCK_BOOTTIME: clockid_t = 7;
+pub const CLOCK_REALTIME_ALARM: clockid_t = 8;
+pub const CLOCK_BOOTTIME_ALARM: clockid_t = 9;
+// TODO(#247) Someday our Travis shall have glibc 2.21 (released in Sep
+// 2014.) See also musl/mod.rs
+// pub const CLOCK_SGI_CYCLE: clockid_t = 10;
+// pub const CLOCK_TAI: clockid_t = 11;
 
 pub const RLIMIT_CPU: ::c_int = 0;
 pub const RLIMIT_FSIZE: ::c_int = 1;
@@ -666,7 +679,8 @@ extern {
     pub fn fdatasync(fd: ::c_int) -> ::c_int;
     pub fn mincore(addr: *mut ::c_void, len: ::size_t,
                    vec: *mut ::c_uchar) -> ::c_int;
-    pub fn clock_gettime(clk_id: ::c_int, tp: *mut ::timespec) -> ::c_int;
+    pub fn clock_getres(clk_id: clockid_t, tp: *mut ::timespec) -> ::c_int;
+    pub fn clock_gettime(clk_id: clockid_t, tp: *mut ::timespec) -> ::c_int;
     pub fn prctl(option: ::c_int, ...) -> ::c_int;
     pub fn pthread_getattr_np(native: ::pthread_t,
                               attr: *mut ::pthread_attr_t) -> ::c_int;
--- a/src/unix/solaris/mod.rs
+++ b/src/unix/solaris/mod.rs
@@ -1,6 +1,7 @@
 pub type c_char = i8;
 pub type c_long = i64;
 pub type c_ulong = u64;
+pub type clockid_t = ::c_int;
 
 pub type blkcnt_t = i64;
 pub type clock_t = i64;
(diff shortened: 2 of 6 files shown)


nodakai:
@nbaksalyar It would be great if you could check the Solaris part. I don't quite understand what's available as `clockid_t` and what is not on Solaris.

Memo: checker tool
```rust
use std::time::Instant;

#[allow(non_camel_case_types)]
type clockid_t = i32;

extern {
    fn clock_gettime(clkid: clockid_t, t: *mut Instant) -> i32;
    fn clock_getres(clkid: clockid_t, t: *mut Instant) -> i32;
}

fn f(clkid: clockid_t) {
    let mut res = Instant::now();
    let mut now0 = Instant::now();
    let mut now1 = Instant::now();
    assert_eq!(0, unsafe { clock_getres(clkid, &mut res as *mut _) } );
    assert_eq!(0, unsafe { clock_gettime(clkid, &mut now0 as *mut _) } );
    assert_eq!(0, unsafe { clock_gettime(clkid, &mut now1 as *mut _) } );

    println!("{}: {:?} {:?} {:?} {:?}", clkid, res, now0, now1, now1 - now0);
}

fn main() {
    f(0);
    f(1);
    f(2);
    f(3);
}
```

alexcrichton:
We actually need to avoid `enum` as it's basically pretty easy to violate memory safety in the FFI case (as FFI can return any value). Perhaps the can all be `pub const` instead?

src/unix/notbsd/mod.rs
```diff
@@ -176,8 +176,44 @@ pub const SIGTRAP: ::c_int = 5;
...
+/// # Examples
+/// ```
+/// fn main() {
+///     let mut ts = libc::timespec { tv_sec: 0, tv_nsec: 0 };
+///     let rc = unsafe { libc::clock_gettime(libc::CLOCK_REALTIME, &mut ts) };
+///     assert_eq!(0, rc);
+///     let rc = unsafe { libc::clock_gettime(libc::CLOCK_MONOTONIC_COARSE, &mut ts) };
+///     assert_eq!(0, rc);
+/// }
+/// ```
```
alexcrichton: You can leave out the documentation in this library as these are all just straight bindings to libc, no need to document how to use each and every one

src/unix/solaris/mod.rs
```diff
@@ -752,7 +764,8 @@ extern {
     pub fn ioctl(fildes: ::c_int, request: ::c_int, ...) -> ::c_int;
     pub fn mprotect(addr: *const ::c_void, len: ::size_t, prot: ::c_int)
                     -> ::c_int;
-    pub fn clock_gettime(clk_id: ::c_int, tp: *mut ::timespec) -> ::c_int;
+    pub fn clock_getres(clk_id: ::clockid_t, tp: *mut ::timespec) -> ::c_int;
+    pub fn clock_gettime(clk_id: ::clockid_t, tp: *mut ::timespec) -> ::c_int;
```
alexcrichton: This is another example of a breaking change which we can't take into libc, but if this uses `c_int` as `clockid_t` it'll be fine.

nodakai:
@alexcrichton

> it's basically pretty easy to violate memory safety in the FFI case (as FFI can return any value)

Can you elaborate? How can `const X: i32` contribute to safety/robustness in a way that `#[repr(i32)] enum E { X ... }` cannot?

alexcrichton:
If a C function returns an enum then it can actually return any integer. In Rust it's not memory safe to have a value of type `T` where `T` is an `enum` and the value is outside the range.

nodakai:
Right, I had only passing an enum to C API in mind, like in this case of `clockid_t` which is returned from no known C API except OpenBSD [`clock_getcpuclockid(3)`](http://man.openbsd.org/OpenBSD-current/man3/clock_getcpuclockid.3). Do you think there's still a large risk here?

alexcrichton:
Yeah unfortunately there's still the problems of:

* This is a breaking change as the type accepted is no longer `i32` but an emum
* This isn't forward compatible if new clock ids are added over time because crates can exhaustively match the Rust enum

nodakai:
> if new clock ids are added over time

Well, then a downstream user *should* stop and think how to deal with the newly added case, so isn't breakage rather desirable? I'd say it's only propagating a semantic incompatibility brought by the OS.

Anyways, you're telling me this project should be a *literally* "raw bindings" to the system libc (and such considerations should perhaps be addressed by another, high-level wrapper of libc.) Fair enough, I reverted them back to plain int `pub const`s.

Any comments on constants that might not exist in older kernels? Of course an user is expected to check the return value of APIs, but I just want to know about the general principles for handling backward-incompatibilities if there are any.

nodakai:
More than half of Travis failures came from the style checker (on comment lines longer than 80 chars in length,) BUT others indeed highlighted a compatibility issue. For example, the aarch64 cross tools shipped with Trusty/14.04 seems to lack `CLOCK_TAI`

- https://travis-ci.org/rust-lang/libc/jobs/119964887#L447

when the native build tools had no issues with it:

- https://travis-ci.org/rust-lang/libc/jobs/119964881

@alexcrichton Any suggestions? It's true that few people will complain when `CLOCK_TAI` is missing from  the `libc` crate.

alexcrichton:
> Anyways, you're telling me this project should be a literally "raw bindings" to the system libc

Yes

> Any comments on constants that might not exist in older kernels?

It's the job of consumers of libc to handle this, we simply bind everything and assume the kernel retains ABI compatibility with older releases.

> It's true that few people will complain when CLOCK_TAI is missing from the libc crate.

It's fine to leave out this constant on aarch64 for now. We may just need to update the aarch64 glibc or something like that

posborne:
> Anyways, you're telling me this project should be a literally "raw bindings" to the system libc (and such considerations should perhaps be addressed by another, high-level wrapper of libc.)

Hi @nodakai.  Providing greater safety on top of the libc APIs is the primary goal of the nix project (https://github.com/nix-rust/nix).  We would welcome changes there as well once the core bits have made it into libc.  Nix is the core behind a number of libraries such as mio (at least on *nix).

CC @kamalmarhubi

nodakai:
@posborne Thanks, I'll consider merging some of my private binding to it

@alexcrichton Now all Travis failures come from missing `CLOCK_TAI` (as well as similar but much less relevant `CLOCK_SGI_CYCLE`) which requires glibc 2.21 and Linux 3.10 or later.

src/unix/solaris/mod.rs
```diff
@@ -540,8 +541,20 @@ pub const PTHREAD_STACK_MIN: ::size_t = 4096;
...
+// https://illumos.org/man/3c/clock_gettime
+// https://github.com/illumos/illumos-gate/
+//   blob/HEAD/usr/src/lib/libc/amd64/sys/__clock_gettime.s
+// clock_gettime(3c) doesn't seem to accept anything other than CLOCK_REALTIME
+// or __CLOCK_REALTIME0
+//
+// https://github.com/illumos/illumos-gate/
+//   blob/HEAD/usr/src/uts/common/sys/time_impl.h
+// Confusing! CLOCK_HIGHRES==CLOCK_MONOTONIC==4
+// __CLOCK_REALTIME0==0 is an obsolated version of CLOCK_REALTIME==3
```
kamalmarhubi: typo: obsoleted

kamalmarhubi:
@nodakai just to set expectations, we're still figuring out how to handle enum-style constants in nix (https://github.com/nix-rust/nix/issues/254). We'd definitely welcome experiences and ideas on how to to that though, so a PR would be quite welcome!

Also thanks for adding the new stuff in libc. I was unaware of `CLOCK_TAI` on Linux. It appears its in `<linux/time.h>` and not in `<time.h>`, which brings to mind #235.

alexcrichton:
Perhaps we can just leave out the constants for platforms that don't have it yet? I'd also remove the comments here as well as the duplication across the source isn't buying us much and this library isn't really intended to be read for documentation.

nodakai:
@alexcrichton Moved `CLOCK_TAI` to `musl/mod.rs` though it should in principle be as portable as Linux kernel is. Also removed doc comments.

PR #251, opened 2016-04-02
Additional pthread functions for DragonFly.

mneumann:
These are needed in order to fix/implement the guard page code of Rust
for DragonFly.

### DIFF
--- a/src/unix/bsd/freebsdlike/dragonfly/mod.rs
+++ b/src/unix/bsd/freebsdlike/dragonfly/mod.rs
@@ -115,4 +115,12 @@ extern {
     pub fn mprotect(addr: *mut ::c_void, len: ::size_t, prot: ::c_int)
                     -> ::c_int;
     pub fn clock_gettime(clk_id: ::c_ulong, tp: *mut ::timespec) -> ::c_int;
+
+    pub fn pthread_attr_get_np(tid: ::pthread_t,
+                               attr: *mut ::pthread_attr_t) -> ::c_int;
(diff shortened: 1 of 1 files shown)


mneumann:
@dhuseby: This probably applies to FreeBSD as well.

src/unix/bsd/freebsdlike/dragonfly/mod.rs
```diff
@@ -115,4 +115,12 @@ extern {
     pub fn mprotect(addr: *mut ::c_void, len: ::size_t, prot: ::c_int)
                     -> ::c_int;
     pub fn clock_gettime(clk_id: ::c_ulong, tp: *mut ::timespec) -> ::c_int;
+
+    pub fn pthread_attr_get_np(tid: ::pthread_t,
```
dhuseby: do these functions have the same signature on both dragonflybsd and freebsd?  if they do, then please put them in the src/unix/bsd/freebsdlike/mod.rs file instead.  thanks!

mneumann: yes. https://www.freebsd.org/cgi/man.cgi?query=pthread_attr_get_np. Also, guard page code for FreeBSD seems to be missing in Rust. Did you run "gmake check" when building rust? One test that overflows the stack on purpose should also fail for FreeBSD (it fails on DragonFly). I will add those functions then to the common code base for both DragonFly and FreeBSD

mneumann:
@alexcrichton : I will come up with a patch that also includes support for FreeBSD. Closing this pull request.

PR #252, opened 2016-04-03
Add several pthread functions for DragonFly/FreeBSD

mneumann:
These are needed in order to fix/implement the guard page code of Rust
for DragonFly.


mneumann:
Similar to https://github.com/rust-lang/libc/pull/251, just that it is for both DragonFly _and_ FreeBSD.

PR #255, opened 2016-04-07
Add ucontext for linux-musl.

fiveop:
It just copy pasted from the linux-other version.

### DIFF
--- a/src/unix/notbsd/linux/musl/b64/x86_64.rs
+++ b/src/unix/notbsd/linux/musl/b64/x86_64.rs
@@ -1,3 +1,18 @@
+s! {
+    pub struct mcontext_t {
+        __private: [u64; 32],
+    }
+
+    pub struct ucontext_t {
+        pub uc_flags: ::c_ulong,
+        pub uc_link: *mut ucontext_t,
+        pub uc_stack: ::stack_t,
+        pub uc_mcontext: mcontext_t,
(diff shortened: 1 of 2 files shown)


fiveop:
Given that musl certainly exports these methods

    $ git grep getcontext # in a clone of musl's master branch
    include/ucontext.h:int  getcontext(struct __ucontext *);

and `ucontext.h` is imported on linux indiscriminately (see lib-test\build.rs), why does the build fail because of undefined references to this (and similar) functions ( https://travis-ci.org/rust-lang/libc/jobs/121508994 )?

alexcrichton:
I believe that while declared in MUSL the functions aren't actually defined.

fiveop:
You are right: http://wiki.musl-libc.org/wiki/Open_Issues#ucontext.h . But It does provide the struct, so I just removed the functions declarations.

src/unix/notbsd/linux/musl/b64/x86_64.rs
```diff
@@ -1,3 +1,18 @@
 pub const SYS_gettid: ::c_long = 186;
 
 pub const SYS_perf_event_open: ::c_long = 298;
+
+s! {
```
alexcrichton: Currently structs come before constants, but otherwise this looks good to me!

fiveop:
done

PR #257, opened 2016-04-10
Add Linux-specific recvmmsg(2) and sendmmsg(2)

nodakai:
glibc has almost harmless bugs in their prototype declarations.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -343,6 +343,11 @@ fn main() {
             "setrlimit" | "setrlimit64" |    // non-int in 1st arg
             "strerror_r" if linux => true,   // actually xpg-something-or-other
 
+            // int vs uint. Sorry musl, your prototype declarations are "correct" in the sense that
+            // they match the interface defined by Linux verbatim, but they conflict with other
+            // send*/recv* syscalls
+            "sendmmsg" | "recvmmsg" if musl => true,
+
             // typed 2nd arg on linux and android
             "gettimeofday" if linux || android || freebsd || openbsd || dragonfly => true,
 
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -181,6 +181,11 @@ s! {
         #[cfg(target_pointer_width = "64")]
         bits: [u64; 16],
     }
+
+    pub struct mmsghdr {
+        pub msg_hdr: ::msghdr,
+        pub msg_len: ::c_uint,
+    }
 }
 
 pub const ABDAY_1: ::nl_item = 0x20000;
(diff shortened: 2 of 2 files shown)


nodakai:
`recvmmsg(2)` and `sendmmsg(2)` actually require musl-libc v1.1.3 while Trusty has only v0.9.15:

```
$ git describe --contains acb7e049
v1.1.3~16                         
$ git sh v1.1.3
commit de337c9314e4dede7b26e03f29cf768e7b0615b5
Author: Rich Felker <dalias@aerifal.cx>
Date:   Wed Jun 25 16:18:05 2014 -0400
...
```

- https://travis-ci.org/rust-lang/libc/jobs/122061324

```
Get:1 http://us.archive.ubuntu.com/ubuntu/ trusty/universe musl amd64 0.9.15-1 [299 kB]
```

alexcrichton:
Hm we should probably just install musl from source anyway, or use a docker image that already has musl installed. This may be able to set the `DOCKER` variable for the musl build pointing at the [rust-slave-linux](https://hub.docker.com/r/alexcrichton/rust-slave-linux/tags/) container which is what we use for bulidbot (and should have a newer musl). Maybe that'd work?

libc-test/build.rs
```diff
@@ -341,7 +341,9 @@ fn main() {
 
             "getrlimit" | "getrlimit64" |    // non-int in 1st arg
             "setrlimit" | "setrlimit64" |    // non-int in 1st arg
-            "strerror_r" if linux => true,   // actually xpg-something-or-other
+            "strerror_r" |                   // actually xpg-something-or-other
+            "sendmmsg" | "recvmmsg"          // glibc header has wrong declarations
```
alexcrichton: Out of curiosity, what's the wrong declaration in glibc?

nodakai: It's `unsigned int flags` vs `int flags`.  I filed a bug at https://sourceware.org/bugzilla/show_bug.cgi?id=19936

alexcrichton:
I'm not sure we'll want to add an exception here for these functions, the flags argument is signed for other functions like `send`, `sendmsg`, and `sendto`. Any constants defined for those can presumably be passed here as well, and if we were to switch the types then it'll cause unnecessary compile failures.

nodakai:
The original Linux syscall and musl use `unsigned int flags` whereas glibc uses `int flags`. Either of them has to be skipped in the test.

alexcrichton:
Not necessarily, the function has a different signature on musl than it does on glibc, so that just means it needs to be defined differently

nodakai:
I come to realize that sharing the `MSG_` constants (I'll add them in a separate PR) with other `recv*` syscalls is an essential design requirement here, and that the Linux kernel team, not the glibc team, has made a poor decision which doesn't really impact C programmers.

So we should perhaps skip the test only for musl. How do you think about this?

alexcrichton:
Yeah skipping the test on musl should be ok, we unfortunately seem to have to skip it *somewhere*

nodakai:
- https://travis-ci.org/rust-lang/libc/jobs/125989709

> Unpacking libc6-dev:amd64 (2.19-0ubuntu6.7) over (2.19-0ubuntu6.6) ...

glibc older than 2.21 (Feb 2015) has wrong prototypes where `struct timespec` are `const struct timespec`

alexcrichton:
Perhaps some of these could switch to running inside of a ubuntu 16.04 docker container? Is that recent enough to have a new-enough glibc?

alexcrichton:
In https://github.com/rust-lang/libc/pull/316 I've updated how we run all tests to much more easy to update each container, so feel free to update this after that lands.

nodakai:
Sorry for not updating this for a long time... I'll take a look at the new test driver.

alexcrichton:

tokenrove:
Is there any chance of reopening this?  I need these syscalls.  Is there something I can do to help get this in?  (I rebased against master at https://github.com/tokenrove/libc/tree/recvmmsg-sendmmsg and have been using that branch successfully on x86_64-unknown-linux-musl)

alexcrichton:
@tokenrove certainly yes feel free! Want to just send a PR?

PR #261, opened 2016-04-12
64-bit Android timegm fix

myfreeweb:
On aarch64 and x86_64, `timegm64` does not exist (see [time64.h](https://android.googlesource.com/platform/development/+/797351fd3bbb8fe517afafdd5095fd740387e7a4/ndk/platforms/android-L/include/time64.h)).

See rust-lang-deprecated/time#118 for usage. Since that patch, the time library switched to using libc for `timegm` (rust-lang-deprecated/time@51c0019), and it doesn't build on 64-bit Android. This PR fixes that.

### DIFF
--- a/src/unix/notbsd/android/b64.rs
+++ b/src/unix/notbsd/android/b64.rs
@@ -10,3 +10,7 @@ s! {
 }
 
 pub const SYS_gettid: ::c_long = 178;
+
+extern {
+    pub fn timegm(tm: *const ::tm) -> ::time64_t;
+}
--- a/src/unix/notbsd/android/b32.rs
+++ b/src/unix/notbsd/android/b32.rs
@@ -10,3 +10,7 @@ s! {
 }
 
 pub const SYS_gettid: ::c_long = 224;
+
+extern {
+    pub fn timegm64(tm: *const ::tm) -> ::time64_t;
+}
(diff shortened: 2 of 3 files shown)


myfreeweb:
okay, lint didn't like `#[cfg]`, put the function in `b32.rs`/`b64.rs` (I was initially afraid of the circular dependency, but it only needs `time64_t`, which I just put into these modules too)

src/unix/notbsd/android/b64.rs
```diff
@@ -10,3 +11,7 @@ s! {
 }
 
 pub const SYS_gettid: ::c_long = 178;
+
+extern {
+    pub fn timegm(tm: *const ::tm) -> time64_t;
```
alexcrichton: You should actually be able to change these references to `::time64_t` and avoid the duplicate definition of `time64_t` above I believe

myfreeweb: Fixed. Looks like I'm just too used to Haskell, where circular dependencies are not supported :D

alexcrichton:
@bors: r+ 6b0e09c1f417ecb51e16cdf4be41836db729c6e6

FWIW I think circular references in resolve are handled much better today in recent Rust than they were in 1.0, for example. Although I have yet to test them out on this crate just yet :)

PR #263, opened 2016-04-12
Add the family of printf, scanf functions

bluss:
snprintf is apparently supported in MSVC since version 14. I'm curious to take the tests for a spin.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -166,16 +166,6 @@ extern {
                    mode: c_int,
                    size: size_t) -> c_int;
     pub fn setbuf(stream: *mut FILE, buf: *mut c_char);
-
-    pub fn fprintf(stream: *mut FILE, format: *const c_char, ...) -> c_int;
-    pub fn printf(format: *const c_char, ...) -> c_int;
-    pub fn snprintf(s: *mut c_char, n: size_t,
-                    format: *const c_char, ...) -> c_int;
-    pub fn sprintf(s: *mut c_char, format: *const c_char, ...) -> c_int;
-    pub fn fscanf(stream: *mut FILE, format: *const c_char, ...) -> c_int;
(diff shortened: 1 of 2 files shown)


alexcrichton:
@bors: r+ adb400dd0ef2089668f11ab013ce3ced74b8ee1c

Sure! Let's see what happens

bluss:
Here's some more information about the warning https://msdn.microsoft.com/en-us/library/a98sb923.aspx

Exactly the new functions are warned for. They are all variadic, unlike the already present ones in src/lib.rs. There are two variadic functions in src/windows.rs already though.

bluss:
Ok, so the windows situation is not well known to me. I can not fix that here since I don't have a windows environment to play with, so that part will be left to another contributor.

alexcrichton:
Ok, I'm fine with that too! [This documentation](https://msdn.microsoft.com/en-us/library/yd3056cz.aspx?f=255&MSPPError=-2147217396) seems to indicate that the warning is off by default, so it's probably [one of our warning-enablers](https://github.com/alexcrichton/ctest/blob/965d6571ef4cfa0a911e89cee11e2f3cb211d6f0/src/lib.rs#L567) that's turning that on. I'd also be fine just ignoring this warning with `/wd4710` there as well.

Have a preference on which route to take?

alexcrichton:
oops

bluss:
Ah I didn't find that code, with windows specific warnings, thank you for the link.

Discussing with @nagisa it seems the functions don't exist under those names in the windows library, so it seems it would be incorrect to add them.

alexcrichton:
@bors: r+ 1ebfe63f1cf0ec2532d269883243aabbd1bf1a17

Ok, sounds good to me!

bluss:
Thanks. Incremental improvements is the open source way. And /me bravely ran away.

PR #264, opened 2016-04-15
Add support for arm-unknown-linux-musleabi{,hf} targets.

timonvo:
These targets will be similar to the x86_64-unknown-linux-musl
target, in that they'll use MUSL libc to allow fully static binaries
to be generated. To remain consistent with the naming of existing
ARM targets, as well as with the standard naming of MUSL toolchains,
we'll use `musleabi` and `musleabihf` as the target environment
names (analogous to `gnueabi` and `gnueabihf`).

Most of these changes just extend the special casing for x86_64 MUSL
targets to the ARM ones as well.

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -55,7 +55,9 @@ s! {
         pub ru_nvcsw: c_long,
         pub ru_nivcsw: c_long,
 
-        #[cfg(target_env = "musl")]
+        #[cfg(any(target_env = "musl",
+                  target_env = "musleabi",
+                  target_env = "musleabihf"))]
         __reserved: [c_long; 16],
     }
 
@@ -149,8 +151,9 @@ cfg_if! {
     } else if #[cfg(all(not(stdbuild), feature = "use_std"))] {
         // cargo build, don't pull in anything extra as the libstd  dep
         // already pulls in all libs.
-    } else if #[cfg(all(target_env = "musl", not(any(target_arch = "mips",
-                                                     target_arch = "arm"))))] {
+    } else if #[cfg(any(all(target_env = "musl", not(target_arch = "mips")),
+                        target_env = "musleabi",
+                        target_env = "musleabihf"))] {
         #[link(name = "c", kind = "static")]
         extern {}
     } else if #[cfg(target_os = "emscripten")] {
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -93,9 +93,13 @@ s! {
     }
 
     pub struct pthread_cond_t {
-        #[cfg(target_env = "musl")]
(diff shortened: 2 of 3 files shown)


alexcrichton:
Thanks for the PR! The compiler doesn't currently recognize these triples (and transitively rustdoc doesn't) so this may have to leave out the doc links for now.

I also wonder if we should just define `target_env` as `"musl"` for all of these triples and provide the hard/soft float option via a different method... (cc @brson). This is fine to land for now though, it'd work in either situation :)

timonvo:
Got rid of the doc links for now. This whole separate repositories with interdepencies thing makes it quite difficult to land patches. :) I just figured landing preliminary support in the sub-repos first would be better than landing broken changes in the main repo first.

As for the target_env thing. Yes, I agree that the situation is getting unwieldy. However, in order to get my changes in I'd like to just go ahead anyway, and then improve later on. It's been taking quite some effort already to keep my changes up-to-date with master and I'd be grateful if I could finally get my preliminary changes landed.

timonvo:
Also fixed line wrap (80 chars now).

joerg-krause:
I really appreciate the support for musl on ARM and I know using `eabi` and `eabihf` is also used for musl based toolchains, but having three different kinds of target-envs for musl looks somehow weird and redundant.

timonvo:
It's no different from the three existing different `target_env`s of `gnu`, `gnueabi`, and `gnueabihf` though, all of which are also backed by the same glibc implementation yet have different `target_env`s. I agree that this doesn't quite feel right, but I feel like we shouldn't stray away from that existing convention either.

That being said, if this is really a dealbreaker I could:
- use `target_env="musl"` instead. Note though that this means we should probably change the existing `gnueabi`/`gnueabihf` targets to use `gnu` as the `target_env` though, and that may incur some changes in the code as well.
- add a new target variable (e.g. `target_libc`) to cover this case, independent of the `target_env` variable so that it can remain consistent with the toolchain naming conventions.

WDYT?

alexcrichton:
@bors: r+ ad728590d256edd7c6bb2b9908e5bfb3be2e5e0b

Thanks @timonvo! @joerg-krause I agree with the sentiment here, I definitely don't want to see `#[cfg(target_env = "musl")]` explode everywhere to cover all these as well. I think that @timonvo's diagnosis is spot on and my preferred solution as well is to change gnueabihf and gnueabi to only have "gnu" as the `target_env`, but I haven't though too too hard about doing this (it's a breaking change, so shouldn't be taken lightly).

In any case this is pretty minor for liblibc itself, we may as well help these targets come online in the first place anyway :). If we end up choosing to just use "musl" everywhere then we can revert this in libc in time.

alexcrichton:
Hm seems spurious, we'll fix tomorrow

timonvo:
Thanks for the review Alex!

PR #268, opened 2016-04-27
linux: Add prlimit(2) and prlimit64(2)

kamalmarhubi:
As with `getrlimit` and `setrlimit`, the glibc wrappers have a non-`int`
for the `resource` argument, eg:

    extern int prlimit (__pid_t __pid, enum __rlimit_resource __resource,
                        const struct rlimit *__new_limit,
                        struct rlimit *__old_limit) __THROW;

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -341,6 +341,7 @@ fn main() {
 
             "getrlimit" | "getrlimit64" |    // non-int in 1st arg
             "setrlimit" | "setrlimit64" |    // non-int in 1st arg
+            "prlimit" | "prlimit64" |        // non-int in 2nd arg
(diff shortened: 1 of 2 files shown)


kamalmarhubi:
Not sure about the android failure. These are defined in bionic: https://github.com/android/platform_bionic/blob/bca25411b882abc6ab0bf07d25266ba9f8e4d398/libc/include/sys/resource.h#L56-L57

alexcrichton:
Maybe that's not in the NDK we're using?

kamalmarhubi:
Ah interesting though. I've pushed it down to Linux-only for now.

kamalmarhubi:
83 characters!

kamalmarhubi:
Fixed and rebased.

PR #269, opened 2016-04-28
notbsd: Deduplicate definitions

kamalmarhubi:
A bunch of definitions were duplicated across, eg, android and linux.
This commit pulls these up to higher levels where they can be shared.

### DIFF
--- a/src/unix/notbsd/android/mod.rs
+++ b/src/unix/notbsd/android/mod.rs
@@ -648,30 +648,7 @@ extern {
                        serv: *mut ::c_char,
                        sevlen: ::size_t,
                        flags: ::c_int) -> ::c_int;
-    pub fn eventfd(init: ::c_uint, flags: ::c_int) -> ::c_int;
     pub fn ptrace(request: ::c_int, ...) -> ::c_long;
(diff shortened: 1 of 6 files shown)


kamalmarhubi:
I *think* this is right. Don't merge until the build is done so we can check the number of tests are the same. I definitely lost some while doing this!

screenshot from 2016-04-27 18 20 34
screenshot from 2016-04-27 18 19 22

kamalmarhubi:
(That function was found, as were 8 I lost on mips.)

kamalmarhubi:
I think there's something wrong with the current OS X nightly

> rustup: assert_nz url is empty!

kamalmarhubi:
Checked the numbers of tests on all architectures, and they match the most recent passing master build.

alexcrichton:
@bors: r+ b9750b650d10acc8f1c61a5966f309fd940f02d4

Thanks! We could consider also adding CI for detecting duplicate definitions, but also perhaps isn't worth it.

kamalmarhubi:
Yeah, I think that could be good but less important than #270.

PR #271, opened 2016-04-28
OS X: add various process status tests defined by wait.h

mbarnett:
Adds the various process test macros specified in the wait(2) manpage on OS X. ```WCOREDUMP``` is implemented BSD-wide as it seems to be the only macro whose definition is constant across the BSDs.

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -1032,20 +1032,20 @@ f! {
         status >> 8
     }
 
-    pub fn WSTATUS(status: ::c_int) -> ::c_int {
+    pub fn _WSTATUS(status: ::c_int) -> ::c_int {
         status & 0x7f
     }
 
     pub fn WIFCONTINUED(status: ::c_int) -> bool {
-        WSTATUS(status) == _WSTOPPED && WSTOPSIG(status) == 0x13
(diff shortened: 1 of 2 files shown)


src/unix/bsd/apple/mod.rs
```diff
@@ -1025,6 +1025,30 @@ pub const RTLD_NODELETE: ::c_int = 0x80;
...
 pub const RTLD_GLOBAL: ::c_int = 0x8;
 
+pub const _WSTOPPED: ::c_int = 0o177;
+
+f! {
+    pub fn WSTOPSIG(status: ::c_int) -> ::c_int {
+        status >> 8
+    }
+
+    pub fn WSTATUS(status: ::c_int) -> ::c_int {
```
alexcrichton: This doesn't look like it's a `#define` in `sys/wait.h` for me, but there's a `_WSTATUS`, so maybe this is an internal function?

alexcrichton:
Thanks! Just one small nit but otherwise lgtm

mbarnett:
Good call. I've pushed a change to rename it to _WSTATUS to match the header, but the obvious other solution here is to just repeat the ```(status & 0x7f)``` mask in each of the macros if that is undesirable.

alexcrichton:
@bors: r+ 06abf3a

Sounds good to me!

PR #277, opened 2016-05-04
Added strnlen function to all platforms.

raphaelcohn:
strnlen is used to find the length of a C string that may be
lacking a terminal NUL character. Whilst it is possible to
implement the equivalent functionality in rust code, it is
cleaner, simpler and removes the need for duplication of tricky
functionality to get right. It also makes it easier to port
C code snippets to rust. In my case, it's needed for dealing with
the result of `gethostname`.

Note that strnlen is not part of POSIX or C99, but is present on all major platforms.


alexcrichton:
@bors: r+ 29b1ceab3ee715e7919f443d41858dbc89923a06

Thanks!

PR #278, opened 2016-05-04
Adding syslog functions, constants and structs

raphaelcohn:
This commit adds the functions:-
- syslog
- openlog
- closelog
- setlogmask

It adds LOG_* constants.

It adds the `CODE` struct used by the `#define` definitions `prioritynames` and `facilitynames`.

It does not add:-
- the function `vsyslog`; this is an extension to POSIX and uses va_list macros;
- the `#define`s `prioritynames` and `facilitynames`, as usage is not common
- rust functions mirroring the macros:-
  - LOG_PRI
  - LOG_MAKEPRI
  - LOG_MASK
  - LOG_UPTO
  - LOG_FAC
* `CODE` is included in case a third-party unsafe C function returns or takes it.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -95,6 +95,7 @@ fn main() {
         cfg.header("sched.h");
         cfg.header("termios.h");
         cfg.header("poll.h");
+        cfg.header("syslog.h");
     }
 
     if android {
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -1029,6 +1029,13 @@ pub const RTLD_GLOBAL: ::c_int = 0x8;
 
 pub const _WSTOPPED: ::c_int = 0o177;
 
+pub const LOG_NETINFO: ::c_int = 12 << 3;
+pub const LOG_REMOTEAUTH: ::c_int = 13 << 3;
+pub const LOG_INSTALL: ::c_int = 14 << 3;
+pub const LOG_RAS: ::c_int = 15 << 3;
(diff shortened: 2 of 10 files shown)


alexcrichton:
Looks like a header for verification may not be included? You may need to modify `libc-test/build.rs` to add some headers here or there.

raphaelcohn:
Let me know if you want the commits squashed, although done this way, I feel it's a little clearer the level of changes required.

By the way, is there a simple way I can use `cfg` checks to determine if a function is present in libc before `use`ing it, or do I have to mirror the `cfg` checks libc uses in my consuming code? The latter seems like a subtle way to introduce breakages, eg if libc ever gained support for AIX or HP-UX (although given how they're slowly dying, probably not).

alexcrichton:
Looks like android CI is failing?

> Let me know if you want the commits squashed

Nah whatever's fine by me :)

> By the way, is there a simple way I can use cfg checks to determine if a function is present in libc before useing it, or do I have to mirror the cfg checks libc uses in my consuming code?

Unfortunately the only way to do so right now is to mirror the checks that libc has :(

raphaelcohn:
I've not got an Android set up to test this on, so it's not a big surprise. A quick check at what I think is the Android [bionic libc headers on GitHub](https://github.com/AOKP/bionic/blob/kitkat/libc/include/syslog.h) suggests `LOG_NFACILITIES`  isn't defined on that platform... I'll try again....

Oh, how I hate C's cross platform uggh-ness. May be one day when I win the lottery I'll fund a team to re-write libc et al in rust...

raphaelcohn:
Thank you.

PR #279, opened 2016-05-05
unix: Add PIPE_BUF for bsd and notbsd

kamalmarhubi:
This is the maximum size of guaranteed-atomic writes to a pipe.


kamalmarhubi:
Rebased.

PR #281, opened 2016-05-08
Getprogname

raphaelcohn:
Added `getprogname()` and `setprogname()` for all BSDs and Solaris (including Mac OS X).

Added `program_invocation_short_name` global, for Linux (glibc and Musl) which is effectively the same thing, and is what compatibility libraries like `libbsd` use to implement `getprogname()`.

Added `__progname` global for Android, which, whilst not quite the same as `getprogname` or `program_invocation_short_name`, is better than using argv[0], as it (a) avoids a common bug with no arguments (b) avoids a common bug with a NULL string in argv[0] and (c) incorporates Android's chosen name for an unknown process.

### DIFF
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -358,6 +358,8 @@ extern {
                       buf: *mut ::c_char,
                       buflen: ::size_t,
                       result: *mut *mut passwd) -> ::c_int;
+    pub fn getprogname() -> *const ::c_char;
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks! Wanna squash that last commit up into the previous?

raphaelcohn:
Will try to. I'm hors de combat until Tue...

raphaelcohn:
Just squished. Let's see how it goes...

PR #282, opened 2016-05-08
Add correctly sized structs and types for Android aarch64.

Nercury:
So far both android x86_64 and aarch64 configurations used incorrect types from 32-bit architecture.

This PR updates only the aarch64, because it is more common. The x86_64 may be also incorrect, but is out of scope of this PR.


alexcrichton:
@bors: r+ dad0dd2a50557f2e4947160e0f9687da7d14f027

Thanks! Once we get nightlies for this target I'd love to also run the CI for aarch64

PR #285, opened 2016-05-11
Added extensive constants to make use of the BSD's sysctl function.

raphaelcohn:
sysctl usage does differ significantly across the BSDs, and, whilst
some constants overlap, many do not. It is easier to maintain them
in separate modules, rather than trying to tease out common definitions.


alexcrichton:
Thanks! Looks like some of the CI failures are legitimate though?

raphaelcohn:
OK will take a look in the morning...

raphaelcohn:
OK, looks like I missed some duplicate definitions which had been added piecemeal. I'm beginning to wonder, even though it creates apparent duplication, whether it would be better to drop the attempt to create a hierarchy of unices and just push all definitions down to a per-OS file. It could make maintenance easier. Thoughts? Alternatively, a kickstarter campaign to flesh out all remaining libc bits and bobs... I'd be game for that.

alexcrichton:
Heh, yeah the organization as-is may be a bit of an opportunistic optimization that may not be worth it in the end, but we probably don't want to shift it all around *just* yet.

raphaelcohn:
Fair enough

alexcrichton:
Looks like travis is still having some problems though?

raphaelcohn:
OK, I think I've fixed this. (I must say, I'm dead impressed with your CI builds. They're _very_ effective).
Turns out the problem was I grepp'd for duplicate `pub const KERN_PROC:` but missed the ones `pub const KERN_PROC :`. Would this be worth an addition to the style checks?

raphaelcohn:
OK, looks like we're just down to bitrig failing now...I'll try again...

Nercury:
@raphaelcohn It looks like `libc::HW_NCPU` is now missing on `auto-linux-64-cross-netbsd`. Is this intentional? (issue here: https://github.com/rust-lang/rust/pull/33500#issuecomment-219263000)

alexcrichton:
Ah yeah that's accidentally, I'm adding it back in https://github.com/rust-lang/libc/pull/291

PR #288, opened 2016-05-13
Adding sysinfo() and sysinfo struct for Linux and Android.

raphaelcohn:
Sadly, the sysinfo struct varies slightly between Musl and Glibc / Bionic.

This means that users need to be careful when using the uptime, and should
always cast it to a signed value. Why uptime can be signed is beyond me...


raphaelcohn:
Silly mistake on my part...

alexcrichton:
Looks good to me! will r+ when CI is green

alexcrichton:
@bors: r+ 4d37980fc3b9b6727953592c69fda6bc9be32023

Thanks! Also feel free to ping a PR whenever you update it, unfortunately github doesn't send out notifications for an event like that :(

alexcrichton:
Ah and in the future it should be fine to rebase instead of merge (helps keep out the merge commits)

PR #293, opened 2016-05-19
Add the TIOCGWINSZ and TIOCSWINSZ constants

### LINKED ISSUE #292: libc::TIOCGWINSZ is not defined for FreeBSD targets
Despite the `ioctl` call TIOCGWINSZ being available on FreeBSD under the `sys/ioctl.h` header, it's not exposed in libc when built for FreeBSD targets.

If this isn't intentional, I was wondering if maybe it could be add for FreeBSD targets, as this is a blocking issue for a project I'm working on.

Compiling and running the following on FreeBSD should prove that TIOCGWINSZ is available on FreeBSD:

```c
#include <stdio.h>
#include <sys/ioctl.h>

int main(int argc, char **argv) {
    struct winsize sz;

    ioctl(0, TIOCGWINSZ, &sz);
    printf("Screen width: %i  Screen height: %i\n", sz.ws_col, sz.ws_row);
    return 0;
}
```

SShrike:
I've added the `TIOCGWINSZ` and `TIOCSWINSZ` constants/ioctl control codes. I wasn't quite sure where to put them, so hopefully they're in the right place.

As of now it's set to compile on 'FreeBSD-like' platforms, so it will compile on both FreeBSD *and* DragonflyBSD. I've only tested it on FreeBSD though, but it should also work on DragonflyBSD AFAIK.

Fixes #292.

### DIFF
--- a/src/unix/bsd/freebsdlike/mod.rs
+++ b/src/unix/bsd/freebsdlike/mod.rs
@@ -668,6 +668,9 @@ pub const LOG_SECURITY: ::c_int = 13 << 3;
 pub const LOG_CONSOLE: ::c_int = 14 << 3;
 pub const LOG_NFACILITIES: ::c_int = 24;
 
+pub const TIOCGWINSZ: ::c_ulong = 0x40087468;
+pub const TIOCSWINSZ: ::c_ulong = 0x80087467;
+
 #[link(name = "util")]
 extern {
     pub fn getnameinfo(sa: *const ::sockaddr,


alexcrichton:
@bors: r+ ae72fc7e43172abbda0b6b5d826db21256a233b4

Thanks!

SShrike:
Seems like build failure wasn't caused by my code.

alexcrichton:
@bors: retry

Indeed!

SShrike:
The test passed apparently: https://travis-ci.org/rust-lang/libc/builds/131547549.

SShrike:
:tada:

PR #294, opened 2016-05-19
Added nice, setpriority and getpriority along with constants and type…

raphaelcohn:
… definitions

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -19,6 +19,7 @@ pub type fsfilcnt_t = ::c_uint;
 pub type speed_t = ::c_ulong;
 pub type tcflag_t = ::c_ulong;
 pub type nl_item = ::c_int;
+pub type id_t = ::c_uint;
 
 pub enum timezone {}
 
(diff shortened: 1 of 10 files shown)


alexcrichton:
Looks like the CI failures may be legit?

raphaelcohn:
Ok will try to sort this out tomorrow.

raphaelcohn:
OK, think I've got a handle on this. The Linux failures are because glibc's signatures use an enum for which, rather than an int. Yet more gratuitous breakage which deviates from the [man pages](http://man7.org/linux/man-pages/man2/getpriority.2.html). (It seems they've [fiddled here more than once](https://stackoverflow.com/questions/16175217/unknown-type-name-priority-which-t)). I'll try to work something out for them. Fingers crossed musl eventually conquers Debian and RedHat...

The Mac OS X failure is on i686. I don't fully understand what causes the 'bad function pointer' test failure. Given that Darwin on i686 is effectively dead, if I can't fix this expediently I'll just not add nice() for this platform.

raphaelcohn:
Right, the Mac OS i686 issue seems solved. FreeBSD failures are occurring because they don't use `id_t`. Fix applied. GNU glibc issues still happen as I _think_ they're enum definition uses unsigned ints. Trying again... then I'll squash the commits.

raphaelcohn:
OK, just Android left; have adjusted getpriority / setpriority signature to use int and not id_t... Let's see what happens.

raphaelcohn:
Whoops, pushed commits I shouldn't have...

alexcrichton:
Looks like some lint errors in CI?

PR #296, opened 2016-05-20
Mac OS X specific pthread and Mach thread functions

raphaelcohn:
Also includes `SCHED_*` definitions. Tested locally.

The commented pub type definitions are because these cause breakage of the libc tests - I suspect it's in the test logic, but I'm not able to deduce how this code works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ script:
 os:
   - linux
   - osx
+osx_image: xcode6.4
 env:
   matrix:
     - ARCH=x86_64
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -125,6 +125,9 @@ fn main() {
     if apple {
         cfg.header("mach-o/dyld.h");
         cfg.header("mach/mach_time.h");
+        cfg.header("mach/thread_policy.h");
+        cfg.header("mach/thread_act.h");
+        cfg.header("mach/task_info.h");
         cfg.header("malloc/malloc.h");
         cfg.header("util.h");
         if target.starts_with("x86") {
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -288,6 +288,9 @@ fn main() {
             // uuid_t is a struct, not an integer.
             "uuid_t" if dragonfly => true,
             n if n.starts_with("pthread") => true,
+			
+			// Mac OS X
+			n if n.starts_with("thread_") => true,
 
             // windows-isms
             n if n.starts_with("P") => true,
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -29,25 +29,28 @@ pub type kern_return_t = ::c_int;
 pub type thread_latency_qos_t = ::integer_t;
 pub type thread_throughput_qos_t = ::integer_t;
 // libc-test does not like structs
-//pub type thread_policy_t = *mut integer_t;
-//pub type thread_standard_policy_data_t = thread_standard_policy;
-//pub type thread_extended_policy_data_t = thread_extended_policy;
-//pub type thread_time_constraint_policy_data_t = thread_time_constraint_policy;
-//pub type thread_precedence_policy_data_t = thread_precedence_policy;
-//pub type thread_affinity_policy_data_t = thread_affinity_policy;
-//pub type thread_background_policy_data_t = thread_background_policy;
-//pub type thread_latency_qos_policy_data_t = thread_latency_qos_policy;
-//pub type thread_throughput_qos_policy_data_t = thread_throughput_qos_policy;
+// pub type thread_standard_policy_data_t = thread_standard_policy;
+// pub type thread_extended_policy_data_t = thread_extended_policy;
+// pub type thread_time_constraint_policy_data_t =
+// thread_time_constraint_policy;
+// pub type thread_precedence_policy_data_t = thread_precedence_policy;
+// pub type thread_affinity_policy_data_t = thread_affinity_policy;
+// pub type thread_background_policy_data_t = thread_background_policy;
+// pub type thread_latency_qos_policy_data_t = thread_latency_qos_policy;
+// pub type thread_throughput_qos_policy_data_t = thread_throughput_qos_policy;
 // libc-test does not like mutable pointers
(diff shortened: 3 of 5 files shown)


src/unix/bsd/apple/mod.rs
```diff
@@ -19,6 +19,35 @@ pub type fsfilcnt_t = ::c_uint;
...
+//pub type thread_extended_policy_data_t = thread_extended_policy;
+//pub type thread_time_constraint_policy_data_t = thread_time_constraint_policy;
+//pub type thread_precedence_policy_data_t = thread_precedence_policy;
+//pub type thread_affinity_policy_data_t = thread_affinity_policy;
+//pub type thread_background_policy_data_t = thread_background_policy;
+//pub type thread_latency_qos_policy_data_t = thread_latency_qos_policy;
+//pub type thread_throughput_qos_policy_data_t = thread_throughput_qos_policy;
+// libc-test does not like mutable pointers
+//pub type thread_standard_policy_t = * mut thread_standard_policy_data_t;
+//pub type thread_extended_policy_t = * mut thread_extended_policy_data_t;
```
alexcrichton: What's up with these comments? Could you gist the errors you were seeing from libc-test?

alexcrichton:
Travis errors may be legit

raphaelcohn:
Yep. I'm hors de combat until Tue. Will try to gist then.

raphaelcohn:
I'm going to wait a bit until my other pull requests are green.

raphaelcohn:
Gist for struct crash during test: <https://gist.github.com/raphaelcohn/9684ede35f7f0691dabfc41815fd71f2>
Gist for mutable pointer problem: <https://gist.github.com/raphaelcohn/93aa68ad046fc7507800cddaf9f0ae62> (Failing line is `pub type thread_policy_t = *mut integer_t;`).

raphaelcohn:
Build failures may be due to the version of Mac OS X being used. These APIs were enhanced in 10.10 (<https://developer.apple.com/library/mac/documentation/General/Reference/APIDiffsMacOSX10_10SeedDiff/frameworks/Kernel.html>). Is there a way to use `#cfg` to target versions of `darwin` (eg `x86_64-apple-darwin14.5.0`)?

alexcrichton:
Oh you may just need to skip the signededness check (an option in ctests) as part of the libc-test script. You can add some special cases to skip that check on OSX for example.

Unfortunately we don't have much control over the OSX version here as it's just whatever Travis gives us. What APIs are these specifically?

raphaelcohn:
The APIs are Mac OS X Mach threading policy APIs, mostly suitable when using Mac OS X more as a server-like or number crunching OS.

It seems Travis [defaults to OS X 10.9](https://docs.travis-ci.com/user/osx-ci-environment/#OS-X-Version) but that we can choose to use 10.10. I suppose it's a case of do we want to? I know for myself that I do continue to run a OS X 10.9 laptop simply because 10.10 didn't work with a third party SSD...

There is a slightly larger problem here with both this change and the musl change. Effectively, these 'libc' libraries are changing far more rapidly than is traditional - in the Mac OS X case, annually, and in musl's case, even more frequently. This implies a need for versioning. With Mac OS X, in theory, if we could get to the env version (we can't currently), we could accommodate this; it would be legitimate to do so. With libraries like musl, we're pretty stuck, and particularly so if we're linking to system wide dynamic libs. A similar problem is likely with the Android bionic libs (some functions are only available in very recent builds).

One possible horrible workaround might be to do something with a `build.rs` that checks for function presence or OS version and then generates a bit of rust code to be used with the `insert!` macro. A sort of even more revolting autoconf. An alternative is to be explicit on the versions of each libc that we are supporting with each rust / libc combination release, and then stick to that.

For these particular functions, I can move to a 'private' implementation for now, but that simply pushes the problem down the road to when I publish a crate that needs them... In other words, I'm not sure what's right.

alexcrichton:
Oh bumping the OSX version should actually be fine, I had no idea we could do that! In general this crate strives to provide the most recent API for all platforms, and then users of the crate can opt out of the more recent functionality if more compatibility is desired. That is, it's up to you to maintain compatibility with older versions, we just give you all the goodies we can find.

Now that's all built on the assumption that the libraries themselves are always backwards compatible, which is true most of the time but not all...

.travis.yml
```diff
@@ -21,6 +21,7 @@ script:
 os:
   - linux
   - osx
+osx_image: xcode6.4
```
alexcrichton: Oh this actually ended up being upgraded as part of https://github.com/rust-lang/libc/pull/307 as well, wanna back this out for now as that may land ahead of this?

libc-test/build.rs
```diff
@@ -285,6 +288,9 @@ fn main() {
             // uuid_t is a struct, not an integer.
             "uuid_t" if dragonfly => true,
             n if n.starts_with("pthread") => true,
+			
+			// Mac OS X
+			n if n.starts_with("thread_") => true,
```
alexcrichton: tabs may be a bit off here?

src/unix/bsd/apple/mod.rs
```diff
@@ -20,6 +20,36 @@ pub type speed_t = ::c_ulong;
...
+pub type thread_policy_t = *mut integer_t;
+pub type thread_standard_policy_t = * mut thread_standard_policy_data_t;
+pub type thread_extended_policy_t = * mut thread_extended_policy_data_t;
+pub type thread_time_constraint_policy_t =
+* mut thread_time_constraint_policy_data_t;
+pub type thread_precedence_policy_t = * mut thread_precedence_policy_data_t;
+pub type thread_affinity_policy_t = * mut thread_affinity_policy_data_t;
+pub type thread_background_policy_t = * mut thread_background_policy_data_t;
+pub type thread_latency_qos_policy_t =
+* mut thread_latency_qos_policy_data_t;
```
alexcrichton: Perhaps this could be indented one to the right? We also tend to not put a space between "*" and "mut"

alexcrichton:
Looks good to me! Feel free to ping me whenever a PR is ready to go as unfortunately github doesn't send out notifications when the branch is updated :(

alexcrichton:
Closing due to inactivity, but feel free to resubmit with a rebase!

PR #297, opened 2016-05-21
Add pthread mutex type constants

Amanieu:
These are needed to create a pthread mutex that doesn't use lock elision on glibc.


alexcrichton:
@bors: r+ b2b9f29eecaf6f8d5650a676d9669f3f871595df

Thanks!

PR #299, opened 2016-05-25
Adding sched_getcpu() for Android and Linux

raphaelcohn:


raphaelcohn:
The CI fails for Linux-Musl because `sched_getcpu` was added after the release of musl used during CI (added by one of the Alpine Linux maintainers on the 5th March). It seems CI is using 0.9.5, which is antiquated. To make this work, we'd need to use a more recent musl release - and not the Debian package. Debian isn't a primary musl platform. Could we run Alpine Linux for CI? This is the primary Linux distro using musl as its baseline libc, and the most popular musl-based distro for Docker.

alexcrichton:
Unfortunately we probably can't add it just yet as the compiler doesn't run inside alpine, but we could modify the tests to just pull down the most recent version of musl

raphaelcohn:
Right-oh. I've not got any great experience of playing with Travis builds - is this likely to be a case of a little shell script munging? If so, if you can give me a pointer to where to start, I'll have a go.

raphaelcohn:
I need to find time to work on the travis build; this latest change is just to make this mergeable again.

alexcrichton:
Looks like it's just MUSL now, which can perhaps be fixed by installing a newer version of it?

raphaelcohn:
There hasn't been an official musl release with this code in it since February, although code subsequent to this has been cherrypicked by some distros (eg Alpine). In effect, that means there are 'different' versions of musl floating around that are current. Rather than cause gratuitous breakage, I'm going to change the PR to remove musl.

If and when this code is more widely deployed then I may revisit it. For my use case for now I can probably get equivalent functionality via the libnuma-sys crate.

alexcrichton:
@bors: r+ ae664ff911cff82bf19240a69a2fbcd1a6d83ae7

Sounds good to me!

PR #300, opened 2016-05-27
added const SOCK_SEQPACKET,  and SOCK_RDM where applicable

rrichardson:
... which is used for UNIX Seqpacket sockets as well as SCTP IP.

I have verified in source on  a couple Linux flavours as well as freebsd and darwin source that SOCK_SEQPACKET has value 5. 
Solaris is the odd one with a value of 6.  I also added RDM for solaris (value 5)


rrichardson:
Looks like networking failed in the openbsd test.

semarie:
it is due to maintenance on ftp5 : https://marc.info/?l=openbsd-misc&m=146427439004957

PR #289, opened 2016-05-15
Fix ioctl constants for musl target envs.

fnichol:
Heya!

I ran across this issue today while trying to build a portable static binary using the `x86_64-unknown-linux-musl` target. Took a bit of digging to make sure I understood what was going on, and while I may still be off the mark, I believe this is a fix to my issue.

Thanks!!

----

According to musl's source, the `ioctl` [function signature][musl-ioctl-h] takes an `int` as the request argument (i.e. an `i32`) which is reflected in this crate's [ioctl binding][musl-ioctl-rs]. It looks like when the ioctl constants were added that [glibc's default][glibc-ioctl-h] of a `c_ulong` type was used for the musl values as well, rather than a `c_int` type. This change updates these constants to a `c_int` so that they match the expected function call type.

Here is a minimal reproduction of the issue. Given this Rust program:

```rust
extern crate libc;

use libc::{ioctl, winsize, STDOUT_FILENO, TIOCGWINSZ};

fn main() {
    let mut wsize = winsize {
        ws_row: 0,
        ws_col: 0,
        ws_xpixel: 0,
        ws_ypixel: 0,
    };
    unsafe {
        ioctl(STDOUT_FILENO, TIOCGWINSZ, &mut wsize);
    }
    println!("Sizes: {{ rows: {}, cols: {}, xpixel: {}, ypixel: {} }}",
             wsize.ws_row,
             wsize.ws_col,
             wsize.ws_xpixel,
             wsize.ws_ypixel);
}
```

When run against the `x86_64-unknwon-linux-gnu` and
`x86_64-unknown-linux-musl` targets, we see the difference in behavior:

```
> cargo clean

> cargo run --target=x86_64-unknown-linux-gnu
   Compiling libc v0.2.11
   Compiling libc-musl-ioctl v0.1.0 (file:///src/libc-musl-ioctl)
     Running `target/x86_64-unknown-linux-gnu/debug/libc-musl-ioctl`
Sizes: { rows: 28, cols: 211, xpixel: 0, ypixel: 0 }

> cargo clean

> cargo run --target=x86_64-unknown-linux-musl
   Compiling libc v0.2.11
   Compiling libc-musl-ioctl v0.1.0 (file:///src/libc-musl-ioctl)
src/main.rs:13:30: 13:40 error: mismatched types:
 expected `i32`,
    found `u64` [E0308]
src/main.rs:13         ioctl(STDOUT_FILENO, TIOCGWINSZ, &mut wsize);
[...]

PR #301, opened 2016-05-30
Fix ioctl types for non-x86 musl

Amanieu:
This fixes up #289 by changing the type for other platforms as well.


alexcrichton:
@bors: r+ 

Thanks!

PR #308, opened 2016-06-01
change FD_ISSET to use *const raw pointer.

cpjreynolds:
Considering that `FD_ISSET` does not modify `fd_set`, it only needs a `*const` pointer to it.

### DIFF
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -305,7 +305,7 @@ f! {
         return
     }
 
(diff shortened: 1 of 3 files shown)


alexcrichton:
This unfortunately has the potential to be a breaking change, and according to [the spec](http://pubs.opengroup.org/onlinepubs/9699919799/) it looks like this "function" signature uses `*mut` instead of `*const`?

cpjreynolds:
Interesting, I hadn't considered the spec when making the changes. Unfortunate that the spec doesn't specify a `const` pointer. Thanks anyway! I'll close this now.

PR #309, opened 2016-06-04
Add semaphore APIs

sfackler:


sfackler:
Builds pass

src/unix/notbsd/android/mod.rs
```diff
@@ -95,6 +95,10 @@ s! {
         #[cfg(target_pointer_width = "32")]
         __bits: [__CPU_BITTYPE; 1],
     }
+
+    pub struct sem_t {
+        pub count: ::c_uint,
```
alexcrichton: Perhaps this could be opaque like the other platforms? That'd mean the special case in build.rs could be removed as well

src/unix/notbsd/linux/mips.rs
```diff
@@ -189,6 +189,15 @@ s! {
         pub mem_unit: ::c_uint,
         pub _f: [::c_char; 8],
     }
+
+    // FIXME this is actually a union
```
alexcrichton: yeah... binding C is so great

src/unix/notbsd/android/mod.rs
alexcrichton: (also applicable with musl below)

alexcrichton:
Nice! All looks good to me, just a minor nit about making the fields private (to avoid some extra libc-test logic), but other than that r=me

sfackler:
Updated

PR #310, opened 2016-06-05
Add more constants from sys/wait.h

nrc:
And fix WIFEXITED on Linux, which seemed to be wrong.

### DIFF
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -324,16 +324,16 @@ f! {
         }
     }
 
+    pub fn WTERMSIG(status: ::c_int) -> ::c_int {
+        status & 0o177
+    }
+
     pub fn WIFEXITED(status: ::c_int) -> bool {
-        (status & 0x7f) == 0
+        (status & 0o177) == 0
     }
 
     pub fn WEXITSTATUS(status: ::c_int) -> ::c_int {
         status >> 8
-    }
-
-    pub fn WTERMSIG(status: ::c_int) -> ::c_int {
-        status & 0o177
     }
 
     pub fn WCOREDUMP(status: ::c_int) -> bool {
--- a/src/unix/notbsd/mod.rs
+++ b/src/unix/notbsd/mod.rs
@@ -682,16 +682,32 @@ f! {
         }
     }
 
+    pub fn WIFSTOPPED(status: ::c_int) -> bool {
+        (status & 0xff) == 0x7f
+    }
+
+    pub fn WSTOPSIG(status: ::c_int) -> ::c_int {
+        (status >> 8) & 0xff
+    }
+
+    pub fn WIFSIGNALED(status: ::c_int) -> bool {
+        (status & 0x7f) + 1 >= 2
+    }
+
+    pub fn WTERMSIG(status: ::c_int) -> ::c_int {
+        status & 0x7f
+    }
+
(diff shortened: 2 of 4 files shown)


src/unix/bsd/mod.rs
```diff
@@ -324,18 +324,30 @@ f! {
         }
     }
 
-    pub fn WIFEXITED(status: ::c_int) -> bool {
-        (status & 0x7f) == 0
+    pub fn WIFSTOPPED(status: ::c_int) -> bool {
+        (status & 0o177) == 0o177
```
alexcrichton: Locally on OSX for me it looks like this is

```c
(_WSTATUS(x) == _WSTOPPED && WSTOPSIG(x) != 0x13)
```

so maybe this should also check `WSTOPSIG`?

src/unix/notbsd/mod.rs
```diff
@@ -682,17 +682,33 @@ f! {
         }
     }
 
-    pub fn WIFEXITED(status: ::c_int) -> bool {
-        (status & 0xff) == 0
+    pub fn WIFSTOPPED(status: ::c_int) -> bool {
+        (status & 0x7f) == 0x7f
```
alexcrichton: For me locally this is:

```
#define __WIFSTOPPED(status)    (((status) & 0xff) == 0x7f)
```

Maybe an extra bit on the mask of `0x7f`?

src/unix/notbsd/mod.rs
```diff
@@ -682,17 +682,33 @@ f! {
...
     }
 
-    pub fn WEXITSTATUS(status: ::c_int) -> ::c_int {
+    pub fn WSTOPSIG(status: ::c_int) -> ::c_int {
         (status >> 8) & 0xff
     }
 
+    pub fn WIFSIGNALED(status: ::c_int) -> bool {
+        (status & 0x7f) + 1 >= 2
```
alexcrichton: My local definition of this is crazy:

```c
#define __WIFSIGNALED(status) \                     
  (((signed char) (((status) & 0x7f) + 1) >> 1) > 0)
```

I'll... just assume this is the same!

src/unix/notbsd/mod.rs
```diff
@@ -682,17 +682,33 @@ f! {
...
+    }
+
     pub fn WTERMSIG(status: ::c_int) -> ::c_int {
         status & 0x7f
     }
+
+    pub fn WIFEXITED(status: ::c_int) -> bool {
+        (status & 0x7f) == 0
```
alexcrichton: Good catch on this one!

src/unix/bsd/mod.rs
nrc: But this is `WIFSTOPPED`, not `WSTATUS`

nrc: For the bsd headers I see `#define WIFSTOPPED(x)	(_WSTATUS(x) == _WSTOPPED)`
Where `#define	_WSTOPPED	0177` and `#define	_WSTATUS(x)	(_W_INT(x) & 0177)`.

src/unix/notbsd/mod.rs
nrc: fixed

src/unix/notbsd/mod.rs
nrc: I'm pretty sure they are

PR #317, opened 2016-06-13
Add `if_nameindex()` and `if_freenameindex()` from POSIX-2001

cjschneider2:
As the title says, this is adding the functions `if_nameindex` and `if_freenameindex`.

~~The only thing I'm personally unsure of is that the naming of the struct and the function for `if_nameindex` are the same and if this may cause problems with naming in Rust.~~ (This seems to work fine in my test program)


cjschneider2:
About the failing android build... It seems like these have been added in the android [ABI version 24](https://android.googlesource.com/platform/bionic/+/master/libc/include/net/if.h).

@alexcrichton: Putting the fn's in the `#[cfg(not(target_os = "android"))]` segment should disable them correct?

alexcrichton:
Ah yeah that should be fine, curses Android!

alexcrichton:
Also in https://github.com/rust-lang/libc/pull/316 it should be somewhat easier to update the Android API level we test against, although perhaps still not trivial.

cjschneider2:
@alexcrichton Sorry to bother you but I am somewhat stuck. If I do not compile the functions for android it complains that the struct can't be found in libc (which makes sense; it's not there). If I also
 ```Rust
#[cfg(not(target_os = "android"))]
pub struct if_nameindex { ... }
```
the s! macro blows up saying the struct is undefined as it's trying to decorate a struct that technically doesn't exit. What would the best course of action to fix this?

alexcrichton:
Ah yeah that may not work well, feel free to just copy it around to the relevant platforms though, this could just manually exclude Android for now

PR #318, opened 2016-06-15
Add missing constants for wait family

### LINKED ISSUE #315: Include WUNTRACED in `sys/wait.h` constants
For the `wait` family the only constant flag that is defined is WNOHANG.  The POSIX standard defines WUNTRACED as well. I would be happy to create a PR for this except I don't know how to find the value of the constant on Solaris. Linux also defines some more flags like WCONTINUED that could be added as well.

nikklassen:
Fixes #315


alexcrichton:
@bors: r+ 

Thanks!

PR #319, opened 2016-06-20
Add SEEK_DATA/SEEK_HOLE constants

posborne:
Relates to https://github.com/nix-rust/nix/pull/377.  These constants
are only available in systems running the Linux kernel.

### DIFF
--- a/src/unix/notbsd/linux/other/mod.rs
+++ b/src/unix/notbsd/linux/other/mod.rs
@@ -429,6 +429,9 @@ pub const F_SETOWN: ::c_int = 8;
 pub const F_SETLK: ::c_int = 6;
 pub const F_SETLKW: ::c_int = 7;
 
+pub const SEEK_DATA: ::c_int = 3;
+pub const SEEK_HOLE: ::c_int = 4;
+
 pub const SFD_NONBLOCK: ::c_int = 0x0800;
 
 pub const TCSANOW: ::c_int = 0;


alexcrichton:
Thanks! Looks like these constants may not exist yet on musl though?

posborne:
> Thanks! Looks like these constants may not exist yet on musl though?

Yes, that would appear to be the case.  Wasn't sure, so I figured I would let CI let me know.  I'll add an ignore in the tests for that constant for musl (I believe that is the best way to handle).

alexcrichton:
Ah the preferred method is actually to duplicate these constants down to the level where they're defined (e.g. android + non-musl linux) rather than `#[cfg(not(..))]` but other than that sounds good to me.

posborne:
> Ah the preferred method is actually to duplicate these constants down to the level where they're defined (e.g. android + non-musl linux) rather than #[cfg(not(..))] but other than that sounds good to me.

I can do this, although it doesn't seem strictly necessary (or beneficial) and makes usage of the constants in downstream crates more complicated as instead of worrying about the OS we need to be concerned with the libc implementation (which is fine in some cases, but unnecessary here).  These constants are defined by the kernel and musl lets the kernel accept/reject the values for calls where they are used (e.g. [lseek](https://git.musl-libc.org/cgit/musl/tree/src/unistd/lseek.c#n5)).

alexcrichton:
Ah yeah we just typically try to stick to the musl headers, regardless of whether they just forward it to the kernel or not

posborne:
Sure, I'll take a look at fixing musl as well then.

posborne:
Ok, Travis is happy after amending the commit to live under `unix/notbsd/linux/other`

PR #320, opened 2016-06-22
unix: Add lchown

knight42:


knight42:
@alexcrichton 
Travis-ci complaint: [bad lchown function pointer: rust: 2528398520 (0x96b44cb8) != c 2528415600 (0x96b48f70)](https://travis-ci.org/rust-lang/libc/jobs/139429859#L144) only on [i686-apple-darwin](https://travis-ci.org/rust-lang/libc/builds/139429853).

Since I knew little about darwin kernel, I wonder if you could tell me how to fix this error?

alexcrichton:
Ah that typically just means there's some weird symbol mangling going on, a fix similar to that of `close` beneath this function should suffice!

knight42:
@alexcrichton 
Thanks! It is fixed now.

Issue #327, opened 2016-07-09
struct *linger* is missing


sateffen:
Hey,

the `linger {int l_onoff, int l_linger}` struct is missing in this binding. I checked it's implementation in [glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/bits/socket.h;h=2eb95f7e20b8b68685ac2363b4b2db72cab8d98a;hb=HEAD#l402) and [musl](https://git.musl-libc.org/cgit/musl/tree/include/sys/socket.h#n40), but it's missing in this binding.

I guess the struct should look like in the crate [winapi](https://github.com/retep998/winapi-rs/blob/e13b8561738d33f52cbde4fd49ce8dd73025a610/src/winsock2.rs#L113).

alexcrichton:
Ah yeah should be good to add at any time!

sateffen:
Ok, it's a bad idea to take the struct from winapi, because on windows you need to set u_short's, on *nix systems you need to set int's, so we need to set it so c_int, not u16.

alexcrichton:
Now fixed!

PR #329, opened 2016-07-10
Added linger struct

sateffen:
Added linger struct like mentioned in #327 

I looked up documentations on some platforms, and all look the same in the unix world. Additionally I've tested this with a local setup, building it against [this](https://github.com/rust-lang-nursery/net2-rs/pull/44) pullrequest, and it works well.

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -287,11 +287,6 @@ s! {
         pub int_p_sign_posn: ::c_char,
         pub int_n_sign_posn: ::c_char,
     }
-
-    pub struct linger {
-        pub l_onoff: ::c_int,
-        pub l_linger: ::c_int,
-    }
 }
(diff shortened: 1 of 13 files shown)


alexcrichton:
Looks like this is defined the same everywhere, perhaps the definition could go in `src/unix/mod.rs`?

sateffen:
Fugg, I could have foreseen this^^

I looked for each SO_LINGER to have a linger struct with it, but I failed to recognize this is on each platform. Thanks for the hint! I'll change this as soon as I've got some time.

Thanks @alexcrichton for your patience and guidance through my rust learning experience, even though this are embarrassing problems that don't have anything to do with rust^^

sateffen:
Done the refactoring. Works with my test setup (on a cloud9 machine)

alexcrichton:
Awesome, and no worries!

PR #332, opened 2016-07-12
solaris: add dlsym/dlopen RTLD constants

cuviper:
These constants are taken from OpenIndiana's `dlfcn.h`.  In particular,
`RTLD_DEFAULT` is now needed in `fetch` for std's weak symbol support.


alexcrichton:
Thanks!

PR #333, opened 2016-07-12
Linux: add reboot() and constants

bugaevc:
Add declarations from `sys/reboot.h` and `linux/reboot.h`. Initially originated as nix-rust/nix#386.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -165,9 +165,11 @@ fn main() {
         cfg.header("sys/vfs.h");
         cfg.header("sys/syscall.h");
         cfg.header("sys/sysinfo.h");
+        cfg.header("sys/reboot.h");
         if !musl {
             cfg.header("linux/netlink.h");
             cfg.header("linux/magic.h");
+            cfg.header("linux/reboot.h");
 
             if !mips {
                 cfg.header("linux/quota.h");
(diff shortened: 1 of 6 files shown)


bugaevc:
Should the rest of reboot declarations be in `notbsd/mod.rs`, or should I copy the declarations to `notbsd/android`, or are they somehow included from `notbsd/linux/mod.rs` (doesn't seem to be the case, but there are lots of functions there and just several in `notbsd/android`)?

alexcrichton:
If they're shared among all the platforms under `notbsd` (Android + Linux) then it should be fine to place the definitions in `notbsd/mod.rs`, but copying is also fine. Looks like CI is green here so I'd be good merging, but do you wanna juggle around the declarations a bit?

bugaevc:
I'm not really sure whether the same stuff is available under Android: Android is Linux anyway, but it may well be glibc-specific stuff.
Let me try to put them into `notbsd` and see if that breaks the Android build.

bugaevc:
OK, so it does break the Android build. Reverting.

bugaevc:
Feel free to merge this (or should we play the rebasing/squashing game?).

alexcrichton:
Nah this is fine, thanks again!

bugaevc:
Thanks! When can I expect this to be released (e.g. to use it in nix)?

alexcrichton:
Would it be possible for nix to use a git version of libc for now? I can be sure to make a release right before a nix release, but we just made a release yesterday I believe for another nix-requested PR :)

alexcrichton:
That being said, I certainly don't mind making another release!

bugaevc:
Filed nix-rust/nix#387 for nix to track libc's master

alexcrichton:
Thanks! I'll add some comments over there as well.

PR #334, opened 2016-07-15
Cast unsigned literals to signed integers.

fiveop:
This makes it explicity that we know the literals are to large and
will result in a negative number. This gets rid of a warning in
1.9.0 and younger versions, as well as an error when using the
constants as discrimintants of #[repr(i32)] enumerations in older
versions.

For nix-rust/nix#386


alexcrichton:
Thanks!

Issue #338, opened 2016-07-25
Add struct utmp


knight42:
Since utmp structure and its related functions are missing, I want to add them to libc. However, the definition of struct utmp is a bit tricky:

```C
           struct utmp {
           ...
           ...

               /* The ut_session and ut_tv fields must be the same size when
                  compiled 32- and 64-bit.  This allows data files and shared
                  memory to be shared between 32- and 64-bit applications. */
           #if __WORDSIZE == 64 && defined __WORDSIZE_COMPAT32
               int32_t ut_session;           /* Session ID (getsid(2)),
                                                used for windowing */
               struct {
                   int32_t tv_sec;           /* Seconds */
                   int32_t tv_usec;          /* Microseconds */
               } ut_tv;                      /* Time entry was made */
           #else
                long   ut_session;           /* Session ID */
                struct timeval ut_tv;        /* Time entry was made */
           #endif

           ...
           };
```

I wonder is it safe to assume all 64-bit platforms are 32-bit compatible? If not, how should I handle this case?

alexcrichton:
Ah the CI for this repo is typically pretty good about detecting these problems. I'd just submit a PR and see what travis says :)

knight42:
Nice! Thanks for your work.

PR #340, opened 2016-07-27
Add utmpx

knight42:
#338

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -105,6 +105,7 @@ fn main() {
         cfg.header("arpa/inet.h");
         cfg.header("time64.h");
         cfg.header("xlocale.h");
+        cfg.header("utmp.h");
     } else if !windows {
         cfg.header("glob.h");
         cfg.header("ifaddrs.h");
@@ -121,6 +122,12 @@ fn main() {
             if !netbsd && !openbsd {
                 cfg.header("execinfo.h");
                 cfg.header("xlocale.h");
+            }
+
+            if openbsd {
+                cfg.header("utmp.h");
+            } else {
+                cfg.header("utmpx.h");
             }
         }
     }
--- a/src/unix/bsd/openbsdlike/openbsd.rs
+++ b/src/unix/bsd/openbsdlike/openbsd.rs
@@ -137,7 +137,24 @@ s! {
         pub int_p_sign_posn: ::c_char,
         pub int_n_sign_posn: ::c_char,
     }
+
+    pub struct lastlog {
+        ll_time: ::time_t,
+        ll_line: [::c_char; UT_LINESIZE],
+        ll_host: [::c_char; UT_HOSTSIZE],
+    }
+
+    pub struct utmp {
+        pub ut_line: [::c_char; UT_LINESIZE],
+        pub ut_name: [::c_char; UT_NAMESIZE],
+        pub ut_host: [::c_char; UT_HOSTSIZE],
+        pub ut_time: ::time_t,
+    }
 }
+
+pub const UT_NAMESIZE: usize = 32;
+pub const UT_LINESIZE: usize = 8;
+pub const UT_HOSTSIZE: usize = 256;
 
 pub const O_CLOEXEC: ::c_int = 0x10000;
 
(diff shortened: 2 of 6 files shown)


src/unix/notbsd/linux/mod.rs
```diff
@@ -481,6 +506,23 @@ pub const RB_POWER_OFF: ::c_int = 0x4321fedcu32 as i32;
...
+pub const BOOT_TIME: ::c_short = 2;
+pub const NEW_TIME: ::c_short = 3;
+pub const OLD_TIME: ::c_short = 4;
+pub const INIT_PROCESS: ::c_short = 5;
+pub const LOGIN_PROCESS: ::c_short = 6;
+pub const USER_PROCESS: ::c_short = 7;
+pub const DEAD_PROCESS: ::c_short = 8;
+pub const ACCOUNTING: ::c_short = 9;
+pub const _PATH_UTMP: &'static str = "/var/run/utmp";
+pub const _PATH_WTMP: &'static str = "/var/log/wtmp";
```
alexcrichton: Currently we don't include constant strings like this in libc, perhaps these could be omitted for now?

src/unix/notbsd/linux/mod.rs
```diff
@@ -192,6 +192,31 @@ s! {
...
+        pub ut_type: ::c_short,
+        pub ut_pid: ::pid_t,
+        pub ut_line: [::c_char; UT_LINESIZE],
+        pub ut_id: [::c_char; UT_IDSIZE],
+
+        pub ut_user: [::c_char; UT_NAMESIZE],
+        pub ut_host: [::c_char; UT_HOSTSIZE],
+        pub ut_exit: exit_status,
+        pub ut_session: ::int32_t,
+        pub ut_tv: _timeval,
```
alexcrichton: Out of curiosity, is this equivalent to `timeval`?

knight42: Not always.
```
pub struct timeval {
    pub tv_sec: time_t (AKA i64 on 64-bit platform),
    pub tv_usec: suseconds_t (AKA i64 on 64-bit platform),
}
```

However, according the definition of `struct utmp` on linux:
```C
struct utmp {
    ...
    ...

    /* The ut_session and ut_tv fields must be the same size when
       compiled 32- and 64-bit.  This allows data files and shared
       memory to be shared between 32- and 64-bit applications. */
    #if __WORDSIZE == 64 && defined __WORDSIZE_COMPAT32
    int32_t ut_session;           /* Session ID (getsid(2)),
                                     used for windowing */
    struct {
        int32_t tv_sec;           /* Seconds */
        int32_t tv_usec;          /* Microseconds */
    } ut_tv;                      /* Time entry was made */
    #else
    long   ut_session;           /* Session ID */
    struct timeval ut_tv;        /* Time entry was made */
    #endif

    ...
};
```
and it seems that `__WORDSIZE_COMPAT32` is usually defined on 64-bit platform, so the actual definition of `struct utmp` goes like this:
```C
struct utmp {
    ...
    ...

    int32_t ut_session;           /* Session ID (getsid(2)),
                                     used for windowing */
    struct {
        int32_t tv_sec;           /* Seconds */
        int32_t tv_usec;          /* Microseconds */
    } ut_tv;                      /* Time entry was made */

    ...
};
```

It's worth noting that getutent() reads `struct utmp`s from binary file /var/run/utmp, hence the actual size of `struct utmp` matters.

src/unix/notbsd/linux/mod.rs
knight42: Removed.

src/unix/notbsd/linux/mod.rs
alexcrichton: Ah oh well, thanks for the clarification though!

alexcrichton:
Looks like the CI is failing for non-spurious reasons?

knight42:
Yep, according to the stderr, cc could not find the definition of the structs I added. Trying to find out why.

knight42:
@alexcrichton Got it. `libc-test` doesn't include `utmp.h`.

knight42:
@alexcrichton Some irrelated errors happened this time... 

Anyway, I decided to compromise, only add `struct utmp(x)` on certain platforms.

alexcrichton:
@bors: r+ b47e554216fadb9613f4ff8ec271ecd9a5bcbc7f

Thanks!

alexcrichton:
grr bors, but tests passed!

PR #341, opened 2016-07-27
arm-musl targets now use cfg(target_env = "musl")

japaric:
As decided on rust-lang/rust#33244, the upcoming ARM musl targets will use musl (instead of musleabi/musleabihf) as their target_env.

r? @alexcrichton 
cc @timonvo


alexcrichton:
@bors: r+

Thanks!

alexcrichton:
(homu's been slow but CI otherwise passed)

PR #342, opened 2016-07-31
Add functions to get password/group file entry

knight42:
* Add `struct group`
* Add `getgrnam()`,`getgrgid()`,`getpwuid()`, `getpwnam()`,`setpwent()`,`getpwent()`,`endpwent()`,

### DIFF
--- a/src/unix/solaris/mod.rs
+++ b/src/unix/solaris/mod.rs
@@ -966,10 +966,16 @@ extern {
                        serv: *mut ::c_char,
                        sevlen: ::socklen_t,
                        flags: ::c_int) -> ::c_int;
+    pub fn getpwnam_r(name: *const ::c_char,
+                      pwd: *mut passwd,
+                      buf: *mut ::c_char,
+                      buflen: ::c_int) -> *const passwd;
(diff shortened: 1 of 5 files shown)


src/unix/solaris/mod.rs
```diff
@@ -966,10 +966,8 @@ extern {
                        serv: *mut ::c_char,
                        sevlen: ::socklen_t,
                        flags: ::c_int) -> ::c_int;
-    pub fn getpwuid_r(uid: ::uid_t,
-                      pwd: *mut passwd,
-                      buf: *mut ::c_char,
-                      buflen: ::size_t) -> *const passwd;
```
alexcrichton: It looks like the signature of getpwuid_r above is different than this one for solaris, was that intentional?

Unfortunately we don't have CI for the solaris bindings yet, so I'm not sure which one is right :(

knight42: Thanks for the catch! I have updated my patch according to the [man pages](http://docs.oracle.com/cd/E36784_01/html/E36874/getpwuid-r-3c.html).

alexcrichton:
Thanks!

PR #343, opened 2016-07-31
Added Ptrace option parameters

valarauca:
When the tracer process (often parent, but not necessarily) starts tracing it needs to pass the kernel options about what it is trying to accomplish. These constants were not defined in this document, so I'm attempting to add them.


alexcrichton:
Thanks! Looks like some of the CI failures may be legit though?

valarauca:
Copy paste issue I accidentally duplicated a line 3 times.

alexcrichton:
Thanks!

PR #344, opened 2016-08-01
Add clock_nanosleep for linux and solaris

letheed:
Add `clock_nanosleep`.

Support is a bit scarce but it's the only sleep call with options for monotonic clock and absolute time.


alexcrichton:
Thanks!

PR #352, opened 2016-08-07
Treat file descriptors as signed platform sized int's

valarauca:
Every Unix allows for less then <2^32 file descriptors so they likely all can be treated as i32's but I'm not sure about platform specific issues this may arise (I couldn't test this on every unix). 

I really don't see why they're all being handled like `*mut File`. This is just really inconvenient. Since you have to jump though `std::mem::transmute()` to interact with these functions.

I don't understand why the `File{}` type is being used other then to just discourage people from using this library.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -151,50 +151,50 @@ extern {
     #[cfg_attr(all(target_os = "macos", target_arch = "x86"),
                link_name = "fopen$UNIX2003")]
     pub fn fopen(filename: *const c_char,
-                 mode: *const c_char) -> *mut FILE;
+                 mode: *const c_char) -> isize;
     #[cfg_attr(all(target_os = "macos", target_arch = "x86"),
                link_name = "freopen$UNIX2003")]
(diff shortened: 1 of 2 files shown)


posborne:
If you want to use versions of these functions that don't take a file pointer, you should use the ones that don't take a file pointer (the ones you suggest changing simply do in C as well -- look at the headers for fopen for instance: http://man7.org/linux/man-pages/man3/fopen.3.html).  The alternate is to just use open (http://man7.org/linux/man-pages/man2/open.2.html).  The behavior is different but the signatures for functions here do need to match the libc we are linking against.

These are present in this crate: https://doc.rust-lang.org/libc/x86_64-unknown-linux-gnu/libc/fn.open.html (if not, that can be addressed individually).

For ease of use, you are better off using the `std` crate or the `nix` crate to get a higher-level interface.  The libc crate exists to make access possible but not to provide a nice, high-level, or safe interface to libc.

valarauca:
Very sorry I got the libc functions confused with the system calls of the same name.

PR #353, opened 2016-08-08
Make utmp(x) support more platforms

knight42:
Also ignored deprecated functions `readdir_r` and `readdir64_r` that fail the tests.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -283,6 +283,11 @@ fn main() {
         match ty {
             "sockaddr_nl" => musl,
 
+            // On Linux, the type of `ut_tv` field of `struct utmpx`
+            // can be an anonymous struct, so an extra struct,
+            // which is absent in glibc, has to be defined.
+            "__timeval" if linux => true,
+
             // The alignment of this is 4 on 64-bit OSX...
             "kevent" if apple && x86_64 => true,
 
@@ -431,7 +436,9 @@ fn main() {
         // This is a weird union, don't check the type.
         (struct_ == "ifaddrs" && field == "ifa_ifu") ||
         // sighandler_t type is super weird
-        (struct_ == "sigaction" && field == "sa_sigaction")
+        (struct_ == "sigaction" && field == "sa_sigaction") ||
+        // __timeval type is a patch which doesn't exist in glibc
+        (linux && struct_ == "utmpx" && field == "ut_tv")
     });
 
     cfg.skip_field(move |struct_, field| {
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -64,6 +64,8 @@ fn main() {
             cfg.header("ws2tcpip.h");
         }
     } else {
+        cfg.flag("-Wno-deprecated-declarations");
+
         cfg.header("ctype.h");
         cfg.header("dirent.h");
         if openbsd {
--- a/libc-test/build-generated.rs
+++ b/libc-test/build-generated.rs
@@ -10,6 +10,7 @@ fn main() {
         .flag("-Wall")
         .flag("-Wextra")
         .flag("-Werror")
+        .flag("-Wno-deprecated-declarations")
         .flag("-Wno-type-limits")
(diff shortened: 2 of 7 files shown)


src/unix/mod.rs
```diff
@@ -334,10 +334,6 @@ extern {
                link_name = "opendir$INODE64$UNIX2003")]
     #[cfg_attr(target_os = "netbsd", link_name = "__opendir30")]
     pub fn opendir(dirname: *const c_char) -> *mut ::DIR;
-    #[cfg_attr(target_os = "macos", link_name = "readdir_r$INODE64")]
-    #[cfg_attr(target_os = "netbsd", link_name = "__readdir_r30")]
-    pub fn readdir_r(dirp: *mut ::DIR, entry: *mut ::dirent,
-                     result: *mut *mut ::dirent) -> ::c_int;
```
alexcrichton: Despite these being deprecated, we can't remove these due to backwards compatibility, could these stick around for now?

libc-test/build.rs
```diff
@@ -281,6 +281,9 @@ fn main() {
         match ty {
             "sockaddr_nl" => musl,
 
+            // user-defined struct to patch struct utmpx on Linux
+            "__timeval" if linux => true,
```
alexcrichton: Could you elaborate a bit on this exception? How come we can't match what libc does?

knight42: The declaration of `struct utmpx` on Linux:
```
struct utmp {
...
...

    /* The ut_session and ut_tv fields must be the same size when
       compiled 32- and 64-bit.  This allows data files and shared
       memory to be shared between 32- and 64-bit applications. */
#if __WORDSIZE == 64 && defined __WORDSIZE_COMPAT32
    int32_t ut_session;           /* Session ID (getsid(2)),
                                     used for windowing */
    struct {
        int32_t tv_sec;           /* Seconds */
        int32_t tv_usec;          /* Microseconds */
    } ut_tv;                      /* Time entry was made */
#else
     long   ut_session;           /* Session ID */
     struct timeval ut_tv;        /* Time entry was made */
#endif

...
};
```
As shown above, the type of field `ut_tv` can be an anonymous struct, so I have to define a `struct __timeval` which is nonexistent in glibc to handle this case.

src/unix/mod.rs
knight42: I decided to ignore them in the tests.

libc-test/build.rs
```diff
@@ -350,6 +353,9 @@ fn main() {
     cfg.skip_fn(move |name| {
         // skip those that are manually verified
         match name {
+            // glibc-2.24 deprecates these functions
+            "readdir_r" | "readdir64_r" => true,
```
alexcrichton: can these still be tested and the deprecation warning allowed?

libc-test/build.rs
alexcrichton: Ok! Can that be added in a comment here?

libc-test/build.rs
knight42: The error message is pasted below.
```
out/all.c: In function ‘__test_fn_readdir_r’:
out/all.c:2061:17: error: ‘readdir_r’ is deprecated [-Werror=deprecated-declarations]
                 return readdir_r;
                 ^~~~~~
out/all.c:16:0:
/usr/include/dirent.h:183:12: note: declared here
 extern int readdir_r (DIR *__restrict __dirp,
            ^~~~~~~~~
```
Perhaps we can pass `-Wno-deprecated-declarations` to the compiler? Is that acceptable?

alexcrichton: Sounds good to me!

knight42:
@alexcrichton Why did [cargo chose a rather old commit of `ctest`](https://travis-ci.org/rust-lang/libc/jobs/150840469#L1023) in qemu-openbsd?

knight42:
@alexcrichton could be merge now

alexcrichton:
Awesome, thanks!

PR #358, opened 2016-08-16
add missing (common) iff constants

dovreshef:
as seen here: [glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/gnu/net/if.h), [musl](http://git.musl-libc.org/cgit/musl/tree/include/net/if.h).


alexcrichton:
Thanks!

PR #359, opened 2016-08-18
Add some macros from sched.h for android and linux.

fiveop:
I duplicated the code for android, because the member of cpu_set_t
is private, thus cannot be accessed in the parent mod. I would have
had to configure the name anyway based on the platform, because it
differs between the platforms.


alexcrichton:
Thanks!

posborne:
For the record, relates to https://github.com/nix-rust/nix/pull/402

PR #360, opened 2016-08-18
Add sem_timedwait for unix platforms

jynnantonix:
We can almost implement synchronization mechanisms in pure-rust using
just pthread semaphores as the primitive but sem_timedwait is necessary for
implementing timeout-based functions.


alexcrichton:
Looks like OSX may call this something else?

jynnantonix:
Looks like it isn't actually implemented on OS X and iOS :-/

It is available on the BSDs, though.  Is there some way to enable it for all unixes except OSX and iOS?

alexcrichton:
Oh yeah you can just push the definition of the function into each respective module (e.g. duplicate it everywhere but OSX) and that should be fine.

alexcrichton:
Thanks!

PR #366, opened 2016-08-27
add initial support for mips64

japaric:
just enough to be able to cross compile std for the mips64-gnu targets

tested by running libc-test for mips64-unknown-linux-gnuabi64

r? @alexcrichton


japaric:
> for mips64-unknown-linux-gnuabi64

which has not even landed in rust-lang/rust. We meta now.

alexcrichton:
@bors: r+

Nice! We should remember to add to the libc-test matrix as soon as we've got nightlies for mips64 up and running.

japaric:
Fixed the style errors. I don't know if I have bors rights here but ...

@bors r=alexcrichton

PR #369, opened 2016-08-28
Add CI for mipsel-unknown-linux-musl

alexcrichton:
Download a trunk OpenWRT toolchain and verify bindings against that. Also fixup
a few stray errors today

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,9 @@ matrix:
     - os: linux
       env: TARGET=powerpc64-unknown-linux-gnu
       rust: stable
+    - os: linux
+      env: TARGET=mipsel-unknown-linux-musl
+      rust: stable
 
     # beta
     - os: linux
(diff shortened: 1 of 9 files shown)


alexcrichton:
r? @japaric 

Is there perhaps a more stable location to download this toolchain from? It looks like that link may go stale as soon as it's updated...

japaric:
LGTM

> Is there perhaps a more stable location to download this toolchain from? It looks like that link may go stale as soon as it's updated...

Sadly, no. But you could use a command like the one below to get the current toolchain name.

```
curl -sL https://downloads.openwrt.org/snapshots/trunk/malta/generic | grep SDK | cut -d'"' -f2
```

alexcrichton:
@bors: r+ 

Ok, we'll see how long this lasts us and we can always tweak when the URL changes.

PR #372, opened 2016-08-30
added getchar and putchar

philippkeller:
Not sure why they were not in the repo until now, but they are both part of POSIX.1-2008: [getchar](http://pubs.opengroup.org/onlinepubs/9699919799/functions/getchar.html), [putchar](http://pubs.opengroup.org/onlinepubs/9699919799/functions/putchar.html)


alexcrichton:
@bors: r+

Thanks!

Issue #371, opened 2016-08-29
Reorganize the Linux module to reuse more code


GuillaumeLestringant:
This issue is a follow-up of the discussion started with #368 trying to be more general. The main idea is that the separation between different libc as a first branching inside the Linux module actually leads to much more code duplication as could be possible by branching by platform first, and maybe (we shall see if it is possible) have the libc and the platform be orthogonal.

In this issue, I wish to propose a working plan for this severe refactoring, to get approval before anybody, me included, starts working.

## Step 0: the MIPS module

The recently created MIPS and MIPS64 modules are actually MIPS-glibc and MIPS64-glibc. Maybe, to make that clearer, we should put them in their logical place, that is inside `other/b32` and `other/b64` module. Maybe it will induce too much changes to be worth the effort, so let this be optional.

## Step 1: get an appropriate test platform for MIPS(64)

@alexcrichton expressed concern on this matter, and it should be solved before any great amount of work is done. Question: is #369 enough or do we need more?

## Step 2: set up the working space

We can than create a `common` submodule, with its own submodules `b32`, `b64` and every necessary platform, to receive everything that is common to every one of the three libc. This submodule would be included alongside the remaining `musl` and `other` and `mips(64)` modules, to make sure nothing is lost while the work is in progress.

## Step3: put in common everything that can be put in common

To make this easier, a few recommendations.

### Type definitions

As much as possible, use the C types in the right part of the definition. For instance, `clock_t` is defined as `i32` in `mips` and `other/b32`, as `i64` in `other/b64`, and as `c_long` in `musl`. These are actually *all* `long`s. Using `::c_long` for all of them would make it possible to have a single definition for all Linuxes, that would actually be in `notbsd` given that `android` has the same definition.

### Structure definitions

It might not always be possible, but the apparent discrepancies in definitions between different libc are generally less than they seem. An example is given [here](https://github.com/rust-lang/libc/pull/368#issuecomment-242993752).

To make this work, it might be necessary to have some `cfg` elements to mimic the macros like `__USE_FILE_OFFSET64`. How exactly would that be possible?

### Macros and functions

Apart from a few exceptions, these should really be the same across the different libc and only differ in the platforms.

## Step 4: see if libc and platform are orthogonal

That is, make sure that all the definitions remaining in `other` and `musl` are platform independent. if this has been achieved, merge `common/mod.rs` into `linux/mod.rs`, and have a separate `if cfg` for platforms and for libc.

## Step 5: bonus

[...]

PR #373, opened 2016-08-30
Refactoring the Linux module: first try.

GuillaumeLestringant:
So this is a first try of the refactoring proposed in #371. It sets the tree structure up, and gives a go at the `fcntl.h` header. It is an interesting choice because it has type and structure definitions, generic and platform specific macros, and also POSIX and Linux specific definitions, as well as glibc and musl specific ones. So it covers more or less all possible cases.

For clarity’s sake, all definitions of this header have been placed at the end of their respective sections, along with a comment saying:

```rust
/* Header <fcntl.h> */
```

Doing so with all headers should make sure that a specific definition is easier to find than in the mess we have for now.

A few notes on unexpected issues.

- Generally speaking, glibc and uclibc are closer together than with musl. So, for definitions found in glibc but neither musl nor uclibc, I created a `gnu` submodule under `other` (and a `uclibc` one, that remains empty for now).
- The glibc has a structure named `file_handle` whose last element is a zero-sized array. This is rendered by a dynamically-sized array in Rust, but then, the struct cannot implement the `Copy` trait, and makes the `s!` macro bug. I commented that line out, but a better solution would be wished.
- I use a feature `file_offset64` that should be equivalent to `__USE_FILE_OFFSET64`. This feature must be defined in musl targets (!), but I don’t know how to make sure it is the case. As a palliative, I added the condition `target_env = "musl"` everywhere `file_offset64` is used, but this cannot be definitive. **Advice strongly wished here.**

PS: I can’t test every platform at home, so there might be some remaining bugs. Will add some commits when I know where they are.


petrochenkov:
I haven't looked carefully, but "reuse more code" and "+760 −316" diff look a bit contradicting, no? Is some new stuff added?

GuillaumeLestringant:
Yes, there is indeed some added stuff. And a part of it are comments.

GuillaumeLestringant:
So, I finally managed to put most of the bugs out. There are only three remaining, for which I have no clue. So, asking for help.

On `i686-unknown-linux-musl` (job 9), the test script doesn’t find the `flock` type, although it is undoubtedly defined in `common/b32/x86.rs`. Is there something weird going on with the `cfg_if` macro? Am I missing something?

On `aarch64-unknown-linux-gnu` (job 11), 5 macros seem not to be defined in the library against which the tests are done. These macros are defined [here](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/bits/fcntl-linux.h;h=7e5b0aecdcb45213186840667f80b738f5c428b8;hb=HEAD#l132) and [here](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/bits/fcntl-linux.h;h=7e5b0aecdcb45213186840667f80b738f5c428b8;hb=HEAD#l325), each time gated by a `__USE_GNU` macro, so I don’t get it why the AArch64 version doesn’t have them.

On `mipsel-unknown-linux-musl` (job 17), if I understand the error messages correctly, the script complains about `struct flock` having no `l_sysid` field, but the [musl definition](https://git.musl-libc.org/cgit/musl/tree/include/fcntl.h#n24) has no such field: only the [glibc definition](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/mips/bits/fcntl.h;h=6092f2a520c8c406f1f7d4442a581cac31f6dfeb;hb=HEAD#l67) has it (and uClibc too).

I don’t really understand how the tests are done so, maybe I’m missing something, or maybe the tests are wrong somehow, I don’t know. I would really use some insight here.

alexcrichton:
Thanks for the PR @GuillaumeLestringant! 

Before going much farther, though, I think we may want to get on the same page about the motivation of this refactoring. Could you help walk me through what you're thinking?

Today this crate optimizes for *readability* rather than writing new bindings themselves. That is, we greatly favor duplication over a jungle of `#[cfg]`. Additionally this crate favors simple-as-possible bindings, avoiding comments/organization/otherwise. These aspects mean that working within one particular platform is almost always a breeze as there's a "path to the root" in the module hierarchy where you just pick up more bindings as you go along. Additionally when binding a new platform, you frequently just start fresh in a new module and add in everything until it starts compiling.

Finally, the duplication is all validated through the CI infrastructure on this repository. That is, basically all definitions/constants/functions are automatically verified, so there's no risk of duplication silently introducing bugs.

So given all that, could you walk me through the high level goals and rationale of this refactor? Additionally, I'm curious where this'll all end up eventually!

PR #374, opened 2016-09-01
Add system call constants

antonblanchard:
System calls are not consistent across architectures. Even 32 and
64 bit versions of the same architecture are sometimes different.

Instead of individual applications defining system calls when they need
it, lets add them to libc.


antonblanchard:
Here is a first pass at adding system call numbers. Happy for any feedback as to how we should go about this.

alexcrichton:
Awesome, thanks! Could these be added to the existing architecture modules though? I think we should have a module for all the architectures here. It may involve some duplication, but that's fine to just copy/paste.

alexcrichton:
Looks like CI is also indicating that some C header needs to be included to verify these constants?

src/unix/notbsd/linux/syscall/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+    if #[cfg(target_arch = "arm")] {
+        mod arm;
+        pub use self::arm::*;
+    } else if #[cfg(target_arch = "aarch64")] {
+        mod aarch64;
+        pub use self::aarch64::*;
+    } else if #[cfg(target_arch = "powerpc")] {
+        mod powerpc;
+        pub use self::powerpc::*;
+    } else if #[cfg(any(target_arch = "powerpc64"))] {
```
parched: Is the `any` needed?

src/unix/notbsd/linux/syscall/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+        mod powerpc64;
+        pub use self::powerpc64::*;
+    } else if #[cfg(target_arch = "x86")] {
+        mod x86;
+        pub use self::x86::*;
+    } else if #[cfg(any(target_arch = "x86_64"))] {
```
parched: and here?

alexcrichton:
I'm going to close this for now, but we're alwyas willing to add these once tests are passing!

PR #377, opened 2016-09-03
add support for s390x

japaric:
tested by running libc-test on real s390x hardware.

r? @alexcrichton


src/unix/notbsd/mod.rs
```diff
@@ -613,7 +613,10 @@ pub const POSIX_FADV_NORMAL: ::c_int = 0;
 pub const POSIX_FADV_RANDOM: ::c_int = 1;
 pub const POSIX_FADV_SEQUENTIAL: ::c_int = 2;
 pub const POSIX_FADV_WILLNEED: ::c_int = 3;
+// TODO modularize better
+#[cfg(targe_arch = "s390x")]
 pub const POSIX_FADV_DONTNEED: ::c_int = 4;
+#[cfg(targe_arch = "s390x")]
 pub const POSIX_FADV_NOREUSE: ::c_int = 5;
```
alexcrichton: Could these be pushed down and duplicated everywhere except s390x?

alexcrichton:
> tested by running libc-test on real s390x hardware.

Holy cow, nice!

Does `qemu-s390x` not work?

japaric:
> Does qemu-s390x not work?

Sadly, no. At least Ubuntu's doesn't. I also tried compiling QEMU 2.6 myself but that didn't work either. Both qemu-s390xs segfaulted with even the simplest cross compiled C program.

I just saw QEMU 2.7 was released. I'll try that one and hope that the issue got magically fixed.

src/unix/notbsd/mod.rs
japaric: I was going to do that but too tired already :-)

alexcrichton:
> Both qemu-s390xs segfaulted with even the simplest cross compiled C program

One thing I noticed recently was that `qemu-mips`  in a docker container would segfault, but [if you also install](https://github.com/rust-lang/libc/blob/master/ci/docker/mips-unknown-linux-gnu/Dockerfile#L7) the `qemu-system-mips` package then the segfault magically went away. Not sure if that's the case here, but do you perhaps have a "stripped down installation of qemu" however ubuntu defines it maybe?

japaric:
Oh, that sounds weird but I'm not that surprised. I've seen a 64-bit QEMU binary segfault whereas a 32-bit QEMU binary worked fine before. Sadly, the qemu-system-s390x package is only avaliable for the [s390x architecture itself](http://packages.ubuntu.com/xenial/qemu-system-s390x).

japaric:
> Could these be pushed down and duplicated everywhere except s390x?

Done, I think.

PR #379, opened 2016-09-05
Move kqueue-related definitions from nix to libc

asomers:
This PR moves, as much as possible, kqueue and kevent related definitions from the nix crate to libc.  There will be a companion PR to the nix repo.


asomers:
I fixed the macos test failures.  The FreeBSD tests failed because your CI VM is too old.  It's complaining about symbols that are defined in FreeBSD 11 but not in FreeBSD 10.  And I don't know why the OpenBSD test failed.  Could somebody please take a look at it?

asomers:
The macos tests are still failing because OSX is a moving target.  The value of EVFILT_SYSCOUNT has changed twice recently.  What do you want to do about that?  Should we omit it from the bindings?  It's not a very useful definition.

src/unix/bsd/openbsdlike/openbsd.rs
```diff
@@ -392,6 +440,12 @@ extern {
                        serv: *mut ::c_char,
                        servlen: ::size_t,
                        flags: ::c_int) -> ::c_int;
+    pub fn kevent(kq: ::c_int,
+                  changelist: *const ::kevent,
+                  nchanges: ::size_t,
+                  eventlist: *mut ::kevent,
+                  nevents: ::size_t,
+                  timeout: *const ::timespec) -> ::c_int;
```
semarie: `nchanges` and `nevents` should be `c_int`.
But I am unsure about the error message of OpenBSD test: it seems related to return type ?

alexcrichton:
Awesome, thanks! Could perhaps some of the more unstable or up-to-date constants be left out for now? Or do they all need to be added all at once?

asomers:
I can leave out some of the newer constants.  I think I should leave out EVFILT_SYSCOUNT too, because that changes from one release to another.

It strikes me that rather than use FFI, a C module might be more appropriate for these definitions.  It would be able to automatically detect which constants are defined and set their values correctly.  I know how to write such a module for Ruby, but not for Rust.  Is there a way to define publicly visible Rust symbols from C code?

alexcrichton:
For the libc crate itself we wouldn't have any C code, but for something like nix that could work! You can call C from Rust easily with just a `extern { ... }` block and vice-versa if necessary with `#[no_mangle]`.

alexcrichton:
@bors: r+ 

Thanks! Also feel free to ping a PR whenever it's updated :)

PR #383, opened 2016-09-07
Add more s390x definitions

uweigand:
This adds more Linux-specific definitions for s390x, to match the
defines provided for other Linux targets.  Many of these will be
required to build std, rustc, and cargo.

In addition, this changes the type of sigaction.sa_flags to "int"
to match current glibc after this bug fix was applied:
https://sourceware.org/bugzilla/show_bug.cgi?id=16713

Some other (cosmetic) changes to bring types in line with the
exact types used in the glibc headers (e.g. ino64_t instead
of ino_t in statfs64).

Signed-off-by: Ulrich Weigand <ulrich.weigand@de.ibm.com>


alexcrichton:
@bors: r+

Awesome, thanks!

PR #387, opened 2016-09-09
Diff reduction between OpenBSD and Bitrig

asomers:
Bitrig had a number of errors and omissions.
Also, fill out struct glob_t on all BSDs.

A follow-up commit will condense OpenBSD/Bitrig common code into a separate file.


alexcrichton:
@bors: r+

Thanks!

PR #391, opened 2016-09-12
add CI for mips-musl

japaric:
and fix the statvfs struct for this target

r? @alexcrichton 

I didn't really test the CI part but this target now passes libc-test under QEMU

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,6 +80,9 @@ matrix:
       env: TARGET=powerpc64-unknown-linux-gnu
       rust: stable
     - os: linux
+      env: TARGET=mips-unknown-linux-musl
+      rust: stable
+    - os: linux
       env: TARGET=mipsel-unknown-linux-musl
       rust: stable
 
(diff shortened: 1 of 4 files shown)


alexcrichton:
Holy cow. My reaction to this was "surely we don't need to test *both* mips and mipsel, they're so similar!" Of course I'm wrong...

@bors: r+

Thanks @japaric!

japaric:
> "surely we don't need to test both mips and mipsel, they're so similar!" 

#EndiannessMatters

> this target now passes libc-test under QEMU

(Shameless plug) I realized this wasn't working thanks to [smoke](https://github.com/japaric/smoke), which is looking more "green" these days.

alexcrichton:
> which is looking more "green" these days

🎈

alexcrichton:
Looks like the CI failures are legitimate?

japaric:
> Looks like the CI failures are legitimate?

Fixed the SDK install.

Interestingly, glibc doesn't change the order of the f_unused and f_fsid fields according to endianness like musl does but that still works for glibc ... ? (perhaps because of C alignment rules?).

Hiding (making it a private field) f_unused seems to "fix" the problems observed on powerpc. Is that OK? being named __f_unused, I suppose that means the field is supposed to be private by C conventions.

alexcrichton:
Ah yeah hiding it's fine by me!

@bors: r+

alexcrichton:
That one also looks somewhat legitimate?

japaric:
> That one also looks somewhat legitimate?

Yes, I think I stripped too many components and ended up flatteing some directories inside `include`. Will take a look later.

japaric:
OK, that should do it.

japaric:
OK, it's fixed for real this time. (I actually ran the test locally)

PR #394, opened 2016-09-14
Adding accept4 function

raphaelcohn:
accept4 is a version of accept that takes flags.
It is more efficient for certain use cases, and has
a more modern path inside the kernel.


alexcrichton:
@bors: r+

Thanks!

PR #396, opened 2016-09-15
Added SSIZE_MAX for all platforms. It's always isize::MAX

raphaelcohn:
This is a new PR, which replaces the previous one which
became polluted with non-relevant changes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -133,6 +133,9 @@ pub type ssize_t = isize;
 pub enum FILE {}
 pub enum fpos_t {} // TODO: fill this out with a struct
 
+#[cfg(not(any(dox, target_env = "msvc")))]
+pub const SSIZE_MAX: ssize_t = core::isize::MAX as ::ssize_t;
+
 extern {
     pub fn isalnum(c: c_int) -> c_int;
     pub fn isalpha(c: c_int) -> c_int;


src/lib.rs
```diff
@@ -133,6 +133,9 @@ pub type ssize_t = isize;
 pub enum FILE {}
 pub enum fpos_t {} // TODO: fill this out with a struct
 
+#[cfg(not(any(dox, target_env = "msvc")))]
```
alexcrichton: Ideally this `#[cfg]` would get removed. Could the `dox` module get shimmed and is this just not defined on MSVC?

raphaelcohn: I actually don't know what the purpose of `dox` is; I'm not sure what a shim would look like. Do you mean adding a definition of `isize::MAX` inside `doc.rs`?

As I've discovered, the Windows MSCV ABI doesn't define `SSIZE_MAX` but the Windows GNU ABI does. I suppose the correct solution might be to break out a msvc and a gnu modules, and use `cfg_if` to use them. What do you think?

PS: I can't work on any of this now until Monday.

alexcrichton: Ah yeah let's split this between `src/unix` and `src/windows` if that's the case, and then yeah you'll want to add something in `doc.rs` so the travis builds can continue to build documentation for this

raphaelcohn: Ok, I'm just about back in the land of the living after a week long standards love in. I'm happy to add this to `src/unix`, but to add it to `src/windows` would require quite an invasive change - in essence, moving `windows.rs` to `windows/mod.rs` and then introducing `windows/msvc.rs` and `windows/gnu.rs`, say. I'm not completely comfortable doing that for one teensy constant... especially as I don't have the local infrastructure to test my changes thoroughly on windows.

alexcrichton: Ah yeah that's fine, want to put this in `src/unix/mod.rs` for now?

alexcrichton:

PR #402, opened 2016-09-19
Add EAI_SYSTEM getaddrinfo error code.

tmiasko:
This is useful to decide if value returned from getaddrinfo is actual
error or you should look at errno instead.

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -621,6 +621,8 @@ pub const EOWNERDEAD: ::c_int = 105;
 pub const EQFULL: ::c_int = 106;
 pub const ELAST: ::c_int = 106;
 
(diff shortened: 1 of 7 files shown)


bvinc:
This is needed for rust-lang/rust#36546.  I was looking at this earlier.  Looks good to me.

semarie:
bitrig and openbsd definitions could be factorized in `src/unix/bsd/netbsdlike/openbsdlike/mod.rs`

tmiasko:
@semarie moved OpenBSD and Bitrig changes into common openbsdlike mod, thanks.

@bvinc  indeed, the issue rust-lang/rust#36546 was original motivation for this change.

alexcrichton:
Thanks!

tmiasko:
Rebased.

PR #404, opened 2016-09-23
add time, localtime

philippkeller:
time, localtime and strftime are all in POSIX.1 and should IMO all be part of `libc`.

I [tested the bindings](https://github.com/philippkeller/apue-rust/blob/master/src/bin/f06-11-strftime.rs) - works fine on OSX and linux.

And: I added the netbsd aliases based on `time.h` of netbsd


philippkeller:
I'll try to fix the build, it currently fails with `bad strftime function pointer: rust: 2473012243 (0x93672c13) != c 2472821892 (0x93644484)` for `i686-apple-darwin`.

If anyone has any hints what could be wrong I'd be happy. I guess I need to get the headers for OSX i686 somewhere (though I have no idea from where) and run bindgen on them and see how the function definition differs from the other architectures?

alexcrichton:
@bors: r+ 

Thanks!

PR #406, opened 2016-09-26
Add shadow password bindings

philippkeller:
Add shadow password bindings: setspent, endspent, getspent, getspnam plus spwd, the corresponding struct.

[Example code](https://github.com/philippkeller/apue-rust/blob/master/src/bin/e06-01-shadow-pw.rs) (tested on Linux 4.4.0-21-generic)


alexcrichton:
Looks like the CI failures may be legitimate, perhaps a missing header file?

philippkeller:
@alexcrichton thanks for the tip. I added the header file and now the build is green

PR #411, opened 2016-09-29
Add `pthread_mutex_timedlock`

apasel422:


alexcrichton:
Looks like this may not exist on OSX?

apasel422:
Yeah, I have to investigate further.

apasel422:
What's the proper organization for including this in Linux and BSD but not OSX or iOS?

alexcrichton:
It's fine to duplicate the definition in the non-apple BSD modules

PR #413, opened 2016-10-02
Define CLONE_NEWCGROUP

zenhack:
Should be fairly self explanatory.

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -248,6 +248,8 @@ pub const ABMON_10: ::nl_item = 0x20017;
 pub const ABMON_11: ::nl_item = 0x20018;
 pub const ABMON_12: ::nl_item = 0x20019;
 
+pub const CLONE_NEWCGROUP: ::c_int = 0x02000000;
+
 pub const MON_1: ::nl_item = 0x2001A;
 pub const MON_2: ::nl_item = 0x2001B;
 pub const MON_3: ::nl_item = 0x2001C;
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -281,6 +281,9 @@ pub const CODESET: ::nl_item = 14;
 
 pub const CRNCYSTR: ::nl_item = 0x4000F;
 
+pub const RUSAGE_THREAD: ::c_int = 1;
+pub const RUSAGE_CHILDREN: ::c_int = -1;
+
 pub const RADIXCHAR: ::nl_item = 0x10000;
 pub const THOUSEP: ::nl_item = 0x10001;
 
--- a/src/unix/notbsd/linux/musl/b32/arm.rs
+++ b/src/unix/notbsd/linux/musl/b32/arm.rs
@@ -332,7 +332,5 @@ pub const TIOCCONS: ::c_int = 0x541D;
 pub const SYS_gettid: ::c_long = 224;
 pub const SYS_perf_event_open: ::c_long = 364;
 
-pub const POSIX_MADV_DONTNEED: ::c_int = 0;
-pub const RUSAGE_CHILDREN: ::c_int = 1;
 pub const POLLWRNORM: ::c_short = 0x100;
 pub const POLLWRBAND: ::c_short = 0x200;
--- a/ci/docker/aarch64-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/aarch64-unknown-linux-gnu/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:16.10
 RUN apt-get update
 RUN apt-get install -y --no-install-recommends \
   gcc libc6-dev ca-certificates \
--- a/ci/docker/arm-unknown-linux-gnueabihf/Dockerfile
+++ b/ci/docker/arm-unknown-linux-gnueabihf/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:16.10
 RUN apt-get update
 RUN apt-get install -y --no-install-recommends \
   gcc libc6-dev ca-certificates \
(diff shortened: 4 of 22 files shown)


alexcrichton:
Hm looks like the CI errors might be relevant?

zenhack:
I worked out where the travis errors are coming from: the c library versions we're testing with are too old.

We're pulling in musl 1.1.14, and a definition for CLONE_NEWCGROUP was added sometime after that.

I haven't poked at the system headers on ubuntu 16.04, but the kernel for that release is 4.4, and the flag was introduced in to the kernel in 4.6, so it stands to reason that it's absent there as well.

The tests pass on my local system (archlinux, kernel version 4.7.5, glibc).

It would be straightforward enough to pull in a more recent version of musl. I don't see an immediate reason why upgrading to 16.10 would be unworkable, though it's not an LTS release so there's that.

Also, there's a *lot* of code duplication in `ci/docker`. Just saying.

Thoughts?

alexcrichton:
That all sounds like the source to me! Want to update the various docker images used in testing? Note that the duplication is indeed intentional! Means it's easy to have differences among platforms. We don't need to keep things similar amongst platforms

zenhack:
Done. I'm now getting a different error on the musl builds, which I'm still investigating. There's also a failure on android which I hadn't noticed before. It looks like the definition of CLONE_NEWCGROUP hasn't hit a release of bonic yet, and it doesn't seem to be included in the betas for the next release (though it is on master). Don't know how you want to handle that. I'll keep poking at the musl failures.

alexcrichton:
Looks like the 32-bit musl compiler failed during the build:

```
./configure: cannot find a C compiler
```

Also looks like android doesn't have `CLONE_NEWCGROUP` defined? Perhaps it can be omitted on Android?

For 64-bit musl looks like a few constants changed, namely `POSIX_MADV_DONTNEED` and `RUSAGE_CHILDREN`

zenhack:
`git blame` on the musl source reveals that the changed constants are bug fixes, and the new versions agree with the ones in `linux/other`. Since these are defined by the kernel ABI, any mus-specific deviation would be a bug, so I've just removed the definitions from the musl-specific modules. Don't have any easy way of testing short of pushing, so we'll see what travis does with that.

There were also some changes to musl's `configure` re: finding the compiler. When I get the chance I'll poke around with a 16.10 system and see if I can't find the right incantation.

re: android, it seems a shame not to include it just because of the test suite; it's a kernel defined feature, so libc is actually irrelevant here; it should still work. And it's in master so it will land sooner or later.

src/unix/notbsd/linux/musl/b32/arm.rs
```diff
@@ -332,7 +332,5 @@ pub const TIOCCONS: ::c_int = 0x541D;
 pub const SYS_gettid: ::c_long = 224;
 pub const SYS_perf_event_open: ::c_long = 364;
 
-pub const POSIX_MADV_DONTNEED: ::c_int = 0;
-pub const RUSAGE_CHILDREN: ::c_int = 1;
```
alexcrichton: Backwards compatibility requires that we don't delete these definitions, so perhaps the values could just get corrected with the updated ones?

zenhack: Perhaps I've misunderstood how the crate is organized? I was under the impression that it would just fall back to the definition in `linux/mod.rs`. Looking now, I see that RUSAGE_CHILDREN isn't there, but with this change the value is the same across all linux platforms (as it should be, since it's part of the kernel ABI), so (unless I've misunderstood how the crate works) could be moved up into mod.rs.

If my understanding is off I can just correct the values in each location.

zenhack:
Just pushed something that fixed musl/i686 locally, but still breaks on travis (already know why). Don't mind me while I fuss with this...

zenhack:
I fixed the i686-musl problems, and restricted the definition of CLONE_NEWCGROUP to non-android linux. Per the comment, it will be a while before it's available from the NDK.

I also defined those other symbols in the higher-level modules, so they should be still be available on the systems that previously defined the same constants in each file. Please let me know if I've misunderstood something.

All of the CI tests are passing now.

I would like to do a bit of history cleanup before this gets merged, but let me know what you think of the overall diff first.

src/unix/notbsd/linux/mod.rs
```diff
@@ -248,6 +248,11 @@ pub const ABMON_10: ::nl_item = 0x20017;
 pub const ABMON_11: ::nl_item = 0x20018;
 pub const ABMON_12: ::nl_item = 0x20019;
 
+// This is defined here instead of in ../mod.rs because Android doesn't expose
+// it yet; the change is on master, but NDK 12 beta doesn't include it, so it
+// will likely be absent for a while.
```
alexcrichton: It's ok to omit comments like these, we tend to not leave comments generally in libc (as it's just bindings)

src/unix/notbsd/linux/musl/b32/arm.rs
alexcrichton: Ah sorry I think I missed the movement up, in which case that's totally fine!

zenhack:
Quoting Alex Crichton (2016-10-04 11:01:21)

>    Ah sorry I think I missed the movement up, in which case that's totally
>    fine!

No, I actually forgot it on the first pass (though it had been my
intent); you didn't miss anything.

zenhack:
Quoting Alex Crichton (2016-10-04 11:00:49)

> +// This is defined here instead of in ../mod.rs because Android doesn't expose
> +// it yet; the change is on master, but NDK 12 beta doesn't include it, so it
> +// will likely be absent for a while.
>
>    It's ok to omit comments like these, we tend to not leave comments
>    generally in libc (as it's just bindings)

Done. Let me know if there's anything else you'd like addressed. If not
I'll go ahead and do the history cleanup.

alexcrichton:
Looks good to me, thanks! I'll r+ w/ a squash

zenhack:
I just pushed an alternate history. The final diff is identical, but this tells a more coherent story. I don't know what the policy is on intermediate commits failing CI; let me know if this is a problem.

Wasn't sure hwo to interpret your comment; it sounded like you might have planned on just squashing the whole thing? Whatever you prefer.

alexcrichton:
@bors: r+

Thanks!

PR #417, opened 2016-10-05
Added all additional ptrace options.

valarauca:
Added additional PTrace Flags.

All the `PTRACE_O` are flags options values that are passed with `PTRACE_OPTION`. 

The `PTRACE_PEEKSIGINFO_SHARED` is used for setting/fetching signal masks when interacting with thread groups not individual threads. 

`SYSEMU` stuff I pulled from the kernel sources. As far as I can tell these flags haven't been added to `<sys/ptrace.h>`, which is weird because the two `SYSEMU` calls have been live on x64 since 2011. I'm starting a thread on kernel-newies about it.


valarauca:
Okay so Travis-CI is confirming the behavior I'm seeing on my local PC which is that `<sys/ptrace.h>` doesn't have the 2 `SYSEMU` flags populated in them.

I understand if you want to reject the patch on those grounds. I can rework it so only the option flags are being added. But the [`SYSEMU` flags are the correct value.](https://github.com/torvalds/linux/blob/9256d5a308c95a50c6e85d682492ae1f86a70f9b/arch/x86/um/shared/sysdep/ptrace_32.h)

alexcrichton:
@bors: r+

Thanks!

PR #418, opened 2016-10-05
Added the waitid interface

valarauca:
Waitid allows for a non-blocking way to check the status of individual threads, groups of threads, or every thread on the system.

Not 100% sure about how the linking is handled. Also can't confirm that it exists for *every* platform currently. I'm more or less gonna use travis for that.

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -20,9 +20,9 @@ pub type nl_item = ::c_int;
 pub type idtype_t = ::c_int;
 pub enum fpos64_t {} // TODO: fill this out with a struct
 
-pub const P_PID: ::id_t = 1;
-pub const P_PGID: ::id_t = 2;
-pub const P_ALL: ::id_t = 3;
+pub const P_PID: ::idtype_t = 1;
+pub const P_PGID: ::idtype_t = 2;
+pub const P_ALL: ::idtype_t = 3;
 
 
 s! {
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -17,8 +17,13 @@ pub type msgqnum_t = ::c_ulong;
 pub type msglen_t = ::c_ulong;
 pub type nfds_t = ::c_ulong;
 pub type nl_item = ::c_int;
-
+pub type idtype_t = ::c_int;
 pub enum fpos64_t {} // TODO: fill this out with a struct
+
+pub const P_PID: ::id_t = 1;
+pub const P_PGID: ::id_t = 2;
+pub const P_ALL: ::id_t = 3;
+
 
 s! {
     pub struct dirent {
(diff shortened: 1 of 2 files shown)


valarauca:
Invalid pointer type attempting fix.

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -548,6 +548,13 @@ pub const NLA_F_NESTED: ::c_int = 1 << 15;
 pub const NLA_F_NET_BYTEORDER: ::c_int = 1 << 14;
 pub const NLA_TYPE_MASK: ::c_int = !(NLA_F_NESTED | NLA_F_NET_BYTEORDER);
 
+#[repr(C)]
+pub enum idtype_t {
+    P_ALL,
+    P_PID,
+    P_PGID
+}
```
alexcrichton: Right now we actually tend to avoid literal `enum` types as they're not extensible (while the C side is), so perhaps these could just be `pub const P_ALL: c_int = i32`?

alexcrichton: Yeah let's take the route of listing all the constants explicitly.

valarauca: I lifted the the constant definition to `linux/mod.rs` from `/linux/other/mod.rs` because `sys/wait.h` _should_ be standard cross all GNU/Linux platforms. 

The type+explicitly declared constants is added. Waiting for Travis-ci to report.

src/unix/notbsd/linux/mod.rs
```diff
@@ -17,9 +17,14 @@ pub type msgqnum_t = ::c_ulong;
 pub type msglen_t = ::c_ulong;
 pub type nfds_t = ::c_ulong;
 pub type nl_item = ::c_int;
-
+pub type idtype_t = ::c_int;
 pub enum fpos64_t {} // TODO: fill this out with a struct
 
+pub const P_PID: ::id_t = 1;
+pub const P_PGID: ::id_t = 2;
+pub const P_ALL: ::id_t = 3;
```
alexcrichton: Should these be `idtype_t`?

valarauca: Update.

Also while I have your attention. How does one go about adding the `linux/*.h` sources for linking? 

I can see where to add them for the test suite/crate. This was the main issue why I couldn't merge the 2 remaining for `PTRACE` yesterday/last night.

alexcrichton:
Looks like CI may still be failing?

valarauca:
Style error I'll submit a commit in the morning got distracted today

valarauca:
I'm not sure how to explain this failing. 

On a lot of platforms the are running tests for `idtype_t` but this is type not a constant(?)

alexcrichton:
It looks like there's failures across various platforms, for examples mips-musl has `idtype_t` as unsigned and a different value for `P_ALL`

PR #424, opened 2016-10-10
Make the AF_NETLINK constant available for Android

mmatyas:
Currently the `AF_NETLINK` constant is not available for Android, but [it has the same value as on Linux](https://android.googlesource.com/kernel/common/+/android-4.4.y/include/linux/socket.h#177). To fix it, this patch moves the definition to the common `notbsd/mod.rs`.

(this fixes an [issue we found in Servo](https://github.com/servo/servo/issues/13154#issuecomment-251325701))


alexcrichton:
@bors: r+

Thanks!

PR #432, opened 2016-10-20
Add support for Fuchsia

raphlinus:
These patches add support for the Fuchsia operating system. For the
time being, it shares a lot of infrastructure with the Linux target,
because of the use of a musl-based libc.

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -231,7 +231,6 @@ cfg_if! {
     } else if #[cfg(target_os = "fuchsia")] {
         #[link(name = "c")]
         #[link(name = "mxio")]
-        #[link(name = "unwind")]
         extern {}
     } else {
         #[link(name = "c")]
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -228,6 +228,11 @@ cfg_if! {
         #[link(name = "root")]
         #[link(name = "network")]
         extern {}
+    } else if #[cfg(target_os = "fuchsia")] {
+        #[link(name = "c")]
+        #[link(name = "mxio")]
+        #[link(name = "unwind")]
+        extern {}
     } else {
         #[link(name = "c")]
         #[link(name = "m")]
@@ -843,7 +848,8 @@ extern {
 cfg_if! {
     if #[cfg(any(target_os = "linux",
                  target_os = "android",
-                 target_os = "emscripten"))] {
+                 target_os = "emscripten",
+                 target_os = "fuchsia"))] {
         mod notbsd;
         pub use self::notbsd::*;
     } else if #[cfg(any(target_os = "macos",
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -728,6 +728,7 @@ extern {
 
 cfg_if! {
     if #[cfg(any(target_env = "musl",
+                 target_os = "fuchsia",
                  target_os = "emscripten"))] {
         mod musl;
         pub use self::musl::*;
--- a/src/unix/notbsd/mod.rs
+++ b/src/unix/notbsd/mod.rs
@@ -53,7 +53,7 @@ s! {
         pub ai_protocol: ::c_int,
         pub ai_addrlen: socklen_t,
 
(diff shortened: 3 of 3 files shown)


src/unix/mod.rs
```diff
@@ -228,6 +228,11 @@ cfg_if! {
         #[link(name = "root")]
         #[link(name = "network")]
         extern {}
+    } else if #[cfg(target_os = "fuchsia")] {
+        #[link(name = "c")]
+        #[link(name = "mxio")]
+        #[link(name = "unwind")]
```
raphlinus: This doesn't seem right to me, but I haven't been able to figure out a better way to get it to work. I tried adding it as `#cfg[attr(..., link(...))]` at the end of https://github.com/rust-lang/rust/blob/master/src/libunwind/libunwind.rs , but that doesn't seem to show up in the final link line.

src/unix/notbsd/linux/mod.rs
```diff
@@ -728,6 +728,7 @@ extern {
 
 cfg_if! {
     if #[cfg(any(target_env = "musl",
+                 target_os = "fuchsia",
```
raphlinus: So right now this is under linux, because that's where musl lives, and because the diff is really minimal, but I'm not sure that's the best long-term structure. I think it might be better for musl to be a toplevel under notbsd, and have both linux-musl and fuschsia go there, with additional fuchsia `#[cfg]` items there as needed.

posborne:
@raphlinus Would it be possible to add Fuchsia to the set of targets that are tested by CI (see https://github.com/rust-lang/libc/blob/master/ci/README.md)?  That is probably the best way to ensure compatibility and support moving forward.  For this, we would need to be able to run fuchsia under docker or qemu (guessing qemu would be required in this case).

src/unix/notbsd/linux/mod.rs
alexcrichton: This is fine for now, but if the two start to diverge we can just copy musl over and update fuschia as necessary

src/unix/mod.rs
alexcrichton: I think `unwind` actually comes in at the libstd layer rather than the libc layer (e.g. those symbols aren't referenced from this library). Other than that though this looks fine to me!

alexcrichton:
I'd love to add CI as well, but we need to get compiler targets first to ensure we've got a standard library to download. In that sense it's fine by me to punt that to later

src/unix/mod.rs
alexcrichton: Ah yeah note that if it's removed here it'll need to be added to `src/unwind/build.rs` most likely (as that's where the unwinder symbols are defined

raphlinus:
It is certainly possible to run Fuchsia under qemu (see https://fuchsia.googlesource.com/fuchsia/ for instructions how to build the image), and I would like to get CI running. However, things are still stabilizing, so as @alexcrichton suggested I'd like to defer that to later.

src/unix/mod.rs
raphlinus: Thanks! That's exactly what I was missing. Will respin...

src/unix/notbsd/mod.rs
```diff
@@ -53,7 +53,9 @@ s! {
         pub ai_protocol: ::c_int,
         pub ai_addrlen: socklen_t,
 
-        #[cfg(any(target_os = "linux", target_os = "emscripten"))]
+        #[cfg(any(target_os = "linux",
+                  target_os = "emscripten",
```
tedsta: Late nitpick - couldn't this be:

```
        #[cfg(any(target_os = "linux",
                        target_os = "android",
￼                        target_os = "emscripten",
                        target_os = "fuchsia"))]
        pub ai_addr: *mut ::sockaddr,
```

rather than having the android entry be separate at line 63?

alexcrichton:
Looks like CI may be failing on at least Android

src/unix/notbsd/mod.rs
raphlinus: So it turns out that doesn't work, because the structure is serialized, and order matters. Reverted.

tedsta: Whoops

PR #433, opened 2016-10-21
Try to fix powerpc64 and update to crates.io ctest

alexcrichton:


posborne:
@alexcrichton Is there another issue tracking why powerpc64 is missing from 1.12.1?  I ran into this when updating meta-rust-bin for 1.12.1: https://github.com/rust-embedded/meta-rust-bin/pull/9

PR #434, opened 2016-10-21
Add setresuid/setresgid for linux

dgreid:


alexcrichton:
Hm I'll need to debug what's going on there...

PR #439, opened 2016-11-03
Minimal liblibc for Redox

jackpot51:
This implements a minimal liblibc for Redox. Previously, this was talked about here: https://internals.rust-lang.org/t/redox-support-in-liblibc-and-libstd/3954

Merging this would allow Redox to port more Rust software, and will also allow libstd to be ported to Redox.

Changes to src/lib.rs are minimal, and the src/redox.rs file is also minimal but may be expanded in the future.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -265,7 +265,7 @@ cfg_if! {
     if #[cfg(windows)] {
         mod windows;
         pub use windows::*;
-    } else if #[cfg(redox)] {
+    } else if #[cfg(target_os = "redox")] {
         mod redox;
         pub use redox::*;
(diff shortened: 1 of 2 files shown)


jackpot51:
One question I have to whoever is assigned this - should I put the structs I have defined for networking inside of `s!`?

jackpot51:
This is in preperation for a much larger change in Rust's libstd, which can be seen in progress here: https://github.com/rust-lang/rust/compare/master...redox-os:redox

jackpot51:
@alexcrichton How does this look?

jackpot51:
@alexcrichton Can a new cargo version be published?

PR #441, opened 2016-11-05
Add setfsgid()/setfsuid() on linux

billiob:
Tested with the following code ran as root and then looking at the owner/group of `/tmp/foo.txt`.
```rust
extern crate libc;

use std::fs::File;

fn main() {
    unsafe {
        libc::setfsuid(1234u32);
        libc::setfsgid(5678u32);
    }
    File::create("/tmp/foo.txt").ok();
}
```


alexcrichton:
@bors: r+

Thanks!

PR #443, opened 2016-11-06
Fix platform constants not being exposed on mips64el

xen0n:
Just noticed this while trying to bring `rustup` to mips64el, which is almost certainly an oversight.

More work is needed to get `rustup` working on mips64el, but I'm submitting this first. More fixes to come.

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -742,7 +742,7 @@ cfg_if! {
     } else if #[cfg(any(target_arch = "s390x"))] {
         mod s390x;
         pub use self::s390x::*;
-    } else if #[cfg(any(target_arch = "mips64"))] {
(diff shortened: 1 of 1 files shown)


xen0n:
(I believe the Travis error is spurious, likely a HTTP 503 or 504.)

alexcrichton:
Thanks for the PR! I believe though that the target arch is actually `mips64` for `mips64el-unknown-linux-gnu`. Did you find that this fixed compilation failures for you though?

xen0n:
Oops, I realized immediately that mips64el's arch is also `mips64`, this is different from `mipsel`... Hey, the target_arch `mipsel` itself is bogus, the MIPS32 targets also use `mips` as target_arch!

(The reason the missing `libc` constants seemed to not show up was a mistake in overriding deps for Cargo, I believe. I tweaked too many things at once and lost track of what caused what. Closing.)

PR #444, opened 2016-11-06
add pledge(2) syscall.

semarie:
this PR adds pledge(2) syscall to bitrig and openbsd.


alexcrichton:
Looks like CI failures are legit?

PR #445, opened 2016-11-07
Remove mipsel target_arches

xen0n:
The `mipsel` `target_arch` was introduced with the initial MIPSel support (rust-lang/rust@82ec1aef293ddc5c6373bd7f5ec323fafbdf7901), but was subsequently removed with implementation of the Flexible Target Specification (Rust RFC 0131, rust-lang/rust@3a8f4ec32a80d372db2d02c76acba0276c4effd0). Remove these to be consistent with other arches and reduce confusion.

I've very slightly formatted the deletions so as to not require a style cleanup later and obscure `git blame` output.

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -66,21 +66,21 @@ s! {
     }
 
     pub struct pthread_mutex_t {
-        #[cfg(any(target_arch = "mips", target_arch = "mipsel",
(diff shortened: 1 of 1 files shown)


xen0n:
(I was confused by this myself, and this is why I incorrectly done #443 when, in fact, the problem was that `[replace]`s inside sub-`Cargo.toml`s (instead of the root-level one) are ignored, not picking up my changes.)

alexcrichton:
Thanks!

PR #449, opened 2016-11-13
POSIX AIO definitions, for BSD and Linux

asomers:
Where structures included unions, I only defined the most useful member.  This
means that sigevents can't be delivered using SIGEV_THREAD until Rust gets better
union support (SIGEV_THREAD_ID still works).

I'll followup with PRs for nix and mio.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -430,9 +430,13 @@ fn main() {
             // the symbol.
             "uname" if freebsd => true,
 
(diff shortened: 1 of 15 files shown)


asomers:
It seems that most of the ci test failures are due to the liberties I took with unions.  I guess I'll have to expand them all.

alexcrichton:
Awesome, thanks!

PR #451, opened 2016-11-13
Remove constants no longer present in FreeBSD 11

asomers:
FreeBSD svn r262489 removed some *_MAXID definitions in sys/sysctl.h.  They never
should've been used outside of the FreeBSD base system.  r273250 removed
MAP_RENAME and MAP_NORESERVE, flags used by mmap(2).  Old binaries that use them
will still work.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -361,6 +361,19 @@ fn main() {
             "QFMT_VFS_OLD" |
             "QFMT_VFS_V0" if mips && linux => true,
 
+            // These constants were removed in FreeBSD 11 (svn r273250) but will
+            // still be accepted and ignored at runtime.
+            "MAP_RENAME" |
+            "MAP_NORESERVE" if freebsd => true,
+
+            // These constants were removed in FreeBSD 11 (svn r262489),
+            // and they've never had any legitimate use outside of the
+            // base system anyway.
+            "CTL_MAXID" |
+            "KERN_MAXID" |
+            "HW_MAXID" |
+            "USER_MAXID" if freebsd => true,
+
             _ => false,
         }
     });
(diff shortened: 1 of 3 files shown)


asomers:
The only test failure was a failure setting up the VM on Linux mips.

alexcrichton:
Thanks for the PR! We typically don't try to break the API of the libc crate, such as removing constants like these. Was there a particular motivation for removing these?

asomers:
The motivation is that they break the ci-tests on FreeBSD 11.  If you'd prefer, we can leave the MAP_RENAME and MAP_NORESERVE flags, but disable their checks in libc-test/build.rs.  However, we should definitely remove all the *_MAXID definitions.  It's a bug for any 3rd party software to use them.

alexcrichton:
Oh ok yeah for that let's just add some exceptions

asomers:
I combined PR #452 and PR #451 and rewrote history so there should now be only a single commit between them.

alexcrichton:
Could the MAXID constants also stick around?

asomers:
I'd really rather not.  Those constants never should've been used outside of the FreeBSD base system's sysctl(8).  Furthermore, they were never intended to be stable across versions.  For example, KERN_MAXID had a value of 38 in FreeBSD 10, but in FreeBSD 11 (if it still existed), it would be 39.  And since rust-lang/libc isn't maintained separately for different versions of the OS, it is therefore a bug for any program to use those constants.

FWIW, sysctl(8) itself hasn't used those constants since r13239, which was released in FreeBSD 2.2 in 1997.

alexcrichton:
Unfortunately we can't land this as it's a breaking change to the API. Perhaps these could be `#[doc(hidden)]` annotated and excepted in the libc-test build script?

asomers:
It's a bug for any libc customer to be using those symbols.  Wouldn't you rather they catch the bug at build time instead of run time?

alexcrichton:
These are separate concerns. Constants can't be removed from the libc crate, period. What to do about preventing usage once they're added is a separate concern. We could perhaps use deprecated annotations or similar to deter usage, but we simply cannot remove them.

asomers:
Ok, you're the boss.  Would you at least consider removing them when bumping the major version of libc?  Semver allows that.

alexcrichton:
Certainly!

asomers:
Ok, I left them in.  Shall I open an issue so we don't forget to remove them later?  Or how do you want to track that?

alexcrichton:
@bors: r+ 

Thanks!

Also yeah if you want to open an issue I'll make a tag for breakage.

PR #453, opened 2016-11-14
MIPS: Layout refactor

xen0n:
MIPS32 and MIPS64 userspaces are mostly the same save the obvious ABI difference, hence most of the const definitions could be shared.

This commit is not introducing any new significant additions, but mostly just shuffles the existing files around to make the directory structure more similar to the other, more mature platforms. The structs and extern fn definitions are not touched. Plus a couple of new consts for MIPS64 to match MIPS32 coverage:

* `RLIM_INFINITY`
* `SYS_gettid`

Passed `libc-test` on Gentoo MIPS64el (n64 ABI), on a Loongson 3A2000 box. The test suite itself needs some more love to actually compile cleanly on MIPS (`ctest` transitively depends on ancient `libc` version which doesn't build on MIPS), but let's save that for another day.


alexcrichton:
Awesome, thanks! Want to rebase and I'll merge?

xen0n:
@alexcrichton Sure, done! 😹

PR #457, opened 2016-11-23
Added setpgrp

zethra:


alexcrichton:
Thanks for the PR! Looks like the CI failures are vaild?

zethra:
Not sure what those CI errors about. Those systems don't seem to be able to find the function.  Any suggestions?

zethra:
Not sure why those few targets failed.  Any suggestions?

alexcrichton:
It looks like this API may not exist on those platforms, perhaps it could be moved farther down in the sourc ehierarchy?

semarie:
additionnally, not all platforms follows strictly POSIX.

under OpenBSD, `setpgrp()` is identical to `setpgid()`, and is retained for calling convention compatibility with historical versions of BSD (see http://man.openbsd.org/2/setpgrp).

PR #459, opened 2016-11-29
Make openbsdlike to support 64 and 32 bits archs

semarie:
It is a preliminary work to add a new target `i686-unknown-openbsd`.


alexcrichton:
Thanks!

PR #462, opened 2016-12-02
i686-unknown-openbsd libc correction

semarie:
Now I had been able to run testsuite on i686-unknown-openbsd, here a batch of corrections.

- `suseconds_t` is `c_long` (`i64` or `i32` depending the arch)
- struct siginfo_t padding adjustment (the pointer `si_addr` makes the pad to change depending pointer size)

I tested the PR on i686-unknown-openbsd and x86_64-unknown-openbsd.

### DIFF
--- a/src/unix/bsd/netbsdlike/openbsdlike/mod.rs
+++ b/src/unix/bsd/netbsdlike/openbsdlike/mod.rs
@@ -1,5 +1,5 @@
 pub type clock_t = i64;
-pub type suseconds_t = i64;
+pub type suseconds_t = ::c_long;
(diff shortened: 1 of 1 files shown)


semarie:
@alexcrichton is it possible to release a new (minor) version of libc ? it is a bit annoying to have to patch 0.2.18 to have proper libc supports for openbsd i686. thanks.

alexcrichton:
@semarie sure yeah, that's https://github.com/rust-lang/libc/pull/487

semarie:
thanks !

PR #465, opened 2016-12-04
Add needed structs for fd passing

GuillaumeGomez:


alexcrichton:
Looks good to me, but seems CI is failing?

GuillaumeGomez:
Seems like fields are missing, strange. Taking a deeper look.

GuillaumeGomez:
All fixed!

GuillaumeGomez:
But maybe not (I added missing const).

GuillaumeGomez:
Ok, I forget about consts, too much troubles.

src/unix/notbsd/linux/musl/b64/mod.rs
```diff
@@ -114,6 +114,13 @@ s! {
         pub msg_flags: ::c_int,
     }
 
+    pub struct cmsghdr {
+        pub cmsg_len: ::socklen_t,
+        pub __pad1: ::c_int,
```
semarie: does `__pad1` need to be public ?

GuillaumeGomez: I think it should. After all it's a C binding, so people can do whatever they want.

PR #474, opened 2016-12-15
Fix doc generation on travis.

LegNeato:
Documentation generation appears to be failing on master (https://travis-ci.org/rust-lang/libc/jobs/183483333):

```
Collecting ghp-import
  Downloading ghp-import-0.4.1.tar.gz
Collecting travis
  Could not find a version that satisfies the requirement travis (from versions: )
No matching distribution found for travis
```

Basically `--user` doesn't appear to take an option, so `pip install ghp_import --user $USER` makes pip think the user is a package that needs to be installed (in this case `travis`). As there is no `travis` package, it dies.


alexcrichton:
@bors: r+

Thanks!

PR #477, opened 2016-12-21
Added constants from prctl.h

zethra:
I left in the comment from prctl.h I can remove them if you want.


alexcrichton:
Thanks for the PR! Looks like CI is failing due to line length limits, and yeah in general it's ok to remove comments from this library as this isn't really the authoritative source for how to use all these values.

alexcrichton:
Looks good to me, but looks like travis failures may be legitimate?

zethra:
@alexcrichton Any idea how to fix the one that failed?  I'm not sure why a few specific constants are having a problem on just android.

alexcrichton:
@zethra you can just move those constants down a level in the module hierarchy to the linux-specific module so they're not picked up on Android.

alexcrichton:
@bors: r+

Thanks!

PR #478, opened 2016-12-21
Add memalign to Redox (for use in allocator)

jackpot51:
This is required to satisfy a review item here: https://github.com/rust-lang/rust/pull/38401


jackpot51:
Thanks @alexcrichton !

PR #481, opened 2016-12-29
Add execl, execle, execlp

philippkeller:
I tried to be as close as possible to the existing definitions of `execv` and friends.

btw: do you still welcome these PR? Doesn't the [new bindgen/build.rs method](http://fitzgeraldnick.com/2016/12/14/using-libbindgen-in-build-rs.html) offer to automate this crate to 100%

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -377,6 +377,12 @@ extern {
     pub fn close(fd: ::c_int) -> ::c_int;
     pub fn dup(fd: ::c_int) -> ::c_int;
     pub fn dup2(src: ::c_int, dst: ::c_int) -> ::c_int;
+    pub fn execl(path: *const c_char,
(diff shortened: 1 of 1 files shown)


alexcrichton:
@bors: r+

Looks good to me, thanks!

philippkeller:
@alexcrichton I'm planning to add a few more of those additions (gut feeling is about 5-10): should I still do this in the light of the [new bindget/build.rs method](http://fitzgeraldnick.com/2016/12/14/using-libbindgen-in-build-rs.html)? And if yes, is it ok if I package those into smaller PRs (would be easier for me) or would you prefer one bigger PR?

alexcrichton:
Ah yeah for now with libc it's easiest to just add them by hand. Everything is verified as correct so you don't have to worry about mistakes, and in general bindgen is tough to make work across all the platforms that libc works with.

And feel free to throw it all in one PR!

PR #483, opened 2016-12-31
sparc64-linux support

japaric:
This needs to be "cleaned" up to use modules instead of a bunch of `cfg`s ...

Sadly, sparc64 constants are very different from other architectures so cleaning this will result in a bunch of duplication, I think.

While working on this, I was wondering why the constants are not written like this:

``` rust
// linux/mod.rs
const COMMON: ::c_int = 3;

cfg_if! {
    if #[cfg(target_arch = "sparc64")] {
        const FOO: ::c_int = 1;
    } else if #[cfg(any(target_arch = "mips64", target_arch = "x86_64"))] {
        const FOO: ::c_int = 2;
    } else {
        // unsupported/unknown architecture
    }
}
```

I think this might result in less duplicated code. @alexcrichton Has something like that ^ been attempted before?


alexcrichton:
For values that need to be changed for #[cfg]'d, can we just push those down the module hierarchy? It's fine to have a ton of duplication, that's what libc is for!

alexcrichton:
Looks good to me! Travis has a few failures though?

japaric:
should be ready to go this time

PR #488, opened 2017-01-06
Link libc and libm in Redox [critical]

jackpot51:
This is critical to being able to set up a proper cross compilation setup for Redox. Due to the way libc is vendored, I have to have this change merged in and have liblibc updated in the Rust repository as well before I can have a working buildbot for Redox.


jackpot51:
Thanks @alexcrichton!

PR #489, opened 2017-01-06
Add waitid and related constants and types.

zackw:
[`waitid`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/waitid.html) is a variation on `waitpid` with a marginally more convenient way of reporting the status, and a couple of handy
additional features, such as the ability to peek at an exit status without consuming it.  It's in POSIX.1-2008 and should be available on all supported Unixes.

Along with it come the type `idtype_t` and the constants `WEXITED`, `WSTOPPED`, `WCONTINUED`, and `WNOWAIT`.  The constants were already defined for unix/notbsd platforms.

The patch is currently incomplete: I'm pushing it to get CI to test platforms I don't have.  Todo list is

* [x] Add a definition of `siginfo_t` to all platforms that don't have it.
* [x] Verify that the new constants are consistent for all \*BSD platforms.
* [x] Verify that `idtype_t` is consistent across the board.
* [x] Add `link_name` annotations for `waitid` if/as necessary.

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -21,11 +21,6 @@ pub type tcflag_t = ::c_ulong;
 pub type nl_item = ::c_int;
 pub type id_t = ::c_uint;
 pub type sem_t = ::c_int;
-
-// idtype_t is specified as a C enum:
-// http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_wait.h.html
-// However, FFI doesn't currently know how to ABI-match a C enum
-// (rust#28925, rust#34641).
 pub type idtype_t = ::c_uint;
 
 pub enum timezone {}
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -22,6 +22,12 @@ pub type nl_item = ::c_int;
 pub type id_t = ::c_uint;
 pub type sem_t = ::c_int;
 
+// idtype_t is specified as a C enum:
+// http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_wait.h.html
+// However, FFI doesn't currently know how to ABI-match a C enum
+// (rust#28925, rust#34641).
+pub type idtype_t = ::c_uint;
+
 pub enum timezone {}
 
 s! {
@@ -1368,6 +1374,10 @@ pub const WSTOPPED: ::c_int = 0x00000008;
 pub const WCONTINUED: ::c_int = 0x00000010;
 pub const WNOWAIT: ::c_int = 0x00000020;
 
+pub const P_ALL: idtype_t = 0;
+pub const P_PID: idtype_t = 1;
+pub const P_PGID: idtype_t = 2;
+
 f! {
     pub fn WSTOPSIG(status: ::c_int) -> ::c_int {
         status >> 8
@@ -1537,6 +1547,12 @@ extern {
                    flags: ::c_int) -> ::c_int;
 
     pub fn initgroups(user: *const ::c_char, basegroup: ::c_int) -> ::c_int;
+
+    #[cfg_attr(all(target_os = "macos", target_arch = "x86"),
+               link_name = "waitid$UNIX2003")]
+    pub fn waitid(idtype: idtype_t, id: id_t, infop: *mut ::siginfo_t,
+                  options: ::c_int) -> ::c_int;
+
 }
 
 cfg_if! {
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -1363,6 +1363,11 @@ pub const LIO_READ: ::c_int = 1;
 pub const LIO_WAIT: ::c_int = 2;
 pub const LIO_NOWAIT: ::c_int = 1;
 
+pub const WEXITED: ::c_int = 0x00000004;
+pub const WSTOPPED: ::c_int = 0x00000008;
+pub const WCONTINUED: ::c_int = 0x00000010;
+pub const WNOWAIT: ::c_int = 0x00000020;
+
 f! {
     pub fn WSTOPSIG(status: ::c_int) -> ::c_int {
         status >> 8
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -4,11 +4,6 @@ pub type dev_t = u64;
 pub type blksize_t = ::int32_t;
 pub type fsblkcnt_t = ::uint64_t;
 pub type fsfilcnt_t = ::uint64_t;
-
-// idtype_t is specified as a C enum:
-// http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_wait.h.html
-// However, FFI doesn't currently know how to ABI-match a C enum
-// (rust#28925, rust#34641).
 pub type idtype_t = ::c_int;
 
 s! {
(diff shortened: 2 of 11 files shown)


alexcrichton:
Nice! We tend to avoid literal `enum` types so could that be expanded to a list of constants?

zackw:
@alexcrichton The trouble is that `idtype_t` is actually specified as an `enum` on the C side.  Is it even possible to express that any other way?  (See http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_wait.h.html )

alexcrichton:
Sure yeah, but C enums are different than Rust enums, they're extensible.

src/unix/mod.rs
```diff
@@ -447,6 +458,11 @@ extern {
                link_name = "waitpid$UNIX2003")]
     pub fn waitpid(pid: pid_t, status: *mut ::c_int, options: ::c_int)
                    -> pid_t;
+    #[cfg(not(target_os = "openbsd"))] // " if " -- appease style checker
```
alexcrichton: Could the definition here be pushed down into the various modules to avoid the `#[cfg]`?

src/unix/mod.rs
```diff
@@ -16,6 +16,13 @@ pub type cc_t = ::c_uchar;
 pub enum DIR {}
 pub enum locale_t {}
 
+// FIXME: This is technically wrong; idtype_t is specified as a C enum.
+// [ http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_wait.h.html ]
+// However, FFI doesn't currently know how to ABI-match a C enum
+// (rust#28925, rust#34641) and *probably* the underlying type will be
+// c_uint everywhere since all of the enumerators are representable by c_uint.
```
alexcrichton: I think it's ok to remove this FIXME, I don't think it's possible to fix such a fixme.

zackw: In the new rev, it still says this, but not as a FIXME.  I think it's valuable to document that we are manually matching a C enum here.  Especially as, it turns out, the underlying type differs on Android.

src/unix/mod.rs
zackw: Can we defer that conversation until after I have something that _works_ everywhere, please?  I don't think the right tradeoff among tidiness, consistency with other stuff, and minimization of repetition will be apparent till then.

src/unix/mod.rs
```diff
@@ -16,6 +16,20 @@ pub type cc_t = ::c_uchar;
...
+// http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_wait.h.html
+// However, FFI doesn't currently know how to ABI-match a C enum
+// (rust#28925, rust#34641).
+cfg_if! {
+    if #[cfg(target_os = "openbsd")] {
+        // idtype_t is not available
+    } else if #[cfg(target_os = "android")] {
+        pub type idtype_t = ::c_int;
+    } else {
+        pub type idtype_t = ::c_uint;
```
alexcrichton: Instead of a `cfg_if!` block here, could the definition of this just get pushed down into respective modules?

src/unix/mod.rs
```diff
@@ -204,6 +218,22 @@ pub const PRIO_MIN: ::c_int = -20;
 pub const PRIO_MAX: ::c_int = 20;
 
 cfg_if! {
+    if #[cfg(target_os = "openbsd")] {
+        // P_* constants are not available
+    } else if #[cfg(target_os = "freebsd")] {
```
alexcrichton: Similar to above, could this `cfg_if!` be avoided to push the definitions into lower modules? (duplication is fine)

alexcrichton:
@zackw oh if you're writing for a green bill of health on Travis before worrying about organization no worries! Feel free to just ping me when it's all working and I'll come take a look at the style.

zackw:
@alexcrichton This is ready to be reviewed now.  I did wind up pushing stuff down into submodules - the cross-platform variance is bigger than I thought it was.  You probably want to look only at the diff versus master; let me know if you want a squashed version.

There's still a weird problem with NetBSD (which is why the actual function prototype is commented out in that file), and I'm a little surprised _not_ to have had to add any definitions of `siginfo_t`.

src/unix/bsd/apple/mod.rs
```diff
@@ -22,6 +22,12 @@ pub type nl_item = ::c_int;
 pub type id_t = ::c_uint;
 pub type sem_t = ::c_int;
 
+// idtype_t is specified as a C enum:
+// http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_wait.h.html
+// However, FFI doesn't currently know how to ABI-match a C enum
+// (rust#28925, rust#34641).
```
alexcrichton: It's ok to avoid repeating this comment in all locations, we tend to not have a lot of comments in libc due to the large amount of duplication.

src/unix/bsd/netbsdlike/netbsd/mod.rs
```diff
@@ -658,6 +673,11 @@ extern {
     pub fn newlocale(mask: ::c_int,
                      locale: *const ::c_char,
                      base: ::locale_t) -> ::locale_t;
+
+    // This should work, but it causes the netbsd CI build to fail with an
+    // intra-libc.a undefined reference to `wait6`.
+    //pub fn waitid(idtype: idtype_t, id: id_t, infop: *mut ::siginfo_t,
+    //              options: ::c_int) -> ::c_int;
```
alexcrichton: This may just need a `#[link_name]` of some form, but otherwise it's ok to just delete this for now.

src/unix/notbsd/mod.rs
```diff
@@ -9,6 +9,18 @@ pub type clockid_t = ::c_int;
 pub type key_t = ::c_int;
 pub type id_t = ::c_uint;
 
+// idtype_t is specified as a C enum:
+// http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_wait.h.html
+// However, FFI doesn't currently know how to ABI-match a C enum
+// (rust#28925, rust#34641).
+cfg_if! {
+    if #[cfg(target_os = "android")] {
+        pub type idtype_t = ::c_int;
```
alexcrichton: Can this be pushed into the two lower modules as well to avoid the `cfg_if!`?

src/unix/bsd/apple/mod.rs
zackw: Hm, the trouble is that there isn't an obvious "top" or "first" place to put it.  I think I'll leave it in bsd/apple/mod.rs, which is alphabetically first, and remove it from the others; does that sound good?

src/unix/bsd/netbsdlike/netbsd/mod.rs
zackw: Yeah, OK.  To be clear, though, I'm pretty sure this is a bug within (possibly just the rumprun version of) NetBSD libc.  The actual error is

    /usr/local/rumprun-x86_64/lib/libc.a(waitid.o): In function `_waitid':
    /build/rumprun/src-netbsd/lib/libc/gen/waitid.c:52: undefined reference to `wait6'

I don't think there's any way to work around that from within our code.

src/unix/notbsd/mod.rs
zackw: Sure.

src/unix/bsd/apple/mod.rs
alexcrichton: Oh I think it's safe to just elide the comment in general. All the relevant properties of the typedef are already checked against the C representation, so it's basically never going to change, so it's ok to omit the comment.

zackw: OK, done.

zackw:
So I think this is fully baked now; do you want it rebased and squashed?

alexcrichton:
@bors: r+

Thanks!

alexcrichton:
Nah it's ok, we can go ahead and merge if CI passes

alexcrichton:
@bors: retry

* network errors

PR #491, opened 2017-01-06
Add getpeereid function

Roguelazer:
This is a wrapper around getsockopt() for getting the uid/gid of a remote Unix domain socket peer. It was added in FreeBSD 4.6 and present in all modern BSDs I checked (including Mac OS X).


alexcrichton:
@bors: r+

Thanks!

PR #493, opened 2017-01-14
Added defines from ttycom.h for dragonfly and freebsd.

johalun:
I wasn't sure about the convention but looking at other files it seems that whatever is over 0x80000000 is c_ulong so I went with that.


alexcrichton:
@bors: r+

Thanks!

johalun:
Great!  Anyone know when this will be in libc crate?

alexcrichton:
Oh we can basically release whenever, would you like a release to be made?

johalun:
That would be great, but please do it after my other PR which contains some constants for dfly that I  overlooked.

nanpuyue:
The type of some constants seems to be c_ulong instead of c_uint:
https://gitlab.redox-os.org/redox-os/termion/-/merge_requests/152

I got the same error on freebsd:
img

@johalun

nanpuyue:
```
  --> src/term.rs:16:22
   |
16 |         if ioctl(fd, TIOCGWINSZ, winsize.as_mut_ptr()) != 0 {
   |                      ^^^^^^^^^^ expected `u64`, found `u32`
   |
help: you can convert a `u32` to a `u64`
   |
16 |         if ioctl(fd, TIOCGWINSZ.into(), winsize.as_mut_ptr()) != 0 {
   |                                +++++++

For more information about this error, try `rustc --explain E0308`.
error: could not compile `trsh` due to previous error
```

PR #498, opened 2017-01-16
Run ios tests

learnopengles:
This WIP PR runs the tests on the iOS simulator. I've tested it locally using macOS Sierra and XCode 8.2.1. I get this output:

RUNNING ALL TESTS
PASSED 6756 tests

The python script probably needs to be customized for the specific failure output from this test runner (update: newest commit should have this).


alexcrichton:
Awesome, thanks!

Historically most of our python scripts for CI inevitably get rewritten in Rust as we maintain them over time. Perhaps we could preempt that and add this as Rust initially?

Looks like there may also be errors on Travis: https://travis-ci.org/rust-lang/libc/jobs/192426280

learnopengles:
Hi Alex,

Rust would be OK too, but I was planning to reuse this code for the rust compiler tests where we might not have a working compiler? Not sure what you prefer here. Maybe python for now and then a conversion later?

Hmm looks like the Travis Env doesn't have the same list of available simulators. I'll try to see how to reproduce this locally so I can fix it.

alexcrichton:
@learnopengles looks like that did the trick for Travis! (travis is green).

And yeah I figured that if it's in Rust it'll be easier to share with the main repo anyway b/c we're trying to jettison all the python scripts there as well.

learnopengles:
Ok! So what would be a good way to integrate that in Rust? And would that be in this PR or a new one?

alexcrichton:
Oh this is just for the libc crate, not for the Rust repo itself. Let's land this here and then the next step would be to apply similar logic to the rust-lang/rust repo to the test suites.

alexcrichton:
Yeah if the logic were in a crate that'd be ideal for sharing between all the projects I think.

Would you be ok translating the script here to Rust before landing?

learnopengles:
Hmm, in that case what form should it be in? I'm not really sure what the *right* way would be here. Have it be a cargo install command and then cargo run X? This probably wouldn't be a quick PR for me in any case but I'm OK if you want to keep it open until then.

alexcrichton:
Oh here we'd just have a script and do `rustc foo.rs && ./foo` or something like that. We could later extract it to a cargo project and such (w/ libs) later

learnopengles:
Ah alright, that sounds like something with a much quicker turnaround :)

learnopengles:
Hi Alex, here is the new version in Rust. Hope it's ok! :)

alexcrichton:
@bors: r+

Thanks!

Issue #500, opened 2017-01-25
support pipe2 on BSD


oconnor663:
Right now it's only available in the `notbsd` subtree, but it looks like the BSD's support it?

https://www.freebsd.org/cgi/man.cgi?query=pipe&sektion=2
http://netbsd.gw.com/cgi-bin/man-cgi?pipe2+2+NetBSD-current
http://man.openbsd.org/pipe.2

I might try moving it up to `src/unix/mod.rs` in a PR just to see what happens. Maybe Apple doesn't support it? Let me know if there's an easy way to figure this out without just seeing what breaks :)

oconnor663:
Yep, it breaks on OSX. Would the right thing to do be to individually add it to the other BSD flavors?

oconnor663:
Related question: `nix` currently hardcodes the platforms that support this call, https://github.com/nix-rust/nix/blob/master/src/unistd.rs#L617-L619. If we add more platforms, `nix` will want to update those lines. Is there a way for `nix` to avoid this kind of hardcoding?

alexcrichton:
AFAIK there's no way to avoid hardcoding down the road, but yeah you can just add it to the BSD modules

oconnor663:
Closing since #502 has landed.

PR #501, opened 2017-01-25
try moving pipe2 from notbsd to unix

oconnor663:
https://github.com/rust-lang/libc/issues/500

This is mostly to see what breaks in CI.


oconnor663:
Ok yeah, this breaks on OSX.

oconnor663:
Here's the failing build: https://travis-ci.org/rust-lang/libc/builds/195223765

PR #503, opened 2017-01-29
Set the type for S_ISUID, S_ISGID and S_ISVTX correctly

TethysSvensson:


alexcrichton:
Thanks for the PR! Unfortunately this runs a relatively high risk of being a breaking change, so we may not be able to accept this until the next major version :(

Idolf:
That's fair.

For the record, this is related to a nix cleanup I am working. I would be surprised if anybody except nix depends on those constants, but I guess it is better to be safe than sorry.

How far are you from the next major version?

alexcrichton:
We don't currently have a timeline for the next major release, unfortunately, but it's likely to happen at some point at least.

gnzlbg:
@TethysSvensson could you rebase this? We can give this a try and see if it breaks anybody.

TethysSvensson:
@gnzlbg Done!

PR #506, opened 2017-01-31
Fix the sign of the EVFILT_* defines

TethysSvensson:
The sign of the defines are not actually consistent with where they are used.

They are used in the `kevent` struct, which defines the `filter` field as `uint32_t`. This is also consistent with the [netbsd header files](http://ftp.tku.edu.tw/NetBSD/NetBSD-current/src/sys/sys/event.h), where the field is a `uint32_t` and the constants are defined as e.g.:

```c
#define	EVFILT_READ		0U
```


Idolf:
Note that this only affects netbsd.

alexcrichton:
Looks good to me, but unfortunately like https://github.com/rust-lang/libc/pull/503 we'll probably have to wait until the next major version :(

gnzlbg:
@TethysSvensson I think this can be merged without issues. Could you rebase this on top of master ?

TethysSvensson:
This was already fixed independently in a0e683ef9.

TethysSvensson:
As far as I can tell, this was the only change included in #787 -- so I guess no breakage was found?

PR #511, opened 2017-02-03
Fix android SA_* constants

malbarbo:
Trying to crosscompile wait-timeout to armv7-linux-androideabi I found that SA_* had different types in libc, but they have to be the same because they are used in bitwise operations.

### DIFF
--- a/src/unix/notbsd/android/b32/mod.rs
+++ b/src/unix/notbsd/android/b32/mod.rs
@@ -161,6 +161,15 @@ s! {
     }
 }
 
+// These constants must be of the same type of sigaction.sa_flags
+pub const SA_NOCLDSTOP: ::c_ulong = 0x00000001;
+pub const SA_NOCLDWAIT: ::c_ulong = 0x00000002;
+pub const SA_NODEFER: ::c_ulong = 0x40000000;
+pub const SA_ONSTACK: ::c_ulong = 0x08000000;
+pub const SA_RESETHAND: ::c_ulong = 0x80000000;
+pub const SA_RESTART: ::c_ulong = 0x10000000;
+pub const SA_SIGINFO: ::c_ulong = 0x00000004;
+
 pub const RTLD_GLOBAL: ::c_int = 2;
 pub const RTLD_NOW: ::c_int = 0;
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks for the PR! Unfortunately this'll be a breaking change to libc so we'll have to hold off on this for now, but otherwise we can be sure to merge on the next bump!

kamalmarhubi:
@alexcrichton if you have time, could you add a label for all these fixes-that-are-breaking PRs? :-)

alexcrichton:
Sure yeah

malbarbo:
@alexcrichton It this really a breaking change? Can we consider it a bug fix instead? Some others PR with type changes was included in 0.2.x releases, for example https://github.com/rust-lang/libc/pull/516, https://github.com/rust-lang/libc/pull/523, https://github.com/rust-lang/libc/pull/533.

alexcrichton:
@malbarbo I think #489 didn't have any deletions, so was that a mistake? All the others I believe the breakage was contained to Solaris, a tier 3 platform. Android, however, is tier 2 :(

malbarbo:
@alexcrichton  #489 was my bad. Thank you for the clarifications.

roblabla:
So I just looked at sigaction from the android-24 platform in the NDK, and it seems to me that this PR is wrong. In the NDK, both arch-arm and arch-arm64 have `sa_flag` from `sigaction` set to `unsigned long`. You can see for yourself : 

- https://android.googlesource.com/toolchain/prebuilts/ndk/r13/+/master/platforms/android-24/arch-arm/usr/include/asm/signal.h#97
- https://android.googlesource.com/toolchain/prebuilts/ndk/r13/+/master/platforms/android-24/arch-arm64/usr/include/asm-generic/signal.h#101

So, as far as I'm aware, this should really be defined as `c_ulong` directly under `android/`. Am I missing something ?

gnzlbg:
@malbarbo can you rebase this ?

We can give this a try to see if anybody is affected by these changes, and if they aren't then just do them.

malbarbo:
Done.

PR #516, opened 2017-02-08
Correct solaris libc definitions:

### LINKED ISSUE #515: some libc solaris definitions are not correct
There are a few somewhat subtle errors in the definitions found in src/unix/solaris/mod.rs.

1. `pthread_t` should be a `c_uint`, not a `uintptr_t` just as `pthread_key_t` is defined
2. `fd_set` is actually a `long` on Solaris, so needs to change definition based on target_pointer_width; not only is this only working by chance right now, it prevents process spawning from working correctly on sparc since the 32-bit value is placed in the lower half of the 64-bit value passed to select/pselect
3. for lp64, FD_SETSIZE should actually be 65536 as has been true since 1997

I will send a pull request shortly with proposed fixes.  This is the first time I've sent a pull request, so please let me know if I've missed any steps (I've tried to follow the instructions at https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#pull-requests).

/cc @dhduvall

binarycrusader:
* pthread_t is defined as uint_t, so must be c_uint, not uintptr_t, just
  as pthread_key_t is already defined
* fd_set is defined as long, so must be i32/i64 based on
  target_pointer_width; this also fixes an indirect endianness issue
  encountered on sparc
* FD_SETSIZE should be defined as 65536 when target_pointer_width = 64

Fixes #515

### DIFF
--- a/src/unix/solaris/mod.rs
+++ b/src/unix/solaris/mod.rs
@@ -1,3 +1,5 @@
+use dox::mem;
+
 pub type c_char = i8;
 pub type c_long = i64;
 pub type c_ulong = u64;
@@ -27,7 +29,7 @@ pub type off_t = i64;
 pub type useconds_t = ::c_uint;
 pub type socklen_t = u32;
 pub type sa_family_t = u8;
-pub type pthread_t = ::uintptr_t;
+pub type pthread_t = ::c_uint;
 pub type pthread_key_t = ::c_uint;
 pub type blksize_t = u32;
 pub type fflags_t = u32;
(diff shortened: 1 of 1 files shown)


alexcrichton:
@bors: r+

Thanks!

binarycrusader:
I don't understand the test failures; I ran `python x.py test src/tools/tidy` from the top-level rust.git clone and it doesn't show libc in the output or any new messages that I can see.

alexcrichton:
Oh the tidy script in this repo is different, it's `ci/style.rs` instead of `./x.py` (which doesn't exist in this repo)

binarycrusader:
This is a trap for new contributors; any chance of either updating the [contributing document](https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#pull-requests) so that it points out that submodules have different contribution guidelines?

Also, it seems unfortunate that the same rules for the main rust repo don't apply to submodules; a unified contribution process would simplify things for new contributors.

I will try style.rs and see what's wrong, thanks.

alexcrichton:
@bors: r+

Oh right yeah the contributing document is intended for rust-lang/rust and basically just got copied over here unfotunately :(

PR #521, opened 2017-02-12
WIFCONTINUED function on Linux and FreeBSD

radupopescu:
* Adds the missing WIFCONTINUED function to Linux (in `src/unix/notbsd/mod.rs`).
* Modifies the signature of WIFCONTINUED on FreeBSD (`src/unix/bsd/freebsdlike/mod.rs`) to be consistent with the ones on MacOS and Linux.


alexcrichton:
@bors: r+

Thanks!

PR #524, opened 2017-02-16
Added *_setpshared and *_getpshared bindings

mersinvald:
Adding bindings to posix pthreads functions vital for IPC via shared memory.

That's my first PR into libc and I'm not proficient in unix systems programming, so I would really appreciate if someone reviewed it in case I missed something regardrless of the passed tests.
Thanks!


alexcrichton:
@bors: r+

Thanks!

PR #525, opened 2017-02-17
Add EPOLLEXCLUSIVE that is added in Linux 4.5.

monadbobo:
Add [EPOLLEXCLUSIVE](https://kernelnewbies.org/Linux_4.5#head-64f3b13b9026133a232a418a27ac029e21fff2ba) that is added in Linux 4.5.


alexcrichton:
Thanks! Looks like there's CI failures though?

monadbobo:
@alexcrichton  This is my mistake, the flag only in linux, so i will send the new pr for moving this flag to linux/mod.rs. 
Thanks.

PR #527, opened 2017-02-18
add tmpnam and pthread_exit

philippkeller:
tmpnam and readdir are trivial IMO.

About the `pthread_create` change: It needs `unsafe` for passing `C` functions to pthread_create (with rust functions the omission of `unsafe` is working of course).

`bindgen` produces this function definition:

```
pub fn pthread_create(arg1: *mut pthread_t,
                      arg2: *const pthread_attr_t,
                      arg3: Option<unsafe extern "C" fn(arg1: *mut c_void) -> *mut c_void>,
                      arg4: *mut c_void) -> c_int;
```

So it would add an additional `Option` around the function. But that would break existing code which uses `libc::pthread_create` and what use is it to call pthread_create without any function pointer, so I left `Option` out.

For reference: I also opened a [stackoverflow question](http://stackoverflow.com/questions/42284562) where the answers were also suggesting adding `unsafe` to the function definition.

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -549,6 +549,7 @@ extern {
                link_name = "pthread_join$UNIX2003")]
     pub fn pthread_join(native: ::pthread_t,
                         value: *mut *mut ::c_void) -> ::c_int;
+    pub fn pthread_exit(value: *mut c_void);
     pub fn pthread_attr_init(attr: *mut ::pthread_attr_t) -> ::c_int;
     pub fn pthread_attr_destroy(attr: *mut ::pthread_attr_t) -> ::c_int;
     pub fn pthread_attr_setstacksize(attr: *mut ::pthread_attr_t,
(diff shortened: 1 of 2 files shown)


philippkeller:
@alexcrichton the build failed for no apparent reason (iOS build said `No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`). What could I do here? Can I retrigger the build somehow?

alexcrichton:
Thanks for the PR! Unfortunately this is a breaking change though, so could the change to the function pointer be backed out?

philippkeller:
@alexcrichton Can you shed some light on what is breaking? 

I left out the `Option` so everything which is breaking is the change from `fn` to `unsafe fn`. I tested with a safe fn and this still works, so I assume this is non-breaking.

alexcrichton:
@philippkeller I believe that was added in a very recent version of Rust (coercion from safe to unsafe) but historical versions of Rust didn't support such a feature.

philippkeller:
I see, it only got [introduced lately](https://github.com/rust-lang/rfcs/issues/1762), I think it is only available as from rustc 1.15

Issue #528, opened 2017-02-19
Add baudrate constants


berkowski:
The only portable way to set baudrates in `termios` structs is through the `B*` constants defined for each platform.  Given the use of `ctest` in `libc` it was suggested in nix-rust/nix#518 that these constants might be better defined `libc` itself, with `nix` providing access through a type-safe enum.

Seems straight forward (prior art: https://github.com/dcuddeback/termios-rs) and fits the [scope](https://github.com/rust-lang/rfcs/blob/master/text/1291-promote-libc.md).

PR #530, opened 2017-02-20
Added baudrate constants.

berkowski:
Addresses #528, adding baudrate constants for `bsd` and `notbsd` flavors of `unix`.

Passes `libc-test` locally on ubuntu 14.04 LTS with the additional (uncommitted) entries into in `build.rs`:

```
cfg.skip_const( move |name| {
        match name {
            # ... snip ...
            "PTRACE_O_SUSPEND_SECCOMP" |
            "CLONE_NEWCGROUP" |
            "NETLINK_LIST_MEMBERSHIPS" |
            "NETLINK_LISTEN_ALL_NSID" |
            "NETLINK_CAP_ACK" |
            "PR_CAP_AMBIENT_CLEAR_ALL" |
            "PR_CAP_AMBIENT_LOWER" |
            "PR_CAP_AMBIENT_RAISE" |
            "PR_CAP_AMBIENT_IS_SET" |
            "PR_CAP_AMBIENT" |
            "PR_FP_MODE_FRE" |
            "PR_FP_MODE_FR" |
            "PR_GET_FP_MODE" |
            "PR_SET_FP_MODE" |
            "PR_MPX_DISABLE_MANAGEMENT" |
            "PR_MPX_ENABLE_MANAGEMENT" |
            "PR_GET_THP_DISABLE" |
            "PR_SET_THP_DISABLE" |
            "PR_SET_MM_MAP_SIZE" |
            "PR_GET_MM_MAP_SIZE" |
            "PR_SET_MM_MAP" |
            "NLM_F_DUMP_FILTERED" |
            "EPOLLEXCLUSIVE" => true,

            _ => false,
        }
    });
```
I'm assuming this is because I'm stuck using `linux-libc-dev:3.13.0-24.46` for the moment and those constants are defined in newer versions.


berkowski:
The ppc-linux failures look like an endian issue to me.  I know the ppc is big-endian but I'm not sure why the baud constants seem to remain little-endian for rust in the test.  Only immediate difference I see is they're defined as octal literals, not hex or dec.

OpenBSD seems to be reporting the same baud constants as linux?  Is this correct?  I based my constants off [termios-rs](https://github.com/dcuddeback/termios-rs/blob/master/src/os/openbsd.rs) and [ftp://ftp.usa.openbsd.org/pub/OpenBSD/src/sys/sys/termios.h](ftp://ftp.usa.openbsd.org/pub/OpenBSD/src/sys/sys/termios.h)

semarie:
about OpenBSD, it is due to `rpcsvc/rex.h` that define the same names (but with different values). The header is generated from [src/lib/librpcsvc/rex.x](http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/lib/librpcsvc/rex.x?rev=1.5&content-type=text/x-cvsweb-markup).

@alexcrichton any way to deal with this situation ? Ideally `build.rs` should `#undef` constants from  `rpcsvc/rex.h` that aren't used and conflicts with `termios.h`. Is it possible ?

semarie:
@berkowski could you try to add

```rust
cfg.header("termios.h");
```

inside OpenBSD block (but **after** `cfg.header("rpcsvc/rex.h");`) ? it should be enough to redefine values with termios ones.

berkowski:
Pulled in the `cfg.skip_const` changes by accident.  Re-pushed with only the header addition.  Should still fail though.

berkowski:
Ok, so at this point we seem to have workable solutions except for:

- OpenBSD:  Conflicting `#define`s between `termios.h` and `rpcsvc/rex.h`
- PowerPC:  Doesn't seem to have `BOTHER` constant; can just remove.
- sparc64: Not testable.  Could blindly implement defines from linux source.  Probably better to just leave unimpl'd until testable.

alexcrichton:
Looks great to me! I'm fine r+'ing once CI passes

berkowski:
Traveling over the next couple of days - I'll peck at the issues when I can

kamalmarhubi:
The powerpc errors are confusing:
```
bad CBAUDEX value at byte 3: rust: 0 (0x0) != c 16 (0x10)
```
while
```
#define CBAUDEX 0000000
```
—https://github.com/torvalds/linux/blob/v4.9/arch/powerpc/include/uapi/asm/termbits.h#L139

kamalmarhubi:
The openbsd errors are also confusing:
```
bad B50 value at byte 0: rust: 50 (0x32) != c 1 (0x1)
bad B75 value at byte 0: rust: 75 (0x4b) != c 2 (0x2)
```
while
```
#define B50	50
#define B75	75
```
—https://github.com/openbsd/src/blob/5271000b44abe23907b73bbb3aa38ddf4a0bce08/sys/sys/termios.h#L213-L214

(50 is 0x32 and 75 is 0x4b)

semarie:
@kamalmarhubi for openbsd the reason is values from `termios.h` are overrided later by `rpcsvc/rex.h` inclusion. (and due to `#ifdef _SYS_TERMIOS_H_` / `#define _SYS_TERMIOS_H_` the reinclusion didn't reset values).

for powerpc, it could be the same problem (I don't check source code deeper).

berkowski:
@semarie:  I tried including `termios.h` after `rpcsvc/rex.h` in `build.rs` in  e393a2d but it didn't seem to do the trick.  Did I add the include in the correct place you intended?

The PPC stuff is interesting.  I found https://lists.ozlabs.org/pipermail/linuxppc-dev/1999-January/000634.html which looks similar.  Maybe this is an issue with `qemu-ppc` using old header files?

semarie:
@berkowski yes it is where I intended but it isn't enough (I didn't think about that before).

- `termios.h` is included - at [line 100](https://github.com/rust-lang/libc/blob/e393a2d4a09ac7c3792c2545f49b16e1b6983021/libc-test/build.rs#L100).
- `#define` baudrate values are defined, and `_SYS_TERMIOS_H_` header guard is **also** defined
- when `rpcsvc/rex.h` is included, it redefines baudrate values
- when `termios.h` is included a second time, due to `_SYS_TERMIOS_H_` header guard, baudrate values aren't redefined.

berkowski:
@semiarid Ah, ok.  In that case I'll revert the change to build.rs in a new commit.

semarie:
it seems a bit odd that the test still pass without `rpcsvc/rex.h`. Something should have changed and the header isn't required anymore. But it is a good thing for this PR.

berkowski:
I couldn't find anything in `bsd/netbsdlike/openbsdlike/openbsd.rs` that referenced the "rex" specific identifiers so I gave it a shot.  Maybe it got pulled in by accident to support some of the termios related constants.

Anyway, seems to be good to go now.

PR #531, opened 2017-02-21
Add jail functions and constants for FreeBSD

freqlabs:
These are the system calls for creating and managing jails on FreeBSD.


alexcrichton:
@bors: r+

Thanks!

PR #532, opened 2017-02-21
Add definitions to Android to allow fs2 crate compile

malbarbo:
This is the only missing part to allow cargo to compile on android.


alexcrichton:
@bors: r+

Thanks!

PR #533, opened 2017-02-21
Make readdir available on all unix targets

raphlinus:
The readdir_r call has problems, and we'll probably want to move to
readdir on many, if not most, unix targets. This patch makes readdir
available in unix, rather than just solaris as before.

See https://github.com/rust-lang/rust/issues/40021


alexcrichton:
Looks like there may be some CI failures?

raphlinus:
Yeah, let's see if this fixes it. It's possible some will need const (apparently that worked ok with Solaris) and some mut.

PR #534, opened 2017-02-22
epoll: Add EPOLLEXCLUSIVE

kamalmarhubi:
This flag was added in Linux 4.5.


alexcrichton:
Looks like CI may be failing?

kamalmarhubi:
Ah looks like it was added under `linux` in #526.

PR #535, opened 2017-02-23
epoll: Change type of EPOLL flags to uint32_t

kamalmarhubi:
These flags are used the `events` field of `struct epoll_event`, which
has that type.

    pub struct epoll_event {
        pub events: ::uint32_t,
        pub u64: ::uint64_t,
    }


alexcrichton:
Thanks for the PR! This is a breaking change for now so we can't land it just yet, but I've tagged it as such.

kamalmarhubi:
@alexcrichton as a "maybe" for managing 1.0-breakage stuff: close the PRs but leave the label in place? I think it's nicer for contributors and the project not to have open PRs that we don't expect to land in any reasonable time frame.

PR #537, opened 2017-02-23
Add ppoll() for all unix platforms

Susurrus:
I'm unsure of whether there is support in OS X for this, and I can't find anything online (so I'm betting there isn't), but I'm going to let this run through CI to confirm.

### DIFF
--- a/src/unix/bsd/freebsdlike/mod.rs
+++ b/src/unix/bsd/freebsdlike/mod.rs
@@ -922,6 +922,10 @@ extern {
     pub fn pthread_mutex_timedlock(lock: *mut pthread_mutex_t,
                                    abstime: *const ::timespec) -> ::c_int;
     pub fn pipe2(fds: *mut ::c_int, flags: ::c_int) -> ::c_int;
+    pub fn ppoll(fds: *mut ::pollfd,
+                 nfds: nfds_t,
+                 timeout: *const ::timespec,
+                 sigmask: *const sigset_t) -> ::c_int;
 }
 
 cfg_if! {
--- a/src/unix/bsd/freebsdlike/mod.rs
+++ b/src/unix/bsd/freebsdlike/mod.rs
@@ -923,7 +923,7 @@ extern {
                                    abstime: *const ::timespec) -> ::c_int;
     pub fn pipe2(fds: *mut ::c_int, flags: ::c_int) -> ::c_int;
     pub fn ppoll(fds: *mut ::pollfd,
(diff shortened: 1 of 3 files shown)


alexcrichton:
@bors: r+

Thanks!

Susurrus:
I have a local fix for the styling, so I already know about that, but I also ran things through my local test suite and found some errors I'm not understanding:
```
cargo:warning=/home/susurrus/Projects/libc/target/debug/build/libc-test-26b349ee79642880/out/all.c: In function ‘__test_fn_ppoll’:
cargo:warning=/home/susurrus/Projects/libc/target/debug/build/libc-test-26b349ee79642880/out/all.c:2830:24: error: return from incompatible pointer type [-Werror=incompatible-pointer-types]
cargo:warning=                 return ppoll;
cargo:warning=                        ^~~~~
cargo:warning=cc1: all warnings being treated as errors
```

But if I look at that file that function looks correct:
```
int (*_test_fn_ppoll(void))(struct pollfd*, nfds_t, int, const sigset_t*) {
    return ppoll;
}
```

I thought maybe it was because the `sigset_t` struct wasn't getting a `struct` keyword, but `sigwait` doesn't seem to have that problem.

alexcrichton:
Let's see what CI says, it may just be a const/mut or bit width thing

Susurrus:
That same problem is occurring on Travis, so there's definitely something wrong which I can reproduce locally.

I also pushed a style fix, but it looks like that retriggered CI, so maybe I shouldn't have done that yet...

Susurrus:
Alright, I think I got it. Had the wrong type for `timeout`.

Susurrus:
So turns out this doesn't exist on Mac or NetBSD, or at least not by that name. There are similar different names for `pselect()` or one of the other functions. But I couldn't find anything by googling, though I did find the name `ppollts()` for NetBSD, but I don't know if that's right. Any suggestions for how to resolve this one way or another for those platforms?

alexcrichton:
It's ok to just leave out this function on those platforms, you can sink the definition farther down the crate hierarchy

Susurrus:
I just pushed it down the stack, but I'm not certain that was correct. OpenBSD and FreeBSD passed their tests, so it's just NetBSD and Mac that don't have it. Should I also expose it in `/src/unix/bsd/freebsdlike/mod.rs`?

Susurrus:
Alright, looks like I finally figured it out! @alexcrichton Wanna give this a look-see?

PR #538, opened 2017-02-24
Add and fix tests for {i686, aarch64}-linux-android targets

### LINKED ISSUE #536: Add tests for others android platforms
The libc-test runs for arm-linux-android. It would be great if it were also run for aarch64-linux-android and i686-linux-android.

malbarbo:
I think that these changes do not breaks compatibility.

There are some types and constants changes to i686 and aarch64, but I see these changes as bug fixes instead of breaking changes. Also the type time64_t was remove from aarch64 because it is not defined in this arch.

Fixes #536


src/unix/mod.rs
```diff
@@ -750,6 +728,56 @@ extern {
...
+    pub fn sigismember(set: *const sigset_t, signum: ::c_int) -> ::c_int;
+
+    #[cfg_attr(target_os = "netbsd", link_name = "__timegm50")]
+    pub fn timegm(tm: *mut ::tm) -> time_t;
+}
+
+// Android has some weirdness in this definition
+// See weirdness in src/unix/notbsd/android/mod.rs
+#[cfg(any(not(target_os = "android"),     // " if " -- appease style checker
+          not(target_arch = "aarch64")))]
```
alexcrichton: I unfortunately regret the block below this, but could we just duplicate these functions in lower modules to avoid this `#[cfg]` attribute?

src/unix/notbsd/android/mod.rs
```diff
@@ -2,22 +2,27 @@
...
-pub type c_char = u8;
+cfg_if! {
+    if #[cfg(target_arch = "x86")] {
+        pub type c_char = i8;
+        pub type wchar_t = i32;
+    } else {
+        pub type c_char = u8;
+        pub type wchar_t = u32;
+    }
+}
```
alexcrichton: Could this be pushed down further?

src/unix/notbsd/android/mod.rs
```diff
@@ -447,12 +452,21 @@ pub const O_EXCL: ::c_int = 128;
...
+    if #[cfg(target_arch = "x86")] {
+        pub const O_DIRECT: ::c_int = 0x4000;
+        pub const O_DIRECTORY: ::c_int = 0x10000;
+        pub const O_NOFOLLOW: ::c_int = 0x20000;
+    } else {
+        pub const O_DIRECT: ::c_int = 0x10000;
+        pub const O_DIRECTORY: ::c_int = 0x4000;
+        pub const O_NOFOLLOW: ::c_int = 0x8000;
+    }
+}
```
alexcrichton: Could this also be pushed down further?

alexcrichton:
Looks great to me, thanks so much for taking the charge on this! Overall it'd be great to reduce the `#[cfg]` introduced here, but other than that looks great to me. (duplication is fine)

malbarbo:
@alexcrichton All cfg_if! removed.

We got some network issues in the tests.

kamalmarhubi:
@malbarbo maybe try `git commit --amend --no-edit && git push -f` to force travis to run again.

malbarbo:
@alexcrichton All green now.

@kamalmarhubi Thanks for the tip. It worked.

src/unix/notbsd/android/mod.rs
```diff
@@ -727,74 +712,6 @@ pub const NLA_TYPE_MASK: ::c_int = !(NLA_F_NESTED | NLA_F_NET_BYTEORDER);
 pub const SIGEV_THREAD_ID: ::c_int = 4;
 
 f! {
-    pub fn sigemptyset(set: *mut sigset_t) -> ::c_int {
```
alexcrichton: It looks like these functions may not have survived a rebase by accident?

malbarbo: These are removed purposely. The old tests used android-9 api, which do not have this functions. The next api version distributed with android ndk is android-12 (release in 2011 https://en.wikipedia.org/wiki/Android_version_history), which define these functions. The definitions used by android are the same of others unix (src/unix/mod.rs).

I can see that this can be a breaking change... Should I return these implementations? (this will require moving definitions out of src/unix/mod.rs)

Keeping our own implementations of these functions seems odd.

alexcrichton: Ah ok, in that case sounds great to me!

Android-9 is such an ancient NDK that it seems fine to not require support for that here. If project want that level of compatibility (e.g. libstd) we can vendor the impls ourselves.

kamalmarhubi:
Failures look related to yesterday's S3 outage:
- docker pull failures
- falied to download qemu image from S3

kamalmarhubi:
@malbarbo if you get to it before @alexcrichton retries, you could rebase and force push. Push will get Travis to run, rebase will allow homu to use the cached Travis success so long as no other PR is merged in the meantime. Hopefully we can get this out today!

alexcrichton:
I can't even see the failures, but I'll take your word for it!

@bors: retry

PR #541, opened 2017-02-26
Add cfmakeraw and cfsetspeed

Susurrus:
This includes implementations for Android. `cfsetspeed` is basically just a back-to-back call to `cfsetispeed` and `cfsetospeed`, both of which seem to do the same thing here, so I just copied that body as well for `cfsetspeed`. The implementation for `cfmakeraw` was taken from the man pages for `termios(3)`.

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -835,8 +835,10 @@ extern {
     pub fn tcdrain(fd: ::c_int) -> ::c_int;
     pub fn cfgetispeed(termios: *const ::termios) -> ::speed_t;
     pub fn cfgetospeed(termios: *const ::termios) -> ::speed_t;
+    pub fn cfmakeraw(termios: *mut ::termios);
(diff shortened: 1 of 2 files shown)


Susurrus:
I think the problem was that the void datatype needed to be explicit in the weird macro in the Android stuff, so hopefully that's fixed.

fiveop:
http://stackoverflow.com/questions/38896155/what-is-the-bitwise-not-operator-in-rust/38896376#38896376

Susurrus:
Ugh! I've done that so many times coming from C and I thought it was compiling on my computer. Weird.

Anyways, I've updated it. so we'll see how it goes now

alexcrichton:
Looks like CI may be failing?

Susurrus:
Indeed. The macro is picky about void functions still being annotated to return `()`. Should be fixed now.

Susurrus:
Indeed, got it all sorted out now.

Issue #540, opened 2017-02-25
Policy on new system calls?


alexander255:
I see that the `accept4` system call is currently exposed on Linux, but not on FreeBSD, OpenBSD or other system that support it. Here's the Unix compatibility story for that, IMHO quite useful, system call:

 - Linux 2.6.28+ (24.12.2008)
 - FreeBSD 10.0+ (01.02.2014)
 - OpenBSD 5.7+ (01.05.2015)
 - DragonFlyBSD 4.4+ (07.12.2015)
 - BitRig ??? (More that 3 years ago according to `git blame` on https://github.com/bitrig/bitrig/blame/master/sys/kern/syscalls.c, OTOH the BitRig 1.0 release page says it incorporates changes up to OpenBSD 5.6?, very unclear status in general IMHO…)
 - NetBSD 8.0+ (Hopefully released soonish…)
 - macOS/Darwin Nope, expected anything else?

Should these simply be submitted for any platform that supports them? Will that cause any issues for applications that target older platform?

Possibly related issues: #234

alexcrichton:
New functions are definitely welcome! Adding them to this crate doesn't imply that they *must* be used, so it's up to downstream crates to use them or not. So in that sense PRs are definitely welcome!

Susurrus:
Can this be closed since #546 was merged?

alexander255:
I guess so… The policy is basically "**add whatever you want, as long as it exists in some released version of that platform**". So the opening question of this issue has indeed been answered in the process of getting the PR merged, yes.

PR #546, opened 2017-03-02
Add support for the `accept4` system call for FreeBSD* & NetBSD*

alexander255:
As outlined in #540.

However since support for this system call seems to be committed, but not yet released in NetBSD & BitRig (is BitRig actually still be developed at all, btw?), I'm not sure whether it wouldn't be a better idea to only add OpenBSD support instead?
What's the official policy on this?


alexcrichton:
It looks like rumprun may not have accept4?

alexander255:
As I said, support for `accept4` in NetBSD und BitRig has been committed, but not yet release in any stable version. Should I move the reference to the OpenBSD submodule instead?

alexcrichton:
Yeah let's just move it down for now.

alexander255:
Done! Also, out of curiosity, is BitRig still a thing or have they stalled after their initial release?

alexcrichton:
I'm not sure myself unfortunately :(

Looks great though!

@bors: r+

PR #550, opened 2017-03-13
add fdopendir

Mic92:


src/unix/mod.rs
```diff
@@ -328,6 +328,10 @@ extern {
     #[cfg_attr(all(target_os = "macos", target_arch = "x86"),
                link_name = "fdopen$UNIX2003")]
     pub fn fdopen(fd: ::c_int, mode: *const c_char) -> *mut ::FILE;
+
+    #[cfg(not(any(target_os = "macos", target_os = "ios")))]
```
Mic92: Should I define this function here or in every os module except macos?

alexcrichton: Ah yeah let's just inline this definition in modules other than `src/unix/bsd/apple`

Mic92: done

alexcrichton:
@bors: r+

Thanks!

Mic92:
passes the test now

PR #564, opened 2017-03-31
Make networking on Solaris/Illumos great again

vojtechkral:


vojtechkral:
There's actually a couple of other problems in that file, but this one is causing breakage at the moment. I'll try to collect the other ones into one larger fix.

PR #565, opened 2017-04-01
Link libc crate against pthread

anatol:
pthread_* functions are defined in libpthread thus we need to link against
this library.

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -242,6 +242,7 @@ cfg_if! {
         #[link(name = "c")]
         #[link(name = "m")]
         #[link(name = "rt")]
+        #[link(name = "pthread")]
         extern {}
     }
 }


alexcrichton:
Thanks for the PR! Out of curiosity, what platform was this on?

This has a risk of being a breaking change (due to how heavily depended on libc is) so if it's a niche-ish platform then we may be able to have a targeted fix, but if it's a big one then we can try it out.

anatol:
Hi, I was trying to start my rust-only version of `std` similar to what [steed](https://github.com/japaric/steed) is doing. And I hit this problem.

I do not think it is a breaking change. I am looking at rust build files and it links to `pthread` in [this file](https://github.com/rust-lang/rust/blob/master/src/libstd/build.rs). But it would be more correct to move linking dependency to the place where we have `extern`s.

alexcrichton:
Aha excellent point! In that case when we land this in rust-lang/rust we should just be sure to remove pthread from that location

@bors: r+

anatol:
It looks like Apple products do not contain separate `pthread` library.

anatol:
Could you please help me to understand what is the problem with darwin products? When I click to the [test job logs](https://api.travis-ci.org/jobs/218262431/log.txt?deansi=true) I see something like `Sorry, we experienced an error.`

alexcrichton:
@bors: retry

Ah that just looks like a weird travis issue, let's try again

alexcrichton:
@bors: retry

* another travis bug?

PR #568, opened 2017-04-04
Add MINSIGSTKSZ constant to linux sources

anatol:
MINSIGSTKSZ is a minimum size of process stack allowed

### DIFF
--- a/src/unix/notbsd/android/mod.rs
+++ b/src/unix/notbsd/android/mod.rs
@@ -584,6 +584,7 @@ pub const MCL_CURRENT: ::c_int = 0x0001;
 pub const MCL_FUTURE: ::c_int = 0x0002;
 
 pub const SIGSTKSZ: ::size_t = 8192;
+pub const MINSIGSTKSZ: ::size_t = 2048;
 pub const CBAUD: ::tcflag_t = 0o0010017;
 pub const TAB1: ::c_int = 0x00000800;
 pub const TAB2: ::c_int = 0x00001000;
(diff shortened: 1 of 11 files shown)


alexcrichton:
@bors: r+

Thanks!

anatol:
```
nning Install scripts
appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
Error downloading remote file: One or more errors occurred.
Inner Exception: The request was aborted: Could not create SSL/TLS secure channel.
Command exited with code 2
```

Does not look related to my change

anatol:
2 powerpc targets fail with following message

```
RUNNING ALL TESTS
bad MINSIGSTKSZ value at byte 2: rust: 8 (0x8) != c 16 (0x10)
thread 'main' panicked at 'some tests failed', /checkout/target/powerpc-unknown-linux-gnu/debug/build/libc-test-4790929c323c3f02/out/all.rs:11
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

could you please help me to understand what does this error mean?

alexcrichton:
Oh that just means that the value for the constant is different on powerpc than is what's currently defined. You'll want to update the constant's value on powerpc to fix the message. Right now it's just saying that the value a byte 2 differed, in Rust is was defined as 0x08 and in C it's 0x10

anatol:
I can change it to match the system libc. But it is weird it uses a value different from Linux kernel uapi. [The const defined as 2048 there](https://github.com/torvalds/linux/blob/5924bbecd0267d87c24110cbe2041b5075173a25/arch/powerpc/include/uapi/asm/signal.h#L88).

alexcrichton:
@bors: r+

Ah looks like it got sorted out

PR #571, opened 2017-04-13
add dirfd on unix

Mic92:


alexcrichton:
@bors: r+

Thanks!

alexcrichton:
Ah looks like this may not be defined on rumprun?

Mic92:
Interesting, because dirfd already appeared in 4.2BSD in 1983.

Mic92:
How can I exclude rumprun, but include netbsd?

alexcrichton:
I think there should be some examples in the netbsd module as well but the gist is to use target_vendor in the cfg.

Dunno what's up with rumprun to cause this failure...

Susurrus:
@Mic92 Want to try pushing this again and seeing if it was a CI glitch? It'd be great to get this supported so we can support it in nix-rust/nix also.

Mic92:
still the same problem.

Susurrus:
`dirfd` is declared in `dirent.h` in NetBSD. This header file is included by the `build.rs` script in `libc-test`. So I'm not certain what's the problem, though I'm not too familiar with the testing framework used by `libc`. @alexcrichton Any other ideas?

Susurrus:
Could it be because `dirfd` is actually a macro on NetBSD? [Source](http://cvsweb.netbsd.org/bsdweb.cgi/src/include/dirent.h?rev=1.36&content-type=text/x-cvsweb-markup&only_with_tag=MAIN)

alexcrichton:
Want to just duplicate it to each platform and leave it out on NetBSD?

Susurrus:
I think the correct solution here is to just skip the check for this function on NetBSD. There's already precedent for this with many other functions.

Additionally it might be worth pinging some people in the NetBSD IRC and seeing if our understanding here is correct.

Mic92:
In nix we can just inline the netbsd macro.

Susurrus:
I would have suggested inlining that macro within `libc` instead of in a higher-level library as libraries will probably expect it to exist in `libc` on that platform. Would that be rejected if it was submitted in a separate PR?

alexcrichton:
Nah yeah we typically include macros for other platforms for common functionality, so including that for netbsd sounds fine by me!

Susurrus:
@Mic92 Could I convince you to file an additional PR bringing `dirfd` to NetBSD as well? :-)

Mic92:
@Susurrus ^

PR #574, opened 2017-04-19
More constants

Susurrus:
I searched through the codebases for libc and the various platforms to come to these conclusions.


Susurrus:
Looks like the openbsd failed due to an unrelated error with fetching gcc.

alexcrichton:
@bors: r+

Thanks!

PR #576, opened 2017-04-20
Solaris: Add ports support, correct some types, add some errnos

vojtechkral:
This should benefit Illumos too...


alexcrichton:
@bors: r+

Thanks!

Issue #361, opened 2016-08-19
add support for mips-uclibc targets


japaric:
These targets landed in rust-lang/rust#35734 but this crate doesn't properly support them. This
issue tracks adding support for them. Where "adding support" means updating the function signatures
and definition of "constants" (probably via new `cfg` blocks/code) to make the libc-ctest test suite
pass for these targets.

Here are instructions on how to run said test suite:

I've created a docker image (*) with all the stuff needed (e.g. cross toolchain) to run the test
suite. Launch a container with this image using the following command:

```
$ docker run --privileged -it japaric/rust-uclibc:2016-08-17
```

Then, inside docker container run this command:

```
$ curl -sL https://raw.githubusercontent.com/japaric/docker/master/rust-uclibc/libc-ctest.sh | bash -s -- $TARGET
```

(where `$TARGET` is `mips-unknown-linux-uclibc` or `mipsel-unknown-linux-uclibc`).

To cross compile `std` for `$TARGET` and to run the libc-test test suite. The test suite will fail
and then you'll be back to the shell. The `libc` crate will be available in `~/libc`. That `libc`
checkout would have to be modified until the test passes. To run the libc-test test suite again run
these commands:

```
$ cd ~/lib/libc-test
$ cargo run --target $TARGET
```

UPDATE: The image is now [up](https://hub.docker.com/r/japaric/rust-uclibc/tags/). Disregard the note below.

---

(*) NOTE: I'm still uploading the image and it may take a while. In the meantime you can build the image
yourself with these command (it will take a while):

```
$ git clone --depth 1 https://github.com/japaric/docker
$ cd docker
$ docker build -t $IMAGE_NAME:$IMAGE_TAG -f Dockerfile ..
```

cc @felixalias

felixalias:
This is awesome - thanks for doing all of this for a relatively exotic platform!! :100:

japaric:
I forgot to paste the errors I got from running this. For further reference: [gist](https://gist.github.com/japaric/3d8b4acb11e09e2d9e4d2aff8af5e7c5)

alexcrichton:
I believe initial support has been added!

theronic:
@japaric when I run `curl -sL https://raw.githubusercontent.com/japaric/docker/master/rust-uclibc/libc-ctest.sh | bash -s -- mipsel-unknown-linux-uclibc`, it can't find your libc repo:

```
try  + git clone --branch uclibc https://github.com/japaric/libc /home/rust/libc
[...]

PR #578, opened 2017-04-20
Steps towards support for musl-unknown-linux-uclibc

cactorium:
Hello! I've been working towards resolving https://github.com/rust-lang/libc/issues/361 , this PR compiles successfully with a newish compiler (with some minor fixes in `gcc`, `ctest`), and all the tests pass for `libc-ctest`. Basically most of the undefined functions, constants, and structs were just removed from the ctest, and then any constants that weren't correct were fixed. Would it make more sense to conditionally remove them from libc? I wasn't sure when it was appropriate to skip the test for it instead of removing the function/constants, so I just removed all the tests for now because that was a little easier than hunting them down. I'm also guessing the way some of the constants were conditionally set wasn't the correct style, would you guys have any advice on how to do it more correctly? Lemme know how it looks!


libc-test/build.rs
```diff
@@ -387,6 +399,29 @@ fn main() {
...
+            "LC_PAPER_MASK" | "LC_NAME_MASK" | "LC_ADDRESS_MASK" | "LC_TELEPHONE_MASK" | "LC_MEASUREMENT_MASK" | "LC_IDENTIFICATION_MASK" | "LC_ALL_MASK" |
+            "MS_DIRSYNC" | "MS_MOVE" | "MS_REC" | "MS_SILENT" | "MS_POSIXACL" | "MS_UNBINDABLE" | "MS_PRIVATE" | "MS_SLAVE" | "MS_SHARED" | "MS_RELATIME" | "MS_KERNMOUNT" | "MS_I_VERSION" | "MS_STRICTATIME" | "MS_ACTIVE" |
+            "QIF_BLIMITS" | "QIF_SPACE" | "QIF_ILIMITS" | "QIF_INODES" | "QIF_BTIME" | "QIF_ITIME" | "QIF_LIMITS" | "QIF_USAGE" | "QIF_TIMES" | "QIF_ALL" | "IP_TRANSPARENT" |
+            "Q_GETFMT" | "Q_GETINFO" | "Q_SETINFO" | "SHM_EXEC" | "RB_SW_SUSPEND" | "RB_KEXEC" | "RUSAGE_THREAD" |
+            // not entirely sure why these don't work...
+            "LC_CTYPE_MASK" | "LC_NUMERIC_MASK" | "LC_TIME_MASK" | "LC_COLLATE_MASK" | "LC_MONETARY_MASK" | "LC_MESSAGES_MASK" |
+            "MADV_MERGEABLE" | "MADV_UNMERGEABLE" | "MADV_HWPOISON" | "IPV6_ADD_MEMBERSHIP" | "IPV6_DROP_MEMBERSHIP" | "IPV6_MULTICAST_LOOP" | "IPV6_V6ONLY" |
+            "MAP_STACK" | "MAP_HUGETLB" | "RTLD_DEEPBIND" |
+            "SOL_IPV6" | "SOL_ICMPV6" | "SOL_TIPC"  if uclibc => true,
```
alexcrichton: Shouldn't these just all be omitted on uclibc?

libc-test/build.rs
```diff
@@ -309,6 +316,11 @@ fn main() {
             // This is actually a union, not a struct
             "sigval" => true,
 
+            // These structs have changed since uClibc was developed
+            "stat" | "stat64" | "dqblk" if uclibc => true,
```
alexcrichton: Shouldn't these be corrected to match what uclibc has internally?

libc-test/build.rs
```diff
@@ -309,6 +316,11 @@ fn main() {
...
+            // These structs didn't exist when uClibc was developed
+            "aiocb" if uclibc => true,
```
alexcrichton: Should this be removed from the libc crate on uclibc if it doesn't exist?

libc-test/build.rs
```diff
@@ -290,6 +295,8 @@ fn main() {
             // sighandler_t is crazy across platforms
             "sighandler_t" => true,
 
+            "locale_t" if uclibc => true,
```
alexcrichton: Could you add a short comment to why this is ignored?

libc-test/build.rs
```diff
@@ -468,6 +503,14 @@ fn main() {
             // it's in a header file?
             "endpwent" if android => true,
 
+            // These are either unimplemented or optionally built into uClibc
+            // or "sysinfo", where it's defined but the structs in linux/sysinfo.h and sys/sysinfo.h
+            // clash so it can't be tested, or they're tainted with an old struct definition, as in g"getnameinfo"
+            "mkostemp" | "mkostemps" | "mkstemps" | "futimes" | "setns" | "sysinfo" | "fallocate" | "posix_fallocate" | "preadv" | "pwritev" | "dup3" | "openpty" | "forkpty" |
+            "getloadavg" | "process_vm_readv" | "process_vm_writev" | "backtrace" | "posix_madvise" |
+            "aio_read" | "aio_write" | "aio_fsync" | "aio_error" | "aio_return" | "aio_suspend" | "aio_cancel" | "lio_listio" |
+            "newlocale" | "duplocale" | "freelocale" | "uselocale" | "nl_langinfo_l" | "wcslen" | "getnameinfo" | "wcstombs" if uclibc && linux => true,
```
alexcrichton: I wonder, should most of these just not be present in the libc crate? Especially the unimplemented ones seem like they shouldn't exist, but including the optional ones is likely ok. It'd be great though if we could verify against a build of uclibc with all optional extensions compiled in.

libc-test/build.rs
```diff
@@ -505,7 +548,12 @@ fn main() {
         // musl seems to define this as an *anonymous* bitfield
         (musl && struct_ == "statvfs" && field == "__f_unused") ||
         // sigev_notify_thread_id is actually part of a sigev_un union
-        (struct_ == "sigevent" && field == "sigev_notify_thread_id")
+        (struct_ == "sigevent" && field == "sigev_notify_thread_id") ||
+
+        // uClibc differs in signedness
+        (uclibc && struct_ == "sigaction" && field == "sa_flags") ||
+        // same here
+        (uclibc && struct_ == "msghdr" && field == "msg_iovlen")
```
alexcrichton: Could these be corrected in the uclibc definitions to have the corresponding signededness according to the uclibc header files?

libc-test/build.rs
cactorium: My bad, it looks like it builds fine with this removed!

libc-test/build.rs
cactorium: Fixed!

libc-test/build.rs
cactorium: Would you have any advice on how to remove `aiocb`? Just sticking a `#cfg[]` above it breaks because of `s!` macro wrapping it

libc-test/build.rs
cactorium: Fixed (most of them)! The couple that are left appear to exist in the uClibc source code, but they're either optional or I'm missing some compiler flags to trigger their inclusion. I'm pretty sure at least the ones related to locale are optional

libc-test/build.rs
cactorium: Fixed (except for the locale ones and backtrace (which I think is also an optional add on?))

libc-test/build.rs
cactorium: My bad, fixed!

cactorium:
Thank you for the fast reply :)! Would you have any advice on how to separate it out more cleanly? `uclibc` causes issues at both the `notbsd`, `linux`, and `mips` folders, so I'm not sure where it'd make sense to add in the new module files? So far I've moved a lot of the conflicting definitions in `linux` (which is where most of the conflicts are) into a `notuclibc` module, and used `#cfg`s at the other levels

cactorium:
Also it looks like `AF_MAX` is set incorrectly on x86_64-unknown-linux, at least on my system libc-test says that it should be 43 instead of 42. Would it be better to open a new pull request to fix that, or just include that with this one?

alexcrichton:
Oh I'd just start a brand new `uclibc` module instead of a `notuclibc` hierarchy. This'll likely result in a lot of duplication, but that's ok.

I'd probably just add a clause to ignore `AF_MAX`, that's something that's likely to change a lot!

alexcrichton:
In general libc avoids `#[cfg]` directives except inside of the "dispatch to the next layer of the hierarchy" `cfg_if!` statement. This is mostly to assist with readability as otherwise it gets really hard to know what platform defines what.

To reiterate, though, duplication is totally ok. The only risk is that when moving things around something accidentally gets lots, but it's at least all verified to be correct!

cactorium:
Okay, I think I understand now :)! Is adding a `uclibc` module in the `unix` folder an okay place for the split? It looks kinda weird to be me there (because it's the only subdirectory there that isn't an OS), but any lower and there'd be issues because of API differences already. Also would it make sense to bump up `AF_MAX` to 43 anyways? At least to me it seems like it'd make sense to try to keep track of `AF_MAX`, even if it changes pretty often!

alexcrichton:
Yeah adding a `uclibc` folder is fine by me, it's sort of just whatever's the most natural organization today. The idea is to maximize the amount of sharing where possible but that doesn't tend to happen a lot, especially on the Linux side of things.

libc-test/build.rs
```diff
@@ -387,6 +400,15 @@ fn main() {
...
 
+            // uClibc doesn't support a LOT of constants, because its latest release
+            // was before they coud be implemented yet
+            "LC_PAPER" | "LC_NAME" | "LC_ADDRESS" | "LC_TELEPHONE" | "LC_MEASUREMENT" | "LC_IDENTIFICATION" |
+            "LC_PAPER_MASK" | "LC_NAME_MASK" | "LC_ADDRESS_MASK" | "LC_TELEPHONE_MASK" | "LC_MEASUREMENT_MASK" | "LC_IDENTIFICATION_MASK" | "LC_ALL_MASK" |
+            // not entirely sure why these don't work...
+            "LC_CTYPE_MASK" | "LC_NUMERIC_MASK" | "LC_TIME_MASK" | "LC_COLLATE_MASK" | "LC_MONETARY_MASK" | "LC_MESSAGES_MASK" |
+            "MADV_MERGEABLE" | "MADV_UNMERGEABLE" | "MADV_HWPOISON" | "IPV6_ADD_MEMBERSHIP" | "IPV6_DROP_MEMBERSHIP" | "IPV6_MULTICAST_LOOP" | "IPV6_V6ONLY" |
+            "MAP_STACK" | "RTLD_DEEPBIND" | "MAP_HUGETLB" |
+            "SOL_IPV6" | "SOL_ICMPV6" if uclibc => true,
```
alexcrichton: Since uclibc has its own module now, most of these can be removed, right?

src/unix/mod.rs
```diff
@@ -837,7 +837,10 @@ extern {
 }
 
 cfg_if! {
-    if #[cfg(any(target_os = "linux",
+    if #[cfg(target_env = "uclibc")] {
+        mod uclibc;
+        pub use self::uclibc::*;
+    } else if #[cfg(any(target_os = "linux",
                  target_os = "android",
```
alexcrichton: Could you touch up the indentation on this line while you're at it?

alexcrichton:
This is looking great to me, thanks for doing the legwork here! How confident are you in all the various definitions for each architecture? Should this only add i686/x86_64 for now?

libc-test/build.rs
cactorium: Good catch! I'm still trying to work out how to enable the second set of ignored constants, but I unimplemented the first set at least

src/unix/mod.rs
cactorium: Done I think

cactorium:
So I've actually been cross compiling this for mips to use with an OpenWRT router, so I'm fairly confident that it works for that one at least! I just spent some time trying to test against `x86` and `x86_64`, but Rust doesn't seem to have targets for using `x86{,_64}-linux-uclibc`

alexcrichton:
Ok maybe we should start out with just the mips targets in that case then? I figure we can add the others in over time, running tests as they come in.

cactorium:
Sounds good to me! Is that something that I should try to add in this pull request?

alexcrichton:
Yeah let's update this PR as I think it'd be deleting the non-musl modules, right?

cactorium:
Sorry, I don't understand what you mean

alexcrichton:
Oh sure I'll clarify. So this PR adds "support" for a ton of uclibc architectures, but only tested on one. Could all untested architectres be removed from this PR?

cactorium:
Okay, I think I got it. So basically just delete the unsupported modules?

libc-test/build.rs
alexcrichton: Did you want to work on removing this clause as well as the one below that were added?

cactorium: I'm pretty sure these need to be ignored, because you can grep them in the codebase, but compiling fails, so I'm assuming these are optional features that my test case (OpenWRT) doesn't have built in. At least to me it seems like ignoring them is the best solution for now

alexcrichton: Shouldn't these constants just be deleted from uclibc files in this crate? If they don't actually exist in uclibc it seems like we shouldn't define them?

cactorium: Agh, sorry, I'm guessing I wasn't being as clear as I should've been. These do exist in uclibc, but they appear to be optionally enabled, because my test case with the OpenWRT toolchain doesn't have them, and they're in #ifdef guards in the headers. The ones that were a few lines above these that were removed definitely didn't exist

alexcrichton: Oh ok, that's fine yeah. Should we delete the ones that definitely don't exist, update the comment to match the one below, and then merge?

alexcrichton:
@cactorium wanna [update the comment](https://github.com/rust-lang/libc/pull/578#discussion_r114577194) and otherwise get ready to merge?

cactorium:
Hey! Sorry for the delay, I was busy with my sister's graduation. Stuff should be good now I think!

alexcrichton:
@bors: r+ 

Thanks!

PR #581, opened 2017-04-24
Fix size of sigset_t structure

anatol:
Linux uapi defines sigset_t as an array that contains NUMBER_OF_SIGNALS bits.
Number of signals is 64 at all hardware platforms except mips where it is 128.
Linux uapi (and musl) provides a const for number of signals called _NSIG, introduce one
in rust's libc.

### DIFF
--- a/src/unix/notbsd/android/b64/mod.rs
+++ b/src/unix/notbsd/android/b64/mod.rs
@@ -7,9 +7,11 @@ pub type mode_t = u32;
 pub type off64_t = i64;
 pub type socklen_t = u32;
 
+pub const _NSIG: ::size_t = 64;
+
 s! {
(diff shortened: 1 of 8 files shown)


alexcrichton:
@bors: r+

Thanks!

anatol:
Ok, I see what is going on here. libc implementation such as musl do not use `sigset_t` structure provided by kernel uapi. Instead they define their own structure with the same name but different format/size. Musl functions such `sigaction` copy data from libc struct to the kernel version of the struct. Musl also redefines _NSIG const to something that differs from kernel uapi. :(

I'll drop this change as I am trying to work with kernel directly and do not use libc version of `sigset_t`. Thus no reason for me to modify the structure.

PR #582, opened 2017-04-25
Add sigprocmask function

mkilgore:
sigprocmask is currently missing, this adds the definition for sigprocmask into the main list of Unix functions. This should be standard across all the Unix-like systems supported.

Note that the constants for 'how' (SIG_SETMASK, SIG_BLOCK, SIG_UNBLOCK) are already defined, it is only this function that is missing.

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -780,6 +780,12 @@ extern {
     #[cfg_attr(target_os = "netbsd", link_name = "__sigismember14")]
     pub fn sigismember(set: *const sigset_t, signum: ::c_int) -> ::c_int;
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
@bors: r+


Thanks!

mkilgore:
Any chance you could run the failed test again? It seems like it could be a fluke somehow. The failed test is i386-apple-ios and it just hung there with no output. The [same test](https://travis-ci.org/rust-lang/libc/jobs/225482521) ran fine on my original pull request.

alexcrichton:
@bors: retry

yeah let's try again

PR #585, opened 2017-04-27
Add res_init

jonhoo:
The `res_init` function, while deprecated, is critical for making networked applications work correctly when switching between networks, or between being offline and online. By default, `getaddrinfo` and friends use a cached version of `/etc/resolv.conf`, which means that network changes are not picked up by applications after they first start up. This has bitten [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=214538), [Pidgin](https://developer.pidgin.im/ticket/2825), [MongoDB](https://jira.mongodb.org/browse/DOCS-5700), and more in the past.

The logic behind exposing only `res_init` is that it is the only `res_*` function that is frequently called directly by user applications. The other `res_*` functions provide low-level access to domain lookups, whereas `res_init` is useful even if the application itself is not concerned with doing lookups.

Behind the scenes, `getaddrinfo` in glibc [ultimately calls](https://github.com/lattera/glibc/blob/a2f34833b1042d5d8eeb263b4cf4caaea138c4ad/resolv/nss_dns/dns-host.c#L196) `res_nsearch` with `&_res`. `res_init` operates directly on this global reference, and is thus more useful to expose for most applications than the non-deprecated `res_ninit` function (which operators on an arbitrary `res_state`).

As far as I can tell, `res_init` is available in [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=res_init&manpath=SuSE+Linux/i386+11.3), [NetBSD](http://netbsd.gw.com/cgi-bin/man-cgi?res_init+3+NetBSD-6.1), [OpenBSD](http://man.openbsd.org/resolver.3), [Solaris](http://www.polarhome.com/service/man/?qf=res_init&tf=2&of=Solaris&sf=), [Linux](https://linux.die.net/man/3/res_init), and [macOS](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/res_init.3.html).

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -206,9 +206,22 @@ pub const PRIO_MAX: ::c_int = 20;
 cfg_if! {
     if #[cfg(dox)] {
         // on dox builds don't pull in anything
-    } else if #[cfg(all(not(stdbuild), feature = "use_std"))] {
+    } else if #[cfg(all(not(stdbuild),
+                        feature = "use_std",
+                        not(any(target_os = "macos",
+                                target_os = "ios",
+                                target_os = "android",
+                                target_os = "openbsd",
+                                target_os = "bitrig")
+                        )))] {
         // cargo build, don't pull in anything extra as the libstd  dep
         // already pulls in all libs.
+    } else if #[cfg(all(not(stdbuild), feature = "use_std"))] {
+        // except on macOS and iOS, where we must link with lib resolv
+        // for res_init, despite libsystem_info including it:
+        // http://blog.achernya.com/2013/03/os-x-has-silly-libsystem.html
+        #[link(name = "resolv")]
+        extern {}
     } else if #[cfg(any(all(target_env = "musl", not(target_arch = "mips"))))] {
         #[link(name = "c", kind = "static", cfg(target_feature = "crt-static"))]
         #[link(name = "c", cfg(not(target_feature = "crt-static")))]
@@ -229,6 +242,7 @@ cfg_if! {
                         target_os = "bitrig"))] {
         #[link(name = "c")]
         #[link(name = "m")]
+        #[link(name = "resolv")]
         extern {}
     } else if #[cfg(target_os = "haiku")] {
         #[link(name = "root")]
@@ -694,6 +708,17 @@ extern {
                        res: *mut *mut addrinfo) -> ::c_int;
     pub fn freeaddrinfo(res: *mut addrinfo);
     pub fn gai_strerror(errcode: ::c_int) -> *const ::c_char;
+    #[cfg_attr(all(unix,
+                   not(target_os = "macos"),
+                   not(target_os = "ios"),
+                   not(target_os = "netbsd"),
+                   not(target_os = "solaris"),
+                   not(target_env = "musl")
+                   ),
+               link_name = "__res_init")]
+    #[cfg_attr(any(target_os = "macos", target_os = "ios"),
+               link_name = "res_9_init")]
+    pub fn res_init() -> ::c_int;
 
     #[cfg_attr(target_os = "netbsd", link_name = "__gmtime_r50")]
     pub fn gmtime_r(time_p: *const time_t, result: *mut tm) -> *mut tm;
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -206,19 +206,9 @@ pub const PRIO_MAX: ::c_int = 20;
 cfg_if! {
     if #[cfg(dox)] {
         // on dox builds don't pull in anything
-    } else if #[cfg(all(not(stdbuild),
-                        feature = "use_std",
-                        not(any(target_os = "macos",
-                                target_os = "ios")
-                        )))] {
+    } else if #[cfg(all(not(stdbuild), feature = "use_std"))] {
         // cargo build, don't pull in anything extra as the libstd  dep
         // already pulls in all libs.
-    } else if #[cfg(all(not(stdbuild), feature = "use_std"))] {
-        // except on macOS and iOS, where we must link with lib resolv
-        // for res_init, despite libsystem_info including it:
-        // http://blog.achernya.com/2013/03/os-x-has-silly-libsystem.html
-        #[link(name = "resolv")]
-        extern {}
     } else if #[cfg(any(all(target_env = "musl", not(target_arch = "mips"))))] {
         #[link(name = "c", kind = "static", cfg(target_feature = "crt-static"))]
         #[link(name = "c", cfg(not(target_feature = "crt-static")))]
@@ -233,12 +223,8 @@ cfg_if! {
         #[link(name = "m")]
         extern {}
     } else if #[cfg(any(target_os = "macos",
-                        target_os = "ios"))] {
-        #[link(name = "c")]
-        #[link(name = "m")]
-        #[link(name = "resolv")]
-        extern {}
-    } else if #[cfg(any(target_os = "android",
+                        target_os = "ios",
+                        target_os = "android",
                         target_os = "openbsd",
                         target_os = "bitrig"))] {
         #[link(name = "c")]
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -210,8 +210,6 @@ cfg_if! {
                         feature = "use_std",
                         not(any(target_os = "macos",
                                 target_os = "ios",
-                                target_os = "android",
-                                target_os = "openbsd",
                                 target_os = "bitrig")
                         )))] {
         // cargo build, don't pull in anything extra as the libstd  dep
@@ -237,12 +235,15 @@ cfg_if! {
         extern {}
     } else if #[cfg(any(target_os = "macos",
                         target_os = "ios",
-                        target_os = "android",
-                        target_os = "openbsd",
                         target_os = "bitrig"))] {
         #[link(name = "c")]
         #[link(name = "m")]
         #[link(name = "resolv")]
+        extern {}
+    } else if #[cfg(any(target_os = "android",
+                        target_os = "openbsd"))] {
+        #[link(name = "c")]
+        #[link(name = "m")]
         extern {}
     } else if #[cfg(target_os = "haiku")] {
         #[link(name = "root")]
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -209,8 +209,7 @@ cfg_if! {
     } else if #[cfg(all(not(stdbuild),
                         feature = "use_std",
                         not(any(target_os = "macos",
-                                target_os = "ios",
-                                target_os = "bitrig")
+                                target_os = "ios")
                         )))] {
         // cargo build, don't pull in anything extra as the libstd  dep
         // already pulls in all libs.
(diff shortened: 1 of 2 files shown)


alexcrichton:
@bors: r+

Thanks!

jonhoo:
@alexcrichton looking at the [Travis build](https://travis-ci.org/rust-lang/libc/builds/226252590) thus far, it looks like a number of tests are failing with empty logs.. Is this a known issue?

alexcrichton:
I think Travis is doing some DB maintenance right now, and tha may be causing issues with logs :(

alexcrichton:
Maybe a header file isn't `#include`'d?

jonhoo:
Ah, good catch! Added.
Any chance you could also take a look at the related https://github.com/rust-lang/rust/issues/41570 ?

alexcrichton:
Oh hm, is this required to link to the `resolv` library? That changes the calculus a bit here as it runs a high risk of causing accidental breakage :(

jonhoo:
@alexcrichton yeah, unfortunately it seems like it does, at least on some platforms. Which is kind of strange, given that `getaddrinfo` internally calls these methods (and thus must somehow be able to get at the relevant `resolv` symbols). `musl` doesn't require this, since `res_init` is implemented directly, nor do the x86_64 targets (why?), but some other glibc targets do...

jonhoo:
Also, interestingly, `cargo` complains about "undefined reference to `res_init`" even with `-lresolv`, despite

jonhoo:
Actually, we may be in luck @alexcrichton. See ae5cf307737c439dfea23fafa630ddca31d0d275

alexcrichton:
Nice! Perhaps we could leave in the leading double undescore? 

Or is that just the symbol name and the C identifier is `res_init`?

jonhoo:
The C identifier (and what shows up in all the manpages) is `res_init`. The rename to `__res_init` is completely masked from C applications by the `resolv.h` macro:
```c
#define res_init __res_init
```

aidanhs:
I can only see `res_init` in musl, not `__res_init`.

(it's also a no-op in musl, but that's an aside)

alexcrichton:
@bors: r+

Ok thanks for the info @jonhoo!

@aidanhs ah we'll probably get a link error then, but let's see what travis says.

jonhoo:
See https://stackoverflow.com/questions/16741732/using-getaddrinfo-only-checks-nscd-cache-first-time-if-dns-times-out for another discussion of why `res_init` is necessary

jonhoo:
@alexcrichton @aidanhs I think I tracked down all the major ones in fe94999e52b603c40a87dd4faf93d2918d04da00

jonhoo:
@alexcrichton can you cancel the various outstanding Travis jobs except the latest? Fixed a syntax error, a style error, and squashed, so they aren't useful.

alexcrichton:
Ah looks like I got here a little late :(

Some platforms may still be failing though?

jonhoo:
Yeah, this is quite the web of implementations to navigate. It seems like there are primarily two dimensions: whether `res_init` has been renamed to `__res_init` or not (modulo macOS/iOS), and whether `libresolv` must be linked or not. The last commit had Android and OpenBSD both failing, which I suspect is because they both require `libresolv` :/

jonhoo:
Hmm, I have no idea what is going on with Android.. It doesn't accept `res_init` nor `__res_init`, regardless of whether `libresolv` is linked...

jonhoo:
cc @malbarbo @Nercury maybe?
As far as I can tell, the Android [NDK](https://android.googlesource.com/platform/prebuilts/abi-dumps/ndk/+/master/current/x86/source-based/libc.so.lsdump#47773) does list `res_init` as available in `libc`..

malbarbo:
@jonhoo According with https://mail.gnome.org/archives/commits-list/2013-May/msg01329.html `res_init` exist but it is not declared in any header. I checked android ndk and can confirm this.

A workaround is to [skip](https://github.com/rust-lang/libc/blob/master/libc-test/build.rs#L394) the function test for android.

jonhoo:
Ah, good catch!
Hmm, ignoring `res_init` for android seems to have made OpenBSD pass as well though, which seems weird?

semarie:
about OpenBSD: libm shouldn't be require. `res_init`should be libc only. As hint, Bitrig is a OpenBSD-clone, and your PR don't manage Bitrig and OpenBSD in the same way.

Else, under OpenBSD, `res_init` is called automatically on the first use of resolver routines. The documentation regarding it is [available online](http://man.openbsd.org/res_init.3).

jonhoo:
@semarie The libc crate already links with `libm` for OpenBSD/bitrig independently of this PR. As far as I can tell, that's for other system-calls than `res_init`. You're right that I incorrectly treated bitrig differently from OpenBSD though -- that should be fixed in 6eb8d1c.

`res_init` being called automatically is sort of orthogonal to this change. The reason exposing `res_init` is important is to force a reload of the cached nameserver information (see rust-lang/rust#41570).

semarie:
I am unsure to understand the point. Under OpenBSD you could call `res_init` function as many you want, but the initialization code will be executed [only once](https://github.com/openbsd/src/blob/master/lib/libc/asr/res_init.c#L52).

It seems to me you rely to a not POSIX behaviour. You couldn't infer that all OS will behave the same way.

I will test you testcase to see how OpenBSD manage a change inside `/etc/resolv.conf` for a running program.

semarie:
ok, so for OpenBSD it is a no-issue. The libc do its jobs alone. Your testcase works here: The program starts with errors (no nameserver available) and next it connects when `/etc/resolv.conf` is rewritten externally.

As said, it does hurt to call `res_init` multiple times. So go ahead.

malbarbo:
I also tested the example in https://github.com/rust-lang/rust/issues/41570 at it worked on a real android phone. The program prints connected when the phone is back online.

jonhoo:
Interesting.. I suspect this is because Android probably uses nscd or something similar, which does automatically refresh the cache. If you're *not* running a network manager, that is not the case, and you must call `res_init` again (at least on Linux). I guess we *could* then expose `res_init` only for Linux, since that then seems like the only place it is needed.. Thoughts @alexcrichton?

semarie:
Android doesn't use glibc as Linux does (their libc version has at least [parts from OpenBSD](http://undeadly.org/cgi?action=article&sid=20140506132000), but I dunno for details).

@jonhoo your underline problem seems to be unexpected behaviour in glibc (maybe a bug should be opened over there  ?). But it could be more simple to workaround it with a global solution (like calling `res_init` unconditionnally). From your bug report, it seems Mozilla already solves it this way, so it works, no need to diverge from a working solution.

jonhoo:
@semarie I don't know that this is actually unexpected behavior as far as glibc goes. It seems quite reasonable for it to cache `/etc/resolv.conf` so that it doesn't need to do an additional two system calls for every new connection. My question above was more whether, since this only *seems* to be an issue on Linux when no network manager is in use, it's sufficient for us to just expose `libc::res_init` on Linux (and not on other targets)? And place the call to `res_init` in a `cfg!(linux)` in the modified [`lookup_host`](https://github.com/rust-lang/rust/pull/41582).

jonhoo:
@alexcrichton would you prefer just exposing the call on `Linux`, or keep the current version which exposes it on all UNIX platforms?

src/unix/mod.rs
```diff
@@ -694,6 +708,19 @@ extern {
...
     pub fn freeaddrinfo(res: *mut addrinfo);
     pub fn gai_strerror(errcode: ::c_int) -> *const ::c_char;
+    #[cfg_attr(all(unix,
+                   not(target_os = "macos"),
+                   not(target_os = "ios"),
+                   not(target_os = "netbsd"),
+                   not(target_os = "openbsd"),
+                   not(target_os = "bitrig"),
+                   not(target_os = "solaris"),
+                   not(target_env = "musl")
```
alexcrichton: Perhaps this could be converted to a whitelist? It's mostly just targeting Linux/android/FreebSD it looks like?

src/unix/mod.rs
```diff
@@ -206,9 +206,19 @@ pub const PRIO_MAX: ::c_int = 20;
...
+                        not(any(target_os = "macos",
+                                target_os = "ios")
+                        )))] {
         // cargo build, don't pull in anything extra as the libstd  dep
         // already pulls in all libs.
+    } else if #[cfg(all(not(stdbuild), feature = "use_std"))] {
+        // except on macOS and iOS, where we must link with lib resolv
+        // for res_init, despite libsystem_info including it:
+        // http://blog.achernya.com/2013/03/os-x-has-silly-libsystem.html
+        #[link(name = "resolv")]
```
alexcrichton: I'm a little wary of doing this, especially outside the `stdbuild` blocks. Just in terms of possible breakage on new platforms. I do realize though that this is required to link correctly as libc on crates.io otherwise doesn't pull in new libs (it relies on libc in std to do that).

Perhaps we could elide this section (or move it down below after the clause above) and then update the copy in libstd? We can rely on rust-lang/rust CI to block changes if it'd otherwise cause linkage problems. In the meantime we could just disable verification of the `res_init` symbol in libc as we already know it works!

jonhoo: I'm not sure I follow this? What do you mean by "move it down below after the clause above"? And by "update the copy in libstd"?

jonhoo:
Hmm, following afebd981318c61b45f35ad7792c7ea32e741ebcc, it doesn't look like `res_init` is tester for any of the platform..? :/

src/unix/mod.rs
alexcrichton: Ah sorry, basically what I mean is:

* Ideally, no libs are added
* If a lib must be added, it should be added without adding clauses to `cfg_if!`
* If a lib must be added, disable tests in libc-test and we'll test that it works in rust-lang/rust instead

jonhoo: So we should simply not link with `resolv` on macOS/iOS, both on `stdbuild` and `not(stdbuild)` , and disable the test for `res_init if apple`?

alexcrichton: Let's start off by disabling linking to `resolv` everywhere. Let's then only enable it in the `not(stdbuild)` blocks to ensure we don't accidentally break crates.io. By only enabling it in `not(stdbuild)` then we can't test it in this PR, but it's ok to leave it enabled on all platforms you have it defined on currently.

jonhoo: Okay, I've removed all the linking with libresolv in fcf5fc0257bb01e84aabfa0efd0d09638145d105. As noted in my [discussion comment](https://github.com/rust-lang/libc/pull/585#issuecomment-298983214) though, it seems like `res_init` is just never tested any more following the `res_init if android => true` addition made in afebd98.

alexcrichton:
@jonhoo hm but the current failures on CI seem to contradict your findings that it's not being tested?

jonhoo:
Huh, weird.. You're right. I guess I must have actually fixed some linking errors then :p
And now you'd like me to skip tests for `if apple`?

jonhoo:
@alexcrichton all right, I think that brings the changes in line with what you wanted?

alexcrichton:
@bors: r+

Looks great! Thanks and sorry for the runaround!

jonhoo:
No worries at all! It's important to get the details of these things right, and gave me some further insight into libc's innards. Now onto rust-lang/rust#41570!

PR #587, opened 2017-04-30
Add major, minor, makedev functions

sfackler:
Only for Linux and Android currently, though they also exist on BSDs


alexcrichton:
Wow I had no idea the identifiers `major` and `minor` were macros in C...

@bors: r+

PR #590, opened 2017-05-03
Add x86_64-linux-android test

malbarbo:
We extract the linker and some libraries from an android image and copy then to `/system`. This allows the android test to run directly (or with qemu).

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -105,7 +105,10 @@ case "$TARGET" in
 esac
 
 case "$TARGET" in
-  arm-linux-androideabi | aarch64-linux-android | i686-linux-android | x86_64-linux-android)
+  # Android emulator for x86_64 does not work on travis (missing hardware
+  # acceleration). Tests are run on case *). See ci/android-sysimage.sh for
+  # informations about how tests are run.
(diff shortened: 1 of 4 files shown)


malbarbo:
@alexcrichton I've have being testing this strategy for some time now. It works fine. It is capable of running all (excluding debuginfo) rustc tests! Considering that x86_64 android emulator does not run on travis, what do you think about this?

ci/run.sh
```diff
@@ -122,6 +125,10 @@ case "$TARGET" in
     grep "^PASSED .* tests" /tmp/out
     ;;
 
+  x86_64-linux-android)
+    $CARGO_TARGET_DIR/$TARGET/debug/libc-test
+    ;;
```
alexcrichton: I think this can be omitted right to fall down to the fall through at the bottom?

alexcrichton:
Looks great to me! Nice!

(just one tiny nit)

ci/run.sh
malbarbo: Yes it can. Thanks.

malbarbo:
Changed compiler to beta. There is no x86_64-linux-android on stable.

PR #591, opened 2017-05-07
Add experimental L4re support

humenda:
This PR introduces a UNIX-alike target for the l4re microkernel
operating system. While this system is not a UNIX system at all, it is
easiest to use the POSIX interface and hence benefit from common
definitions.

This version is a very early draft, only the basic data types have been
verified and most of the complex data types (structs and unions) have
not been ported yet.

Thanks to TobiasSchaffner for providing his patched version, to verify some assumptions.


src/unix/mod.rs
```diff
@@ -16,14 +16,17 @@ pub type cc_t = ::c_uchar;
 pub enum DIR {}
 pub enum locale_t {}
 
+#[cfg(not(target_os = "l4re"))]
```
alexcrichton: We typically prefer to avoid these sorts of `#[cfg]` annotations in libc wherever possible, mind pushing these structs further down the hierarchy to skip them on l4re?

alexcrichton:
@humenda looks like there may be a merge conflict? Mind also pushing some more of those `#[cfg]` tagged functions and such higher down the hierarchy?

humenda:
I just realized that there are other efforts to define a separate uclibc target env. I spoke to the author of these lines and agreed with him, that I'll move my changes there. I'll let you know as soon as I'm done and the integration tests work fine.

alexcrichton:
Ok sounds good to me!

alexcrichton:
@bors: r+

Thanks!

PR #592, opened 2017-05-07
Define XATTR flags, and add XATTR functions for OSX

PlasmaPower:
As found in:

http://elixir.free-electrons.com/linux/latest/source/include/uapi/linux/xattr.h

https://opensource.apple.com/source/xnu/xnu-2050.7.9/bsd/sys/xattr.h

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -145,6 +145,7 @@ fn main() {
         cfg.header("mach/mach_time.h");
         cfg.header("malloc/malloc.h");
         cfg.header("util.h");
+        cfg.header("sys/xattr.h");
         if target.starts_with("x86") {
             cfg.header("crt_externs.h");
         }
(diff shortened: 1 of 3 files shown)


PlasmaPower:
Oh, it looks like we don't have xattr functions for OSX yet. Adding those now.

PlasmaPower:
Done. It should be noted that OSX doesn't have lgetxattr, lsetxattr, etc. but rather uses a flag passed to the base versions of those functions.

PlasmaPower:
It should also be noted that in regards to OSX I'm relying on Travis to test my changes as I don't have an OSX machine. I was hoping this would be a constant only change.

PlasmaPower:
Ok, Travis is passing so it looks like my function signatures are finally correct.

alexcrichton:
Looks great to me, thanks!

@bors: r+

PR #593, opened 2017-05-15
Add AT_FDCWD for FreeBSD

PlasmaPower:


alexcrichton:
@bors: r+

Thanks!

PlasmaPower:
Oops, looks like AT_FDWCD was already defined in submodules for netbsd (though they were the same). I'm moving those definitions to the main thing. If I ammend the commit it won't break bors right?

PlasmaPower:
Actually since some netbsd variants have more AT_* definitions, I think I'll just leave netbsd the way it was.

PlasmaPower:
Okay should be good to go now, unless we do want to change the AT_FDCWD definition for netbsd.

PR #594, opened 2017-05-15
Add ENOATTR for Linux

PlasmaPower:
It's defined differently for OSX and *BSD, so having a Linux definition is helpful for cross-platform code.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -404,6 +404,10 @@ fn main() {
             "MADV_MERGEABLE" | "MADV_UNMERGEABLE" | "MADV_HWPOISON" | "IPV6_ADD_MEMBERSHIP" | "IPV6_DROP_MEMBERSHIP" | "IPV6_MULTICAST_LOOP" | "IPV6_V6ONLY" |
             "MAP_STACK" | "RTLD_DEEPBIND" | "SOL_IPV6" | "SOL_ICMPV6" if uclibc => true,
 
+            // Defined by libattr not libc on linux (hard to test).
+            // See constant definition for more details.
+            "ENOATTR" if linux => true,
+
             _ => false,
         }
     });
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -700,6 +700,11 @@ pub const PR_CAP_AMBIENT_CLEAR_ALL: ::c_int = 4;
 pub const XATTR_CREATE: ::c_int = 0x1;
 pub const XATTR_REPLACE: ::c_int = 0x2;
 
+// On Linux, libc doesn't define this constant, libattr does instead.
+// We still define it for Linux as it's defined by libc on other platforms,
+// and it's mentioned in the man pages for getxattr and setxattr.
+pub const ENOATTR: ::c_int = ::ENODATA;
+
 f! {
     pub fn CPU_ZERO(cpuset: &mut cpu_set_t) -> () {
         for slot in cpuset.bits.iter_mut() {


libc-test/build.rs
```diff
@@ -171,6 +171,7 @@ fn main() {
         if !uclibc {
             // optionally included in uclibc
             cfg.header("sys/xattr.h");
+            cfg.header("attr/xattr.h");
```
PlasmaPower: Not sure about this, but we test ulibc right?

PlasmaPower:
The `#include <attr/xattr.h>` seems more complicated than expected. It's defined by libattr, which we don't have on all platforms. That's not necessary for the definition itself though, just the tests. How should I handle this? Just disable the tests for the constant?

PlasmaPower:
Also, this change is failing the linter as the constant definition comes at the end, but I think it needs to be there to use the platform specific `ENODATA` definition. Once again, how should I handle this?

alexcrichton:
Oh right now we don't actually test uclibc on CI yet, so it's ok to back out those changes if you'd like. Also no need to worry about definition order, it doesn't have to literally come after `ENODATA`, the definitions can be in either order.

PlasmaPower:
How should I deal with needing the attr headers for tests? They aren't present on some test systems.

alexcrichton:
Whatever gets past CI is fine by me, I don't particularly have opinions how it gets past CI

PlasmaPower:
Do you mind if I just ignore the constant in the tests? The setxattr and getxattr man pages on linux specifically say `EATTR == ENODATA`.

alexcrichton:
Hm man pages typically tend to be right for one architecture and I've seen they don't always tend to be right across all of them. Do you know why this header can't be found?

PlasmaPower:
It comes from `libattr`, which we don't have for a lot of Linux platforms.

PlasmaPower:
Because of that, maybe it shouldn't be in this Rust crate, but I think it's very important as other platforms define it and it's explicitly mentioned in the man page.

PlasmaPower:
From the libattr source: http://download.savannah.gnu.org/releases/attr/attr-2.4.47.src.tar.gz

`include/xattr.h`:

```c
#include <errno.h>
#ifndef ENOATTR
# define ENOATTR ENODATA        /* No such attribute */
#endif
```

If defined, `ENOATTR` will always be an alias for `ENODATA` on Linux.

alexcrichton:
If this constant is only available on architecture, perhaps it could just be included there? The package should be installable via the various docker images we've got.

PlasmaPower:
Okay, I'll see if I can install them in docker.

ENOATTR is available for all architectures. It's just that libattr isn't installed by default on all platforms.

alexcrichton:
Hm if this isn't actually available in libc itself but requires an extra package to be installed, should it be included in the libc crate itself?

PlasmaPower:
I'm not sure. It is in libc for other platforms, so it's useful for cross-platform code. It's also mentioned in the `setxattr` and `getxattr` man pages (which don't come from libattr but Linux).

alexcrichton:
Ok, well it sounds like the story here is fishy enough that it seems safe to just skip the constant, mind adding these comments to a comment as to why we're skipping it?

PlasmaPower:
Done. I'm assuming we don't want rustdoc comments for the constant definition? We don't have them for any other constants.

alexcrichton:
Nah that's ok, we typically don't have lots of rustdoc docs in libc (we rely on libc itself documentation for that)

@bors: r+

PR #596, opened 2017-05-18
Use NOCANCEL variants for close on Mac

### LINKED ISSUE #595: plain libc close(fd) is unreliable on MacOSX
Should probably switch to the methods reccomended here

lionheart/openradar-mirror#4506

**An alternative version available on Mac OS X, close$NOCANCEL$UNIX2003 (32-bit x86) and close$NOCANCEL (x86_64) is available.**

Susurrus:
The default `close()` on OS X does not provide clear errors, instead the `$NOCANCEL` variants should be used. These are available as of OS X 10.6 onwards.

Fixes #595.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -500,7 +500,7 @@ fn main() {
             "backtrace" |
             "sysinfo" | "newlocale" | "duplocale" | "freelocale" | "uselocale" |
             "nl_langinfo_l" | "wcslen" | "wcstombs" if uclibc => true,
-          
+
             // Apparently res_init exists on Android, but isn't defined in a header:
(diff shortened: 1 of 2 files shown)


alexcrichton:
Heh that's to be expected! It's ok to just add an exception for that as we're conciously expecting it to be different than the default `close` function.

Susurrus:
Alright, looks like this is good-to-go. @docbrown Any comments on this, since you originally discovered the issue. Maybe you wanna take a shot at this and confirm it fixes #595 as expected.

docbrown:
@Susurrus I think you mean @DanielJoyce. :)

Susurrus:
@docbrown Sorry, you're right! Too many PRs going around!

PR #599, opened 2017-05-23
Adding missing structs from linux/user.h

xd009642:
Adding structs from linux/user.h (or sys/user.h depending on OS version). This adds the linux user_regs_struct and user struct from sys/user.h into libc for x86 and x86_64. Also, bumping the version number.

As an aside, I was wondering if this was also the right place to add the bit offsets for the [EFLAGS register ](https://en.wikibooks.org/wiki/X86_Assembly/X86_Architecture#EFLAGS_Register)? These aren't defined anywhere in the linux system libraries instead the definition seems to just be in the x86 and x86_64 documentation but they are useful for anyone looking to interpret eflags. If libc is the right place for these constants I'd like the opportunity to add them before it's merged in. :)


alexcrichton:
Thanks for the PR! Looks like there's some style errors though?

Could you also back out the version bump? That typically happens through dedicated PRs.

xd009642:
Done and done.

Any thoughts on constants for the EFLAGS register?

xd009642:
so there appears to still be some errors, it seems some of the generated tests fail? And also "error: requires `freeze` lang_item". cargo build and test both pass on my own pc and I'm not too familiar with the testing involved. Any guidance would be appreciated although I'll still be looking into it so hopefully will figure it out...

alexcrichton:
> Any thoughts on constants for the EFLAGS register?

Ah sorry I missed that! If these are constants in glibc or in standard C headers, sure! Otherwise though I'd recommend a separate crate for now.

For the test failures the nightly failures are ok (I'll work around them elsewhere) but for there's a few other non-nightly failures I think?

philipc:
@xd009642 To test on your local PC you can run `cd libc-test && cargo run`.

You may need to change the test to include `<sys/user.h>`.

Note that `_libc_fpstate` is from `<sys/ucontext.h>`. It's not the same as  `user_fpregs_struct` from `<sys/user.h>`. It looks like you renamed `user_regs_struct` to `_libc_state` to match that, but you should keep the original name instead. You'll have to fix this to get the test to pass.

xd009642:
Ah okay, thanks @philipc, I probably won't have a chance to do this until Tuesday as I'm away from my home pc.

xd009642:
Right I've addressed the issues and everything passes on my home machine. Hopefully it should all be good with travis now. Sorry about the delay!

alexcrichton:
@bors: r+

Looks great, thanks!

PR #602, opened 2017-05-25
Added the sync() function from unistd.h

### LINKED ISSUE #601: Missing sync() function
Hello, I'm trying to implement a service that requires the `sync()` function from `unistd.h` in Linux, but I cannot find it. More information [here](https://linux.die.net/man/2/sync).

Razican:
This fixes #601.


Razican:
Seems it fails in Android (the nightly failure seems unrelated to my changes). Should I add `#[cfg(not(target_os = "android"))]` to the function?

alexcrichton:
Ah that just means you'll want to push it further down the module hierarchy. That may required duplication but it's ok to do so.

alexcrichton:
@bors: r+

Thanks!

PR #604, opened 2017-05-26
Fill in sys/sem.h as well as getitimer and setitimer

Uplifting:
struct semun is not implemented and neither is semtimedop.


alexcrichton:

Uplifting:
So getitimer and setitimer have different function signatures on glibc and everything else.
Glibc uses `enum __itimer_which` while everything else uses `int`.  What should be done about this case?

alexcrichton:
@bors: r+

Looks good now!

PR #605, opened 2017-05-27
fill out common unix syscalls

Uplifting:
i'm expecting a few build failures, particularly from brk/sbrk.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -105,6 +105,7 @@ fn main() {
         cfg.header("syslog.h");
         cfg.header("semaphore.h");
         cfg.header("sys/statvfs.h");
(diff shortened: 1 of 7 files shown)


Uplifting:
`settimeofday` takes a `const void*` instead of a `struct timezone` on netbsd...is there a way to change the type signature for just that arch?

In order to get the tests to pass it also looks like it is necessary to define `clock_t` in mips64.

alexcrichton:
Ah yeah currently you can handle that by just duplicating the signature, it's ok to do so.

Uplifting:
At this point I think I'll just move some functions to the notbsd folder.

alexcrichton:
@bors: r+

Looks great now!

PR #607, opened 2017-06-03
Add x86_64 module to uclibc/mod.rs

humenda:


src/unix/uclibc/mod.rs
```diff
@@ -1759,6 +1759,9 @@ cfg_if! {
     if #[cfg(target_arch = "mips")] {
         mod mips;
         pub use self::mips::*;
+    } if #[cfg(target_arch = "x86_64")] {
```
alexcrichton: I think this may need to be `else if`?

src/unix/uclibc/mod.rs
```diff
@@ -1759,6 +1759,9 @@ cfg_if! {
...
+        mod x86_64;
+        use self::x86_64::*;
```
alexcrichton: This'll probably need to be `pub use`

humenda:
Thanks for the watchful review! Issues are corrected.

PR #608, opened 2017-06-07
add a number of linux syscalls

Uplifting:
vfork, wait4, getresuid, getresgid, personality, vhangup,
swapon, swapoff, getdomainname, setdomainname, iopl, ioperm


alexcrichton:
Thanks for the PR! Looks like some CI is failing though?

PR #610, opened 2017-06-08
Add test support for emscripten targets

malbarbo:
Also fix some definitions to make the tests pass for asmjs-unknown-emscripten. The build fails for wasm-unknown-emscripten with an assert error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,10 +104,9 @@ matrix:
     - os: linux
       env: TARGET=asmjs-unknown-emscripten
       rust: stable
-    # https://github.com/rust-lang/libc/pull/610#issuecomment-307264794
-    #- os: linux
-    #  env: TARGET=wasm32-unknown-emscripten
-    #  rust: stable
+    - os: linux
+      env: TARGET=wasm32-unknown-emscripten
+      rust: stable
 
     # beta
     - os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,9 +104,10 @@ matrix:
     - os: linux
       env: TARGET=asmjs-unknown-emscripten
       rust: stable
-    - os: linux
-      env: TARGET=wasm32-unknown-emscripten
-      rust: stable
+    # https://github.com/rust-lang/libc/pull/610#issuecomment-307264794
+    #- os: linux
+    #  env: TARGET=wasm32-unknown-emscripten
+    #  rust: stable
 
     # beta
     - os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,10 +103,10 @@ matrix:
       rust: beta
     - os: linux
       env: TARGET=asmjs-unknown-emscripten
-      rust: beta
+      rust: stable
     - os: linux
       env: TARGET=wasm32-unknown-emscripten
-      rust: beta
+      rust: stable
 
     # beta
     - os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,10 +103,10 @@ matrix:
       rust: beta
     - os: linux
       env: TARGET=asmjs-unknown-emscripten
-      rust: stable
+      rust: beta
     - os: linux
       env: TARGET=wasm32-unknown-emscripten
-      rust: stable
+      rust: beta
 
     # beta
     - os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,6 +101,12 @@ matrix:
     - os: linux
       env: TARGET=mips-unknown-linux-gnu
       rust: beta
+    - os: linux
+      env: TARGET=asmjs-unknown-emscripten
+      rust: stable
+    - os: linux
+      env: TARGET=wasm32-unknown-emscripten
+      rust: stable
 
     # beta
     - os: linux
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,7 +2,7 @@
 name = "libc-test"
 version = "0.1.0"
 dependencies = [
- "ctest 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "ctest 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.23",
 ]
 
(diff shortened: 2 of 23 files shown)


malbarbo:
I'm waiting the build fails for wasm-unknown-emscripten so I can open a issue and link the error.

malbarbo:
[This](https://travis-ci.org/rust-lang/libc/jobs/241020070#L990) failure is expected, I will disable wasm test an report the assert faillure to emscripten.

malbarbo:
Sadly, this changes will limit the file size that can be addressed to 2GB, but this is emscripten [issue](https://github.com/kripken/emscripten/issues/3641#issuecomment-135853958)... It [seems](https://github.com/kripken/emscripten/blob/6dc4ac5f9e4d8484e273e4dcc554f809738cedd6/system/include/libc/stdio.h#L187) that all *64_t types and *64 functions related with file system are alias to they "normal" version...

alexcrichton:
This looks awesome, thanks! Would it be possible to avoid the large whitelist of functions in `libc-test/build.rs`? could those functions be pushed down a layer perhaps?

malbarbo:
>  could those functions be pushed down a layer perhaps?

I think this will make libstd (and others crates) more difficult to maintain. Take for example the fork function, it is not available on emscripten (it is in the whitelist) but it is used in libstd. An user will only get an error if he tries to use a function that calls fork, but he can use all the others functions. This is the same approach used in emscripten, that also defines the prototype of some functions but does not provides an implementation.

I don't know how many of the whitelisted functions are used in libstd, but I think removing them from libc would make it impossible to make libstd compile for emscripten.

alexcrichton:
Oh ok in that case mind clarifying something? So does `fork` exist in emscripten? If it does then how come tests are failing? If not, how does libstd link correctly?

malbarbo:
So, now I'm confused. It toke me some time to create the whitelist, and now I think I need to review it. fork and others functions can be removed and tests still pass, but the majority of the functions are still there.

Some functions, for example `aio_*`, are compiled ok when used in all.c, but fails when linking libc-test. I think that all.c is only compiled and not linked, and `aio_*` prototypes are defined in emscripten headers, so it works. But when libc-test are linked, the symbols are not found and the linking fails.

Now I don't know what is the best approach here. If we remove these definitions, a library that uses them cannot be compiled, and must cfg_out functions that use these definitions. On the other hand, if we leave it, the library can be compiled but will fail to link (in case an unimplemented function is necessary, but I think it will link ok if no such a function is used).

alexcrichton:
So my general rule of thumb is that if you call a function in `libc` you shouldn't get a link error. In that sense if it's impossible to use these functions with wasm then they should be removed, right? If libstd is using them then presumably they're not impossible to use, so there's *something* for us to verify?

malbarbo:
I will have a couple of busy weeks from now, when things slow down, I will return to work on this.

alexcrichton:
No worries, thanks for the update!

lilianmoraru:
The `emscripten` issue(the assert @malbarbo reported) was fixed and merged upstream.
I re-enabled the wasm target on Travis, but it still fails because it uses only prebuilt Emscripten SDKs: https://travis-ci.org/lilianmoraru/libc

alexcrichton:
Great! I'm not sure we have the capacity on Travis though to use a non-prebuilt emscripten sdk, but is there perhaps a docker image to use?

alexcrichton:
(prebuilt docker image)

lilianmoraru:
@alexcrichton `1.37.14` was tagged just now(includes the fix).
Do you know(I don't) what needs to happen now to bring that in?

alexcrichton:
AFAIK it's just updating the docker images in this PR, but that's mostly just a guess!

lilianmoraru:
`emsdk install sdk-1.37.14-64bit` doesn't work yet(of course, it was tagged ~30 minutes ago).
I'll be back here to continue spamming the issue when it will be available :smile:

lilianmoraru:
The SDKs are [built and stored](https://github.com/juj/emsdk/blob/master/emsdk#L23) on https://s3.amazonaws.com/mozilla-games/emscripten/.
Searched for `mozilla-games` but didn't find much. No idea who does the builds...

malbarbo:
Update the emscripten sdk, let's see how travis goes. I didn't work on the whitelist issue.

lilianmoraru:
The wasm build passed.
@malbarbo, this is fantastic!!
Thanks for moving this forward.

alexcrichton:
Awesome, thanks @malbarbo!

My only hesitation is the large whitelist of functions to ignore in `libc-test/build.rs`. Does libstd use these functions? If so, how come we can't check them? If not, can they be omitted on emscripten? (presumably they don't exist)

lilianmoraru:
@alexcrichton What needs to be done(finished-up) here to push this forward?
Proper emscripten and wasm support in Rust is basically blocked by this and it would be great to have this in.

alexcrichton:
@lilianmoraru I'd like to understand more about the whitelist of functions here that aren't tested, but other than that it should just be a rebase away from landing

lilianmoraru:
I rebased and solved the merge conflicts but it seems that I did it wrong(had to force-push...).
I guess I have to read how to correctly rebase and try another day again(meanwhile maybe you will address the [`.travis.yml` issue](https://github.com/rust-lang/libc/pull/652#issuecomment-322603670), if it is an issue - so this can be rebased more cleanly).

pepyakin:
> - no release yet for WebAssembly/binaryen#1111 ?( rust-lang/rust#42630 )
> - no release yet for WebAssembly/binaryen#1106 ?( rust-lang/rust#42629 )

I might be terribly wrong, but isn't binaryen builds from the master?

https://github.com/rust-lang/rust/blob/master/src/ci/docker/scripts/emscripten-wasm.sh#L31

```
curl "https://storage.googleapis.com/wasm-llvm/builds/linux/lkgr.json" \
  | jq '.repositories.binaryen.hash'
```
at time of this writing, points on [`21d06ae`](https://github.com/WebAssembly/binaryen/tree/21d06ae) witch seems to be master, meaning that mentioned fixes there already.

> had to force-push..

It doesn't mean that you did something wrong.

alexcrichton:
Thanks for taking this on @lilianmoraru! Other than the .travis.yml weirdness, was there anything I could help out with?

lilianmoraru:
@alexcrichton Nah. I pushed my changes to @malbarbo's branch( https://github.com/malbarbo/libc/pull/1 ), now I'll just wait.

alexcrichton:
I've rebased this in https://github.com/rust-lang/libc/pull/742 with what I had in mind, basically just copying around a lot more things to avoid many changes to libc-test and ensure everything is validated on emscripten.

alexcrichton:
Thanks again though for the patch @malbarbo!

lilianmoraru:
Personally, I would of liked to preserve the history of @malbarbo's work :(

alexcrichton:
Egad sorry! I've merged it all into the other PR

malbarbo:
Thanks @alexcrichton for finishing this! I was worried that removing the functions white list would be too disturbing. Moving emscripten up in the hierarchy was the right choice.

@lilianmoraru sorry  for not replaying to you messages..., I've been very busy lately. And thank you for helping moving this forward.

PR #612, opened 2017-06-10
Add unix seteuid()

shawnanastasio:
Adds prototype for unix seteuid() as defined [here](http://pubs.opengroup.org/onlinepubs/009695399/functions/seteuid.html).


alexcrichton:
@bors: r+

Thanks!

PR #613, opened 2017-06-10
OpenBSD kqueue EV_RECEIPT and EV_DISPATCH

jan-schreib:
OpenBSDs kqueue now has EV_RECEIPT and EV_DISPATCH.

I wrote a patch and the devs merged it into current:
https://marc.info/?l=openbsd-tech&m=149621427511219&w=2

This change extends the rust libc crate to provide the feature.


jan-schreib:
Test fails because its using OpenBSD 6.0 which does not have the patch. OpenBSD-current has, though.

alexcrichton:
Thanks! If this is super-new, want to add an exception to the testing script with a comment as to why it's ignored?

jan-schreib:
Sorry for the late response. I don't see where to change that. Its in the buildprocess on travis...

alexcrichton:
Ah you can [find some instructions in the README](https://github.com/rust-lang/libc#test-before-you-commit).

semarie:
@jan-schreib I expected #728 to automatically mark this PR as merged. I should have miss something in my git command to import your patch in my PR with additionnal commit (to skip the added constants)

alexcrichton:
I'm going to close this due to inactivity, but feel free to reopen with tests fixed!

PR #616, opened 2017-06-13
Add structs defined in linux/input.h

Arvamer:
I was using definitions for these structs from `ioctl` but because @cmr decide to deprecate his crate (and yanked all versions :() I think that `libc` is the best place for them.

In original C header, primitive types uses aliases like `__u16` or `__s32`; for now I replaced them with Rust's counterparts but I'm not sure if it is ok, especially because tests were failing for `u64` (`unsigned long long` vs `unsigned long` on x86_64). Also, should I do something special with union in `ff_effect`?

I would like also to add all constants form `linux/input.h` and `linux/input-event-codes.h` if this PR will be accepted.

### DIFF
--- a/ci/docker/i686-unknown-linux-musl/Dockerfile
+++ b/ci/docker/i686-unknown-linux-musl/Dockerfile
@@ -18,6 +18,13 @@ RUN curl https://www.musl-libc.org/releases/musl-1.1.15.tar.gz | \
     CC=gcc CFLAGS=-m32 ./configure --prefix=/musl-i686 --disable-shared --target=i686 && \
     make CROSS_COMPILE= install -j4 && \
     cd .. && \
-    rm -rf musl-1.1.15
+    rm -rf musl-1.1.15 && \
+# Install linux kernel headers sanitized for use with musl
+    curl -L  https://github.com/sabotage-linux/kernel-headers/archive/v3.12.6-5.tar.gz | \
+    tar xzf - && \
+    cd kernel-headers-3.12.6-5 && \
+    make ARCH=i386 prefix=/musl-i686 install -j4 && \
+    cd .. && \
+    rm -rf kernel-headers-3.12.6-5
 ENV PATH=$PATH:/musl-i686/bin:/rust/bin \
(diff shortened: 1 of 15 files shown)


Arvamer:
So mips64 and powerpc64 doesn't like `c_ulonglong == __u64` while x86_64 (and probably aarch64) are against `u64 == __u64`.

And musl targets don't have `linux/input.h` – but shouldn't this come from kernel package? At least on Archlinux, it's owned by `linux-api-headers`.

alexcrichton:
Ah there's [an issue](https://github.com/rust-lang/libc/issues/235) about what to do for `linux/*.h` which has what I remember as some scary warnings about including `linux/*.h` headers? Not all of them are apparently intented for consumption in user-space?

Arvamer:
I would be surprised if `linux/input.h` wasn't intended for consumption in user-space. It's the only place where types for working with evdev are defined, and for example SDL2 include it directly.

I also tried running test suite locally with `CC=musl-gcc HOST=x86_64-unknown-linux-musl` and it works.

alexcrichton:
I personally don't know much about linux header files and whether or not they should be included. This can't merge until the CI is green though, and I don't have a particular preference on how that's done.

Arvamer:
CI is green now.

alexcrichton:
@bors: r+

Looks great, thanks!

cmr:
Note that those types were redeclared in my evdev crate. ioctl has been
deprecated for a long time, all the functionality was merged into nix
and the types belong in their own crate somewhere.

On Thu, Jun 15, 2017, at 10:29, bors wrote:
> :lock: Merge conflict
>
> --
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/rust-lang/libc/pull/616#issuecomment-308750828


--
cmr
http://octayn.net/
+16038524272

Susurrus:
Was there consensus on what to do with this?

alexcrichton:
Yes this is fine, sorry for the delay! @Arvamer feel free to ping me whenever a PR is updated, unfortunately GitHub doesn't send out notifications for that :(

PR #617, opened 2017-06-13
Add dirfd on netbsd

Mic92:
follow up of https://github.com/rust-lang/libc/pull/571


alexcrichton:
Thanks! Could the `DIR` struct be defined to avoid the unsafe cast/load?

Mic92:
It is already defined as incomplete type in `src/unix/mod.rs`

```
pub enum DIR {}
```

alexcrichton:
Ok in that case we can go with this. Looks like CI is failing though?

Mic92:
here we go!

Susurrus:
@Mic92 Thanks for submitting this!

PR #619, opened 2017-06-15
Fixed missing EPOLLONESHOT for android

### LINKED ISSUE #548: Missing implementation `libc::EPOLLONESHOT` on the android
epoll EPOLLONESHOT is implemented since linux kernel 2.6.2, but on android platform EPOLLONESHOT is missing in rust libc.

Should be an bug in rust libc implementation.
Because EPOLLONESHOT is available on android platform.
see [https://doc.rust-lang.org/libc/arm-linux-androideabi/libc/index.html?search=EPOLLONESHOT](https://doc.rust-lang.org/libc/arm-linux-androideabi/libc/index.html?search=EPOLLONESHOT)

gdf8gdn8:
resolve #548


alexcrichton:
@bors: r+

Thanks!

PR #620, opened 2017-06-15
struct tm for Windows

pornel:
As documented here:

https://msdn.microsoft.com/en-us/library/windows/hardware/ff567981(v=vs.85).aspx

I've verified it matches `corecrt_wtime.h` file from Win10 SDK.


alexcrichton:
@bors: r+

Thanks!

PR #622, opened 2017-06-18
Unify some of BSD flags

equal-l2:


alexcrichton:

equal-l2:
I don't have any idea about the CI failing.
Maybe should I simply copy `O_NONBLOCK` value to `O_NDELAY`, instead of assigning?

equal-l2:
@alexcrichton Any thought?

equal-l2:
I've found that all I need is unification of BSD flags.
I'll make another PR for that.

PR #623, opened 2017-06-19
Unify some BSD flags

equal-l2:
These BSD flags have common value in all OSes, so I've merged them into root `mod.rs`.
- `O_ACCMODE`
- `O_RDONLY`
- `O_WRONLY`
- `O_RDWR`
- `O_APPEND`
- `O_CREAT`
- `O_TRUNC`
- `O_EXCL`
- `O_SYNC`
- `O_NONBLOCK`
- `O_SHLOCK`
- `O_EXLOCK`
- `O_FSYNC`
- `O_NDELAY`

PR #624, opened 2017-06-19
Add BSD flags

equal-l2:


equal-l2:
That will be fine with #623.

alexcrichton:

PR #625, opened 2017-06-22
Add more termios constants

Susurrus:
Went through and found some more termios constants that were missing and added them. I skipped some platforms (solaris, uclibc) and assumed Android has the same constants are Linux, as I couldn't find a good reference for what Android exposes, but it generally follows linux, so I assumed the same was true for some of those values.

## References:
 FreeBSD:
   * https://github.com/freebsd/freebsd/blob/d39171bb1f2256bd3bf018314aec600f79b89bc6/sys/sys/_termios.h
   * https://github.com/freebsd/freebsd/blob/d39171bb1f2256bd3bf018314aec600f79b89bc6/include/termios.h
 NetBSD: https://github.com/NetBSD/src/blob/0bff031265b50be8e8b7719aed82212928d6c1df/sys/sys/termios.h
 musl: /usr/x86_64-linux-musl/include/bits/termios.h (identical to Linux-x86_64)
 linux: /usr/include/bits/termios.h
 OpenBSD: https://github.com/openbsd/src/blob/5271000b44abe23907b73bbb3aa38ddf4a0bce08/sys/sys/termios.h
 DragonFlyBSD:
   * https://github.com/DragonFlyBSD/DragonFlyBSD/blob/ddad68003a1b070d6955ae737c834cd3267d3ead/sys/sys/_termios.h
   * https://github.com/DragonFlyBSD/DragonFlyBSD/blob/ddad68003a1b070d6955ae737c834cd3267d3ead/sys/sys/termios.h
 Android: Just use the same as Linux
 uclibc: skipped
 haiku-os: http://cgit.haiku-os.org/haiku/tree/headers/posix/termios.h
 mac: /usr/include/sys/termios.h and /usr/include/sys/_types/_posix_vdisable.h
 solaris: skipped
 bitrig (follows openbsd): https://github.com/bitrig/bitrig/blob/d31f5220a98fae4e74abafe27c8d150843cc8105/sys/sys/termios.h


Susurrus:
Tests pass and I've successfully used these constants in a [downstream PR](https://github.com/nix-rust/nix/pull/527), so I think this should be RTM.

Susurrus:
Nvm, missing some constants on a few platforms I think.

src/unix/bsd/apple/mod.rs
```diff
@@ -1226,22 +1226,22 @@ pub const NOTE_TRACK: ::uint32_t = 0x00000001;
...
-pub const CR2: ::c_int  = 0x00002000;
-pub const CR3: ::c_int  = 0x00003000;
-pub const FF0: ::c_int  = 0x00000000;
-pub const FF1: ::c_int  = 0x00004000;
-pub const BS0: ::c_int  = 0x00000000;
-pub const BS1: ::c_int  = 0x00008000;
-pub const TAB3: ::c_int = 0x00000004;
-pub const VT0: ::c_int  = 0x00000000;
-pub const VT1: ::c_int  = 0x00010000;
+pub const NL0: ::tcflag_t  = 0x00000000;
```
alexcrichton: Unfortunately while likely more correct this is a breaking change, so it'll have to happen in a future PR.

Susurrus: I can file that as a separate PR for that change then. Unfortunate that breaking changes can't be done to `libc` even though it's < 1.0. Is there a roadmap for getting addressing these breaking changes in a reasonable timeframe?

alexcrichton: Not currently, no.

Susurrus: So I've not changed the datatypes for the non-apply BSD platforms in this PR, can I keep those as `tcflag_t` types then and just make sure I don't touch the old ones?

Susurrus: Or do you want those constants to be `c_int` across all platforms?

alexcrichton: nah switching to `tcflag_t` seems fine by me.

Susurrus:
This failed CI because the Mac builds mostly timed out, but they should pass. Could someone with Travis permissions rerun the build or r+ it so homu will?

Susurrus:
Still failing because OS X builds in Travis aren't even running. I've pushed this branch again to force a new test run, and hopefully this one passes.

Susurrus:
Looks like something's wrong with the mac runners on Travis right now, and I don't know how to fix that. travis's status page doesn't indicate anything bad is going on.

Susurrus:
Man, Travis was super difficult on this PR. Thanks tho for helping push this through!

PR #626, opened 2017-06-24
[uclibc] add more structs to x86_64

humenda:
This adds more structs to the uclibc/x86_64 port.

### DIFF
--- a/src/unix/uclibc/x86_64/mod.rs
+++ b/src/unix/uclibc/x86_64/mod.rs
@@ -2,6 +2,7 @@
 
 pub type blkcnt_t = i64;
 pub type blksize_t = i64;
+pub type clock_t = i64;
 pub type c_char = u8;
 pub type c_long = i64;
 pub type c_ulong = u64;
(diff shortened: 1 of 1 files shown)


Susurrus:
Removing the `in_addr` and `in6_addr` structs is a breaking change and will prevent this PR from being merged. I suggest you remove those changes, either fixing those structs or leaving a comment there like you did stating they may be incorrect, in order to get this PR merged in a timely fashion.

alexcrichton:
This is actually a tier 3 platform so breakage is ok (nothing is automatically verified here). @humenda despite this though was there a reason to remove those structs?

humenda:
L4Re doesn't have networking and this is what I am targeting with uclibc/x86_64.
That made me a bit ignorant and I simply removed the unverified structs. I'll
add them again, since they might be useful on other platforms.

alexcrichton:
Ok, I think you're the main user of uclibc on x86_64 for now, so sounds good!

@bors: r+

PR #629, opened 2017-06-27
Correct datatype for various termios constants

Susurrus:
This was a part of #625, but I moved it to a separate PR as it's a breaking change and those are currently blocked for Tier 1 platforms.


Susurrus:
This also need to address `NLDLY` and friends.

Susurrus:
`OCRNL`, `ONOCR`, `OFILL`, `ONLRET`, and `OFDEL` need to be changed to `tcflag_t` as well.

gnzlbg:
@Susurrus can you rebase, and maybe merge this PR with the one that removes termios for musl ?

I would like to do a libc release without this, and then, if this is ready, do a release that only contains this change. 

We can then try to update libc on rust upstream, and see if there is any downfall (if anybody complains). If somebody does, we can always revert the change, and yank the libc version. If nobody complains, then this would be ok.

Susurrus:
Rebased and merged #1015 into this PR.

Susurrus:
@gnzlbg I updated and rebased this PR. What're next steps to merge this?

gnzlbg:
> What're next steps to merge this?

Nothing. There are a couple of breaking changes in the pipeline. I'll prepare a release containing them, and see if anything breaks. If nothing breaks then that would be it. Otherwise we'll need to revert them.

PR #630, opened 2017-06-28
Fix broken html_root_url link

hngnaig:
All `html_root_url` redirect to 404 github page. I fixed this with add path `/libc`.


hngnaig:
cc @alexcrichton

alexcrichton:
Thanks for the PR! I think these are mostly just intended for rustdoc itself, though, which I believe already appends this information?

alexcrichton:
I'm going to close this because I think the URLs are correct as-is

PR #335, opened 2016-07-20
Add support for pthread_condattr_t on Unix platforms.

tmiasko:
Following types and function have been added:
* pthread_cond_init
* pthread_condattr_t
* pthread_condattr_init
* pthread_condattr_destroy
* pthread_condattr_getclock [0]
* pthread_condattr_setclock [0]

[0] Not available on OSX.

PR #633, opened 2017-06-28
Add pthread_rwlockattr APIs

main--:
I'm basically trying to mirror the existing code for pthread_mutexattr.

Hopefully the non-linux parts are correct - finding the right definitions is rather confusing and occasionally even contradicts existing definitions (e.g. [this](https://github.com/practicalswift/osx/blob/a26375e7b3e01c83e26a23977f909438ed9b31b4/src/libpthread/src/internal.h#L295) `pthread_condattr_t` doesn't match #335).


alexcrichton:
@bors: r+

Awesome, thanks!

PR #636, opened 2017-07-01
Add constants for termios ioctls

Susurrus:
These should be `c_int` from what I could see for Linux and I assumed the others were the same.


alexcrichton:
Looks like CI may be failing?

Susurrus:
Indeed, evidently i forgot how to write out octal literals.

This actually brings up a question I had, but never asked: should I try to preserve constants in their original base where possible? For example I used octal when defining these constants on NetBSD and family because that's how they're written in their source files, though these values are actually identical across every platform. Was this the right approach? I haven't seen a comment on any of my PRs stating a preference one way or another, but I know for the `winapi` crate, the maintainer prefers keeping things looking similar.

Susurrus:
Alright, this should be GTG!

PR #640, opened 2017-07-03
Add missing termios constants for Mac

Susurrus:
Didn't add these the first time around, because it says they're unimplemented, but so are the `NL0` and related constants I added. This also makes testing a little easier having these constants available, and they are in the termios headers on Mac.


Susurrus:
Alright, tests pass and this should be GTM.

PR #646, opened 2017-07-06
Add experimental newlib bindings

FenrirWolf:
I'm not sure how much desire there is for something like this in the libc crate, but I've been working with a newlib-based toolchain for a while and thought I'd throw this PR up and see what happens.

These bindings are more specifically targeted towards the devkitARM toolchain from http://devkitpro.org rather than newlib in general. I'd be happy to try making things more platform and toolchain-agnostic, but I'm not completely sure what the best way to do that is. I can move more of the arch-specific bindings to the `arm` folder, but should there also be a `devkitarm` subdirectory to further separate specific bindings from other newlib implementations?

There's also the question of if the bindings should live in the `unix` directory in the first place. Newlib aims to provide a unix/posix-like environment and it would be nice to inherit common unix definitions by default, but it can target anything from embedded devices to custom userlands, and as such it often lacks APIs that are common in other libc variants, such as pthreads.

### DIFF
--- a/src/unix/newlib/mod.rs
+++ b/src/unix/newlib/mod.rs
@@ -1,0 +1,604 @@
+use dox::mem;
+
+pub type blkcnt_t = i32;
+pub type blksize_t = i32;
+pub type clock_t = i32;
+pub type clockid_t = ::c_ulong;
+pub type dev_t = u32;
+pub type fsblkcnt_t = u64;
+pub type fsfilcnt_t = u32;
+pub type id_t = u32;
+pub type ino_t = u32;
+pub type key_t = ::c_int;
+pub type loff_t = ::c_longlong;
+pub type mode_t = ::c_uint;
+pub type nfds_t = u32;
+pub type nlink_t = ::c_ushort;
+pub type off_t = i64;
+pub type pthread_t = ::c_ulong;
+pub type pthread_key_t = ::c_uint;
+pub type rlim_t = u32;
+pub type sa_family_t = u8;
+pub type socklen_t = u32;
+pub type speed_t = u32;
+pub type suseconds_t = i32;
+pub type tcflag_t = ::c_uint;
+pub type time_t = i32;
+pub type useconds_t = u32;
+
+s! {
+    pub struct sockaddr {
+        pub sa_family: sa_family_t,
+        pub sa_data: [::c_char; 14],
+    }
+
+    pub struct sockaddr_in {
+        pub sin_family: sa_family_t,
+        pub sin_port: ::in_port_t,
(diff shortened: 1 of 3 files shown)


src/unix/newlib/mod.rs
```diff
@@ -0,0 +1,566 @@
...
+                       serv: *mut ::c_char, servlen: socklen_t,
+                       flags: ::c_int) -> ::c_int;
+}
+
+cfg_if! {
+    if #[cfg(target_arch = "arm")] {
+        mod arm;
+        pub use self::arm::*;
+    } else {
+        pub use unsupported_target;
```
alexcrichton: Mind adding a comment as to the intention here? (a semi-nicer compiler error message)

src/unix/newlib/mod.rs
```diff
@@ -0,0 +1,566 @@
+use dox::mem;
```
alexcrichton: I think this may be unused?

src/unix/newlib/mod.rs
FenrirWolf: Basically I've only verified bindings for ARM so it needs to fail in some fashion for things that aren't ARM. the uclibc module does something similar so I just imitated that: https://github.com/rust-lang/libc/blob/master/src/unix/uclibc/mod.rs#L1771-L1782

src/unix/newlib/mod.rs
FenrirWolf: Right, I forgot to remove that earlier. I'll do that when I push another fix that I need to make.

src/unix/newlib/mod.rs
alexcrichton: Oh yeah this is fine, it'd just be good to have a comment here

FenrirWolf: Sure, I can definitely add a comment. Should I also change the `unsupported target` module to something like `target_arch_not_yet_supported` or something similar?

alexcrichton: Either's fine by me!

src/unix/newlib/mod.rs
FenrirWolf: Or not. Turns out I actually use that import after all. `FD_CLR()` and friends depend on it and I'd just forgotten to include them last time around.

ZetaNumbers:
Hello @FenrirWolf. I've made #3345, please take a look at it. I would like to know where did you get these values from, so i could try to not break it with this PR. As i've said, i've looked at [devkitPro/newlib](https://github.com/devkitPro/newlib) and i haven't found anything like your values there.

PR #647, opened 2017-07-07
harmonize dirent->d_ino according to posix

Mic92:
POSIX 2004 states that dirent shall include the field `d_ino` of type `ino_t`:
http://pubs.opengroup.org/onlinepubs/009695399/basedefs/dirent.h.html

Updated darwin/dragonfly/freebsd and netbsd definitions to met POSIX and make
portable use easier.

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -113,7 +113,7 @@ s! {
     }
 
     pub struct dirent {
-        pub d_ino: u64,
+        pub d_ino: ino_t,
         pub d_seekoff: u64,
         pub d_reclen: u16,
         pub d_namlen: u16,
--- a/src/unix/bsd/freebsdlike/dragonfly/mod.rs
+++ b/src/unix/bsd/freebsdlike/dragonfly/mod.rs
@@ -55,7 +55,7 @@ s! {
     }
 
     pub struct dirent {
-        pub d_fileno: ::ino_t,
+        pub d_ino: ::ino_t,
         pub d_namlen: u16,
         pub d_type: u8,
         __unused1: u8,
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -44,7 +44,7 @@ s! {
     }
 
     pub struct dirent {
-        pub d_fileno: u32,
+        pub d_ino: ::ino_t,
         pub d_reclen: u16,
         pub d_type: u8,
         pub d_namlen: u8,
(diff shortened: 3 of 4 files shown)


Mic92:
motivation see `nix-rust`.

semarie:
changing the name of `dirent` struct's member looks like a breaking change in libc. any Rust program relying on old name will not compile any more.

I am also a bit surprised that travis doesn't complain: a quick look at source code of BSDs show no ` d_ino` field in `dirent`:
- [FreeBSD](http://bxr.su/FreeBSD/sys/sys/dirent.h#50)
- [NetBSD](http://bxr.su/NetBSD/sys/sys/dirent.h#50)
- [DragonFly](http://bxr.su/DragonFly/sys/sys/dirent.h#59) (here it is present for KERNEL)

Maybe there is a `#define` somewhere ?

POSIX is only a specification, not all OSs follow it strictly.

Mic92:
I suppose it its a define. However we have not equivalent in Rust. In the current form the definition useless as it is unnecessary inconsistent also operating systems follow POSIX here from the semantics (Somebody just blindly copied the definition from the man pages). I can also work around this in nix in instead.

Mic92:
It should be at least `ino_t` on all platforms.

Mic92:
I fixed this in nix now. Once the pull request there is merged, I will also close this pull request:
https://github.com/nix-rust/nix/pull/558/commits/6ed79b03262d4b88f8001b762cca414c369e4700#diff-03012aeef26d4c7454d6dc08d0e31312R57

alexcrichton:
Thanks for the PR! As this is a breaking change we can't merge this at this time, but I'll tag it as such.

gnzlbg:
@Mic92 could you rebase this ? 

Also, do you have a link to the headers of each of the OSes this PR modifies? I'd like to check the name of this field in those headers. libc should do what those headers do. Like @semarie says, not all OSes follow POSIX properly, and this library should reflect that (`nix` doesn't have to do that though).

Mic92:
@gnzlbg I can rebase once, it becomes clear that it will be merged.
I strongly disagree that this crate should use the same names as the underlying libc's. This only makes platform-independent development unnecessary difficult for not good reason.

gnzlbg:
> This only makes platform-independent development unnecessary difficult for not good reason.

The only goal of this crate is to provide raw FFI bindings to the platform APIs "as is". This is particularly important for projects like bindgen, where they currently translate struct field names "as is" from the platform to Rust, and they expect that if the type exists in libc, its fields are named just like in the platform. 

If we were to start translating platform API names here because we don't like their names, all C FFI binding generators would need to apply those exact translations to be able to use the libc crate. 

If you are looking for a library supporting platform-independent development for unix-like systems, check out the `nix` crate. That's exactly what its goal is.

Issue #655, opened 2017-07-09
SCM_TIMESTAMP missing in multiple targets


Wolvereness:
The constant SCM_TIMESTAMP appears to be missing on many platforms.

ndusart:
@Wolvereness #656 is merged, can you confirm this issue is fixed now ?

Wolvereness:
Appears to be fixed as of 472ffbb75fcbb771d59208641846db196295c021.

PR #656, opened 2017-07-09
add missing socket constants

ndusart:
Add some missing socket constants that are hardcoded by `nix` (nix-rust/nix#647)

I took the opportunity to merge some constants in a upper module when applicable.


ndusart:
Fixes #655

ndusart:
I've got some problems defining `SO_RESTRICTION` and `SO_RESTRICT_*` constants for apple platforms.

These are defined in `sys/socket.h`: https://github.com/opensource-apple/xnu/blob/dc0628e187c3148723505cf1f1d35bb948d3195b/bsd/sys/socket.h#L188-L190
in a `#ifdef PRIVATE` block, but even when defining it in build.rs (using `cfg.define("PRIVATE", None)`), it fails to get these constants.

@alexcrichton Do I misused the ctest generator for this ?

alexcrichton:
Nah I can see `-DPRIVATE` passed but it looks like that's not working? Maybe these can't be defined for OSX yet?

Susurrus:
Reproducing my comment in nix-rust/nix#647 here:

Those constants don't actually exist on Mac as of 10.11.6 (I have one locally). I did `grep SO_RESTRICT -r /usr/include` and turned up nothing. Additionally I looked through `/usr/include/sys/socket.h` and those constants are gone (though the values aren't reused by other constants).

I think the `PRIVATE` constant is a good hint that these should never have been used by external code anyways, so it's unsurprising that they've removed them. I don't think they should be added to `libc`.

ndusart:
OK it's RTM then.

ndusart:
Just added missing `SOCK_RDM` for linux and android.

Really ready to be merged now :)

PR #660, opened 2017-07-10
Add some more constants found on FreeBSD and DragonFly

mneumann:
Those are used by nix crate.


ndusart:
`SCM_RIGHTS` ans `TCP_MAXSEG` were just added in #656. You can remove them, they are present on BSD platforms now.

PR #661, opened 2017-07-12
Add CMSPAR to linux/mips

Susurrus:


alexcrichton:
Looks like CI is failing?

Susurrus:
It's weird this is failing. This is declared similarly to `CRTSCTS` and `CIBAUD`, both of which are defined for mips. The sources for `linux` and `glibc` seem to suggest that `CMSPAR` is available as well. @alexcrichton Do you know where the headers for mips are as maybe I'm not looking at the right ones? I was using [this](https://github.com/torvalds/linux/tree/master/kernel) for Linux and [this](https://github.molgen.mpg.de/git-mirror/glibc) for glibc.

alexcrichton:
I'd look inside the containers rather than online, that's what's being tested against.

Susurrus:
Alright, did a lot of rooting around in the docker image. Though it's not necessary as the headers come from `libc6-dev-mips64-cross` so I simply could have installed and searched `/usr/mips64-linux-gnuabi64` instead of doing it in the docker image, but I had to learn that first.

Turns out that including `#include <asm/termios.h>` works but `#include <termios.h>` directly does not. And you can't have them both included at the same time. I'm unsure what's "correct" to do in this situation. As in I'm not really certain `CMSPAR` **is** available on mips64 given where it is in the headers and I'm also not certain how to modify the build script to get this working. Should I try to always import `asm/termios.h` and see if that works?

alexcrichton:
As long as CI is green I don't really have a preference as to how this is done.

Susurrus:
I can't figure out a way to get the includes to work to support this constant, changing the includes leads to duplicate datatype errors. I'm pretty certain this exists on mips/linux so is it fine if I disable checking for it on mips?

alexcrichton:
If this is just one constant and it's already defined on a bunch of other platforms and you've verified the value manually, yeah it seems ok to disable checking with a precise comment as to why.

Susurrus:
@alexcrichton So I disabled the check for `CMSPAR` on exclusively the platforms that have this problem and tried to document it clearly. Hopefully that's good enough.

PR #668, opened 2017-07-13
Add more termios s390x constants

Susurrus:
These are all that remain for nix to compile on s390x.


Susurrus:
Just pushed a new version removing `BOTHER` (though I think this should be defined...), so this should pass now.

PR #669, opened 2017-07-16
Add missing constants for mac

GuillaumeGomez:
I didn't find these constants on other systems than mac so only mac is updated...


GuillaumeGomez:
All good!

PR #671, opened 2017-07-17
Add signalfd and related to android

Susurrus:


alexcrichton:
Looks like CI is failing?

Susurrus:
@alexcrichton Yes, 64-bit android builds are failing because the 64-bit fields within `signalfd_siginfo` are the wrong size I guess. This doesn't jive with what I'm seeing in the source headers anywhere, however, so I'm trying to figure out what's going on here. I'll come back to this in the coming days and see if I can sort it out.

Susurrus:
@ndusart You updated the Android NDK version recently, so I'm wondering if you have any familiarity with Android APIs here. As far as I can see when searching the same headers I find that `signalfd_siginfo` field is defined thesame for all architectures. I even checked the `__u64` is the same on all platforms, as it is. Is there something I'm missing about 64-bit versus 32-bit Android platforms?

ndusart:
These fields are 64bits on each platforms but there are not the same type.

`__u64` is a typedef to `unsigned long long` (defined in [`asm-generic/int-ll64.h`](https://android.googlesource.com/platform/prebuilts/ndk/+/dev/platform/sysroot/usr/include/asm-generic/int-ll64.h), included from [`asm-generic/types.h`](https://android.googlesource.com/platform/prebuilts/ndk/+/dev/platform/sysroot/usr/include/asm-generic/types.h))

and `uint64_t` is only `unsigned long long` on 32-bits architectures, not 64-bits where it is `unsigned long` (see [`stdint.h`](https://android.googlesource.com/platform/prebuilts/ndk/+/dev/platform/sysroot/usr/include/stdint.h#42))

you should use `::c_ulonglong` here

Susurrus:
So `uint64_t` will be 64-bits on all platforms while `__u64` will vary? Ugh that's some confusing naming. I figured that had to be the case but couldn't believe it given their naming.

@alexcrichton This brings up a good point though with typedefs, what should I be using for datatypes here? `__u64` and family is defined on Linux, but not android. Should I be trying to use the exact same types with all the typedef facade (define `__u64` and friends and use them), should I use equivalent types (`c_ulonglong` instead of `__u64`), or should I use explicit types (`u32` and `u64`) depending on the platform?

Susurrus:
BTW @ndusart Thanks for looking at this!

ndusart:
Explicit types will not work either as [`ctest` translate them to the `stdint` types](https://github.com/alexcrichton/ctest/blob/f574bfb76096de30597089a1e9b6c8c69f5a6454/src/lib.rs#L825-L849), so you'll have to same behavior as using `uint64_t`.

I didn't look if `__u64` was defined in this crate, It's probably better to define `__u64` in android as well then.

alexcrichton:
@Susurrus I'd only require a type with the same size/align/sign, other than that I don't think it's necessary to strive to have the same exact type as C, there's no way that we're "correct" in all other structs.

Susurrus:
> No, __u64 should be 64-bits on all platforms. It's just that one decided to use unsigned long long on all platforms while the other chose to use unsigned long on 64 bits. These are different C types.

Ahh, I think I got it! While the bit-width is the same the types are different and the compiler complains. Makes sense.

> @Susurrus I'd only require a type with the same size/align/sign, other than that I don't think it's necessary to strive to have the same exact type as C, there's no way that we're "correct" in all other structs.

Alright, I did it the minimally-invasive way, using `c_ulonglong` directly instead of defining the `__u64` type. Looks ugly, but it should be correct.

alexcrichton:
Looks like CI is still red?

Susurrus:
@alexcrichton Fixed.

PR #672, opened 2017-07-17
Add syscall table for x86 and x86_64

marmistrz:
I'm wondering if we should use `SYS_write` or `__NR_write`. Both are available to C programs.


alexcrichton:
Thanks! Looks like CI may be failing? 

I don't mind which is added, we can always add both eventually!

marmistrz:
I think we should stick with one set, two sets will unnecessarily enlarge the binary. Let's stick with `SYS_*`, since it's what I've seen in most tutorials.

Well, it didn't like a `pub const` after `extern fn`. Fixed that (I think :))

alexcrichton:
Looks like CI is still failing?

ndusart:
@marmistrz @alexcrichton In linux, [`syscall`](https://github.com/rust-lang/libc/blob/c00686f3a22e419ddf13ab01db99ce828c5b74b5/src/unix/notbsd/mod.rs#L900) expect a `c_long`
We should rewrite all these constants to be of type `c_long`.

nix crate cannot be built for these targets anymore

alexcrichton:
Oops sorry I missed that breakage! I'll send a PR.

marmistrz:
I wasn't fast enough to fix it myself :)

ndusart:
Thanks for the quick fix !

PR #673, opened 2017-07-17
Add all current fallocate flags

g2p:
Also move them to the common Linux module, as they are not platform-specific.
See include/uapi/linux/falloc.h.

### DIFF
--- a/ci/docker/aarch64-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/aarch64-unknown-linux-gnu/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.10
+FROM ubuntu:17.10
 RUN apt-get update
 RUN apt-get install -y --no-install-recommends \
   gcc libc6-dev ca-certificates \
--- a/ci/docker/arm-unknown-linux-gnueabihf/Dockerfile
+++ b/ci/docker/arm-unknown-linux-gnueabihf/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.10
+FROM ubuntu:17.10
 RUN apt-get update
 RUN apt-get install -y --no-install-recommends \
   gcc libc6-dev ca-certificates \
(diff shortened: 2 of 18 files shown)


g2p:
The Travis failure is because Travis doesn't have sufficiently recent headers.
IIRC the linux-libc-dev should be upgraded.

alexcrichton:
Unfortunately this can't land until Travis is green

g2p:
I'll send another pull request to upgrade linux-libc-dev.

g2p:
See https://github.com/rust-lang/libc/pull/676

alexcrichton:
It looks like 17.10 is still failing?

alexcrichton:
Looks like there's still CI failure?

g2p:
Android breaks with the base image upgrade, I'll revert that.

It appears musl brings its own kernel headers unfortunately.
On x86_64 and i686 they come from https://github.com/sabotage-linux/kernel-headers which is unmaintained since 2014, and on mips and mipsel they come from an OpenWRT archive.

ndusart:
@g2p For Android, the headers are not taken from the glibc of the host but the headers in the NDK toolchain that is built during the test.
So, effectively, the version of the host does not matter much. As long as it can build the toolchain and the emulator, it's ok.

I don't quite understand why it fails here, but yes, as it worked before, just let the older version for Android dockerfiles.

g2p:
I have the Android fix locally, I'm just not sure how to proceed with musl.
Either I exclude newer fallocate flags from the musl version, which is stupid but will make the tests pass, or I need someone with far more expertise than I have to find a set of headers that doesn't break everything.
I guess I'll push what I have so people can comment.

alexcrichton:
If these are kernel-defined constants anyway then it seems fine to add a clause to `libc-test/build.rs` with a comment to ignore the constants on musl.

g2p:
I didn't know you could do that…
Updated to ignore the relevant flags in build.rs when musl is used.

g2p:
All green!

PR #676, opened 2017-07-18
Upgrade the glibc toolchain to get newer headers

g2p:


g2p:
All relevant targets are green.

alexcrichton:
Can you describe what this is doing? I'm not sure what these ppa commands are personally.

g2p:
It adds a newer version of the glibc toolchain, including newer linux headers.
This fixes missing C defines in #673; current tests require all flags defined in Rust to be defined in C headers.
The packages come from Canonical: https://wiki.ubuntu.com/ToolChain#PPA_packages

alexcrichton:
Is there documentation for what this is doing? The string "artful" doesn't show up in the link you gisted. Also can this be combined with #673 instead of separate to ensure it actually fixes the errors seen?

g2p:
Artful is Ubuntu 17.10.
I'm adding the repository as a PPA, so that its GPG key is registered, then I add it as a full Apt line so that the Ubuntu 17.10 headers (for Linux 4.11) are used everywhere.

g2p:
I've rebased #673 on top of this to make a combined pull request.

alexcrichton:
Could the docker images just be changed to `FROM ubuntu:17.10`?

g2p:
I guess. I'll try that.

g2p:
The base image upgrade is in https://github.com/rust-lang/libc/pull/677

alexcrichton:
Can these changes happen directly on https://github.com/rust-lang/libc/pull/673?

g2p:
Sorry. Done.

PR #680, opened 2017-07-19
Restore SYS_* constants to the c_long type

alexcrichton:
cc https://github.com/rust-lang/libc/pull/672#issuecomment-316406664


marmistrz:
lgtm ;)

PR #685, opened 2017-07-20
Actually test the targets specified on Travis

retep998:
Today we discovered that libc is actually broken on some targets and CI never actually tested those targets correctly.


clarcharr:
I'm taking this commit, printing it out, and framing it on my wall.

Others:
The Travis failures look intentional, ie. they are failures that are only caught because of this ci fix.

alexcrichton:
I'm going to close this due to inactivity, but feel free to resubmit with tests fixed!

PR #689, opened 2017-07-21
unix: add fexecve()

lucab:
Add missing fexecve() - execute program specified via file descriptor.

Reference: http://pubs.opengroup.org/onlinepubs/9699919799/functions/fexecve.html

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -457,7 +457,8 @@ fn main() {
             "execv" |       // crazy stuff with const/mut
             "execve" |
             "execvp" |
-            "execvpe" => true,
+            "execvpe" |
+            "fexecve" => true,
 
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
@alexcrichton: This PR isn't working.

lucab:
It seems to be complaining about pointer type. The generated test is as follow:
```
int (*__test_fn_fexecve(void))(int, const char* const *, const char* const *) {
  return fexecve;
}
```

while the real signature should be `(int fd, char *const argv[], char *const envp[])`.

However the Rust signature matches the pattern from `execve()`. I see that testing `exec*()` is disabled because of `// crazy stuff with const/mut`. Should this one also be excluded from tests?

alexcrichton:
Yeah seems good to me!

lucab:
Rebased, test is now disabled.

PR #698, opened 2017-07-25
change type of INADDR constants to in_addr_t

ndusart:
These should be `in_addr_t` (u32) instead of `c_int`

This is a breaking change but it is just breaking newlib platforms for now.

@alexcrichton Do you prefer that I split the definition in submodules, to keep `c_int` in newlib and update the others ?


alexcrichton:
@bors: r+

Nah looks fine to me!

PR #686, opened 2017-07-21
Add offsets in user_regs_struct from sys/reg.h

marmistrz:

Issue #665, opened 2017-07-12
AF_MAX/PF_MAX incorrect on local system


Susurrus:
I'm running x86_64 Linux and when running `libc-test` I the following:
```
bad AF_MAX value at byte 0: rust: 42 (0x2a) != c 43 (0x2b)
bad PF_MAX value at byte 0: rust: 42 (0x2a) != c 43 (0x2b)
```

I don't remember getting these in the past, but I just upgraded to Fedora 26, so maybe newer libs changed this constant. Tests are also still passing in CI...

marmistrz:
The same for me, Arch Linux, glibc 2.25

marmistrz:
glibc 2.25 has added Qualcomm IPC Router and had to increment `AF_MAX/PF_MAX`:
https://fossies.org/diffs/glibc/2.24_vs_2.25/sysdeps/unix/sysv/linux/bits/socket.h-diff.html

Susurrus:
I think this ties into how does `libc` deal with API changes between OS versions. There's some work to be done to figure that out, and it should include things like this.

alexcrichton:
We should probably just ignore these constants in libc-test, the values don't seem too critical for keeping correct

Susurrus:
I was going to suggest they might just be worth removing, but that doesn't seem like the right answer either.

marmistrz:
Maybe we should instead generate such files during build based on the libc that is currently installed? It can possibly be a problem when rust-libc exports some constants libc doesn't.

Susurrus:
Digging around the internet more, I'm inclined to say that `AF_MAX` and `PF_MAX` should just be removed. I think this would need to wait to 1.0 to do given the backwards compatibility guarantees provided by `libc`, but I can't find any real code that uses `AF_MAX` and PF_MAX`.

In `nix` we don't expose these constants in our socket API which makes me even more inclined to say that these constants are unused.

I searched through all Rust projects on GitHub and in the first 20 pages couldn't find a use for `AF_MAX` that wasn't just declaring it.

mati865:
Since https://github.com/rust-lang/libc/pull/904 tests now skip these 2 values.
Confirmed to pass tests on openSUSE Tumbleweed (rolling release distro with glibc 2.27 right now).

gnzlbg:
I've submitted a PR that deprecates these #1375 . After a couple of releases, if nobody complains, we can remove these.

gnzlbg:
[...]

PR #699, opened 2017-07-25
Port #672 and #686 to musl.

marmistrz:


alexcrichton:
Looks like CI may be failing?

marmistrz:
It looks like offending syscalls were added in musl 1.16, you're using 1.15. 
http://www.openwall.com/lists/musl/2016/12/25/2
What about updating musl on your side?

alexcrichton:
Oh certainly! You can update the various scripts in `ci/docker` for the musl builders

marmistrz:
It's looks like we've run into #665 on musl too:
```
bad AF_MAX value at byte 0: rust: 42 (0x2a) != c 43 (0x2b)
bad PF_MAX value at byte 0: rust: 42 (0x2a) != c 43 (0x2b)
```

alexcrichton:
Want to update them?

marmistrz:
It's not so easy. These constants are defined in `src/unix/notbsd/linux/mod.rs`. This means that updating them will break the `glibc` CI, since Ubuntu 17.10 still has glibc 2.24. We would need to have glibc 2.25 in our glibc CI.

alexcrichton:
Oh yeah that's fine, want to duplicate the definition and push them down the module hierarchy?

marmistrz:
The same will have to be added on x86 and x86_64 for Android. 
Maybe we should rethink the whole thing? There'll be three copies of the same syscall table which is Linux-specific. 

Maybe it would be better to simply generated them based on the installed libc?

alexcrichton:
Duplication is fine, we're not going to be generating the libc crate based on what was found at compile time

marmistrz:
Why not?

alexcrichton:
Sorry this isn't really quite the PR to debate this on, that question is much broader than just this. If you'd like to land this then the best way will be to duplicate the definitions I believe, otherwise if you'd like to wait and hold off we can continue discussion on perhaps a dedicated issue.

marmistrz:
So let's land it with duplication now and discuss it in a separate issue.

PR #700, opened 2017-07-25
Use cfg_attr to gate musl linking

### LINKED ISSUE #684: Won't compile on musl without libstd
See the linked thread at rust-lang/rust#37406.

I tried compiling my `loop-forever` crate on musl to reduce binary size, but it failed to compile due to an error that appears to be a bug; the `link_cfg` feature should never be called outside of the std build.

Others:
Fixes #684. Previously building libc for musl would fail, because we don't want to use #[link(...)] outside a stdbuild. This commit fixes that issue. (Although it may leave other targets broken, I'm not sure.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ services:
 install:
   - if [ -z "$NO_ADD" ]; then rustup target add $TARGET; fi
 script:
-  - cargo build
-  - cargo build --no-default-features
+  - cargo build --target=$TARGET
+  - cargo build --no-default-features --target=$TARGET
   - cargo generate-lockfile --manifest-path libc-test/Cargo.toml
   - if [[ $TRAVIS_OS_NAME = "linux" ]]; then
       sh ci/run-docker.sh $TARGET;
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -236,8 +236,8 @@ cfg_if! {
         // cargo build, don't pull in anything extra as the libstd  dep
         // already pulls in all libs.
     } else if #[cfg(any(all(target_env = "musl", not(target_arch = "mips"))))] {
-        #[link(name = "c", kind = "static", cfg(target_feature = "crt-static"))]
-        #[link(name = "c", cfg(not(target_feature = "crt-static")))]
+        #[cfg_attr(stdbuild, link(name = "c", kind = "static", cfg(target_feature = "crt-static")))]
+        #[cfg_attr(stdbuild, link(name = "c", cfg(not(target_feature = "crt-static"))))]
         extern {}
     } else if #[cfg(target_os = "emscripten")] {
         #[link(name = "c")]
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -232,16 +232,12 @@ pub const INADDR_NONE: in_addr_t = 4294967295;
 cfg_if! {
     if #[cfg(dox)] {
         // on dox builds don't pull in anything
-    } else if #[cfg(all(not(stdbuild), feature = "use_std"))] {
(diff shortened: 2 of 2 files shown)


alexcrichton:
Looks like CI is failing? Can you also include tests from https://github.com/rust-lang/libc/pull/685?

Others:
@alexcrichton I'm already seeing more failures on travis, even with the style fix. How would you feel about gating all the `link` commands behind `stdbuild`? One way of doing that would be changing:
```
else if #[cfg(all(not(stdbuild), feature = "use_std"))] {
```
To:
```
else if #[cfg(any(not(stdbuild), feature = "use_std"))] {
```

Others:
Alternatively I can just introduce a fix for the [rumprun target](https://travis-ci.org/rust-lang/libc/jobs/257762233) (which looks to be the only broken one), and leave the rest as-is.

alexcrichton:
Ah yeah it'd be great to add this same fix there as well!

Others:
@alexcrichton Okay, I've made the CI build work for all platforms. I did, however, use the first method I outlined and prevent any linking from happening if `not(stdbuild)`. I think that the logic is cleaner this way--and could prevent future issues. 

If you agree with the above assessment, then I think this is ready to be merged.

alexcrichton:
Hm yeah so the only gotcha there is that if you pull in the `libc` crate without the standard library you may get unresolved symbol errors. Is that the use case you have in mind or is libstd still pulled in elsewhere?

Others:
@alexcrichton Yeah, you're right. Fixing it this way won't work. But I don't think the earlier `#[cfg_attr(...)]` fix works either. (Since it prevents the necessary libc.a from being pulled in on musl if you aren't using the standard library.)

I'm not sure how to proceed here. Do you know of any other options we have?

alexcrichton:
It's true yeah, we could avoid the `cfg` inside the `link` attribute perhaps though? Basically have a `#[cfg_attr(not(stdbuild), link(...))]`?

Others:
@alexcrichton How can we avoid the `cfg` inside the link? Don't we want different behavior depending on whether `target_feature = "crt-static"`?

alexcrichton:
Yeah we wouldn't get precisely the same behavior, we'd have to select one kind of linakge for musl, for example.

Others:
@alexcrichton Do you have any thoughts on what we should do then? #684 is a serious problem for using Rust on a MUSL platform. If there is no quick fix for now, perhaps we should introduce some sort of work around?

alexcrichton:
I think our only option is to have `#[link(name = "c")]` and pray it works.

clarcharr:
Is there any progress on this? I'd be willing to help out with a fix but I'm not sure where to go from here.

Others:
@clarcharr I'm still paying attention to this, although I've been busy at work this past week. I'm dissatisfied with the current solution / every solution outlined thus far. I feel like the `#[link(name = "c")]`approach is likely to break `-crt-static` linking (which I personally depend on). @alexcrichton do you have further thoughts on this? Do you want me to just forge ahead with the `#[link(name = "c")]` and pray strategy?

alexcrichton:
Ideally this would look like:

```rust
#[cfg_attr(all(not(stdbuild), target_feature = "crt-static"), link(name = "c", kind = "static")]
#[cfg_attr(all(not(stdbuild), not(target_feature = "crt-static")), link(name = "c")]
```

But right now `target_feature = "crt-static"` is unstable to match against. Short of that I'm not entirely sure what to do.

alexcrichton:
I'm going to close this due to inactivity, but feel free to resubmit with a rebase!

PR #704, opened 2017-07-30
Change TIO* constants to c_ulong, as ioctl() requires c_ulong

mneumann:
This for example fixes code like:

https://github.com/a8m/pb/blob/master/src/tty/unix.rs#L22

where we would need an "as c_ulong" otherwise.


alexcrichton:
Unfortunately this is a breaking change for now so I'll tag this as 1.0 cleanup

mneumann:
The problem is that for some operating systems these constants are already defined in (rusty) libc as ```u_long```, whereas for some others these are (incorrectly) defined as ```c_int```. Many libraries are only tested on Linux, where these constants are defined as u_long, so these libraries are broken on other OSes like FreeBSD or DragonFly. This leads to a lot of work to fix these small things. Even worse, this leads to code like ```TIOCGETD as c_ulong``` all over the place in these libraries, which is only correct as long as ```ioctl()``` really is ```c_ulong```. 

I assume a 1.0 release is light years away. Having things broken on BSDs is bad IMHO (it frustrates me to spend so much time fixing things like that :(). Can we at least introduce a ```type ioctl_fun = c_ulong```, so that other libraries can safely cast any TIO* constant (any ioctl() request code) to the correct type?

alexcrichton:
Oh I agree it's frustrating! On the other hand though this is a breaking change. I hope that 1.0 is not lightyears away, though. Hopefully just on the order of months.

mneumann:
@alexcrichton: Thanks! I understand it's a breaking change, but I'd argue that it's currently broken on these platforms as dependent crates won't currently build on Free- and DragonFly BSD. A few months I can wait :)

asomers:
It's not a breaking change if it only affects code that doesn't compile as-is.  Can we please have this pre-1.0, unless somebody's aware of code that is using the incorrect types yet somehow actually works?

alexcrichton:
@asomers code which works around these being the wrong types will break if this lands, and it's definitely possible to write working code today that breaks if these types change

mneumann:
@asomers @alexcrichton I think code that works around this will do an "as u32" cast or an "into()", so whatever type libc exports, it will not break anything.

alexcrichton:
This breaks code like `let x: c_uint = TIOCEXCL` because the type changes

mneumann:
@alexcrichton : Correct, that would break this code. But, that code would be broken anyway, as `x` would be passed to `ioctl`, which expects an `u_long` ;-). Well, I can live with the workaround, as it doesn't involve much code, but on the other hand, this workaround will very likely never be removed from the code :). A more general question arises, how to proceed with bugs in the API (and this is one). Imagine that we would have exposed `ioctl` with an invalid signature... would we change it and break the API?

alexcrichton:
Sorry but we're just not accepting breaking changes right now, if and when 1.0 happens we can change this but until that time FreeBSD is a "high enough tier" that we don't accept breakage for it.

asomers:
Speaking as a FreeBSD committer, the potential breakage that would be caused by merging this PR is less bad than the actual breakage that is caused by not merging it.  This bug breaks a lot of crates that were originally developed on Linux when they try to go to FreeBSD.

alexcrichton:
If a crater-like run could be performed showing that this doesn't have any practical breakage in the ecosystem on FreeBSD then I think we would be able to make this change, but without that knowledge we simply cannot make breaking changes blindly

Mark-Simulacrum:
Would it be sufficient to patch libc with this merged and then run Crater in check-only mode targeting FreeBSD? Or is more than that needed?

alexcrichton:
I believe so yeah, but I don't think there's a lot of precedent for doing this on non-Linux platforms

gnzlbg:
@mneumann could you rebase this on top of master ?

While this only affects FreeBSD, the APIs it touches are not super rare. I'd like to merge this tentatively and do a release that only contains this change, and then, update libc in rust-lang/rust with this change. 

If nobody complains for a reasonable amount of time, we can assume that this did not broke anybody. Otherwise, we'll have to yank that libc release, and revert this change from master, but we can hopefully do that during the nightly cycle as usual.

asomers:
@gnzlbg isn't this what crater is for?

gnzlbg:
@asomers crater can check whether a rust-lang/rust PR breaks something, so when we upgrade libc in rust-lang/rust, we could do a crater run to check that. However, AFAIK, crater cannot be used yet to check whether a PR to a crate on crates.io would break its dependencies. 

IIRC, using the bundled libc actually requires a nightly feature (`feature(libc)`), so I don't know how many crates are doing that instead of using libc from crates.io, so I don't know how useful the crater run would be. We should do this anyways though, but it might well be that doing a crates.io release breaks users that the crater run won't detect.

PR #709, opened 2017-08-01
Add setresgid & setresuid to FreeBSD and OpenBSD

wezm:
These functions are both present in [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=setresuid&apropos=0&sektion=0&manpath=FreeBSD+11.1-RELEASE+and+Ports&arch=default&format=html) and [OpenBSD](http://man.openbsd.org/setresuid).


alexcrichton:
@bors: r+

Thanks!

PR #711, opened 2017-08-01
Add BOTHER & termios2 to Android & Linux

Susurrus:
This should be the same for both the struct and the constant across all Linux/Android platforms, but we'll see!

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -377,6 +377,10 @@ fn main() {
             // FIXME: unskip it for next major release
             "stat" | "stat64" if android => true,
 
+            // These are tested as part of the linux_fcntl tests since there are
+            // header conflicts when including them with all the other structs.
+            "termios2" => true,
+
             _ => false
         }
     });
@@ -671,8 +675,7 @@ fn main() {
     // fails on a lot of platforms.
     let mut cfg = ctest::TestGenerator::new();
     cfg.skip_type(|_| true)
-        .skip_struct(|_| true)
-        .skip_fn(|_| true);
+       .skip_fn(|_| true);
     if android || linux {
         // musl defines these directly in `fcntl.h`
         if musl {
@@ -695,8 +698,18 @@ fn main() {
                 _ => true,
             }
         });
+        cfg.skip_struct(|s| {
+            s != "termios2"
+        });
+        cfg.type_name(move |ty, is_struct| {
+            match ty {
+                t if is_struct => format!("struct {}", t),
+                t => t.to_string(),
+            }
+        });
     } else {
         cfg.skip_const(|_| true);
+        cfg.skip_struct(|_| true);
     }
     cfg.generate("../src/lib.rs", "linux_fcntl.rs");
 }
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -489,6 +489,7 @@ fn main() {
             "F_CANCELLK" | "F_ADD_SEALS" | "F_GET_SEALS" => true,
             "F_SEAL_SEAL" | "F_SEAL_SHRINK" | "F_SEAL_GROW" | "F_SEAL_WRITE" => true,
             "QFMT_VFS_OLD" | "QFMT_VFS_V0" | "QFMT_VFS_V1" if mips && linux => true, // Only on MIPS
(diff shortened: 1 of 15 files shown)


alexcrichton:
A whitelist is added for this but is it verified anywhere?

Susurrus:
What do you mean verified? I looked through the Linux sources to figure out these values and they should be correct.

alexcrichton:
Oh I just mean verified in the sense of libc-test/build.rs. If it's ignored there on all linux platforms, and the value varies across architectures on linux platforms, how do we know it's right?

Susurrus:
@alexcrichton I looked at the Linux sources to figure out what these values are. For situations like this, what is the bar for proving this is correct? I haven't had to do it for `CMSPAR` when I added it, so I'm not certain what you're looking for.

alexcrichton:
Constants like CMSPAR are only ignored on mips-unknown-linux-gnu it looks like, meaning it's verified on almost all platforms. BOTHER, however, is validated nowhere.

Can this constant not be accessed from C code? If not why not, and how is it supposed to be used in C? If so, how come we can't access it that way?

Susurrus:
BOTHER is only defined in `asm/termbits.h`, which cannot be included along with `termios.h`. To test this we'd need a separate test that only included this header to verify the value. There are various posts on the internet that ask how you're supposed to use the termios2 API since it conflicts with termios1, and the response is normally to inline the necessary datatypes to work around the conflict.

Susurrus:
I'd also like to include `NCCS` and `struct termios2` from the Linux `asm/termbits.h` header, but I'm uncertain how to do this. `NCCS` is already defined in `termios.h` and is a different value between these two headers on most platforms. There's no `const_name` method within `ctest` to allow setting it to `NCCS_TERMIOS2` in the source and still allowing it to be translated to just `NCCS` for testing.

I could for now at least support the struct without this constant, though that's definitely less than ideal.

Susurrus:
Note that this branch is now based on #720 as that's needed to implement testing.

alexcrichton:
Sure yeah sounds reasonable, I'd probably just not expose the constant if possible

Susurrus:
Just rebased and updated this now that #720 landed. One thing I wasn't able to do was to test the `VINTR` and related constantsto check they have the proper values in both the `linux_fcntl` and `main` test binaries. It looks like those binaries are linked together as part of the final test suite, which means there are linking errors about test functions when the same constants are tested in both binaries. These should be the same, but I thought an extra check would be nice so I wanted to mention this issue.

Susurrus:
I'm not certain why there're failures on `mips` targets, because the declaration of `termios` follows pretty closely to `termios2`, which passes all CI tests. I think it might be in the type name mapping, which i don't really understand. Any idea @alexcrichton?

libc-test/build.rs
```diff
@@ -653,9 +659,18 @@ fn main() {
     // fails on a lot of platforms.
     let mut cfg = ctest::TestGenerator::new();
     cfg.skip_type(|_| true)
-        .skip_struct(|_| true)
+        .skip_struct(move |name| {
+            match name {
+                // The termios2 struct is the same as `termios` on powerpc, so it's actually
+                // undeclared in the headers. We'll still expose it there to still provide
+                // some type safety, but we can't check it directly.
+                "termios2" if !powerpc => false,
```
alexcrichton: Could this use `struct_name` to rename the struct to assert `termios2` is the same as `termios` on powerpc?

alexcrichton:
This just looks like `c_cc` has a different type in the struct than is defined in Rust?

alexcrichton:
I'm gonna close this due to inactivity, but feel free to reopen with a rebase!

Susurrus:
Looks like I can't reopen this PR, but I'm definitely interested in merging it.

Susurrus:
@alexcrichton This crashes for me with really weird errors when trying to rename `termios2` to `termios` on `powerpc` platforms, but I figured I'd submit it, see what CI says, and maybe get some help.

Susurrus:
Looks like a spurious failure on `aarch64-unknown-linux-gnu`, so if this looks good should be fine to merge.

PR #715, opened 2017-08-07
Add some permission and misc (mostly 'mode_t' related) constants to Redox.

goyox86:
Hi!

While porting some software ((exa)[https://github.com/ogham/exa])to Redox I found these missing constants. 

Cheers!

//cc @ids1024 @jackpot51


jackpot51:
@goyox86 The mode_t values should probably be hex

goyox86:
Ah I left them as decimal as @ids1024 mentioned it would be good to leave them like Linux ones. I'm gonna change them then.

goyox86:
Done @jackpot51

jackpot51:
@goyox86 Thanks very much for that!

PR #718, opened 2017-08-10
FD_ISSET does not need to be mutable

Susurrus:


alexcrichton:
Tagging as 1.0-brekage, but it's also worth noting that [this isn't how these functions are prototyped](http://pubs.opengroup.org/onlinepubs/7908799/xsh/select.html).

Susurrus:
@alexcrichton What do you mean? Those functions look correct to me.

alexcrichton:
The prototype is:

```c
int FD_ISSET(int fd, fd_set *fdset);
```

not

```c
int FD_ISSET(int fd, const fd_set *fdset);
```

Susurrus:
Sure, so you're saying all pointer function arguments within libc should be `mut` unless they've been explicitly labeled as `const`? So even though in this case this function shouldn't modify the argument, we can't assume that since the original API didn't declare that?

alexcrichton:
What I'm saying is that we're not the one making the API decisions here, libc is. These are macros in libc so all we have to go off is the opengroup specification, which is what these APIs *would* look like (in theory) if they were actual functions.

Susurrus:
Roger dodger. Then let's go ahead and close this.

PR #719, opened 2017-08-10
Add statvfs constants for Linux/musl targets

Susurrus:


Susurrus:
Whoops, looks like this didn't need to be done for `musl` targets. Misread some errors on my end I guess.

PR #720, opened 2017-08-10
Add more fcntl and seal constants for Android/Linux

Susurrus:


Susurrus:
@alexcrichton I wanted to ask for your help here. The `linux/fcntl.h` and `fcntl.h` headers cannot be included simultaneously, so to test them I generated an additional binary just to test `linux/fcntl.h` includes. The file generation is working, but I'm not certain what the best way to hook things up in CI is. I think the best solution here is to generate this test executable for all platforms only including the appropriate constants on the correct systems and then have CI run both the main executable `libc-test` and also `linux_fcntl` on all platforms. This should just take some modifications to `ci/run.sh` if I'm understanding that correctly, is that what I should do? I want to make sure this is extensible because I plan on doing something similar to resolve #711.

alexcrichton:
Yeah this seems pretty reasonable to me! To suggest a way to simplify this, though, perhaps:

* Instead of binaries, these could be `tests/*.rs`-style tests that are listed as `harness = false`
* We could use Cargo's "test runner" support to specify how to execute binaries

I *think* if we do that it means that `cargo test` in the `libc-test` folder should "just work"?

Susurrus:
@alexcrichton I implemented that and it's definitely a lot nicer. This also sets us up nicely for running additional specific tests (like for #711) for all platforms.

There are still build failures on OpenBSD, but that's a really complicated setup. Any way I could get some help debugging it? Comparing [a good run](https://travis-ci.org/rust-lang/libc/jobs/263533684) and [my recent failed run](https://travis-ci.org/rust-lang/libc/jobs/263663237) didn't reveal much. I think `OPT_LEVEL` being undefined may be part of the problem, but I don't see that anywhere in the project.

ci/run.sh
```diff
@@ -95,12 +96,14 @@ fi
 
 case "$TARGET" in
   *-apple-ios)
-    cargo rustc --manifest-path libc-test/Cargo.toml --target $TARGET -- \
-        -C link-args=-mios-simulator-version-min=7.0
+    cargo rustc --manifest-path libc-test/Cargo.toml --target $TARGET \
+        --test main -- -C link-args=-mios-simulator-version-min=7.0
+    cargo rustc --manifest-path libc-test/Cargo.toml --target $TARGET \
+        --test linux-fcntl -- -C link-args=-mios-simulator-version-min=7.0
```
alexcrichton: It may be worth testing out `RUSTFLAGS` nowadays. I think at the time this code was written `RUSTFLAGS` didn't exist but nowadays I think it'd suffice for this, allowing us to basically just use `cargo build --tests`

alexcrichton:
Ah drat I thought we removed some of these pieces awhile ago...

The OpenBSD testing is way crazier than everything else because it generates tests on the host and compiles them on the target (in the emulator). I think we should just jettison all that though, it's crazy complicated and not really worth it :(

semarie:
the OPT_LEVEL problem isn't just that cargo in the image is out-of-date ? the used image is OpenBSD 6.0 so cargo should be 0.10.0.

I could try to update the image to 6.1 (latest stable). It will be cargo 0.16.0 and rustc 1.16.0.

@alexcrichton feel free to ping me for OpenBSD related issues.

alexcrichton:
@semarie my worry isn't necessarily exactly what's happening here but rather that the mechanism of testing OpenBSD is pretty crazy in this repo, and I think it's not pulling its weight any more. All other platforms cross-compile from linux and then we figure out how to emulate the binary, but we can't cross-compile to OpenBSD which makes it too hard to get a program running in there.

Susurrus:
Been a few days on this, do we know what direction we want to go with this? I'm willing to help push this forwards, but it looks like we are at an impasse in terms of direction here.

alexcrichton:
I think let's just remove OpenBSD support for now, and we can continue to shim how the other platforms are tested.

alexcrichton:
I've actually gone ahead and removed OpenBSD from CI as we needed to do it anyway to get CI working again

Susurrus:
Sounds good. Hopefully OpenBSD gets some official rustup/cargo support (whatever it is that's missing) so we can get better CI support for it.

Anyways, I updated this PR and it passed all tests.

ci/run.sh
Susurrus: I've fixed this on the latest build, but it failed on i386-ios because there were two `main-*` binaries. Which is weird, but maybe it's a caching issue. I'm not too certain what the caching policy is when set to `cargo` on Travis, but maybe that'll need to be turned off? Any other ideas here?

alexcrichton:
To avoid duplication of all the various test invocations, can this use the `runner` support now present in Cargo to override how tests are executed?

Susurrus:
Is that different from the `harness = false` I already set?

alexcrichton:
Oh it's actually different from that (we'll want that as well though). This is related to the `runner` configuration on [this page](http://doc.crates.io/config.html)

Susurrus:
Any way we can leave that out of this PR? I'm already a little out of my understanding and I don't really understand what you're asking me to do.

Susurrus:
I realized after thinking about this for a bit that this is the correct fix, as cargo knows the exact names of the executables and it solves our ios problem. I have a version here that uses a custom runner in CI for non-Android/NetBSD targets. This is because those targets use a simulator where executing the binary and receiving the input is not trivial.

Susurrus:
I think the solution to the Android and NetBSD targets is something that looks very similar to the iOS program at `ci/ios/deploy_and_run_on_ios_simulator.rs`

alexcrichton:
Ah sorry I started working on this locally and didn't get a chance to push it up. This wasn't quite what I had in mind as I personally like to isolate all target-specific information as much as possible.

I believe I've handled everything except Android, which I'll try to get to soon but if you'd like to try to tackle it then it should look similar to the rumprun solution.

Susurrus:
I'll try to hack at this tomorrow for Android and see what I can come up with.

alexcrichton:
@bors: r+

Let's see if that works...

Susurrus:
It doesn't seem like bors is doing anything as Travis doesn't have any active builds. This applies to #737 as well.

Susurrus:
Failed due to timeout, but at least Travis has recently fixed the issues with the Mac builds, but there still is a massive backlog.

Susurrus:
This failed again on Travis with spurious errors on the Mac builders. Travis's backlog is getting down to a reasonable number, so trying this again will likely be successful.

Susurrus:
Omg it finally happened! Travis really needs to up their build slaves for Mac.

Anyways, thanks for staying on this @alexcrichton!

PR #727, opened 2017-08-13
skip siginfo_t.si_addr type check on OpenBSD

semarie:
the type changed from *c_char to *c_void in 6.1


alexcrichton:
@bors: r+

Thanks!

PR #728, opened 2017-08-13
Openbsd kqueue

semarie:
The PR supersede #613 to add exception on the testsuite.


semarie:
@alexcrichton the homu fail is a timeout on i386-apple-ios.

PR #731, opened 2017-08-17
Make fd_set.fds_bits public

### LINKED ISSUE #730: fd_set.fds_bits is not public
Was there a reason for this? I know that there are `FD_CLR` and friends, but it seems odd that this one field is private while almost all struct fields for all `libc` types are public.

Susurrus:
Closes #730.

### DIFF
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -64,10 +64,10 @@ s! {
     pub struct fd_set {
         #[cfg(all(target_pointer_width = "64",
                   any(target_os = "freebsd", target_os = "dragonfly")))]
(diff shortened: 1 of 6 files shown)


alexcrichton:
CI is failing?

Susurrus:
Interesting. I'm not really certain what that error is. Looks like a bug in `ctest`? I'll try to dig into it some more as I can reproduce it locally at least.

Susurrus:
This looks to be a parsing error within `libc-test`. Looks like it doesn't like expressions in the length part of an array declaration, but this only applies to `pub` fields. Maybe I'll just close this since the `FD_*` macros should take care of this anyways.

alexcrichton:
Up to you!

PR #734, opened 2017-08-17
L4Re improvements

humenda:
This commit improves the x86_64 uClibc port and splits of L4Re-specific structs into a separate submodule. It defines additional constants and fixes some unfinished structs.

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -230,8 +230,10 @@ pub const INADDR_BROADCAST: in_addr_t = 4294967295;
 pub const INADDR_NONE: in_addr_t = 4294967295;
 
 cfg_if! {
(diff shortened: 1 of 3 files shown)


src/unix/mod.rs
```diff
@@ -230,7 +230,7 @@ pub const INADDR_BROADCAST: in_addr_t = 4294967295;
 pub const INADDR_NONE: in_addr_t = 4294967295;
 
 cfg_if! {
-    if #[cfg(dox)] {
+    if #[cfg(any(dox, target_os = "l4re"))] {
```
alexcrichton: Could you separate this out explicitly to a separate clause below? (just to be clear that this branch is the "doc only" branch.

PR #747, opened 2017-08-30
Add QFMT_VFS_V1 on MIPS and test all QFMT_ constants

Susurrus:


Susurrus:
Cool, looks like this should be correct as tests are all passing!

PR #749, opened 2017-08-30
apple: Add VM_* constants from mach/vm_statistics.h

### LINKED ISSUE #736: Add VM_FLAGS_* constants on Darwin
These flags are needed for using superpages on Darwin with `mmap`, and are defined in `mach/vm_statistics.h`.

joshlf:
Closes #736.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,8 +38,10 @@ matrix:
     - env: TARGET=i686-unknown-linux-gnu
     - os: osx
       env: TARGET=x86_64-apple-darwin NO_ADD=1
+      osx_image: xcode8.3
     - os: osx
       env: TARGET=i686-apple-darwin
+      osx_image: xcode8.3
     - env: TARGET=arm-linux-androideabi
     - env: TARGET=aarch64-linux-android
     - env: TARGET=i686-linux-android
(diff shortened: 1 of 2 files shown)


joshlf:
@alexcrichton Any advice on what I should do about the errors that some of the constants don't exist in the version of libc that Travis is using? They exist in [my laptop's copy of `mach/vm_statistics.h`](https://gist.github.com/joshlf/d552f4804b0185eb2659e747ac71b6fe).

alexcrichton:
Is a header missing? Or maybe Travis's version is too old? Can travis's version be newer?

joshlf:
My guess is that a header is missing, yeah. I'm not sure when those variables were added, but my `vm_statistics.h` lists the copyright as "2000-2016", so maybe if we get the 2016 version, we'd be OK. I'll look into whether that's possible with Travis.

joshlf:
OK, it looks like we're currently building with the Travis default of XCode 7.3 and OS X 10.11. The iOS builds, on the other hand, build with XCode 8.2 and OS X 10.12, and those tests pass. I'm going to bump the failing builds to XCode 8.3/OS X 10.12 (which is the latest stable version available) and see if that fixes it.

joshlf:
OK that worked.

PR #750, opened 2017-08-30
Fix dev_t minor() bitmasking on Linux

agrover:
This code appears to be modeled on the macros in glibc bits/sysmacros.h
(since Glibc 2.26). Fix the masking of bits for minor() to match that
implementation, which also corresponds with the explanatory comment in
that file.

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -962,7 +962,7 @@ f! {
 
     pub fn minor(dev: ::dev_t) -> ::c_uint {
         let mut minor = 0;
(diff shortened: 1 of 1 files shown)


agrover:
@sfackler can you take a look? Using this test program:

```
extern crate libc;

fn main() {
    unsafe {
        let dev = libc::makedev(1, 2);

        println!("major {} minor {}", libc::major(dev), libc::minor(dev));
    }
}
```
I've verified that it prints "major 1 minor 0" without the fix (incorrect) but "major 1 minor 2" with the fix (correct).

sfackler:
Oops, this seems right to me!

PR #755, opened 2017-09-03
Add Solaris 10 support

bgermann:
Unlike Illumos and Solaris 11, Solaris 10 does not support posix_memalign, so memalign needs to be added for Solaris 10 support. Additionally add pthread_create to Solaris, which is used in stdlib.


bgermann:
I do not think the tests failed because of this change. It does not even test on Solaris...

PR #756, opened 2017-09-03
Haiku updates

jessicah:


jessicah:
I've been looking at the travis builds, the ones that fail all seem to have `error: missing key: 'url'` in the logs, is that why they're failing?

PR #759, opened 2017-09-09
Add timerfd API on Linux

slyrz:
This change adds the Linux-specific timerfd API to libc.

```C
#include <sys/timerfd.h>

int timerfd_create(int clockid, int flags);
int timerfd_gettime(int fd, struct itimerspec *curr_value);
int timerfd_settime(int fd, int flags, 
                    const struct itimerspec *new_value, 
                    struct itimerspec *old_value);
```

The timerfd API has been available since kernel 2.6.25 and in glibc since version 2.8. 

I'm not sure if I put the changes in the right places, so please review with great care.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -188,6 +188,7 @@ fn main() {
         cfg.header("sys/shm.h");
         cfg.header("sys/user.h");
         cfg.header("sys/fsuid.h");
+        cfg.header("sys/timerfd.h");
         cfg.header("shadow.h");
         if !emscripten {
             cfg.header("linux/input.h");
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -200,6 +200,11 @@ s! {
         _pad: [::uint8_t; 48],
     }
 
+    pub struct itimerspec {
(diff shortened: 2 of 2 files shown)


alexcrichton:
Looks like CI is failing?

slyrz:
The MIPS targets complain about the value of the `TFD_NONBLOCK` constant:

```
bad TFD_NONBLOCK value at byte 2: rust: 8 (0x8) != c 0 (0x0)
bad TFD_NONBLOCK value at byte 3: rust: 0 (0x0) != c 128 (0x80)
```

I could alias `TFD_NONBLOCK` and `TFD_CLOEXEC` to the respective values of `O_NONBLOCK` and `O_CLOEXEC` to fix this:

```rust
pub const TFD_CLOEXEC: ::c_int = O_CLOEXEC;
pub const TFD_NONBLOCK: ::c_int = O_NONBLOCK;
```

But I noticed signalfd/eventfd have their constants (`SFD_CLOEXEC` and `EFD_CLOEXEC`) manually declared inside each architecture, shall I go this route as well?

alexcrichton:
Either way should be fine, so long as CI passes!

PR #762, opened 2017-09-11
Define additional IPPROTO_* constants.

jack-pappas:
Define additional IPPROTO_* constants, e.g. for DCCP and SCTP. These constants are standard across all system types -- the values are those assigned to the protocol by the IANA:
https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -1039,6 +1039,223 @@ pub const MINCORE_MODIFIED: ::c_int = 0x4;
 pub const MINCORE_REFERENCED_OTHER: ::c_int = 0x8;
 pub const MINCORE_MODIFIED_OTHER: ::c_int = 0x10;
 
+//
+// sys/netinet/in.h
+// Protocols (RFC 1700)
+// NOTE: These are in addition to the constants defined in src/unix/mod.rs
+
+// IPPROTO_IP defined in src/unix/mod.rs
+/// IP6 hop-by-hop options
+pub const IPPROTO_HOPOPTS: ::c_int = 0;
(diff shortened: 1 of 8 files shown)


alexcrichton:

jack-pappas:
Yeah -- I'll fix it. These constants are the same across all platforms so I thought it might be fine to just put them all in the base module. libc-test is strict and only allowing the constants defined on each particular platform though, so I'll need to go back and move some of the constants into the platform-specific modules instead. (Will do that when I have a minute then update the PR.)

jack-pappas:
I've updated the PR, and the build succeeds on my machine and passes most of the travis CI builds now. However, a few of the travis builds are still failing with the latest code and I don't see why that's the case. A couple of them appear to be failing in the C compiler tests with a constant redefinition error, any ideas why that's happening?

alexcrichton:
Some of these constants may not be defined on OSX?

jack-pappas:
Yes -- there were a few constants that'd been removed from the OSX / Darwin headers when they were ported over from FreeBSD, and I'd left them in there by accident. There were also a couple of other places where I'd added a duplicate ``IPPROTO_RAW`` constant in. I've fixed these and updated the PR.

One small note -- I had to disable (comment out) the ``IPPROTO_L2TP`` constant for NetBSD. It's defined in the latest NetBSD source, but it looks as though the docker image used by the CI tests is using an older version of NetBSD which doesn't have that constant defined. The constant was added here:
https://github.com/NetBSD/src/commit/7c6df79fc2d2c616a886db2f5e64951d33d313ef#diff-dab66d274da3dd4ba168b2ba7f3b4374R108

alexcrichton:
@bors: r+

Ok, thanks!

PR #764, opened 2017-09-12
Make stdbuild a cargo feature

tamird:
Also invert the default feature / no std arrangement. The default
still uses std, but no_std now requires opt-in instead of opt-out.

This is work toward removing the shim in rustc.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -83,18 +83,18 @@
 ))]
 
 // Attributes needed when building as part of the standard library
-#![cfg_attr(stdbuild, feature(no_std, staged_api, custom_attribute))]
-#![cfg_attr(stdbuild, feature(link_cfg))]
-#![cfg_attr(stdbuild, no_std)]
-#![cfg_attr(stdbuild, staged_api)]
-#![cfg_attr(stdbuild, allow(warnings))]
-#![cfg_attr(stdbuild, unstable(feature = "libc",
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -18,8 +18,8 @@ travis-ci = { repository = "rust-lang/libc" }
 appveyor = { repository = "rust-lang-libs/libc" }
 
 [features]
-default = ["use_std"]
```
alexcrichton: This was intentionally set up like this before (enabling std as a feaeture), could this be reverted?

tamird: Sure. It's a bit more awkward because this setup allows `stdbuild` to depend on `no_std`, which can't be done in reverse, that's all.

tamird: Done.

tamird:
This is green now, with the changes you requested.

Cargo.toml
```diff
@@ -20,6 +20,7 @@ appveyor = { repository = "rust-lang-libs/libc" }
 [features]
 default = ["use_std"]
 use_std = []
+stdbuild = []
```
alexcrichton: Can this be avoided in this Cargo.toml? Ideally this isn't exported as part of the public API

tamird:
I suppose, but won't that mean we can never get rid of the shim in rustc?
Anyway, done.

alexcrichton:
@bors: r+

AFAIK we have no path forward for using the libc crate as-is in rustbuild, we've got a long way to go to enable that.

PR #770, opened 2017-09-19
Define DCCP constants

jack-pappas:
This PR defines constants needed to utilize [DCCP](https://en.wikipedia.org/wiki/Datagram_Congestion_Control_Protocol) in Linux and NetBSD.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -239,6 +239,13 @@ fn main() {
         }
     }
 
+    if linux || android {
+        // DCCP support
+        if !uclibc && !musl && !emscripten {
+            cfg.header("linux/dccp.h");
+        }
+    }
+
     if freebsd {
         cfg.header("pthread_np.h");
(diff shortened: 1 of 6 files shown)


alexcrichton:
@bors: r+

Thanks!

jack-pappas:
I removed redundant definition of a couple of constants from the NetBSD module which was causing some build failures.

On the Linux side -- the C program generated by ``libc-test`` is failing because it can't find the definitions of some constants. The C program would need to include ``uapi/linux/dccp.h`` -- is there a way to tell if it's doing that, and if not, add the header to what it pulls in for the tests?

alexcrichton:
You'll probably need to edit `libc-test/build.rs` to include the various headers I think

jack-pappas:
I've fixed ``build.rs`` to include the needed additional headers and made a few tweaks to the constant definitions, and the changes are passing the CI tests now.

Issue #776, opened 2017-09-21
clock_gettime for macOS 10.12+


ekmartin:
macOS 10.12 and Xcode 8 added support for `clock_gettime` and [related functions](https://linux.die.net/man/3/clock_gettime). Would it be possible to add these to the macOS libc bindings?

I'd send a PR, but not sure how to do it conditionally since it doesn't seem like the OS version number is included in `cfg`.

Relevant: https://github.com/zeromq/libzmq/issues/2175

alexcrichton:
Seems good to me to add! (PRs are more than welcome!)

ekmartin:
@alexcrichton: but would it be okay to add it to the general `apple/mod.rs` even if it's only supported in newer versions of macOS?

alexcrichton:
Yeah that's fine, this isn't the first symbol of its kind :)

PR #777, opened 2017-09-21
Add clock_gettime related functions to macOS

ekmartin:
Fixes https://github.com/rust-lang/libc/issues/776. The constants were retrieved on macOS 10.12.6. Tried adding `clock_nanosleep` too, but it doesn't seem like macOS' libc includes that yet.


ekmartin:
Hm, `clock_settime` isn't available on iOS. @alexcrichton: is there any way to include it only on macOS? The other two functions are available on iOS as well.

alexcrichton:
@bors: r+

Ah ok looks like travis is green so let's try again...

PR #781, opened 2017-09-25
Haiku fixes for nix

jessicah:


alexcrichton:
Looks like there are CI failures?

jessicah:
Ah, this is because of `#[cfg(not(target_os = "haiku"))]` in unix/mod.rs.

I guess I'll need to move definitions of POLL* & RTLD_LAZY into all of the submodules...

PR #782, opened 2017-09-25
Add support for aarch64-unknown-linux-musl

tjkirch:
This adds support for aarch64-unknown-linux-musl as requested in rust-lang/rust#44779 by @alexcrichton 

The new file `ci/docker/aarch64-unknown-linux-musl/Dockerfile` essentially merges the aarch64-unknown-linux-gnu and x86_64-unknown-linux-musl Dockerfiles.

The bigger changes are under `src/unix/notbsd/linux/`, though they're fairly superficial:

Previously, some constants could be shared between all 64-bit musl triples, but aarch64 differs, so a number of things were moved from `musl/b64/mod.rs` to specific arches: `musl/b64/powerpc64.rs` and `musl/b64/x86_64.rs`, with the aarch64-specific differences being added to `musl/b64/aarch64.rs`.

Similarly, some constants moved from `musl/mod.rs` to lower levels: `musl/b32/mod.rs`, `musl/b64/powerpc64.rs`, and `musl/b64/x86_64.rs`, with the aarch64-specific differences added to `musl/b64/aarch64.rs`.

Finally, some things that were true of all Linux builds moved from `mod.rs` into lower levels: `mips/mod.rs`, `other/mod.rs`, `musl/b32/mod.rs`, `musl/b64/powerpc64.rs`, and `musl/b64/x86_64.rs`, with the aarch64-specific differences added to `musl/b64/aarch64.rs`.

Testing:

All linux-based triples under `ci/docker` were run through `ci/run-docker.sh` successfully, which checks that the size and alignment of definitions match for each triple, among other tests.  (The local build of rust from rust-lang/rust#44779 was set for the aarch64-unknown-linux-musl build.)

I also confirmed that it has "good style!" according to `ci/style.rs`.


alexcrichton:
Nice!

If you want to back out the changes to `src/lib.rs` and `.travis.yml` into a separate PR I think we can land this *right now* and then we can land those other changes once the target has landed in upstream rust-lang/rust?

tjkirch:
@alexcrichton Done!  When this is in, I'll send the PR for `src/lib.rs` and `.travis.yml` to enable the changes in CI.  Thanks!

PR #787, opened 2017-10-01
Fix kqueue filter consts type on NetBSD [WAS: Provide EV_SET]

vojtechkral:
Using BSD kqueue's `kevent` structure is tedious, because some BSD variants define the structure slightly differently. This forces user's code to differentiate between BSD variants and initialize `kevent` accordingly, which is annoying and error-prone.

For an example, refer to [MIO](https://github.com/carllerche/mio/blob/master/src/sys/unix/kqueue.rs#L38).

This is an attempt to fix it - provide a ctor function with the same signature across BSD variants.

Is an `impl` piece for a C structure allowed in libc?

**edit:** I noticed the `kevent` function has a similar problem, maybe I should include a fix for that too... 
**edit:** ^ Done

### DIFF
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -649,13 +649,13 @@ pub const PTHREAD_MUTEX_ERRORCHECK: ::c_int = 1;
 pub const PTHREAD_MUTEX_RECURSIVE: ::c_int = 2;
 pub const PTHREAD_MUTEX_DEFAULT: ::c_int = PTHREAD_MUTEX_NORMAL;
 
-pub const EVFILT_AIO: ::int32_t = 2;
-pub const EVFILT_PROC: ::int32_t = 4;
-pub const EVFILT_READ: ::int32_t = 0;
-pub const EVFILT_SIGNAL: ::int32_t = 5;
-pub const EVFILT_TIMER: ::int32_t = 6;
-pub const EVFILT_VNODE: ::int32_t = 3;
-pub const EVFILT_WRITE: ::int32_t = 1;
+pub const EVFILT_AIO: ::uint32_t = 2;
+pub const EVFILT_PROC: ::uint32_t = 4;
+pub const EVFILT_READ: ::uint32_t = 0;
+pub const EVFILT_SIGNAL: ::uint32_t = 5;
+pub const EVFILT_TIMER: ::uint32_t = 6;
+pub const EVFILT_VNODE: ::uint32_t = 3;
+pub const EVFILT_WRITE: ::uint32_t = 1;
 
 pub const EV_ADD: ::uint32_t = 0x1;
 pub const EV_DELETE: ::uint32_t = 0x2;


alexcrichton:
Thanks for the PR! Right now though this crate does not provide functions like this, this'd be an implementation detail in a downstream crate.

vojtechkral:
Ok, understandable. Are macros out of question too? (Ie. maybe provide an `EV_SET` equivalent?) If so, I'll just deal with it in my own code, feel free to close this PR is this isn't applicable.

alexcrichton:
Nah macros should be fine to add, there's already a number here

vojtechkral:
Allrighty then, I'll refactor the code...

alexcrichton:
Oh so actually you may want to look for invocations of the `f!` macro, we typically provide functions for C macros rather than literal Rust macros

vojtechkral:
Ok, take three using `f!`. I ended up figuring out defining a couple of helper type aliases is probably the best way. I also noticed kevent filter consts have a wrong type on netbsd.

src/unix/bsd/apple/mod.rs
```diff
@@ -23,6 +23,10 @@ pub type nl_item = ::c_int;
 pub type id_t = ::c_uint;
 pub type sem_t = ::c_int;
 pub type idtype_t = ::c_uint;
+pub type kevent_nevents = ::c_int;
+pub type kevent_filter = ::int16_t;
+pub type kevent_flags = ::uint16_t;
+pub type kevent_fflags = ::uint32_t;
```
alexcrichton: Hm if these don't actually exist, how come we're defining them?

src/unix/bsd/apple/mod.rs
```diff
@@ -1969,6 +1973,18 @@ f! {
     pub fn WIFSTOPPED(status: ::c_int) -> bool {
         _WSTATUS(status) == _WSTOPPED && WSTOPSIG(status) != 0x13
     }
+
+    pub fn EV_SET(ident: ::uintptr_t, filter: kevent_filter, flags: kevent_flags,
```
alexcrichton: These macros aren't like their C counterparts where they take a `kevent` as the first argument?

src/unix/bsd/apple/mod.rs
vojtechkral: Well, the idea is that, for example, `kevent_nevents` is "the type that the `kevent` function uses for the `nevents` param on this BSD variant" (it varies). So that in user code I can cast a variable `as kevent_nevents` without having to worry about which particular BSD I'm currently on. Same for the other `kevent_*` types.

src/unix/bsd/apple/mod.rs
vojtechkral: Oh, yes, I wrote it this way so that you don't need to first create an uninitialized `kevent` struct. But I don't have a strong opinion on this...

vojtechkral:
I'll fix those long lines...

src/unix/bsd/apple/mod.rs
alexcrichton: That makes sense, yes, but it's not what libc does. We don't add things that aren't in headers.

src/unix/bsd/apple/mod.rs
alexcrichton: I think we should stick to the C macros as much as possible

vojtechkral:
Ok, so, since platform abstractions aren't allowed, I guess I'll just make this into the netbsd const fixes.

The only other way of making `EV_SET` independent of the underlying integer type would be to make it generic with `Into<someCommonInteger>`, but that's not supported by the `f!` macro and I don't know whether it's allowed in terms of policy (I presume not). So, there's no real reason to include `EV_SET` since it wouldn't really do anything...

alexcrichton:
Ok, sounds like `EV_SET` should be left out?

vojtechkral:
Sorry for the delay. So I just reduced this to the fix for NetBSD. (Waiting for CI now.)

Issue #788, opened 2017-10-01
glibc 2.26 incompatibilities


vojtechkral:
glibc 2.26 introduces some changes that make `libc-test` not work anymore.

- It [removes `xlocale.h`](https://sourceware.org/glibc/wiki/Release/2.26#Removal_of_.27xlocale.h.27)
- It [removes `SIGUNUSED`](http://man7.org/linux/man-pages/man7/signal.7.html)

alexcrichton:
Thanks for the report! A PR to update this would be most welcome!

vojtechkral:
Fixed by #798

PR #789, opened 2017-10-03
Add pthread_getname_np, requires #define _GNU_SOURCE

dns2utf8:
I noticed two things:
1. `pthread_setname_np` was not included in libc so I started adding it.
2. inside rustc's `libstd/sys/unix/thread.rs` the linux variant `pthread_getname_np` is implemented with a raw syscall. So I added it here too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,7 +16,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 [[package]]
 name = "ctest"
 version = "0.1.6"
-source = "git+https://github.com/alexcrichton/ctest#621f64e78e25e71aca65f023591508dec188ae92"
+source = "git+https://github.com/alexcrichton/ctest#29d33e26d8f1297736c3716f4f2495dd068849ef"
 dependencies = [
  "cc 1.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "syntex_syntax 0.59.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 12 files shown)


alexcrichton:
Looks great! I think CI has some errors though?

dns2utf8:
All tests pass on my local machine with the latest Arch Linux.
I am working now on the MacX and Windows calls to rename a thread

vojtechkral:
@dns2utf8 

> All tests pass on my local machine with the latest Arch Linux.

I'm wondering how you got the `libc-test` to work on current Arch. I'm running into #788. What's your glibc version?

alexcrichton:
@dns2utf8 looks like there may still be a lot of CI errors?

dns2utf8:
I am testing it again now. 
My current version of glibc is 2.26-6

dns2utf8:
@alexcrichton regarding the CI errors. Should we switch to a container solution with a more recent linux?
I did that for kiss3d, but it is not merged yet: https://github.com/sebcrozet/kiss3d/pull/102

dns2utf8:
I noticed that the repo already has a CI folder, we could use the scripts or wrap them inside a container.

I am currently developing the extension for mac os x. Sadly it broke with the latest updates.

dns2utf8:
Hi all

I finished for the tier 1 plattform. Unfortunatly there is a style violation because android and musl do not support `getname`.
Do you have an idea how to fix this?
After that I would like to squash and merge it.

Best

alexcrichton:
Oh for style I think it can just be tweaked a bit, and for platforms missing this it's ok to just omit the definition on those platforms

alexcrichton:

PR #790, opened 2017-10-03
add utimensat to NetBSD and OpenBSD

semarie:


semarie:
futimens and utimensat should be defined on FreeBSD. For Apple, I am unsure. Let see what Travis will say.

semarie:
so apple seems to not have it, and FreeBSD 10 neither. The [man page](http://man.openbsd.org/FreeBSD-11.0/utimensat.2) for FreeBSD 11 is saying that there are here, but as i don't have the platform, I prefer to not declare it and mask it in test.

so just add `utimensat` to netbsdlike.

semarie:
looks like it failed on android for unrelated problem

semarie:
@alexcrichton do you think a release will be possible after this PR to be landed ?

Currently, I have a problem for building rustc-nightly on OpenBSD because of lake of `lutimes()` function. filetime uses it by default. So with `utimensat()` I could make filetime to use it instead of `lutimes()` for OpenBSD.

semarie:
after a push force, I still have the same errors on android:

```
Refresh Sources:

  Fetching URL: https://dl.google.com/android/repository/repository-11.xml

  Failed to fetch URL https://dl.google.com/android/repository/repository-11.xml, reason: SSLPeerUnverified peer not authenticated

There is nothing to install or update.

+ + android create avd --name arm --target android-24 --abi armeabi-v7a

echo no

Error: Target id is not valid. Use 'android list targets' to get the target ids.

The command '/bin/sh -c sh /android/android-install-sdk.sh $ANDROID_ARCH' returned a non-zero code: 1
```

@alexcrichton any ideas ?

semarie:
@alexcrichton with #791 it resolvs travis-ci test. thanks.

alexcrichton:
@semarie oh for releases there's a [section in the README](https://github.com/rust-lang/libc#releasing-your-change-to-cratesio) of how to do it

PR #793, opened 2017-10-05
Attempt to fix musl build.

clarcharr:
Follow-up to #685 and #700.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target
 Cargo.lock
 *~
+/style
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ services:
 install:
   - if [ -z "$NO_ADD" ]; then rustup target add $TARGET; fi
 script:
-  - cargo build
-  - cargo build --no-default-features
+  - cargo build --target=$TARGET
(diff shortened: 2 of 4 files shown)


alexcrichton:
Thanks! Looks good to me, but CI is failing?

clarcharr:
Fixed the style errors; waiting for Travis again.

clarcharr:
It appears to be passing!

alexcrichton:
@bors: r- retry

er actually, can you verify this works in the rustc repo?

clarcharr:
I made a PR for the rustc repo to test it out; I think that's how this is supposed to work?

clarcharr:
@alexcrichton it seems to work in the PR I made, so, I think this is okay to merge?

alexcrichton:
Er what I mean is to actually test this. For example actually build libstd for these targets and make sure it works.

One error, for example, is that `stdbuild` is no longer a thing. For example this PR as-is should break linkage on all these targets.

marioidival:
@clarcharr you can fix the conflict? 
@alexcrichton after conflict, need more changes here?

alexcrichton:
Ah I think I'll close for inactivity for now, but this'll just need to be tested against rust-lang/rust before merging.

PR #794, opened 2017-10-06
bump to 0.2.32

semarie:
- add DCCP constant definitions
- add clock_gettime related functions to macOS
- add fstatat64 on linux/android
- add preadv64/pwritev64 on linux/android
- add utimensat on solaris, netbsd and openbsd
- add IP_BINDANY on freebsd


semarie:
in order to unbreak rustc nightly under openbsd, I need a published version with utimensat.

semarie:
thanks

PR #798, opened 2017-10-09
Fix glibc 2.26 incompatibilities #788

vojtechkral:
Hopefully the `xlocale.h` → `locale.h` change won't break something else somewhere...


alexcrichton:
Looks like CI is failing?

PR #799, opened 2017-10-11
Define fdatasync on FreeBSD.  It was introduced in FreeBSD 11.1.

asomers:


alexcrichton:
Looks like CI may be failing?

asomers:
It looks like your QEMU image is still based on FreeBSD 10.  This PR will fail until you update that image.

alexcrichton:
Unfortunately this can't land until CI is green :(

asomers:
Then what's the procedure to upgrade the QEMU image, and is there any reason to hold it back?  The concerns around FreeBSD 12 compatibility don't apply here.

alexcrichton:
There should be some documentation in the `ci` folder about how it was originally created, and IIRC the result is just a QEMU image file which once you get working locally I can upload and the URLs here can get updated.

myfreeweb:
Just add it to `skip_fn` in ` libc-test/build.rs`

wezm:
@asomers the CI image has been updated to 11.1 so this might pass now.

asomers:
Thanks @wezm !  You have generously rewarded my laziness.

PR #802, opened 2017-10-12
add setfsgid/setfsuid on Linux

Mic92:


Mic92:
Interesting. Actually it is defined here, but seems to be missing in the headers used for th build: https://github.com/aosp-mirror/platform_bionic/blob/master/libc/include/sys/fsuid.h#L37

libc-test/build.rs
```diff
@@ -212,6 +211,7 @@ fn main() {
         cfg.header("sched.h");
         cfg.header("sys/epoll.h");
         cfg.header("sys/eventfd.h");
+        cfg.header("sys/fsuid.h");
```
Mic92: Now it is hopefully also in android available.

libc-test/build.rs
```diff
@@ -240,6 +239,8 @@ fn main() {
     }
 
     if linux || android {
+        cfg.header("sys/fsuid.h");
```
Mic92: turns out not to be available in emscripten.

PR #806, opened 2017-10-16
define SYS_renameat2 + flags on linux

Mic92:
Motivation: there is no glibc wrapper for this syscall


Mic92:
For comparison: https://github.com/systemd/systemd/blob/master/src/basic/missing_syscall.h#L201

alexcrichton:
@bors: r+

Thanks!

Mic92:
I had to exclude musl libc from `renameat2`.

PR #807, opened 2017-10-16
Local peer credential support for Darwin.

djg:
Darwin uses SOL_LOCAL/LOCAL_PEERCRED instead of SOL_SOCKET/SO_PEERCRED parameters to getsocktopt to retrieve the credentials of a socket's peer.

(Documented in man 4 unix)


alexcrichton:
@bors: r+

Thanks!

PR #810, opened 2017-10-18
Introduce pthread_cancel() for terminating threads

oherrala:
pthread_cancel() is safer way to terminate thread than pthread_kill().

This function can be found from ISO/IEC 9945-1:1996 (“POSIX.1”) and first introduced into Single Unix Specification version 2 on 1997.


alexcrichton:
@bors: r+

Thanks!

alexcrichton:
@bors: r- 

got back in the queue...

PR #812, opened 2017-10-18
Initial fixes for linux x32

malbarbo:
Tested with a local build of rustc. The main test can be compiled but fails to execute (receives SIGTRAP, which I think in this case indicates memory violation, I will open an issue in the rustc repo).

This PR is important because it fixes the definition of `c_long` and `c_ulong`. Without these fixes, rustc crash with the error:

```
Cannot emit physreg copy instruction
UNREACHABLE executed at /checkout/src/llvm/lib/Target/X86/X86InstrInfo.cpp:5778!
```

Related https://github.com/rust-lang/rust/issues/37976


alexcrichton:
@bors: r+

Thanks!

alexcrichton:
@bors: r+

...?

alexcrichton:
Thanks! I think this may need a rebase though now?

malbarbo:
Rebased. All green.

PR #813, opened 2017-10-19
SYS_* constants are of type long in Android

ndusart:
The SYS_* constants should be `c_long` as `syscall` expect a `c_long` as first argument and in Android headers, there is no sign that these constants should be `unsigned long`.

So the compilation targeting ARM Android fails for dependent crates using `syscall`.


alexcrichton:
@bors: r+

Oops, thanks!

Susurrus:
This actually [broke nix on almost all platforms](https://travis-ci.org/nix-rust/nix/builds/289807221?utm_source=github_status&utm_medium=notification). It appears that 9029cfb9eb76647cf9122a47252f1129eb18a562 is the culprit which changed existing `SYSCALL_*` constants and added more and they were all changed to `unsigned long` instead of `long`, which I don't think should have been done as @ndusart stated.

Note that this is currently blocking all PRs to `nix` as we pull the git version of `libc` until we decide to pull a release

Susurrus:
@alexcrichton Should I raise another issue about the `SYSCALL_` constants being incorrectly typed on non-Android platforms as I posted just after your response?

alexcrichton:
Do you mean the `SYS_` constants? If anything broken in https://github.com/rust-lang/libc/commit/9029cfb9eb76647cf9122a47252f1129eb18a562 that was accidental, and if anything is wrong that's also accidental. Let's just fix it?

Susurrus:
Yes, all changes to the `SYS_` constants changed datatypes in that commit and all those changes were incorrect as far as I can tell. I'd rather fix it than remove it, but I'm unfortunately not going to be able to do that today. @ndusart would you be willing to file another PR fixing the other platforms touched by 9029cfb?

asomers:
This PR should fix issue #814

ndusart:
@Susurrus, I could make another PR.
I just don't target linux arm in my projects so I didn't know it was broken for these archs too.

Susurrus:
It'd be most appreciated if you could!

ndusart:
@Susurrus #815

PR #817, opened 2017-10-24
add flags needed for SYS_memfd_create

Mic92:


libc-test/build.rs
```diff
@@ -213,6 +213,7 @@ fn main() {
         cfg.header("sched.h");
         cfg.header("sys/epoll.h");
         cfg.header("sys/eventfd.h");
+        cfg.header("sys/memfd.h");
```
Mic92: there is a good chance, I need to restrict platforms here to android/linux.

Mic92:
Turns out, I don't need this at the moment and maybe I fix this later.

PR #820, opened 2017-10-27
Run s390x tests on qemu system

malbarbo:
qemu in user mode cannot run binaries produced by rustc (it hangs), so we run the tests in a fully virtualized s390x machine using qemu-system-s390x.

Some constants had to be fixed.


malbarbo:
Android tests are failing. Maybe https://travis-ci.org/rust-lang/libc/jobs/293607888#L4848

> Download finished with wrong checksum. Expected b8130e7c390496cff12bf9355739bd41eed6a0a5, got 668ff8e319715175ff628ad52b124f154275fe2d.

alexcrichton:
Nice!

Something about the android tests do indeed look funny, I'll take a look. Looks like s390x is also failing?

alexcrichton:
@malbarbo so it looks like the Android docker images no longer have `adb`, presumably something about the downloads changed?

I think you've done some work with Android/CI in the past, mind helping me investigate? It looks like the `android` command we're using is deprecated in favor of some others, and I'm curious if you happen to already know off the top of your head how to move to the non-deprecated versions?

malbarbo:
@alexcrichton I'm trying to solve the android issue.

alexcrichton:
Awesome, thanks!

malbarbo:
I've tried many things but none made s390x builder works... It works locally...

Next week I will take a look at this again.

alexcrichton:
Oops sorry didn't mean to close.

I see you installed binfmt support in the container, is that needed to run foreign binaries? I haven't looked too closely at this yet, but historically I've seen the need for installing that package outside the container as well, for example directly on Travis , maybe that's what is up?

malbarbo:
I have changed to a simpler (and faster) approach that do not use qemu-user. Hopefully this will work on travis. I'm testing this approach with sparc64.

alexcrichton:
@bors: r+

Nice!

alexcrichton:
This feels "more correct" to me in terms of emulation in QEMU, so if this is easily adaptable to other architectures I'd be totally on board with doing this across the board for all cross-compiled arches

malbarbo:
I opened https://github.com/rust-lang/libc/issues/829 issue to track the use of qemu system.

Issue #822, opened 2017-10-27
Enable travis test for sparc64-unknown-linux-gnu


malbarbo:
Using qemu in user mode segfaults. Generating a image like in https://github.com/rust-lang/libc/pull/820 to run with qemu system also segfaults...

PR #823, opened 2017-10-27
Add sparc64-unknown-linux-gnu to CI (with disabled tests)

malbarbo:
Tests are disabled because qemu segfaults, see https://github.com/rust-lang/libc/issues/822

The builder is still useful to catch some errors.


alexcrichton:
@bors: r+

👍

PR #824, opened 2017-10-27
Fixes and test for x86_64-unknown-linux-gnux32

malbarbo:
This fixes all libc tests and almost all rustc tests (3 libstd tests and one run-pass test fails).

Fixes https://github.com/rust-lang/libc/issues/816https://github.com/rust-lang/libc/issues/816


src/unix/notbsd/linux/other/b64/x86_64.rs
```diff
@@ -164,7 +167,11 @@ s! {
         pub start_stack: ::c_ulonglong,
         pub signal: ::c_longlong,
         __reserved: ::c_int,
+        #[cfg(target_pointer_width = "32")]
+        __pad1: u32,
         pub u_ar0: *mut user_regs_struct,
+        #[cfg(target_pointer_width = "32")]
+        __pad2: u32,
```
malbarbo: Considering the [definition](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/x86/sys/user.h;hb=a4cc02c7413af4826fe8d13b6250c3fc47f15699#l85) of the user struct,  it's seems odd to add a pad before the field, but without it, the test fails...

.travis.yml
```diff
@@ -7,8 +7,8 @@ services:
 install:
   - if [ -z "$NO_ADD" ]; then rustup target add $TARGET; fi
 script:
-  - cargo build
-  - cargo build --no-default-features
+  - cargo build $OPT
+  - cargo build $OPT --no-default-features
```
alexcrichton: I think this may not be necessary due to the clause in `run.sh` below?

alexcrichton: Oh nvmd, I see!

ci/run.sh
```diff
@@ -68,4 +68,10 @@ if [ "$QEMU" != "" ]; then
   exec grep "^PASSED .* tests" $CARGO_TARGET_DIR/out.log
 fi
 
-exec cargo test --manifest-path libc-test/Cargo.toml --target $TARGET
+# FIXME: x86_64-unknown-linux-gnux32 fail to compile wihout --release
```
alexcrichton: Perhaps an issue could be filed to tag here as well?

src/unix/notbsd/linux/other/b64/x86_64.rs
alexcrichton: *shrug*

alexcrichton:
Nice! Everything looks great here modulo one nit for the issue to tag, and otherwise r=me

PR #825, opened 2017-10-27
Fix android ci

malbarbo:


malbarbo:
Tested locally (arm), let's see for the others archs.

alexcrichton:
@bors: r+ p=1

Awesome, thanks so much!

malbarbo:
I think the emulator for i686 does not run anymore without [hardware acceleration](https://travis-ci.org/rust-lang/libc/jobs/293825882#L4046)...

```
qemu-system-i386: -device virtio-blk-pci,drive=system,iothread=disk-iothread,modern-pio-notify: ioeventfd is required for iothread
emulator: WARNING: encryption is off
emulator: WARNING: x86 emulation may not work without hardware acceleration!
```

I tried to run i686 tests like the x86_64 tests (without the emulator), but I get an error:

```
assertion failed: signal(libc::SIGPIPE, libc::SIG_IGN) != libc::SIG_ERR
```

The same I [got](https://github.com/japaric/cross/pull/132) when adding android testing support to cross. A partial solution was run the tests with qemu-i686, but it needs a patched qemu.

What we do now?

alexcrichton:
Hmm that's unfortunate :(

Is hardware acceleration something that needs lots of privs in the cloud? (Aka we will never be able to do this bare on Travis)

If so maybe we should just remove the tests for now?

alexcrichton:
@bors: r+

Ok I've pushed a commit to disable it for now, but we can of course reenable if we can get it working again!

PR #827, opened 2017-10-29
Add WCONTINUED const for OpenBSD

worr:
Simple addition of `WCONTINUED` for OpenBSD-like systems. Built nix against this modified libc and used it to test.


alexcrichton:
@bors: r+

Thanks!

PR #828, opened 2017-10-31
Fix sparc64-unknown-linux-gnu tests

malbarbo:
Run sparc64 test in qemu system like s390x. Fix O_TMPFILE const.

Fixes https://github.com/rust-lang/libc/issues/822


malbarbo:
Based on https://github.com/rust-lang/libc/pull/820

alexcrichton:
@bors: r+

Nice!

PR #830, opened 2017-11-01
Add constants used by getrandom linux syscall

malbarbo:
getrandom syscall was added in kernel 3.17. Musl and android seems to use old kernel headers, but considering that they define SYS_getrandom, the constants GRND_NONBLOCK and GRND_RANDOM should also be defined.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -250,6 +250,10 @@ fn main() {
         }
     }
 
+    if linux {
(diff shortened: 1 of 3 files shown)


alexcrichton:
Looks great, thanks! r=me with musl CI back to green

alexcrichton:
@bors: r+

Seems fine by me!

PR #831, opened 2017-11-02
Compile an empty library on wasm32 non-Emscripten

alexcrichton:
In preparation for eventually having a non-Emscripten based wasm32 target, this
commit makes `libc` the crate an empty library on wasm32 targets that are not
with `target_os = "emscripten"`. This may eventually get filled out over time,
but for now it's all empty!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -100,176 +100,183 @@ extern crate std as core;
 #[macro_use] mod macros;
 mod dox;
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
r? @malbarbo 

Note that the [whitespace ignoring view](https://github.com/rust-lang/libc/pull/831/files?w=1) is more illuminating.

malbarbo:
@bors r+

I didn't know about the whitespace ignoring view. Thanks!

PR #833, opened 2017-11-03
Add pid_t to Windows

shaleh:


malbarbo:
`c_int` is not in scope, try `::c_int`.

shaleh:
Thanks @malbarbo, I had issues compiling and testing locally due to missing depends.

alexcrichton:
Thanks! Looks like the pid may be wrong though on WIndows for some platforms?

shaleh:
According to MS docs it is int. Not i64 or i32.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

alexcrichton:
According to the code though the docs are wrong :(

shaleh:
https://sourceforge.net/p/mingw-w64/mailman/message/34497805/ a 2 year old question about why ming64 defines `pid_t` as a 64bit value when it should be 32. Yay.

shaleh:
And the Window Visual compiler tests are failing because the Windows headers do not define `pid_t` and this is 100% true. Rust libc is being asked to have a typedef to ease porting. If I understand the tests correctly I can add code to the `skip_type` section of `build.rs` when the compilation is running on Windows and that will get the tests closer to passing. I will update the PR and see how it goes...

libc-test/build.rs
```diff
@@ -302,6 +302,7 @@ fn main() {
             // Fixup a few types on windows that don't actually exist.
             "time64_t" if windows => "__time64_t".to_string(),
             "ssize_t" if windows => "SSIZE_T".to_string(),
+            "pid_t" if windows => "pid_t".to_string(),
```
alexcrichton: I think this may be caught by the clause below, right?

shaleh: Drat. I totally missed that. Too right.

shaleh: i have been reading too much lisp code. I read the "t if" as "true if" :-)

alexcrichton: Heh no worries!

shaleh: I have a fresh Win10 VM running so I should be able to test this locally finally.

shaleh:
Goal achieved with the latest commit:

    extern crate libc;

    fn main() {
        let pid: libc::pid_t = unsafe { libc::getpid() };
        println!("Hello, world! {}", pid);
    }

libc-test/build.rs
```diff
@@ -302,6 +302,7 @@ fn main() {
             // Fixup a few types on windows that don't actually exist.
             "time64_t" if windows => "__time64_t".to_string(),
             "ssize_t" if windows => "SSIZE_T".to_string(),
+            "pid_t" if windows => "int".to_string(),
```
alexcrichton: Hm this is completely renaming the type? Does Windows not have a `pid_t` type in MinGW or MSVC?

shaleh: Correct.

shaleh: `getpid` aka `_getpid` just returns 'int'.

alexcrichton: Ah ok, in that case we tend to prefer to not try to abstract over the underlying platform too much in the sense that we don't want to manufacture types in Rust-lang which don't actually exist on the underlying system.

shaleh: Grumble. This is a common thing to have in cross platform code. Windows is the only platform without pid_t. Leaving it out of libc forces every lib and program to have ifdefs in their own code.

shaleh: I also learned how to fix the compilation error by copying existing Windows work arounds so this isn't the first one to be added.

alexcrichton: Oh sure yeah it makes sense that this is a pain, but it's not `libc`'s job to sovle this, that's up to another crate.

marioidival:
@alexcrichton about this PR, should change something?

alexcrichton:
Ah sorry I meant to close this awhile ago, it looks like MinGW doesn't have a `pid_t`

PR #836, opened 2017-11-05
Add constants for memfd_create syscall

Susurrus:


Susurrus:
Alright, we're green, finally! Should be good to merge.

Susurrus:
@alexcrichton So I left off many of the musl targets because these constants aren't declared there. However, since this is a kernel interface, I'd imagine it's appropriate for those targets as well. Is that a correct understanding?

alexcrichton:
Yeah I think that'd be fine.

PR #839, opened 2017-11-07
Add net interface device flag

LuoZijun:
Add net interface device flag for all platforms.


LuoZijun:
@alexander255 
 the failing is about [libc/ci/style.rs](https://github.com/rust-lang/libc/blob/master/ci/style.rs#L55)

like some lines longer than 80 chars, maybe we turn off `panic` and use one `rustfmt.toml`  ?

alexander255:
@LuoZijun I take it you were talking to @alexcrichton, not me? 🙂

LuoZijun:
@alexander255 Oh ... mistake :D

alexcrichton:
@LuoZijun for now let's stick with `style.rs`, it in theory shouldn't be too too hard to adhere to. Eventually we can use rustfmt when it's ready

PR #842, opened 2017-11-11
Add dl_iterate_phdr and related types

sfackler:
A lot of this is more broadly supported than just Linux, but support for
those can be added later.

r? @alexcrichton

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -23,6 +23,17 @@ pub type __s16 = ::c_short;
 pub type __u32 = ::c_uint;
 pub type __s32 = ::c_int;
 
+pub type Elf32_Half = u16;
+pub type Elf32_Word = u32;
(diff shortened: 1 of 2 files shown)


src/unix/notbsd/linux/mod.rs
```diff
@@ -391,6 +402,52 @@ s! {
...
+        pub dlpi_phdr: *const Elf64_Phdr,
+        #[cfg(target_pointer_width = "32")]
+        pub dlpi_phdr: *const Elf32_Phdr,
+
+        #[cfg(target_pointer_width = "64")]
+        pub dlpi_phnum: Elf64_Half,
+        #[cfg(target_pointer_width = "32")]
+        pub dlpi_phnum: Elf32_Half,
+
+        pub dlpi_adds: ::c_ulonglong,
```
sfackler: These four fields weren't present when this type was initially defined, but hopefully they were added long enough ago that every glibc version we care about supports them.

PR #844, opened 2017-11-13
Use more convenient and UNIX-agnostic shebang

dereckson:
Pure sh scripts should use /bin/sh as it's available on every platform.

When using bash-specific features, use env to find it, as bash can
be installed in different places according the OS.


malbarbo:
Thanks!

@bors r+

PR #847, opened 2017-11-14
Add aarch64 support to newlib bindings

roblabla:
It's all in the title. I grabbed the definition from  a small test in with a gcc toolchain, I hope it's correct.

There's currently no tests around newlib in the CI. Would it be possible to add them ? If so, how ?


alexcrichton:
@bors: r+

Thanks! Yeah we don't have newlib tests right now because upstream there are no newline releases for the rust standard library

PR #848, opened 2017-11-15
Update Fuchsia open flags

cramertj:
These are being changed internally-- see [here](https://fuchsia-review.googlesource.com/c/zircon/+/83959/15/third_party/ulib/musl/include/fcntl.h#57).

I wasn't sure about the best way to cfg the entries that were architecture specific, so I just blocked them out and inserted them at a higher level. Let me know if there's a better way to organize the cfgs.

cc @smklein

### DIFF
--- a/src/unix/notbsd/linux/musl/b64/aarch64.rs
+++ b/src/unix/notbsd/linux/musl/b64/aarch64.rs
@@ -62,10 +62,15 @@ s! {
     }
 }
 
-pub const O_DIRECT: ::c_int = 0x10000;
-pub const O_DIRECTORY: ::c_int = 0x4000;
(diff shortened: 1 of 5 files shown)


smklein:
Double checked the O_* values with the planned change from Fuchsia, I can vouch that they match.

alexcrichton:
Thanks! Perhaps fuchsia though could get it's own hierarchy now? We tend to prefer inline cfg blocks like this as it can be quite difficult to maintain over time

cramertj:
Sure, that seems reasonable. I was trying to keep this particular diff minimal, but I'll open a new CL with a new Fuchsia target.

PR #849, opened 2017-11-15
Separate and Update Fuchsia

cramertj:
The first commit pulls Fuchsia out into its own directory, while the second commit updates the "open" flags as in #848.

cc @smklein, @alexcrichton

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -286,6 +286,9 @@ cfg_if! {
     } else if #[cfg(target_os = "redox")] {
         mod redox;
         pub use redox::*;
+    } else if #[cfg(target_os = "fuchisa")] {
+        mod fuchsia;
+        pub use fuchsia::*;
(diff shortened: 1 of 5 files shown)


cramertj:
@alexcrichton Should I open a PR against rustc updating the submodule, or is there a more official process for updating libc?

alexcrichton:
Nah that's it!

src/lib.rs
```diff
@@ -286,6 +286,9 @@ cfg_if! {
     } else if #[cfg(target_os = "redox")] {
         mod redox;
         pub use redox::*;
+    } else if #[cfg(target_os = "fuchisa")] {
```
petrhosek: There's a typo: `fuchisa`, so this is likely not going to work.

cramertj: Gah! https://github.com/rust-lang/libc/pull/854

PR #859, opened 2017-11-28
Use IFF_OACTIVE and IFF_RUNNING even on FreeBSD. Deprecate the DRV ones.

mwanner:
According to @asomers, libc should propagate the use of the portable constants `IFF_OACTIVE` and `IFF_RUNNING` for user-space applications, instead of `IFF_DRV_OACTIVE` and `IFF_DRV_RUNNING`. It least that's my understanding of [his comment](https://github.com/nix-rust/nix/pull/667#discussion_r152396984) in nix-rust/nix#667.


alexcrichton:

asomers:
The failure is due to a "line longer than 80 characters" error in the style checker.

mwanner:
Overly long lines corrected, CI now happy.

semarie:
`#[deprecated()]` seems to make problem when used in rustc.

see https://github.com/rust-lang/rust/pull/46760#issuecomment-352234236

PR #860, opened 2017-11-28
Add a couple pthread function definitions, get_sched and backtrace for musl

mwanner:
trying to compile a binary statically using `x86_64-unknown-linux-musl`, I figured the `pthread_{g,s}etaffinity` functions were missing. I added these - and a couple others found in the vicinity - from `linux/other/mod.rs`.

I'm also wondering if that really is the right approach, since there seems to be quite a lot of duplication between all of the `linux/*/mod.rs` files...


alexcrichton:
Thanks! Looks like CI may be failing though?

tamird:
@mwanner are you sure musl implements these functions? I can't find `backtrace`, `pthread_attr_{get,set}affinity_np`, or `pthread_rwlockattr_{get,set}kind_np` in the musl sources.

alexcrichton:

PR #865, opened 2017-12-04
Minor fixes for mips-unknown-linux-uclibc

cactorium:
Just adding a few definitions so the toolchain compiles


alexcrichton:
@bors: r+

Thanks!

PR #867, opened 2017-12-07
Ucreds

Susurrus:
I'm not positive on the `SOCKCREDSIZE()` implementation, but we'll see what CI says I guess.

### DIFF
--- a/src/dox.rs
+++ b/src/dox.rs
@@ -146,5 +146,6 @@ mod imp {
 
     pub mod mem {
         pub fn size_of_val<T>(_: &T) -> usize { 4 }
+        pub fn size_of<T>(_: &T) -> usize { 4 }
     }
 }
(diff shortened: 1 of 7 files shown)


Susurrus:
Hmm, it looks like I imported things wrong. Can someone provide some guidance on how I can safely use `core::mem::size_of` in my types here?

Susurrus:
Just fixed FreeBSD, so hopefully that passes CI now, but I'm not certain why nightly was failing, maybe some change in module path names?

alexcrichton:
The documentation build is slightly different, there's a few other examples throughout of importing `mem` (it's a bit of a hack right now...)

Susurrus:
I dug around enough to know something weird was going on, but didn't quite understand it, so I was hoping for some more details in your response as it didn't give me any new information. Anyways, I hacked around at it a bit more and I think what I've got now is right. We'll see what CI says.

Susurrus:
Sweet, that worked!

PR #873, opened 2017-12-14
Add `net/if_utun.h` and `linux/if_tun.h` constants for macos and linux platform

LuoZijun:
Add `net/if_utun.h` constants for macos platform.

Add `linux/if_tun.h` constants for linux platform.

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -2071,7 +2071,6 @@ pub const MH_MAGIC: u32 = 0xfeedface;
 pub const MH_MAGIC_64: u32 = 0xfeedfacf;
 
 // net/if_utun.h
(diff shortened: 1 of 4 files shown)


src/unix/bsd/apple/mod.rs
```diff
@@ -2070,6 +2070,11 @@ pub const LC_SEGMENT_64: u32 = 0x19;
 pub const MH_MAGIC: u32 = 0xfeedface;
 pub const MH_MAGIC_64: u32 = 0xfeedfacf;
 
+// net/if_utun.h
+pub const UTUN_CONTROL_NAME: &'static str = "com.apple.net.utun_control";
```
alexcrichton: Unfortunately static strings don't map greatly to C right now and we don't have any other instances of `str` right now, could this be left out for now?

LuoZijun: okay.

PR #875, opened 2017-12-16
Add additional aarch64 linux syscalls

kpcyrd:
I've noticed some syscalls are missing for aarch64 linux, I've added those I needed. The numbers have been taken from a raspberry pi 3.

Adds:
- `SYS_fcntl`
- `SYS_lseek`
- `SYS_newfstatat`
- `SYS_fstat`
- `SYS_mmap`
- `SYS_open`
- `SYS_stat`
- `SYS_lstat`
- `SYS_pipe`
- `SYS_getdents`
- `SYS_poll`


alexcrichton:

kpcyrd:
Oh, didn't notice the tests didn't work. The problem seems to be that for these:

```
cargo:warning=             static  long __test_const_SYS_open_val = SYS_open;
cargo:warning=             static  long __test_const_SYS_stat_val = SYS_stat;
cargo:warning=             static  long __test_const_SYS_lstat_val = SYS_lstat;
cargo:warning=             static  long __test_const_SYS_pipe_val = SYS_pipe;
cargo:warning=             static  long __test_const_SYS_getdents_val = SYS_getdents;
cargo:warning=             static  long __test_const_SYS_poll_val = SYS_poll;
```

there's a number defined, eg `#define __NR_open 1024`, but there's no `#define SYS_open __NR_open` for those, causing the test to fail.

I'm not 100% sure what's the best thing to do. Unless there are objections with a better way, I would try to remove those again and drop them from my seccomp filter using `#[cfg(not(target_arch = "aarch64"))]`.

kpcyrd:
The tests are passing now. It turned out those syscalls are not actually in use on aarch64 and I don't need to add them to seccomp when compiling for that architecture.

alexcrichton:
@bors: r+

Ok!

PR #877, opened 2017-12-17
remove #[deprecated(...)] attributes

semarie:
it makes problems with `#[feature(staged_api)]` when used inside rustc code.

introduced by #859
see https://github.com/rust-lang/rust/pull/46760


alexcrichton:
Ah sorry but I think https://github.com/rust-lang/libc/pull/876 beat this PR to the punch!

semarie:
ok, just see than #876 already covers it. thanks @asomers.

PR #880, opened 2017-12-19
Add SECCOMP_MODE_* flags on Linux

Mic92:


src/unix/notbsd/linux/mod.rs
```diff
@@ -1084,6 +1084,10 @@ pub const PR_CAP_AMBIENT_CLEAR_ALL: ::c_int = 4;
 pub const GRND_NONBLOCK: ::c_uint = 0x0001;
 pub const GRND_RANDOM: ::c_uint = 0x0002;
 
+pub const SECCOMP_MODE_DISABLED: ::c_ulong = 0;
+pub const SECCOMP_MODE_STRICT: ::c_ulong = 1;
+pub const SECCOMP_MODE_FILTER: ::c_ulong = 2;
```
malbarbo: Should it be c_uint? http://man7.org/linux/man-pages/man2/seccomp.2.html

Mic92: probably. I was looking at `prctl`, when implementing this.

PR #884, opened 2017-12-27
Add support for CloudABI.

EdSchouten:
CloudABI is a sandboxed UNIX-like runtime environment, based on the
principle of capability-based security. As CloudABI is intended to be
cross-platform, the system call layer is specified here:

    https://github.com/NuxiNL/cloudabi/blob/master/cloudabi.txt

From these definitions, we automatically generate C and Rust bindings.
The latter is published on crates.io:

    https://crates.io/crates/cloudabi

My goal is to implement libstd for CloudABI in such a way that it uses
the C library as little as possible; only in places where it would ease
interfacing with C code (e.g., thread creation). In places where
constants in the C library are directly based on the CloudABI
specification (e.g., errnos), use the constants provided by the cloudabi
crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,22 +9,9 @@ version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
-name = "bitflags"
-version = "1.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
 name = "cc"
 version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
-name = "cloudabi"
-version = "0.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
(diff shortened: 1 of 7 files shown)


alexcrichton:
Thanks for the PR! Unfortunately though `libc` for various technical reasons can't depend on other crates, could the definitions from the `cloudabi` crate be inlined here?

EdSchouten:
Hi Alex,

Thanks for taking a look! The `cloudabi` package also contains bindings/accessors for all of CloudABI's system calls, meaning that libstd for CloudABI will also depend on it. I've solved this by adding a `cloudabi_shim` to the Rust source tree and letting libc and libstd depend on that: https://github.com/EdSchouten/rust/tree/cloudabi/src/rustc. As libstd has to use it, I guess it's also less of a problem to depend on it here, right?

That said, I've just pushed a second commit to drop the dependency on the `cloudabi` package. If you want me to roll back to the version that does depend on the `cloudabi` package, just let me know.

Ed

alexcrichton:
We can in general manage the dependency with libstd better, yeah, but it'd be best to not depend on it in libstd as well if possible.

@bors: r+

PR #886, opened 2018-01-02
Add support for copyfile on Darwin

mre:
This PR adds support for the [`copyfile`](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/copyfile.3.html) syscall. It is similar to [splice](http://man7.org/linux/man-pages/man2/splice.2.html) on Linux and offers fast copying of pages in kernel-space.

Right now, the tests fail, because `copyfile_state_t` is not a struct, but a pointer.
I thought, using `state: *mut copyfile_state_t` in the function definition might be the way to go, but that doesn't work. Any help would be appreciated here.

Also, once this is fixed I also want to add `fcopyfile`, which should [work on FreeBSD](https://stackoverflow.com/a/2180157/270334), too.
There are some other helper functions defined in the [manpage](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/copyfile.3.html) (e.g. `copyfile_state_alloc`, `copyfile_state_get`,...) and we could add them, too if we want.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -170,17 +170,21 @@ fn main() {
         cfg.header("sys/xattr.h");
         cfg.header("sys/sys_domain.h");
         cfg.header("net/if_utun.h");
+        cfg.header("net/bpf.h");
         if target.starts_with("x86") {
             cfg.header("crt_externs.h");
         }
         cfg.header("net/route.h");
+        cfg.header("netinet/if_ether.h");
         cfg.header("sys/proc_info.h");
+        cfg.header("copyfile.h");
     }
 
     if bsdlike {
         cfg.header("sys/event.h");
         cfg.header("net/if_dl.h");
         if freebsd {
+            cfg.header("net/bpf.h");
             cfg.header("libutil.h");
         } else {
             cfg.header("util.h");
@@ -251,7 +255,8 @@ fn main() {
     if linux || android {
         cfg.header("sys/fsuid.h");
         cfg.header("linux/seccomp.h");
-
+        cfg.header("linux/if_ether.h");
+        
         // DCCP support
         if !uclibc && !musl && !emscripten {
             cfg.header("linux/dccp.h");
--- a/src/cloudabi/aarch64.rs
+++ b/src/cloudabi/aarch64.rs
@@ -1,0 +1,4 @@
+pub type c_char = i8;
+pub type c_long = i64;
+pub type c_ulong = u64;
+pub type wchar_t = u32;
--- a/src/cloudabi/mod.rs
+++ b/src/cloudabi/mod.rs
@@ -1,0 +1,161 @@
+pub type in_addr_t = u32;
+pub type in_port_t = u16;
+pub type pthread_key_t = usize;
+pub type pthread_t = usize;
+pub type sa_family_t = u8;
+pub type socklen_t = usize;
+
+s! {
+    pub struct addrinfo {
+        pub ai_flags: ::c_int,
+        pub ai_family: ::c_int,
+        pub ai_socktype: ::c_int,
+        pub ai_protocol: ::c_int,
+        pub ai_addrlen: ::socklen_t,
+        pub ai_addr: *mut ::sockaddr,
+        pub ai_canonname: *mut ::c_char,
+        pub ai_next: *mut addrinfo,
+    }
+
+    pub struct in_addr {
+        pub s_addr: in_addr_t,
+    }
+
+    pub struct in6_addr {
+        pub s6_addr: [u8; 16],
+    }
+
+    pub struct pthread_attr_t {
+        __detachstate: ::c_int,
+        __stacksize: usize,
+    }
+
+    pub struct sockaddr {
+        pub sa_family: sa_family_t,
+        pub sa_data: [::c_char; 0],
+    }
+
+    pub struct sockaddr_in {
+        pub sin_family: ::sa_family_t,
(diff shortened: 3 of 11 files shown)


mre:
Humm... don't really know what to try next. Any ideas @alexcrichton?

alexcrichton:
@mre it looks like the verification program, `libc-test`, failed? I think maybe it's a missing header file?

mre:
I thought `copyfile.h` would be enough. It was already added to the build script [here](https://github.com/rust-lang/libc/pull/886/files#diff-747c41e91bd79d9208c2953b72047176R180) and according to [the docs](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/copyfile.3.html), that should be all.

For a test, I also tried adding this:

```rust
cfg.header("sys/cdefs.h");
cfg.header("stdint.h");
```

These are the includes from the [copyfile headerfile](https://opensource.apple.com/source/copyfile/copyfile-85.1/copyfile.h.auto.html), but that didn't help. 🤔

alexcrichton:
Sorry I dunno myself, I'm not too familiar with this API

mre:
Just in case anybody wants to give it a shot, here's the first error message I'm currently struggling with:

```
cargo:warning=/Users/travis/build/rust-lang/libc/target/i686-apple-darwin/debug/build/libc-test-f149ec9caa38d75d/out/main.c:10040:28: error: offsetof requires struct, union, or class type, 'copyfile_state_t' (aka 'struct _copyfile_state *') invalid
cargo:warning=                    return offsetof(copyfile_state_t, src);
cargo:warning=           ```

**Comment by asomers** [comment]
The author of that stackoverflow post is completely wrong when he says that fcopyfile is supported on FreeBSD.

**Comment by Susurrus** [comment]
The problem here is that you didn't implement the C API exactly as it is. In your Rust code `copyfile_state_t` refers to an entire struct while in the C code it refers to a `_copyfile_state *`. Does that make sense? You should change your code to also declare the `_copyfile_state` struct and also a typedef to use for `copyfile_state_t`.

@alexcrichton Is that the right way to go about this? I've not seen an API like this reproduced in `libc`, where it tries to hide the underlying type through a pointer typedef. The header says "this API is not stable" and I'm worried about if we implement this that the `libc` stability guarantees will bite us in the butt in the future.

**Comment by alexcrichton** [comment]
Hm yeah if possible I think it'd be best to avoid binding unstable APIs (leave them to downstream consumers), but @mre did you have a use case in mind when adding this?

**Comment by mre** [comment]
Thanks @asomers. Just tested on FreeBSD 10.3 and it doesn't work indeed. Added a comment to the StackOverflow thread.

Thanks for your suggestions @Susurrus. Highly appreciated! I tried to change the code according to your suggestions. It still doesn't build, but I'm pretty convinced it's entirely my fault. I've updated the PR. Currently, the first error message is:

```
cargo:warning=/Users/mendler/Code/upstream/libc/target/debug/build/libc-test-50dafc47d3c66791/out/main.c:9959:74: error: use of undeclared identifier '_copyfile_state_t'; did you mean 'copyfile_state_get'?
cargo:warning=            uint64_t __test_size__copyfile_state_t(void) { return sizeof(_copyfile_state_t); }
cargo:warning=                                                                         ^~~~~~~~~~~~~~~~~
cargo:warning=                                                                         copyfile_state_get
```

Maybe that has something to do with the underscore in `'_copyfile_state_t`?

@alexcrichton, @Susurrus: the API is marked as unstable, but it has been unchanged since ten years. I doubt that it will change in the future. If anything, it might be removed eventually. 😉 But so far, I don't know of any removal plans or alternatives.
Basically, it's the only way to copy files performantly, without going through userspace, similar to splice and sendfile for Linux. This stuff is nice to have for building webservers. NGINX supports [sendfile](http://nginx.org/en/docs/http/ngx_http_core_module.html#sendfile), for example (but not copyfile).
My use-case is rather simple: implement a faster `cat`.

I'm totally fine if we don't want to merge that into libc. We could host that in a separate crate. For that, I would basically take libc, strip out everything that is not needed, and push it to crates.io. Your call.
Would still be happy for any input on how to get it working, though. 😁

asomers:
@mre are you sure [copyfile(3)](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/copyfile.3.html) is zero-copy?  The man page doesn't say anything about zero-copy, splice, or page flipping.  Indeed, the fact that the man page is in section 3 and there's no corresponding entry in section 2 suggests that there are no special system calls going on.  I don't think it's zero-copy at all.  I think it's basically just cp(1) turned into a library routine.  It's main benefit seems to be that it knows how to copy a file's extended attributes.

mre:
@asomers, you might be right. My initial assumption was based on [this answer on StackOverflow](https://stackoverflow.com/a/2180157/270334):
> The following code snippet should work on the most OS X (as of 10.5), (Free)BSD, and Linux (as of 2.6.33). The implementation is "zero-copy" for all platforms, meaning all of it is done in kernelspace and there is no copying of buffers or data in and out of userspace. Pretty much the best performance you can get.

I did some research now, and I also can't find any evidence of zero-copy. Therefore I stated the question on the SO thread. Lesson learned.
Thanks again for the input y'all. I guess we can close this. Seems like there's no way to do file-to-file zero-copy on Mac right now.

alexcrichton:
Ah ok in that case I'll close for now, but we can always reopen later!

PR #893, opened 2018-01-11
feat(linux): add mntent struct and related functions

kelseasy:
Useful when playing around with mounting points, sadly not a common interface. :(


alexcrichton:
@bors: r+

Thanks!

PR #896, opened 2018-01-12
added EPOLL(5)  for Illumos (Solaris fork) 

cneira:
EPOLL(5)  is being used in crates like mio and iovec , this change allows those crates and other that depends on EPOLL(5)  to be built in illumos systems.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -215,7 +215,7 @@ fn main() {
             cfg.header("sys/reg.h");
         }
     }
-
+ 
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! I think there's some tidy failures, and could the changes to src/lib.rs get backed out as well? (it looks like nothing necessary?)

cneira:
Fixed code style, and removed not needed changes in lib.rs

cneira:
Thanks a lot @bors

PR #902, opened 2018-01-17
Missing linux intrinsics

gnzlbg:


gnzlbg:
This obviously fail because `musl` does not support `mallinfo`, and [never intends to support it](http://www.openwall.com/lists/musl/2018/01/17/1):

I've fixed that. The breakage in the other build bots looked spurious. Let's see what breaks next.

alexcrichton:
Thanks! Looks like CI may be uncovering a few style errors?

gnzlbg:
So @bors is right, those are real errors, it looks like:

* `malloc_usable_size`
* `memfd_create`
* `getauxval`

are not available in some glibc targets, the question is, why? The only thing that I can come up with is that these targets are using a way older version of glibc than the others.

gnzlbg:
@japaric ?

malbarbo:
It seems that all targets included in linux/others are failing, so it seems not to be a glibc version issue.

I take a quick look at each failing function, here are some observatios:

> cargo:warning=/checkout/target/x86_64-unknown-linux-gnux32/release/build/libc-test-47f843c814227a35/out/main.c:18933:24: error: return from incompatible pointer type [-Werror=incompatible-pointer-types]
> cargo:warning=                 return malloc_usable_size;

The parameter is not const: http://man7.org/linux/man-pages/man3/malloc_usable_size.3.html

> cargo:warning=/checkout/target/x86_64-unknown-linux-gnux32/release/build/libc-test-47f843c814227a35/out/main.c:18938:24: error: 'memfd_create' undeclared (first use in this function); did you mean 'timerfd_create'?
> cargo:warning=                 return memfd_create;

There is no glibc wrapper for this system call: http://man7.org/linux/man-pages/man2/memfd_create.2.html

I don't known if rust libc should provide such a wrapper. What do you think @alexcrichton?

> cargo:warning=/checkout/target/x86_64-unknown-linux-gnux32/release/build/libc-test-47f843c814227a35/out/main.c:18943:24: error: 'getauxval' undeclared (first use in this function)
> cargo:warning=                 return getauxval;

Missing include `<sys/auxv.h>`  in `build.rs`?

gnzlbg:
> There is no glibc wrapper for this system call: http://man7.org/linux/man-pages/man2/memfd_create.2.html

Is there precedence of libc providing these kinds of wrappers?

malbarbo:
I think there is no precedence for these kinds of wrappers. The only functions defined in libc are for c macros.

gnzlbg:
@malbarbo then I'll leave it out of this PR so that the rest can be merged.

gnzlbg:
Wait a second, what is this doing here?

https://github.com/rust-lang/libc/blob/master/libc-test/build.rs#L271

See https://github.com/rust-lang/libc/commit/c0935ac343ebc9bed86be30bea7e7a2fb1adf189 :/ 

So the constants are there, but the `memfd_create` function isn't ?

malbarbo:
The [`memfd.h`](https://github.com/torvalds/linux/blob/master/include/uapi/linux/memfd.h) file only contains the constants.

There is no function `memfd_create`, only the syscall. The issue is that the syscall is documented like a function...

gnzlbg:
This should be good to go now :)

malbarbo:
@bors r+

Thanks @gnzlbg

PR #903, opened 2018-01-18
Bsd domainname

### LINKED ISSUE #868: bsd: {get,set}domainname are missing
Currently {get,set}domainname are exposed on Linux and emscripten only. However they are implemented on BSDs too so they should be exposed there are as well:
 * https://www.freebsd.org/cgi/man.cgi?sektion=3&query=gethostname
 * https://man.openbsd.org/setdomainname.3
 * http://netbsd.gw.com/cgi-bin/man-cgi?getdomainname+3+NetBSD-current
 * https://www.dragonflybsd.org/cgi/web-man?command=getdomainname&section=3

gnzlbg:
Closes #868 .

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -2352,7 +2352,8 @@ extern {
     pub fn querylocale(mask: ::c_int, loc: ::locale_t) -> *const ::c_char;
     pub fn getpriority(which: ::c_int, who: ::id_t) -> ::c_int;
     pub fn setpriority(which: ::c_int, who: ::id_t, prio: ::c_int) -> ::c_int;
-
+    pub fn getdomainname(name: *mut ::c_char, len: ::c_int) -> ::c_int;
+    pub fn setdomainname(name: *const ::c_char, len: ::c_int) -> ::c_int;
     pub fn getxattr(path: *const ::c_char, name: *const ::c_char,
                     value: *mut ::c_void, size: ::size_t, position: u32,
                     flags: ::c_int) -> ::ssize_t;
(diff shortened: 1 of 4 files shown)


gnzlbg:
These can probably be refactored to an outer level but I just want to see if any of the BSDs breaks because of this first.

lucab:
It looks good on the surface, but I don't have BSD knowledge or a test environment to double-check.
/cc @asomers  and @Susurrus who originally pointed this out.

PS: I tend to add a `Fixes: #868` note in PRs body to cross-reference and ease navigation.

gnzlbg:
> PS: I tend to add a Fixes: #868 note in PRs body to cross-reference and ease navigation.

The second commit has a `Closes #868` so once this PR is squashed and merged, this issue will be automatically closed and the whole thing cross-referenced by github.

lucab:
@gnzlbg it tests *some* BSDs, yup. I think it doesn't cover the whole genealogy, but it is often enough.

And yes, I was specifically mentioning PR body so the bug gets a direct and stable link to here as opposed to the commit (I'm lazy, plus it becomes noisy when rebasing many times). Anyway, I'm going off-topic and it was just a sidenote, don't worry and thanks for the fix!

alexcrichton:
Thanks! Looks like OSX has a failure though?

Susurrus:
`gethostname` is already implemented for all unices, so it can simply be removed from this PR.

asomers:
Looks like the `namelen` argument needs to be of type `int` on FreeBSD, Dragonfly, and OSX.

Susurrus:
And on mac, it's `getdomainname(name: *mut ::c_char, len: ::c_int)` and `etdomainname(name: *const ::c_char, len: ::c_int) -> ::c_int`

asomers:
And unlike the other functions, `gethostname` is actually standardized.  So that can probably go in the unix directory.

PR #905, opened 2018-01-20
add DT_UNKNOWN

wfraser:
The `d_type` field of `struct dirent` can be `DT_UNKNOWN` if a
filesystem doesn't support returning the file type in directory entries.
Linux's readdir(3) man page, for example, says: "Currently, only some
filesystems [...] have full support for returning the file type in
d_type. All applications must properly handle a return of DT_UNKNOWN."


wfraser:
looks like one of the tests had an infrastructure failure

alexcrichton:
@bors: r+

Thanks!

PR #908, opened 2018-01-22
Do not assume dynamic linking in musl/mips targets

malbarbo:


malbarbo:
Required for https://github.com/rust-lang/rust/pull/47663

alexcrichton:
Due to the usage of `target_feature` I think this won't break the current target when landing, right?

malbarbo:
I think it will break... The target definition is initialized using `..super::linux_base::opts()`, which returns `crt_static_respected = false`...

malbarbo:
Considering that we are not changing the default linking mode in https://github.com/rust-lang/rust/pull/47663 anymore, can this be merged?

alexcrichton:
@malbarbo hm I'm not sure I could say w/o testing, mind testing before and after this change to see if the targets build correctly?

malbarbo:
I had tested it with both static and dynamic linking. I tested it again after changing `crt_static_respected = false`. Both works as expected.

I had forgotten to commit the -fPIC changes, but I pushed it now.

alexcrichton:
Ok!

@bors: r+

PR #909, opened 2018-01-26
Add posix_spawn bindings

alexcrichton:


bdrewery:
@alexcrichton Were you working on using `posix_spawn` in `Command` or just adding the bindings? I get significant performance improvement on FreeBSD by using `vfork`, so was looking into writing `posix_spawn` support.

alexcrichton:
@bdrewery initially yeah I was curious if we could [implement an optimization](https://github.com/alexcrichton/rust/commit/a7d5852ca13116a146a5f1a695b7bb8493686f53) for `Command` to at least attempt to use `posix_spawn` for some cases, but unfortunately the semantics of `posix_spawn` on a nonexistent executable are different than fork/exec, so it ended up being a dead end for libstd. 

That being said, I think it'd be great to implement this in crates.io!

bdrewery:
@alexcrichton What's the difference you're seeing?

alexcrichton:
`Command::new("nonexistent").spawn()` returns an error today, but with `posix_spawn` it returns a success and the call to `wait()` fails. I wasn't sure how you could actually tell whether the process exec'd correctly or not.

bdrewery:
I think we could just `try_wait` (to use `waitpid(WNOHANG)`) after and check for exit status _127_ here https://github.com/rust-lang/rust/compare/master...alexcrichton:posix-spawn#diff-86e199423278a641f73ed69c42c9f733R40 and return the expected error. Some platforms would allow skipping that such as FreeBSD which just returns _ENOENT_ in errno after `posix_spawn`.

bdrewery:
For context I picked up Rust just last week and the first thing I prototyped hit major overhead in `fork` on FreeBSD https://github.com/rayon-rs/rayon/issues/540 which `vfork` or `posix_spawn` fixes. So I'd really like to see this fixed by default for FreeBSD at least.

alexcrichton:
@bdrewery isn't `try_wait` racy though? We wouldn't know if the process exited quickly or if it failed to spawn I'd imagine. Additionally would that be guaranteed to catch all process spawn failures?

bdrewery:
@alexcrichton yeah true. At least on FreeBSD and DragonFlyBSD because `vfork` is used, the parent process is suspended until the `exec`; but a _ENOENT_/_ENOEXEC_ is returned for those platforms anyway. OpenBSD doesn't use `vfork`, and NetBSD's syscall implementation doesn't appear to have the suspend behavior always. Using `posix_spawn` for FreeBSD/DragonFlyBSD rather than `vfork` is still preferred since there's some special handling for threads in FreeBSD's `posix_spawn` that could not be reproduced otherwise. I'm going to take your commit and polish it on FreeBSD.

bdrewery:
@alexcrichton looking over glibc it looks like its latest `posix_spawn` can return ENOENT in the case you're talking about as it uses a pipe to send the error code back up to the parent https://github.com/bminor/glibc/commit/9ff72da471a509a8c19791efe469f47fa6977410#diff-35e1fecdd831511357340526ff4c5bacR285. There are also some tests for the behavior: https://github.com/bminor/glibc/commit/9ff72da471a509a8c19791efe469f47fa6977410#diff-d5ff29c03cf203a4861e0ed3e2ec63ddR41

https://sourceware.org/bugzilla/show_bug.cgi?id=10354#c12
```
      2. Child must synchronize with parent to enforce stack deallocation and
         to possible return execv issues.
```

I guess the only problem with Linux now would be ensuring its glibc has this newer `posix_spawn` at runtime.

I made a few changes to your implementation (namely the `*sys::os::environ` one was needed and no need to zero objects which have their own initializers):
```
diff --git src/libstd/sys/unix/process/process_unix.rs src/libstd/sys/unix/process/process_unix.rs
index d66c237514..fa66245abb 100644
--- src/libstd/sys/unix/process/process_unix.rs
+++ src/libstd/sys/unix/process/process_unix.rs
@@ -24,7 +24,6 @@ impl Command {
                  -> io::Result<(Process, StdioPipes)> {
         use sys;

-
         const CLOEXEC_MSG_FOOTER: &'static [u8] = b"NOEX";

         let envp = self.capture_env();
@@ -237,7 +236,7 @@ impl Command {
     }

     #[cfg(not(any(target_os = "linux", target_os = "macos")))]
-    fn posix_spawn(&mut self, stdio: &ChildPipes, envp: Option<&CStringArray>)
+    fn posix_spawn(&mut self, _stdio: &ChildPipes, _envp: Option<&CStringArray>)
         -> io::Result<Option<Process>>
     {
         Ok(None)
@@ -280,8 +279,8 @@ impl Command {
         }

         unsafe {
-            let mut file_actions = PosixSpawnFileActions(mem::zeroed());
-            let mut attrs = PosixSpawnattr(mem::zeroed());
+            let mut file_actions = PosixSpawnFileActions(mem::uninitialized());
+            let mut attrs = PosixSpawnattr(mem::uninitialized());

             libc::posix_spawnattr_init(&mut attrs.0);
             libc::posix_spawn_file_actions_init(&mut file_actions.0);
@@ -302,7 +301,7 @@ impl Command {
                                                            libc::STDERR_FILENO))?;
             }

-            let mut set: libc::sigset_t = mem::zeroed();
+            let mut set: libc::sigset_t = mem::uninitialized();
             cvt(libc::sigemptyset(&mut set))?;
             cvt(libc::posix_spawnattr_setsigmask(&mut attrs.0,
                                                  &set))?;
@@ -315,7 +314,7 @@ impl Command {
             cvt(libc::posix_spawnattr_setflags(&mut attrs.0, flags as _))?;

             let envp = envp.map(|c| c.as_ptr())
-                .unwrap_or(sys::os::environ() as *const _);
+                .unwrap_or(*sys::os::environ() as *const _);
             let ret = libc::posix_spawnp(
                 &mut p.pid,
                 self.get_argv()[0],
```

As a sidenote it turns out that `posix_spawn` still has like 12% overhead vs `vfork` on FreeBSD due to the file actions using `malloc`. That's an implementation detail that we'll improve in FreeBSD's libc though.

PR #910, opened 2018-01-26
Add basic thread affinity for MacOS.

seeday:
Doesn't pass libc-test yet, but I'm not quite sure why.

Functions fine in practice though.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -179,6 +179,8 @@ fn main() {
         cfg.header("netinet/if_ether.h");
         cfg.header("sys/proc_info.h");
         cfg.header("sys/kern_control.h");
+        cfg.header("mach/thread_policy.h");
+        cfg.header("mach/mach_init.h");
     }
 
     if bsdlike {
@@ -616,6 +618,10 @@ fn main() {
             // Deprecated on OSX
             "sem_destroy" if apple => true,
             "sem_init" if apple => true,
+
+            // Commented out in header but still callable?
+            "thread_policy_set" if apple => true,
+            "thread_policy_get" if apple => true,
 
             // These functions presumably exist on netbsd but don't look like
             // they're implemented on rumprun yet, just let them slide for now.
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! I think there's CI errors though?

seeday:
I've taken a closer look at the CI errors and figured a few out, but I'm still getting one that I've got no idea where to start. Maybe you've got an idea?

`bad syslog function pointer: rust: 140734905748683 (0x7fff66108ccb) != c 140734905749265 (0x7fff66108f11)`

Somewhat unrelated question. Is it correct to think that the types defined must be some kind of integer? `pub type thread_affinity_policy_t = *mut ::thread_affinity_policy;` fails as an alias, trying to emulate the C equivalent `typedef struct thread_affinity_policy *thread_affinity_policy_t`.

alexcrichton:
Heh I was actually getting that same error about the `syslog` function last night on my local computer. I think it's related to function pointers slightly changing over time due to differences in behavior, and you can sort of see that with the `#[link_name]` and a bunch of other functions on OSX. For now though you can probably ignore it.

Also yeah that sort of `pub type` is fine, you'll just need to add a whitelist to `build.rs` to not check it for signededness.

alexcrichton:
Hm is there any header we can verify this against though?

seeday:
According to the (apparently out of date) [apple documentation](https://developer.apple.com/library/content/releasenotes/Performance/RN-AffinityAPI/index.html), that's where things should be defined. The functions are commented out and moved to another file which I've added, but now the verification fails because of a pointer type weakening on the C side. 

`error: incompatible pointer types returning 'kern_return_t (thread_act_t, thread_policy_flavor_t, thread_policy_t, mach_msg_type_number_t)' (aka 'int (unsigned int, unsigned int, int *, unsigned int)') from a function with result type 'int (*)(thread_t, thread_policy_flavor_t, struct thread_affinity_policy *, mach_msg_type_number_t)' (aka 'int (*)(unsigned int, unsigned int, struct thread_affinity_policy *, unsigned int)'` is the actual error.

The C version uses an int* as thread_policy_t because there's a few different structs for that argument. Is that something we can represent in rust? I tried using traits but those seem to cause 
 other verification issues. I assume enums wouldn't have the right internal structure either.

alexcrichton:
Do the header files have `typedef thread_policy_t int*` or something like that? In general we try to just stick to the headers wherever possible to avoid premature abstractions and such.

alexcrichton:

PR #911, opened 2018-01-30
Add netfilter constants to Linux

faern:
Constants and enums from [`netfilter.h`] and [`nf_tables.h`].

I was not really sure what to do with the enums, for example `nf_tables_msg_types`. But since I could not find a C enum mapped into a Rust enum in this crate I assumed you wanted them as constants instead.

I also wanted to get the `nf_inet_addr` union in here. But I did not find any unions in this crate so I was not sure how you would prefer it represented.

There are of course a billion more netfilter constants and types to map. But this is a good start.

[`netfilter.h`]: https://github.com/torvalds/linux/blob/ead751507de86d90fa250431e9990a8b881f713c/include/uapi/linux/netfilter.h
[`nf_tables.h`]: https://github.com/torvalds/linux/blob/ead751507de86d90fa250431e9990a8b881f713c/include/uapi/linux/netfilter/nf_tables.h

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -247,6 +247,7 @@ fn main() {
             cfg.header("linux/netlink.h");
             cfg.header("linux/magic.h");
             cfg.header("linux/reboot.h");
+            cfg.header("linux/netfilter/nf_tables.h");
 
             if !mips {
                 cfg.header("linux/quota.h");
@@ -255,7 +256,7 @@ fn main() {
     }
     if solaris {
         cfg.header("sys/epoll.h");
-   }
+    }
 
     if linux || android {
         cfg.header("sys/fsuid.h");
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks! Looksl ike CI may be failing though?

faern:
Yup. Was trying to figure out why.
Do you happen to know why the netlink definitions is not included under musl? Can't I communicate over netlink if I statically link? 😕

alexcrichton:
Hm I wonder if it's a missing header file? I'm not too familiar with netlink unfortunately :(

In any case it should be fine to only define the constants on glibc (but still linux) if necessary!

faern:
Yes, it seems `nf_tables.h` is not available with musl. Why, I don't know, but I'm going to skip it for now at least.

Now there is a lot of duplication between Linux and Android. But I don't find any other place where you share values between them, so I guess this is the preferred method.

Issue #915, opened 2018-02-04
IFF_ constants should all be c_int


Susurrus:
#873 added some Linux-specific `IFF_*` constants as `c_short`s, but I believe they should been `c_int`s as all other `IFF_*` constants are `c_int`.

kristate:
@Susurrus @LuoZijun @alexcrichton 
with the #916 merge, can this now be closed?

Susurrus:
#916 didn't change the types for c_int, so no. That will require a 0.3 libc release to do.

gnzlbg:
How come ctest did not catch this ?

alexcrichton:
@gnzlbg constant in C don't really have a type, so ctest can only test the bit pattern of constants, not the types unfortunately :(

gnzlbg:
Ah, I thought that there was a problem with their value.  Maybe one could do a crater run to see what changing these to `c_int` would break ?

alexcrichton:
Seems plausible!

gnzlbg:
I'd be willing to give this a try if somebody submits a PR fixing this.

PR #916, opened 2018-02-04
Add `IFF_*` constants  to linux, android and fuchsia

LuoZijun:
Add constants to android and fuchsia:

*    IFF_TUN
*    IFF_TAP
*    IFF_NO_PI

Type problem: https://github.com/rust-lang/libc/issues/915

so change constants type `libc::c_short` to `libc::c_int` :

*    IFF_*

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -265,7 +265,7 @@ fn main() {
         cfg.header("sys/fsuid.h");
         cfg.header("linux/seccomp.h");
         cfg.header("linux/if_ether.h");
-
+        cfg.header("linux/if_tun.h");
         // DCCP support
         if !uclibc && !musl && !emscripten {
             cfg.header("linux/dccp.h");
@@ -280,7 +280,6 @@ fn main() {
         cfg.header("linux/random.h");
         cfg.header("elf.h");
         cfg.header("link.h");
-        cfg.header("linux/if_tun.h");
         cfg.header("spawn.h");
     }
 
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks! Unfortunately we can't change the types as that'd be a breaking change, but the additions are certainly possible!

LuoZijun:
@alexcrichton ok, updated.

src/unix/notbsd/mod.rs
```diff
@@ -445,6 +445,16 @@ pub const IFF_PORTSEL: ::c_int = 0x2000;
 pub const IFF_AUTOMEDIA: ::c_int = 0x4000;
 pub const IFF_DYNAMIC: ::c_int = 0x8000;
 
+cfg_if! {
+    if #[cfg(target_os = "android")] {
+        pub const IFF_TUN: ::c_int = 0x0001;
```
alexcrichton: Could these be moved into the android folder to avoid the `cfg_if!`?

LuoZijun: okay, no problem.

by the way, should keep `IFF_TUN/IFF_TAP/IFF_NO_PI` type is same as linux ?

alexcrichton: Nah I think it's fine to fix the type for new definitions.

LuoZijun: ok, updated.

LuoZijun:
@alexcrichton Done.

PR #919, opened 2018-02-14
Add termios ioctl constants to NetBSD

Susurrus:
These are all declared as `c_ulong` because that's what BSD's `ioctl()` function is declared with.


Susurrus:
Whoa, when did highfive start being used with `libc`?! Guess I haven't submitted a PR here in a while...

PR #920, opened 2018-02-17
Fix typo in O_TMPFILE definition on linux-sparc64

glaubitz:
While working on the libc bindings for sparc-linux, I stumbled across this typo in the O_TMPFILE definition on linux-sparc64.

Compare with:

> https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/sparc/bits/fcntl.h;hb=16efad5171ac1ac2c8728405f2703045f08c494b#l42
> https://sources.debian.org/src/glibc/2.26-6/sysdeps/unix/sysv/linux/sparc/bits/fcntl.h/#L42

CC @jrtc27


jrtc27:
8 zeros in octal is `8*3 = 24` bits, and 6 zeros in hex is `6*4 = 24` bits too, so they do have the same value. Having said that it still makes sense to match the glibc definition rather than having a single one in octal.

glaubitz:
Right, I completely forgot about octal and was somewhat surprised it worked ;).

Probably makes sense to rephrase the commit message.

Issue #921, opened 2018-02-19
Interest in more netlink constants in libc


jbaublitz:
I realize that netlink is a pretty specific use case but is there any interest in me doing a PR to include some of the standard but less common netlink constants? Currently for my crate, I compile a C extention and link to it from Rust to get these constants in my crate and I'd love to have this incorportated into this crate to keep all of the C dependencies in one place. I'm mainly focusing on the constants that signify netlink attribute types in `<linux/netlink.h>` and `<linux/genetlink.h>` as well as potentially `<linux/rtnetlink.h>` as that is the next part of netlink that I plan to implement in my library. There were a number of netlink constants that were not included in libc when I started that now are so I'd be happy to get the rest of them in there if this is a use case you'd be willing to support.

faern:
I'm working on a PR to include these now. I had to add some netlink structs anyway for another use case so I'll bring in the constants you need. See #922, however, the constants have not been pushed there yet.

faern:
However, since `GENL_ID_GENERATE` was deprecated in 2016 (https://github.com/torvalds/linux/commit/a07ea4d9941af5a0c6f0be2a71b51ac9c083c5e5) I'm going to exclude that one.

jbaublitz:
Thanks for this!

alexcrichton:
@jbaublitz there's certainly interest! Basically if it's in headers and it verifies, it should be fine to add.

kristate:
@jbaublitz @faern What is the current status?

faern:
This is merged and done. I'm pretty sure my PR included all the parts @jbaublitz needed.

alexcrichton:
Great!

PR #922, opened 2018-02-21
Add netlink structs to all notbsd platforms

faern:
Adding the remaining structs from `<linux/netlink.h>`. Sadly very duplicated, but that's how it ended up when added to all "notbsd" platforms except musl. If there is a better way to structure it to reduce duplication I'm all ears!

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -1270,6 +1270,65 @@ pub const POSIX_SPAWN_SETSIGMASK: ::c_int = 0x08;
 pub const POSIX_SPAWN_SETSCHEDPARAM: ::c_int = 0x10;
 pub const POSIX_SPAWN_SETSCHEDULER: ::c_int = 0x20;
 
+pub const NLMSG_NOOP: ::c_int = 0x1;
+pub const NLMSG_ERROR: ::c_int = 0x2;
+pub const NLMSG_DONE: ::c_int = 0x3;
+pub const NLMSG_OVERRUN: ::c_int = 0x4;
+pub const NLMSG_MIN_TYPE: ::c_int = 0x10;
+
+pub const GENL_NAMSIZ: ::c_int = 16;
+
+pub const GENL_MIN_ID: ::c_int = NLMSG_MIN_TYPE;
+pub const GENL_MAX_ID: ::c_int = 1023;
+
+pub const GENL_ADMIN_PERM: ::c_int = 0x01;
+pub const GENL_CMD_CAP_DO: ::c_int = 0x02;
+pub const GENL_CMD_CAP_DUMP: ::c_int = 0x04;
+pub const GENL_CMD_CAP_HASPOL: ::c_int = 0x08;
+cfg_if! {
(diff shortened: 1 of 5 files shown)


faern:
Working on adding some `genetlink.h` constants as well. So consider this PR work in progress for a little longer!

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -517,12 +555,6 @@ pub const NLM_F_EXCL: ::c_int = 0x200;
 pub const NLM_F_CREATE: ::c_int = 0x400;
 pub const NLM_F_APPEND: ::c_int = 0x800;
 
-pub const NLMSG_NOOP: ::c_int = 0x1;
-pub const NLMSG_ERROR: ::c_int = 0x2;
-pub const NLMSG_DONE: ::c_int = 0x3;
-pub const NLMSG_OVERRUN: ::c_int = 0x4;
-pub const NLMSG_MIN_TYPE: ::c_int = 0x10;
```
faern: Was able to move these up one level. Covers more platforms and with less duplication.

faern:
This PR now fixes #921. Should include everything to make https://github.com/jbaublitz/neli/blob/master/src/c/netlink.c obsolete. Ping @jbaublitz

src/unix/notbsd/linux/mod.rs
```diff
@@ -1264,6 +1270,65 @@ pub const POSIX_SPAWN_SETSIGMASK: ::c_int = 0x08;
...
+pub const GENL_NAMSIZ: ::c_int = 16;
+
+pub const GENL_MIN_ID: ::c_int = NLMSG_MIN_TYPE;
+pub const GENL_MAX_ID: ::c_int = 1023;
+
+pub const GENL_ADMIN_PERM: ::c_int = 0x01;
+pub const GENL_CMD_CAP_DO: ::c_int = 0x02;
+pub const GENL_CMD_CAP_DUMP: ::c_int = 0x04;
+pub const GENL_CMD_CAP_HASPOL: ::c_int = 0x08;
+cfg_if! {
```
alexcrichton: Could the location of these definitions get tweaked a bit to avoid the `cfg_if!` annotations here?

alexcrichton:
Looks good to me, thanks! Just one small nit but otherwise seems good to go

src/unix/notbsd/linux/mod.rs
faern: Yes, done!

faern:
Now the tests started to fail on a constant added by me in a previous PR. The value of the `NFT_*_MAXNAMELEN` constants seems to have been increased from 32 to 256 on most platforms. So I had to add some `cfg_if!` trickery to get it to pass the tests again.

kristate:
@faern

> Now the tests started to fail on a constant added by me in a previous PR. The value of the `NFT_*_MAXNAMELEN` constants seems to have been increased from 32 to 256 on most platforms.

Okay, once #924 or 95befb4b20b39d2b257ed69a10a22ab68995fc12 lands, I will rebase and push #924. thanks.

PR #925, opened 2018-02-25
add IPV6_MULTICAST_IF and IPV6_MULTICAST_HOPS

bluejekyll:
These were missing from the net2 crate. Adding them here so that the higher functions can be added to net2 later. They are necessary for declaring the interface for ipv6 multicast packets (route), and limits on the number of hops an ipv6 packet can travel

Linux reference: https://github.com/torvalds/linux/blob/master/include/uapi/linux/in6.h#L168-L170
macOS reference: https://github.com/apple/darwin-xnu/blob/master/bsd/netinet6/in6.h#L538-L540


bluejekyll:
Thank you!

PR #926, opened 2018-02-25
Add netfilter/nftables fields

faern:
Adding a few missing `NF_` constants to more complete the story around that header. Then most importantly add a lot of `NFT_` constants needed in my nftables library. Some were added back in #911, but as the library grew I realized I needed more constants.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,6 +148,9 @@ cfg_if! {
         pub enum FILE {}
         pub enum fpos_t {} // TODO: fill this out with a struct
 
+        pub const INT_MIN: c_int = -2147483648;
+        pub const INT_MAX: c_int = 2147483647;
+
         extern {
             pub fn isalnum(c: c_int) -> c_int;
             pub fn isalpha(c: c_int) -> c_int;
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -148,6 +148,9 @@ cfg_if! {
         pub enum FILE {}
         pub enum fpos_t {} // TODO: fill this out with a struct
 
+        pub const INT_MIN: c_int = -2147483648;
+        pub const INT_MAX: c_int = 2147483647;
```
faern: These are only two of the limits from `limits.h`. Needed since I need to reach the min and max integers for other constants within `libc`. I could not use `c_int:min_value()` since that does not work as a constant function on Rust 1.0. And I could not use `core::i32::MIN` because I could not use `core` on some platforms. Is the latter expected or did I do something wrong?

alexcrichton: Nah yeah redefining these here are fine, but we can probably relax CI in the future!

alexcrichton:
@bors: r+

Thanks!

PR #931, opened 2018-02-28
Add posix_spawn bindings for FreeBSD

bdrewery:


bdrewery:
I think these would apply for OpenBSD as well but I only am able to test FreeBSD right now.
This may not apply for NetBSD as their ABI is slightly different. For FreeBSD and OpenBSD posix_spawnattr_t is an opaque pointer:
```
typedef struct __posix_spawnattr                *posix_spawnattr_t;
typedef struct __posix_spawn_file_actions       *posix_spawn_file_actions_t;
```
But on NetBSD it's not:
```
struct posix_spawnattr {
        short                   sa_flags;
        pid_t                   sa_pgroup;
        struct sched_param      sa_schedparam;
        int                     sa_schedpolicy;
        sigset_t                sa_sigdefault;
        sigset_t                sa_sigmask;
};

enum fae_action { FAE_OPEN, FAE_DUP2, FAE_CLOSE };
typedef struct posix_spawn_file_actions_entry {
        enum fae_action fae_action;

        int fae_fildes;
        union {
                struct {
                        char *path;
#define fae_path        fae_data.open.path
                        int oflag;
#define fae_oflag       fae_data.open.oflag
                        mode_t mode;
#define fae_mode        fae_data.open.mode
                } open;
                struct {
                        int newfildes;
#define fae_newfildes   fae_data.dup2.newfildes
                } dup2;
        } fae_data;
} posix_spawn_file_actions_entry_t;

struct posix_spawn_file_actions {
        unsigned int size;      /* size of fae array */
        unsigned int len;       /* how many slots are used */
        posix_spawn_file_actions_entry_t *fae;
};

typedef struct posix_spawnattr          posix_spawnattr_t;
typedef struct posix_spawn_file_actions posix_spawn_file_actions_t;
```

Do the field and sizes matter for declaring the types in the NetBSD case or would this still work?
```
pub type posix_spawnattr_t = *mut ::c_void;         
```

alexcrichton:
Field sizes and such do need to be correct, but looks like this passed CI so it's all verified correct anyway!

@bors: r+

PR #933, opened 2018-02-28
Add IPV6_UNICAST_HOPS to fuchsia module

### LINKED ISSUE #932: IPV6_UNICAST_HOPS not defined on Fuchsia
#925 added a bunch of `IPV6_MULTICAST_*` definitions, and _mostly_ applied them consistently between notbsd unix and fuchsia, but seems to have left off `IPV6_UNICAST_HOPS` in particular from the fuchsia module. From a shallow look, it seems like a simple oversight, the fix is probably to add it with the same value as notbsd.

This is causing net2 0.2.32 to fail to compile on Fuchsia, I'm temporarily working around this by pinning to net2 0.2.31.

Happy to provide a PR if it's the simple one-liner it appears to be, but wanted to sanity-check first.

ping @bluejekyll as author of the patch

raphlinus:
Recent change #925 added constants for IPV6 multicast, but seems to have inadvertently left off IPV6_UNICAST_HOPS from the fuchsia module.

Fixes #932

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -1486,6 +1486,7 @@ pub const IP_HDRINCL: ::c_int = 3;
 pub const IP_ADD_MEMBERSHIP: ::c_int = 35;
 pub const IP_DROP_MEMBERSHIP: ::c_int = 36;
 pub const IP_TRANSPARENT: ::c_int = 19;
+pub const IPV6_UNICAST_HOPS: ::c_int = 16;
 pub const IPV6_MULTICAST_IF: ::c_int = 17;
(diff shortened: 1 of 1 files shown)


raphlinus:
Note: I confirmed that this is the correct constant in [our fork of musl](https://fuchsia.googlesource.com/zircon/+/master/third_party/ulib/musl/include/netinet/in.h#326).

bluejekyll:
Thanks for fixing this! Definitely an oversight.

bluejekyll:
Hm, this may cause issues on the Redox targets as well.

PR #934, opened 2018-03-01
Add passwd/group APIs needed for nix-rust/nix#864

ctrlcctrlv:
Hope I did this right. I only added platforms I could personally test. . . 

cc: @gnzlbg

### DIFF
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -50,6 +50,13 @@ s! {
         pub pw_fields: ::c_int,
     }
 
+    pub struct group {
+        pub gr_name: *mut ::c_char,
+        pub gr_passwd: *mut ::c_char,
+        pub gr_gid: ::gid_t,
+        pub gr_mem: *mut *mut ::c_char
+    }
+
     pub struct ifaddrs {
         pub ifa_next: *mut ifaddrs,
         pub ifa_name: *mut ::c_char,
@@ -388,6 +395,10 @@ extern {
     pub fn getpwent() -> *mut passwd;
     pub fn setpwent();
     pub fn endpwent();
+    pub fn setgrent();
+    pub fn endgrent();
+    pub fn getgrent() -> *mut group;
+
     pub fn getprogname() -> *const ::c_char;
     pub fn setprogname(name: *const ::c_char);
     pub fn getloadavg(loadavg: *mut ::c_double, nelem: ::c_int) -> ::c_int;
@@ -513,6 +524,18 @@ extern {
                       buf: *mut ::c_char,
                       buflen: ::size_t,
                       result: *mut *mut passwd) -> ::c_int;
+    #[cfg_attr(target_os = "netbsd", link_name = "__getpwent_r50")]
+    #[cfg_attr(target_os = "solaris", link_name = "__posix_getpwent_r")]
+    pub fn getpwent_r(pwd: *mut passwd,
+                      buf: *mut ::c_char,
+                      buflen: ::size_t,
+                      result: *mut *mut passwd) -> ::c_int;
+    #[cfg_attr(target_os = "netbsd", link_name = "__getgrent_r50")]
+    #[cfg_attr(target_os = "solaris", link_name = "__posix_getgrent_r")]
+    pub fn getgrent_r(grp: *mut group,
+                      buf: *mut ::c_char,
+                      buflen: ::size_t,
+                      result: *mut *mut group) -> ::c_int;
     #[cfg_attr(all(target_os = "macos", target_arch ="x86"),
                link_name = "sigwait$UNIX2003")]
     #[cfg_attr(target_os = "solaris", link_name = "__posix_sigwait")]
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -1,5 +1,4 @@
 use dox::{mem, Option};
-use unix::group;
 
 pub type wchar_t = i32;
 pub type off_t = i64;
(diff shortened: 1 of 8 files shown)


ctrlcctrlv:
The build is failing on some platforms due to automatically generated tests, I'm not sure how to fix that...

gnzlbg:
> I only added platforms I could personally test. . .

Don't worry about this that much. This is what CI is for :)

src/unix/bsd/mod.rs
```diff
@@ -388,6 +389,10 @@ extern {
     pub fn getpwent() -> *mut passwd;
     pub fn setpwent();
     pub fn endpwent();
+    pub fn setgrent();
+    pub fn endgrent();
+    pub fn getgrent() -> *mut group;
```
gnzlbg: prefer `::unix::group` here, removing the `use unix::group` above

src/unix/notbsd/linux/mod.rs
```diff
@@ -1468,9 +1469,13 @@ extern {
     pub fn setpwent();
     pub fn endpwent();
     pub fn getpwent() -> *mut passwd;
+    pub fn setgrent();
+    pub fn endgrent();
+    pub fn getgrent() -> *mut group;
```
gnzlbg: same here

src/unix/bsd/mod.rs
```diff
@@ -513,6 +518,18 @@ extern {
...
                       result: *mut *mut passwd) -> ::c_int;
+    #[cfg_attr(target_os = "netbsd", link_name = "__getpwent_r50")]
+    #[cfg_attr(target_os = "solaris", link_name = "__posix_getpwent_r")]
+    pub fn getpwent_r(pwd: *mut passwd,
+                      buf: *mut ::c_char,
+                      buflen: ::size_t,
+                      result: *mut *mut passwd) -> ::c_int;
+    #[cfg_attr(target_os = "netbsd", link_name = "__getgrent_r50")]
+    #[cfg_attr(target_os = "solaris", link_name = "__posix_getgrent_r")]
+    pub fn getgrent_r(grp: *mut group,
```
gnzlbg: getgrent_r and getpwent_r are not available on macosx (or looks like it), you should move them to the netbsd and solaris modules.

src/unix/notbsd/linux/mod.rs
```diff
@@ -1826,6 +1831,18 @@ extern {
...
+    pub fn getpwent_r(pwd: *mut passwd,
+                      buf: *mut ::c_char,
+                      buflen: ::size_t,
+                      result: *mut *mut passwd) -> ::c_int;
+    #[cfg_attr(target_os = "netbsd", link_name = "__getgrent_r50")]
+    #[cfg_attr(target_os = "solaris", link_name = "__posix_getgrent_r")]
+    pub fn getgrent_r(grp: *mut group,
+                      buf: *mut ::c_char,
+                      buflen: ::size_t,
+                      result: *mut *mut group) -> ::c_int;
```
gnzlbg: getpwent_r and getgrent_r are not available in `musl`. I think you have to move them to src/unix/notbsd/linux/other/mod.rs

src/unix/bsd/mod.rs
```diff
@@ -388,6 +389,10 @@ extern {
     pub fn getpwent() -> *mut passwd;
     pub fn setpwent();
     pub fn endpwent();
+    pub fn setgrent();
```
gnzlbg: setgrent is not available in freebsd (or it looks like this is the case), you are going to have to put it in the folders of all bsd targets except for the freebsd one :/

alexcrichton:

src/unix/bsd/mod.rs
ctrlcctrlv: setgrent is available in FreeBSD. What gives you the impression it isn't?

https://github.com/freebsd/freebsd/blob/d5842c73935af7f7a7eeaeaec75ad8f3cf6cdf50/include/grp.h#L80

gnzlbg: > setgrent is available in FreeBSD. What gives you the impression it isn't?

Indeed, this buildbot: https://travis-ci.org/rust-lang/libc/jobs/347737123#L1402

Says that the pointer types are incompatible. It looks like its signature was changed in: https://github.com/freebsd/freebsd/commit/786c3926a65879b5dd6c301ca9ccdfda486fd195

from `int setgrent()` to `void setgrent()`. That commit was 2 years ago, but it might be that we are testing against a FreeBSD release that is older as the release where that commit was included. 

You might want to change the signature to `int setgrent()` and see if that fixes the build. If so, please open an issue about updating the FreeBSD version (we can fix that and fix setgrent later).

gnzlbg: If it turns out to be the case that changing the signature to `int setgrent()` fixes this issue, what you can also do is change it back afterwards to `void setgrent()` and whitelist the function here:  https://github.com/rust-lang/libc/blob/master/libc-test/build.rs#L606

By adding a match arm like this:

```rust
// FIXME: need to upgrade FreeBSD version. Issue: rust-lang/libc/xxxxxx
"setgrent" if freebsd => true,
```

ctrlcctrlv:
@alexcrichton The build is now passing :)

gnzlbg:
lgtm

alexcrichton:
Thanks! I think lockfile may need another update though?

ctrlcctrlv:
@alexcrichton Done

gnzlbg:
@ctrlcctrlv nice thanks :) I don't know if nix-rust has some more open issues on libc for missing interfaces (they used to have some at least), but now you guys should be able to quickly add to libc whatever you find its missing :)

PR #936, opened 2018-03-01
Link against kevent@FBSD_1.0 to fix ABI compat with FreeBSD12.

bdrewery:
struct kevent was modified in FreeBSD12.  The @FBSD_1.0 symbol supports the old
structure ABI still.

This allows the `mio` crate tests to now pass on FreeBSD12.


bdrewery:
To be clear, this symbol is available on all FreeBSD versions and not just 12.

PR #937, opened 2018-03-02
Use pre-ino64 FreeBSD symbols to resolve binary compatibility.

bdrewery:
This follows the same method as other platforms like OSX and NetBSD.

This will fix rustup and building from git (once libc is updated for bootstrap)
on FreeBSD12 post-ino64 in https://github.com/freebsd/freebsd/commit/f713b08c027de35b0b8ed9990eabc32e100bee7a.
It also avoids having to hotpatch the stage0 compiler, and HOME/.cargo
libc files on FreeBSD12 to build rust.

The only real pitfall is that this will prevent interaction with inodes that
have an ino_t above the 32-bit limit due to truncation.  On the other hand
Rust won't work at all on 12 without doing this currently.  In general
it should not be a problem for users and if they need 64-bit ino_t they
can use a patched libc, rather than the current state of affairs in
requiring a patched libc to use Rust on 12.

A better, or complementary, approach would be something like proposed in
https://github.com/rust-lang/rfcs/pull/2048 to allow targetting a specific
version of FreeBSD. This would allow Rust to default to this compatibility
mode by targetting FreeBSD10 and still allow targetting FreeBSD12 for 64-bit
ino_t.

The symbol versions used were taken from the old version in
https://github.com/freebsd/freebsd/commit/f713b08c027de35b0b8ed9990eabc32e100bee7a#diff-61a32fcfb7ecd4517665fed591813c57
and
https://github.com/freebsd/freebsd/commit/f713b08c027de35b0b8ed9990eabc32e100bee7a#diff-7f67ccf8b5f44ff2f54eaab0207abb8d.

The scope of functions versioned here differs from other platforms as
not all structs were modified that were on others, such as DIR for
`opendir`, `telldir`, etc.  Only functions using dirent, stat, glob_t,
and dev_t need the changes.

Fixes https://github.com/rust-lang/rust/issues/42681

### DIFF
--- a/src/unix/bsd/freebsdlike/mod.rs
+++ b/src/unix/bsd/freebsdlike/mod.rs
@@ -1151,6 +1151,7 @@ extern {
 
     pub fn fdopendir(fd: ::c_int) -> *mut ::DIR;
 
+    #[cfg_attr(target_os = "freebsd", link_name = "mknodat@FBSD_1.1")]
     pub fn mknodat(dirfd: ::c_int, pathname: *const ::c_char,
                   mode: ::mode_t, dev: dev_t) -> ::c_int;
     pub fn mkfifoat(dirfd: ::c_int, pathname: *const ::c_char,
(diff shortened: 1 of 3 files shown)


bdrewery:
FYI @dumbbell @asomers

alexcrichton:
@bors: r+

Awesome, thanks @bdrewery!

FWIW when we can I think it'd be fine to add things like `stat64` and such like we have on Linux. Although those symbols probably don't actually exist in FreeBSD we could probably get away with pretending they do in this crate (using `#[link_name]`)

PR #941, opened 2018-03-07
Added ucred (sockpeercred) for OpenBSD and Bitrig

wictory:
OpenBSD and Bitrig has `struct ucred`implemented in the way it is used in `getsockopt(2)` with `SOL_SOCKET` and `SO_PEERCRED`. However, it is called `struct sockpeercred`.

Source: https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/sys/sys/socket.h?rev=1.65&content-type=text/x-cvsweb-markup

I added a type alias for `ucred` to enable Linux compatibility, but I'm not sure this is the correct way to do it.

### DIFF
--- a/src/unix/bsd/netbsdlike/openbsdlike/mod.rs
+++ b/src/unix/bsd/netbsdlike/openbsdlike/mod.rs
@@ -14,6 +14,7 @@ pub type pthread_cond_t = *mut ::c_void;
 pub type pthread_condattr_t = *mut ::c_void;
 pub type pthread_rwlock_t = *mut ::c_void;
 pub type pthread_rwlockattr_t = *mut ::c_void;
+pub type ucred = ::sockpeercred;
 
 s! {
     pub struct dirent {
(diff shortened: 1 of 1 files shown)


alexcrichton:
@bors: r+

Thanks!

semarie:
if I have no problem with `struct sockpeercred` addition, I don't
understand the type alias with `ucred` name.

it is confusing as this struct name already exists in OpenBSD but for
something else:
https://github.com/openbsd/src/blob/master/sys/sys/ucred.h#L43

I found accepting such alias is really bad: it makes libc to behave as
compatibility layer (which it is not), and it makes the real `struct
ucred` unable to be used in libc if somebody needs it some day.

Did I miss something ?

wictory:
I don't think you missed anything. I wanted the `SO_PEERCRED` functionality available in OpenBSD from rust. The `pub type...` thing is a hack to achieve direct compatibility with software written with Linux in mind because the interfaces seemed rather similar and the type alias would fix everything out of the box. However, I realize now that this is a mistake for multiple reasons. From my point of view the `ucred` type should be retracted.

alexcrichton:
@semarie without automatic validation we don't really have any way of knowing whether anything added for OpenBSD is correct, so I'm just r+'ing PRs.

PR #942, opened 2018-03-07
Arguments of pledge(2) changed.

wictory:
The arguments of pledge(2) have changed. However, the changes have to yet been propagated to Bitrig.

https://marc.info/?l=openbsd-tech&m=151268831628549&w=2


semarie:
@alexcrichton did you see you accepted a breaking change ?

https://github.com/openbsd/src/commit/4ea7ed56bd1c71560696216439fcb657286d6443

the second argument switched from `const char **` to `const char *` in C
and from `*mut *const ::c_char` to `*const ::c_char` in Rust.

alexcrichton:
Yes I did notice this was breaking, and figured it's ok to accept as OpenBSD isn't validated right now so breakage is acceptable.

PR #948, opened 2018-03-12
Fix test suite on FreeBSD 11

wezm:
This PR addresses failures in the test suite on FreeBSD amd64 11.1-RELEASE-p6:

* Exclude `NET_MAXID` from tests as it was [removed from FreeBSD](https://svnweb.freebsd.org/base?view=revision&revision=262489).
* Update the definition of `blksize_t` to match [its FreeBSD definition](https://github.com/freebsd/freebsd/blob/095a5e9be73cf121b1f54cf7c2099d549843732c/sys/sys/_types.h#L38).
* Update various rlimit constants to match [their FreeBSD definition](https://github.com/freebsd/freebsd/blob/095a5e9be73cf121b1f54cf7c2099d549843732c/sys/sys/resource.h#L119).
* Update various locale constants to match [their FreeBSD definitions](https://github.com/freebsd/freebsd/blob/e086fcfb8c1cbcf7df952c4932d2a991447c6bd9/include/xlocale/_locale.h#L37).

With these changes all test pass.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -509,6 +509,7 @@ fn main() {
             "CTL_MAXID" |
             "KERN_MAXID" |
             "HW_MAXID" |
+            "NET_MAXID" |
             "USER_MAXID" if freebsd => true,
 
             // These constants were added in FreeBSD 11
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -3,7 +3,7 @@ pub type clock_t = i32;
 pub type ino_t = u32;
 pub type lwpid_t = i32;
 pub type nlink_t = u16;
-pub type blksize_t = u32;
+pub type blksize_t = i32;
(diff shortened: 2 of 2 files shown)


wezm:
So CI failed due to it running FreeBSD 10 and not having some of the constants I added (`RLIMIT_KQUEUES`, `RLIMIT_UMTXP`). This appears to be an issue in https://github.com/rust-lang/libc/pull/799#issuecomment-335807786 as well. I can look into generating an updated qemu image if that's the right way to go about it.

alexcrichton:
Thanks! Looks like CI may be failing though?

wezm:
@alexcrichton yep see previous comment. Let me know the best way to proceed.

alexcrichton:
Ah sorry was a bit too tired! I'd be fine either recreating the image yeah! If you need help with that I *think* the old steps are documented in the readme in the `ci` dir, but I can try to help out otherwise.

If that fails though then we can also whitelist the conflicting constants to not be checked.

wezm:
Merged into #950

PR #949, opened 2018-03-12
redefine pthread_t for l4re-uclibc

humenda:


humenda:
I was forced to add a other.rs to unix/uclibc/x86_64, because the style script forbids the usage of #[cfg], please let me know if there is a better solution.

PR #955, opened 2018-03-24
Add MCL_CURRENT and MCL_FUTURE to s390.

afck:


afck:
I'd like to [add `mlockall` to nix](https://github.com/nix-rust/nix/pull/876), and it turns out these two constants are missing in libc from s390.

If I'm not mistaken, s390 uses the generic Linux definitions, which set these constants to 1 and 2? I apologize if I'm misreading this!

alexcrichton:
@bors: r+

Thanks!

PR #959, opened 2018-04-04
Add PTHREAD_MUTEX_ADAPTIVE_NP for glibc

glandium:


glandium:
Actually, let me merge this with #960.

PR #960, opened 2018-04-04
Add PTHREAD_*_MUTEX_INITIALIZER_NP for glibc

glandium:
`pthread_mutex_t` varies across architectures, in several ways:
- endianness alters the ordering of bytes, since the contents of the
  struct are larger than 8-bit.
- its length varies.
- the location of the mutex kind (`PTHREAD_MUTEX_RECURSIVE`,
  `PTHREAD_MUTEX_ERRORCHECK` or `PTHREAD_MUTEX_ADAPTIVE_NP`) varies
  between 32-bit and 64-bit: On 32-bit architectures, it is preceded by
  three int/unsigned int, while on 64-bit architectures, it is preceded
  by four of them.

These initializers are only available from <pthread.h> when _GNU_SOURCE
is defined.


glandium:
rust 1.0 didn't like it:
```
src/unix/notbsd/linux/other/b64/not_x32.rs:1:5: 1:41 error: unresolved import
src/unix/notbsd/linux/other/b64/not_x32.rs:1 use unix::notbsd::linux::pthread_mutex_t;
                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

glandium:
Working around the rust 1.0 issue is an abuse of style.rs, but in the end, I think it's much better this way.

alexcrichton:
Could the `cfg_if!` be avoided and instead these definitions copied to the respective modules?

glandium:
> Could the cfg_if! be avoided and instead these definitions copied to the respective modules?

See the original patch and how it blantantly failed to build on rust 1.0. I also find it's a mess to have this spread across multiple files.

alexcrichton:
Hm I no longer see the logs, but presumably there exist some construction that works, right?

glandium:
Apparently, rust 1.0 didn't support `use` of ancestor module items. So it's not possible to `use unix::notbsd::linux::pthread_mutex_t` from `unix::notbsd::linux::something`. The only work around I can think of is to move and duplicate the definition of `pthread_mutex_t` in all the submodules. I think there's already way too much duplication in the libc crate, I don't think this is a great idea.

glandium:
I tried `::foo` already, and that didn't work:
```
src/unix/notbsd/linux/other/b64/not_x32.rs:1:5: 1:43 error: unresolved import
src/unix/notbsd/linux/other/b64/not_x32.rs:1 use ::unix::notbsd::linux::pthread_mutex_t;
                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

alexcrichton:
Er yes, I mean in the code itself, aka `fn foo(a: ::c_int)` instead of `fn foo(a: c_int)`

glandium:
`::pthread_mutex_t` doesn't work either.

glandium:
Ah ok, I could make it work.

glandium:
ping @alexcrichton

alexcrichton:
@glandium hm this still has all the definitions in one location? Could the definitions not be pushed down into their respective modules and avoid the need for `cfg_if!`?

glandium:
Argh, it looks like I rebased and pushed the wrong branch after #962.

PR #963, opened 2018-04-05
Use #[repr(align(32))] instead of private field for in6_addr.

tmccombs:
This allows constructiong in6_addr instances as a constant from other
crates.

See https://github.com/rust-lang/rust/issues/44582#issuecomment-378488339


tmccombs:
hmm, I just realized there are travis builds for older versions of rust. If libc needs to support older version of rust, this might not work.

PR #964, opened 2018-04-06
Default RHS to Self for Div and Shl

npmccallum:
This fixes a consistency issue with the other operator traits.

### DIFF
--- a/src/dox.rs
+++ b/src/dox.rs
@@ -68,13 +68,13 @@ mod imp {
     }
 
     #[lang = "div"]
-    pub trait Div<RHS> {
+    pub trait Div<RHS=Self> {
         type Output;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! These are just used during documentation, however, so is this change required for correctness?

npmccallum:
@alexcrichton I assumed it would be required for correct documentation. :)

alexcrichton:
Oh we don't actually document these traits as part of the build, they're just required to document other things (it's sort of a hack today)

npmccallum:
I don't take this patch to be urgent. I just took it to be easy while I was fixing the actual code in the rust bug. :)

https://github.com/rust-lang/rust/pull/49630

PR #970, opened 2018-04-11
Refresh Cargo.lock

### LINKED ISSUE #968: PTHREAD_*_MUTEX_INITIALIZER_NP constants added in #960 are not tested
Because ctest never sets `target_endian`, no unit tests are generated for the constants I added in #960.
Before doing any release of libc with those constants, ctest should be updated with a version that has https://github.com/alexcrichton/ctest/pull/29 merged.

glandium:
Fixes #968.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "cc"
-version = "1.0.8"
+version = "1.0.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


glandium:
Funnily enough, this reveals problems in statvfs.

glandium:
So big endian glibc apparently doesn't have the same statvfs as big endian musl, and 3df005724 broke the former by adding the latter. Then I guess this was cargo culted to statvfs64 on mips glibc in a36da11fb.

alexcrichton:
@bors: r+

Thanks!

PR #971, opened 2018-04-11
Make netlink struct fields public

faern:
I realized my netlink structs were a bit unusable if their fields are not public :man_facepalming: 

Follow up and improvement to #922

### DIFF
--- a/src/unix/notbsd/android/mod.rs
+++ b/src/unix/notbsd/android/mod.rs
@@ -187,47 +187,47 @@ s! {
     }
 
     pub struct genlmsghdr {
-        cmd: u8,
-        version: u8,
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks! Looks like some platforms are failing though?

faern:
@alexcrichton The powerpc and mips fails are completely unrelated to this code from what I can see, I have not touched the `statvfs` struct. 
For 32 bit iOS I can't see the error, all it prints is that it does fail? But I doubt just making things public would break anything? All memory alignment etc should be identical to before this PR.

alexcrichton:
@bors: r+

Ah ok, those may be fixed on master then

PR #972, opened 2018-04-15
Fix undefined-behavior on MacOSX structs in stdbuilds 

gnzlbg:
Some MacOSX structs have an incorrect layout that results in undefined behavior. This is because on `x86_64` the MacOSX kernel headers define these using `#pragma pack 4`. 

This PR fixes their layout using `repr(packed(4))` . Since it is only available on nightly, it is only enabled for stdbuilds .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,9 @@ env:
     secure: "e2/3QjgRN9atOuSHp22TrYG7QVKcYUWY48Hi9b60w+r1+BhPkTseIJLte7WefRhdXtqpjjUJTooKDhnurFOeHaCT+nmBgiv+FPU893sBl4bhesY4m0vgUJVbNZcs6lTImYekWVb+aqjGdgV/XAgCw7c3kPmrZV0MzGDWL64Xaps="
 matrix:
   include:
-    # 1.0.0 compat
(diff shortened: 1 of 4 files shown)


gnzlbg:
@alexcrichton this would require updating the stable Rust version used in CI. Is there a policy about that?

alexcrichton:
What would it require updating to?

gnzlbg:
I think `attr_literals` but I'll bisect the first version that works.

gnzlbg:
An alternative could be tuning repr(packed) syntax to something that is compatible with 1.0, maybe instead of `repr(packed(N))` we could add `repr(packed1)`, `repr(packed4)`, etc.

gnzlbg:
@alexcrichton I can also put these two structs into two different files, and conditionally include one or the other depending on the `unstable` cargo feature.

alexcrichton:
Nah Rust 1.13.0 is ancient at this point (aka pre-stable-Serde) so I doubt anything before that is still in use. If you'd like to raise the minimum version in CI to 1.13.0 I think we'll be good to go!

gnzlbg:
@alexcrichton done !

PR #974, opened 2018-04-16
feat: add fread_unulock for linux

hawkingrei:
add fread_unulock for linux


alexcrichton:
Thanks! Could this instead though be placed in the Linux-specific module to avoid the `#[cfg]`?

alexcrichton:
Looks good! I think there may be a style issue though?

hawkingrei:
Fix all

PR #985, opened 2018-05-10
Add sem_getvalue to unix

longshorej:
Adds `sem_getvalue` for supported platforms.

Reference: https://www.gnu.org/software/libc/manual/html_node/Semaphores.html


longshorej:
CI failed for sparc because https://cdimage.debian.org/cdimage/ports/debian-9.0-sparc64-NETINST-1.iso is serving up a 404. Any advice on how to fix, or do we ignore?

alexcrichton:
Oh dear! Want to move that to `allow_failures` in Travis and then we can fix that up later and land this PR?

alexcrichton:
Hm so OSX is somewhat spurious, but want to move s390x into ignored as well?

PR #990, opened 2018-05-22
android: add POSIX sysconf constants

lucab:
Ref: https://android.googlesource.com/platform/bionic/+/1c19194c9d2518dbe86973cd313a277ecb70d75c%5E%21/libc/include/sys/sysconf.h


asomers:
@lucab they weren't included simply because I didn't know which Android sdk to use as a reference.

lucab:
I think it is safe to add them now. The oldest supported android API at the moment is from 2015, and this commit predates it. Worst case is that older bionic will return `EINVAL` on these.

PR #991, opened 2018-05-23
Add `pthread_sigqueue` to linux/other

martinhath:
I'm not too sure about where to put this, as it seems to be a GNU extension, but other pthread functions that were also GNU extensions, like `pthread_rwlockattr_setkind_np` was in the same file, so I just put it directly above.

### DIFF
--- a/src/unix/notbsd/linux/other/mod.rs
+++ b/src/unix/notbsd/linux/other/mod.rs
@@ -864,6 +864,9 @@ extern {
                                          val: *mut ::c_int) -> ::c_int;
     pub fn pthread_rwlockattr_setkind_np(attr: *mut ::pthread_rwlockattr_t,
                                          val: ::c_int) -> ::c_int;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Looks like there's some style errors on CI though?

alexcrichton:
I think that error is because in C the signature is different, maybe sigval is taken by a pointer?

martinhath:
Both are supposed to be `sigval`, but in C the `sigval` is a
```c
union sigval {
    int     sival_int;
    void   *sival_ptr;
};
```

The current `sigval` that we have is just the `void *` variant, as shown in my previous comment. It seems that enum arguments are a problem for `libc` in any case? `rg "union"` shows quite a few "Actually a union of ..."

alexcrichton:
Looks like on CI the signature may still not match? 

Especially with signals I've found that the documented standard and the actual headers tend to differ a bit...

alexcrichton:

PR #999, opened 2018-05-31
add struct ifmap

canndrew:


semarie:
`struct ifmap` doesn't exist in BSDs (checked OpenBSD, and quickly grep
source tree of DragonFly, FreeBSD and NetBSD).

canndrew:
@semarie Thanks :). I already moved it down into the `notbsd` part of the hierarchy.

alexcrichton:
Should the fields here be `pub` so they can be accessed externally?

canndrew:
> Should the fields here be `pub` so they can be accessed externally?

Yes. Good catch :p

canndrew:
Worth a shot.

@alexcrichton: this one is ready to go as well

alexcrichton:

PR #1000, opened 2018-05-31
add __caddr_t typedef

canndrew:


canndrew:
Anyone know how to fix these errors on linux?:

```
cargo:warning=/checkout/target/x86_64-unknown-linux-gnu/debug/build/libc-test-c97fd473a3299b51/out/main.c: In function '__test_signed___caddr_t':
cargo:warning=/checkout/target/x86_64-unknown-linux-gnu/debug/build/libc-test-c97fd473a3299b51/out/main.c:825:42: error: ordered comparison of pointer with integer zero [-Werror=extra]
cargo:warning=                 return (((__caddr_t) -1) < 0);
```

alexcrichton:
I think you'll need to tweak `libc-test/build.rs` which has a few `skip_signededness` checks for typedefs like this

canndrew:
I might just not bother with this one. Looks like it's extremely non-portable, it's only used in one struct (and not even used there consistently - other versions of headers I found online use `char *` or `void *` instead), also a SO answer says that it's obsolete and should never be used.

PR #1003, opened 2018-05-31
add some missing netlink symbols

canndrew:


canndrew:
This is blocked on https://github.com/rust-lang/libc/pull/1004

canndrew:
This in now blocked on #1007

canndrew:
error[E0015]: calls in constants are limited to struct and enum constructors
        --> src/unix/notbsd/linux/mod.rs:1252:6
         |
    1252 |     (mem::size_of::<nlmsghdr>() + NLMSG_ALIGNTO - 1) & !(NLMSG_ALIGNTO - 1);
         |      ^^^^^^^^^^^^^^^^^^^^^^^^^^

Is there any way we can support this? Rust-1.13 is pretty old now.

canndrew:
IMHO if people want to stick with an old version of rust-stable for some reason they can stick with an old version of `libc` as well. But I'm curious to hear what @alexcrichton et al. think of that.

alexcrichton:
Why are the `const` fns and new traits needed? Couldn't the constant's value just be inlined?

It's not required to build on ancient versions of Rust but this crate has also tended to not really require newer version of rust either

canndrew:
> Why are the `const` fns and new traits needed?

I'm trying to keep the definitions the same as they are in the headers. In this case we have:

```c
#define NLMSG_ALIGNTO	4U
#define NLMSG_ALIGN(len) ( ((len)+NLMSG_ALIGNTO-1) & ~(NLMSG_ALIGNTO-1) )
#define NLMSG_HDRLEN	 ((int) NLMSG_ALIGN(sizeof(struct nlmsghdr)))
```

Which I would like to translate as

```rust
pub const NLMSG_ALIGNTO: usize = 4;
pub const fn NLMSG_ALIGN(len: usize) -> usize {
    (len + NLMSG_ALIGNTO - 1) & !(NLMSG_ALIGNTO - 1)
}
pub const NLMSG_HDRLEN: usize = NLMSG_ALIGN(mem::size_of::<nlmsghdr>());
```

However that has two problems: Firstly we can't define const fns on stable yet, secondly we can't call const fns on older Rust versions. To work around the first problem I've instead written this:

```rust
pub const NLMSG_ALIGNTO: usize = 4;
pub fn NLMSG_ALIGN(len: usize) -> usize {
    (len + NLMSG_ALIGNTO - 1) & !(NLMSG_ALIGNTO - 1)
}
// FIXME: use this definition once const fns are stable
//pub const NLMSG_HDRLEN: usize = NLMSG_ALIGN(mem::size_of::<nlmsghdr>());
pub const NLMSG_HDRLEN: usize = (mem::size_of::<nlmsghdr>() + NLMSG_ALIGNTO - 1) & !(NLMSG_ALIGNTO - 1);
```

But that still leaves us with the second problem. Note that i can't just inline the value of `mem::size_of::<nlmsghdr>()` because only the compiler really knows what it is.

canndrew:
I suppose I could make `NLMSG_HDRLEN` a function instead. Would that be preferable? It's a constant in the original headers though so it feels like a constant would be more appropriate.

alexcrichton:
I'd personally prefer to make the definitions just constants, would that be possible here?

alexcrichton:

PR #1013, opened 2018-06-01
mlock/munlock stubs

MggMuggins:
The `mlock` and `munlock` routines are unimplemented on redox (I don't think redox even _has_ swap yet), but the symbols are needed for compilation of some deps.

### DIFF
--- a/src/redox/mod.rs
+++ b/src/redox/mod.rs
@@ -128,6 +128,10 @@ extern {
     pub fn close(fd: ::c_int) -> ::c_int;
 }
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! Currently, though, this crate is used to export headers for functions/macros defined in C, typically in libc. In that sense we don't accept stub implementation or other sorts of compatibility implementations, that's left to downstream crates to take care of

MggMuggins:
Oh well, worth a shot I guess.

PR #1015, opened 2018-06-02
Remove termios2 from musl targets

### LINKED ISSUE #1006: termios2 shouldn't exist on musl targets
`musl` doesn't include the termios2 API, but those structs are declared by libc. They should be removed.

Susurrus:
This wasn't caught because the tests disable checking for `termios2` on all targets, so I enabled it on `musl` targets to prevent this from recurring.

Closes #1006

### DIFF
--- a/src/unix/notbsd/linux/musl/b32/arm.rs
+++ b/src/unix/notbsd/linux/musl/b32/arm.rs
@@ -138,17 +138,6 @@ s! {
         pub f_namemax: ::c_ulong,
         __f_spare: [::c_int; 6],
     }
-
-    pub struct termios2 {
-        pub c_iflag: ::tcflag_t,
-        pub c_oflag: ::tcflag_t,
-        pub c_cflag: ::tcflag_t,
-        pub c_lflag: ::tcflag_t,
-        pub c_line: ::cc_t,
-        pub c_cc: [::cc_t; 19],
-        pub c_ispeed: ::speed_t,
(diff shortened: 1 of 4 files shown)


alexcrichton:
Has this been published? If so I think this'd be a breaking change?

If so, do we know as well how this passed ctest without being caught?

Susurrus:
> Has this been published? If so I think this'd be a breaking change?

Yes it has. musl targets are all Tier-2 or worse targets, so it depends on how big of a problem it is to break things on these platforms. Though this is a bug fix. I think it'd be very useful as a contributor to have the breaking changes policy addressed in the README or CONTRIBUTOR documentation to address this.

> If so, do we know as well how this passed ctest without being caught?

I'll need to dig into this a bit more. Looking at the testing, `termios2` is excluded from the regular test because of import conflicts with the C headers. So it's tested as part of the second fcntl tests. `termios2` is the only struct checked on non-musl platforms, but for `musl` targets, I'd expect that it gets checked. The relevant code is:
```rust
        if !musl {
            cfg.skip_struct(|s| {
                s != "termios2"
            });
        }
```

alexcrichton:
Oh dear, that sounds bad! Currently we don't ever remove APIs from libc, so perhaps these could be marked as deprecated with a message saying that they were added mistakenly and are guaranteed to not work on musl?

Susurrus:
Merged into #629 so closing this.

PR #1016, opened 2018-06-04
added renamex_np, renameatx_np

debris:
these function can be used to atomically swap files

- added 2 macos specific functions - `renamex_np` and `renameatx_np` and two constants `RENAME_SWAP` and `RENAME_EXCL`
  - [macos stdio.h](https://github.com/apple/darwin-xnu/blob/0ddccd8/bsd/sys/stdio.h#L45-L48)
  - [docs](http://www.manpagez.com/man/2/renameatx_np/osx-10.12.3.php)
- ~added linux specific `renameat2`. according to the [docs](http://man7.org/linux/man-pages/man2/rename.2.html#VERSIONS) it was added to Linux in kernel 3.15.~

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -1616,9 +1616,6 @@ extern {
                            len: ::off_t) -> ::c_int;
     pub fn readahead(fd: ::c_int, offset: ::off64_t,
                      count: ::size_t) -> ::ssize_t;
-    pub fn renameat2(olddirfd: ::c_int, oldpath: *const ::c_char,
-                     newdirfd: ::c_int, newpath: *const ::c_char,
-                     flags: ::c_int) -> ::c_int;
     pub fn getxattr(path: *const c_char, name: *const c_char,
                     value: *mut ::c_void, size: ::size_t) -> ::ssize_t;
     pub fn lgetxattr(path: *const c_char, name: *const c_char,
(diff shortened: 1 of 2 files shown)


src/unix/notbsd/linux/mod.rs
```diff
@@ -1616,6 +1616,9 @@ extern {
                            len: ::off_t) -> ::c_int;
     pub fn readahead(fd: ::c_int, offset: ::off64_t,
                      count: ::size_t) -> ::ssize_t;
+    pub fn renameat2(olddirfd: ::c_int, oldpath: *const ::c_char,
+                     newdirfd: ::c_int, newpath: *const ::c_char,
+                     flags: ::c_int) -> ::c_int;
```
debris: according to the linux [docs](http://man7.org/linux/man-pages/man2/rename.2.html#NOTES), this should be `c_uint`, but `RENAME_EXCHANGE` which can be passed here is already defined in `libc` as `c_int`. [docs](https://docs.rs/libc/0.2.42/libc/constant.RENAME_EXCHANGE.html)

debris:
actually these functions do not work on macos versions lower than `10.12`. Can I conditionally compile these functions only for newer versions of osx?

alexcrichton:
@bors: retry

Oh no worries, we've got a lot of functions defined that don't work in previous versions of an OS, it's up to each crate to decide how to use the apis

debris:
but these do not exist on older versions of OSX, so linking will fail

alexcrichton:
Indeed and no worries! That's ok.

PR #1018, opened 2018-06-08
add fdopendir on macOS

### LINKED ISSUE #1017: fdopendir on macOS
I'd like to use `fdopendir`, but it's missing on macOS.

Looks like 99a524d added `fdopendir` to...

   * `src/unix/bsd/freebsdlike/mod.rs`
   * `src/unix/bsd/netbsdlike/mod.rs`
   * `src/unix/haiku/mod.rs b/src/unix/haiku/mod.rs`
   * `src/unix/notbsd/mod.rs`
   * `src/unix/solaris/mod.rs`

and it was copied into `src/unix/uclibc/mod.rs` when that was created in 6e4beb6.

so it's there for everything but macOS and "newlib" (which I guess is [this](https://sourceware.org/newlib/)?). As far as I can tell, both those platforms do have it. macOS apparently added it in 10.10 (see chneukirchen/nq#6).

Could this just be in `src/unix/bsd/mod.rs` rather than the freebsdlike/netbsdlike subdirs? or even in `src/unix/mod.rs` (like `fdopen`) since all existing platforms support it?

scottlamb:
Fixes #1017

I moved it up to src/unix/mod.rs, as it's specified in POSIX.1-2008 and
appears to be implemented on every Unix-like system.

The symbol names on macOS appear similar to those for opendir; I found
them via the commands below. I tested the x86_64 version;
fdopendir$INODE64 worked as expected.

$ nm -arch x86_64 /usr/lib/system/libsystem_c.dylib | grep fdopendir
000000000007ea6d T _fdopendir
000000000002ba97 T _fdopendir$INODE64
$ nm -arch i386 /usr/lib/system/libsystem_c.dylib | grep fdopendir
00082d1e T _fdopendir
0002b528 T _fdopendir$INODE64$UNIX2003
00082d1e T _fdopendir$UNIX2003

### DIFF
--- a/src/unix/bsd/freebsdlike/mod.rs
+++ b/src/unix/bsd/freebsdlike/mod.rs
@@ -1170,8 +1170,6 @@ extern {
     pub fn setpriority(which: ::c_int, who: ::c_int, prio: ::c_int) -> ::c_int;
     pub fn rtprio(function: ::c_int, pid: ::pid_t, rtp: *mut rtprio) -> ::c_int;
 
-    pub fn fdopendir(fd: ::c_int) -> *mut ::DIR;
-
     #[cfg_attr(target_os = "freebsd", link_name = "mknodat@FBSD_1.1")]
     pub fn mknodat(dirfd: ::c_int, pathname: *const ::c_char,
                   mode: ::mode_t, dev: dev_t) -> ::c_int;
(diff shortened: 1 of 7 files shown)


scottlamb:
Hmm, the CI failures look unrelated to my changes. Maybe a transient network problem:

> error: component download failed for rust-std-mips64-unknown-linux-gnuabi64
info: caused by: could not download file from 'https://static.rust-lang.org/dist/2018-06-05/rust-std-1.26.2-mips64-unknown-linux-gnuabi64.tar.xz' to '/home/travis/.rustup/downloads/5553de45482f6b808208523e218a5ec132bd6fb05f47f1bd4a5f0e77ee7f8ed1.partial'

Can I make it retry somehow?

PR #1019, opened 2018-06-08
Simplify the stdbuild section

est31:
Found this when encountering the code in the rustc submodule and changing the allow for the warnings to deny.

* `no_std` is stable so it does not have to be listed in the `feature` attribute
* `no_std` as an attribute for the crate is already implied by the `#![cfg_attr(not(feature = "use_std"), no_std)]` below
* `staged_api` as an attribute gives a warning. That also matches my knowledge.


est31:
trying next libs team member. This PR shouldnt sit here not being looked at just because alex is gone!

r? @BurntSushi

est31:
@alexcrichton yes, I've done that back when I filed the PR.

PR #1023, opened 2018-06-18
libc: changes to ppc64le musl branch to support building of rust on A…

mksully22:
…lpine

This PR includes changes to the libc musl branch to include the correct defines & declarations to support powerpc64.  Values that needed changes to a definition for powerpc64.rs that existed higher in the branch also resulted in a change that moved the definition down to the b32/mod.rs, b64/x86_64.rs to ensure that builds continued to work on those architectures. 

Verification was done building rust for both ppc64le and x86_64 on Alpine as described in the git project
https://github.com/mksully22/ppc64le_alpine_rust_1.26.2


alexcrichton:
@bors: r+

Thanks!

PR #1025, opened 2018-06-20
Fix FreeBSD ioctl request constants

dcuddeback:
On FreeBSD `ioctl()` is defined as taking `c_ulong` as the `request` parameter:

```rust
fn ioctl(fd: c_int, request: c_ulong, ...) -> c_int
```

Constants meant to be passed as the `request` parameter were being defined as a mixture of `c_uint` and `c_ulong`, which can cause type errors. For example, the following program

```rust
extern crate libc;

fn main() {
    unsafe {
        libc::ioctl(1, libc::TIOCEXCL);
    }
}
```

results in a type error:

```
 --> src/main.rs:5:24
  |
5 |         libc::ioctl(1, libc::TIOCEXCL);
  |                        ^^^^^^^^^^^^^^ expected u64, found u32
```

This commit changes ioctl request constants on FreeBSD to `c_ulong`, which fixes the compilation error in the above test program.


dcuddeback:
Duplicate of #704. Sorry, I overlooked that there was an existing PR before opening this one.

PR #1030, opened 2018-07-04
Add ENOATTR for Android

mati865:
PR based on https://github.com/rust-lang/libc/pull/594

It's defined in Android sysroot so it should work without test workaround (CI should catch it otherwise?).

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -560,7 +560,7 @@ fn main() {
 
             // Defined by libattr not libc on linux (hard to test).
             // See constant definition for more details.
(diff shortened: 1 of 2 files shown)


mati865:
```
cargo:warning=/checkout/target/x86_64-linux-android/debug/build/libc-test-a83d36f94edd2cc5/out/main.c:17374:51: error: 'ENOATTR' undeclared here (not in a function)
cargo:warning=             static int __test_const_ENOATTR_val = ENOATTR;
cargo:warning=                                                   ^
```
Apparently tests fail without workaround.

PR #1031, opened 2018-07-05
Add linux musl powerpc (32-bit) support

dusxmt:
This change adds support for musl on 32-bit powerpc.  Most of the general constants in src/unix/notbsd/linux/musl/b32/mod.rs were different from those of the powerpc port, so I moved them over to the individual architecture-specific submodules.  The same with the ipc_perm structure.


alexcrichton:
@bors: r+

Thanks!

PR #1033, opened 2018-07-06
Add if_packet.h headers

SmilingNavern:
Some structs for implementing af_packet(7)


SmilingNavern:
I missed types(used char instead of unsigned char). I will fix it

PR #1034, opened 2018-07-06
Link to libgcc when statically linking musl

Amanieu:
On some architectures (e.g. aarch64) musl depends on some libgcc functions (`__addtf3`, `__multf3`, `__subtf3`) for `long double` arithmetic that it uses internally. Unfortunately we don't provide these functions in compiler-builtins, so we instead need to get them from libgcc.

Fixes:
https://github.com/rust-lang/rust/issues/46651
https://github.com/rust-lang-nursery/compiler-builtins/issues/201
https://github.com/rust-lang-nursery/compiler-builtins/issues/217

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -285,6 +285,18 @@ cfg_if! {
         // cargo build, don't pull in anything extra as the libstd dep
         // already pulls in all libs.
     } else if #[cfg(target_env = "musl")] {
+        // On some architectures (e.g. aarch64) musl depends on some libgcc
+        // functions (__addtf3, __multf3, __subtf3) for long double arithmetic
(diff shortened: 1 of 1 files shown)


alexcrichton:
Could you try bootstrapping rustc with this change to confirm that it works? Once that's working should be good to merge!

Amanieu:
I bootstrapped rustc with the aarch64-unknown-linux-musl target and it fixes the linker errors in the issues mentioned.

There's just a small change required in the CI scripts: https://github.com/rust-lang/rust/pull/52157

Amanieu:
Seems to be a spurious failure on travis? Not sure why this change would cause iOS to fail...

alexcrichton:
Ok great! In that case let's merge this and fold it with the rust PR

@bors: r+

PR #930, opened 2018-02-27
Fix --no-default-features builds

gnzlbg:
Closes #928 #929 #684

PR #1038, opened 2018-07-12
Fix netbsd build with "stdbuild"

Amanieu:
A bug was introduced in #930 which caused netbsd to not link to libm.

r? @alexcrichton


Amanieu:
cc @gnzlbg

gnzlbg:
duh! sorry!

Amanieu:
Travis failure seems to be a spurious failure when trying to downloading the Android NDK.

This is ready to be merged.

PR #1040, opened 2018-07-14
Openbsd fix

semarie:
put in several fixes for OpenBSD (one fix per commit).
testsuite ran on OpenBSD 6.3-current (upcoming 6.4)

```
RUNNING ALL TESTS
PASSED 6474 tests
```


semarie:
it is a network failure.

```
error: component download failed for rust-std-i686-unknown-linux-gnu
info: caused by: could not download file from 'https://static.rust-lang.org/dist/2018-07-10/rust-std-1.27.1-i686-unknown-linux-gnu.tar.xz' to '/home/travis/.rustup/downloads/68f028911a1606e93eb71bdf95fe7d91e2b4cbd247663f3ab1e59d291dcd61b7.partial'
info: caused by: error during download
info: caused by: [18] Transferred a partial file (transfer closed with 47035510 bytes remaining to read)
```

semarie:
@alexcrichton could you make @bors to retry please ?

gnzlbg:
@semarie is there an easy way to test openbsd in travis-ci ? (e.g. for crates in the nursery?)

semarie:
@gnzlbg I am unsure :-/ The way would be to follow what is done for FreeBSD: crossbuilding the testsuite and run it on QEMU.

I have in my TODO list to work on crossbuilding OpenBSD from Linux for Rust. I think it could be doable now: I recently saw an openbsd developer with ld-gold on OpenBSD (with patches), and we now have a full LLVM on base but we still use your old and customized ld by default. ld.lld is usable. So I except minimal patches on Linux side to have a way to link binaries for OpenBSD. The plumbing still need to be done...

gnzlbg:
Let me know if you get a "simple" set up going in travis-ci. Ideally we would have a docker Linux image with the cross-compilation toolchain required to cross-compile Rust on Linux for OpenBSD, and then a similar script to the ones used here that uses qemu-system with an OpenBSD image as a test runner.

PR #1041, opened 2018-07-15
Add MODULE_INIT_ constants

bachp:
These are flags required to implement the linux kernel loading mechanism.
Specifically finit_module.

### DIFF
--- a/ci/docker/aarch64-unknown-linux-musl/Dockerfile
+++ b/ci/docker/aarch64-unknown-linux-musl/Dockerfile
@@ -3,14 +3,14 @@ FROM ubuntu:17.10
 RUN apt-get update && apt-get install -y --no-install-recommends \
   gcc make libc6-dev git curl ca-certificates \
   gcc-aarch64-linux-gnu qemu-user
-RUN curl https://www.musl-libc.org/releases/musl-1.1.16.tar.gz | \
+RUN curl https://www.musl-libc.org/releases/musl-1.1.19.tar.gz | \
     tar xzf - && \
-    cd musl-1.1.16 && \
+    cd musl-1.1.19 && \
     CC=aarch64-linux-gnu-gcc \
     ./configure --prefix=/musl-aarch64 --enable-wrapper=yes && \
     make install -j4 && \
     cd .. && \
-    rm -rf musl-1.1.16 && \
+    rm -rf musl-1.1.19 && \
 # Install linux kernel headers sanitized for use with musl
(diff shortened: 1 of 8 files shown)


bachp:
I'm not sure if this constants should also be added to the android platform as the kernel module loading should be the same there.

bachp:
That platforms where it is failing is Linux and Linux is the only one it is available. So excluding doesn't make much sense :wink: 

The constants are defined in the linux uapi header file [`module.h`](https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/module.h#L6). Do I need to do something to makes this file known to the build?

alexcrichton:
You may need to tweak `libc-test/build.rs` to include the right header file

bachp:
@alexcrichton It looks like the musl kernel-headers do not include the `linux/module.h` file. The issue seems to be similar to: https://github.com/rust-lang/libc/pull/616/

I'm unsure how to handle this. As this are only constant definitions that will be used later in a syscall (not a c function) from the nix crate. So I don't just want to exclude musl here.

I tried manually copying the `module.h` file to the right location, and this works, but feels like a very dirty hack.

I see several ways to fix this:
1. Add the module.h to the kernel headers (very hacky)
2. Don't define the constants in libc but in nix (seems to be not the  proper way)
3. Exclude musl (makes me unable to use them in nix when building against musl)

Any ideas.

alexcrichton:
The docker images we're using I think are copying the kernel heders from #616 I think? Are newer versions of the headers needed though?

bachp:
Newer version of the "sanetized headers" are not available. I will check if I can bring the missing header in there and then use the newer version.

bachp:
Sparc64 is still failing as debian doesn't ship the required header file. Ubuntu does that's why it is working there.

bachp:
@alexcrichton I got CI to pass but with two caveats:

1. The musl build depend if https://github.com/sabotage-linux/kernel-headers/pull/14 getting accepted to be properly working.
2. The sparc64 build needs a hack to copy over the `linux/module.h` from the kernel headers uapi as it is not included in Debian.

alexcrichton:
This looks great, thanks! I think doing something weird for one platform is fine. It looks like upstream is merged so should the urls switch back to that?

bachp:
@alexcrichton Yes I will update the PR

bachp:
@alexcrichton Urls are switched back to upstream. I think this is good to go now.

Issue #1042, opened 2018-07-15
Replace __align fields with #[repr(align(x))]


faern:
This has been briefly touched in #963 and #58 but this issue intend to discuss it from a different angle.

When I grep in the libc repository I find 123 instances of `__align:`. The main problem is that instantiating a struct that uses this way to achieve alignment requires:
```rust
let mut foo: libc::some_struct = unsafe { mem::uninitialized() };
foo.field = my_value;
```
Which has many downsides, mainly
* needing `unsafe`
* has to be assigned to a mutable variable
* and can't be done in a constant expression.

`#repr(align(x))]` was introduced in Rust [1.25](https://blog.rust-lang.org/2018/03/29/Rust-1.25.html), which at the time of writing this is only two releases behind current stable. Thus I understand we probably can't bump the required rustc version for libc that high yet. But I could not find any information about how backwards compatible libc intends to be, going forward. Since #972 bumped the required rustc version from 1.0 to 1.13 without much fanfare or breaking version bumping etc I guess there is some unofficial point where even libc don't care about your old compiler anymore.

Switching to `#repr(align(x))]` would, among removing the need for a lot of unsafe, help in making more things constant expressions. For example this issue where we need to create an `in6_addr` in an associated constant: https://github.com/rust-lang/rust/issues/44582

Would it count as a breaking change to remove the `__align` fields? It would change the struct, but not the public part of it, so I guess not. Any code using the `mem::uninitialized()` dance should still work exactly the same I think.

I'm playing with the thought to somehow make this work in Edition 2018. But I guess compiling a crate with `edition = "2018"` that has edition 2015 dependencies only compiles libc once, even if crates from both editions depend on it? If it is compiled separately for each edition then something like the following could work (given that there will be such a cfg value to conditionally compile on to begin with):
```rust
#[cfg_attr(not(edition = "2015"), repr(align(4)))]
pub struct in6_addr {
    pub s6_addr: [u8; 16],
    #[cfg(edition = "2015")]
    __align: [u32; 0],
}
```

Another issue that discusses using newer rustc features in order to improve the libc implementation is #1020. Both implementing that issue and this one would raise the rustc version requirement. Although, the other issue would require rustc 1.19, while this is a lot more extreme as it requires 1.25.

alexcrichton:
[...]

PR #1044, opened 2018-07-20
Add alignment feature and use #[repr(align(x))]

faern:
Trying to solve #1042.

Here I introduce the discussed feature that will allow going from struct alignment with a private `__align` field to using `#[repr(align(x))]`. However, I have not implemented it for all structs that require alignment yet, only `in6_addr`. This because I did not want to spend too much time before we have discussed and solved the remaining questions regarding this.

One thing to discuss is testing. I have so far not done anything to the CI scripts. So currently they will still test the crate only with the `align` feature disabled. Thus they will make sure the `__align` fields are still correct. But no automatic tests make sure everything is correct when the `align` feature is turned on. What do we want to do about that? Can we insert another `cargo test` with `--features align` to make all the Travis jobs run the test suite twice, or will that slow things down too much?

I have tried using this version of libc in rustc and the standard library. And successfully changed `Ipv6Addr::new` to not use any `unsafe` and to become a `const fn`. Whether or not we want that is out of scope for this PR, but my point was that the changes introduced with this PR allow much more flexible usage of the libc structs that have alignment.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ appveyor = { repository = "rust-lang/libc", project_name = "rust-lang-libs/libc"
 [features]
 default = ["use_std"]
 use_std = []
+align = []
 
 [workspace]
 members = ["libc-test"]
--- a/README.md
+++ b/README.md
@@ -35,6 +35,16 @@ this via:
 libc = { version = "0.2", default-features = false }
 ```
 
+By default libc uses private fields in structs in order to enforce a certain
+memory alignment on them. These structs can be hard to instantiate outside of
+libc. To make libc use `#[repr(align(x))]`, instead of the private fields,
+activate the *align* feature. This requires Rust 1.25 or newer:
+
+```toml
+[dependencies]
+libc = { version = "0.2", features = ["align"] }
+```
+
 ## What is libc?
 
 The primary purpose of this crate is to provide all of the definitions necessary
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -79,7 +79,7 @@ if [ "$QEMU" != "" ]; then
   exec grep "^PASSED .* tests" $CARGO_TARGET_DIR/out.log
 fi
 
-# FIXME: x86_64-unknown-linux-gnux32 fail to compile wihout --release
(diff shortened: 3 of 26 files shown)


faern:
Experimented with running an extra iteration of `cargo test` with the feature activated if the build job is using Rust >=1.25.

faern:
If you like this approach and find it OK to merge, ping me and I'll add the same things to the remaining `__align` structs.

alexcrichton:
I think some tests may still be failing?

faern:
Now I got a bit further. But are there maybe nicer ways to write what I have already done?

Also, I struggle with figuring out the test failure on `i686-unknown-linux-gnu` where the test says `bad pthread_cond_t align: rust: 8 (0x8) != c 4 (0x4)`. The old alignment there is `__align: [::c_longlong; 0],`, so alignment should be `c_longlong` which should be 64 bit on all platforms(?), and align(8) works on all other 32 bit platforms.

alexcrichton:
Ah I think u64 has a 4 byte alignment on 32-bit platforms (or at least on Linux), so that may need a conditional alignment attribute

faern:
Finally had time to finish the rest of the structs and their corresponding constants. Some open questions:

* Do you think I should document this feature in some way? Either by comment just in `Cargo.toml`, or a section in the README?
* Is the extra test, `no_private_align.rs`, an acceptable/reasonable approach to make sure future changes don't accidentally make `in6_addr` not possible to create instances of without `unsafe`?
* Is the `use pthread_mutex_t;` in order to change `::pthread_mutex_t` into `pthread_mutex_t` acceptable? I did not find a way to make the macro accept anything other than a `ident` for the `$t2` argument.

alexcrichton:
Thanks! All the changes look good to me yeah, although I'd be fine without the `no_private_align.rs` tests as well, we have no more reason to add fields that don't exist!

For the feature, yeah, mind adding a blurb to the README about it?

faern:
Done. I removed the commit that added the custom test and added a commit with an extra paragraph in the README.

alexcrichton:
@bors: r+

Thanks!

PR #1048, opened 2018-07-30
Add libc definitions for HermitCore (https://hermitcore.org)

ColinFinck:
HermitCore is based on lwIP, newlib, and pthread-embedded.
Some definitions are similar to other targets using newlib, however some are different enough to justify an own "hermit" port and not base on the existing "newlib" port.

### DIFF
--- a/src/unix/hermit/aarch64.rs
+++ b/src/unix/hermit/aarch64.rs
@@ -1,0 +1,2 @@
+pub type c_char = u8;
+pub type wchar_t = u32;
--- a/src/unix/hermit/mod.rs
+++ b/src/unix/hermit/mod.rs
@@ -696,22 +696,29 @@ f! {
 }
 
 extern {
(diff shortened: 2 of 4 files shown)


ColinFinck:
This patch is part of a patch series to add HermitCore support to Rust:

1. https://github.com/rust-lang/llvm/pull/122 / https://reviews.llvm.org/D48964
2. https://github.com/rust-lang/libc/pull/1048
3. https://github.com/rust-lang/rust/pull/52861

CC @stlankes

alexcrichton:
Looks great to me, thanks! I think the style check on CI is getting tripped up, but if that gets fixed I'll r+!

ColinFinck:
@alexcrichton Thanks for the prompt reply! I've fixed the coding style.

alexcrichton:
@bors: retry

er... sorry about that!

PR #1050, opened 2018-07-31
Add libc definitions for Megaton-Hammer, a Switch Homebrew toolchain

roblabla:
I'm working on a pure-rust toolchain to write homebrew for the Nintendo Switch called [Megaton-Hammer](http://github.com/megatonhammer/megaton-hammer). I'm hoping to get those definitions upstreamed to simplify my life :).

This toolchain does not depend on a C compiler or a libc (it reimplements everything in rust) - but given many crates in the Rust ecosystem rely on the libc crate for the definition of various common types, this is what I came up with.

I was wondering what a good target triple would be ? I currently gate the implementation behind `target_os = "switch"`, but if this goes upstream I figure that might cause trouble for people using the Nintendo SDK (they might already be using `target_os = "switch"` for some things). Would it be better to go with `target_env = "megatonhammer"`?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -104,6 +104,9 @@ mod dox;
 cfg_if! {
     if #[cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))] {
         // empty ...
+    } else if #[cfg(target_os = "switch")] {
+        // On the Switch, we only define some useful universal types for
+        // convenience. Those can be found in the switch.rs file.
     } else {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! I think `target_os` is fine to use, but could this use a separate file instead of inlining the definitons? (like most other platforms use)

roblabla:
It's fixed :)

PR #1056, opened 2018-08-05
Bump version to 0.2.43

faern:
Would be nice to have the new align feature from #1044 available for general use. But mostly I want this released since I have problems using the align feature for a PR on libstd, and I suspect it's somehow because I try to use an unpublished libc (https://github.com/rust-lang/rust/pull/52872).


alexcrichton:
@bors: r+ 

Thanks!

PR #1063, opened 2018-08-15
NetBSD: correct link_name of some time-related functions

jakllsch:
The futimes(), lutimes(), mq_timedreceive(), and mq_timedsend()
functions were linking against legacy library symbols that need 32-bit
time_t in structures, resulting in an ABI mismatch with 64-bit time_t.


gnzlbg:
@alexcrichton I've been getting this sparc64 failure for a while now, so often, that it is not spurious I think, and just came to libc to see if it was happening here too. rust-lang/rust probably will get this as well, so it might be worth opening an issue, pinging @malbarbo and trying to find a different image or upload a backup somewhere else that we can update the buildbots to.

malbarbo:
Its seems that the site is in maintenance mode. See https://lists.debian.org/debian-infrastructure-announce/2018/08/msg00001.html

alexcrichton:
Thanks @jakllsch!

I'll work on CI in a bit

PR #1071, opened 2018-09-02
Add mallopt

Lytigas:
First time contributor. Thanks for the excellent contributing guide.
I'm not quite sure I put the method in the right place. [The GNULib docs](https://www.gnu.org/software/gnulib/manual/html_node/mallopt.html) say its not on Android 7.1, but there are [mentions of it in some android headers](https://android.googlesource.com/platform/bionic/+/master/libc/include/malloc.h).


Lytigas:
So from the Travis failures it's evident this doesn't exist in musl. Is there a place for glibc-specific APIs? Or does that place it beyond the scope of this crate?

alexcrichton:
@bors: r+ 

Looks great to me, thanks!

PR #1073, opened 2018-09-03
Add stdin, stdout, stderr globals

tbu-:
CC #7.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -152,6 +152,12 @@ cfg_if! {
         pub enum FILE {}
         pub enum fpos_t {} // TODO: fill this out with a struct
 
+        extern {
+            pub static mut stdin: *mut FILE;
+            pub static mut stdout: *mut FILE;
+            pub static mut stderr: *mut FILE;
+        }
+
         pub const INT_MIN: c_int = -2147483648;
         pub const INT_MAX: c_int = 2147483647;
 


tbu-:
How do I test this?

tbu-:
(On Linux with glibc, it seems to work.)

alexcrichton:
Hm unfortunately I don't think `libc-test` is capable of testing `static` imports right now. I'd prefer to fix that before landing this, as having this in `src/lib.rs` seems unlikely to work for *all* platforms. Would you be interested in sending a PR to `ctest` to add support for a `static`?

tbu-:
Summarizing the failures:
macOS: `std*` not defined, not investigated further.
musl: `std*` defined as `FILE *const` instead of glibc's `FILE *`.
Android: Unrelated failure, due to the now-existing static checking: `__progname` not found in the included headers, declared in [src/unix/notbsd/android/mod.rs:1516](https://github.com/rust-lang/libc/blob/267375e99957d97317b24f19210af576cb58b2f2/src/unix/notbsd/android/mod.rs#L1516).
NetBSD: `std*` is just a macro for something else (permitted by the C standard, `std*` just needs to be a macro which resolves to an expression for the relevant stream): NetBSD source: [include/stdio.h:215-217](https://github.com/rumpkernel/src-netbsd/blob/b8b951e911a2fc555848a2785a9998bc128530b6/include/stdio.h#L215-L217), [include/stdio.h:149](https://github.com/rumpkernel/src-netbsd/blob/b8b951e911a2fc555848a2785a9998bc128530b6/include/stdio.h#L149).
FreeBSD: `std*` is a macro referencing a static with a different name (`__std*p`): FreeBSD source: [include/stdio.h:240-242](https://github.com/freebsd/freebsd/blob/91d3358201172058408ca55206d177de3b7e33c7/include/stdio.h#L240-L242), [include/stdio.h:169-171](https://github.com/freebsd/freebsd/blob/91d3358201172058408ca55206d177de3b7e33c7/include/stdio.h#L169-L171).

tbu-:
Since it's only required to be a value by the C standard, maybe declaring it as a static is not the right way. Would you consider adding it as functions `fn stdin() -> *mut FILE`, similar to how the `stdin` identifier might only represent a value in C?

It seems we don't have bindings for errno so far, a similar thing could be done for that, I think.

alexcrichton:
The `libc` crate only defines what's already in libc, so we can use things like `#[link_name]` to rename it, but we can't insert our own shims.

tbu-:
`stdin` is part of libc. Rust doesn't have a 1-1 replacement for macros.

Maybe this could be part of a module, `macros`, so it doesn't clash with a potentially existing symbol `stdin`?

alexcrichton:
I don't really have a preference on how this is bound, but it cannot be done through in-crate shims

thomcc:
It seems like NetBSD is the only place where this needs a shim. The other places where macros are used (freebsd/darwin), it can be done with `#[link_name]`.

Given that the libc crate already doesn't expose everything availiable on every OS, it seems plausibly okay to leave that out for NetBSD (and I guess any other OS like this).

Would a patch that did that have a shot of being accepted?

PR #1075, opened 2018-09-05
Theoretically test statics

tbu-:
There are none of them in `libc` except for `__progname` on Android, but
that one cannot be tested because it's not present in any header files.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -432,7 +432,6 @@ fn main() {
     cfg.skip_struct(move |ty| {
         match ty {
             "sockaddr_nl" => musl,
-            "ucontext_t" => true,
 
             // On Linux, the type of `ut_tv` field of `struct utmpx`
             // can be an anonymous struct, so an extra struct,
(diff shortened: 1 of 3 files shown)


alexcrichton:
Yes let's revert those changes in `ctest`, I'll be yanking 0.2.1 for that reason

alexcrichton:
er, didn't mean to close

libc-test/build.rs
```diff
@@ -432,6 +432,7 @@ fn main() {
     cfg.skip_struct(move |ty| {
         match ty {
             "sockaddr_nl" => musl,
+            "ucontext_t" => true,
```
alexcrichton: How come this is unconditionally ignored? Should it only be ignored on one platform?

libc-test/build.rs
```diff
@@ -639,6 +640,7 @@ fn main() {
             s if ios && s.starts_with("RTV_") => true,
             s if ios && s.starts_with("DLT_") => true,
 
+            "NFT_MSG_MAX" => true,
```
alexcrichton: Could this have a comment indicating why it's ignored?

tbu-:
My bad, I added these ignores because they failed on my machine, didn't mean to add them to the PR.

Will open an issue about them.

PR #1078, opened 2018-09-10
Add new FreeBSD socket option SO_REUSEPORT_LB.

johalun:
FreeBSD 12, which is scheduled to be released soon, has a new socket option SO_REUSEPORT_LB.  
From setsockopt man page:
     SO_REUSEPORT_LB allows completely duplicate bindings by multiple
     processes if they all set SO_REUSEPORT_LB before binding the port.
     Incoming TCP and UDP connections are distributed among the sharing
     processes based on a hash function of local port number, foreign IP
     address and port number. A maximum of 256 processes can share one socket.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -556,7 +556,8 @@ fn main() {
             "PD_CLOEXEC" | "PD_ALLOWED_AT_FORK" if freebsd => true,
 
             // These constants were added in FreeBSD 12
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! It look like though this is failing CI? It may be that our version of FreeBSD is too old, in which case can this be added to the whitelist to not test in `libc-test/build.rs`?

johalun:
OK, added. I hope it was in the right place :)

PR #1082, opened 2018-09-17
Re-export core::ffi::c_void if it exists

IsaacWoods:
This is the second part of the implementation of [RFC 2521](https://github.com/rust-lang/rfcs/pull/2521), replacing the definition of `c_void` in libc with a re-export of the type from `core::ffi::c_void` on builds it exists for.

This uses the re-export for rustc version `1.31.0` or greater, as `1.30.x` was the current nightly when [the PR for the changes to libcore and libstd](https://github.com/rust-lang/rust/pull/53910) was merged, so I'm assuming the first nightly they will appear in will be `1.31.0`; is this acceptable?

cc rust-lang/rust#53856

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ description = """
 A library for types and bindings to native C functions often found in libc or
 other common platform libraries.
 """
+build = "build.rs"
 
 [badges]
 travis-ci = { repository = "rust-lang/libc" }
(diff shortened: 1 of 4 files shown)


IsaacWoods:
Done :)

I think the one Travis failure (`stable, TARGET=aarch64-linux-android`) is spurious, as the rest pass and it just appears to be timing out. Could someone issue a retry?

PR #1085, opened 2018-09-25
Expose futimens and utimensat on FreeBSD and DragonFly

jmmv:
Should have sent this out with #1084 but I didn't notice then that these were missing (see https://github.com/nix-rust/nix/pull/944).

Haven't been able to test these locally as I do not have access to these platforms. Will rely on CI to verify these for me.

### DIFF
--- a/src/unix/bsd/freebsdlike/mod.rs
+++ b/src/unix/bsd/freebsdlike/mod.rs
@@ -1136,6 +1136,9 @@ extern {
                         timeout: *const ::timespec) -> ::c_int;
     pub fn sigwaitinfo(set: *const sigset_t,
                        info: *mut siginfo_t) -> ::c_int;
+    pub fn futimens(fd: ::c_int, times: *const ::timespec) -> ::c_int;
+    pub fn utimensat(dirfd: ::c_int, path: *const ::c_char,
+                     times: *const ::timespec, flag: ::c_int) -> ::c_int;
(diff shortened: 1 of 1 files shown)


jmmv:
The two failures in the Travis build look like flakes due to a network download issue. What's the process here? I do not have the permissions to make Travis re-run those two failed builds.

alexcrichton:
@bors: retry 

Ah we just try again!

alexcrichton:
@bors: retry

???

jmmv:
This is unfortunate...

jmmv:
One last attempt today maybe? :-P

jmmv:
This is sad :(

Wondering if it's possible to just hit retry on the individual Travis build failures in the hope that we can iterate faster and get the flakes through? Or maybe these builds should be marked non-fatal until they can be made to work again?

alexcrichton:
Unfortunately our CI doesn't work by retrying individual builds, I'll need to allocate some time to investigate and try to fix these android builds

jmmv:
Give it another try? I have created a test repository that downloads the problematic files on Travis, using the same settings as this project, and cannot reproduce it: https://travis-ci.org/jmmv/test/builds/434537625. So either something else is going on, or the problem has been resolved.

alexcrichton:
@bors: retry

Sure yeah, can always try that!

PR #1086, opened 2018-10-01
Revisit work on cvoid

IsaacWoods:
At the recommendation of @SimonSapin [here](https://github.com/rust-lang/rust/issues/53856#issuecomment-422932721), I have revisited the build script to check whether `core::ffi::c_void` resolves, instead of relying on a particular `rustc` version. This allows use on `1.30.x` builds with `core::ffi::c_void`.

I also noticed that `c_void` is defined twice in this crate, once in `lib.rs` and again in `switch.rs`. This instead defines `c_void` for every target except `wasm32`. As far as I can tell, this shouldn't actually change functionality on any existing targets.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -101,6 +101,28 @@ extern crate std as core;
 #[macro_use] mod macros;
 mod dox;
 
+/*
+ * `c_void` should be defined for all targets except wasm.
+ */
+#[cfg(not(all(target_arch = "wasm32", not(target_os = "emscripten"))))]
+cfg_if! {
+    if #[cfg(core_cvoid)] {
+        pub use core::ffi::c_void;
+    } else {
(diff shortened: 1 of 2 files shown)


SimonSapin:
The use of `#![feature]` and `-Z` make this incompatible with release channels other than Nightly. Could they both be replaced by `--crate-type lib`?

alexcrichton:
Thanks for the PR! I would not personally recommend this sort of feature detection though, there's precisely one right answer we know ahead of time for "when is `c_void` included?" based on versions, so I think we can just amend that here.

Executing `rustc -V` should be much quicker as well than compiling a program which requires involving a lot more rustc infrastructure.

IsaacWoods:
Okay, I've dropped the first commit with the feature detection. Do you think the de-duplication bit is still worth merging, or should I close this?

alexcrichton:
@bors: r+

Nah that's fine, looks good to me!

PR #1089, opened 2018-10-05
Added ptrace constants for apple

xd009642:
I noticed the ptrace constants were missing for apple so I've added them referencing this [header](https://opensource.apple.com/source/xnu/xnu-792.13.8/bsd/sys/ptrace.h). This doesn't add any new functions just constants so I haven't added any tests which seems to be correct.

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -747,7 +747,7 @@ pub const PROT_READ: ::c_int = 1;
 pub const PROT_WRITE: ::c_int = 2;
 pub const PROT_EXEC: ::c_int = 4;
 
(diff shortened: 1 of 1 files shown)


xd009642:
So xcode stable is an allowed failure but not xcode nightly which seems to be why homu failed. However, the failures seem to be from the same cause. Is this a legit issue?

alexcrichton:
iOS is allowed to fail but the main OSX targets aren't allowed to fail, I don't think it's stable/nightly?

xd009642:
I'm not sure tbh I was just going off the fact only stable is mentioned in allowed failures. Either way I'm unsure why the addition of some constants would introduce a failure?

screenshot from 2018-10-05 18-05-26

PR #1091, opened 2018-10-08
Arch specific ptrace commands for BSDs

xd009642:
This PR is to add architecture specific ptrace requests etc.

Don't merge this yet if it passes I want to go over the other BSDs and architectures to see if I've missed anything I'm just creating the PR to get that initial CI approval that it's right as I expect this bit to be more fiddly.


xd009642:
So there might be architecture specific commands for apple. But in all honesty I can't find anything about it online, so I'm going to count this as done once the CI passes and then that should be the last of ptrace related PRs!

PR #1093, opened 2018-10-13
Add acct(2) to BSD module

jabedude:
The `acct` syscall is supported on BSD platforms, the PR aims to add it to the BSD module.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,6 @@ matrix:
            RUSTFLAGS=-Clink-arg=-mios-simulator-version-min=7.0
       before_install:
         rustc ./ci/ios/deploy_and_run_on_ios_simulator.rs -o $HOME/runtest
-    - env: TARGET=x86_64-rumprun-netbsd
     - env: TARGET=powerpc-unknown-linux-gnu
     - env: TARGET=powerpc64-unknown-linux-gnu
     - env: TARGET=powerpc64le-unknown-linux-gnu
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Looks like ci is failing though?

jabedude:
Looks like it failed here:
`          /checkout/target/x86_64-rumprun-netbsd/debug/build/libc-test-b256b20c8099faff/out/main.c:5622: undefined reference to `acct'`
I'm not familiar with the `x86_64-rumprun-netbsd` triple, but netbsd supports the acct(2) syscall. 
Any idea why it fails to build on that platform?

alexcrichton:
Presumably the API isn't available on that platform? Could it be omitted on that platform?

jabedude:
Oh definitely! I was just confused why x86_64-rumprun-netbsd didn't compile when netbsd supports acct(2). I will omit with a #[cfg]

alexcrichton:
Oh the target is a unikernel fork that's super old and hasn't been updated in forever, which may explain that!

jabedude:
Do you know how to exclude that specific target triple? The Rust documentation doesn't show a way to be more specific than `netbsd`. [See here.](https://doc.rust-lang.org/reference/attributes.html#conditional-compilation)

alexcrichton:
I think that you found the right way to test for it, but it looks like CI is failing?

jabedude:
I'm sort of at a loss on how to use #[cfg] and keep the linter happy. CI also failed when I used `cfg_if!`

alexcrichton:
The current iteration [also fails due to using `target_vendor` on OSX](https://travis-ci.org/rust-lang/libc/jobs/441410418) it looks like. 

TBH I think we should go back to the original version of this PR and just remove `x86_64-rumprun-netbsd` from the list of tested targets, AFAIK no one's ever even used the target... Want to go ahead and do that?

jabedude:
Sounds good. I reverted to the original version

jabedude:
Did @bors get stuck?

Issue #1094, opened 2018-10-16
support for RS485


xoac:
Hi,
There exist crate [rs485](https://github.com/mbr/rs485-rs/blob/master/src/lib.rs#L49) that define const from libc. The problem is that `ioctls` change signature depending on system platform (for example it accept `c_int` when using musl on x86 and `c_ulong` when use glibc on x86_64). So compilation fill when musl is involved:

```
rust-musl-builder-armv7 cargo build --examples                                                                                                              18:21:01 
   Compiling rs485 v0.1.0
error[E0308]: mismatched types
  --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/rs485-0.1.0/src/lib.rs:92:45
   |
92 |         let rval = unsafe { libc::ioctl(fd, TIOCGRS485, &mut conf as *mut SerialRs485) };
   |                                             ^^^^^^^^^^ expected i32, found u32

error[E0308]: mismatched types
   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/rs485-0.1.0/src/lib.rs:186:45
    |
186 |         let rval = unsafe { libc::ioctl(fd, TIOCSRS485, self as *const SerialRs485) };
    |                                             ^^^^^^^^^^ expected i32, found u32

```  

It would be nice to add  this constants (`TIOCGRS485`, `TIOCSRS485`) to libc. I am not sure how to do it properly. For musl target constant can be found here:
http://git.musl-libc.org/cgit/musl/tree/arch

in most cases looks like this:
```
#define TIOCGRS485	0x542E
#define TIOCSRS485	0x542F
```
but there are different here:
http://git.musl-libc.org/cgit/musl/tree/arch/mips64/bits/ioctl.h#n61

alexcrichton:
Looks like you've got a [PR for this](https://github.com/rust-lang/libc/pull/1095), which sounds great to me!

xoac:
@gnzlbg I think no -- I just added implementation to musl target and it's still missing for others. I don't need more since I use just musl -- but I think rs485 crate will still fail to compile on others platform.

gnzlbg:
The RS485 crate define the constants itself, so it cannot fail to compile AFAICT.

xoac:
It can fail `c_int` and `c_ulong` are different types. Since libc (this crate) didn't define it. So rs485 failed to compile on musl buf successful for example on glibc.
[...]

PR #1095, opened 2018-10-16
Add TIOCGRS485 and TIOCSRS485 for musl #1094

xoac:
I need help with adding `TIOCSRS485` and `TIOCSRS485` constants to libc.

### DIFF
--- a/src/unix/notbsd/linux/musl/b32/arm.rs
+++ b/src/unix/notbsd/linux/musl/b32/arm.rs
@@ -467,6 +467,9 @@ pub const TIOCMSET: ::c_int = 0x5418;
 pub const FIONREAD: ::c_int = 0x541B;
 pub const TIOCCONS: ::c_int = 0x541D;
 
+pub const TIOCGRS485: ::c_int = 0x542E;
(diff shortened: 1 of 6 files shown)


alexcrichton:
Looks good to me! I think CI is failing though?

xoac:
It's not good. This lacks some definitions. I just defined this for musl target.
We should add this also here:
```
src/unix/notbsd/linux/other/*
src/unix/notbsd/linux/mips/ *
src/unix/newlibc/*
src/unix/uclibc/*
```
and maybe in more places but I am not sure how to check values that paramter should have. It would nice to have a link to sources in c in every module.

xoac:
This CI is failing because when I forked `master`  master was failing

alexcrichton:
@xoac I don't think that's correct? There's syntax errors in this PR?

alexcrichton:
Looks like there may still be some CI failures?

xoac:
I think u missed comment about missing definition for other implementations. It will be available for musl target but will not for exmple for glibc x86. It's because I don't know where to verify that variables..

alexcrichton:
Oh that's fine! We're ok with incremental updates an additions

PR #1096, opened 2018-10-17
Add truncate(2) to Unix platforms

jmmv:


jmmv:
Haven't tested this locally as I cannot predict its implications. Will rely on a CI run to see whether this truly works for all platforms, if that's OK.

alexcrichton:
@bors: r+

👍

PR #1098, opened 2018-10-19
Add CMSG macros for unix/bsd and unix/notbsd

andrewtj:
This adds the ["Ancillary Data Object Macros" from RFC 2292](https://tools.ietf.org/html/rfc2292#section-4.3). My C's weak but I think I've translated them faithfully. Tested on macOS and Linux/glibc.

### DIFF
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -332,6 +332,43 @@ pub const POLLRDBAND: ::c_short = 0x080;
 pub const POLLWRBAND: ::c_short = 0x100;
 
 f! {
+    pub fn CMSG_FIRSTHDR(mhdr: *const msghdr) -> *mut cmsghdr {
+        if (*mhdr).msg_controllen as usize >= mem::size_of::<cmsghdr>() {
+            (*mhdr).msg_control as *mut cmsghdr
+        } else {
+            0 as *mut cmsghdr
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


andrewtj:
Any feedback?

alexcrichton:
Thanks for the PR and sorry for the delay! These look good to me though, thanks!

@bors: r+

src/unix/bsd/mod.rs
```diff
@@ -332,6 +332,43 @@ pub const POLLRDBAND: ::c_short = 0x080;
 pub const POLLWRBAND: ::c_short = 0x100;
 
 f! {
+    pub fn CMSG_FIRSTHDR(mhdr: *const msghdr) -> *mut cmsghdr {
+        if (*mhdr).msg_controllen as usize >= mem::size_of::<cmsghdr>() {
+            (*mhdr).msg_control as *mut cmsghdr
+        } else {
+            0 as *mut cmsghdr
```
gnzlbg: please use `ptr::null()` here

andrewtj: I assumed the crate's convention was to not use that (it's not used anywhere that I've seen)

gnzlbg: duh, indeed, libc might predate https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#zero_ptr we can always do a cargo fix for these later

gnzlbg:
r? me

PR #1100, opened 2018-10-19
Added support for dup3() function

koutheir:
Added support for `dup3()` function.


asomers:
@koutheir your build failed because it created a duplicate definition for Linux, and an unresolved symbol on OSX.  It looks like OSX doesn't support dup3 yet.  Instead of adding the definition to src/unix/mod.rs, you'll have to add it to various subdirectories, as appropriate.  Don't forget to check the man pages for OSes like OpenBSD and NetBSD that aren't actually run in CI.

asomers:
So did you decide that you don't need `dup3` after all?

koutheir:
No, I made another pull request to perform the operation correctly.

PR #1102, opened 2018-10-20
New module for bare metal architecture which still have a libc

### LINKED ISSUE #375: The thumb* target family (which have newlib libc) are not supported
I have embedded hardware without an OS. I build Rust code for it with `--target thumbv7em-none-eabi` and this `thumbv7em-none-eabi.json` file:

``` json
{
    "arch": "arm",
    "cpu": "cortex-m4",
    "data-layout": "e-m:e-p:32:32-i64:64-v128:64:128-a:0:32-n32-S64",
    "disable-redzone": true,
    "executables": true,
    "llvm-target": "thumbv7em-none-eabi",
    "morestack": false,
    "os": "none",
    "relocation-model": "static",
    "target-endian": "little",
    "target-pointer-width": "32",
    "no-compiler-rt": true,
    "pre-link-args": [
        "-mcpu=cortex-m4", "-mthumb",
        "-Tlayout.ld"
    ],
    "post-link-args": [
        "-lm", "-lgcc", "-lnosys"
    ]
}
```

If I add a Cargo dependency on the `libc` crate and build with this target, I get many build errors. The first and last are below. The entire thing is at https://gist.github.com/anonymous/780fe74b2ae2834fc9c738dbff60acc0

``` rust
   Compiling libc v0.2.15
error[E0412]: type name `c_char` is undefined or not in scope
   --> /home/simon/tmp/cargo-home/registry/src/github.com-1ecc6299db9ec823/libc-0.2.15/src/lib.rs:153:35
    |
153 |     pub fn fopen(filename: *const c_char,
    |                                   ^^^^^^ undefined or not in scope
    |
    = help: no candidates by the name of `c_char` found in your project; maybe you misspelled the name or forgot to import an external crate?


// Removed: ~4000 more lines


error[E0204]: the trait `Copy` may not be implemented for this type
   --> /home/simon/tmp/cargo-home/registry/src/github.com-1ecc6299db9ec823/libc-0.2.15/src/macros.rs:44:9
    |
44  |         impl ::dox::Copy for $i {}
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^ field `rlim_cur` does not implement `Copy`
    | 
   ::: /home/simon/tmp/cargo-home/registry/src/github.com-1ecc6299db9ec823/libc-0.2.15/src/lib.rs
    |
261 | cfg_if! {
    | - in this macro invocation

error: aborting due to 5 previous errors

error: Could not compile `libc`.

```

gbip:
This should fix #375 :smile:
Please tell me if there is anything wrong in this PR.


gbip:
I think this is now ready for review :smile:

alexcrichton:
Thanks for the PR! This has come up a few times before:

* https://github.com/rust-lang/libc/pull/891
* https://github.com/rust-lang/libc/pull/753

Right now `libc` isn't targeted at providing a compatibility layer or polyfilling certain types and such on various platforms, but rather it's providing the bindings for a particular platform that match the C standard and/or popular C compilers. That means that we can't really add this target as it doesn't really match up to existing compilers as there's not really a concept of "no platform".

gbip:
Could you please join the discussion in #375 ?

gnzlbg:
@gbip Thank you for submitting this PR! I am going to close this PR since it appears that a change like this would require more consensus. Feel free to re-open if you disagree, or after more consensus is achieved.

PR #1108, opened 2018-10-28
NetBSD: fix pthread types on arm and powerpc

jakllsch:


alexcrichton:
Can the `cfg_if!` declarations be pushed down into submodules instead of being expanded inline?

jakllsch:
@alexcrichton I was able to eliminate all added cfg_if! declarations.

PR #1109, opened 2018-10-29
revert i386-apple-ios to xcode 9.2

gnzlbg:


gnzlbg:
Seems that linking is broken as well, i've opened: https://github.com/rust-lang/rust/issues/55477

PR #1110, opened 2018-10-29
upgrade x86_64-apple-ios to xcode10

gnzlbg:


gnzlbg:
libc fails to compile for `x86_64-apple-ios` with the latest SDK, seems related to: https://github.com/rust-lang/rust/issues/55477

PR #1114, opened 2018-10-30
Add extattr(2) and extattr(3) definitions for FreeBSD and NetBSD

asomers:
DragonflyBSD does define 3 of these 14 functions.  But I elected not to
add it to the PR because I can't find any evidence that these syscalls
are used on Dragonfly, even in the base system.  And by themselves,
those three are insufficient for the xattr crate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,8 +53,6 @@ matrix:
     - env: TARGET=arm-unknown-linux-musleabihf
     - env: TARGET=aarch64-unknown-linux-gnu
     - env: TARGET=aarch64-unknown-linux-musl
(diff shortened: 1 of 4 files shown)


asomers:
cc @Stebalien

alexcrichton:
Thanks! Looks like something is failing, but it may just need and update to the rustc version on its CI?

asomers:
What do you mean?  Are you suggesting that Travis's rustc is too old?

alexcrichton:
For that one particular builder, yeah, we're pinned to an older version (presumably done a long time ago to work around some rustc bug that has since been fixed)

PR #1117, opened 2018-10-31
NetBSD: fix extattr_set_* return types

jakllsch:


alexcrichton:
The loose policy we have is that if something is tested on CI it can't break, but otherwise breakage is allowed. That's not really codified anywhere but is a rule of thumb I've used historically!

alexcrichton:
@gnzlbg oh oops just redeployed, try again?

alexcrichton:
@bors: r=gnzlbg

Oops, managed to update all the configs but one... I'll redeploy once the container is rebuilt!

PR #1118, opened 2018-11-01
Add F_RDLCK/F_WRLCK/F_UNLCK to several platforms

smaeul:
These are used by the flock wrapper in rustc_data_structures. The constants were already present in x86_64-linux-gnu and BSD (since 4928bd986907).

Currently the `flock` wrapper in `rustc_data_structures` does not work on several 32-bit musl targets, because they expect the `LARGEFILE64` version of the `F_SETLK`/`F_SETLKW` constants. This PR is a prerequisite to converting that code to use the `libc` types and constants, instead of (inaccurately) duplicating them, since `libc` already provides architecture-specific definitions of the relevant constants.

### DIFF
--- a/src/unix/haiku/mod.rs
+++ b/src/unix/haiku/mod.rs
@@ -351,6 +351,10 @@ pub const F_SETLK: ::c_int = 0x0080;
 pub const F_SETLKW: ::c_int = 0x0100;
 pub const F_DUPFD_CLOEXEC: ::c_int = 0x0200;
 
+pub const F_RDLCK: ::c_int = 0x0040;
+pub const F_UNLCK: ::c_int = 0x0200;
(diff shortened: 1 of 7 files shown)


gnzlbg:
The failure is spurious. I've restarted the `x86_64-apple-darwin` job a couple of times but it doesn't find the rust toolchain :/

gnzlbg:
@smaeul thank you for the PR. Currently `x86_64-apple-darwin` is missing from nightly (see The failure is caused by https://github.com/rust-lang/rust/issues/55571 ). I'll `r+` it as soon as the upstream issue with is solved. 

-- 

@alexcrichton it seems that i'm still not in reviewers

alexcrichton:
@gnzlbg try again now?

PR #1122, opened 2018-11-10
the f macro is unused in some targets

gnzlbg:


gnzlbg:
@alexcrichton this fails because the beta of `x86_64-apple-darwin` is still not shipped with rustup (even though the beta of `i686-apple-darwin` now started shipping again). I manually merged a PR before that had the same issue.

PR #1123, opened 2018-11-16
Use OR in the license field

newpavlov:
According to [The Manifest Format](https://doc.rust-lang.org/cargo/reference/manifest.html):
>Multiple licenses can be separated with a `/`, although that usage is deprecated.  Instead, use a license expression with AND and OR operators to get more explicit semantics.


gnzlbg:
r? @gnzlbg 

bors: r+

PR #1124, opened 2018-11-19
Add SGX target

jethrogb:
This adds support for the `x86_64-fortanix-unknown-sgx` target. See https://github.com/rust-lang/rust/pull/56066 for details.


src/lib.rs
```diff
@@ -158,12 +158,15 @@ cfg_if! {
         pub type uintptr_t = usize;
         pub type ssize_t = isize;
 
+        #[cfg(not(target_env = "sgx"))]
```
gnzlbg: Please use `cfg_if!` here to provide these for all targets except `sgx`.

src/lib.rs
```diff
@@ -158,12 +158,15 @@ cfg_if! {
...
         pub enum FILE {}
+        #[cfg(not(target_env = "sgx"))]
         pub enum fpos_t {} // TODO: fill this out with a struct
 
         pub const INT_MIN: c_int = -2147483648;
         pub const INT_MAX: c_int = 2147483647;
 
+        #[cfg(not(target_env = "sgx"))]
```
gnzlbg: Same for these (use `cfg_if`). 

Does this mean that `sgx` does not have `strlen`, `memcpy`, etc ?

src/lib.rs
```diff
@@ -284,7 +287,7 @@ cfg_if! {
 
         // These are all inline functions on android, so they end up just being entirely
         // missing on that platform.
-        #[cfg(not(target_os = "android"))]
+        #[cfg(not(any(target_os = "android", target_env = "sgx")))]
```
gnzlbg: Switch this to use `cfg_if` as well. Are these also inline functions on `sgx` ? In any case, it might be worth it to update the comment.

src/lib.rs
jethrogb: SGX doesn't have a `libc`. The definitions in here are basically just what would be in a hypothetical `ctypes` crate.

`memcpy` is provided by `compiler_builtins`

I have at times compiled a static library with some of the C functions (strcpy, etc.) to link to existing C libraries and an implementation of `malloc` in terms of the Rust allocator. Would it makes sense to add these to this crate (behind a feature)?

jethrogb: In order to add a target to `std`, it is currently required to make changes to `libc`.

> `libstd` depends on the target having a full-featured `libc`.

No it doesn't, `libstd` depends on exactly those features it uses from `libc`. In case of the SGX port, that is none.

> Also, this crate links against a `libc` library.

No it doesn't, it expects the linker to insert this dependency automatically. This doesn't happen in the SGX port.

All this basically works the same as in wasm32-unknown-unknown.

gnzlbg: > All this basically works the same as in wasm32-unknown-unknown.

`libc` should not support `wasm32-unknown-unknown` because it does not have a `libc` implementation.

alexcrichton:
It's sort of a longstanding bug that the `libc` crate has so many function definitions in its root because they're not always available. Ideally these functions should (historically) have been pushed down into the various other modules. That requires duplication, but that's ok with the `libc` crate!

Would you be ok pushing all the functions into lower modules (duplicating them) and that way the top module should only have types shared by all targets?

jethrogb:
> Would you be ok pushing all the functions into lower modules (duplicating them) and that way the top module should only have types shared by all targets?

Maybe, that seems contrary to efforts like https://github.com/rust-lang/libc/pull/1086

alexcrichton:
Hm I'm not sure I understand, how's it contrary to `c_void` and such?

jethrogb:
Sorry, I specifically meant the de-duplication that was done as part of that PR

alexcrichton:
Oh I just mean duplicating the definition, for any one platform it's still just defined once. (this is also just functions too I'm talking about as `sgx` has all the types it looks like)

jethrogb:
Made the changes suggested by @alexcrichton. This is now based on #1126.

jethrogb:
CI failure for nightly is the same as #1126

gnzlbg:
This appears to contain an spurious commit :/

alexcrichton:
Duplication is the bread and butter of the `libc` crate and has been tried and true for quite some time now. I do realize it's pretty backwards compared to how we normally work, but it really is the easiest thing to maintain in this crate, and these are the "weirdest" PRs where items are pushed down to lower platforms.

I am personally in favor of this PR and would like to merge. If there's a standard C compiler for targets then that means we have type definitions, but it may mean that there's no C functions themselves (such as for this target and wasm)

gnzlbg:
> Duplication is the bread and butter of the libc crate and has been tried and true for quite some time now. 

I kind of wish we would just put all this code which is shared by all `libc`s except `sgx` in a single place instead of duplicating it.

PR #1126, opened 2018-11-19
Factor out platforms for which libc is empty

gnzlbg:
This change shouldn't change any functionality. It just separates the platforms for which `libc` is currently empty (only `wasm32-unknown-unknown`), from those for which it isn't. This is a non-functional change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,10 @@ matrix:
     # QEMU based targets that compile in an emulator
     - env: TARGET=x86_64-unknown-freebsd
 
+    - env: TARGET=wasm32-unknown-unknown
+      install: rustup target add $TARGET
+      script: cargo build --no-default-features --target $TARGET --release
+
     - name: "Shellcheck"
       install: true
       script:
--- a/README.md
+++ b/README.md
@@ -1,8 +1,7 @@
 libc
 ====
 
-A Rust library with native bindings to the types and functions commonly found on
-various systems, including libc.
+Rust wrapper over the system's `libc`.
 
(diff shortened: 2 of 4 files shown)


jethrogb:
I'm not sure this is acceptable because of the `c_void` interoperability problem.

gnzlbg:
Note that `libc` does not export `c_void` for `wasm32-unknown-unknown`.

alexcrichton:
r=me, but looks like ci may be broken?

We'll fix this eventually for `wasm32-unknown-unknown` once the Clang ABI stabilizes and/or is released, but until that time we're gonna leave this as empty

gnzlbg:
Yes, so summarizing the conversation from IRC, `libc` should always compile for all targets, but it should be empty for targets without a libc (or similar). 

That is, `wasm32-unknown-unknown` would check that box at the moment (we can change that in the future), and the `target_env = "none"` targets would do so as well. 

So i'm going to keep this PR more or less as is until its merged, but at some point we might want to go from explicitly listing the non-supported target, to "somehow" only have items in the supported targets and be empty by default.

alexcrichton:
I believe https://github.com/rust-lang/libc/pull/1129 fixes that though

gnzlbg:
> I believe #1129 fixes that though

Damn, I'm so sorry. I was just going to ask whether that PR fixes it.

alexcrichton:
No worries! I had no idea about this as well :)

gnzlbg:
I'm trying to get `rust-semverver` up to a point that can be used to detect breaking changes to libc.

PR #1127, opened 2018-11-19
Formatting and shellcheck

gnzlbg:


.travis.yml
```diff
@@ -90,6 +81,29 @@ matrix:
     # QEMU based targets that compile in an emulator
     - env: TARGET=x86_64-unknown-freebsd
 
+    - name: "rustfmt"
+      install: rustup component add rustfmt-preview
+      script: cargo fmt --all -- --check
```
alexcrichton: I'd personally prefer to leave this out for now if that's ok, I feel like as a community we haven't figured out a solution for rustfmt that doesn't make it harder for contributors to contribute :(

gnzlbg: Ok!

PR #1132, opened 2018-11-21
core::ffi::c_void is available since rustc 1.30

glandium:


glandium:
It should also be noted that there hasn't been a release of libc with the merged use of core::ffi::c_void, while it has been available for multiple weeks now, since it was part of 1.30. Could there be a libc release with this merged?

gnzlbg:
> Could there be a libc release with this merged?

Sure I can do one. From #1082 and #1086 it appears that requiring 1.31 was an oversight, but I'll wait till @alexcrichton signs on that.

glandium:
from irc:

`<Arnavion>` It was marked as 1.31 because it was added in the 1.30 nightly, so there would be 1.30 nightlies that didn't have it
`<Arnavion>` But of course it also makes sense that it'll be in 1.30 stable, so it should be marked as 1.30
`<Arnavion>` Dunno how libc has done it in the past

gnzlbg:
That's pretty much what the #1082 PR message says.

PR #1133, opened 2018-11-21
Add missing condition for musl mips64

glandium:
unix::uclibc::mips has both support for mips and mips64, but it's
currently only imported for mips.


gnzlbg:
@bors: r+ 

There maybe was a time when `mips` used to match both.

PR #1134, opened 2018-11-21
Prepare for being included via crates.io into std

alexcrichton:
This commit prepares the `libc` crate to be included directly into the
standard library via crates.io. More details about this can be found on
rust-lang/rust#56092, but the main idea is that this crate now depends
on core/compiler-builtins explicitly (but off-by-default).

The main caveat here is that this activates `no_core` when building as
part of libstd, which means that it needs to explicitly have an `iter`
and `option` module for the expansion of `for` loops to work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,6 @@
 [package]
-
 name = "libc"
-version = "0.2.43"
+version = "0.2.44"
 authors = ["The Rust Project Developers"]
 license = "MIT OR Apache-2.0"
 readme = "README.md"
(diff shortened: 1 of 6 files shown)


gnzlbg:
LGTM.

Cargo.toml
```diff
@@ -19,10 +18,14 @@ exclude = ["/ci/*", "/.travis.yml", "/appveyor.yml"]
...
 appveyor = { repository = "rust-lang/libc", project_name = "rust-lang-libs/libc" }
 
+[dependencies]
+rustc-std-workspace-core = { version = "1.0.0", optional = true }
+
 [features]
 default = ["use_std"]
 use_std = []
 align = []
+rustc-dep-of-std = ['align', 'rustc-std-workspace-core']
```
gnzlbg: I don't know if this should enable `use_std` as well to be sure - it should be enabled by default anyways as long as libstd doesn't use `default-features=false`.

Issue #1135, opened 2018-11-22
nl_pad is private and is required to be initialized for sockaddr_nl


WillYee:
https://github.com/rust-lang/libc/blob/5b403753da9ec8ff501adf34cb6d63b319b4a3ae/src/unix/uclibc/mod.rs#L82

I was able to fix this buy marking nl_pad as pub

gnzlbg:
PR welcome :)

gnzlbg:
FWIW you don't need to access `nl_pad` to initialize the struct, e.g., you can use `MaybeUninit<sockaddr_nl>::zeroed()` to zero the struct, and then only initialize the relevant fields.

gnzlbg:
Alternatively, you can use `let x: sockaddr_nl = unsafe { mem::zeroed() };` on stable Rust, although if you can use nightly `MaybeUninit` is preferable.

PR #1137, opened 2018-11-22
Marking nl_pad as public so we can construct sockaddr_nl

WillYee:
Addressing issue 1135

https://github.com/rust-lang/libc/issues/1135

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -384,7 +384,7 @@ s! {
 
     pub struct sockaddr_nl {
         pub nl_family: ::sa_family_t,
-        nl_pad: ::c_ushort,
+        pub nl_pad: ::c_ushort,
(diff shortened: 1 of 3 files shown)


gnzlbg:
I suppose that `nl_pad` are padding bits required for the ABIs to match. I wonder if one shouldn't be using here`repr(align)`, `repr(packed)`, etc. instead. @alexcrichton ?

alexcrichton:
`repr(align)` is fine to use yeah, `repr(packed)` I believe is still feature gated

gnzlbg:
Any progress here? @WillYee please use `repr(align)`/`repr(packed)`/etc. to remove the padding field instead of making it public.

gnzlbg:
I'm closing this, since it is intended for this to be an internal implementation detail. See #1135 for how to initialize these structs.

Issue #1136, opened 2018-11-22
0.2.44 breaks fairly recent compiler 


przygienda:
```
error[E0432]: unresolved import `core::ffi::c_void`
    --> /b/prz/PVT_RIFT_184/obj//cargo_home/registry/src/github.com-1ecc6299db9ec823/libc-0.2.44/src/unix/mod.rs:1151:17
     |
1151 |         pub use core::ffi::c_void;
     |                 ^^^^^^^^^^^^^^^^^ no `c_void` in `ffi`

error[E0432]: unresolved import `core::ffi::c_void`
    --> /b/prz/PVT_RIFT_184/obj//cargo_home/registry/src/github.com-1ecc6299db9ec823/libc-0.2.44/src/unix/mod.rs:1151:17
     |
1151 |         pub use core::ffi::c_void;
     |                 ^^^^^^^^^^^^^^^^^ no `c_void` in `ffi`

error[E0641]: cannot cast to a pointer of an unknown kind
   --> /b/prz/PVT_RIFT_184/obj//cargo_home/registry/src/github.com-1ecc6299db9ec823/libc-0.2.44/src/unix/bsd/freebsdlike/mod.rs:864:56
    |
864 | pub const PTHREAD_MUTEX_INITIALIZER: pthread_mutex_t = 0 as *mut _;
    |                                                        ^^^^^------
    |                                                             |
    |                                                             help: consider giving more type information
    |
    = note: The type information given here is insufficient to check whether the pointer cast is valid

error[E0641]: cannot cast to a pointer of an unknown kind
   --> /b/prz/PVT_RIFT_184/obj//cargo_home/registry/src/github.com-1ecc6299db9ec823/libc-0.2.44/src/unix/bsd/freebsdlike/mod.rs:865:54
    |
865 | pub const PTHREAD_COND_INITIALIZER: pthread_cond_t = 0 as *mut _;
    |                                                      ^^^^^------
    |                                                           |
    |                                                           help: consider giving more type information
    |
    = note: The type information given here is insufficient to check whether the pointer cast is valid

error[E0641]: cannot cast to a pointer of an unknown kind
   --> /b/prz/PVT_RIFT_184/obj//cargo_home/registry/src/github.com-1ecc6299db9ec823/libc-0.2.44/src/unix/bsd/freebsdlike/mod.rs:866:58
    |
866 | pub const PTHREAD_RWLOCK_INITIALIZER: pthread_rwlock_t = 0 as *mut _;
[...]

PR #1138, opened 2018-11-22
Add build job for Rust 1.30 version compatibility

gnzlbg:
From that version on we use `core::ffi::c_void` instead of our own `c_void`.


alexcrichton:
Is this still needed?

gnzlbg:
No, I thought this was going to break CI, but it didn't. See #1136 .

PR #1144, opened 2018-11-23
Move fuchsia to unix/notbsd/linux/fuchsia

gnzlbg:
cc @cramertj It appears that fuchsia has items both in the library top level (libc/src/fuchsia ) but that also all linux items should be included for it (since there are target_os macros for fuchsia in unix/notbsd and unix/notbsd/linux). 

Is this change correct?


alexcrichton:
Is this verified to compile? Is there a reason to change this?

gnzlbg:
That's a good question. It's either moving fuchsia into `notbsd`, or deleting the `target_os = fuchsia` from `notbsd` since they are never reached. I'll wait till @cramertj or someone from fuchsia answers here.

cramertj:
`cargo build --target=x86_64-fuchsia` gives lots of errors on duplicate imports:

```
error[E0659]: `c_char` is ambiguous                                                                                                                                                                                
    --> src/unix/notbsd/linux/mod.rs:2148:31                                                                                                                                                                       
     |                                                                                                                                                                                                             
2148 |                  mode: *const c_char) -> *mut ::FILE;                                                                                                                                                       
     |                               ^^^^^^ ambiguous name                                                                                                                                                         
     |                                                                                                                                                                                                             
note: `c_char` could refer to the name imported here                                                                                                                                                               
    --> src/unix/notbsd/linux/mod.rs:2247:17                                                                                                                                                                       
     |                                                                                                                                                                                                             
2247 |         pub use self::musl::*;                                                                                                                                                                              
     |                 ^^^^^^^^^^^^^                                                                                                                                                                               
note: `c_char` could also refer to the name imported here                                                                                                                                                          
    --> src/unix/notbsd/linux/mod.rs:2264:17                                                                                                                                                                       
     |                                                                                                                                                                                                             
2264 |         pub use self::fuchsia::*;    
```

gnzlbg:
@cramertj do you think we could add fuchsia to `libc`'s CI somehow ?

Issue #570, opened 2017-04-07
How to deal with breaking changes on platform ? [OpenBSD]


semarie:
I open an issue on libc because it is here the problems will start to show up. Depending the solution or the way to deal with, modifications could occurs in rustc repository too.

At OpenBSD, we don't care about breaking API/ABI between releases. Once a release is done, the API/ABI is stable, but there is no guarantee that it will be compatible with the next release.

Currently, in the upcoming 6.2 version of OpenBSD (6.1-current), there is a breaking change that will affect libc : [si_addr should be of type void *, not char * (caddr_t)](http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/sys/sys/siginfo.h.diff?r1=1.11&r2=1.12). Here the current definition in [libc](https://github.com/rust-lang/libc/blob/master/src/unix/bsd/netbsdlike/openbsdlike/mod.rs#L110).

Under OpenBSD, we deal with ABI under LLVM by using a triple like: `amd64-unknown-openbsd6.1`. For Rust, instead we use an unversioned platform, resulting all OpenBSD versions to define the same ABI (which isn't properly right).

Do you think it possible to switch from `*-unknown-openbsd` to `*-unknown-openbsd6.0`, `*-unknown-openbsd6.1`, ... without having to duplicate all code in libc for each target ? and without having to add a new target in rustc for each OpenBSD release ?

Any others ideas on the way to deal with it ?

alexcrichton:
Unfortunately I don't really know how we'd handle this, I just figured that platforms wouldn't do this. 

If this happens a lot we'll just need to document what's wrong and stop adding new bindings, it'll be up to crates to implement version compatibility.

semarie:
I think it isn't just a "version compatibility" issue. My purpose isn't to have a compatibility layer for missing/removed functions or types.

The problem is OpenBSD triple is versioned, meaning that API/ABI of one version could be different from another version.

I checked some others system (running `llvm-config --host-target` to see if the triple is versioned or not), and it seems it is a common situation in not-Linux world:
  - x86_64-apple-darwin16.0.0
  - x86_64-unknown-freebsd12.0
  - x86_64-unknown-freebsd11.0
  - i386-unknown-openbsd5.8
  - x86_64-unknown-netbsd7.99

I also checked in LLVM source tree: the OS version is a part of the triple definition. 
[...]

PR #1145, opened 2018-11-23
dragonfly: add back missing errnos

levex:
A couple of errno's went missing from libc in commit 8293ced
(accidentally, I think). However, these are still in DragonflyBSD's
`sys/sys/errno.h` file and this breakage causes the `nix` crate to not
build.

One can see the `errno.h` file [here](http://gitweb.dragonflybsd.org/dragonfly.git/blob/25fa222:/sys/sys/errno.h#l188).

Signed-off-by: Levente Kurusa <lkurusa@acm.org>

### DIFF
--- a/src/unix/bsd/freebsdlike/dragonfly/mod.rs
+++ b/src/unix/bsd/freebsdlike/dragonfly/mod.rs
@@ -376,6 +376,14 @@ pub const CTL_P1003_1B_SIGQUEUE_MAX: ::c_int = 24;
 pub const CTL_P1003_1B_TIMER_MAX: ::c_int = 25;
 pub const CTL_P1003_1B_MAXID: ::c_int = 26;
 
+pub const ENOMEDIUM: ::c_int = 93;
+pub const EUNUSED94: ::c_int = 94;
+pub const EUNUSED95: ::c_int = 95;
+pub const EUNUSED96: ::c_int = 96;
+pub const EUNUSED97: ::c_int = 97;
+pub const EUNUSED98: ::c_int = 98;
+pub const EASYNC: ::c_int = 99;
+
 pub const EVFILT_READ: ::int16_t = -1;
 pub const EVFILT_WRITE: ::int16_t = -2;
 pub const EVFILT_AIO: ::int16_t = -3;


asomers:
I don't think it's wise for Rust to define the`EUNUSED*` variables.  Presumably there is no valid reason for an application to use them, and their names suggest that they might disappear in a future version of Dragonfly.  If one of those variables gets replaced by a real errno, then libc won't compile on future versions of Dragonfly.

gnzlbg:
>  If one of those variables gets replaced by a real errno, then libc won't compile on future versions of Dragonfly.

How come ? If Dragonfly removes these, we can deprecate them, but these are just `const` values. They aren't linked against anything.

levex:
> I don't think it's wise for Rust to define the`EUNUSED*` variables. Presumably there is no valid reason for an application to use them, and their names suggest that they might disappear in a future version of Dragonfly. If one of those variables gets replaced by a real errno, then libc won't compile on future versions of Dragonfly.

They are already part of `nix`: https://github.com/nix-rust/nix/blob/master/src/errno.rs#L525

(which is the principal reason behind introducing these back to `libc`)

gnzlbg:
@asomers @levex it would be nice if any of the dragonflybsd stakeholders could get CI for it up and running in libc. Otherwise dragonfly would "continue" to break.

gnzlbg:
> libc's CI (currently not running on Dragonfly) checks constants as well as functions. It will not allow an undefined constant.

Actually, libc's CI checks only those constants that it does not skip - there are many constants that CI doesn't check. 

> And libc can't deprecate constants on new versions of Dragonfly only; libc must be the same on all versions. 

libc can deprecate constants on all Dragonfly versions as well as change the version of Dragonfly that it supports.

> libc must be the same on all versions. Basically, libc can never bind any symbol that might ever change.

Dragonfly is not tested on CI, so I cannot comment about it. But pretty much every single target tested on libc's CI has changed _many_ of the symbols that libc exposes. libc doesn't target all versions of a platform - it only usually targets a single version.

levex:
> @asomers @levex it would be nice if any of the dragonflybsd stakeholders could get CI for it up and running in libc. Otherwise dragonfly will "continue" to break.

I would love to do this, but currently I'm not sure if any CI supports Dfly and/or there is a CI where I can upload a custom ISO. Any suggestions?

gnzlbg:
You can pretty much run everything on travis CI on top of qemu - libc uses qemu for freebsd for example.

levex:
> You can pretty much run everything on travis CI on top of qemu - libc uses qemu for freebsd for example.

Sweet, thanks. I'll build it over the weekend!

asomers:
> > libc's CI (currently not running on Dragonfly) checks constants as well as functions. It will not allow an undefined constant.
> 
> Actually, libc's CI checks only those constants that it does not skip - [there are many constants that CI doesn't check](https://github.com/rust-lang/libc/blob/master/libc-test/build.rs#L504).
> 
> > And libc can't deprecate constants on new versions of Dragonfly only; libc must be the same on all versions.
> 
> libc can deprecate constants on all Dragonfly versions as well as change the version of Dragonfly that it supports.
> 
> > libc must be the same on all versions. Basically, libc can never bind any symbol that might ever change.
> 
> Dragonfly is not tested on CI, so I cannot comment about it. But pretty much every single target tested on libc's CI has changed _many_ of the symbols that libc exposes, particularly the values of constants. As designed and implemented, libc doesn't target all versions of a platform - it only targets a single version, and that's the version used in libc's CI.

Yes, many targets have changed their symbols.  It's a _major_ problem in libc (https://github.com/rust-lang/libc/issues/570).  But until that problem is fixed, we can't ignore it.  The best thing to do is not to bind symbols that might change in backwards-incompatible ways.  Because if they do change, then libc's bindings will be incorrect on either the old or new version.  And incorrect bindings are worse than incomplete bindings.

So I'll ask again: is there any legitimate reason for an application to use one of the `EUNUSED` errnos?  If not, then libc absolutely shouldn't bind them.

asomers:
> You are assuming that libc should support different versions of a platform. Whether it should do so or not is unclear - doing so would require a different library with different trade-offs (e.g. binding generations, or cfg for platform versions, or ...).
> 
> Right now, however, libc is for better or worse designed to support a single version of a platform. If we don't want to add any symbols that could possibly change, then we should not add any `const`s, `struct`, `static`s, `fn`, ... items because most platforms have already broken most items of these kinds.

That's not quite true.  libc works fine to support multiple versions of a platform when successive versions add symbols.  It just doesn't work when new versions change or remove symbols.  The `EUNUSED*` constants are perfect examples of symbols that are likely to be removed in the future.

levex:
FYI: DragonflyBSD is currently discussing whether to remove these values from the system altogether.

gnzlbg:
@levex if that's the case the path forward for nix is probably to stop relying on these for a build. Should this PR be closed?

levex:
> @levex if that's the case the path forward for nix is probably to stop relying on these for a build. Should this PR be closed?

Let's wait a little more to see if anyone on the Dragonfly team is raising an issue with the removal. I'll get back to this PR when there's a consensus.

levex:
So we've reached consensus and DragonflyBSD will be removing the `EUNUSED*` values. Thus, @gnzlbg I think we can safely close this PR. @asomers I will submit a separate PR to `nix` to remove the values from there too, once the dfly patch is committed upstream.

asomers:
What's the fate of `ENOMEDIUM` and `EASYNC`?

gnzlbg:
Awesome, thanks for looking into this, i'm closing this PR, feel free to open new ones for more stable `const`s!

asomers:
So do `ENOMEDIUM` and `EASYNC` not matter?

gnzlbg:
If they do, please open a different PR to add those just those, or apply the necessary changes to this PR and re-open it.

levex:
@asomers , `ENOMEDIUM` is used by other platforms, too, I believe it's in dfly for compatibility reasons with Linux applications. `EASYNC` as far as I can tell is defined but not used in the codebase: https://grok.dragonflybsd.org/search?project=dports&q=&defs=&refs=EASYNC&path=&hist=&type=&si=refs&n=25

I'll see if we can remove that, too.

For future reference, the commit that removed `EUNUSED`: http://gitweb.dragonflybsd.org/dragonfly.git/commit/af7bdca177816bfbe15f961d0b0e1968219402e3

PR #1146, opened 2018-11-23
Add a couple pthread function definitions and get_sched for musl.

mwanner:
I missed `pthread_setaffinity_np` for the musl variant, so I added a few missing function declarations for musl.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,12 +7,13 @@ readme = "README.md"
 repository = "https://github.com/rust-lang/libc"
 homepage = "https://github.com/rust-lang/libc"
 documentation = "http://doc.rust-lang.org/libc"
-description = """
-A library for types and bindings to native C functions often found in libc or
-other common platform libraries.
-"""
+keywords = ["libc", "ffi", "bindings", "operating", "system" ]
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! Looks like there's some errors on CI though?

mwanner:
> Thanks! Looks like there's some errors on CI though?

Hm.. interesting, test coverage without writing a test. Nice.

How do I run these tests locally?  Fails for me due to not finding `linux/input.h` (for target `x86_64-unknown-linux-musl`) or `linux/module.h` (against default system glibc).

mwanner:
> You need to use the docket containers, check .travis.yml for how the run_docker script is called.

Ah, thanks, I'll try that.

mwanner:
> Thanks! Looks like there's some errors on CI though?

I dropped the definitions for functions that musl really doesn't provide, squashed the two commits and rebased onto master.  This seems to pass CI just fine, now.  Please have another look.

gnzlbg:
Looks like you've rebased one commit too much :/  

Otherwise LGTM. @alexcrichton this can be merged as is.

Issue #981, opened 2018-04-23
tests failing on macOS High Sierra


mcginty:
This is from running the tests on my laptop.
```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.13.4
BuildVersion:   17E199
```

failure output:
```
bad syslog function pointer: rust: 140734611868871 (0x7fff548c4cc7) != c 140734611869453 (0x7fff548c4f0d)
bad VM_FLAGS_USER_ALLOCATE value at byte 0: rust: 27 (0x1b) != c 31 (0x1f)
bad VM_FLAGS_USER_MAP value at byte 0: rust: 27 (0x1b) != c 31 (0x1f)
```

and here's the relevant snippet of my `/usr/include/mach/vm_statistics.h` of the conflicting VM_* constants. I'm not sure how to help with the syslog function pointer error, but happy to give more info as requested.

```
/* These are the flags that we accept from user-space */
#define VM_FLAGS_USER_ALLOCATE	(VM_FLAGS_FIXED |		\
				 VM_FLAGS_ANYWHERE |		\
				 VM_FLAGS_PURGABLE |		\
				 VM_FLAGS_4GB_CHUNK |		\
				 VM_FLAGS_RANDOM_ADDR |		\
				 VM_FLAGS_NO_CACHE |		\
				 VM_FLAGS_OVERWRITE |		\
				 VM_FLAGS_SUPERPAGE_MASK |	\
				 VM_FLAGS_ALIAS_MASK)
#define VM_FLAGS_USER_MAP	(VM_FLAGS_USER_ALLOCATE |	\
				 VM_FLAGS_RETURN_4K_DATA_ADDR |	\
				 VM_FLAGS_RETURN_DATA_ADDR)
```

alexcrichton:
Thanks for the report! These sorts of things can change over time and it should be fine to tweak the values as they change.

dpogretskiy:
MacOS's libc, which is somewhere in their SDK, is not based on glibc, there are a lot of inconsistencies, such as different struct layouts, same consts with different values and so on. Maybe disable libc build for OS X entirely since it's not reliable in any way, unfortunately. Other then full sanity check and somewhat reimplemenation for OSX and probably BSD targets.

gnzlbg:
>  Thanks for the report! These sorts of things can change over time and it should be fine to tweak the values as they change.

[...]

Issue #1142, opened 2018-11-22
Policy of supported platform version for constant values


gnzlbg:
We have many bugs where the value of a constant does not match the value of the constant in the users system. 

I think that, at least for constants, we should support the latest released version of the platform, and that we do not consider upgrading a constant value a breaking change. 

If the `libc` crate ever gets support for multiple versions of the same platform, we should revisit this policy.

alexcrichton:
I agree! For now upating constants as they're updated on the system hasn't caused problems and seems like we should continue to do so

PR #1147, opened 2018-11-23
OS X version cfg

palfrey:
Using `xcrun --show-sdk-version` I think we can get an accurate version of the current in-use OS X SDK. This is often, but not always equal to the OS version.

Using this, I then fix the `NET_RT_MAXID` issue I flagged in #981

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -10,6 +10,12 @@ fn main() {
     if rustc_minor_version().expect("Failed to get rustc version") >= 30 {
         println!("cargo:rustc-cfg=core_cvoid");
     }
+
+    if cfg!(any(target_os = "macos", target_os = "ios")) {
+        let xcrun = Xcrun::version();
+        println!("cargo:rustc-cfg=osx_major=\"{}\"", xcrun.0);
+        println!("cargo:rustc-cfg=osx_minor=\"{}\"", xcrun.1);
+    }
 }
 
 fn rustc_minor_version() -> Option<u32> {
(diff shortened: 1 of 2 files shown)


gnzlbg:
If `HOST == TARGET` then this will work and be correct, the problem is that this will be incorrect if `HOST != TARGET`. To make this correct, we would have to find out the OS version of the `TARGET` but right now there is no way to do that. 

I think the `HOST != TARGET` is not very common for people using the `libc` crate directly, but it is probably the case that applies when building `libstd`, which is probably how most people are obtaining their `libc`. 

Iff we want to allow users to target multiple OSX versions, then I think we have to think this a bit more through:
  * we still have no way to tell the OSX version of the target, we would have to somehow pass it as part of the target triple, and we might want to have the default targets still target the last OSX version anyways for convenience 
  * we would have to test the different OSX on CI - [`mach` supports 5 OSX versions and 5 OSX targets (travis-CI)](https://travis-ci.org/fitzgen/mach), that's a big increase in CI load.
  * most people get `libc` via `libstd`, so we would need to make sure that `libstd` builds `libc` for the OSX version they use

For this particular "breaking" change, given all of the above, I think we should just update the constants to the values of the latest version which is what probably most MacOS X users will be using as a `HOST` and a `TARGET` (see #1142 ). If we want to do better here by supporting multiple OSX versions, we might end up having the default `apple` targets targeting the latest OSX version by default for convenience anyways.

palfrey:
Ok. In this particular case it's broken by Travis not having a 10.14 image yet (and it's taking a while - see https://github.com/travis-ci/travis-ci/issues/10155), but I'll close this PR now.

PR #1151, opened 2018-11-26
Add some TCP constants

Fensteer:
Add some TCP constants according to the sources : https://github.com/freebsd/freebsd/blob/master/sys/netinet/tcp.h#L164

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -1549,7 +1549,6 @@ pub const IPV6_LEAVE_GROUP: ::c_int = 13;
 pub const IPV6_PKTINFO: ::c_int = 46;
 pub const IPV6_RECVPKTINFO: ::c_int = 61;
 
-pub const TCP_NODELAY: ::c_int = 0x01;
 pub const TCP_KEEPALIVE: ::c_int = 0x10;
 
 pub const SOL_LOCAL: ::c_int = 0;
--- a/src/unix/bsd/freebsdlike/dragonfly/mod.rs
+++ b/src/unix/bsd/freebsdlike/dragonfly/mod.rs
@@ -682,6 +682,11 @@ pub const IPPROTO_DONE: ::c_int = 257;
 /// Used by RSS: the layer3 protocol is unknown
 pub const IPPROTO_UNKNOWN: ::c_int = 258;
 
+// sys/netinet/tcp.h
+pub const TCP_SIGNATURE_ENABLE:   ::c_int = 16;
+pub const TCP_KEEPINIT:   ::c_int = 32;
+pub const TCP_FASTKEEP:   ::c_int = 128;
+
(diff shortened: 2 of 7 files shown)


asomers:
Did you check the DragonflyBSD sources?  I bet that many of these symbols are defined identically there.  If so, they belong in  `src/unix/bsd/freebsdlike/mod.rs`.  Unless, of course, NetBSD and OpenBSD also define them identically, in which case they can go in `src/unix/bsd/mod.rs`.

Fensteer:
I tried to put it in `src/unix/bsd/mod.rs` but darwin targets failed: https://travis-ci.org/rust-lang/libc/builds/459839121
Dragonfly doesn't defined the same constants: https://github.com/DragonFlyBSD/DragonFlyBSD/blob/e7ab884bd49753f8884eb597d10d6569a08fa0df/sys/netinet/tcp_var.h#L164

asomers:
That's the wrong file.  You want tcp.h, not tcp_var.h.  Dragonfly does define some of the same constants.
https://github.com/DragonFlyBSD/DragonFlyBSD/blob/e7ab884bd49753f8884eb597d10d6569a08fa0df/sys/netinet/tcp.h

Fensteer:
Thanks for the file. I factor common TCP variables between different OS.
Sources: 
https://github.com/DragonFlyBSD/DragonFlyBSD/blob/master/sys/netinet/tcp.h#L156
https://github.com/freebsd/freebsd/blob/master/sys/netinet/tcp.h#L164
https://ftp.netbsd.org/pub/NetBSD/NetBSD-current/src/sys/netinet/tcp.h
https://github.com/openbsd/src/blob/master/sys/netinet/tcp.h#L126
https://github.com/bitrig/bitrig/blob/master/sys/netinet/tcp.h

Fensteer:
Done, I tried to add TCP_CA_NAME_MAX but it's not available in userspace so I removed it.

gnzlbg:
Could you rebase ?

semarie:
FYI, it broke the build on OpenBSD (there is no CI). I am working on unbreak it

gnzlbg:
Damn. Would it be possible to contribute an OpenBSD build bot to CI similar to how the FreeBSD build bot works ?

asomers:
@gnzlbg Are you thinking of Nix's buildbot?  libc doesn't use buildbot, only Travis.  libc's FreeBSD builder runs in QEMU on Travis.  AFAIK the same technique should be possible for OpenBSD.

semarie:
for FreeBSD, the CI does a crosscompilation from linux to freebsd, and run the executable via qemu. Last time I tried to do crosscompilation from linux to openbsd it was really difficult. Maybe it could be tested again as now we recently switched to clang + ld.lld.

gnzlbg:
@asomers no I was thinking of what @semarie mentions, cross compiling from linux to OpenBSD and then running the binaries under qemu.

Issue #1148, opened 2018-11-24
Docker build issues on Travis


palfrey:
Seeing this for various builds on #1130 and #1146 and probably some other PRs. Errors vary, but some examples are:
* https://travis-ci.org/rust-lang/libc/jobs/459072481
```
Step 1/4 : FROM ubuntu:18.04
error parsing HTTP 404 response body: invalid character 'p' after top-level value: "404 page not found\n"
```
* https://travis-ci.org/rust-lang/libc/jobs/459075930
```
Step 1/14 : FROM ubuntu:16.04
unauthorized: authentication required
```
* https://travis-ci.org/rust-lang/libc/jobs/459075937
```
Step 1/5 : FROM ubuntu:17.10
17.10: Pulling from library/ubuntu
age configuration: unknown blob
```

None of them AFAIK have anything to do with the PRs in question, but are making testing/merging much harder.

gnzlbg:
I've retried this, sometimes CI times out for a variety of reasons. I don't know if there is something that we can do about this specific error.

palfrey:
I'm pretty sure this is actually just something wrong with Travis, and so I've finally given in and emailed their support email :)

gnzlbg:
cc @kennytm

palfrey:
Build is improved (at least for #1130), but still slightly off. https://travis-ci.org/rust-lang/libc/jobs/459715534 is the only failure of the last one, and that's a `curl: (28) Operation timed out after 300808 milliseconds with 0 out of 0 bytes received` which might eventually be worth using a [travis_retry](https://docs.travis-ci.com/user/common-build-problems/#travis_retry) on

PR #1152, opened 2018-11-26
Retry all curl operations

palfrey:
As the docker pulls from #1148 appear mostly fixed, but we're seeing occasional curl timeouts still (e.g. https://travis-ci.org/rust-lang/libc/jobs/459715534, https://travis-ci.org/rust-lang/libc/jobs/459611627), so this should improve things at least a bit.


gnzlbg:
Thanks!

PR #1153, opened 2018-11-27
Remove unreachable fuchsia branches

gnzlbg:
cc @cramertj

So this should be a non-functional change because these branches are unreachable, but please try to build fuchsia with this change to make sure this doesn't break anything.

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -347,10 +347,6 @@ cfg_if! {
         #[link(name = "root")]
         #[link(name = "network")]
         extern {}
-    } else if #[cfg(target_os = "fuchsia")] {
-        #[link(name = "c")]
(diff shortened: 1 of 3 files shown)


cramertj:
LGTM-- the `unix/` code will never be included on fuchsia now because of [this if](https://github.com/rust-lang/libc/blob/master/src/lib.rs#L187).

gnzlbg:
@cramertj is there a Rust project building fuchsia on CI ? it would be cool if we could had some way to test that fuchsia doesn't break.

cramertj:
No, there isn't any currently Rust project running tests on an up-to-date Fuchsia in CI. This is something we'd like to have and there have been efforts towards getting this, but there have been some obstacles.

PR #1158, opened 2018-12-02
define `TIMER_ABSTIME` on BSD

strake:


gnzlbg:
Looks like this constant is not available on Apple systems.

gnzlbg:
Are you sure its available on all bsd systems? (OpenBSD, DragonflyBSD, NetBSD, FreeBSD, etc.) ?

jbeich:
> Looks like this constant is not available on Apple systems.

Not surprising given Apple rarely if ever updates BSD bits.

> Are you sure that it's available on all bsd systems?

See http://fxr.watson.org/fxr/trackident?i=TIMER_ABSTIME

gnzlbg:
These still need to be omitted from Apple platforms.

strake:
@gnzlbg done

PR #1159, opened 2018-12-02
Add support for SysV semaphores on Apple platform

cmbrandenburg:


gnzlbg:
Thanks! It appears that the style needs to be fixed, and that there are some error on the apple build jobs.

alexcrichton:
It looksl ike `semctl` may have a newer name and `repr(packed)` may still be unstable?

cmbrandenburg:
@gnzlbg and @alexcrichton, thanks. The CI test logs can be a bit overwhelming to a newcomer on this project.

Here are the function declarations from the `<sys/sem.h>` header on my Apple:

```c
int     semctl(int, int, int, ...) __DARWIN_ALIAS(semctl);                                
int     semget(key_t, int, int);                                                          
int     semop(int, struct sembuf *, size_t);
```

The `__DARWIN_ALIAS` looks suspicious, and I have no idea what it does. I'll need to look into that.

Also, the `semid_ds` structure definition is `#ifdef`'d, with one definition using `#pragma pack(4)` and the other being not-packed. I need to figure out what's going on there, too. Maybe I need to add conditional compilation to my Rust code.

I'll keep working on this PR. However, just as a heads up, this is for research on the side—not my main work—so my time for this is limited during the weekdays, and it may take a while to push this PR over the goal line.

gnzlbg:
> Also, the semid_ds structure definition is #ifdef'd, with one definition using #pragma pack(4) and the other being not-packed.

I see. That's pretty normal for Apple headers, and it means you have to use `repr(packed(4))` which is unstable. Search the library for `packed(4)` and you can copy how the other `struct`s in the Apple API do it.

cmbrandenburg:
Looks like there's maybe an unrelated build problem happening?

From the test log, `TARGET=aarch64-linux-android`:

```text
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

cmbrandenburg:
I rebased and squashed to clean up the commit history.

alexcrichton:
@bors: r+

Thanks!

alexcrichton:
@gnzlbg FWIW if this keeps happening I think we should just remove Android CI, it's failing practically every other build to prevent it from actually being useful...

@bors: retry

cmbrandenburg:
Thanks, @gnzlbg and @alexcrichton, for all the help!

PR #1161, opened 2018-12-04
Add AF_XDP, PF_XDP and SOL_XDP consts for linux

tklauser:
These are needed to work with `AF_XDP` sockets, see https://www.kernel.org/doc/html/v4.19/networking/af_xdp.html

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -672,6 +672,9 @@ fn main() {
             // shouldn't be used in code anyway...
             "AF_MAX" | "PF_MAX" => true,
 
+            // These are not in a glibc release yet, only in kernel headers.
+            "AF_XDP" | "PF_XDP" | "SOL_XDP" if linux => true,
+
             // Present on historical versions of iOS, but now removed in more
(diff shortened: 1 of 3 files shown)


tklauser:
The failed libc-test is because these consts are not yet in a released version of glibc. I guess this has to wait until there is a glibc release with these defined.

gnzlbg:
The libc Version in the docker containers might be outdated. You might want
to bump the Ubuntu version, or skip these constants in the libc-Test/
build.rs

tklauser:
Thanks @gnzlbg. I added these constants to be skipped in `libc-test/build.rs` since I couldn't find a libc version containing them in any of the Ubuntu releases available for the docker containers.

gnzlbg:
@tklauser please open an issue so that we don't forget testing these once we update the docker images in the future.

@bors: r+

tklauser:
> @tklauser please open an issue so that we don't forget testing these once we update the docker images in the future.

@gnzlbg Thanks, filed issue #1162 for that.

PR #1164, opened 2018-12-05
musl: 64-bit: Ensure proper (c)msghdr layout on big-endian

awilfox:
Discovered while debugging: djg/audioipc-2#52


awilfox:
Neither of these failures seem like issues with the change in this PR.  The mipsel-musl one seems to be because the toolchain can't be found, and the arm64 Android one seems like an issue with the Android device/emulator.

alexcrichton:
@bors: retry

Nah just flakiness in our CI!

PR #1165, opened 2018-12-05
Added setegid function

xd009642:
Adding the `setegid` function to libc. I had a look and this is in Apple, Linux and Fuchsia so I'm hoping it works for all the BSDs etc. But if anything fails in CI I'll change it accordingly.


gnzlbg:
> so I'm hoping it works for all the BSDs etc

OpenBSD, NetBSD, FreeBSD, and DragonflyBSD all have it AFAIK so this should work fine.

xd009642:
I'm guessing by the retries this isn't something wrong with the PR but a CI/testing issue? (Or believed to be)

gnzlbg:
@xd009642 no I don't think anything is wrong with the PR, so don't worry about it. CI just happens to be a bit "picky" today.

PR #1167, opened 2018-12-07
Move FreeBSD testing from Travis/QEMU to Cirrus-CI

### LINKED ISSUE #1163: Consider moving FreeBSD tests to Cirrus-CI
libc's FreeBSD tests are currently using QEMU on Travis-CI.  This is bad for three reasons:
1) It's slow
2) It's complicated
3) Somebody (@wezm I believe) has to build a new image for every release.

Cirrus-ci.com just launched their new FreeBSD service.  It works much like Travis, but it has the option of using full VMs instead of just containers.  They currently provide images for four different versions of FreeBSD, and building new images is probably easier than with QEMU.  https://cirrus-ci.org/guide/FreeBSD/ .  Here's an example of a Rust project using Cirrus: https://github.com/asomers/tokio-file/pull/6 .

asomers:
Fixes #1163


bors.toml
```diff
@@ -0,0 +1,4 @@
+# Gate on both Travis CI and Cirrus-ci
+status = ["continuous-integration/appveyor/branch",
+	  "continuous-integration/travis-ci/push",
+	  "stable x86_64-unknown-freebsd"]
```
alexcrichton: I don't think the homu/bors combo we have recognizes this file?

alexcrichton: I think [this may need to be modified](https://github.com/rust-lang/rust-central-station/blob/d28d4fd292bc3a7db495ba93b4065858c9428230/homu.toml.template#L244-L265) though to gate on Cirrus CI

asomers: The docs say that it's required, and don't mention any default values.  As is, I suspect that bors would allow you to merge a PR with a failing build status.

asomers: Does libc have its own fork of bors or something?

alexcrichton: No it uses the same as rust-lang/rust, which doesn't need this file

.cirrus.yml
```diff
@@ -0,0 +1,14 @@
...
+  # This name gets reported as a build status in GitHub
+  name: stable x86_64-unknown-freebsd
+  setup_script:
+    - pkg install -y curl
+    - curl https://sh.rustup.rs -sSf --output rustup.sh
+    - sh rustup.sh -y
+  test_script:
+    - . $HOME/.cargo/env
+    - cd libc-test
+    - cargo test
```
gnzlbg: Would it be possible for this to just use `sh ./ci/run.sh` ?

gnzlbg:
Thank you!

PR #1172, opened 2018-12-10
Fix for rust-lang/rust issue #50583

piersfinlayson:
Rationale for this fix is in the issue - see https://github.com/rust-lang/rust/issues/50583.

To test I've created a container with the new Dockerfile and verified that the libc.a doesn't have the illegal instruction reported in that issue.  Not sure how to do a more comprehensive test of the CI that presumably produces a liblibc.rlib with this libc.a in it though.

### DIFF
--- a/ci/docker/arm-unknown-linux-musleabihf/Dockerfile
+++ b/ci/docker/arm-unknown-linux-musleabihf/Dockerfile
@@ -7,7 +7,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 RUN curl --retry 5 https://www.musl-libc.org/releases/musl-1.1.19.tar.gz | tar xzf -
 WORKDIR /musl-1.1.19
 RUN CC=arm-linux-gnueabihf-gcc \
-    CFLAGS="-march=armv6 -marm" \
+    CFLAGS="-march=armv6 -marm -mfpu=vfp" \
     ./configure --prefix=/musl-arm --enable-wrapper=yes
 RUN make install -j4
 


alexcrichton:
Thanks for the PR! Did this mean to go to rust-lang/rust though?

piersfinlayson:
Sorry if I've followed the wrong process here - this is my first PR for a submodule.  Should I submit the pull request via rust-lang/rust by updating src/liblibc to point to my forked version of libc?

gnzlbg:
What do you expect to achieve with this PR? This will change how `libc` is tested here, but that won't affect how the `libc` that's shipped via `rust-lang/rust` is compiled.

piersfinlayson:
@gnzlbg could you or someone else share how the liblibc.rlib - that gets delivered as part of the relevant target when installing via rustup -  gets built?  I assumed it was via this container (as this is the only thing I've found in the rust source which generates a libc.a - which the delivered musl liblibc.rlib is statically linked with), but don't quite understand how the rust CI pipeline and deliver fits together.

gnzlbg:
When building `rust-lang/rust`, the git submodules are fetched, which includes fetching some commit from this repository. Then `libc` is built there as part of building the rust standard library (`libstd`). 

You'd have to ask @alexcrichton about which service and containers are used to actually build the release artifacts that are distributed via rustup, but I'd suppose that's done on Travis-CI, using the docker containers in the rust-lang/rust repo: https://github.com/rust-lang/rust/tree/master/src/ci/docker

piersfinlayson:
When building rust-lang/rust from source I sadly can't seem to persuade the liblibc.rlib to statically link with musl libc.a using the appropriate arm target.  But it's mostly the release artifacts I'm interested in getting corrected so I don't even need to build rust from source.  

Pretty sure that the rust (as opposed to libc) CI docker containers don't deal with the non-x86 musl targets which is why I've surmised is the libc one which is the culprit for armv6 musl.

alexcrichton:
Oh sure yeah I can help you there! @gnzlbg is right in that the docker files in this repository exclusively affect this crate's CI, not the binaries we distribute or the rust-lang/rust repository. 

For rust-lang/rust the [`src/ci/docker` folder](https://github.com/rust-lang/rust/tree/master/src/ci/docker) is what you're looking for, specifically the `dist-various-1` container and [probably a line somewhere in here](https://github.com/rust-lang/rust/blob/3499575282b5cda1e98220baae4f6c87e1863926/src/ci/docker/dist-various-1/Dockerfile#L46-L74).

The libc.a built for musl is included later in liblibc.rlib, so changing the original musl build will affect those later artifacts

piersfinlayson:
Great thanks.  I'll leave this PR open, as I think it would be sensible to take the same fix to this repo for consistency, and I'll submit a PR to the rust repo.

alexcrichton:
Ok!

PR #1174, opened 2018-12-12
Add MFD_HUGETLB const for linux

tklauser:


tklauser:
It seems `MFD_HUGETLB` is not available in some (older?) glibc versions. The tests failed on some CI builders on my first push which only ignored it `if !mips && musl => true`.

Sorry, I should have added a comment. I can add one if needed.

alexcrichton:
Ah ok, good to know! It looks like some linux tests at least passed which should help verify its value, but yeah a comment would be great and then I think we can merge!

tklauser:
Done, thanks.

gnzlbg:
If it passed in some build jobs, maybe it could be enabled on at least one of them ?

tklauser:
> If it passed in some build jobs, maybe it could be enabled on at least one of them ?

Is there a way to skip based on libc version of builder type? Or should I just assume on arch (e.g. i686 or x86_64) where it passed consistently on the builders and use that?

gnzlbg:
> Is there a way to skip based on libc version of builder type? 

Not yet, but currently the only version "that matters" is the one in the Docker containers of this crate, so I would just let CI run once, and skip those versions that fail (adding a comment explaining what the failure was for each target, but I suppose that in this case the failure is the same for all of them - too old glibc version without the type/const/...).

tklauser:
Thanks. I now enabled it for `x86_64` and `i686` for which it seems to be available on all targets.

alexcrichton:
👍

PR #1175, opened 2018-12-14
Expose unix ECN-related constants

Ralith:
I'm not certain of the correctness of the BSD definition, but I assume that's what CI's for.

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -1575,6 +1575,8 @@ pub const IP_DROP_MEMBERSHIP: ::c_int = 13;
 pub const IP_PKTINFO: ::c_int = 26;
 pub const IPV6_JOIN_GROUP: ::c_int = 12;
 pub const IPV6_LEAVE_GROUP: ::c_int = 13;
+pub const IPV6_RECVTCLASS: ::c_int = 35;
+pub const IPV6_TCLASS: ::c_int = 36;
 pub const IPV6_PKTINFO: ::c_int = 46;
 pub const IPV6_RECVPKTINFO: ::c_int = 61;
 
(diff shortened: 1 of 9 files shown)


gnzlbg:
@Ralith thank you for the PR! It appears that C failed, could you take a look?

semarie:
> Perhaps it was added recently? https://github.com/openbsd/src/blob/master/sys/netinet/ip.h#L126

oops. It seems my grep pattern on my system was wrong when I quickly
checked. You're right, constants are here (and since 2002).

Sorry about that.

Ralith:
CI failure looks spurious?

gnzlbg:
> but I assume that's what CI's for.

Our CI for the *BSDs is incomplete at best. Do you have links for NetBSD and DragonflyBSD ? Otherwise this looks good to go (I've restarted CI but should be green).

Ralith:
NetBSD has IP_TOS = 3 in [in.h](https://ftp.netbsd.org/pub/NetBSD/NetBSD-current/src/sys/netinet/in.h), a superset of the ECN defines in [ip.h](https://ftp.netbsd.org/pub/NetBSD/NetBSD-current/src/sys/netinet/ip.h), and the TCLASS definitions in [in6.h](https://ftp.netbsd.org/pub/NetBSD/NetBSD-current/src/sys/netinet6/in6.h).

DragonflyBSD has IP_TOS in [in.h](https://github.com/DragonFlyBSD/DragonFlyBSD/blob/master/sys/netinet/in.h), the ECN defines in [ip.h](https://github.com/DragonFlyBSD/DragonFlyBSD/blob/master/sys/netinet/ip.h), and the TCLASS definitions in [in6.h](https://github.com/DragonFlyBSD/DragonFlyBSD/blob/master/sys/netinet6/in6.h) but does not appear to have IP_RECVTOS, so I've moved that down a level.

Ralith:
Looks like I was overly optimistic in placing the `IPTOS_ECN_...` definitions at the top level of `unix`; e.g. solaris (in any form I can find) doesn't supply them. I've narrowed them to notbsd/bsd for now.

Ralith:
What's needed for this to proceed?

alexcrichton:
@bors: r+

Just a notification that it's ready!

PR #1176, opened 2018-12-15
add signal and raise bindings for windows

xmclark:
This PR adds `signal` and `raise` bindings for windows.

I don't know these functions or linux very well, so I leaned on other overrides of signal in the linux bindings, and the [cppreference page](https://en.cppreference.com/w/cpp/header/csignal) for adding the bindings.

I added some constants that were shown on the [microsoft signal page](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=vs-2017) and used the default values I found spelunking through my own copy of `signal.h` that came along with visual studio.

The automated tests pass and my toy apps use the `signal` and `raise` as expected. Let me know if there is anything else I need to do, or any extra tests to write. 

EDIT:
currently working on getting a nice isolated msys2 environment I can use to dev against. 
- [x] msys2 env setup and building

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -76,6 +76,7 @@ fn main() {
         cfg.header("windows.h");
         cfg.header("process.h");
         cfg.header("ws2ipdef.h");
+        cfg.header("signal.h");
 
         if target.contains("gnu") {
             cfg.header("ws2tcpip.h");
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -508,6 +508,7 @@ fn main() {
             n if n.starts_with("P") => true,
             n if n.starts_with("H") => true,
             n if n.starts_with("LP") => true,
+            "__p_sig_fn_t" if mingw => true,
             _ => false,
         }
     });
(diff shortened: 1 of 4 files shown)


xmclark:
Hello @alexcrichton! I'm having a little trouble getting `stable-x86_64-pc-windows-gnu` to build [Appveyor build](https://ci.appveyor.com/project/rust-lang-libs/libc/builds/21050996/job/xovi2a7gk7novnu6). 

I can't explain why the gnu compiler is failing. The pointer arithmetic seems valid. 

I think I could also use a sanity check. Make sure everything looks right. I have never contributed to an FFI library before.

alexcrichton:
Oh I think this is largely just a bug in our testing, you'll need to edit `libc-test/build.rs` to update the `skip_signedness` function to include this typedef as well. It's a typedef to a function pointer type so it doesn't have a signedness

xmclark:
I believe this is ready for review.

I've added the signedness exception for mingw. 

Both windows targets have different names for the `signalhandler_t` type. I added the a alias for both targets.

Issue #1059, opened 2018-08-09
Route netlink C header #define's PR?


jbaublitz:
This is a follow up to #921 which tracks the PR that implemented genetlink constants. I maintain a [netlink library](https://github.com/jbaublitz/neli) but am now implementing the next subsystem in netlink (rtnetlink). I would like to add the constants in the rtnetlink subsystem now that I have a request to provide support for this. If I provide the PR, would you be willing to review and merge it?

alexcrichton:
Definitely! Always willing to have more constants and defines here :)

jbaublitz:
Great! I'll gather all of the rtnetlink constants into a PR when I get a moment.

jbaublitz:
I'm mostly done with this and will be ready to try CI soon. Mainly just making sure this doesn't get closed as I'm actively working on it again. I got pulled away from my side projects for a long while.

gnzlbg:
Glad to hear that!

PR #1179, opened 2018-12-17
Commit for if_addr.h constants for rtnetlink struct flags

jbaublitz:
This PR is meant to add more constants to be used in my netlink library. This is specifically for the rtnetlink subsection of the netlink protocol. It will allow strong typing for [this](https://github.com/jbaublitz/neli/blob/64273782838ad825eb1dc38710b8bbe46fc2b150/src/rtnl.rs#L78) field in the struct I've added which cannot be typed the way other flag fields have been in the library thus far as an enum. See issue #1059 for more information.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -110,6 +110,7 @@ fn main() {
         cfg.header("sys/socket.h");
         if linux && !musl {
             cfg.header("linux/if.h");
+            cfg.header("linux/if_addr.h");
             cfg.header("sys/auxv.h");
         }
         cfg.header("sys/time.h");
--- a/src/unix/notbsd/linux/other/mod.rs
+++ b/src/unix/notbsd/linux/other/mod.rs
@@ -621,6 +621,21 @@ pub const NF_NETDEV_NUMHOOKS: ::c_int = 1;
 pub const NFPROTO_INET: ::c_int = 1;
 pub const NFPROTO_NETDEV: ::c_int = 5;
 
+// linux/if_addr.h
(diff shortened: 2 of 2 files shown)


jbaublitz:
Hey @alexcrichton, it appears that CI is targeting the wrong platform? Let me know if there's something I'm doing wrong. I read over the contributing guidelines and don't think I'm missing anything, but obviously you'd be more familiar here than I would be as it's my first PR to libc. Let me know if there's something I need to change and I'll be happy to fix it. Thanks!

jbaublitz:
I'm finding more constants that I will need. Going to mark this as a work in progress until I find all of the constants I will need so we can get them all added in one review so I don't drag you through multiple review processes.

alexcrichton:
It's looking good so far! FWIW the linux headers and musl targets are a bit weird, I know they generally don't work by default but I forget why...

jbaublitz:
@alexcrichton Some of these flags seem to be glibc specific based on build jobs and what I can tell from documentation. Given that their use is not documented anywhere, I'm going to add the portable ones only in the portable module. Sound good?

alexcrichton:
Yeah that sounds fine by me!

jbaublitz:
Any idea what's going on [here](https://travis-ci.org/rust-lang/libc/jobs/469306461#L1837)? It's the only failure in the CI pipeline now and it seems to be unrelated to my code but consistently failing.

gnzlbg:
That's a spurious failure (sometimes, it happens). I've restarted the build. Let me know if it fails here again.

PR #1180, opened 2018-12-20
add stdio constants for windows

xmclark:
A few useful constants for working with the stdio file descriptors.

I referenced the [msdn](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fileno?view=vs-2017) docs for the numbers, although I think 0,1,2 are standard  on most platforms for the stdio.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -11,6 +11,7 @@ fn main() {
     let x86_64 = target.contains("x86_64");
     let x32 = target.ends_with("gnux32");
     let windows = target.contains("windows");
+    let msvc = target.contains("windows-msvc");
     let mingw = target.contains("windows-gnu");
     let linux = target.contains("unknown-linux");
     let android = target.contains("android");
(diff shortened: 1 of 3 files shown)


alexcrichton:

xmclark:
Hi @alexcrichton. These constants are useful for both gnu and msvc targets. MSVC does not actually seem to define these constants anywhere, so I manually set them in the `build.rs` file for msvc target.

gnzlbg:
@xmclark I think that's fine - these constants are defined in POSIX so they are valid on msvc targets even though windows headers do not define them. But also see https://stackoverflow.com/a/27583123 .

alexcrichton:
The goal of libc is to bind what the platform provides, so if MSVC doesn't provide these constants then they shouldn't be defined in libc and they should only be defined on MinGW

alexcrichton:
It's true it's convenient! It's just something that happens in downstream crates though, not `libc`

xmclark:
I have reverted my previous commits and move the constants into the windows gnu target. Thank you for feedback!

PR #1181, opened 2018-12-23
Add statx

### LINKED ISSUE #1178: Add statx linux system call so we can implement fs::Metadata::created on Linux
Right now, `std::fs::Metadata::created` returns an `Err` on Linux. But the creation time is available with the `statx` system call.

https://unix.stackexchange.com/a/407305/289105

ariasuni:
Fix #1178

According to [this website](https://fedora.juszkiewicz.com.pl/syscalls.html), syscall number for `statx` on `arm64` should be 291 but it’s already the value of `SYS_syscalls`, is it an error?

https://github.com/rust-lang/libc/blob/c1ae35dad55ce8c9226979ba12340409705e5bc4/src/unix/notbsd/linux/other/b64/aarch64.rs#L821


ariasuni:
I don’t understand from the log why some of the Travis builds failed, especially since one of them is the exact platform I use and did the test on (Rust stable, `x86_64-unknown-linux-gnu`).

Note that I still got, before and after my changes, some errors (which might be related to a more recent than average glibc?):

```
bad ucontext_t size: rust: 936 (0x3a8) != c 968 (0x3c8)
bad NFT_MSG_MAX value at byte 0: rust: 22 (0x16) != c 25 (0x19)
```

gnzlbg:
> especially since one of them is the exact platform I use and did the test on

You are probably just running the tests on a Linux version that differs from the one in `libc`'s Docker container.

ariasuni:
@gnzlbg do I need to change something, and if yes, what?

gnzlbg:
@ariasuni the "problem" is that the linux version on your machine, and the linux version that `libc` "supports" / "targets" are slightly different, which is why we see these errors in CI, but you don't see them in your machine. 

The error message is not 100% clear, but it basically tells you that:

* `bad NFT_MSG_MAX value at byte 0: rust: 22 (0x16) != c 25 (0x19)`, that is, the value of the first byte of `NFT_MSG_MAX` is `0x16` in `libc`, but should be `0x19` for CI. So that needs to be changed.

* `bad ucontext_t size: rust: 936 (0x3a8) != c 968 (0x3c8)`, that is, the size of `ucontext_t` differs between the one in `libc` and the one in CI. This can be due to many factors, you might want to check the linux headers of the version of linux used in libc's ci in ci/docker/x86_64-unknown-linux-gnu, and check whether that type is missing some fields, or whether it has some C attributes that might affect its alignment, field padding, etc.

ariasuni:
Note that the errors I get locally are very different from the ones on the CI, and I should probably not change things that are working on CI and not related to my changes.

gnzlbg:
Any progress on this?

gnzlbg:
> According to this website, syscall number for statx on arm64 should be 291 but it’s already the value of SYS_syscalls, is it an error?

It is often best to just check out the source of the kernel itself. While there are "generic" syscall numbers, each architecture often has different ones. For `arm64`, this file (https://github.com/torvalds/linux/blob/f1c2f8857c5aa6c92aa903bc06437503422e5925/arch/arm64/include/asm/unistd32.h#L818) shows that:

```c
#define __NR_statx 397
```

the syscall number for `statx` might be `397` (i'm not very familiar with the kernel). 

The easiest way to be sure is to just write a small program that performs the syscall and tests whether it behaves correctly.

gnzlbg:
I'm closing this due to lack of inactivity, please feel free to reopen if you want to keep working on this.

PR #1182, opened 2018-12-23
add some socket functions and a SOCKET type for windows

xmclark:
This PR adds a few of the socket functions for windows. Some targets use the `stdcall` calling convention for these functions, so I put them in an `extern "system"` block ([see the nomicon](https://doc.rust-lang.org/nomicon/ffi.html#foreign-calling-conventions)).

### DIFF
--- a/src/windows/mod.rs
+++ b/src/windows/mod.rs
@@ -50,6 +50,8 @@ pub type ino_t = u16;
 pub enum timezone {}
 pub type time64_t = i64;
 
+pub type SOCKET = uintptr_t;
+
 s! {
     // note this is the struct called stat64 in Windows. Not stat, nor stati64.
     pub struct stat {
@@ -92,6 +94,11 @@ s! {
     pub struct timespec {
         pub tv_sec: time_t,
         pub tv_nsec: c_long,
+    }
+
+    pub struct sockaddr {
+        pub sa_family: c_ushort,
+        pub sa_data: [c_char; 14],
     }
 }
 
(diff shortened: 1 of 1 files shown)


xmclark:
@gnzlbg I would like a review of this code! This is the first time I have ever had to think about calling convention. Feedback welcome. 

It's unclear to me why this [nightly linux build](https://travis-ci.org/rust-lang/libc/jobs/471678077) is failing. There is another PR that is failing for what seems to be [similar reasons](https://travis-ci.org/rust-lang/libc/jobs/471512546).

alexcrichton:
@bors: r+

Thanks!

gnzlbg:
FWIW the failures are spurious, i've reported them to the docs team yesterday, but with christmas going on, progress is slow. I'll try to review this by the end of the week.

xmclark:
@gnzlbg no worries. Happy holidays!

alexcrichton:
@bors: retry

Different spurious failure this time though!

retep998:
Weren't these removed a long time ago from `libc` due to them being not part of the CRT and just normal windows system library stuff?

retep998:
These definitions are *not* part of the UCRT though!

retep998:
Well, this is even more awkward now that 0.2.46 was published with this change.

@alexcrichton What is the official word on this? Is `libc` expanding in scope to cover things that are not in the CRT, or was this just an oversight that warrants a yank + revert + publish?

xmclark:
Thanks @retep998 for pointing out. I misunderstood that this was a requirement (my bad! 😢 ). 

 RFC 1291 says this about libc on windows:
> This library is currently intended to be distinct from the winapi crate as well as bindings to common system DLLs found on Windows, so the current scope of libc will be pared back to just what the CRT contains.

alexcrichton:
We can probably just avoid new bindings in the future, this isn't really worth a revert I think.

PR #1186, opened 2018-12-25
Define missing F_RDLCK, F_WRLCK and F_UNLCK constants on linux-mips -and s390x

glaubitz:
These are defined in the fcntl.h glibc header on MIPS and s390x
systems on Linux but missing in the libc crate, so add them as they
are required for the file locking API in rustc.


alexcrichton:
Sure thing, release instructions are in the [README](https://github.com/rust-lang/libc#releasing-your-change-to-cratesio)!

glaubitz:
Ah, thanks for the pointer. Never done this before :).

Issue #1080, opened 2018-09-13
Test for `NFT_MSG_MAX` fails on my Linux machine


tbu-:
```
RUNNING ALL TESTS
bad ucontext_t size: rust: 936 (0x3a8) != c 968 (0x3c8)
bad NFT_MSG_MAX value at byte 0: rust: 22 (0x16) != c 25 (0x19)
```

glibc 0.28
Linux 4.18.6.arch1
Arch Linux

`NFT_MSG_MAX`, from /usr/include/linux/netfilter/nf_tables.h:
```
enum nf_tables_msg_types {
        NFT_MSG_NEWTABLE,
        NFT_MSG_GETTABLE,
        NFT_MSG_DELTABLE,
        NFT_MSG_NEWCHAIN,
        NFT_MSG_GETCHAIN,
        NFT_MSG_DELCHAIN,
        NFT_MSG_NEWRULE,
        NFT_MSG_GETRULE,
        NFT_MSG_DELRULE,
        NFT_MSG_NEWSET,
        NFT_MSG_GETSET,
        NFT_MSG_DELSET,
        NFT_MSG_NEWSETELEM,
        NFT_MSG_GETSETELEM,
        NFT_MSG_DELSETELEM,
        NFT_MSG_NEWGEN,
        NFT_MSG_GETGEN,
        NFT_MSG_TRACE,
        NFT_MSG_NEWOBJ,
        NFT_MSG_GETOBJ,
        NFT_MSG_DELOBJ,
        NFT_MSG_GETOBJ_RESET,
        NFT_MSG_NEWFLOWTABLE,
        NFT_MSG_GETFLOWTABLE,
        NFT_MSG_DELFLOWTABLE,
        NFT_MSG_MAX,
};
```

It seems like
```
        NFT_MSG_NEWFLOWTABLE,
        NFT_MSG_GETFLOWTABLE,
        NFT_MSG_DELFLOWTABLE,
```
were added and thus `NFT_MSG_MAX` increased. Linux doesn't consider this a breaking change. I'd assume that more messages might be added in the future, so just adjusting the constant in libc as a breaking change doesn't seem to be useful.

gnzlbg:
This is expected, the values of constants changes across different kernel versions. Sadly, the different in sizes here will probably lead to undefined behavior.

pusateri:
Same issue on Ubuntu 18.10 cosmic
```
$ uname -a
[...]

PR #1187, opened 2018-12-25
Adjust for newer glibc on sparc64

karcherm:
This commit removes special cases for sparc64 which caused test failures on a sparc64
system (Debian, glibc 2.28). With these changes, the ABI tests pass except for the
known test failure from #1080

### DIFF
--- a/src/unix/notbsd/linux/other/mod.rs
+++ b/src/unix/notbsd/linux/other/mod.rs
@@ -40,23 +40,19 @@ s! {
         pub ut_exit: __exit_status,
 
         #[cfg(any(target_arch = "aarch64",
-                  target_arch = "sparc64",
                   all(target_pointer_width = "32",
                       not(target_arch = "x86_64"))))]
         pub ut_session: ::c_long,
         #[cfg(any(target_arch = "aarch64",
-                  target_arch = "sparc64",
                   all(target_pointer_width = "32",
                       not(target_arch = "x86_64"))))]
         pub ut_tv: ::timeval,
 
         #[cfg(not(any(target_arch = "aarch64",
-                      target_arch = "sparc64",
                       all(target_pointer_width = "32",
                           not(target_arch = "x86_64")))))]
         pub ut_session: ::int32_t,
         #[cfg(not(any(target_arch = "aarch64",
-                      target_arch = "sparc64",
                       all(target_pointer_width = "32",
                           not(target_arch = "x86_64")))))]
         pub ut_tv: __timeval,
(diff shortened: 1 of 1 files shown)


karcherm:
The CI failure is most likely due to an older libc version (Debian libc6-dev-sparc64-cross 0.24.x) compared to the host I created this patch for (Debian libc6-dev 0.28-3). How do we decide the target libc version for the libc crate?
As long as the reference versions on sparc64 stay unchanged, this PR is unneccessary. I made it because `cargo test` failed on the sparc64 system I use to create a different PR that is not yet ready.

glaubitz:
Do we need this commit to fix any failures on Debian sparc64 unstable with glibc 2.28?

karcherm:
Yes, as I said, it fixes all test failures (compiler warning that purposedfully gets escalated to an error for a member of utmpx and several outdated constants for netfilter stuff) that currently occur on Debian sparc64 unstable, except for one netfilter-related test failure that already has its own issue: #1080.

karcherm:
Well, this is about libc ABI tests, *not* about rustc tests. AFAIK libc ABI tests are not part of `./x.py test`. It does not fix any of them.

gnzlbg:
> How do we decide the target libc version for the libc crate?

Historically we have chosen it "as old as possible" to be as compatible with as many libc versions as possible, but we have slowly moved to "as new as possible" because some things like `const`s (C enums, macro values, etc.) do change across libc versions in some OSes often. 

If you need a newer libc version here, just update it as part of this PR.

alexcrichton:
Agreed with @gnzlbg we can just update libc in this repo's CI (which is installed [here](https://github.com/rust-lang/libc/blob/master/ci/linux-sparc64.sh)) and that should do the trick!

gnzlbg:
Any progress on this?

glaubitz:
@gnzlbg I'll take care of this. I just recently built fresh installation images for Debian sparc64. I'm actually the one who is responsible for the installation images in Debian Ports :). We can update the ones for ```powerpc``` and ```ppc64``` as well, if necessary.

gnzlbg:
@karcherm could you check if these issues persist with libc master branch ?

glaubitz:
@gnzlbg I think what needs to be done first is switching the CI to use a newer Debian sparc64 image from https://cdimage.debian.org/cdimage/ports/. I'm the person at Debian responsible for creating these images, so it's probably best when I send the PR for that to avoid pointing to links that might get moved.

gnzlbg:
@glaubitz I did that as part of #1337 (IIRC I updated to the latest image available there).

glaubitz:
@gnzlbg Okay, the 10.0 image you are using is reasonably new enough, I will overwrite it at some point with a new 10.0 one when Debian Buster is officially released.

Then @karcherm should be able to rebase and push his changes which require the CI to use the newer image.

gnzlbg:
> Okay, the 10.0 image you are using is reasonably new enough, I will overwrite it at some point with a new 10.0 one when Debian Buster is officially released.

That would be awesome.

gnzlbg:
This needs rebasing

glaubitz:
I'll take care of that.

glaubitz:
I just had a look at the patch and it looks like most of @karcherm 's changes are already in.

The only thing where the current code disagrees with @karcherm 's code is ```NFT_MSG_MAX``` which @karcherm has explicitly set to **22** while the current code uses **25**.

gnzlbg:
The 25 is verified against the debian version used in CI, which is technically the only one that we fully-support. It should be forward compatible with newer kernel versions, but not with older versions.

gnzlbg:
FYI I'm closing this PR, it appears that most changes are in. Feel free to reopen or submit a new PR for newer changes.

glaubitz:
> The 25 is verified against the debian version used in CI, which is technically the only one that we fully-support. It should be forward compatible with newer kernel versions, but not with older versions.

Okay, makes sense then. Thanks for having taken care of this.

PR #1188, opened 2018-12-28
NetBSD 8 adds new address family changing AF_MAX and new ifaddrs flags.

pusateri:
libc-test on NetBSD 8.0  fails without these changes.
ifa_addrflags was added to struct ifaddrs to support AF_CAN.
AF_MAX got bumped up one by AF_CAN.
RT_IFLIST format changed to support ifa_addrflags.

### DIFF
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -57,7 +57,9 @@ s! {
         pub ifa_addr: *mut ::sockaddr,
         pub ifa_netmask: *mut ::sockaddr,
         pub ifa_dstaddr: *mut ::sockaddr,
-        pub ifa_data: *mut ::c_void
+        pub ifa_data: *mut ::c_void,
+        #[cfg(target_os = "netbsd")]
+        pub ifa_addrflags: ::c_uint
     }
 
     pub struct fd_set {
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -523,15 +523,16 @@ pub const AF_BLUETOOTH: ::c_int = 31;
 pub const AF_IEEE80211: ::c_int = 32;
 pub const AF_MPLS: ::c_int = 33;
 pub const AF_ROUTE: ::c_int = 34;
-pub const AF_MAX: ::c_int = 35;
+pub const AF_MAX: ::c_int = 36;
 
 pub const NET_MAXID: ::c_int = AF_MAX;
 pub const NET_RT_DUMP: ::c_int = 1;
 pub const NET_RT_FLAGS: ::c_int = 2;
-pub const NET_RT_OOIFLIST: ::c_int = 3;
-pub const NET_RT_OIFLIST: ::c_int = 4;
-pub const NET_RT_IFLIST: ::c_int = 5;
-pub const NET_RT_MAXID: ::c_int = 6;
+pub const NET_RT_OOOIFLIST: ::c_int = 3;
+pub const NET_RT_OOIFLIST: ::c_int = 4;
+pub const NET_RT_OIFLIST: ::c_int = 5;
+pub const NET_RT_IFLIST: ::c_int = 6;
+pub const NET_RT_MAXID: ::c_int = 7;
 
 pub const PF_OROUTE: ::c_int = AF_OROUTE;
 pub const PF_ARP: ::c_int = AF_ARP;


semarie:
the NetBSD changes is an interesting breaking changes. the commit is https://github.com/NetBSD/src/commit/6f66312e8fef4fd00df5571fa67c3a8cfa479ff1#diff-b3f45d20f63d013e46f59f3f88a84d3dL477

for explaining a bit, NetBSD changed the name of constants, for example `NET_RT_OOIFLIST` (old-old-iflist) was renamed to `NET_RT_OOOIFLIST` (old-old-old-iflist), but keeping the binary compatibility (an old compiled binary using `NET_RT_OOIFLIST` (`3` as value), still work the same way on differents systems versions.

Regarding the proposed PR, I dunno how it will be resolved. Commited as it, it will break NetBSD (pre 8) as `NET_RT_OOIFLIST` doesn't mean the same thing on pre-8 and post-8. And if not commited, it will break NetBSD (post 8).

@alexcrichton @gnzlbg something has to be done. The minimal thing would be policy on what is the supported version for *each* OS supported by libc ("last stable version" would be a good start for example). It would at least permit to know which change should be considered and which one is not.

More advanced stuff would be https://github.com/rust-lang/rfcs/pull/2048 as it would permit supporting several OS versions (but the rfc stands for defining supported versions also).

gnzlbg:
Does NetBSD still expose symbols of the APIs that were using the old version of `ifaddrs` ? If so, we could add an `ifaddrs2` type that includes these changes, and `_2` version of the APIs using `ifaddrs2`, and use `#[link_name]` to select what gets linked.

pusateri:
No, NetBSD 8.0 does not expose an older version of struct ifaddrs.

gnzlbg:
> No, NetBSD 8.0 does not expose an older version of struct ifaddrs

This is not exactly what I asked. I asked whether NetBSD 8.0 system libraries contain older versions of symbols that use the old `ifaddrs` struct on their ABI (note that if these are exposed they won't show up on any headers).

pusateri:
I copied the NetBSD 7.2 binary for /sbin/route to NetBSD 8.0 and ran 'route monitor' which should print routing socket ifaddr messages when interfaces go up/down or add/delete addresses. After unplugging the ethernet interface, it didn't print the same route info under NetBSD 8.0 that it did under 7.2 and then crashed when I plugged the ethernet cable back in which worked fine in 7.2. It appears that 8.0 is not binary backward compatible with 7.2.

semarie:
@jakllsch @ryoon as you seems involved in rust on netbsd side, I think having feedback from you will be appreciate: some netbsd version could be not longer supported by the Rust libc (or more complex to support in pkgsrc).

if you are not the right netbsd developers for this question, sorry about that, but please ping someone else.

pusateri:
I just tried running `cargo test` on NetBSD 7.2 and it fails there too.
/usr/include/netinet/dccp.h is missing (It was deleted before 7.1 release) which means 7.1 will probably fail too.
```
   Compiling libc-test v0.1.0 (/home/pusateri/Projects/libc/libc-test)
error: failed to run custom build command for `libc-test v0.1.0 (/home/pusateri/Projects/libc/libc-test)`
process didn't exit successfully: `/home/pusateri/Projects/libc/target/debug/build/libc-test-aa7af2bb6c9032d2/build-script-build` (exit code: 101)
--- stdout
-----------------------------------------
cargo:rerun-if-changed=../src/lib.rs
cargo:rerun-if-changed=../src/macros.rs
cargo:rerun-if-changed=../src/dox.rs
cargo:rerun-if-changed=../src/unix/mod.rs
cargo:rerun-if-changed=../src/unix/bsd/mod.rs
cargo:rerun-if-changed=../src/unix/bsd/netbsdlike/mod.rs
cargo:rerun-if-changed=../src/unix/bsd/netbsdlike/netbsd/mod.rs
cargo:rerun-if-changed=../src/unix/bsd/netbsdlike/netbsd/x86_64.rs
OPT_LEVEL = Some("0")
HOST = Some("x86_64-unknown-netbsd")
CC_x86_64-unknown-netbsd = None
CC_x86_64_unknown_netbsd = None
HOST_CC = None
CC = None
CFLAGS_x86_64-unknown-netbsd = None
CFLAGS_x86_64_unknown_netbsd = None
HOST_CFLAGS = None
CFLAGS = None
DEBUG = Some("true")
running: "cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-Wall" "-Wextra" "-Wall" "-Wextra" "-Werror" "-Wno-unused-parameter" "-Wno-type-limits" "-Wno-address-of-packed-member" "-Wno-deprecated-declarations" "-Wno-deprecated-declarations" "-D_NETBSD_SOURCE=1" "-o" "/home/pusateri/Projects/libc/target/debug/build/libc-test-beca0575d5180d93/out/main.o" "-c" "/home/pusateri/Projects/libc/target/debug/build/libc-test-beca0575d5180d93/out/main.c"
cargo:warning=/home/pusateri/Projects/libc/target/debug/build/libc-test-beca0575d5180d93/out/main.c:67:26: fatal error: netinet/dccp.h: No such file or directory
cargo:warning= #include <netinet/dccp.h>
cargo:warning=                          ^
cargo:warning=compilation terminated.
exit code: 1

--- stderr
error: attributes on struct pattern or literal fields are unstable (see issue #38814)
   --> ../src/unix/bsd/netbsdlike/netbsd/mod.rs:711:5
    |
711 | /     #[cfg(any(target_arch = "sparc", target_arch = "sparc64",
712 | |               target_arch = "x86", target_arch = "x86_64"))]
    | |____________________________________________________________^
    | 
   ::: ../src/lib.rs
    |
177 | / cfg_if! {
178 | |     if #[cfg(windows)] {
179 | |         mod windows;
180 | |         pub use windows::*;
...   |
201 | |     }
202 | | }
    |
    = help: add #![feature(struct_field_attributes)] to the crate attributes to enable

error: attributes on struct pattern or literal fields are unstable (see issue #38814)
   --> ../src/unix/bsd/netbsdlike/netbsd/mod.rs:715:5
    |
715 | /     #[cfg(any(target_arch = "sparc", target_arch = "sparc64",
716 | |               target_arch = "x86", target_arch = "x86_64"))]
    | |____________________________________________________________^
    | 
   ::: ../src/lib.rs
    |
177 | / cfg_if! {
178 | |     if #[cfg(windows)] {
179 | |         mod windows;
180 | |         pub use windows::*;
...   |
201 | |     }
202 | | }
    |
    = help: add #![feature(struct_field_attributes)] to the crate attributes to enable

thread 'main' panicked at '

Internal error occurred: Command "cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-Wall" "-Wextra" "-Wall" "-Wextra" "-Werror" "-Wno-unused-parameter" "-Wno-type-limits" "-Wno-address-of-packed-member" "-Wno-deprecated-declarations" "-Wno-deprecated-declarations" "-D_NETBSD_SOURCE=1" "-o" "/home/pusateri/Projects/libc/target/debug/build/libc-test-beca0575d5180d93/out/main.o" "-c" "/home/pusateri/Projects/libc/target/debug/build/libc-test-beca0575d5180d93/out/main.c" with args "cc" did not execute successfully (status code exit code: 1).

', /home/pusateri/.cargo/registry/src/github.com-1ecc6299db9ec823/cc-1.0.25/src/lib.rs:2260:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

semarie:
> ABI compatibility breaks seems to basically never happen on the "tier 1" platforms 

should I point to you the MSVC 2015 entry present in the RFC 2048 ? breaking changes already occurs on Windows in the past. The RFC pointed gets() function removal as example.

retep998:
The C standard library on Windows absolutely has made breaking ABI changes in the past, hence `msvcr100.dll` `msvcr110.dll` `msvcr120.dll` `vcruntime140.dll` and so on. Of course `libc` doesn't actually expose that much on Windows, nor is it used that much, so `libc` would be unlikely to notice those breaking changes.

inferiorhumanorgans:
IMO it would be worth defining a support matrix and including the ABI version with the host triplet here especially as there are similar issues in FreeBSD and DragonFlyBSD.

inferiorhumanorgans:
FWIW:

NetBSD 7.0: syn failed to compile
NetBSD 7.1: internal error w/ gcc
NetBSD 7.2: working on it

gnzlbg:
@bors: r+

---

Feel free to submit a PR adding information about the supported NetBSD version to the readme. Also, if any of you is involved in NetBSD development, it might be interesting to discuss how we could avoid breakage in the future.

Some simple "C guidelines" like:

* only extending C enums by adding new values at their end
* when the layout of a type changes (e.g. new fields added), keep the old API in the library binary so that we can link to that from here (e.g. by using a special link name), and so that we can add "versioned" version of the types (e.g. foo -> foo2, api_foo -> api_foo2) that call the different symbols

PR #1195, opened 2019-01-02
Bump version to 0.2.46

glaubitz:
New release required for https://github.com/rust-lang/rust/pull/57121


glaubitz:
@alexcrichton Looks like we're still missing a tag. Not sure whether I can add the tag through a PR.

alexcrichton:
Just a lock file update in the rustc repo and then it should be good to go!

PR #1200, opened 2019-01-08
add FIO* for FreeBSD and NetBSD

alecmocatta:
Calculated [here](https://repl.it/repls/PaltryTroubledMonitors) from FreeBSD [source](https://github.com/freebsd/freebsd/blob/9867e6d8fe1cac34121e5f2df9357363f5659617/sys/sys/filio.h).

Calculated [here](https://repl.it/repls/ProfitableImpracticalComputeranimation) from NetBSD [source](https://github.com/NetBSD/src/blob/64b8a48e1288eb3902ed73113d157af50b2ec596/sys/sys/filio.h).

Some of these are likely identical across all BSDs – if this can be confirmed they could be elevated to `unix/bsd/mod.rs` alongside `FIOCLEX` and `FIONBIO`.

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -447,6 +447,19 @@ pub const TIOCSIG: ::c_uint = 0x2004745f;
 pub const TIOCM_DCD: ::c_int = 0x40;
 pub const H4DISC: ::c_int = 0x7;
 
+pub const FIONCLEX: ::c_ulong = 0x20006602;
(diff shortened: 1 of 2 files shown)


alecmocatta:
@alexcrichton Is something up with homu? I would expect it to have tested and/or merged by now?

alexcrichton:
I've kicked a build to try to make progress

alecmocatta:
Failure is due to https://github.com/rust-lang/rust/issues/57488. @alexcrichton If you're able to retry when that's resolved, thanks!

inferiorhumanorgans:
This appears to break the build on NetBSD 8.0.

PR #1202, opened 2019-01-09
glibc x86: add the register offset constants for mcontext_t.gregs

acfoltzer:
Constants calculated from the [glibc source](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/x86/sys/ucontext.h;h=3d883c2bf1220ef50cedc6b1491d699c0727c508;hb=HEAD). These are needed when using `mcontext_t`. Thanks!

### DIFF
--- a/src/unix/notbsd/linux/other/b32/x86.rs
+++ b/src/unix/notbsd/linux/other/b32/x86.rs
@@ -758,6 +758,27 @@ pub const EFL: ::c_int = 14;
 pub const UESP: ::c_int = 15;
 pub const SS: ::c_int = 16;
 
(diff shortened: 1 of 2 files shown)


acfoltzer:
A design question I had for this PR is: the constants are defined from a C enum, so they should be `c_int`, however they are only used to index into the `gregs` array, so they always have to be cast to `usize`. Should they be defined in this library as `usize` to be more convenient, or is it more important to have the correct types from the C side?

acfoltzer:
These failures appear to be something going wrong with the rustup step. Is there a way to tell bors to re-test?

PR #1211, opened 2019-01-15
Fix linking aio_read(3) and friends on NetBSD

asomers:
The aio functions require librt on NetBSD.


asomers:
This is the 2nd time today that a Cirrus test has mysteriously aborted.  Please retry the merge.

PR #1212, opened 2019-01-15
Fix CMSG_DATA(3) and friends on BSD

### LINKED ISSUE #1210: Incorrect CMSG functions on BSD
PR #1098 introduced Rust functions that duplicate the `CMSG_FIRSTHDR`, `CMSG_NXTHDR`, `CMSG_DATA`, `CMSG_SPACE`, and `CMSG_LEN` C macros.  However, several of these are incorrect on BSD.  It looks like the author wrote the Linux versions first, and then simply copied them into src/unix/bsd/mod.rs.  For example, on both FreeBSD and OpenBSD `CMSG_DATA` is defined as
```rust
    pub fn CMSG_DATA(cmsg: *const cmsghdr) -> *mut ::c_uchar {
        cmsg.offset(1) as *mut ::c_uchar
    }
```
even though the actual C macro is
```c
#define CMSG_DATA(cmsg) \
        ((unsigned char *)(cmsg) + _ALIGN(sizeof(struct cmsghdr)))
```

asomers:
PR #1098 added the CMSG_DATA(3) family of functions into libc.  Because
they're defined as macros in C, they had to be rewritten as Rust
functions for libc.  Also, they can't be tested in CI for the same
reason.  But that PR erroneously used the same definitions in BSD as in
Linux.

This commit corrects the definitions for OSX, FreeBSD, DragonflyBSD,
OpenBSD, and NetBSD.  I renamed a few variables and collapsed a few
macros in order to combine the definitions where possible.

Fixes #1210

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -1,6 +1,7 @@
 //! Apple (ios/darwin)-specific definitions
 //!
 //! This covers *-apple-* triples currently
+use dox::mem;
 
 pub type c_char = i8;
 pub type clock_t = c_ulong;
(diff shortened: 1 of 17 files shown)


asomers:
One question.  The `CMSG_DATA`, `CMSG_LEN`, and `CMSG_SPACE` functions are always safe to use (the other two are not).  But the `f!` macro marks them as `unsafe extern`, even though they are neither.  Should I bypass the macro to declare them as safe Rust functions?  I hesitate because I don't see any precedent for doing that, even for other obviously safe functions like `WEXITSTATUS`.

alexcrichton:
@bors: r+

Thanks!

I think we covered the safety in another issue where we're hoping for consistency here, but we'll see how that pays off over time!

Issue #1208, opened 2019-01-14
setjmp/longjmp


jeff-davis:
I would like support for setjmp/longjmp (and friends sigsetjmp/siglongjmp). The use case is as follows:

I am trying to integrate with postgres (written in C), which means calling into postgres functions. The postgres exception mechanism uses sigsetjmp/siglongjmp.

The idea is, right before calling a postgres function via FFI call, I would do what postgres does in its PG_TRY() / PG_CATCH() macros, which involves sigsetjmp() and some local variable manipulation, and updating a global pointer. That would allow me to catch the siglongjmp() before it bypasses any rust frames, turn it into a rust panic, and properly unwind.

Later, before returning to postgres, I could catch this panic unwind in the highest rust frame, and turn it back into a siglongjmp(). I could be careful that all non-trivial variables have gone out of scope before the siglongjmp() takes place, so that no important destructors are skipped.

If this is a reasonable thing to do, I'd be willing to try to implement it for the libc crate and offer a patch, though I may need some help getting the platform stuff right (because the jmp_buf struct size is platform-dependent).

If it's not a reasonable thing to do, please let me know.

alexcrichton:
Seems reasonable to me to add!

tgross35:
Based on discussion at https://github.com/rust-lang/libc/pull/1216, it sounds like the correct thing to do would be to add intrinsics rather than binding the functions in `libc`. And that if we were to add these to `libc`, we would first need an attribute to indicate that a function may return twice.

There is also an open RFC issue at https://github.com/rust-lang/rfcs/issues/2625 and I think some resolution there is necessary in order to do anything here, so I am just going to close this for now.

PR #1216, opened 2019-01-18
Add support for setjmp/longjmp on x86_64.

jeff-davis:
Fix issue #1208

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -58,6 +58,7 @@ fn main() {
         .header("fcntl.h")
         .header("limits.h")
         .header("locale.h")
+        .header("setjmp.h")
         .header("stddef.h")
         .header("stdint.h")
         .header("stdio.h")
@@ -920,11 +921,7 @@ fn main() {
     });
 
     cfg.fn_cname(move |name, cname| {
-        if windows {
-            cname.unwrap_or(name).to_string()
-        } else {
-            name.to_string()
-        }
+        cname.unwrap_or(name).to_string()
     });
 
     cfg.generate("../src/lib.rs", "main.rs");
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -920,8 +920,24 @@ fn main() {
                                            field == "ssi_arch"))
     });
 
-    cfg.fn_cname(move |name, cname| {
-        cname.unwrap_or(name).to_string()
+    cfg.fn_cname(move |name, link_name_opt| {
+        match link_name_opt {
+            Some(link_name) => {
+                let mut use_link_name = true;
+                for c in link_name.chars() {
+                    if !c.is_ascii_alphanumeric() && c != '_' {
+                        use_link_name = false;
+                        break;
+                    }
+                }
+                if use_link_name {
+                    link_name.to_string()
+                } else {
+                    name.to_string()
+                }
+            },
+            None => name.to_string(),
+        }
     });
 
     cfg.generate("../src/lib.rs", "main.rs");
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -657,19 +657,16 @@ s! {
     }
 
     #[repr(C)]
-    pub struct jmp_buf {
+    pub struct __jmp_buf_tag {
         __jmpbuf: [i64; 8],
         __mask_was_saved: ::c_int,
         __saved_mask: ::sigset_t,
     }
 
-    #[repr(C)]
-    pub struct sigjmp_buf {
-        __jmpbuf: [i64; 8],
-        __mask_was_saved: ::c_int,
-        __saved_mask: ::sigset_t,
-    }
 }
+
+pub type jmp_buf = [__jmp_buf_tag; 1];
+pub type sigjmp_buf = [__jmp_buf_tag; 1];
 
 pub const ABDAY_1: ::nl_item = 0x20000;
 pub const ABDAY_2: ::nl_item = 0x20001;
@@ -2274,11 +2271,11 @@ extern {
         stream: *mut ::FILE
     ) -> ::size_t;
 
-    pub fn setjmp(env: *mut ::jmp_buf) -> ::c_int;
+    pub fn setjmp(env: ::jmp_buf) -> ::c_int;
     #[link_name="__sigsetjmp"]
-    pub fn sigsetjmp(env: *mut ::sigjmp_buf, savesigs: ::c_int) -> ::c_int;
-    pub fn longjmp(env: *mut ::jmp_buf, val: ::c_int);
-    pub fn siglongjmp(env: *mut ::sigjmp_buf, val: ::c_int);
+    pub fn sigsetjmp(env: ::sigjmp_buf, savesigs: ::c_int) -> ::c_int;
+    pub fn longjmp(env: ::jmp_buf, val: ::c_int);
+    pub fn siglongjmp(env: ::sigjmp_buf, val: ::c_int);
 }
 
 cfg_if! {
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -2278,7 +2278,7 @@ extern {
     ) -> ::size_t;
 
     pub fn setjmp(env: ::jmp_buf) -> ::c_int;
-    #[link_name="__sigsetjmp"]
+    #[cfg_attr(target_env="gnu", link_name="__sigsetjmp")]
     pub fn sigsetjmp(env: ::sigjmp_buf, savesigs: ::c_int) -> ::c_int;
     pub fn longjmp(env: ::jmp_buf, val: ::c_int);
     pub fn siglongjmp(env: ::sigjmp_buf, val: ::c_int);
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -2280,8 +2280,6 @@ extern {
     pub fn setjmp(env: ::jmp_buf) -> ::c_int;
     #[cfg_attr(target_env="gnu", link_name="__sigsetjmp")]
     pub fn sigsetjmp(env: ::sigjmp_buf, savesigs: ::c_int) -> ::c_int;
-    pub fn longjmp(env: ::jmp_buf, val: ::c_int);
-    pub fn siglongjmp(env: ::sigjmp_buf, val: ::c_int);
 }
 
 cfg_if! {
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -38,6 +38,9 @@ pub type Elf64_Sxword = i64;
 pub type Elf32_Section = u16;
 pub type Elf64_Section = u16;
 
+pub type jmp_buf = [__jmp_buf_tag; 1];
+pub type sigjmp_buf = [__jmp_buf_tag; 1];
+
 pub enum fpos64_t {} // TODO: fill this out with a struct
 
 s! {
@@ -664,9 +667,6 @@ s! {
     }
 
 }
-
-pub type jmp_buf = [__jmp_buf_tag; 1];
-pub type sigjmp_buf = [__jmp_buf_tag; 1];
 
 pub const ABDAY_1: ::nl_item = 0x20000;
 pub const ABDAY_2: ::nl_item = 0x20001;
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -655,6 +655,20 @@ s! {
         pub updated: ::c_ulong,
         pub ha: [::c_uchar; ::MAX_ADDR_LEN],
     }
+
+    #[repr(C)]
+    pub struct jmp_buf {
+        __jmpbuf: [i64; 8],
+        __mask_was_saved: ::c_int,
+        __saved_mask: ::sigset_t,
+    }
+
+    #[repr(C)]
+    pub struct sigjmp_buf {
+        __jmpbuf: [i64; 8],
+        __mask_was_saved: ::c_int,
+        __saved_mask: ::sigset_t,
+    }
 }
 
 pub const ABDAY_1: ::nl_item = 0x20000;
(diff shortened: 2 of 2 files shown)


newpavlov:
IIUC they will be mostly used for FFI, and other uses should be heavily discouraged.

gnzlbg:
@newpavlov independently of what their intended usage is, i am wondering whether they can be used correctly at all. E.g. `longjmp` allows you to return from multiple functions without actually going through a return statement (by changing the program counter, stack pointer, etc.). Is that allowed by LLVM ?

jeff-davis:
Seems easy enough to answer the LLVM question: I wrote a simple C program using setjmp/longjmp and compiled it with:

clang-6.0 -emit-llvm -S setjmp.c -o setjmp.S

The resulting LLVM has:

```
  ...
  declare i32 @_setjmp(%struct.__jmp_buf_tag*) #1
  ...
  %3 = call i32 @_setjmp(%struct.__jmp_buf_tag* %2) #4
  ...
  declare void @longjmp(%struct.__jmp_buf_tag*, i32) #3
  ...
  call void @longjmp(%struct.__jmp_buf_tag* %4, i32 0) #5
  ...
```

So it looks like LLVM is capable of handling a call to setjmp/longjmp.

gnzlbg:
> So it looks like LLVM is capable of handling a call to setjmp/longjmp.

Note that just because it appears to sometimes work does not mean that it always work. 
Also, there is a crucial difference between C and Rust: C does not have destructors. 

C++, like Rust, does have destructors, and the [C++ standard support.runtime](http://eel.is/c++draft/support.runtime#csetjmp.syn-2) states:

> If any automatic objects would be destroyed by a thrown exception transferring control to another (destination) point in the program, then a call to `longjmp`(jbuf, val) at the throw point that transfers control to the same (destination) point has undefined behavior.

That is, the C++ standard includes explicit wording that states that calls to `longjmp` that skip destructors cannot happen (e.g., an optimizer is allowed to remove all such code as unreachable).

To me it is completely unclear how these "functions" should interact with Rust (e.g. how would `miri` execute them - cc @ralfj @oli-obk).  We should at least clarify that calling `longjmp` isn't always undefined behavior before we add it, because if that ends up being the case, then adding it is a mistake.

jeff-davis:
> > So it looks like LLVM is capable of handling a call to setjmp/longjmp.
> 
> Note that just because it appears to sometimes work does not mean that it always work.

But in this case clang is generating the calls to libc's setjmp/longjmp. If clang (written by the authors of LLVM) think it's OK to call setjmp/longjmp in LLVM, I don't see how it would be some inherent problem with LLVM.

> Also, there is a crucial difference between C and Rust: C does not have destructors.

I understand that. I have no desire to skip over destructors. I only want to use setjmp in the lowest rust frame before an FFI call to catch any longjmp. That *protects* me from skipping over rust frames, and there is currently *no other way to protect me from that*.

> To me it is completely unclear how these "functions" should interact with Rust (e.g. how would `miri` execute them - cc @ralfj @oli-obk). We should at least clarify that calling `longjmp` isn't always undefined behavior before we add it.

longjmp is a fact of life in C. If we ignore it, we are allowing those longjmps to unsafely skip over rust destructors. setjmp is the only solution that I see.

I also added longjmp for two reasons:

1. also helpful to me to turn rust panics back into longjmp before returning to a C function
1. symmetry after adding setjmp.

Though I don't need longjmp() quite so much, and can work around it more easily. If you want to only add setjmp, I am fine with that for now.

gnzlbg:
> I only want to use setjmp in the lowest rust frame before an FFI call to catch any longjmp. That protects me from skipping over rust frames, and there is currently no other way to protect me from that.

Why aren't you doing this in C? That is, why aren't you calling from Rust a C wrapper over postgres that uses `setjmp`/`longjmp` toi avoid having to use those from Rust ?

jeff-davis:
> > I only want to use setjmp in the lowest rust frame before an FFI call to catch any longjmp. That protects me from skipping over rust frames, and there is currently no other way to protect me from that.
> 
> Why aren't you doing this in C? That is, why aren't you calling from Rust a C wrapper over postgres that uses `setjmp`/`longjmp` to avoid having to use those from Rust ?

1. It would require a separate C wrapper for each C function I intend to call. For my use case, which is making it easier for people to write postgres extensions in rust, that would be wrapping hundreds of functions. In rust, I can use a macro, or wrap the function in rust, or do slightly different things for some functions when necessary.
1. It would be a separate build step in C with its own portability issues.

> > If we ignore it, we are allowing those longjmps to unsafely skip over rust destructors.
> 
> Adding these functions to `libc` does not make them magically work (these don't have to be part of libc for you to use it). AFAIK if I merge this PR I am adding a function to `libc` that always invokes undefined behavior when its called. The only fix for that would be for us to define the behavior of this function.

There are two functions: setjmp() and longjmp(). Which has undefined behavior, and by what reasoning is it undefined while other FFI calls are defined?

What is the cost of defining this behavior? One of the things that's great about rust is how well it integrates with C or any language that can integrate with C. Adding support for setjmp() and longjmp() seem to fit that spirit.

gnzlbg:
> here are two functions: setjmp() and longjmp(). Which has undefined behavior

`longjmp`

> and by what reasoning is it undefined while other FFI calls are defined?

All functions called via Rust FFI that we support either return their return value to the caller at the call site, or they never return. In particular, we forbid `extern` functions from performing other actions like unwinding the stack (e.g. if the C code calls C++ code that throws an exception).

`longjmp` does not return a value to the caller at the callsite but instead jumps to a different point in the program stack.

> Adding support for setjmp() and longjmp() seem to fit that spirit.

To me that looks like RFC territory.

gnzlbg:
Also, [C11 7.13.1.1p4](http://port70.net/~nsz/c/c11/n1570.html#7.13.1.1p4) states the following about `setjmp`:  

> An invocation of the setjmp macro shall appear only in one of the following contexts:
> 
> * the entire controlling expression of a selection or iteration statement;
> * one operand of a relational or equality operator with the other operand an integer constant expression, with the resulting expression being the entire controlling expression of a selection or iteration statement;
> * the operand of a unary ! operator with the resulting expression being the entire controlling expression of a selection or iteration statement; or
> * the entire expression of an expression statement (possibly cast to void).

While [C11 7.13.1.1p5](http://port70.net/~nsz/c/c11/n1570.html#7.13.1.1p5) states:

> If the invocation appears in any other context, the behavior is undefined.

That is, if you wanted to assign the result of `setjmp` to a "variable" in Rust, using `let x = setjmp(...);` the C standard says that at least in C that would be undefined behavior. Is that UB in Rust ? No idea, probably? But then we are adding a function with a return type that cannot be used where the return types of functions can normally we used. That sounds like RFC territory as well. 

Also, functions that use `setjmp` (and the extern `setjmp` function if you want to use it via C FFI) probably need to use LLVM [`return_twice`](https://llvm.org/docs/LangRef.html) attribute, which is specified as:

> This attribute indicates that this function can return twice. The C setjmp is an example of such a function. The compiler disables some optimizations (like tail calls) in the caller of these functions.

which specifically mentions `setjmp` and hints at possible mis optimizations if not used.

For example, this program ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8b3ad9e3f703b7e19639078b099c65ce)) passes in debug but fails in release due to a mis-optimization:

```rust
extern crate libc;
use libc::c_int;

pub type jmp_buf = [i64; 8];

extern "C" {
    fn setjmp(env: &mut jmp_buf) -> c_int;
    fn longjmp(env: &jmp_buf, val: c_int);
}

unsafe fn foo() -> i32 {
    let mut buf: jmp_buf = [0; 8];
    let mut x = 42;
    if setjmp(&mut buf) != 0 {
        return x;
    }
    x = 13;
    longjmp(&mut buf, 1);
    x
}

fn main() {
    assert_eq!(unsafe { foo() }, 13);
}
```

cc @rkruppe

jeff-davis:
> `longjmp`

I am willing to remove ``longjmp()`` from my PR if ``setjmp()`` is acceptable by itself. I think ``setjmp()`` is a nice safety feature when working with C.

gnzlbg:
Using `setjmp` as proposed in this PR is only ok as long as it only returns once. That is, Rust code calls it, it returns 0, and the program never jumps to it again. 

AFAIK that would make `setjmp` useless, since its raison-d'être is being able to `longjmp` to it from somewhere else.

For that, AFAICT, it cannot be added as a simple extern function. We'd have to add it as a `core::intrinsic` to Rust C, and apply the appropriate LLVM attributes to it to prevent mis-optimizations like the one I showed above.

gnzlbg:
> That can be done by a (Rust-level) unstable attribute instead of via a new intrinsic, but some compiler integration is needed. 

Note that `libc` is a normal crate on `crates.io`. If we do this via an attribute, `libc` would need to use it, and it will need to become stable at some point. Making `setjmp` an intrinsic allows us to stabilize the intrinsic without stabilizing the attribute, but as you mention the attribute might be required by any code trying to wrap `setjmp` anyways..

LLVM also exposes `setjmp`/`longjmp` intrinsics, so it might be worth it to consider exposing those instead of manually linking to the system symbols.

> And about longjump skipping over destructors [...] adding it as an unsafe function to libc doesn't cause any unsoundness,

Rust does not guarantee that destructors run, so skipping them isn't unsound, just annoying for unsafe code that assume that it does not happen. My worry about `longjmp` is that it is not 100% clear to me which stack state it restores, e.g., could it restore the value of drop flags creating double drops? I don't know, but that would be unsound.

This is not a good example, but it has a double drop ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=51d6407335cd5156e54eb088ed584cac)): 

```rust
fn foo() {
    let mut buf: jmp_buf = [0; 8];
    let a = A;
    if unsafe { setjmp(&mut buf) } != 0 {
        return;
    }
    std::mem::drop(a);
    unsafe { longjmp(&mut buf, 1) };
}
```

gnzlbg:
@jeff-davis I think we should definitely encourage people to "play" with these intrinsics, and for interfacing with C Rust is going to need to provide some kind of support for these.

Would you be ok with adding them behind an experimental cargo feature ? (e.g. `unstable_setjmp_longjmp`) ? This allows us to test the platform specific details here in CI, and gives those wanting to play with them a semi reliable way to do so. This also allows us to move them to `core::intrinsics` at some point in the future without breaking stable users. 

In the meantime, we should open an issue in rust-lang/rust about what to do about these. We'll probably need to write down what code that uses these is allowed to do to avoid undefined behavior and that's going to be hard.

How does that sound?

gnzlbg:
So I've opened an issue in the RFC repo: https://github.com/rust-lang/rfcs/issues/2625 

@jeff-davis mentioned:

> 1. It would require a separate C wrapper for each C function I intend to call. For my use case, which is making it easier for people to write postgres extensions in rust, that would be wrapping hundreds of functions. In rust, I can use a macro, or wrap the function in rust, or do slightly different things for some functions when necessary.
>
> 2. It would be a separate build step in C with its own portability issues.

Even if we add `setjmp` to libc in this PR behind a feature gate, using it would be instantaneous undefined behavior, result in mis-optimizations, uses after move, etc. and there would be very little that you could do to avoid that. 

The main advantage of writing a C wrapper over postgres that catches postgres exceptions and translates them to Rust error codes (where in Rust you can then raise them again as a panic if you want, or do something else) is that this approach would at least work and that there is experience from doing this same thing to C++ libraries that throw exceptions (we don't have tools in Rust to catch C++ exceptions, C++ thin wrappers that catch them and translate them to error codes need to be written here).

I understand that this solution is sub-optimal, but coming up with ways to make `setjmp`/`longjmp` "barely work" will take time.

burdges:
Is there no way to abuse variable length arguments in C to create one C wrapper function that supports calling all the C functions?

jeff-davis:
Thank you for the detailed replies. They still leave a few puzzles:

I translated your rust example to C, and I get:

```
jdavis@jdavis:~/code/c$ ./a.out
foo(): 13
jdavis@jdavis:~/code/c$ clang-6.0 -O2 setjmp2.c
jdavis@jdavis:~/code/c$ ./a.out
foo(): 42
jdavis@jdavis:~/code/c$ gcc setjmp2.c
jdavis@jdavis:~/code/c$ ./a.out
foo(): 13
jdavis@jdavis:~/code/c$ gcc -O2 setjmp2.c
jdavis@jdavis:~/code/c$ ./a.out
foo(): 42
```

So that means that either (a) the behavior of that pattern of use of setjmp is also undefined in C; or (b) both gcc and clang have the same correctness bug. I have to assume (a), because (b) seems quite unlikely.

The example would be more compelling if you showed a correctness bug in rust for well-defined behavior in C.

But, I agree that we at least should have the ``returns_twice`` attribute on ``setjmp()``. I am fine making this an RFC.

jeff-davis:
@gnzlbg: Yes, the experimental cargo feature sounds like a nice way to proceed for now. Thank you.

gnzlbg:
> I have to assume (a), because (b) seems quite unlikely.

In my translations, both GCC and Clang always return 13 =/  

Maybe you forgot to make `x` `volatile` ? In C, per [C 7.13.2.1p3](http://port70.net/~nsz/c/c11/n1570.html#7.13.2.1p3):

> All accessible objects have values, and all other components of the abstract machine249) have state, as of the time the `longjmp` function was called, except that the values of objects of automatic storage duration that are local to the function containing the invocation of the corresponding `setjmp` macro **that do not have volatile-qualified type and have been changed between the setjmp invocation and longjmp call are indeterminate**.

one **must** make `x` `volatile` because otherwise its value is _indeterminate_. AFAIK Rust does not have a concept of "indeterminate" values, so I don't know whether it makes sense to use that term in Rust. 

Also, Rust does not have a "`volatile`" storage specifier, so `x` is ""_indeterminate_"" and there is not much that one can do about that. In [this comment](https://github.com/rust-lang/libc/pull/1216#issuecomment-455810681) I mentioned how one can use volatile pointer load and stores to workaround this issue for that particular example ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e6ca15a0066efe3548b30ef2ed29a8a6)).  However, that won't help you with the load / stores that the compiler inserts for code that you don't directly write though (drop code, etc.).

Basically, there is no way to write Rust code that's equivalent to the correct C code because of `Drop`.

jeff-davis:
> > I have to assume (a), because (b) seems quite unlikely.
> 
> In my translations, both GCC and Clang always return 13 =/

Oh, you're right. I forgot to mark x volatile.

PR #1217, opened 2019-01-19
RFC 2235 - Implement PartialEq,Eq,Hash,Debug for all types

Susurrus:
First pass at implementing [RFC2235](https://github.com/rust-lang/rfcs/blob/master/text/2235-libc-struct-traits.md). I just started with `PartialEq`/`Eq` and tested locally on my x64 Linux system. Definitely going to run into other types for other platforms that need this treatment, but thought I'd start small.

Open question is whether this should also rely on the `align` feature, which should improve which types can be auto-derived versus manually implemented (as it removed a lot of odd-sized padding arrays). My first pass here doesn't do that, but I might test it out to see if it does simplify quite a few structs. I think it might also be nice to have as it makes it easier for contributors to add new structs.

Part of rust-lang/rust#57715


Cargo.toml
```diff
@@ -27,6 +27,7 @@ default = ["use_std"]
 use_std = []
 align = []
 rustc-dep-of-std = ['align', 'rustc-std-workspace-core']
+extra_traits = []
```
gnzlbg: We might want to start by using more fine grained features here, and having an `extra_traits` feature that enables them all.

src/unix/notbsd/linux/mod.rs
```diff
@@ -657,6 +549,169 @@ s! {
     }
 }
 
+s_no_extra_traits!{
```
gnzlbg: One could manually derive the traits for these, right? Might want to add a TODO here.

gnzlbg:
> Open question is whether this should also rely on the align feature, which should improve which types can be auto-derived versus manually implemented (as it removed a lot of odd-sized padding arrays). 

I think so. `align` is stable and it is only available as a feature to support older Rust versions. Since this is a new feature, if `align` simplifies it, then it should depend on it. I don't see any downsides here. The same applies to `packed`.

Susurrus:
> We might want to start by using more fine grained features here (e.g. traits_eq), and having an extra_traits feature that enables them all.

I'd prefer to not implement it like this. When we talk about maintainability costs, having more features is harder than less features. The RFC also had discussion about this, but I'd like to find out more about why it can't be done as the single trait as I've already implemented it before changing this to individual features.

> LGTM. I worry about the extra maintainability cost of the structs for which the traits cannot be derived automatically (because they contain arrays with more than 32 elements).

Do you still have this concern after seeing the implementation? I'm uncertain what the maintainability cost will be that you're concerned about. There isn't much change to types after they're added to `libc`, so I'm wondering what you're referring to here.

> I think so. align is stable and it is only available as a feature to support older Rust versions. Since this is a new feature, if align simplifies it, then it should depend on it. I don't see any downsides here. The same applies to packed(N) - by the time this hits stable, packed(N) will already be stable.

I did this, though looking at the code I didn't see any types I manually implemented this for that would actually benefit from this. I still think it's the right call, as it may help in structs that haven't been added yet though.

So I've completed implementing everything here, though there are some open questions on what to do for the fields that I skipped in my manual implementations. I'll see if they're trivial to implement manually or I'll ask if we can set guidelines such that some fields can be skipped to make things easier.

gnzlbg:
Oh, I see that the documentation build bots are still failing. Could you fix those? 

This error sounds familiar: cc @QuietMisdreavus 

```
rustdoc -o target/doc/x86_64-unknown-linux-gnu --target x86_64-unknown-linux-gnu src/lib.rs --cfg cross_platform_docs --crate-name libc
error[E0433]: failed to resolve: maybe a missing `extern crate clone;`?
  --> src/macros.rs:42:22
   |
42 |             #[derive(Clone, Copy)]
   |                      ^^^^^ maybe a missing `extern crate clone;`?
error[E0433]: failed to resolve: maybe a missing `extern crate marker;`?
  --> src/macros.rs:42:29
   |
42 |             #[derive(Clone, Copy)]
   |                             ^^^^ maybe a missing `extern crate marker;`?
error[E0433]: failed to resolve: maybe a missing `extern crate clone;`?
  --> src/macros.rs:54:22
   |
54 |             #[derive(Clone, Copy)]
   |                      ^^^^^ maybe a missing `extern crate clone;`?
error[E0433]: failed to resolve: maybe a missing `extern crate marker;`?
  --> src/macros.rs:54:29
   |
54 |             #[derive(Clone, Copy)]
   |                             ^^^^ maybe a missing `extern crate marker;`?
```

QuietMisdreavus:
It looks like the `cross_platform_docs` cfg set by the doc builders removes libcore? It doesn't seem like anything is being put in its place, for the derive macros to be able to use...

https://github.com/rust-lang/libc/blob/fb2b3dac03d4a9ed002893706ff28b364b969721/src/lib.rs#L16-L17

Maybe some kind of import for `dox::{Clone, Copy}` that put the traits in the right module paths would fix this? Doesn't seem like a rustdoc-specific problem, though.

QuietMisdreavus:
Actually, after digging into it, why are those doc builders still going if libc's docs are going to be hosted on docs.rs (which doesn't use `--cfg cross_platform_docs`)?

gnzlbg:
That's a good question. I'd expect that `libc` docs are shipped with Rust (because one can directly do  `extern crate libc` to get the same `libc` as `libstd`) but AFAICT that's not the case.

Susurrus:
I also wanted to ask about the `Cargo.lock` file. I had to update it for this, but since this is a lib, the `Cargo.lock` doesn't need to be committed. Should I remove it in a first commit and add it to the `.gitignore`?

Susurrus:
I've also been thinking that we should probably have a test that checks that each type has the right extra traits. This will ensure any manually-implemented types won't miss traits they should have.

gnzlbg:
@Susurrus don't worry about the update to the `Cargo.lock`. 

> I've also been thinking that we should probably have a test that checks that each type has the right extra traits. This will ensure any manually-implemented types won't miss traits they should have. But I don't know how to start that implementation; any ideas anyone?

IMO the right way to do this is via a clippy lint. That is, we would have (or add) a lint to clippy, e.g., `missing_debug_impl_on_public_items`, that requires all public types to implement, e.g., `Debug`. This lint would be disabled by default in clippy for obvious reasons, and we would opt-into it by enabling it for libc at the top level:

```rust
#![cfg_attr(feature = "cargo-clippy", deny(missing_debug_impl_on_public_items))]
```

This lint will error if a public item does not implement `Debug`. If there is an item for which we explicitly don't want to implement `Debug`, then we can opt-out for that item manually:

```rust
#[cfg_attr(feature = "cargo-clippy", allow(missing_debug_impl_on_public_items))]
struct LookMaNoDebug;
```

For every trait, we would need such a lint. You can open an issue about this in clippy upstream and cite this comment as motivation. Implementing such a lint is not hard, I can try to mentor you if you want, but clippy maintainers have a quick turnaround time and if you are in the Discord clippy channel they can tell you which existing lint you can base yours on to get it done quickly.

FYI there is prior art of doing this. For example, in `stdsimd` we need most of the API to be `#[inline]` or `#[inline(always)]`, so I added a clippy lint [`missing_inline_in_public_items`](https://rust-lang.github.io/rust-clippy/master/index.html#missing_inline_in_public_items) that checks that, and we use it in `stdsimd` and `packed_simd` successfully (it has caught a couple of bugs).

alexcrichton:
r? @gnzlbg 

everything looks good to me so far, thanks! I'll defer to @gnzlbg from here

alexcrichton:
wrong button...

Susurrus:
I would like some clarity on `Cargo.lock`. Can I remove it from the repo and add it to `.gitignore`? It seems like unnecessary noise for a library.

alexcrichton:
That is an artifact of when a `git` dependency on `ctest` was used. That's no longer the case so I think it's fine to remove

Susurrus:
Removed `Cargo.lock` (it was already in `.gitignore` coincidentally) and also added the `style` executable to `.gitignore` as that shouldn't ever be committed. Fixed FreeBSD builds and rebased.

Susurrus:
Need a comment on db7b0c69 and whether I should keep it, cause it breaks the builds for 1.13. I don't know how much value it is, so it's probably not a problem to remove it.

Note that because `Copy` is manually implemented, I can't derive `Debug`, `Hash`, and `PartialEq` as I get the following warning:
```
> cargo build --features extra_traits                                                                                                                       ±[rfc_2235]
   Compiling libc v0.2.48 (/.../Projects/libc)                                                                                                                                  
warning: #[derive] can't be used on a #[repr(packed)] struct that does not derive Copy (error E0133)                                                                                      
   --> src/macros.rs:42:57                                                                                                                                                                
    |                                                                                                                                                                                     
42  |               #[cfg_attr(feature = "extra_traits", derive(Debug, Eq, Hash, PartialEq))]                                                                                             
    |                                                           ^^^^^                                                                                                                     
    |                                                                                                                                                                                     
   ::: src/unix/notbsd/mod.rs:18:1                                                                                                                                                        
    |                                                                                                                                                                                     
18  | / s! {                                                                                                                                                                              
19  | |     pub struct sockaddr {                                                                                                                                                         
20  | |         pub sa_family: sa_family_t,                                                                                                                                               
21  | |         pub sa_data: [::c_char; 14],                                                                                                                                              
...   |                                                                                                                                                                                   
212 | |     }                                                                                                                                                                             
213 | | }                                                                                                                                                                                 
    |                                                                                                                                                                                     
    = note: #[warn(safe_packed_borrows)] on by default                                                                                                                                    
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!                                                     
    = note: for more information, see issue #46043 <https://github.com/rust-lang/rust/issues/46043>          
```

I'm wondering if there's a way we can instead derive `Clone` and `Copy`. I originally had this PR doing that, but it broke the `cross_platform_docs` stuff. Any suggestions?

Additionally, what's the verdict on whether a lint needs to be implemented for `Hash`, `Eq`, and `PartialEq` for this to get merged?

I also did a compilation test of the feature on versus off. Took 2.93s to compile with `extra_traits` enabled and 1.38s without with a clean build.

gnzlbg:
> Need a comment on db7b0c6 and whether I should keep it, cause it breaks the builds for 1.13. 

We have some code in the `build.rs` to handle differences in Rust versions: https://github.com/rust-lang/libc/blob/master/build.rs#L10

I don't know if it is worth it to handle this for Rust 1.13, but if you are motivated, that could be a solution.

Susurrus:
This is a syntax issue that can't be worked around like that. We'd either need to move the lint check to be specified at the command line or remove it. I'm actually in favor of removing it, so I'm gonna do that.

Susurrus:
@asomers This work has a direct impact on `nix` and exposing `Debug` in our types there. Thought I'd CC you here in case you wanted to review this as well.

src/unix/bsd/apple/mod.rs
```diff
@@ -35,6 +35,10 @@ pub type key_t = ::c_int;
 pub type shmatt_t = ::c_ushort;
 
 pub enum timezone {}
```
asomers: Why does freebsdlike's timezone get Debug but apple's does not?

src/unix/notbsd/emscripten.rs
```diff
@@ -36,6 +36,10 @@ pub type c_ulong = u32;
 pub type nlink_t = u32;
 
 pub enum fpos64_t {} // TODO: fill this out with a struct
```
asomers: Why does Linux's `fpos64_t` get Debug but emscripten's does not?

src/unix/notbsd/linux/mod.rs
```diff
@@ -657,6 +554,255 @@ s! {
...
+        self.size.iter().zip(other.size.iter()).all(|(a,b)| a == b)
+    }
+}
+#[cfg(feature = "extra_traits")]
+impl Eq for pthread_cond_t {}
+#[cfg(feature = "extra_traits")]
+impl std::fmt::Debug for pthread_cond_t {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        f.debug_struct("pthread_cond_t")
+            // FIXME: .field("size", &self.size)
```
asomers: I don't think there's anything to fix.  This entire structure is opaque.

src/unix/notbsd/mod.rs
```diff
@@ -230,6 +212,152 @@ s! {
...
+        self.sun_path.hash(state);
+    }
+}
+
+#[cfg(feature = "extra_traits")]
+impl PartialEq for sockaddr_storage {
+    fn eq(&self, other: &sockaddr_storage) -> bool {
+        self.ss_family == other.ss_family
+            && self
+                .__ss_pad2
```
asomers: I think you need to compare the __ss_align fields too, because they could alias to a useful field of some other sockaddr struct.

src/unix/notbsd/mod.rs
```diff
@@ -230,6 +212,152 @@ s! {
...
+            .field("ss_family", &self.ss_family)
+            // FIXME: .field("__ss_pad2", &self.__ss_pad2)
+            .finish()
+    }
+}
+#[cfg(feature = "extra_traits")]
+impl std::hash::Hash for sockaddr_storage {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.ss_family.hash(state);
+        self.__ss_pad2.hash(state);
```
asomers: Similarly, you should hash the __ss_align field too.

gnzlbg:
r? me

gnzlbg:
> Are the BSD platforms not linted in the same way?

If a target is not built in CI, its code is `cfg`'ed out, so none of the lints will see it. Cirrus CI should build FreeBSD and test that, but some of the other *BSDs are not built in CI at all. 

Some of the other BSDs are not built at all on CI because setting qemu to run the libc-tests for them is not trivial. We could add build bots for these that just cross-compile libc for these targets, without running libc-test. That would be better than what we currently do, since that checks that libc at least cross-compiled properly, and it would also run all these lints to check for missing impls. 

We should probably do that in subsequent PRs instead of here (I'll open an issue).

asomers:
Oh, I see now.  The only reason that CI is passing for FreeBSD is because CI isn't testing with `extra_traits` enabled, for any OS.  .travis.yml and .cirrus.yml should both be updated to test with that flag.

ci/run.sh
```diff
@@ -96,4 +96,8 @@ fi
 if [ "$(rustc --version | sed -E 's/^rustc 1\.([0-9]*)\..*/\1/')" -ge 25 ]; then
   cargo test $opt --features align --manifest-path libc-test/Cargo.toml --target "${TARGET}"
 fi
+# Test the #[repr(align(x))] feature if this is building on Rust >= 1.25
```
gnzlbg: Replace `#[repr(align(x))]` with `extra_traits`.

gnzlbg:
It appears that the lints and tests have uncovered a couple of issues (missing Copy impls, Debug impls that cannot be derived, etc.).

asomers:
Why use clippy? What's wrong with rustc's `#[deny(missing_debug_impls)]`

Susurrus:
> Why use clippy? What's wrong with rustc's #[deny(missing_debug_impls)]

I'm already going with `rustc`'s `Debug` and `Copy` lints. I think @gnzlbg's point was just to use lints in general and they didn't really care if they were `rustc` or Clippy ones at the end of the day.

Susurrus:
I think I've sorted everything out at this point. I probably need to do another visual run through of the complete diff, but it's shaping up nicely. Only blocker here is what to do about implementing `Hash` for a struct that has an `f64` in it. Any suggestions?

I've also been making some headway with rust-lang/rust#58070 and might the desired lints for every trait we want done this weekend, but I don't think we'll want to block on that, since once it's merged it'll need to ride the train to stable.

Susurrus:
Down to just `s390x-unknown-linux-gnu` failing due to hashing an `f64`, which you can't do by default. I'll try to investigate this on my own unless someone has some ideas.,

README.md
```diff
@@ -44,6 +44,16 @@ activate the *align* feature. This requires Rust 1.25 or newer:
 libc = { version = "0.2", features = ["align"] }
 ```
 
+All structs implemented by the libc crate have the `Copy` and `Clone` traits
+implemented for them. The additional traits of `Debug, `Eq`, `Hash`, and
```
auscompgeek: Looks like there's a missing backtick after Debug.

gnzlbg:
> Only blocker here is what to do about implementing Hash for a struct that has an f64 in it. Any suggestions?

I'd say, don't derive Hash for this, and open an issue to discuss what to do. No need to block this PR on that.

Susurrus:
Note that all failures in CI here were sporadic and things should work if someone goes in and re-runs the failures.

Susurrus:
So this is pretty much ready, just needs someone to poke at CI for me as we should be all good here.

semarie:
just to point that targets without CI are a bit broken now (openbsd in mind). I will try to take a look

PR #1230, opened 2019-01-26
Added inotify bindings.

vdagonneau:
Hi, I'd like to add inotify bindings. This is a first throw at it. Can you guide me through the process of merging it ?

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -655,6 +655,13 @@ s! {
         pub updated: ::c_ulong,
         pub ha: [::c_uchar; ::MAX_ADDR_LEN],
     }
+
+    pub struct inotify_event {
+        pub wd: ::c_int,
+        pub mask: ::uint32_t,
+        pub cookie: ::uint32_t,
+        pub len: ::uint32_t
+    }
 }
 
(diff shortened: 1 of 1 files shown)


gnzlbg:
> Hi, I'd like to add inotify bindings. This is a first throw at it. 

Thank you for working on this, and I apologize for the delay!

> Can you guide me through the process of merging it ?

Sure! The error message in the travis-ci builds says that the type used "is incomplete". This probably means that the C code that we are generating to validate the API does not know about this type. The reason for this is probably that an `#include ...` is missing in this generated code. 

The comment that you have added hints that these types and constants are in the `uapi/linux/inotify.h` file, so this might be the file that's missing.

We automatically generate the `#include ...` directives for these headers for linux somewhere around here in this build.rs file: https://github.com/rust-lang/libc/blob/master/libc-test/build.rs#L215 

Try committing a change that adds that header there and we'll see if that makes the tests pass. 

Please don't hesitate asking for clarification if you have any further questions. I'm also on Discord and Zulip if you need interactive help.

vdagonneau:
So, everything seems to be working except for the android arm build. It seems to have just timeout-ed.

gnzlbg:
@vdagonneau  yes that's correct, that failure is probably spurious. I've told bors to merge this but the queue appears to be having some issues.

PR #1233, opened 2019-01-29
[WIP] Test all targets with Rust nightly, beta, stable, and 1.13.0

### LINKED ISSUE #1232: Libc on osx now requires rust 1.24.0+
The rand crate tries to maintain compatibility with rust 1.22.0, but a recent release of libc now implicitly requires [1.24.0 on osx](https://github.com/rust-random/rand/issues/719) due to this patch: rust-lang/libc#1212, which uses a constant function that was stablized in 1.24. It appears that the Travis config is trying to maintain compatibility with 1.13, but this test only runs on Linux.

Now I personally don't care what is the minimum version you want to support, but given how libc is now the most actively used dependency, you drive the minimum version of a lot of the ecosystem, so would it be possible to at least document this requirement, or setup some more test targets to make sure this doesn't [happen again accidentally](https://github.com/rust-lang/libc/issues/1229)?

gnzlbg:
Closes #1232 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,16 @@ services: docker
 
 stages:
   - min-checks-and-tier1
-  - other-targets
+  - tier2-1
+  - tier2-2
+  - tier2-3
+  - tier2-4
+  - tier25
+  - tier3
 
 matrix:
   include:
+    # Min Checks and Tier 1 targets
     - name: "Documentation"
       env: TARGET=x86_64-unknown-linux-gnu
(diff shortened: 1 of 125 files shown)


vks:
Could you please make sure to document the minimal supported Rust version in the README?

gnzlbg:
@vks I've added a Rust version support table to the readme, specifying the different levels of Rust version support that libc requires.

gnzlbg:
This will be cleaned up in a new PR.

PR #1235, opened 2019-02-01
Fix cmsg(3) bugs for musl and OSX

asomers:
This PR fixes bugs in the cmsg(3) family of functions for Linux/musl and OSX, introduced by PR #1098 and PR #1212 .  It also adds an integration test which hopefully will validate these functions on every platform.

### DIFF
--- a/ci/docker/x86_64-rumprun-netbsd/runtest.rs
+++ b/ci/docker/x86_64-rumprun-netbsd/runtest.rs
@@ -47,7 +47,8 @@ fn find_ok(input: &mut Read, tx: mpsc::Sender<()>) {
     for line in BufReader::new(input).lines() {
         let line = line.unwrap();
         println!("{}", line);
-        if line.starts_with("PASSED ") && line.contains(" tests") {
+        if (line.starts_with("PASSED ") && line.contains(" tests")) ||
+            line.starts_with("test result: ok"){
             tx.send(()).unwrap();
         }
     }
--- a/ci/ios/deploy_and_run_on_ios_simulator.rs
+++ b/ci/ios/deploy_and_run_on_ios_simulator.rs
@@ -129,8 +129,11 @@ fn run_app_on_simulator() {
 
     let stdout = String::from_utf8_lossy(&output.stdout);
     let passed = stdout.lines()
-                       .find(|l| l.contains("PASSED"))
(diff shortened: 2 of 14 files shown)


gnzlbg:
Thank you for working on this @asomers. It appears that the tests are failing on some targets, maybe some more fine tuning is necessary.

asomers:
I'm pretty sure that the s390x and sparc64 failures are not my fault.  On s390x the test program hangs.  On sparc64, it gets SIGILL.  In both cases, the problem happens before it invokes the first cmsg test function.  It's not Travis's fault; I can duplicate the problem locally.  It might be QEMU bugs.  If nobody else has a better idea, I'll disable the test for those architectures.

asomers:
The two Android failures looks spurious; somebody should restart them.  However, the s390x and sparc64 failures are more serious.  They seem to be failing early in the test harness.  Cargo doesn't provide a way to disable the entire test program on a per-target basis.  So I'm stuck.  If the decision were up to me; I would just remove those platforms from the test matrix.

gnzlbg:
@asomers could you fill in a bug in rust-lang/rust about the sparc64 and s390x failures ? We are able to run a lot of tests in other rust-lang repos like packed_simd, jemallocator, etc. in these platforms, so I find it weird that they fail for something "this simple".

gnzlbg:
@bors: r-

(let me know when you rebase / squash)

---

We should definitely figure out why these particular tests break on those targets (maybe a qemu bug, maybe a rustc/llvm bug), but we don't have to do that here. I don't think disabling all testing on those targets makes things better for these targets. So skipping these tests until we figure things out seems reasonable to me.

asomers:
Ok, now there's an ACTUAL failure on sparc64.  But given how much time I've already spent on this, I'm inclined just to skip that test on sparc64 and open an issue for it.  I'll let somebody who actually uses sparc64 and/or Linux fix it.

The android failures are a trivial mistake.  I'm locally investigating the s390x failure now.

asomers:
Ok @gnzlbg everything's passing now.  I only had to disable the cmsg test on s390x and sparc64.  Let me know when you're satisfied review-wise and I'll squash things.

gnzlbg:
The Cargo.lock needs updating. @alexcrichton should we just remove the `Cargo.lock` ?

PR #1241, opened 2019-02-06
unbreak openbsd after #1217

semarie:


gnzlbg:
How are you testing openbsd ? Rustup does not appear to ship libcore for `i686-unknown-openbsd` nor `x86_64-unknown-openbsd`.

semarie:
I ran rustc natively on OpenBSD -current (upcmoing 6.5). here, test was done using stable (1.32.0).

gnzlbg:
Could you test if https://github.com/rust-lang/libc/pull/1233 also solves that ?

semarie:
@gnzlbg `ci/run.sh` terminate with success, but I am unsure to understand all the details: in order to make my PR to pass tests I needed to manually implement all traits of rfc2235 for several structs (union, or containing big static array as struct member like `[::c_char; 256]`), and I don't see them in your PR.

gnzlbg:
That's because my PR handwaves that by using `#[allow(missing_debug_implementations)]`. The `extra_traits` PR broke pretty much most libc targets, and we did not realize cause we weren't testing them.

gnzlbg:
@semarie it would be great if you could rebase on top of master and test that the build on OpenBSD still works. You might also want to try to see if more cross-compiling OpenBSD targets could be shipped with rustup for linux (or if there are already some more available), and add them to the `ci/build.sh` script to at least ensure that `libc` contains no compilation errors for OpenBSD targets.

semarie:
@gnzlbg OpenBSD has no support at all in rustup. Last time I looked at it it was required to me to do all the work to provide a working bootstrap environment for OpenBSD on Linux (which I don't use). At this time, it was a very difficult task as previous OpenBSD versions used hightly customized linker (aka `ld`) which didn't compile under Linux. The current situation should be better now as we use `ld.lld` (but I didn't check if all patches are upstreamed or not).

Issue #1243, opened 2019-02-06
Tracking issue for the extra_traits features


gnzlbg:
[RFC2235](https://github.com/rust-lang/rfcs/blob/master/text/2235-libc-struct-traits.md).

https://github.com/rust-lang/rust/issues/57715

* [ ] not all types derive all traits yet (e.g. grep for `missing_debug_implementations`)

cc @Susurrus

gnzlbg:
let's track this here instead: https://github.com/rust-lang/rust/issues/57715

PR #1247, opened 2019-02-07
Build all platforms

gnzlbg:
cc @alexcrichton this needs a more thorough review. It turns out libc failed to build on a lot of older Rust versions for various reasons and platforms, so it took a while to fix the build on all of them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,11 @@ Raw FFI bindings to platform libraries like `libc`.
 **NOTE:** The minimum supported Rust version is **Rust 1.13.0** . APIs requiring
 newer Rust features are only available on newer Rust versions:
 
-| Feature                 | Version |
-|-------------------------|---------|
-| `union`                |  1.19.0 |
+| Feature              | Version |
+|----------------------|---------|
+| `union`              |  1.19.0 |
 | `const mem::size_of` |  1.24.0 |
-| `repr(align)`         |  1.25.0 |
+| `repr(align)`        |  1.25.0 |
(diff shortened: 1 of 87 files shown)


gnzlbg:
cc @Susurrus I had to disable some `extra_trait`s in targets in which they were failing to compile (due to too large arrays, `repr(packed)`, etc.). These should be easy to find and fix, i've opened #1243 to track some of these.

README.md
```diff
@@ -8,7 +8,19 @@ Raw FFI bindings to platform libraries like `libc`.
...
 [![Documentation](https://docs.rs/libc/badge.svg)](https://docs.rs/libc)
-![License](https://img.shields.io/crates/l/libc.svg)
+![License](https://img.shields.io/crates/l/libc.svg**
+
+**NOTE:** The minimum supported Rust version is **Rust 1.13.0** . APIs requiring
+newer Rust features are only available on newer Rust versions:
+
+| Feature                 | Version |
+|-------------------------|---------|
+| `union`                |  1.19.0 |
```
Susurrus: I'd recommend manually aligning this table with spaces so it's easier to read un-rendered

gnzlbg: Can do - I hope github uses a monospace font for rendering the diffs.

gnzlbg:
I'm merging this because the `extra_traits` PR broke the library in a couple of platforms and this PR fixes those and ensures that they don't break again. We can iterate on this on master before the next release.

PR #1253, opened 2019-02-12
Always use the latest ctest

gnzlbg:


gnzlbg:
hmm, it appears that high-five still has you as a reviewer

alexcrichton:
Ah it just takes some time to update so it'll be there for a bit, but lgtm regardless :)

PR #1257, opened 2019-02-13
Add IP_ORIGDSTADDR and others on FreeBSD.

vi:
A follow-up from #1252.

CC @asomers.

r? @gnzlbg

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -725,6 +725,16 @@ fn do_ctest() {
             s if ios && s.starts_with("RTV_") => true,
             s if ios && s.starts_with("DLT_") => true,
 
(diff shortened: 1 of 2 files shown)


vi:
I'm not sure what that test does and why it fails.

gnzlbg:
> Maybe that constnats were added to FreeBSD too recently?

We don't test FreeBSD 12 on CI, only 11. You probably need to skip these constants in the `build.rs` for `freebsd`.

vi:
> You probably need to skip these constants in the build.rs for freebsd.

I'm not sure how to do it.

gnzlbg:
Something like this: https://github.com/rust-lang/libc/blob/master/libc-test/build.rs#L567

PR #1259, opened 2019-02-14
Cleanup dox mess - correction

semarie:
unbreak openbsd after 7ac0fe53e a `}` was missing


gnzlbg:
@bors: r+ 

FYI I have a PR in the pipeline that uses `xargo` to compile `libc`+`libcore` on targets that do not ship a std component on CI. Hopefully after that's merged you won't have to send these kinds of PRs anymore :)

gnzlbg:
@semarie Thank you for the PR, I'm closing it because this should be fixed by #1263 , which should prevent this from breaking again, and also generate docs for openbsd (we'll see if the docs part works out).

PR #1261, opened 2019-02-18
Add AF_ALG constants and structures

glebpom:


src/unix/notbsd/linux/other/mod.rs
```diff
@@ -321,7 +321,6 @@ pub const SOL_PNPIPE: ::c_int = 275;
 pub const SOL_RDS: ::c_int = 276;
 pub const SOL_IUCV: ::c_int = 277;
 pub const SOL_CAIF: ::c_int = 278;
-pub const SOL_ALG: ::c_int = 279;
```
gnzlbg: Why is this constant being removed?

src/unix/notbsd/mod.rs
```diff
@@ -253,6 +240,19 @@ s_no_extra_traits!{
         __ss_pad2: [u8; 128 - 2 * 8],
     }
 
+    pub struct sockaddr_alg {
```
gnzlbg: Why was this code moved, it makes it hard to inspect what changed.

src/unix/notbsd/linux/other/mod.rs
glebpom: The idea is to support AF_ALG related constants for both Linux and Android.

src/unix/notbsd/mod.rs
glebpom: Only `SOL_ALG` was moved, other types were not defined prior to this PR

gnzlbg: Weird, you are right. I'm on the train without quality internet, and github was showing me a negative diff here for some reason. Forget that.

src/unix/notbsd/linux/other/mod.rs
gnzlbg: I guess what I am asking is, this constant was in `notbsd/linux/other/mod`, and now it is in `notbsd/mod.rs`. Have you checked that the new value is correct for `emscripten`, `musl`, etc. ?

glebpom: [x] emscripten - https://github.com/emscripten-core/emscripten/blob/6e4b98636618989bcd99308391e51aa1b81f4c61/system/include/libc/sys/socket.h#L253
[x] musl https://git.musl-libc.org/cgit/musl/tree/include/sys/socket.h#n266
[x] android https://github.com/aosp-mirror/platform_bionic/blob/3e5a32ce2c24969932634d84e2755755094fa582/libc/include/sys/socket.h#L288

I think it  doesn't make any sense for `emscripten`. What it the best location for constants applicable only for Linux and Android?

gnzlbg: Thanks, LGTM. 

> I think it doesn't make any sense for `emscripten`. What it the best location for constants applicable only for Linux and Android?

Add them independently to `linux/mod.rs` and `android/mod.rs`, but in this case since they are also available in `emscripten` (even though they are probably useless), adding them to the `notbsd/mod.rs` module is fine.

glebpom:
Tests which explicitly download the kernel version 3.12.6 are failing. Minimum kernel version 3.19 is required for `ALG_SET_AEAD_ASSOCLEN` and `ALG_SET_AEAD_AUTHSIZE`. Any ideas on how to fix it? I would suggest migrating to [musl-cross-make](https://github.com/richfelker/musl-cross-make). That would be far more maintainable but will take more time to build. Use [travis caching](https://docs.travis-ci.com/user/caching/)?

gnzlbg:
> Any ideas on how to fix it?

The best would be to skip these in the `build.rs` of `libc-test`.

glebpom:
The build is passing now. Would you review the code again?

src/unix/notbsd/android/mod.rs
```diff
@@ -451,6 +464,81 @@ cfg_if! {
...
+           }
+        }
+
+        impl Eq for sockaddr_alg {}
+
+        impl ::fmt::Debug for sockaddr_alg {
+            fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                f.debug_struct("sockaddr_alg")
+                    .field("salg_family", &self.salg_family)
+//                    .field("salg_type", &self.salg_type)
```
gnzlbg: why are these commented out?

libc-test/build.rs
```diff
@@ -627,6 +630,8 @@ fn do_ctest() {
             | "RENAME_NOREPLACE"
             | "RENAME_EXCHANGE"
             | "RENAME_WHITEOUT"
+            | "ALG_SET_AEAD_ASSOCLEN"
```
gnzlbg: Could you add a comment stating whats the oldest kernel with these?

glebpom:
Should be good now

src/unix/notbsd/android/mod.rs
```diff
@@ -451,6 +464,81 @@ cfg_if! {
...
+        impl Eq for sockaddr_alg {}
+
+        impl ::fmt::Debug for sockaddr_alg {
+            fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                f.debug_struct("sockaddr_alg")
+                    .field("salg_family", &self.salg_family)
+                    .field("salg_type", &self.salg_type)
+                    .field("salg_feat", &self.salg_feat)
+                    .field("salg_mask", &self.salg_mask)
+//                    .field("salg_name", &self.salg_name)
```
gnzlbg: this and anothr field below are also commented, would you mind fixing those too?

src/unix/notbsd/linux/mod.rs
```diff
@@ -748,7 +748,7 @@ cfg_if! {
         impl ::fmt::Debug for af_alg_iv {
             fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
                 f.debug_struct("af_alg_iv")
-                    // .field("iv", self.as_slice())
+                    .field("iv", &self.as_slice().iter().collect::<Vec<_>>())
```
gnzlbg: I think that &self.as_slice() should be enough, no need to call iter and collect (which would allocate memory).

glebpom: Unfortunately, it doesn't compile with error `doesn't have a size known at compile-time`

PR #1263, opened 2019-02-18
Test that targets without a shipping libcore / libstd build properly

gnzlbg:
This PR tests that `libc` builds on targets that currently don't ship a `libcore`/`libstd` by cross-compiling to them from Linux and MacOSX using `Xargo`.

This fixes the build on DragonflyBSD (superseeds #1237), Haiku, bitrig, and OpenBSD (superseeds #1259). 

cc @asomers @strangelittlemonkey @semarie


ci/build.sh
```diff
@@ -161,5 +168,40 @@ case "${OS}" in
...
+x86_64-pc-windows-gnu \
+x86_64-pc-windows-msvc
+x86_64-unknown-bitrig \
+x86_64-unknown-haiku \
+x86_64-unknown-openbsd
+"
+
+if [ "${RUST}" = "nightly" ]; then
+    rustup component add rust-src || true
+    cargo install xargo || true
```
asomers: Why the `|| true`?  Won't that just lead to confusing error messages, for example when the internet isn't available?

gnzlbg:
> So with this PR, Travis will cross compile OSes like Dragonfly from both Linux _and_ OSX? Isn't that redundant?

For the targets implemented in this PR, yes. I'm unsure if any of these tier2/tier3 targets can only be targeted by some toolchains, but AFAIK not all toolchains can cross-compile to all targets, e.g., the x86_64-apple-darwin can cross-compile to `aarch64-apple-ios` but the x86_64-unknown-linux-gnu toolchain cannot (_I think_, I haven't been able to cross-compile to that).

> What coverage does libc gain by cross-compiling from two different hosts?

None I think, it was just easier CI wise to enable this, but I guess I should just enable this for Linux, and if we need something more in the future, then we'll enable it then.

ci/build.sh
gnzlbg: Without any of these the first `xargo` invocation right after will fail hard, so the error won't be far away (not as far as, e.g., when running tests under qemu inside a Docker container). 

OTOH if you have `xargo` installed, `cargo install xargo` errors, the same applies to `rust-src`, so this is the easiest way of supporting that, which allows you to run this locally without much issues.

strangelittlemonkey:
While this builds, it still doesn't pass the tests that were failing in my PR. Would it not make sense to merge my PR and then rebase this on top of it?

gnzlbg:
> Would it not make sense to merge my PR and then rebase this on top of it?

Yes, this only fix the "build" part of it, it does not run the `libc-test`s on DragonflyBSD.

ci/build.sh
gnzlbg: So i've removed the `|| true`, making this more reliable.

gnzlbg:
@asomers I've addresses the issues, thoughts?

asomers:
> @asomers I've addresses the issues, thoughts?

Honestly, no.  This is fairly far outside of my area of expertise.

gnzlbg:
cc @alexcrichton could you review this ?

gnzlbg:
So i've removed the `CMSG_` APIs from bitrig since they never compiled correctly, and one would need to add platform-specific code for the platform that must be manually verified. 

I'll merge this as soon as CI is green, cc @Susurrus if you want to continue the work with the `extra_traits` feature, after this PR is merged it should be pretty safe to do so without causing libc to stop compiling in some of the supported platforms.

alexcrichton:
All looks reasonable to me!

gnzlbg:
There was some trouble with building the documentation of some of the targets that are built with `xargo`, but now that's all fixed. I initially used `xargo` to build the docs of all targets, but that made the documentation bot take over 50 min. The last commit uses `cargo` to build the docs, and if that fails, then it tries `xargo`, dialing the doc bot back to under 15 min. 

That's a more general approach that we could use in the "build-only jobs", to avoid having to keep track of which targets can be built with cargo, and which one need cargo, but we can always do that later.

@bors: r+

PR #1265, opened 2019-02-21
Fix ALG_OP_* types

glebpom:
According to https://elixir.bootlin.com/linux/v4.9/source/crypto/af_alg.c#L462 ALG_OP_* constants should be `u32`

### DIFF
--- a/src/unix/notbsd/android/mod.rs
+++ b/src/unix/notbsd/android/mod.rs
@@ -1787,8 +1787,8 @@ pub const ALG_SET_OP: ::c_int = 3;
 pub const ALG_SET_AEAD_ASSOCLEN: ::c_int = 4;
 pub const ALG_SET_AEAD_AUTHSIZE: ::c_int = 5;
 
-pub const ALG_OP_DECRYPT: ::c_int = 0;
-pub const ALG_OP_ENCRYPT: ::c_int = 1;
+pub const ALG_OP_DECRYPT: u32 = 0;
+pub const ALG_OP_ENCRYPT: u32 = 1;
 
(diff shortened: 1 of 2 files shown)


src/unix/notbsd/android/mod.rs
```diff
@@ -1787,8 +1787,8 @@ pub const ALG_SET_OP: ::c_int = 3;
 pub const ALG_SET_AEAD_ASSOCLEN: ::c_int = 4;
 pub const ALG_SET_AEAD_AUTHSIZE: ::c_int = 5;
 
-pub const ALG_OP_DECRYPT: ::c_int = 0;
-pub const ALG_OP_ENCRYPT: ::c_int = 1;
+pub const ALG_OP_DECRYPT: u32 = 0;
```
asomers: Shouldn't this be `::c_uint`?

asomers:
Actually, I think signed ints are right.  See where they're used:
https://elixir.bootlin.com/linux/v4.9/source/include/crypto/if_alg.h#L47 .

glebpom:
I was thinking that there is no guarantee that `c_int` is always 32-bits. But from what I see in the `libc` code, this is guaranteed. If this is correct, we can just close this PR

gnzlbg:
> I was thinking that there is no guarantee that `c_int` is always 32-bits

There is no general guarantee that holds for all platforms, but every platform has to define their c_int type. This is something that platforms cannot easily change after the fact without breaking binary compatibility with pretty much all code, and this is also something that we test for Android, Linux, and a bunch of other platforms on CI.

PR #1270, opened 2019-02-21
Do not import the std library with use_std

gnzlbg:


gnzlbg:
@alexcrichton there is one single use of `use_std` in the library remaining, and that is here:

https://github.com/rust-lang/libc/blob/master/src/unix/mod.rs#L306

I would actually expect that whole cfg_if to be actually implemented in the `build.rs` instead of using `#[link(name)]` on empty extern blocks, but 🤷‍♂️ 

AFAICT, that `use_std` branch can be removed, because if libstd does something, it better be the exact same thing we do here, which should mean that libc requires linking against the same libraries that libstd does.

alexcrichton:
There's a ton of historical gotchas to libc as it's been around since the beginning of time. That block exists because libc not only existed before build scripts existed but also when used as a submodule it didn't have the same build script. Times change though! It can probably move to a build script nowadays.

Linkage is sort of hard to change at this layer and not break things on all platforms. I suspect that musl may have the biggest wrench with its static linkage by default (libc-the-crate on crates.io pulls in libc-the-C-library through libc-built-in-the-sysroot-for-libstd for example)

PR #1271, opened 2019-02-22
Implement more extra_traits

Susurrus:
Finishing the implementation of rust-lang/rust#57715 now that all platforms are tested in CI. I plan to expand this CI to all platforms that need this treatment (solarish, empscripten, freebsd, and netbsd), but thought I'd get the part I've already started running in CI since I can't test this changes locally.

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -482,8 +482,6 @@ s! {
 }
 
 s_no_extra_traits!{
-    // FIXME: https://github.com/rust-lang/libc/issues/1243
-    #[allow(missing_debug_implementations)]
     #[cfg_attr(libc_packedN, repr(packed(4)))]
     pub struct kevent {
         pub ident: ::uintptr_t,
@@ -494,8 +492,6 @@ s_no_extra_traits!{
         pub udata: *mut ::c_void,
     }
 
-    // FIXME: https://github.com/rust-lang/libc/issues/1243
-    #[allow(missing_debug_implementations)]
     #[cfg_attr(libc_packedN, repr(packed(4)))]
     pub struct semid_ds {
         // Note the manpage shows different types than the system header.
@@ -509,8 +505,6 @@ s_no_extra_traits!{
         pub sem_pad3: [::int32_t; 4],
     }
 
-    // FIXME: https://github.com/rust-lang/libc/issues/1243
-    #[allow(missing_debug_implementations)]
     #[cfg_attr(libc_packedN, repr(packed(4)))]
     pub struct shmid_ds {
         pub shm_perm: ipc_perm,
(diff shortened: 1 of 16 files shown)


gnzlbg:
Cool! Thanks! Ping me when CI is green!

Susurrus:
So 1.24 is the cutoff for using `unsafe` around `packed` struct instances, but older versions are tested in CI. I think we need to do the same macro trick that you used elsewhere to get this code compiling on all CI targets.

gnzlbg:
The problem is that, if creating a reference to a packed struct field creates an unaligned reference, then the behavior is undefined. 

That's the case irrespectively of whether one uses an unsafe block post Rust 1.24, or no unsafe block in older Rust versions.

gnzlbg:
You could just add an attribute to those methods to silence the unused unsafe warning ("`#[allow(unused_unsafe)]`", but that does not make performing these operations less undefined.

gnzlbg:
Instead of using references, you are probably better off copying the struct fields out of the struct into the stack frame, and then passing references to that.

cc @Centril @RalfJung

gnzlbg:
Alternatively, one can take a reference to the struct, cast it to a pointer, and use pointer offsets to read from the struct fields. One'd need to compute the offsets manually, but the structs are `repr(C)`, so this is possible (the rules are in the C standard). If these structs change, then all offset would need to be manually updated though.

RalfJung:
> you are probably better off copying the struct fields out of the struct into the stack frame, and then passing those (or references to those) to the formatter.

Yes, that is the best approach IMO. You need a reference to call `.field`, so it needs to be aligned, so you need to make a copy.

src/unix/bsd/apple/mod.rs
```diff
@@ -640,6 +634,133 @@ cfg_if! {
...
+                    && self.fflags == other.fflags
+                    && self.data == other.data
+                    && self.udata == other.udata
+            }
+        }
+        impl Eq for kevent {}
+        impl ::fmt::Debug for kevent {
+            fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                f.debug_struct("kevent")
+                    .field("ident", unsafe { &self.ident })
```
gnzlbg: So instead of 

```rust
. field("ident", unsafe { &self.ident })
```

I think the simplest solution would be to write:

```rust
// unnecessary copy but ¯\_(ツ)_/¯ 
let ident = self.ident;  
// ... 
. field("ident", &ident)
```

I'm not sure if all field types are copy (we do have large arrays), but I hope so.

RalfJung: > unnecessary copy but ¯\_(ツ)_/¯

There's nothing unnecessary about it. `.field` will perform an aligned load from this address, we need to put stuff into an aligned address.

Susurrus:
Remaining failures are not due to this code change and for targets that are allowed to fail. I think this is all good now.

PR #1277, opened 2019-02-23
CUDA syscalls

denzp:
This introduces basic CUDA (`nvptx64-nvidia-cuda` target) syscalls.

I defined only `c_char` and `size_t` types since they were necessary for the syscalls.

### DIFF
--- a/src/cuda.rs
+++ b/src/cuda.rs
@@ -1,0 +1,98 @@
+pub type c_char = u8;
+pub type size_t = usize;
+
+cfg_if! {
+    if #[cfg(libc_core_cvoid)] {
+        pub use ::ffi::c_void;
+    } else {
+        // Use repr(u8) as LLVM expects `void*` to be the same as `i8*` to help
+        // enable more optimization opportunities around it recognizing things
+        // like malloc/free.
+        #[repr(u8)]
+        #[allow(missing_copy_implementations)]
+        #[allow(missing_debug_implementations)]
+        pub enum c_void {
+            // Two dummy variants so the #[repr] attribute can be used.
+            #[doc(hidden)]
+            __variant1,
+            #[doc(hidden)]
+            __variant2,
+        }
+    }
+}
+
+extern "C" {
+    /// Syscall to be used whenever the *assert expression produces a `false` value*.
+    ///
+    /// Syscall arguments:
+    /// * `message`: The pointer to the string that should be output.
+    /// * `file`: The pointer to the file name string associated with the assert.
+    /// * `line`: The line number associated with the assert.
+    /// * `function`: The pointer to the function name string associated with the assert.
+    /// * `char_size`: The size in bytes of the characters contained in the provided strings. The only supported character size is 1.
+    ///
+    /// Source:
+    /// [PTX Interoperability](https://docs.nvidia.com/cuda/ptx-writers-guide-to-interoperability/index.html#system-calls).
+    pub fn __assertfail(
+        message: *const c_char,
+        file: *const c_char,
+        line: i32,
+        function: *const c_char,
+        char_size: size_t,
+    );
+
+    /// Print formatted output from a kernel to a host-side output stream.
+    ///
+    /// Syscall arguments:
+    /// * `status`: The status value that is returned by `vprintf`.
+    /// * `format`: A pointer to the format specifier input (uses common `printf` format).
+    /// * `valist`: A pointer to the valist input.
+    ///
+    /// ```
+    /// #[repr(C)]
+    /// struct PrintArgs(f32, f32, f32, i32);
+    ///
+    /// vprintf(
+    ///     "int(%f + %f) = int(%f) = %d\n".as_ptr(),
+    ///     transmute(&PrintArgs(a, b, a + b, (a + b) as i32)),
+    /// );
+    /// ```
+    ///
+    /// Sources:
+    /// [Programming Guide](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#formatted-output),
+    /// [PTX Interoperability](https://docs.nvidia.com/cuda/ptx-writers-guide-to-interoperability/index.html#system-calls).
+    pub fn vprintf(format: *const c_char, valist: *const c_void) -> i32;
+
+    /// Allocate memory dynamically from a fixed-size heap in global memory.
+    ///
+    /// The CUDA in-kernel `malloc()` function allocates at least `size` bytes
+    /// from the device heap and returns a pointer to the allocated memory
+    /// or `NULL` if insufficient memory exists to fulfill the request.
+    ///
+    /// The returned pointer is guaranteed to be aligned to a 16-byte boundary.
+    ///
+    /// The memory allocated by a given CUDA thread via `malloc()` remains allocated
+    /// for the lifetime of the CUDA context, or until it is explicitly released
+    /// by a call to `free()`. It can be used by any other CUDA threads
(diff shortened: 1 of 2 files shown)


src/cuda.rs
```diff
@@ -0,0 +1,98 @@
...
+    ///
+    /// The memory allocated by a given CUDA thread via `malloc()` remains allocated
+    /// for the lifetime of the CUDA context, or until it is explicitly released
+    /// by a call to `free()`. It can be used by any other CUDA threads
+    /// even from subsequent kernel launches.
+    ///
+    /// Sources:
+    /// [Programming Guide](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#dynamic-global-memory-allocation-and-operations),
+    /// [PTX Interoperability](https://docs.nvidia.com/cuda/ptx-writers-guide-to-interoperability/index.html#system-calls).
+    pub fn malloc(size: size_t) -> *mut c_void;
```
gnzlbg: In clang this is `__attribute__((malloc,nothrow)), the returning pointer should be LLVM `noalias`.

src/cuda.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// the call to `free()` is ignored.
+    ///
+    /// Any CUDA thread may free memory allocated by another thread, but care should be taken
+    /// to ensure that the same pointer is not freed more than once. Repeated calls to `free()`
+    /// with the same `ptr` has undefined behavior.
+    ///
+    /// Sources:
+    /// [Programming Guide](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#dynamic-global-memory-allocation-and-operations),
+    /// [PTX Interoperability](https://docs.nvidia.com/cuda/ptx-writers-guide-to-interoperability/index.html#system-calls).
+    pub fn free(ptr: *mut c_void);
```
gnzlbg: This should also be `nothrow`. I wonder whether this is the right place to hook these, or whether we should be doing that through the `#[allocator]`/`GlobalAllocator` infrastructure.

src/cuda.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// * `line`: The line number associated with the assert.
+    /// * `function`: The pointer to the function name string associated with the assert.
+    /// * `char_size`: The size in bytes of the characters contained in the provided strings. The only supported character size is 1.
+    ///
+    /// Source:
+    /// [PTX Interoperability](https://docs.nvidia.com/cuda/ptx-writers-guide-to-interoperability/index.html#system-calls).
+    pub fn __assertfail(
+        message: *const c_char,
+        file: *const c_char,
+        line: i32,
```
gnzlbg: line is `unsigned` in clang, which would map to `c_uint`, but the syscall docs state that this should indeed be a 32-bit integer. Maybe use either `u32` here, or define a `c_uint = u32` ?

src/cuda.rs
```diff
@@ -0,0 +1,98 @@
...
+        function: *const c_char,
+        char_size: size_t,
```
gnzlbg: While the syscall API takes `char_size`, e.g., the API exposed by clang does not (this is always known at compile-time).

src/cuda.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// vprintf(
+    ///     "int(%f + %f) = int(%f) = %d\n".as_ptr(),
+    ///     transmute(&PrintArgs(a, b, a + b, (a + b) as i32)),
+    /// );
+    /// ```
+    ///
+    /// Sources:
+    /// [Programming Guide](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#formatted-output),
+    /// [PTX Interoperability](https://docs.nvidia.com/cuda/ptx-writers-guide-to-interoperability/index.html#system-calls).
+    pub fn vprintf(format: *const c_char, valist: *const c_void) -> i32;
```
gnzlbg: In clang `valist` is also a `*const c_char` and the return type is a `c_int`.

gnzlbg:
The alternative would be to expose these in `core::arch::nvptx`, they do lower to intrinsics, so that might be a better fit. For the memory allocation routines, `core::arch::wasm32` does expose memory allocation intrinsics, so there is some precedent of that in other architectures. 

If we expose these in `core::arch`, instead of using C types, we would use Rust types on the API, except for the `void*` where we could directly use `core::ffi::c_void` since we don't have to support ancient Rust toolchains there. So maybe these would be a better fit for that.

But I'm not sure. 🤷‍♂️

src/cuda.rs
denzp: Is there a way to accomplish this? I haven't found a way to specify those attributes.

src/cuda.rs
denzp: Surprisingly, the documentation states incorrect information while Clang has a correct implementation! I tried to call it with `-1` as `line`, and it printed it as `uint(-1)`!

```
src/main.rs:4294967295: (null): block: [1,0,0], thread: [1,0,0] Assertion `Panicked` failed.
```

src/cuda.rs
denzp: Anyway, we need to emit the correct signature. It's up to higher level crates to provide a convenient API without specifying the `char_size`, I think.

src/cuda.rs
denzp: Is there any difference from FFI or safety point of view between `*const c_char` and `*const c_void`?

From the logic point of view, having `*const c_void` is more correct because the second argument can be of any type, e.g. simple case:
``` rust
vprintf("c_int: %d".as_ptr(), transmute(&10));
```

src/cuda.rs
gnzlbg: so there is a supper unstable `#[allocator]` attribute, but we probably want to expose that only through the global allocator.

src/cuda.rs
gnzlbg: Sure. I guess my point was that maybe this API is too low-level even for `libc`. 

Me as a libc maintainer and myself as a Rust CUDA person are two different people :P

src/cuda.rs
gnzlbg: So I don't know actually. I kind of wish NVIDIA had a CUDA header for their "kernel-only" APIs that we could just follow here.

denzp:
I must say, I like the idea of moving these syscalls to `core::arch::nvptx`! They are indeed quite low-level for normal usage by end-users.

Assume, missing `malloc` and `nothrow` attributes problem should be addressed anyway, could you please suggest how it can be achieved?

gnzlbg:
So for malloc we would just use the `#[allocator]` attribute, for the others that are nothrow I think there is a `#[rustc_nothrow]` attribute that we can use. In any case, `core::arch` seems like a better fit for these to me as well, so maybe we should close this and you can pretty much copy this PR 1:1 there (we don't have to get all attributes perfect for the initial commit there).

src/cuda.rs
denzp: I only found a meention of `printf` declaration in original CUDA headers:
``` c
__host__ __device__ __device_builtin__ __cudart_builtin__ int __cdecl printf(const char*, ...);
```

At least for now it seems to be safe to use `*const c_void` because LLVM emits same PTX with both types:
```
.extern .func  (.param .b32 func_retval0) vprintf
(
	.param .b64 vprintf_param_0,
	.param .b64 vprintf_param_1
)
```

denzp:
Closing in favor of [stdsimd#699](https://github.com/rust-lang-nursery/stdsimd/pull/699).

PR #1279, opened 2019-02-24
Clean up Redox, CloudABI, DragonflyBSD, and NetBSD in libc-test/build.rs

gnzlbg:


gnzlbg:
cc @EdSchouten - this factors some parts of the libc-test system for cloudabi. AFAIK, this might break some things, but until last week, the cloudabi bindings for libc did not even compile (I don't know for how long this has been the case). We have native CI jobs for FreeBSD using Cirrus-CI (see `libc/.cirrus.yml`). It should be possible to add another job there that uses x86_64 FreeBSD 11, which should come with `cloudabi-run` (or one should at least be able to install it). Then doing something like: 

```rust
CARGO_TARGET_X86_64_UNKNOWN_CLOUDABI_RUNNER=cloudabi-run cargo test
```

should be "close enough" to get the libc testsuite running. While the testsuite will try to use file-descriptors for output, AFAIK the only important thing here is whether each test process returns success or failure. So while getting the file-descriptors working properly would be nice-to-have for debugging CI failures, it might not be necessary to do so for an MVP.

gnzlbg:
cc @jackpot51 we recently started testing that libc builds for redox at all, which it wasn't doing, so this PR might not actually break anything that was working before, but it refactors a bit of the code intended to run `libc-test`s under redox, and without any sort of CI for this it is impossible to do the refactor with 100% confidence. This does not affect the libc API, only libc's own testing system. So it might well be that if you are running libc tests somewhere in your CI, after this PR those might error (I don't know how that could happen, since AFAICT libc was not building properly for redox before, but just in case).

gnzlbg:
cc @strangelittlemonkey I've cleaned up the build.rs for dragonflybsd as well. It should be a non-functional change, but it would be nice to try to slim everything that's skipped in there or unnecessary includes down as much as possible. There might be broken ABIs that are currently not tested because someone previously had skipped it for a different operating system, but used the more general bsdlike flag instead.

jackpot51:
@gnzlbg the changes specific to Redox are fine with me

PR #1280, opened 2019-02-24
openbsd: glob_t recently changed to conform posix (use size_t)

semarie:
on openbsd, upcoming stable version have changed glob_t definition to follow posix definition (using `size_t` instead of `int` on some fields)

the change preserves bitrig definition

r? @gnzlbg

### DIFF
--- a/src/unix/bsd/netbsdlike/openbsdlike/bitrig/mod.rs
+++ b/src/unix/bsd/netbsdlike/openbsdlike/bitrig/mod.rs
@@ -1,6 +1,21 @@
 pub type c_char = i8;
 
 s! {
+    pub struct glob_t {
+        pub gl_pathc:   ::c_int,
+        pub gl_matchc:  ::c_int,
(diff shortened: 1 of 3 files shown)


gnzlbg:
@semarie i'm slowly cleaning up libc-test/build.rs to make it 100% clear what headers impact which targets, and what is tested and what is skipped on each target. 

Since we don't run any tests on OpenBSD ? Would you be able to give doing that a try ? (see how it is done for Apple, windows, etc.). Basically it requires extracting copying the whole main function into a `test_openbsd` function, and then disable everything from the main function that was used only by OpenBSD. Once that works, you can try disabling everything that OpenBSD does not need from its test function. And if you are feeling brave, trying to enable all the tests (there are some utilities like `cfg.verbose_skipped(true)` that will tell you which stuff is actually skipped, so that you can remove all the code to skip stuff that's actually not encountered. I'm on Discord and Zulip in case you want to give that a try, and have questions.

PR #1281, opened 2019-02-24
cleanup libc-test for OpenBSD

semarie:
here a cleanup for libc-test for OpenBSD

Some elements (compat for old and now unsupported OpenBSD versions) could be removed, but I think it is better to address them after this PR is merged.

the testsuite for OpenBSD still pass with it (well, with #1280)

r? @gnzlbg


gnzlbg:
Awesome, thank you!

@bors: r+

PR #1285, opened 2019-02-24
[WIP] Import C types from the cty crate

gnzlbg:
So this is more or less how solving https://github.com/japaric/cty/issues/13 and https://github.com/japaric/cty/issues/14 would look like. I've kept the whole history of the cty crate.

I think that I found some bugs in both libc and cty in the process.

cc @japaric @Amanieu @retep998 @alexcrichton @denzp


cty/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+                    cfg_if! {
+                        if #[cfg(any(
+                            target_arch = "powerpc",
+                            target_arch = "powerpc64",
+                            target_arch = "s390x",
+                        ))] {
+                            // FIXME: ??
+                            pub type wchar_t = i32;
+                        } else {
+                            pub type wchar_t = u32;
```
retep998: Windows also supports `aarch64` and `arm` where `wchar_t` is still `u16`.

gnzlbg: I think `wchar_t` is just too OS and target-dependent to try to shoe-horn it here.

alexcrichton:
I would personally pretty strongly discourage this. The `libc` crate is a fundamental crate in the Rust ecosystem which is designed by RFC and led by the libs team. The same does not apply to the `cty` crate. 

Duplication may be unfortunate but it doesn't matter except for the `c_void` type really, otherwise all other types are just typedefs which can already unify across the two crates.

gnzlbg:
After doing this the feeling I got is that it makes more sense to keep both of these separate for now (not change libc to re-export cty), though I'd prefer to host the cty crate here so that we can use libc's CI to make sure its types are correct, both by using `ctest` to ensure that in the same platforms that `libc`'s support, and maybe also by adding a build test for all targets that makes sure that both crates implement the same API on all targets that libc supports. 

I don't think any change here is going to be RFC worthy. The Unsafe Code Guidelines is already documenting that some Rust types exactly match some C types (e.g. `size_t`, `intptr_t`, `ptrdiff_t`, `int16_t`, etc.), and all of that will be RFC'ed sometime this year. If that gets merged, it makes now more sense to me to just add these types to `core::ffi` directly.

alexcrichton:
In terms of hosting I don't think it matters too much where it lives, but I think structurally these should continue to be separate crates for the time being (although an RFC could change that)

gnzlbg:
So i'm closing this to submit a different PR that does that, don't want to lose this branch because it discovered a couple of bugs.

PR #1286, opened 2019-02-26
[WIP] Integrate the cty crate into libc's CI

gnzlbg:
So this the successor to #1285 (cc @japaric @Amanieu @alexcrichton ).

It imports the cty crate into the libc repository preserving its commit history, and integrates it with libc's CI. That basically amounts to running the same tests that libc runs on the cty crates type for those targets that we support at run-time (type size, alignment, sign, etc. ).

This PR also adds a crate that tests at compile-time that all types between the libc and the cty crates match except for `c_void`. As @alexcrichton mentioned, `c_void` matches if both use `core::ffi::c_void`, and differs otherwise because each crate defines its own in old compiler versions.

WIP: while doing #1285 I noticed that the types differed in some targets because of bugs. I will manually disable the tests in the archs in which they fail (adding the appropriate comments and filling issues), so that we can decide later on a 1:1 case basis if we want to fix them, and how (we can probably fix cty bugs by just bumping the semver major version, but for libc bugs that's a different story). This PR should essentially be a non-functional change for both crates, except that `cty` now also supports older Rust targets without `core::ffi::c_void`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ travis-ci = { repository = "rust-lang/libc" }
 appveyor = { repository = "rust-lang/libc", project_name = "rust-lang-libs/libc" }
 
 [dependencies]
+cty = { version = "0.2", path = "cty" }
 rustc-std-workspace-core = { version = "1.0.0", optional = true }
 
 [features]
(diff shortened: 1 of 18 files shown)


alexcrichton:
@gnzlbg it looks like this may be an accident, but it looks like `libc` depends on `cty`?

gnzlbg:
Yes that's an accident. I tried to rebase the old branch to remove the commits that added that, but somehow git rebasing does not work across the commit that merged one repository into the other. 

I have this on my queue, but haven't had time yet to look at it again.

alexcrichton:
Ok just wanted to confirm, no worries! Apart from that looks good to me!

JohnTitor:
Closing as the author is inactive, if someone is interested in this change, let us know!

PR #1288, opened 2019-02-28
Add arm-uclibc definitions

red75prime:
I used `buildroot-2017.05` to build armv7 cross compiler, `uclibc-1.0.24` and `libc-test`. I've executed `libc-test` on armv7-based SoC, using modified `libc-test` to account for the environment. Modified libc-test is not included in this PR because it is too dependent on cross compilation and execution environment.


gnzlbg:
> Modified libc-test is not included in this PR because it is too dependent on cross compilation and execution environment.

Do you have a docker container with those ? 

We are currently not even testing that the `uclib` targets build, which means that if we break the build for them by accident, we won't notice that. If you need more stability for `uclib` targets, you should consider contributing a target specification file to libc, so that we can at least build it using `xargo` to make sure it always compiles (see `ci/build.sh`). If you need ABI stability, a docker container that set ups the tools required, and running the libc tests under qemu, is probably the only way forward.

gnzlbg:
You need to fix the style, otherwise, this LGTM.

red75prime:
> Do you have a docker container with those ?

Currently, no. I need to modify container build script to include all the changes I made. I'll work on that.

red75prime:
@gnzlbg Do you want me to include testing harness for arm-uclibc into this PR? It will take some time.

gnzlbg:
> Do you want me to include testing harness for arm-uclibc into this PR?

No, I was just wondering whether that would be possible, to prevent this from breaking again, we can always do that later! Also note, that being able to build for the target, which does not require linking anything, would already be enough - supporting libc-test could be done as a second step afterwards. 

FYI I just don't get pinged when Travis turns green (feel free to ping me next time as soon as it happens!). 

@bors: r+

red75prime:
> 💔 Test failed - [checks-travis](https://travis-ci.com/rust-lang/libc/builds/102803105)

Errors seem to be unrelated to this PR.

gnzlbg:
Thank you a lot ! Rust could really use better `uclib` support!

PR #1291, opened 2019-03-02
adjust PTHREAD_STACK_MIN to current value on OpenBSD

semarie:
adjust `PTHREAD_STACK_MIN` to current value on OpenBSD, and while here, generate the right file in test_openbsd()

I should say that I am a bit disappointed. If I didn't miss anything, the related change for `PTHREAD_STACK_MIN` is near to one year old. And clearly with the name error in `test_openbsd()` something wasn't clean.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -1088,7 +1088,7 @@ fn test_openbsd(target: &str) {
         (struct_ == "siginfo_t" && field == "si_addr")
     });
 
-    cfg.generate("../src/lib.rs", "linux_fcntl.rs");
+    cfg.generate("../src/lib.rs", "main.rs");
 }
 
 fn test_windows(target: &str) {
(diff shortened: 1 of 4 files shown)


gnzlbg:
> And clearly with the name error in test_openbsd() something wasn't clean.

I just checked the other tests and at least the copy past error here happened only in `test_openbsd`. I shouldn't have missed that in review. Sorry.

> I didn't miss anything, the related change for PTHREAD_STACK_MIN is near to one year old.

Damn. We really should get the libc-test running under qemu for openbsd. We used to do that for FreeBSD so it can't be that hard.

semarie:
> Damn. We really should get the libc-test running under qemu for openbsd. We used to do that for FreeBSD so it can't be that hard.

For FreeBSD, the method was to crosscompile the test binaries from Linux, and next run them under qemu. Trying to compile under qemu is very slow.

gnzlbg:
Can't we cross-compile libc-test using xargo ?

gnzlbg:
Even if we can't, note that the slow part of building libc-test is running the `build.rs` which generates code. This step builds ctest, syntex, and 100 dependencies, but we can do this on Linux. This step generates two files, the Rust test, and the C helper library. If we can't cross-compile these two files, compiling them under qemu is probably fast enough. The Rust test have `libc` as a dependency, but we can cross-compile that from Linux to OpenBSD using `Xargo` (CI already does this in the build tests).

PR #1292, opened 2019-03-02
add HermitCore support even if it doesn't have a UNIX interface

stlankes:
Currently, we redefine the interface between the HermitCore kernel (https://hermitcore.org) and Rust’s standard library. In the future, it will not depend on a POSIX-compatible C library. Consequently, we add the support of HermitCore if the "unix" environment isn’t set. It will be great to integrate this patch because it would simplify our development. The classical interface is still supported and part of the subdirectory "unix".


src/hermit/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+pub type wint_t = u32;
+pub type wctype_t = i64;
+
+pub type regoff_t = size_t;
+pub type off_t = c_long;
+
+cfg_if! {
+    if #[cfg(target_arch = "aarch64")] {
+        mod aarch64;
+        pub use self::aarch64::*;
```
gnzlbg: AFAICt, `c_char` and `w_char` are defined in this module, and in the `aarch64` module, so when the code imports them here, they are defined twice, and this won't compile.

src/hermit/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+    } else if #[cfg(target_arch = "x86_64")] {
+        mod x86_64;
+        pub use self::x86_64::*;
```
gnzlbg: Same here, this won't compile AFAICT.

stlankes:
You are right. I oversaw the redefinition and the typo.

Until now, HermitCore war missing in `ci/build.sh`. Do you have an tutorial, how I could add HermitCore? I know xargo and use it for our project.

gnzlbg:
There is not tutorial, but that you are already using xargo is great. I assume that you pass `xargo` the target specification files (e.g. `x86_64-...-hermit.json` and `aarch64-...-hermit.json`), is that correct? 

If so, it should be enough to add the targets here: https://github.com/rust-lang/libc/blob/master/ci/build.sh#L169

And to copy the json files in the `libc/ci` directory. 

You probably want to try that locally. For that the best you can do is put them as the first two targets of that environment variable, and comment this other loop (https://github.com/rust-lang/libc/blob/master/ci/build.sh#L163) so that only the "no core" targets get tested.

src/unix/hermit/mod.rs
```diff
@@ -19,7 +19,9 @@
 pub type c_long = i64;
 pub type c_ulong = u64;
 
+#[allow(unused)]
```
gnzlbg: So this is weird, the only way in which these could be unused is if they are not exported by the library API, right?

src/unix/hermit/mod.rs
```diff
@@ -19,10 +19,8 @@
 pub type c_long = i64;
 pub type c_ulong = u64;
 
-#[allow(unused)]
-pub type uid_t = u16;
-#[allow(unused)]
-pub type gid_t = u16;
+//pub type uid_t = u16;
```
gnzlbg: My point here was that if rustc says that these are unused, is because you can't access them via `libc::uid_t`. This might hint that the contents of the `hermit` module are not properly exported by the root crate.

stlankes: You are right. At export the module in [src/unix/mod.rs](https://github.com/rust-lang/libc/blob/ff97bdb074782fc5bbcae88c405d408672726f3d/src/unix/mod.rs#L1146-L1148). For me isn't clear, why I got these warnings.

gnzlbg: These are also defined in the general `unix/mod.rs`: https://github.com/rust-lang/libc/blob/ff97bdb074782fc5bbcae88c405d408672726f3d/src/unix/mod.rs#L38

So they are probably being shadowed, and therefore not re-exported (hence the `unused` warning - these are dead code).

gnzlbg: Basically you'd need to move `gid_t` and `uid_t` from `unix/mod.rs` to all other sub-modules of the `unix/` directory.

stlankes: Ok, I oversaw this. `u32` is also fine for me. I will delete these types and use `u32` in our lib.

PR #1293, opened 2019-03-02
x86_64-uclibc-l4re: fix syntax error + linter issues

humenda:
-   fix syntax error in expand_align macro
-   fix errors due to doubly defined symbols / functions for
    x86_64-uclibc


gnzlbg:
LGTM!

@bors: r+

---

Unrelated question, could you contribute the target-specification json file that you are using with xargo to build whatever application you are building ? If you could add it to the libc/ci directory we could build libc for this target, and make sure that at least compilation does not break for the target. Thanks!

gnzlbg:
There are some style errors that should be fixed.

humenda:
Thanks, fixed.

gnzlbg:
FYI i've added a test that checks that this target properly builds on CI here: #1296

PR #1294, opened 2019-03-03
Add vm_size_t

GuillaumeGomez:


gnzlbg:
@GuillaumeGomez Thank you for the PR, why do you need this? I would prefer to stop exposing mach headers in libc, deprecate what we expose, and recommend people to use the `mach` crate instead.

gnzlbg:
Ah wait, nvm, this is for all *BSDs, for some reason I thought this was only for macosx. 

@bors: r+

GuillaumeGomez:
Just like you said. ;)

semarie:
this type doesn't exist on OpenBSD: `error: use of undeclared identifier 'vm_size_t'`

PR #1297, opened 2019-03-03
Upgrade bunch of EOL docker images to Ubuntu 18.04 

mati865:
When upgrading other Dockerfiles (not modified in this PR) they fail with different errors:
https://travis-ci.com/mati865/libc/jobs/181623075
https://travis-ci.com/mati865/libc/jobs/181623059
https://travis-ci.com/mati865/libc/jobs/181623067

I think it's because of these 2 commits in glibc:
https://sourceware.org/git/?p=glibc.git;a=commit;h=b05cb613e7ae507680c8d914b3e788eea782c0d6
https://sourceware.org/git/?p=glibc.git;a=commit;h=2dba5ce7b8115d6a2789bf279892263621088e74

I have no idea how to fix them so leaving it for somebody who knows what to do.

### DIFF
--- a/src/unix/notbsd/linux/other/mod.rs
+++ b/src/unix/notbsd/linux/other/mod.rs
@@ -670,8 +670,7 @@ pub const NFPROTO_NETDEV: ::c_int = 5;
 
 // linux/netfilter/nf_tables.h
 cfg_if!{
-    if #[cfg(any(target_arch = "arm", target_arch = "powerpc",
-                 target_arch = "powerpc64", target_arch = "aarch64"))] {
+    if #[cfg(any(target_arch = "arm", target_arch = "aarch64"))] {
         pub const NFT_TABLE_MAXNAMELEN: ::c_int = 32;
         pub const NFT_CHAIN_MAXNAMELEN: ::c_int = 32;
         pub const NFT_SET_MAXNAMELEN: ::c_int = 32;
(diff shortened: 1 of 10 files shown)


src/unix/notbsd/linux/other/mod.rs
```diff
@@ -670,8 +670,7 @@ pub const NFPROTO_NETDEV: ::c_int = 5;
 
 // linux/netfilter/nf_tables.h
 cfg_if!{
-    if #[cfg(any(target_arch = "arm", target_arch = "powerpc",
-                 target_arch = "powerpc64", target_arch = "aarch64"))] {
```
gnzlbg: Why this change?

mati865: It's because of the kernel upgrade, see here:
- Ubuntu 17.10 with 4.13.x kernel: https://elixir.bootlin.com/linux/v4.13.16/source/include/uapi/linux/netfilter/nf_tables.h
- Ubuntu 18.04 with 4.15.x kernel: https://elixir.bootlin.com/linux/v4.15.18/source/include/uapi/linux/netfilter/nf_tables.h

mati865:
Probably these 2 could go away once all the images are updated:
- https://github.com/rust-lang/libc/blob/8d94e00051ad6d3e5563ddada294f7c31fd20e47/libc-test/build.rs#L539-L551
- https://github.com/rust-lang/libc/blob/8d94e00051ad6d3e5563ddada294f7c31fd20e47/libc-test/build.rs#L595-L609

In fact this PR modifies API only for `powerpc*` because other files that I changed use pinned `musl` and I found nothing about powerpc in `libc-test/build.rs`.

gnzlbg:
Thank you!

Issue #1304, opened 2019-03-14
Deprecate the fixed-width integer aliases


gnzlbg:
We do guarantee that the fixed-width integer types `{i,u}{8,16,32,64}` are layout compatible with C's `{u,}int{8,16,32,64}_t`.

There is no need to provide type aliases (e.g. `libc::uint64_t`) for these, in the same way that we don't provide a `libc::c_bool` type alias for `bool`.

---

Note that `c_float` and `c_double` aren't necessarily layout compatible with `f32` and `f64`.

PR #1307, opened 2019-03-27
Add intial support for wasm32-unknown-wasi

alexcrichton:
This target is [being proposed][LINK] int he rust-lang/rust repository
and this is intended to get coupled with that proposal. The definitions
here all match the upstream reference-sysroot definitions and the
functions all match the reference sysroot as well. The linkage here is
described more in detail on the Rust PR itself, but in general it's
similar to musl.

Automatic verification has been implemented in the same manner as other
targets, and it's been used locally to develop this PR and catch errors
in the bindings already written (also to help match the evolving sysroot
of wasi). The verification isn't hooked up to CI yet though because
there is no wasi target distributed via rustup just yet, but once that's
done I'll file a follow-up PR to execute verification on CI.

[LINK]: https://github.com/rust-lang/rust/pull/59464

### DIFF
--- a/ci/docker/wasm32-unknown-wasi/Dockerfile
+++ b/ci/docker/wasm32-unknown-wasi/Dockerfile
@@ -1,0 +1,69 @@
+FROM ubuntu:18.04 as reference-sysroot
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    ca-certificates \
+    clang \
+    cmake \
+    curl \
+    g++ \
+    git \
+    libc6-dev \
+    libclang-dev \
+    make \
+    ssh \
+    xz-utils
+
+RUN curl http://releases.llvm.org/8.0.0/clang+llvm-8.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz | tar xJf -
+RUN mv /clang+llvm-8.0.0-x86_64-linux-gnu-ubuntu-18.04 /wasmcc
+
+RUN git clone https://github.com/CraneStation/reference-sysroot-wasi && \
+  cd reference-sysroot-wasi && \
+  git reset --hard d5a609fe63926533e1054e539ba5f2693d51bdf5
+RUN make -C reference-sysroot-wasi install -j $(nproc) WASM_CC=/wasmcc/bin/clang INSTALL_DIR=/wasm-sysroot
+COPY docker/wasm32-unknown-wasi/clang.sh /wasm-sysroot/bin/clang
+
+FROM ubuntu:18.04 as wasmtime
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    ca-certificates \
+    clang \
+    cmake \
+    curl \
+    g++ \
+    git \
+    libclang-dev \
+    make \
+    ssh
+
+RUN curl -sSf https://sh.rustup.rs |  sh -s -- -y
+ENV PATH=/root/.cargo/bin:$PATH
+
+RUN apt-get install -y --no-install-recommends python
+RUN git clone https://github.com/CraneStation/wasmtime-wasi wasmtime && \
+  cd wasmtime && \
+  git reset --hard a7ac05df74759a7536b2b1e30adc6ff4867e36c3
+
+# Install wasmtime in /usr/bin, but make sure to remove rust afterwards because
+# we don't want it conflicting with the main Rust we're using to compile
+# `libc`.
+RUN cargo build --release --manifest-path wasmtime/Cargo.toml
+
+FROM ubuntu:18.04
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    gcc \
+    libc6-dev \
+    libxml2
+
+COPY --from=reference-sysroot /wasmcc /wasmcc/
+COPY --from=reference-sysroot /wasm-sysroot/ /wasm-sysroot/
+COPY --from=wasmtime /wasmtime/target/release/wasmtime /usr/bin/
+
+ENV CARGO_TARGET_WASM32_UNKNOWN_WASI_RUNNER=wasmtime \
+  CARGO_TARGET_WASM32_UNKNOWN_WASI_LINKER=/wasm-sysroot/bin/clang \
+  CC_wasm32_unknown_wasi=/wasm-sysroot/bin/clang \
+  PATH=$PATH:/rust/bin \
+  RUSTFLAGS=-Ctarget-feature=-crt-static
--- a/ci/docker/wasm32-unknown-wasi/Dockerfile
+++ b/ci/docker/wasm32-unknown-wasi/Dockerfile
@@ -1,4 +1,7 @@
-FROM ubuntu:18.04 as reference-sysroot
+# In the first container we want to assemble the `wasi-sysroot` by compiling it
+# from source. This requires a clang 8.0+ compiler with enough wasm support and
+# then we're just running a standard `make` inside of what we clone.
+FROM ubuntu:18.04 as wasi-sysroot
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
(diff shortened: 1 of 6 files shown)


libc-test/build.rs
```diff
@@ -1866,3 +1867,67 @@ fn test_dragonflybsd(target: &str) {
...
+            // wasi shenanigans for defining CLOCK_REALTIME and such
+            s if s.starts_with("__CLOCK") => true,
+            _ => false,
+        }
+    });
+
+    // Looks like LLD doesn't merge duplicate imports, so if the Rust
+    // code imports from a module and the C code also imports from a
+    // module we end up with two imports of function pointers which
+    // improt the same thing bug have different function pointers
```
gnzlbg: typo: "bug"/"but" 
typo: "improt"/"import"

So IIUC the issue here is that "libc" (wasi/the sysroot) is linked statically into both the Rust code (due to "-C target-feature=+crt-static"), and the C code, and that when these two are linked, two copies of the "libc" are present in the final binary. This is super-unfortunate. 

Have you tried whether `-C lto=fat` solves the issue ? That might be too brittle. Alternatively, maybe for this target it might make sense to not link the C code (or the Rust code) to the sysroot when compiling, so that when Rust and C are linked together, there is only one sysroot.

src/wasi.rs
```diff
@@ -0,0 +1,792 @@
...
+pub type c_ushort = u16;
+pub type c_long = i32;
+pub type c_ulong = u32;
+pub type c_longlong = i64;
+pub type c_ulonglong = u64;
+pub type size_t = usize;
+pub type ssize_t = isize;
+pub type off_t = i64;
+pub type pid_t = i32;
+pub type int8_t = i8;
```
gnzlbg: I'd rather not add the `int8_t` and similar type-aliases. We do guarantee that `i8 == int8_t` on all targets, and people should just use `i8` on C FFI in the same way they just use `bool` - we should just deprecate these aliases instead for the other targets (see #1304 ).

src/wasi.rs
```diff
@@ -0,0 +1,792 @@
...
+        d: *const tm,
+    ) -> size_t;
+    pub fn gmtime(a: *const time_t) -> *mut tm;
+    pub fn gmtime_r(a: *const time_t, b: *mut tm) -> *mut tm;
+    pub fn localtime_r(a: *const time_t, b: *mut tm) -> *mut tm;
+    pub fn asctime_r(a: *const tm, b: *mut c_char) -> *mut c_char;
+    pub fn ctime_r(a: *const time_t, b: *mut c_char) -> *mut c_char;
+
+    #[link_name = "CLOCK_REALTIME"]
+    pub static __CLOCK_REALTIME_ADDR: __clockid;
```
gnzlbg: So this is weird: `__clockid` is a ZST but C does not have ZSTs. So.. this is a zero-sized static at some address? and when taking the address to it the right thing happens?

src/wasi.rs
```diff
@@ -0,0 +1,792 @@
...
+        pub st_nlink: __wasi_linkcount_t,
+        pub st_size: __wasi_filesize_t,
+        pub st_atim: __wasi_timestamp_t,
+        pub st_mtim: __wasi_timestamp_t,
+        pub st_ctim: __wasi_timestamp_t,
+    }
+
+    pub struct __wasi_ciovec_t {
+        pub buf: *const ::c_void,
+        pub buf_len: usize,
```
gnzlbg: I'd prefer if these would use the "right" C types (e.g. `size_t`, `ptrdiff_t`, `uintptr_t`, etc.) Same pretty much everywhere where `usize` is used below.

src/wasi.rs
```diff
@@ -0,0 +1,792 @@
...
+        pub userdata: __wasi_userdata_t,
+        pub error: __wasi_errno_t,
+        pub type_: __wasi_eventtype_t,
+        pub u: __wasi_event_u,
+    }
+
+    #[allow(missing_debug_implementations)]
+    pub union __wasi_event_u {
+        pub fd_readwrite: __wasi_event_u_fd_readwrite_t,
+        _bindgen_union_align: [u64; 2usize],
```
gnzlbg: I think the `usize` in `2usize` is unnecessary, but if you want to emphasize it, clippy lints all code to prefer writing `2_usize` instead (and that's what a lot of code does as a consequence).

src/wasi.rs
```diff
@@ -0,0 +1,792 @@
...
+    pub union __wasi_prestat_u {
+        pub dir: __wasi_prestat_u_dir_t,
+    }
+
+}
+
+pub static CLOCK_REALTIME: clockid_t = unsafe { &__CLOCK_REALTIME_ADDR };
+pub static CLOCK_MONOTONIC: clockid_t = unsafe { &__CLOCK_MONOTONIC_ADDR };
+pub static CLOCK_MONOTONIC_RAW: clockid_t =
+    unsafe { &__CLOCK_MONOTONIC_RAW_ADDR };
```
gnzlbg: So how are these expressend in C? 

I see that the `__CLOCK_..._ADDR` are zero-sized statics, and that this takes a `&'static` reference to them. Are they macros in C?

ci/docker/wasm32-unknown-wasi/Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+    make \
+    ssh \
+    xz-utils
+
+RUN curl http://releases.llvm.org/8.0.0/clang+llvm-8.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz | tar xJf -
+RUN mv /clang+llvm-8.0.0-x86_64-linux-gnu-ubuntu-18.04 /wasmcc
+
+RUN git clone https://github.com/CraneStation/reference-sysroot-wasi && \
+  cd reference-sysroot-wasi && \
+  git reset --hard d5a609fe63926533e1054e539ba5f2693d51bdf5
```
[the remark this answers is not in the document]
alexcrichton: Oh nothing special, just the current HEAD. The thinking is that there's likely going to be breaking changes to this sysroot, so we can try to mitigate that impact on libc's own CI by only deliberately upgrading when necessary. I'll be sure to update the comment!

ci/docker/wasm32-unknown-wasi/Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+ENV PATH=/root/.cargo/bin:$PATH
+
+RUN apt-get install -y --no-install-recommends python
+RUN git clone https://github.com/CraneStation/wasmtime-wasi wasmtime && \
+  cd wasmtime && \
+  git reset --hard a7ac05df74759a7536b2b1e30adc6ff4867e36c3
+
+# Install wasmtime in /usr/bin, but make sure to remove rust afterwards because
+# we don't want it conflicting with the main Rust we're using to compile
+# `libc`.
```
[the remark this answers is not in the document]
alexcrichton: Ah this was an outdated comment, my bad!

libc-test/build.rs
```diff
@@ -1866,3 +1867,67 @@ fn test_dragonflybsd(target: &str) {
...
+        t if t.starts_with("__wasi") && t.ends_with("_u") => format!("union {}", t),
+        t if t.starts_with("__wasi") && is_struct => format!("struct {}", t),
+        t if t.ends_with("_t") => t.to_string(),
+        t if is_struct => format!("struct {}", t),
+        t => t.to_string(),
+    });
+
+    // This is an opaque struct but we go through shenanigans to define values
+    // for it
+    cfg.skip_struct(move |ty| ty == "__clockid");
```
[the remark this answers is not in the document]
alexcrichton: I'm just gonna go ahead an exclude `__clockid` because it's sufficiently "weird enough" and not entirely necessary yet that I think it's fine to defer it to later

libc-test/build.rs
alexcrichton: Nah this is actually a different issue that's somewhat deeper in LLD and unrelated to linkage/LTO/etc. It has to do with an LLVM module that looks like:

```llvm
define void @foo() {
    call void @bar1()
    call void @bar2()
}

declare void @bar1() #1 
declare void @bar2() #2 

attributes #1 = { "wasm-import-module"="foo" }
attributes #1 = { "wasm-import-module"="foo" "wasm-import-name"="bar1" }
```

Both `bar1` and `bar2` reference the same function, but due to how LLD is implemented these show up as two different functions in the final binary, where the same name is imported twice (giving the same function two different function indices).

It's not really an issue, just a minor thing that will likely get fixed in LLD at some point.

src/wasi.rs
alexcrichton: This has come up historically though where for better or worse these types are used in standard Rust types and whatnot, so I think it's better to include them on all platforms for better compatibility and consider the deprecation story separately?

src/wasi.rs
alexcrichton: This is actually all generated by `bindgen` where `size_t` is the same as `usize`. I think they're basically interchangeable at this point (wasi is unlikely to ever have the insanity that is differences in the three types you mentioned), but would you still prefer `s/usize/size_t/` everywhere?

src/wasi.rs
gnzlbg: We definitely can do that. I think that some projects use these and some projects don't. Changing a project from using `int8_t` to just using `i8` is a "backwards compatible" change, and if someone complaints that wasi does not have these we can always add them later.

gnzlbg: And by that I mean that I'll leave it up to you.

src/wasi.rs
gnzlbg: They are definitely interchangeable for this target, I think that's all that matters. My preference for `size_t` here is for clarity, since that's what the C code does. IIRC bindgen typically preserves the C types if one explicitly tells it to use the `libc::` types, but... there is no libc for wasi yet... If bindgen also generated the `size_t` type alias, it maybe has a "bug".

libc-test/build.rs
```diff
@@ -1866,3 +1867,67 @@ fn test_dragonflybsd(target: &str) {
 
     cfg.generate("../src/lib.rs", "main.rs");
 }
+
+fn test_wasi(target: &str) {
+    assert!(target.contains("wasi"));
+
+    let mut cfg = ctest::TestGenerator::new();
+    cfg.define("_GNU_SOURCE", None);
```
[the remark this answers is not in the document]
alexcrichton: IIRC it was to get some more of the libc APIs to show up, but I can play around with it once we turn this verification on

src/wasi.rs
alexcrichton: Ok in that case I think I'd be inclined to go ahead and add them because we've already had at least one expicit request to add it to the wasm32-unknown-unknown target (which this still doesn't do)

src/wasi.rs
alexcrichton: Ok, I'll go ahead and switch it all to `size_t`

src/wasi.rs
gnzlbg: Sure, fine with me. I'm marking this as resolved. 

Do you have a link to the request ? I think the biggest thing blocking the deprecation is telling bindgen to just use the Rust types always.

alexcrichton: Ah not currently, I forget where the request came from...

gnzlbg:
@bors: retry

---

something weird is going on

PR #1310, opened 2019-03-29
Add CloudABI pthread_mutex related functions and types

faern:
Adding types and functions related to creating and using pthread mutexes and condvars on CloudABI. Taken from [these headers](https://github.com/NuxiNL/cloudlibc/blob/92cb7670f864adc625c24eb214ff1c6d888adf6b/src/include/pthread.h).

This is in preparation for implementing a proper `ThreadParker` for CloudABI in `parking_lot`. There is a [pthread one already](https://github.com/Amanieu/parking_lot/blob/56e313f8b23557da8b3b48fd30d3e9837c231ca7/core/src/thread_parker/unix.rs). So hopefully we can just plug that one in when the platform has these libc bindings.

Ping @EdSchouten, who seem knowledgeable about CloudABI synchronization primitives. Do you think this looks reasonable?

### DIFF
--- a/src/cloudabi/mod.rs
+++ b/src/cloudabi/mod.rs
@@ -34,6 +34,9 @@ pub type sa_family_t = u8;
 pub type socklen_t = usize;
 pub type time_t = i64;
 
+pub type clockid_t = *const __clockid;
+pub type cloudabi_clockid_t = u32;
+
 s! {
     pub struct addrinfo {
         pub ai_flags: ::c_int,
@@ -96,13 +99,17 @@ s! {
     }
 
     pub struct pthread_condattr_t {
-        __clock: usize,
+        __clock: clockid_t,
         __pshared: ::c_int,
     }
 
     pub struct timespec {
         pub tv_sec: ::time_t,
         pub tv_nsec: ::c_long,
+    }
+
+    pub struct __clockid {
+        pub id: cloudabi_clockid_t,
     }
 }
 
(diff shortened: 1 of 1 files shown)


EdSchouten:
Hi there,

The pthread datatypes exposed by the C library were never intended to have a stable ABI. This is not a true requirement for CloudABI yet, mainly due to a strong focus on static linkage. That said, we'd better keep the number of data structures copied from cloudlibc to an absolute minimum. It also increases the amount of stuff from libc on which we depend to a minimum.

For CloudABI, I intentionally implemented the mutex datatypes directly on top of CloudABI's futex system call API. By doing so, we achieved locking primitives whose implementation might not exactly be as optimised as parking_lot, but came pretty close. Mutexes and rwlocks may be implemented in as little as 4 bytes, while recursive mutexes would be 8 bytes in size. The locks would implement priority inheritance properly and also provide fairness.

This is why I'm of the opinion that implementing this on top of pthreads is actually a step backwards. It increases the reliance on parts of the C library that have no strong ABI.

faern:
@EdSchouten So for adding CloudABI support to `parking_lot`, would you suggest I instead implement the parking mechanism via the `cloudabi::poll(&subscription, ...)` calls, in the same way the current libstd primitives are implemented?

EdSchouten:
Yes, exactly!

CloudABI is somewhat opposite of most UNIX flavours: the system call API/ABI is intended to be stable, whereas the libc API/ABI is not. The reason for that is that the former has to be stable anyway to be able to support multiple OSes/runtimes properly.

faern:
Ok, thank you. I will then start writing a `ThreadParker` in that way. Will ping you for some input when it's getting ready if that is OK?

Will keep this PR open a bit longer in case some more information surfaces soon.

EdSchouten:
Sure thing. Thanks!

faern:
@EdSchouten There is no such thing as a simple futex/atomic wait? Where I can just park the thread waiting on an integer address to receive a notification? That's what we do on other systems. See here for example: https://github.com/Amanieu/parking_lot/blob/master/core/src/thread_parker/wasm.rs#L50

If mutexes and condvars is all I have at my disposal it looks like I need a solution very similar to the pthread design in the unix impl I linked in the initial post of this PR. Where the parker holds both a mutex and a condvar and needs to lock the mutex and wait for the condvar in order to park.

PR #1312, opened 2019-04-02
Add an issue template

gnzlbg:


alexcrichton:
I might recommend putting it in `<!-- ... -->` so it doesn't show up in the bug report (but shows up to users, but other than that seems fine by me!

PR #1314, opened 2019-04-05
 Add binding for new `__wasilibc_find_relpath` API 

alexcrichton:
This'll be used in libstd upstream, and this also enables wasi on CI!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -204,6 +204,9 @@ matrix:
       stage: tier2
     - env: TARGET=x86_64-unknown-linux-musl
       stage: tier2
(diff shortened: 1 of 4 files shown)


gnzlbg:
@alexcrichton maybe it would be better to allow CI to fail for wasi for now, so that the other changes here can land. You can always fix CI in a subsequent PR.

alexcrichton:
Oh I think that's just a mistake in the container, the CI should be pretty reliable because all the relevant wasi pieces are pinned. If it becomes an issue though then it can of course be disabled

PR #1316, opened 2019-04-12
[API-breaking] Expose si_addr on siginfo_t. Refs #716

alex:
This is the bare minimum to expose this. It also seems to have required making the unions `pub`, which I wasn't wild about, and I had no clue what to name the anonymous union. But it's a starting point!

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -2,6 +2,126 @@ pub type pthread_t = c_ulong;
 pub type __priority_which_t = ::c_uint;
 pub type __rlimit_resource_t = ::c_uint;
 
+cfg_if! {
+    if #[cfg(libc_union)] {
+        #[repr(C)]
+        #[derive(Copy, Clone)]
+        struct addr_bnd_t {
+            lower: *mut ::c_void,
+            upper: *mut ::c_void,
+        }
+
+        #[repr(C)]
+        #[derive(Copy, Clone)]
+        union sigfault_t_anonymous_union {
+            addr_bnd: addr_bnd_t,
+            pkey: u32,
+        }
+
+        #[repr(C)]
+        #[derive(Copy, Clone)]
+        struct sigfault_t {
+            addr: *mut ::c_void,
+            #[cfg(target_arch = "sparc")]
+            trapno: ::c_int,
+            addr_lsb: ::c_short,
+            anonymous_union: sigfault_t_anonymous_union,
+        }
+
+        cfg_if! {
+            if #[cfg(target_pointer_width = "64")] {
+                type SI_PADDING = [::c_int; 28];
+            } else {
+                type SI_PADDING = [::c_int; 29];
+            }
+        }
+
+        #[repr(C)]
+        #[derive(Copy, Clone)]
+        union sifields_t {
+            _pad: SI_PADDING,
+            sigfault: sigfault_t,
+        }
+
+        impl ::fmt::Debug for sigfault_t_anonymous_union {
+            fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                // Excludes the fields because we can't know which is valid
+                f.debug_struct("sigfault_t_anonymous_union")
+                    .finish()
+            }
+        }
+
+        impl ::fmt::Debug for sigfault_t {
+            fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                // TODO: include trapno on Sparc
+                f.debug_struct("sigfault_t")
+                    .field("addr", &self.addr)
+                    .field("addr_lsb", &self.addr_lsb)
+                    .field("anonymous_union", &self.anonymous_union)
+                    .finish()
+            }
+        }
+
+        impl ::fmt::Debug for sifields_t {
+            fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                // No way to print anything more detailed without the
+                // discriminant from siginfo_t
+                f.debug_struct("sifields_t").finish()
+            }
+        }
+    } else {
+        type sifields_t = [::c_int; 29];
+    }
+}
+
+s_no_extra_traits! {
+    pub struct siginfo_t {
+        pub si_signo: ::c_int,
+        pub si_errno: ::c_int,
+        pub si_code: ::c_int,
+        sifields: sifields_t,
+
+        #[cfg(target_arch = "x86_64")]
+        _align: [u64; 0],
+        #[cfg(not(target_arch = "x86_64"))]
+        _align: [usize; 0],
+    }
+}
+
+cfg_if! {
+    if #[cfg(libc_union)] {
+        impl siginfo_t {
+            pub unsafe fn si_addr(&self) -> *mut ::c_void {
+                self.sifields.sigfault.addr
+            }
+        }
+    } else {
+        impl siginfo_t {
+            pub unsafe fn si_addr(&self) -> *mut ::c_void {
+                #[repr(C)]
+                struct siginfo_sigfault {
+                    _si_signo: ::c_int,
+                    _si_errno: ::c_int,
+                    _si_code: ::c_int,
+                    si_addr: *mut ::c_void
+                }
+                (*(self as *const siginfo_t as *const siginfo_sigfault)).si_addr
+            }
+        }
+    }
+}
+
+impl ::fmt::Debug for siginfo_t {
+    fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+        // TODO: include fields from sifields
+        f.debug_struct("siginfo_t")
+            .field("si_signo", &self.si_signo)
+            .field("si_errno", &self.si_errno)
+            .field("si_code", &self.si_code)
+            .finish()
+    }
+}
+
 s! {
     pub struct statx {
         pub stx_mask: u32,
@@ -179,19 +299,6 @@ s! {
         pub rt_mtu: ::c_ulong,
         pub rt_window: ::c_ulong,
         pub rt_irtt: ::c_ushort,
-    }
-}
-
-impl siginfo_t {
-    pub unsafe fn si_addr(&self) -> *mut ::c_void {
-        #[repr(C)]
-        struct siginfo_sigfault {
-            _si_signo: ::c_int,
-            _si_errno: ::c_int,
-            _si_code: ::c_int,
-            si_addr: *mut ::c_void
-        }
-        (*(self as *const siginfo_t as *const siginfo_sigfault)).si_addr
     }
 }
 


alex:
Hmm, so the macros for unions don't automatically mark them `Eq`/`Debug`/etc. but we do for structs. These are fields in a public struct, so they need to be `Eq` somehow...

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -1,5 +1,18 @@
 pub type __priority_which_t = ::c_uint;
 
+s_no_extra_traits! {
+    pub union XXX_not_sure_what_to_call_this {
+        addr_bnd: addr_bnd_t,
+        pkey: u32,
+    }
+
+    #[repr(C, packed)]
+    pub union sifields_t {
```
gnzlbg: Why is this union `packed` ? 

cc @joshtriplet why do we accept `repr(packed)` on `union`s at all ? (it is a concept that IMO does not make much sense, at least for `repr(C)` unions, which are always packed.

alex: https://github.com/rust-lang/libc/issues/716#issuecomment-331675843

gnzlbg: cc @joshtriplett, @rkruppe: why do we accept `repr(packed)` on `union`s at all ? Is that something that makes sense? AFAICT for `repr(C)` unions at least it does not, they are "always" packed (all fields start at offset 0, there is no padding, etc.).

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -44,11 +57,24 @@ s! {
...
+        addr_lsb: ::c_short,
+        xxx_name_unclear: XXX_not_sure_what_to_call_this
+    }
+
     pub struct siginfo_t {
         pub si_signo: ::c_int,
         pub si_errno: ::c_int,
         pub si_code: ::c_int,
-        pub _pad: [::c_int; 29],
+        sifields: sifields_t,
```
gnzlbg: In the open issue #716 I was hoping that the padding field would be private, which would have allowed us to change this, but for some reason it is public, so this would be a hard breaking change, which I don't think we can make until the next breaking version is released.

src/unix/notbsd/linux/other/mod.rs
rkruppe: `repr(packed)` also lowers the alignment requirement of the annotated type (`repr(align(N))` can only raise it), which is also useful for unions and -- from a cursory glance at the linked issue -- apparently exactly what's needed here.

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -232,6 +258,12 @@ s_no_extra_traits! {
     }
 }
 
+impl siginfo_t {
+    pub unsafe fn si_addr(&self) -> *mut ::c_void {
+        self.sifields.sigfault.addr
+    }
+}
```
gnzlbg: IIRC the C macros work like methods right? E.g. `#define si_addr sifields.__sigfault__.addr` or something like that ? 

We usually add C macros as normal functions, but I think in this case it makes sense to add them as methods, since that is how they would be used in C.

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -1,5 +1,18 @@
 pub type __priority_which_t = ::c_uint;
 
+s_no_extra_traits! {
+    pub union XXX_not_sure_what_to_call_this {
```
gnzlbg: How is this called in C ? Or is it an anonymous `union` ?

src/unix/notbsd/linux/other/mod.rs
alex: Yup, exactly.

src/unix/notbsd/linux/other/mod.rs
alex: Anonymous :-(

src/unix/notbsd/linux/other/mod.rs
gnzlbg: > repr(packed) also lowers the alignment requirement of the annotated type 

Ah, that makes sense then. When used on `union`s, is it then equivalent to `repr(align(1))` ? 

---

I thought that it lowered the alignment requirements of the fields, which impacted how the alignment of the annotated was computed.

src/unix/notbsd/linux/other/mod.rs
gnzlbg: cc @alexcrichton I think this is the first time we add methods to `libc` structs, but that seems justified here.

src/unix/notbsd/linux/other/mod.rs
rkruppe: > I thought that it lowered the alignment requirements of the fields, which impacted how the alignment of the annotated was computed.

You can also view it that way if you want, then lowering the alignment of the whole type is a consequence that also applies to unions (since the requirements of the field(s) at offset 0 also matters). `repr(align(1))` won't help, `repr(align)` can only *raise* alignment.

gnzlbg:
> Hmm, so the macros for unions don't automatically mark them Eq/Debug/etc. but we do for structs. These are fields in a public struct, so they need to be Eq somehow...

The problem is that a `union` does not tell which field is active, so automatically deriving these cannot in general work. I think it is ok for these `union`s to not implement those traits. You could manually implement `Debug` to at least print the `union` name.

Usually, these `union`s are members of `struct`s, where the `struct` stores a discriminant that allows it to tell which field is the active one. For such `struct`s one can manually implement `PartialEq`, `Debug`, etc.

src/unix/notbsd/linux/other/mod.rs
gnzlbg: Then maybe just call it `sigfault_t_anonymous_union`. It's a private field, so I don't think we have to make this type public, right?

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -44,11 +57,24 @@ s! {
         pub ss_size: ::size_t
     }
 
+    pub struct addr_bnd_t {
```
gnzlbg: This type is only used in a private field, I don't think it needs to be public either.

src/unix/notbsd/linux/other/mod.rs
alex: The `s!` macro only works with `pub` types, so I just went that direction for simplicity.

gnzlbg: I think it would be ok to implement these outside of `s!`.

src/unix/notbsd/linux/other/mod.rs
alexcrichton: It seems reasonable to add methods here to me, but especially due to types like `siginfo_t` showing up on multiple platforms there will probably be a slow and steady trickle of issues to add the methods to all other platforms as well. I think that's fine though!

For guidelines I'd recommend that methods should probably be as simple as possible when they can be, ideally only for this use case of C has some macro that accesses a bunch of sub-fields which doesn't translate well to Rust. I wouldn't want to get more expansive to other more complicated computations.

alex:
For bonus points 1.13 doesn't like unions :-(

gnzlbg:
@alex yes, I hope I mentioned that in the issue. You need to use `cfg(libc_union)`, generating both type variants (with and without `union`), and implementing the methods for both cases (e.g. without unions, you just read the appropriate type from the beginning of the padding field).

alex:
Hmm, so I attempted to make `s_no_extra_traits!` support non-`pub` types, and that didn't seem to go so well. Am I barking up the strong tree here?

gnzlbg:
That should be possible, but I don't know how easy / hard doing that is. I would keep it simple for now, and just implement these outside any `s_...!` macros. We can figure out how to add support for these in the macros later. That might mean that the `style` lint might fail temporarily for this, that's ok.

alex:
Ok, after much pain this now builds!

It's failing the size_of check, which is confusing to me, because it looks like it's failing in precisely the way you'd get if the `repr(packed)` was missing. A local test seems to indicate that this ought to work...

```
alexgaynor@penguin /tmp> cat t.rs 
use std::mem;

#[repr(C)]
#[derive(Copy, Clone)]
struct addr_bnd_t {
    lower: *mut (),
    upper: *mut (),
}

#[repr(C)]
#[derive(Copy, Clone)]
union sigfault_t_anonymous_union {
    addr_bnd: addr_bnd_t,
    pkey: u32,
}

#[repr(C)]
#[derive(Copy, Clone)]
struct sigfault_t {
    addr: *mut (),
    #[cfg(target_arch = "sparc")]
    trapno: i32,
    addr_lsb: i16,
    anonymous_union: sigfault_t_anonymous_union,
}

#[repr(C, packed)]
#[derive(Copy, Clone)]
union sifields_t {
    _pad: [i32; 29],
    sigfault: sigfault_t,
}

pub struct siginfo_t {
    pub si_signo: i32,
    pub si_errno: i32,
    pub si_code: i32,
    sifields: sifields_t,

    #[cfg(target_arch = "x86_64")]
    _align: [u64; 0],
    #[cfg(not(target_arch = "x86_64"))]
    _align: [usize; 0],
}

fn main() {
    println!("{}", mem::size_of::<siginfo_t>());
}
alexgaynor@penguin /tmp> rustc +nightly t.rs; and ./t
warning: type `siginfo_t` should have an upper camel case name
  --> t.rs:34:12
   |
34 | pub struct siginfo_t {
   |            ^^^^^^^^^ help: convert the identifier to upper camel case: `SiginfoT`
   |
   = note: #[warn(non_camel_case_types)] on by default

warning: field is never used: `sifields`
  --> t.rs:38:5
   |
38 |     sifields: sifields_t,
   |     ^^^^^^^^^^^^^^^^^^^^
   |
   = note: #[warn(dead_code)] on by default

128
```

alex:
Let's go around the room, what's the dumbest thing you've done so far today?

I'll go first: I didn't actually push the commit where I added `packed`.

No one else needs to go :-)

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -1,5 +1,104 @@
...
+        #[repr(C, packed)]
+        #[derive(Copy, Clone)]
+        union sifields_t {
+            _pad: [::c_int; 29],
+            sigfault: sigfault_t,
+        }
+
+        impl ::fmt::Debug for sigfault_t_anonymous_union {
+            fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                // TODO: no idea how to tell which of the members is used
```
gnzlbg: You can't tell, so that's not possible.

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -1,5 +1,104 @@
...
+}
+
+cfg_if! {
+    if #[cfg(libc_union)] {
+        impl siginfo_t {
+            pub unsafe fn si_addr(&self) -> *mut ::c_void {
+                self.sifields.sigfault.addr
+            }
+        }
+    }
```
gnzlbg: This method should also be implemented if `#[cfg(libc_union)]` is false.

alex: How would you want it implemented? `panic!()`? Manually compute the offset and hardcode it?

src/unix/notbsd/linux/other/mod.rs
alex: Depending on how long we expect it to take to get this into a mergable state, perhaps it would make sense for me to send a seperate PR to make that field private, which could be merged/released sooner?

src/unix/notbsd/linux/other/mod.rs
gnzlbg: > How would you want it implemented?

Such that it always works correctly.

> Manually compute the offset and hardcode it?

All `repr(C)` `union` fields are at offset 0, or what offset do you mean?

alex: Sure, in this particular case it's easy, it's just the first `*mut c_void` in `_pad`, however if we get to exposing other fields the logic will be more complicated. In any event, this one is easy enough.

alex:
(Realized that a force push doesn't trigger a notification -- I think all the comments have been resolved)

gnzlbg:
So this looks good. 

The only issue that remains is how to deal with the breaking API change. I'll think about that.

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -1,5 +1,110 @@
...
+    if #[cfg(libc_union)] {
+        impl siginfo_t {
+            pub unsafe fn si_addr(&self) -> *mut ::c_void {
+                self.sifields.sigfault.addr
+            }
+        }
+    } else {
+        impl siginfo_t {
+            pub unsafe fn si_addr(&self) -> *mut ::c_void {
+                *(self.sifields.as_ptr() as *const *mut ::c_void)
```
alex: Yeah, this isn't correct. There should be 4 bytes of padding between `si_code` and `sifields`, but you won't get that since `[i32; 29]` has an alignment of 4.

Is there a sane way to fix this?

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -1,5 +1,110 @@
...
+            #[cfg(target_arch = "sparc")]
+            trapno: ::c_int,
+            addr_lsb: ::c_short,
+            anonymous_union: sigfault_t_anonymous_union,
+        }
+
+        #[repr(C, packed)]
+        #[derive(Copy, Clone)]
+        union sifields_t {
+            _pad: [::c_int; 29],
```
alex: So this is interesting. There's a disagreement between `/usr/include/asm-generic/siginfo.h` and `/usr/include/x86_64-linux-gnu/bits/siginfo.h` on my system about whether `_pad` is 28 or 29 elements. They produce the same overall struct size because of padding/alignment, but I'm not sure what the right value to put here is?

gnzlbg: Hm, this is weird. If the struct size is the same, I think I'd rather pick 29, but... this might be a bug that you might want to report upstream.

alex:
Is there anything I can be doing to help push this forward at this point, or are we fully blocked on needing to consider how to handle the backwards incompatibility?

gnzlbg:
I've tried to inspect crates using the struct, but sourcegraph is broken AFAICT. We can't really use `crater` for this, since that won't catch code using `libc` from crates.io. I think it might be better to submit a parallel PR that marks the field as "deprecated", and do a release, telling people to report on an issue here, to see who (if anybody) is using the field.

jedisct1:
`si_addr` is now a function on Linux, but a field on MacOS, *BSD and Solaris derivatives.

For consistency, could it be a function everywhere?

gnzlbg:
A consistent API is provided in the `nix` crate, would that solve your problem?

jedisct1:
`nix` only reexports `libc::siginfo_t`.

gnzlbg:
@jedisct1 then I think it might be worth it to raise this issue there. `libc` just exposes the system APIs, and the API of linux differs from that of many other OSes when it comes to `siginfo_t`.

DoumanAsh:
Just curious, I'm currently using `_pad` to get out provided pointer with my data when getting signal.

Currently compiler throw warning on it with reference to this PR, can I assume that any API breakage will lead to libc 0.2?

alex:
@DoumanAsh don't use `_pad`, it leads to alignment issues -- instead cast the whole struct, as done here: https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/linux/gnu/mod.rs#L185-L196

DoumanAsh:
I'm not sure how you expect me to cast the whole struct, which I already have, but libc doesn't expose `sigval` field for me.
Regardless my question is whether this PR will bump minor version or patch version?

gnzlbg:
@DoumanAsh which field are you trying to access precisely ? Maybe we can add a method here to provide this field, and you could start using that instead.

DoumanAsh:
@gnzlbg `si_sigval`

I think most platforms aside freebsd has it as part of enum since it is used by Posix timers
But if libc would be able somehow imitate C like enums that would be actually great

gnzlbg:
> But if libc would be able somehow imitate C like enums that would be actually great

We want to use a proper rust `union` here, but we can't do that without breaking those who are accessing the padding field directly. I think we can just add a `si_value` method that returns that, and you can migrate to use that. 

Would you like to send a PR implementing it ? The current `siginfo_t` already has a method exposing `si_addr`, you could add a `si_value` method that internally accesses the padding, just like you are doing right now in your code.

DoumanAsh:
@gnzlbg To be honest I'm not exactly sure my method is the best one.
I wouldn't mind libc switching to unions, and then relying on it rather than trying to read particular offsets.
The problem is that I imagine libc doesn't want to bump minor version which means breaking changes are not allowed...

About PR, I'll try to take a look and add it for platforms with which I'm familiar enough

gnzlbg:
The plan is to try to make the padding field private with a patch bump, hoping that it doesn't break anybody. If that success, we can modify the internals to use unions like this PR does, and once that works correctly, we can make them public. 

We will need to keep the methods around for backward compatibility with toolchains that do not support unions (libc supports up to Rust 1.13.0...).

alex:
I'm going to close this for now, it's significantly bit-roting, and from my perspective the current implementation works well enough.

LunarLambda:
was pointed here by the source code comment on siginfo_t.

I'd like access to si_pid and si_status to be able to passively (rather than actively) wait for a child process exit (via SIGCHLD and examining siginfo_t). I checked, and they're part of POSIX (https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html), though I suppose the issue is differing struct field order between OSs.

is there anywhere I can/should direct my questions/searches for this?

Issue #1318, opened 2019-04-17
Solarish __posix_readdir_r symbol is only for 32-bit libc


jasonbking:
As discovered building the latest 'fd', the `__posix_readdir_r` symbol is only available in the Solaris & illumos 32-bit libc.  For 32-bit libraries, there is some confusing symbol shuffling that occurs to provide compatibility between traditional Solaris, draft POSIX, and official POSIX behavior for `readdir_r` (which is where `__posix_readdir_r` originates).

By the time 64-bit was introduced in Solaris, that was resolved and there is only `readdir_r` for 64-bit binaries.  Since rust (currently) only supports 64-bit targets on Solarish (Solaris & illumos) targets, the symbol renaming in mod.rs is unnecessary.

gnzlbg:
How do we fix this ?

jasonbking:
The PR to fix the issue was merged a month ago.  Is there some other problem that's come up?

gnzlbg:
No, just checking. Thanks!

PR #1319, opened 2019-04-17
This fixes issue #1318 - on Solarish systems __posix_readdir_r is only

jasonbking:
defined for 32-bit Solarish targets.  Since rust doesn't currently support 32-bit Solarish targets, the line is both unnecessary and wrong.

As found with the latest fd binary, removing the line fixed the fd build (and fd then works as expected).

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -609,6 +609,12 @@ extern {
     #[cfg_attr(target_os = "macos", link_name = "readdir$INODE64")]
     #[cfg_attr(target_os = "netbsd", link_name = "__readdir30")]
     #[cfg_attr(target_os = "freebsd", link_name = "readdir@FBSD_1.0")]
+    /// The 64-bit libc on Solaris and illumos only has readdir_r.  If a
+    /// 32-bit Solaris or illumos target is ever created, it should use
+    /// __posix_readdir_r.  See libc(3LIB) on Solaris or illumos:
+    /// https://illumos.org/man/3lib/libc
+    /// https://docs.oracle.com/cd/E36784_01/html/E36873/libc-3lib.html
+    /// https://www.unix.com/man-page/opensolaris/3LIB/libc/
     pub fn readdir(dirp: *mut ::DIR) -> *mut ::dirent;
     #[cfg_attr(target_os = "macos", link_name = "readdir_r$INODE64")]
     #[cfg_attr(target_os = "netbsd", link_name = "__readdir_r30")]
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -612,8 +612,6 @@ extern {
     pub fn readdir(dirp: *mut ::DIR) -> *mut ::dirent;
     #[cfg_attr(target_os = "macos", link_name = "readdir_r$INODE64")]
     #[cfg_attr(target_os = "netbsd", link_name = "__readdir_r30")]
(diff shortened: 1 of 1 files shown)


gnzlbg:
It would be nice if someone with solaris experience could take a look, maybe @bgermann  ?

jasonbking:
It can be confirmed by running `nm` on `/lib/libc.so` vs. `/lib/amd64/libc.so` on a Solaris system (the former should have `__posix_readdir_r` while the latter (the 64-bit lib) will not.  Due to all the symbol gymnastics in `dirent.h`, it's very easy to miss the difference if you just try looking at the the header file.

jasonbking:
I would be surprised if Solaris has changed behavior.  This appears to have been the case since at least 2006 (looking at what symbols are exported in libc in the commit history which goes back to the OpenSolaris launch) which predates the Solaris/illumos split, and was strictly for backwards compatibility.

It looks like what happened was prior to 64-bit systems arriving on the scene, the interface was changed by POSIX, so the old interface (taking only two arguments) was left for backwards compatibility, while the final POSIX standard interface was added (using some header tricks for new programs to use the new version -- `__posix_readdir_r`) as well as the large file variant for 32-bit apps.  Since 64-bit apps post-date the interface change, only the final POSIX version was exposed for 64-bit applications.

bgermann:
The change is accurate. Please also see
https://docs.oracle.com/cd/E36784_01/html/E36873/libc-3lib.html and
https://www.unix.com/man-page/opensolaris/3LIB/libc/
which list __posix_readdir_r at a section titled "The following interfaces are unique to the 32-bit version of this library"

gnzlbg:
Thank you for taking a look. Would it be possible to add a documentation
comment explaining that and containing the relevant links? And maybe extend
the commit message with this information as well?

There is a lot of knowledge that remains spread in the PRs that is hard to
find when looking at a blame or when somebody tries to make a change. Right
now me and Alex are the single points of failure when it comes to remember
where to find what or who to ping, and I am trying to slowly change that.
Particularly for linkage, which is specially tricky due to how libc in
crates.io interacts with the built-in one.

jasonbking:
Sure -- would something like
```
/// 64-bit Solaris and illumos libc only has readdir_r.  A 32-bit Solaris or illumos target (if ever created)
/// should use __posix_readdir_r.  See libc(3LIB) on Solaris or illumos (https://illumos.org/man/3lib/libc).
```
work?  I figure it'll be easier to hash out the wording here than do a lot of churn on the PR.  I used the illumos man link since that's likely to be a more stable URL than docs.oracle.com (I think they've reorganized things at least once or twice on that site, making older links no longer work).

gnzlbg:
I'd rather err in the side of too many links, and add the links that @bgermann mentioned to the comment, but otherwise LGTM. I'll merge when the PR is updated. Thank you both!

src/unix/mod.rs
```diff
@@ -609,11 +609,15 @@ extern {
     #[cfg_attr(target_os = "macos", link_name = "readdir$INODE64")]
     #[cfg_attr(target_os = "netbsd", link_name = "__readdir30")]
     #[cfg_attr(target_os = "freebsd", link_name = "readdir@FBSD_1.0")]
+    /// The 64-bit libc on Solaris and illumos only has readdir_r.  If a
```
jasonbking: ... and I put this in the wrong spot.. ENOCAFFEINE

PR #1321, opened 2019-04-19
Add more WASI libc bindings

sunfishcode:
I was writing some WASI code and needed some more bits from libc. Here's a patch adding more things from the intersection of things that WASI libc defines and things that other platforms in libc have bindings for.


src/wasi.rs
```diff
@@ -87,16 +114,108 @@ s! {
...
+    }
+
+    pub struct stat {
+        pub st_dev: dev_t,
+        pub st_ino: ino_t,
+        pub st_nlink: nlink_t,
+        pub st_mode: mode_t,
+        pub st_uid: uid_t,
+        pub st_gid: gid_t,
+        __pad0: c_uint,
```
gnzlbg: Is there a way to avoid the padding here by using `repr(align(N))` or `repr(packed(N))` instead?

sunfishcode: Possibly, though explicitly naming the padding like this is what all other platforms do. It's possible that a future version of `stat` could repurpose the padding fields to return additional information in a backwards-compatible way. To support that, we need to give it a name now, so that struct copies never skip copying those bits.

sunfishcode:
> Some of the symbols are named __wasilibc, but others aren't. Is there a reason for this?

The `__wasilibc_*` functions are functions which are WASI-specific and not part of a typical libc API.

alexcrichton:
All looks good to me!

@sunfishcode looks like there's some verification errors on CI, and `libc-test/build.rs` may need to be updated for including more headers files or such on wasi.

While you're here, want to update the wasi check in `src/lib.rs` to `any(target_env = "wasi", target_os = "wasi")`?

src/wasi.rs
alexcrichton: Ah yeah FWIW most other platforms long predate `#[repr(align)]` hence the use of explicit fields. Nowadays I think it's best to use `#[repr]` instead of hidden fields if possible, but I wouldn't go out of your way because it's a backwards-compatible change to remove private fields. 

(AKA I agree with @gnzlbg but would agree we should iterate in tree instead of blocking anything)

sunfishcode:
Investigating this more, this has turned up some bugs in on the wasi-sysroot side, so I'm going to step back here momentarily, fix those, and them come back to this.

src/wasi.rs
gnzlbg: @alexcrichton I am actually convinced by @sunfishcode reply that if we use `repr`s for these, we won't copy these fields, and if the libc version linked to changes to one that uses these fields for something (e.g. relying on the user having zeroed this fields), we would be invoking UB (because Rust won't write, copy, etc. "implicit" padding). 

That is, we should only use `repr(packed)` and `repr(align)` if the C headers do so as well, but even if they do not, we might want to still use manual padding to remain forward compatible.  I've opened https://github.com/rust-lang/libc/issues/1324 and am marking this as resolved. Thank you @sunfishcode .

PR #1325, opened 2019-04-23
Add more WASI libc bindings

sunfishcode:
This picks up #1321, adding more WASI libc bindings, adding several more fixes. In particular, `cargo test --all` including libc-test now builds successfully.


gnzlbg:
@sunfishcode note that if you want to use travis for testing you can comment out all "irrelevant" build jobs in a commit, so that travis only tests the relevant ones for the PR, and once those are green, kill the commit.

sunfishcode:
@gnzlbg That's good to know, thanks! In any case though, the CI is now green here.

PR #1327, opened 2019-04-28
Don't bundle static libc.

hvenev:


gnzlbg:
Ping me when CI is green.

hvenev:
Ping.

alexcrichton:
This PR would unfortunately break both these targets as distributed with rust-lang/rust. The usage of `static` here is the precise use case for where `static` was designed for, which is updating the artifacts so C library doesn't have to be distributed separately but rather it's bundled with the Rust distribution.

alexcrichton:
Yes, linking with `static` will find `libc.a` and then take all the object files from there and put them in `liblibc.rlib`. With `static-nobundle` that doesn't happen and simply passes `-lc`, which would break cross compilation use cases for musl and wasi support which doesn't require installing the sysroot.

gnzlbg:
So one more thing. How does cross compiling for a `#![no_std]` binary that uses the libc crate from crates.io without the `use_std` feature work ? No `link(name = "c")` attributes are added in that case, so this is just a bare `extern {}` block.

alexcrichton:
I think it just basically doesn't work, building a fully linked binary with `#[no_std]` is pretty unstable and never something we've really optimized for

hvenev:
I don't see how musl is that different from glibc, uclibc, or dietlibc. Shouldn't something like `target_vendor = "rust"` be used to decide whether to bundle `libc` in `liblibc.rlib` versus using the one provided by the C toolchain?

gnzlbg:
> I don't see how musl is that different from glibc

IIUC glibc is dynamically linked, while musl appears to be always statically linked. I suppose that if we were to statically link glibc, then we'd have a similar issue.

hvenev:
Not necessarily. My platform does not support static linking.

gnzlbg:
And when compiling with rustc musl gets dynamically linked? Which target triple is that? (I thought rustc always statically linked musl for the musl targets).

hvenev:
A slightly modified `mips-openwrt-linux-musl`. With this patch applied and a similar change to libunwind in rustc, code compiles and runs just fine. Even unwinding seems to be there (not tested) and backtrace on panic works.

gnzlbg:
@alexcrichton to support that it might make sense to move the logic of whether musl is statically or dynamically linked to the target specification file.

hvenev:
Isn't the `crt-static` target feature supposed to be doing that?

alexcrichton:
Dealing with crt linkage on musl is basically an endless world of pain unfortunately in what we have set up in rustc. It's all workable one way or another but it's very meticulously crafted right now and I would discourage large changes without having a fuller picture of what's happening.

hvenev:
I agree with you. I've opened a new issue at https://github.com/rust-lang/rfcs/issues/2695. I'd like to hear your suggestions about how to deal with this.

alexcrichton:
I can be around to answer questions from time to time, but I do not have the time to help design a new system and or help architect changes.

gnzlbg:
I think that the fix here would be to just set `target_env` to `dynmusl`, such that the target triplet becomes `mips-openwrt-linux-dynmusl`, as proposed here: https://github.com/rust-lang/rust/issues/59302 , at least until a solution for https://github.com/rust-lang/rfcs/issues/2695 would be RFCed and implemented.

mati865:
Such `target_env` has to be implemented first.

gnzlbg:
@hvenev I'm closing this, I think that the next step here is to design a solution in rust-lang/rust for how to support both statically and dynamically-linked musl targets. And then propose a combined set of patches that implements support for this in rustc and libc, without breaking the statically linked musl targets that we already support.

@hvenev it seems that https://github.com/rust-lang/rust/issues/59302 is a good place to discuss and organize work around this.

petrochenkov:
Posted a plan for the fixing this in rustc repo https://github.com/rust-lang/rust/issues/72274.

PR #1329, opened 2019-05-01
Deprecate _pad field on siginfo_t

alex:
As discussed in https://github.com/rust-lang/libc/pull/1316

### DIFF
--- a/src/unix/notbsd/linux/other/mod.rs
+++ b/src/unix/notbsd/linux/other/mod.rs
@@ -48,6 +48,7 @@ s! {
         pub si_signo: ::c_int,
         pub si_errno: ::c_int,
         pub si_code: ::c_int,
+        #[deprecated(since="0.2.54", note="Please leave a comment on https://github.com/rust-lang/libc/pull/1316 if you're using this field")]
         pub _pad: [::c_int; 29],
         #[cfg(target_arch = "x86_64")]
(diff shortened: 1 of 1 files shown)


alex:
Hmm, what's teh appropriate way to wrap this long line?

gnzlbg:
You can just break line at `since` and `note`.

Could you add the methods to access the field here as well ? The note should tell people to report here with their use case, and that they can upgrade their code to use the methods instead (but the available methods might not be enough).

alex:
A method to access `_pad` you mean?

gnzlbg:
I mean a method like `si_addr` in the other PR. 

The problem is, we are telling people that `pad` is deprecated, so when they see that message, they would be like "ok, I shouldn't use pad, what should I use instead?". That's a fair question. Ideally, we would have enough methods implemented already so that we can just tell people to use the type's methods to access the appropriate fields. 

That might not be feasible to do right now, and not all methods might be needed, but at least we should be able to tell people that "if you need a method that's not available, look how `si_addr` is implemented, and please submit a PR with the method you need and we will release that quickly", to minimize friction.  

Only after everybody is able to port their code can we make the method private, and then continue the refactor towards using an union when possible.

gnzlbg:
So this LGTM. @bors: r+

PR #1330, opened 2019-05-01
Open 'val' field in 'fsid_t' to the public

alesharik:


gnzlbg:
This doesn't "open" the `val` field (it is still private). It's unclear to me what the motivation for this is.

alesharik:
Oops, sorry. The reason to '"open" `val` field is to compare `fsid_t` structs, because `extra_traits` feature doesn't work.
This dependency config: 
```toml
[dependencies]
libc = { git = "https://github.com/rust-lang/libc", features = ["extra_traits"] }
```
Produces this error:
```
the package depends on `libc`, with features: `extra_traits` but `libc` does not have these features.
```

gnzlbg:
What `libc` commit are you getting from git? And what does `cargo update` report ?

alesharik:
Thanks, `cargo update` solved this issue. Sorry for the wasted time.

gnzlbg:
No problem, and thank you for the PR nevertheless.

PR #1332, opened 2019-05-04
redox: convert to target_family unix

jackpot51:
This is the first step to supporting https://github.com/rust-lang/rust/issues/60139.

In order to have a smooth transition, there will need to be a change made in rust at the same time, switching Redox over to the unix target family. See https://github.com/rust-lang/rust/pull/60547


gnzlbg:
@jackpot51 I think it would be better to make the rust-lang/rust PR use the libc of this branch, and once bors try passes there, then we can merge this here, release, and update the rust-lang/rust PR to use the released version, and merge it with high priority. What do you think?

jackpot51:
I think that sounds good to me. I will do that soon

jackpot51:
@gnzlbg it is done: https://github.com/rust-lang/rust/pull/60547, https://github.com/rust-lang/rust/pull/60547/commits/2ab50afb0d50505d05547bb831dc211d8ba1a9e6

gnzlbg:
So let's merge this, do a release, and then you can update the PR upstream !

jackpot51:
@gnzlbg what should I do about the build failure? I checked and the failures do not seem to be introduced by this PR, but am open to including fixes if possible and necessary.

mati865:
It's https://github.com/rust-lang/libc/issues/1333

The thing is upgrading system on CI is not trivial task for this repo because the code is verified against the system.

jackpot51:
Woohoo!

PR #1334, opened 2019-05-09
Use FreeBSD SO_ constants on newlib

leo60228:
This is necessary to set socket options on the Nintendo Switch, which uses a socket implementation directly ported from FreeBSD.


leo60228:
i'm incredibly confused as to how i appear to have erased the history of my fork

leo60228:
fixed, i think, except github won't let me reopen it

Issue #1333, opened 2019-05-08
Ubuntu 17.10 repository is no longer available


mati865:
It will result in following jobs failure: https://github.com/rust-lang/libc/search?l=Dockerfile&q=17.10

cc https://github.com/rust-lang/libc/pull/1297

gnzlbg:
We should probably update directly to the latest Ubuntu version.  Will do that after #1337 is merged.

jackpot51:
@gnzlbg I recommend to update to 18.04, not the latest. It is a long term support version and will be supported for at least another 4 years, whereas 19.04 will only be supported until January of next year.

gnzlbg:
The main issue with the upgrades is that, instead of doing small incremental upgrades every year, we do a huge one every 4 years. Also, we technically only support the latest release of `libc` on most platforms, and we have to add many workarounds to do that while using a 4 year old version in CI.

jackpot51:
That is a good point. If it is possible to keep it up to date every 6 months, then that would be best. However, with many things still running 17.10... you will already have to jump three fourths the length of an LTS cycle to get to 19.04

gnzlbg:
Yeah, I'll focus first on getting CI back running again, so that we can merge the open PRs and unblock work in rust-lang/rust. Once that happens, we can start slowly migrating to newer Ubuntu versions in CI.

PR #1335, opened 2019-05-09
Improve newlib constants

leo60228:
This is necessary to set socket options on the Nintendo Switch, which uses a socket implementation directly ported from FreeBSD. Reopened from #1334 because I accidentally closed that.


leo60228:
From the person that originally added the constants (not that it matters much, but...):

image

mati865:
Bots failure: https://github.com/rust-lang/libc/issues/1333

gnzlbg:
@leo60228 do you have a link to these definitions in newlib upstream ?

gnzlbg:
Also, the Nintendo Switch libc bindings live here: https://github.com/rust-lang/libc/blob/master/src/switch.rs 

But maybe those aren't necessary anymore and can be deleted ?

---

It would be great if you could add at least a test that builds libc for your target.

leo60228:
@gnzlbg There's two Switch Rust toolchains; one based on newlib, and one based on pure Rust. switch.rs is for the pure Rust one.

leo60228:
Newlib upstream doesn't support sockets beyond copying the headers from Linux when built for Linux. The Switch's sockets are ported directly from FreeBSD, so I copied them over. I could possibly add a cfg flag so that the constants are only used on the Switch?

leo60228:
That .travis.yml says that it's using the dkp linker solely due to the outdated GBA CPU having problems with lld. I've been using dkp's linker for the Switch. Should I try to get it working with lld, or just use dkp's linker like the GBA does?

leo60228:
All of the newlib Switch-specific stuff is in the libnx repo: https://github.com/switchbrew/libnx/tree/master/nx/external/bsd/include/sys

I don't know why it's done this way when newlib includes headers itself for Cygwin and Linux, but...

leo60228:
@gnzlbg The CI seems to work.

leo60228:
#1333 still hasn't been fixed yet. I'm on mobile, but I'm pretty sure that's what's causing Travis to fail.

PR #1337, opened 2019-05-14
Update CI from ubuntu 16.04 (EOL) to 19.04

### LINKED ISSUE #1333: Ubuntu 17.10 repository is no longer available

gnzlbg:
Closes #1333 .


libc-test/build.rs
```diff
@@ -1768,51 +2470,184 @@ fn test_dragonflybsd(target: &str) {
     cfg.skip_type(move |ty| {
         match ty {
             // sighandler_t is crazy across platforms
+            // FIXME: is this necessary?
             "sighandler_t" => true,
 
+            // FIXME: these types have a p_type member but some header
+            // has a define p_type __p_type macro that breaks their tests:
+            "Elf64_Phdr" | "Elf32_Phdr" => true,
```
mati865: Not in scope of this PR:
I suppose it's from `resolv.h`, maybe headers should be tested separately so they don't conflict with each other?

gnzlbg: It is from `resolv.h`, the problem is, why is this issue triggering now, and was not triggering before.

mati865: `Resolv.h` doesn't change a lot: https://sourceware.org/git/?p=glibc.git;a=history;f=resolv/resolv.h;hb=7ac6fad987c28dc5e0e7964084260f163179be68

Maybe doing diff could show something.

gnzlbg: I worry that this is a header file ordering issue.

gnzlbg: I've just split the test for now.

mati865:
`SYS_syscalls` error comes from: https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=2dba5ce7b8115d6a2789bf279892263621088e74;hp=5f3b183d198b39ca993a41aadb02bddd9fde078d

gnzlbg:
Let's see if including the `bits/syscalls.h` header fixes those!

gnzlbg:
So that did not work, that header cannot be included directly.

mati865:
It seems `SYS_syscalls` is no longer generated, at least it's absent in 18.04 container with `libc6-dev` and `linux-headers-5.0.0` installed.

gnzlbg:
That would happen if the kernel does not define `__NR_syscalls`, but AFAICT it does ? Or maybe `syscalls` is just missing from the list that glibc uses.

gnzlbg:
Yes, that appears to be the case: https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=2dba5ce7b8115d6a2789bf279892263621088e74;hp=5f3b183d198b39ca993a41aadb02bddd9fde078d

Shows that "pseudo-system calls" that define ranges, and unused / reserved system calls, are skipped.

gnzlbg:
AFAICT, that's a backwards incompatible breaking change from glibc.

gnzlbg:
cc @fweimer

mati865:
Seems it's not present for every arch:

<details>

```
rg "__NR_syscalls" /usr/
/usr/include/asm-generic/unistd.h
736:#undef __NR_syscalls
737:#define __NR_syscalls 292
781:#undef __NR_syscalls
782:#define __NR_syscalls (__NR3264_lstat+1)
799:#undef __NR_syscalls
800:#define __NR_syscalls (__NR_signalfd+1)
834:#undef __NR_syscalls
835:#define __NR_syscalls (__NR_mmap+1)
897:#undef __NR_syscalls
898:#define __NR_syscalls (__NR_fork+1)

/usr/src/linux-headers-5.0.0-15/arch/arm/tools/syscallnr.sh
29:        echo "#define __NR_syscalls $nr"

/usr/src/linux-headers-5.0.0-15/include/uapi/asm-generic/unistd.h
744:#undef __NR_syscalls
745:#define __NR_syscalls 295

/usr/src/linux-headers-5.0.0-15/arch/alpha/kernel/syscalls/syscallhdr.sh
32:     printf "#define __NR_syscalls\t%s\n" "${nxt}"

/usr/src/linux-headers-5.0.0-15/arch/alpha/include/asm/unistd.h
7:#define NR_SYSCALLS   __NR_syscalls

/usr/src/linux-headers-5.0.0-15/arch/arm64/include/asm/unistd.h
56:#define NR_syscalls (__NR_syscalls)

/usr/src/linux-headers-5.0.0-15/arch/ia64/kernel/syscalls/syscallhdr.sh
32:     printf "#define __NR_syscalls\t%s\n" "${nxt}"

/usr/src/linux-headers-5.0.0-15/arch/microblaze/kernel/syscalls/syscallhdr.sh
32:     printf "#define __NR_syscalls\t%s\n" "${nxt}"

/usr/src/linux-headers-5.0.0-15/arch/m68k/kernel/syscalls/syscallhdr.sh
32:     printf "#define __NR_syscalls\t%s\n" "${nxt}"

/usr/src/linux-headers-5.0.0-15/arch/parisc/kernel/syscalls/syscallhdr.sh
32:     printf "#define __NR_syscalls\t%s\n" "${nxt}"

/usr/src/linux-headers-5.0.0-15/arch/ia64/include/asm/unistd.h
13:#define NR_syscalls          __NR_syscalls /* length of syscall table */

/usr/src/linux-headers-5.0.0-15/arch/riscv/include/asm/unistd.h
23:#define NR_syscalls (__NR_syscalls)

/usr/src/linux-headers-5.0.0-15/arch/parisc/include/asm/unistd.h
7:#define __NR_Linux_syscalls   __NR_syscalls

/usr/src/linux-headers-5.0.0-15/arch/m68k/include/asm/unistd.h
7:#define NR_syscalls   __NR_syscalls

/usr/src/linux-headers-5.0.0-15/arch/powerpc/kernel/syscalls/syscallhdr.sh
32:     printf "#define __NR_syscalls\t%s\n" "${nxt}"

/usr/src/linux-headers-5.0.0-15/arch/arm/include/asm/syscall.h
18:#define NR_syscalls (__NR_syscalls)

/usr/src/linux-headers-5.0.0-15/arch/mips/kernel/syscalls/syscallhdr.sh
32:     printf "#define __NR_syscalls\t%s\n" "${nxt}"

/usr/src/linux-headers-5.0.0-15/arch/sh/kernel/syscalls/syscallhdr.sh
32:     printf "#define __NR_syscalls\t%s\n" "${nxt}"

/usr/src/linux-headers-5.0.0-15/arch/powerpc/include/asm/unistd.h
14:#define NR_syscalls  __NR_syscalls

/usr/src/linux-headers-5.0.0-15/arch/sparc/kernel/syscalls/syscallhdr.sh
32:     printf "#define __NR_syscalls\t%s\n" "${nxt}"

/usr/src/linux-headers-5.0.0-15/arch/xtensa/kernel/syscalls/syscallhdr.sh
32:     printf "#define __NR_syscalls\t%s\n" "${nxt}"

/usr/src/linux-headers-5.0.0-15/arch/sh/include/asm/unistd.h
8:#define NR_syscalls   __NR_syscalls

/usr/src/linux-headers-5.0.0-15/arch/xtensa/include/asm/unistd.h
25:#define NR_syscalls                          __NR_syscalls

/usr/src/linux-headers-5.0.0-15/arch/sparc/include/asm/unistd.h
20:#define NR_syscalls  __NR_syscalls

/usr/src/linux-headers-5.0.0-15/arch/arc/include/uapi/asm/unistd.h
30:#define NR_syscalls  __NR_syscalls

/usr/src/linux-headers-5.0.0-15/arch/sh/include/uapi/asm/unistd_64.h
420:#define __NR_syscalls               394

/usr/src/linux-headers-5.0.0-15/arch/sh/include/uapi/asm/unistd_32.h
400:#define __NR_syscalls               383
```
</details>

gnzlbg:
@mati865 the error is (e.g. on `aarch64`):

```
error: 'SYS_arch_specific_syscall' undeclared here (not in a function); did you mean '__NR_arch_specific_syscall'?
cargo:warning=             static const  long __test_const_SYS_arch_specific_syscall_val = SYS_arch_specific_syscall;
cargo:warning=                                                                             ^~~~~~~~~~~~~~~~~~~~~~~~~
cargo:warning=                                                                             __NR_arch_specific_syscall
cargo:warning=/checkout/target/aarch64-unknown-linux-gnu/debug/build/libc-test-bf57c51a205470a0/out/main.c:29556:64: error: 'SYS_syscalls' undeclared here (not in a function); did you mean '__NR_syscalls'?
cargo:warning=             static const  long __test_const_SYS_syscalls_val = SYS_syscalls;
cargo:warning=                                                                ^~~~~~~~~~~~
cargo:warning=                                                                __NR_syscalls
```

This means that `SYS_syscalls` is not defined in the headers, but `__NR_syscalls` is. Before `glibc` used to export `SYS_syscalls` for these targets. Now it does not, even though `__NR_syscalls` still exists.

gnzlbg:
The same happens in other targets with the reserved/unused syscalls.

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -689,7 +689,7 @@ pub const NFPROTO_NETDEV: ::c_int = 5;
 
 // linux/netfilter/nf_tables.h
 cfg_if!{
-    if #[cfg(any(target_arch = "arm", target_arch = "aarch64"))] {
+    if #[cfg(target_arch = "aarch64")] {
```
mati865: All architectures use 256 now: https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/netfilter/nf_tables.h#L6

mati865: `Sparc64` runs older kernel on Debian (because of https://github.com/rust-lang/libc/commit/ac4afe746f23766204fe2545c19e18d4ffb2d4d5) https://github.com/rust-lang/libc/blob/39ca01c8b3631db48c8e390f93f62e52b51a68b9/ci/docker/sparc64-unknown-linux-gnu/Dockerfile#L1

It should be upgraded if it works.

gnzlbg:
So most build jobs are green now. I have no idea why s390x is failing, the error message is weird. For example, see:

```
cargo:warning=/checkout/target/s390x-unknown-linux-gnu/debug/build/libc-test-bf57c51a205470a0/out/main.c:24441:73: error: 'PTRACE_GETREGS' undeclared here (not in a function); did you mean 'PTRACE_GETREGSET'?
cargo:warning=             static const unsigned int __test_const_PTRACE_GETREGS_val = PTRACE_GETREGS;
cargo:warning=                                                                         ^~~~~~~~~~~~~~
cargo:warning=                                                                         PTRACE_GETREGSET
cargo:warning=/checkout/target/s390x-unknown-linux-gnu/debug/build/libc-test-bf57c51a205470a0/out/main.c:24447:73: error: 'PTRACE_SETREGS' undeclared here (not in a function); did you mean 'PTRACE_GETREGS'?
cargo:warning=             static const unsigned int __test_const_PTRACE_SETREGS_val = PTRACE_SETREGS;
cargo:warning=                                                                         ^~~~~~~~~~~~~~
cargo:warning=                                                                         PTRACE_GETREGS
```

Notice how the first error says that `PTRACE_GETREGS` is undeclared, but the suggestion of the second error does suggest using `PTRACE_GETREGS`, which can only happen if that has been declared somewhere =/

mati865:
I was also puzzled by this one. Definitions seems to be absent though:
- PPC: https://elixir.bootlin.com/linux/latest/source/arch/powerpc/include/uapi/asm/ptrace.h#L182
- S390X: https://elixir.bootlin.com/linux/latest/source/arch/s390/include/uapi/asm/ptrace.h
- architectures with the define: https://elixir.bootlin.com/linux/latest/ident/PTRACE_GETREGS

src/unix/notbsd/linux/other/mod.rs
```diff
@@ -689,12 +689,7 @@ pub const NFPROTO_NETDEV: ::c_int = 5;
 
 // linux/netfilter/nf_tables.h
 cfg_if!{
-    if #[cfg(any(target_arch = "arm", target_arch = "aarch64"))] {
```
mati865: Sparc64 was so different because it had ancient kernel, if 19.04 build succeeds this will cause an error.

gnzlbg: I'll adapt this if the 19.04 succeeds, but it appears that 19.04 fails for other reasons everywhere.

mati865: It changed values all over the place (quite expected) and sparc64 executes on old Debian with very old glibc: https://github.com/rust-lang/libc/blob/master/ci/linux-sparc64.sh#L8

gnzlbg:
Most build jobs fail again with ubuntu 19.04, the errors are of the form:

```
bad ucontext_t size: rust: 936 (0x3a8) != c 968 (0x3c8)
bad fscanf function pointer: rust: 140035897326624 (0x7f5ca5ea7c20) != c 140035897330928 (0x7f5ca5ea8cf0)
bad scanf function pointer: rust: 140035897326816 (0x7f5ca5ea7ce0) != c 140035897330688 (0x7f5ca5ea8c00)
bad sscanf function pointer: rust: 140035897327024 (0x7f5ca5ea7db0) != c 140035897331136 (0x7f5ca5ea8dc0)
bad NFT_MSG_MAX value at byte 0: rust: 22 (0x16) != c 25 (0x19)
```

The incorrect const and field size are easily fixable, but the incorrect function pointers means that Rust points to a different function than C in our tests. Maybe a different version of glibc is being pulled via the rust standard library than the one C is using ? No idea yet.

Also, the s390x build job fails with:

```
/prog: /lib/libc.so.6: version `GLIBC_2.27' not found (required by /prog)
```

but AFAICT ubuntu 19.04 ships glibc 2.29 =/

gnzlbg:
From the 2.29 release notes:

```
An archaic GNU extension to scanf, under which '%as', '%aS', and
  '%a[...]' meant to scan a string and allocate space for it with
  malloc, is now restricted to programs compiled in C89 or C++98 mode
  with _GNU_SOURCE defined.  This extension conflicts with C99's use of
  '%a' to scan a hexadecimal floating-point number, which is now
  available to programs compiled as C99 or C++11 or higher, regardless
  of _GNU_SOURCE.
```

We do define `_GNU_SOURCE` in the tests.

gnzlbg:
So I've removed the switch to 19.04 since it doesn't seem straightforward.

src/unix/notbsd/linux/other/mod.rs
gnzlbg: Ah, the sparc64 tests run on qemu-system using the disk image in that file. In that case, changing the Ubuntu version of the Docker container should have any effect right? What matters is the glibc version of the image used within qemu-system.

mati865: There is this failure:
```
cargo:warning=/checkout/target/sparc64-unknown-linux-gnu/debug/build/libc-test-bf57c51a205470a0/out/main.c:21088:28: error: returning '__int32_t *' {aka 'int *'} from a function with incompatible return type 'long int *' [-Werror=incompatible-pointer-types]
cargo:warning=                     return &b->ut_session;
cargo:warning=                            ^~~~~~~~~~~~~~
cargo:warning=/checkout/target/sparc64-unknown-linux-gnu/debug/build/libc-test-bf57c51a205470a0/out/main.c: In function '__sparc_get_pc_thunk.l7':
cargo:warning=cc1: error: unrecognized command line option '-Wno-address-of-packed-member' [-Werror]
cargo:warning=cc1: all warnings being treated as errors
```
And if you build executable on Ubuntu with glibc 2.29 it won't run on Debian with much older glibc.

gnzlbg: Is that so? I was pretty certain that glibc is dynamically linked, so as long as you only use features of glibc 2.29 that are also available in older glibc versions that should work fine.

src/unix/notbsd/linux/align.rs
```diff
@@ -7,17 +7,14 @@ macro_rules! expand_align {
...
                        repr(align(4)))]
             #[cfg_attr(not(any(target_pointer_width = "32",
                                target_arch = "x86_64",
                                target_arch = "powerpc64",
                                target_arch = "mips64",
                                target_arch = "s390x",
-                               target_arch = "sparc64",
-                               all(target_arch = "aarch64",
-                                   target_env = "musl"))),
+                               target_arch = "sparc64")),
```
mati865: Missing `target_arch = "aarch64"`?

src/unix/notbsd/linux/other/mod.rs
mati865: The thing is it tries to pick some 2.27 feature, latest build:
```
/prog: /lib/libc.so.6: version `GLIBC_2.27' not found (required by /prog)
```

mati865: Debian 10 has glibc 2.28, it could possibly work for now.

gnzlbg: So I've updated the debian images used in the sparc64 and s390x build jobs, we'll see what happens.

mati865:
Hopefully one of the last fixes:
```
cargo:warning=/checkout/target/sparc64-unknown-linux-gnu/debug/build/libc-test-bf57c51a205470a0/out/main.c:21088:28: error: return from incompatible pointer type [-Werror=incompatible-pointer-types]
cargo:warning=                     return &b->ut_session;
cargo:warning=                            ^~~~~~~~~~~~~~
```
Error on Ubuntu 19.04 was much more useful:
```
cargo:warning=/checkout/target/sparc64-unknown-linux-gnu/debug/build/libc-test-bf57c51a205470a0/out/main.c:21088:28: error: returning '__int32_t *' {aka 'int *'} from a function with incompatible return type 'long int *' [-Werror=incompatible-pointer-types]
cargo:warning=                     return &b->ut_session;
cargo:warning=                            ^~~~~~~~~~~~~~
```

https://github.com/rust-lang/libc/blob/11f30a8837d051741d0119078acabdb606fac14c/src/unix/notbsd/linux/other/mod.rs#L227-L246

gnzlbg:
So I think things are mostly green for 19.04 now. There is a breaking change where we callign `_scanf` and friends before, but are calling `__isoc99_scanf` now. These are available since glibc 2.7 (2007) so I hope this doesn't break anybody.

gnzlbg:
So everything is green now, except for the wasi build jobs which still fail with target not found (cc @alexcrichton @sunfishcode) and the android build jobs which are allowed to fail.

gnzlbg:
So I've pulled the renaming changes only from #1342 .

gnzlbg:
Now the documentation build job fails because bitrig was remove in the latest nightly, which was just released. Perfect timing.

gnzlbg:
This was, by far, the largest game of code golf I've ever played.

gnzlbg:
Thank you @mati865 for the feedback and help with the changes!

alexcrichton:
@gnzlbg FWIW if this happens in the future I'd recommend simply disabling targets on CI. I think it's best to keep the wheels turning and the problems fixed here in general should be inconsequential because the current state of the crate isn't exactly full of bugs to the point that it's unusable. By enabling PRs to land quicker it can help reduce some stress from landing big changes/refactorings like this all at the same time.

gnzlbg:
@alexcrichton The Ubuntu version that was being used went end-of-life and the Docker images for it were removed, so all build jobs using it (almost all of them) started failing. That is, we would have had to turn off CI completely. I suppose we could have done that, to be able to continue merging PRs without any testing on CI, but I don't think we should have done a new release until CI was up again, and at that point, merged PRs might have introduced errors, that I would have had to fix in the PR updating CI, so... I don't know if that would have been better.

alexcrichton:
@gnzlbg oh sure yeah I understand but it probably wasn't too hard to get the major targets working again on CI and the main difficulty was getting **everything** working, right? I was mostly just saying that this could probably get staged where the main targets are brought back up and then the remaining targets would be follow-ups.

gnzlbg:
Well, maybe? The part that took most of the time here was getting x86_64 and aarch64 gnu Linux targets working. This PR did not fix Android (there is a PR open for this) nor Emscripten, but getting the tests to pass on the Linux gnu targets without breaking the other targets more required splitting the Linux bits from the libc-test/build.rs, which was done as simple as possible (there are a million fixmes, which are now being cleaned up in other PRs). The other Linux targets (e.g. the musl ones) kind of just worked, or were easy to fix. 

The sparc64 and s390x fixes were probably unnecessary and could have been done later, but they weren't very time consuming either.

What costed most of the time is having to test this on travis-ci, which gives me a 30min iteration time, because the docker containers that we have do not work on MacOS X.

alexcrichton:
Ah ok, well whatever works works! Not being able to test locally I can also see would significantly complicate things...

Issue #1339, opened 2019-05-15
MAP_FIXED_NOREPLACE is missing on Linux


ibabushkin:
`MAP_FIXED_NOREPLACE`, documented in `mmap(2)`, is missing, but is supported by linux >= 4.17. As far as I can tell, all linux targets of versions recent enough should support this flag, here's the relevant chunk from the manpage:

```
MAP_FIXED_NOREPLACE (since Linux 4.17)
       This flag provides behavior that is similar to MAP_FIXED with
       respect to the addr enforcement, but differs in that
       MAP_FIXED_NOREPLACE never clobbers a preexisting mapped range.
       If the requested range would collide with an existing mapping,
       then this call fails with the error EEXIST.  This flag can
       therefore be used as a way to atomically (with respect to other
       threads) attempt to map an address range: one thread will
       succeed; all others will report failure.

       Note that older kernels which do not recognize the
       MAP_FIXED_NOREPLACE flag will typically (upon detecting a
       collision with a preexisting mapping) fall back to a
       "non-MAP_FIXED" type of behavior: they will return an address
       that is different from the requested address.  Therefore,
       backward-compatible software should check the returned address
       against the requested address.
```

Here's a code example:
```Rust
let base = unsafe {
    libc::mmap(
        addr,
        len,
        libc::PROT_READ | libc::PROT_WRITE,
        libc::MAP_ANON | libc::MAP_PRIVATE | libc::MAP_FIXED_NOREPLACE,
        -1,
        0,
    )
};
```

If noone has any objections, I'll go ahead and open a PR for this and #1315.

gnzlbg:
Sounds good, go ahead!

gnzlbg:
This is fixed on master.

Issue #1315, opened 2019-04-11
MAP_SHARED_VALIDATE is missing on Linux


Ekleog:
`MAP_SHARED_VALIDATE` (defined in mmap(2)) is missing in libc, for linux >= 4.15.

For instance the following code should, I think, compile:
```rust
let p = libc::mmap(
    std::ptr::null_mut(),
    4096,
    libc::PROT_READ | libc::PROT_WRITE,
    libc::MAP_SHARED_VALIDATE | libc::MAP_ANONYMOUS,
    -1,
    0,
);
```

Thank you for your time! :)

gnzlbg:
This is fixed on master.

PR #1340, opened 2019-05-15
Added MAP_FIXED_NOREPLACE and MAP_SHARED_VALIDATE consts.

ibabushkin:
This addresses #1339 and #1315. I believe the location of the constants is correct as-is, as both flags are linux-specific additions. Let me know if I missed anything.

### DIFF
--- a/ci/docker/aarch64-unknown-linux-musl/Dockerfile
+++ b/ci/docker/aarch64-unknown-linux-musl/Dockerfile
@@ -3,14 +3,14 @@ FROM ubuntu:18.04
 RUN apt-get update && apt-get install -y --no-install-recommends \
   gcc make libc6-dev git curl ca-certificates \
   gcc-aarch64-linux-gnu qemu-user
-RUN curl --retry 5 https://www.musl-libc.org/releases/musl-1.1.19.tar.gz | \
+RUN curl --retry 5 https://www.musl-libc.org/releases/musl-1.1.22.tar.gz | \
     tar xzf - && \
-    cd musl-1.1.19 && \
+    cd musl-1.1.22 && \
     CC=aarch64-linux-gnu-gcc \
     ./configure --prefix=/musl-aarch64 --enable-wrapper=yes && \
     make install -j4 && \
     cd .. && \
-    rm -rf musl-1.1.19
+    rm -rf musl-1.1.22
 # Install linux kernel headers sanitized for use with musl
 RUN curl --retry 5 -L https://github.com/sabotage-linux/kernel-headers/archive/v3.12.6-6.tar.gz | \
     tar xzf - && \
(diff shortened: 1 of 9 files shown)


ibabushkin:
Unsure why the travis build was canceled, and I don't think I can do anything about it short of a force-push.

gnzlbg:
@ibabushkin I canceled it manually because it was going to fail anyways (CI was temporarily broken due to unrelated issues).

So don't do anything, bors will take care of that (it will run the tests and if everything passes, it will merge this, and if not, it will comment here with what failed).

ibabushkin:
Alright, makes sense!

ibabushkin:
Okay, apparently musl, android and some other platforms do not expose these flags. I'll move the code and fix it.

gnzlbg:
Can you check the musl and android headers for these? Maybe we just have to update the versions that we use of those headers (there is a PR for android open).

ibabushkin:
On 2019-05-16, gnzlbg wrote:
> Can you check the musl and android headers for these? Maybe we just have to update the versions that we use of those headers (there is a PR for android open).
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/rust-lang/libc/pull/1340#issuecomment-493084418

Sure thing, will do!

ibabushkin:
Current musl does not expose `MAP_FIXED_NOREPLACE` for some time now, bumping the version to 1.1.22 should work I think, will test.

The android build failed for some unrelated reason it seems: https://travis-ci.com/rust-lang/libc/jobs/200704897

Same for emscripten: https://travis-ci.com/rust-lang/libc/jobs/200704899 https://travis-ci.com/rust-lang/libc/jobs/200704900

I haven't yet isolated the issues there.

gnzlbg:
Ok, so don't worry about Android and Emscripten for now. The build jobs for these targets are allowed to fail, so as you've noticed, they might fail for unrelated reasons.

ibabushkin:
Alright, I updated musl for most docker containers (sorry for the delay). mips and mipsel use OpenWRT toolchains though, and I haven't yet found musl versions current enough there. Is there a reason for that or does it make sense to try and use an upstream musl there as well? I suppose the issue could be difficulty obtaining a working toolchain, but I don't see that documented anywhere so far.

gnzlbg:
> mips and mipsel use OpenWRT toolchains though, and I haven't yet found musl versions current enough there.

It is possible to update this, but this would probably require doing the update in a synchronized fashion with rust-lang/rust since OpenWRT is used there to create the Rust toolchains for these targets. That is, it would be better to do that in a follow PR, were one branches libc, creates a PR to rust-lang/rust that updates this and uses the branch, makes sure bors: try passes there, and once everything is green in rust-lang/rust, we merge the libc parts here, do a quick release, update the PR upstream to use the released version, and merge that.

gnzlbg:
That is, I think it would be better to not expose these for musl targets here, merge this, and then work on the musl parts later.

ibabushkin:
Alright, I have reverted the mips-related commits. Not the updated musl toolchains for other arches though, I believe that is fine (since I also removed musl support for the two flags for now).

ibabushkin:
Alright, I'll proceed to add musl support separately :)

gnzlbg:
@ibabushkin i had to check the musl support for those targets today, and the musl versions used there are very old, so I think it would be ok to just not provide these constants for musl mips targets.

PR #1350, opened 2019-05-18
Add ttyname_r

dbrgn:
I hope this is correct. Parameter types taken from `man ttyname`.

```c
int ttyname_r(int fd, char *buf, size_t buflen);
```


gnzlbg:
This is failing on osx (bad ttyname_r function pointer: rust: 2805327834 (0xa735e7da) != c 2805003923 (0xa730f693)). This means that the rust function and the C function point to different symbols. You probably need to use `#[link_name]` on OSX (and maybe other BSD) targets to make it point to the right function.

dbrgn:
Thanks @gnzlbg!

PR #1354, opened 2019-05-21
[breaking change] gettimeofday 2nd argument incorrect in some targets

### LINKED ISSUE #1338: gettimeofday has incorrect signature
The signature of `gettimeofday` incorrect on ~~most~~ all targets:

```
error: incompatible pointer types returning 'int (struct timeval *, struct timezone *)' from a function with result type 'int (*)(struct timeval *, void *)' [-Werror,-Wincompatible-pointer-types]
cargo:warning=                return gettimeofday;
```

we should see if we can fix it. Will depend on how many crates are using it.

gnzlbg:
The second argument of `gettimeofday` was a `*mut c_void` on all targets,
but that type is incorrect in the following targets, where it should be
a `*mut timezone` instead:

On these other targets it appears that the signature of gettimeofday was incorrect (it takes a time-zone pointer instead of a void pointer):

*linux+gnu: http://man7.org/linux/man-pages/man2/gettimeofday.2.html
*freebsd: https://www.freebsd.org/cgi/man.cgi?query=gettimeofday&apropos=0&sektion=2&manpath=FreeBSD+11.2-stable&arch=default&format=html
*openbsd: https://man.openbsd.org/gettimeofday.2
*android: https://github.com/ricardoquesada/android-ndk/blob/master/usr/include/sys/time.h
*dragonfly: https://www.dragonflybsd.org/cgi/web-man?command=gettimeofday&section=2

This commit corrects the type on these targets, which is a breaking change. Due
to how this API is commonly used (e.g. passing `ptr::null_mut` to the second
argument), breakage should be minimal or non-existent (AFAICT only `time`, `libstd`, and `parking_lot` use this API, and they all should compile after this change). Users wanting to support both versions can just write `ptr as *mut _` instead.

Closes #1338.

---

On these targets, the signature of `gettimeofday` was correct (the second argument is a `void*`):

* macosx: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/gettimeofday.2.html
* linux+musl: http://git.musl-libc.org/cgit/musl/tree/include/sys/time.h#n11
* linux+newlib: https://chromium.googlesource.com/native_client/nacl-newlib/+/99fc6c167467b41466ec90e8260e9c49cbe3d13c/newlib/libc/include/sys/time.h#74
* netbsd: http://netbsd.gw.com/cgi-bin/man-cgi?gettimeofday+2.i386+NetBSD-8.0
* newlib: https://github.com/devkitPro/newlib/blob/devkitA64/newlib/libc/include/sys/time.h#L370
* solaris/illumos: https://illumos.org/man/3C/gettimeofday
* emscripten: https://chromium.googlesource.com/external/github.com/kripken/emscripten/+/1.35.20/system/include/libc/sys/time.h#11

cc @alexcrichton

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -315,9 +315,6 @@ fn test_openbsd(target: &str) {
         match name {
             "execv" | "execve" | "execvp" | "execvpe" => true,
 
-            // typed 2nd arg
-            "gettimeofday" => true,
-
             // Removed in OpenBSD 6.5
             // https://marc.info/?l=openbsd-cvs&m=154723400730318
             "mincore" => true,
@@ -1113,9 +1110,8 @@ fn test_dragonflybsd(target: &str) {
 
             "getrlimit" | "getrlimit64" |    // non-int in 1st arg
             "setrlimit" | "setrlimit64" |    // non-int in 1st arg
-            "prlimit" | "prlimit64" |        // non-int in 2nd arg
-            // typed 2nd arg on linux
-            "gettimeofday" => true,
+            "prlimit" | "prlimit64"        // non-int in 2nd arg
+             => true,
 
             _ => false,
         }
@@ -1460,10 +1456,6 @@ fn test_android(target: &str) {
             "execvp" |
             "execvpe" |
             "fexecve" => true,
-
-            // typed 2nd arg on android
-            // FIXME: still necessary?
-            "gettimeofday" => true,
 
             // not declared in newer android toolchains
             // FIXME: still necessary?
@@ -1814,9 +1806,6 @@ fn test_freebsd(target: &str) {
             "execvp" |
             "execvpe" |
             "fexecve" => true,
-
-            // FIXME: for some reason, our signature is wrong
-            "gettimeofday" => true,
 
             // The `uname` function in freebsd is now an inline wrapper that
             // delegates to another, but the symbol still exists, so don't check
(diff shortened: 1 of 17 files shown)


gnzlbg:
@semarie thoughts about directly doing a breaking change for OpenBSD ?

@strangelittlemonkey @leo60228 ping on the breaking changes for newlib and dragonflybsd - you might need to patch libstd after this is released.

gnzlbg:
So I just tried to update the crates depending on `gettimeofday` on the broken platforms, and there is nothing to update (most crates will continue to compile after the breaking change due to how this APIs are used) so I'm going to not provide a fallback here.

gnzlbg:
cc @asomers since this is a breaking change on FreeBSD.

semarie:
no problem for me with a breaking change. but it seems to me that if you removed the bad definition of `gettimeofday`, you not added it back for openbsd.

alexcrichton:
I wouldn't personally consider this worth a breaking change, but I think it's fine to land. I do think though that it requires being pretty vigilant after the release and ready to revert if it causes undue breakage.

gnzlbg:
I'll do a release with this change only once its ready, and submit a PR to rust-lang/rust to update libc.

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1249,6 +1249,8 @@ f! {
 }
 
 extern {
+    pub fn gettimeofday(tp: *mut ::timeval,
```
asomers: Why not combine FreeBSD's and Dragonfly's definitions in src/unix/bsd/freebsdlike/mod.rs ?

leo60228:
Looks good. FWIW, that's the Linux-specific time header in newlib, the actual one is here: https://github.com/devkitPro/newlib/blob/devkitA64/newlib/libc/include/sys/time.h#L370

leo60228:
Wait, I misunderstood what the commit did. It should be void* on newlib.

gnzlbg:
thanks @leo60228 for taking a look, i've updated the PR, rebased, and all seems green. 

@bors: r+

PR #1355, opened 2019-05-22
Refactor OpenBSD-like module into OpenBSD

gnzlbg:
The only OpenBSD-like OS that we support after bitrig's removal is OpenBSD itself.

cc @semarie


gnzlbg:
This is a non-functional change, just moves code around. 

@bors: r+

PR #1358, opened 2019-05-23
Add powerpc64-unknown-freebsd target

pkubaj:
Per https://github.com/rust-lang-nursery/stdsimd/pull/765 add powerpc64-unknown-freebsd to prevent further breakages.

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -61,7 +61,6 @@ test_target() {
 }
 
 RUST_LINUX_TARGETS="\
-aarch64-unknown-freebsd \
 aarch64-linux-android \
 aarch64-unknown-linux-gnu \
 arm-linux-androideabi \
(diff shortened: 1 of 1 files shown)


gnzlbg:
Thanks for the PR, you might want to add here aarch64 as well. 

FYI, this repo has nothing to do with stdsimd. My comment there was suggesting setting up a similar system as the one set up here, but on that repo, to be able to test these targets there as well.

gnzlbg:
You need to add these targets to the `RUST_NIGHTLY_LINUX_TARGETS` variable, since these are only available on nightly.

gnzlbg:
Damn, they need to be in the "LINUX_NO_CORE_TARGETS" apparently: https://github.com/rust-lang/libc/pull/1358/files#diff-78b1794b72c7105dbabbe466019395d6L170

gnzlbg:
Thank you! Now it worked! 

@bors: r+

PR #1360, opened 2019-05-23
Add final rtnetlink constants

jbaublitz:
This is a follow up to the PR #1351 which is for issue #1059 as it was closed before I realized there's one last set of netlink constants needed for use with rtnetlink. I'm marking this as WIP until I've verified everything's good with CI and that I've included all necessary constants. I'll remove WIP once it's passing CI and it's ready for review.

### DIFF
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -1830,6 +1830,22 @@ pub const NDA_VNI: ::c_ushort = 7;
 pub const NDA_IFINDEX: ::c_ushort = 8;
 
 // linux/rtnetlink.h
+pub const TCA_UNSPEC: ::c_ushort = 0;
+pub const TCA_KIND: ::c_ushort = 1;
+pub const TCA_OPTIONS: ::c_ushort = 2;
+pub const TCA_STATS: ::c_ushort = 3;
+pub const TCA_XSTATS: ::c_ushort = 4;
+pub const TCA_RATE: ::c_ushort = 5;
+
+// possibly should go in linux/other/mod.rs
+pub const TCA_FCNT: ::c_ushort = 6;
+pub const TCA_STATS2: ::c_ushort = 7;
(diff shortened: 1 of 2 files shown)


gnzlbg:
good bot

jbaublitz:
@gnzlbg All right, should be good to go. Thanks!

jbaublitz:
retry

jbaublitz:
Killjoy bot :slightly_frowning_face:

skrap:
@jbaublitz: was there a reason that you did not include the RTM message type defines in here?  Like RTM_NEWLINK, RTM_DELLINK, etc?

jbaublitz:
@skrap Oversight on my part due to the formatting for the `man 7 rtnetlink` docs. This should be included. I'll put together another PR for `libc` and (as I'm guessing this is why you're asking) I'll provide support in [neli](https://github.com/jbaublitz/neli).

skrap:
Hah, yeah, you're onto me 😆 .  I'd be more than happy to help out in some way — perhaps I could provide something to add to the neli examples collection using some of those new constants?

jeromegn:
I probably don't know enough, but is there a reason these constants can't be added for musl? If I had to guess: they're not defined in musl.

Trying to use rtnetlink with a musl target is not possible right now.

jbaublitz:
@jeromegn Are you using libc directly? If so, only certain constants should not be usuable from musl. The rest are in both GNU and musl.

jeromegn:
I'm not, but rtnetlink requires constants such as `NETLINK_CAP_ACK`. Sorry, this is likely the wrong place to write about this (I realize this now).

I should look into neli and see if it compiles on x86 64 unknown musl.

jbaublitz:
I've tried to make neli compatible with musl because of what I learned doing this PR so please file an issue with neli and I'd be happy to take a look if it doesn't work as expected. If you need additional constants through neli, also file an issue there and I'll do another PR to libc.

Issue #657, opened 2017-07-09
Rename use_std feature to just std


dtolnay:
As recommended by https://github.com/brson/rust-api-guidelines#c-feature. This aligns better with the naming of implicit features inferred by Cargo.

roblabla:
What's the use-case for the `use_std` flag ? Couldn't the crate always be no_std? The only use of that feature flag is at https://github.com/rust-lang/libc/blob/master/src/unix/mod.rs#L286. What happens if we remove that check?

alexcrichton:
Changing it can be a breaking change in some subtle situations so it wasn't turned off by deafult, but for the next major release it'll be turned off by default.

gnzlbg:
Since this feature is enabled by default, this would only be a breaking change iff `--no-default-features --features use_std` is being used, which would be a weird configuration, since doing nothing has the same effect. 

I have a proposal here. 

We could add a `std` feature, in parallel with `use_std`, that is also enabled by default. Then, in the `build.rs`, we could translate `use_std` to `std`, and start warning on its usage. Internally, in the library, we would just use `std` as a feature.

We might not be able to remove `use_std` ever (I know a lot of crates re-exporting `use_std = ["libc/use_std"]`), but we can at least already implement a proper migration path, and start telling users to migrate to it. 

The question is whether doing this is worth it or not.

gnzlbg:
Now the `std` feature can be used as a `cargo` feature, and the `use_std` feature has been deprecated (produces a cargo warning). I'm repurposing this issue to track the removal of the `use_std` feature, which would be a breaking change that will probably have to wait till the next breaking libc release.

mati865:
Is `std` feature necessary?
It caused massive issues upgrading `rand` in Rust repo: https://github.com/rust-lang/rust/pull/61393
More details at: https://github.com/rust-random/rand/issues/852

Short summary:
Somehow when building Rust newer `rand` versions which use more of `libc` were failing during linking of `rustc_main`:
```
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/Scrt1.o: undefined reference to symbol '__libc_start_main@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libc.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```
It was found this can be avoided by disabling `std` feature of `libc` causing it to pull `libc.so` and Rust built fine.

[...]

PR #1361, opened 2019-05-24
Deprecate `use_std` cargo feature: use `std` instead .

gnzlbg:
Related to #657 .


gnzlbg:
cc @dtolnay this essentially fixes #657 , once we release the next breaking version of libc, we can remove the `use_std` feature, completing the fix.

PR #1365, opened 2019-05-24
[breaking change] Cleanup linux and update MUSL

gnzlbg:
* Update MUSL kernel headers to 4.4.2 (non-breaking)
* [breaking] `MADV_SOFT_OFFLINE` is not defined on MIPS
* [breaking] `sendmmsg`/`recvmmsg` take an `unsigned int` flag on MUSL
* [breaking] `pthread_t` is a pointer on MUSL
* `rlimit` resources should use a type alias on GNU (non-breaking)
* Deprecate `SIGNUNUSED` (should use SIGSYS instead)

### DIFF
--- a/ci/install-musl.sh
+++ b/ci/install-musl.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env sh
+#
+# Install musl and musl-sanitized linux kernel headers
+# to musl-{$1} directory
+
+set -ex
+
+MUSL_VERSION=1.1.22
+MUSL="musl-${MUSL_VERSION}"
+
+# Download, configure, build, and install musl:
+curl --retry 5 https://www.musl-libc.org/releases/${MUSL}.tar.gz | tar xzf -
+
+cd $MUSL
+case ${1} in
+    aarch64)
+        musl_arch=aarch64
+        kernel_arch=arm64
+        CC=aarch64-linux-gnu-gcc \
+          ./configure --prefix="/musl-${musl_arch}" --enable-wrapper=yes
+        make install -j4
+        ;;
+    arm)
+        musl_arch=arm
+        kernel_arch=arm
+        CC=arm-linux-gnueabihf-gcc CFLAGS="-march=armv6 -marm -mfpu=vfp" \
+          ./configure --prefix="/musl-${musl_arch}" --enable-wrapper=yes
+        make install -j4
+        ;;
+    i686)
+        # cross-compile musl for i686 using the system compiler on an x86_64
+        # system.
+        musl_arch=i686
+        kernel_arch=i386
+        # Specifically pass -m32 in CFLAGS and override CC when running
+        # ./configure, since otherwise the script will fail to find a compiler.
+        CC=gcc CFLAGS="-m32" \
+          ./configure --prefix="/musl-${musl_arch}" --disable-shared --target=i686
+        # unset CROSS_COMPILE when running make; otherwise the makefile will
+        # call the non-existent binary 'i686-ar'.
+        make CROSS_COMPILE= install -j4
+        ;;
+    x86_64)
+        musl_arch=x86_64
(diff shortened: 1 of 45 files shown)


gnzlbg:
So IPV6_FLOWINFO is here: https://elixir.bootlin.com/linux/v5.0/source/include/uapi/linux/in6.h#L165 
and that header `"linux/in6.h"`  is included, but the tests fail with "undeclared ...". @mati865

gnzlbg:
ah, duh, it appears that these are only included if the linux headers are included first -.-

mati865:
I don't fully understand libc testsuite, it's a bit complicated for newcomer.
Anyway don't worry to ping me if you have issue with something.

ci/install-musl.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# Download, configure, build, and install musl-sanitized kernel headers:
+curl --retry 5 -L \
+     https://github.com/sabotage-linux/kernel-headers/archive/v4.4.2-2.tar.gz | \
+    tar xzf -
+(
+    cd kernel-headers-4.4.2-2
+    make ARCH="${kernel_arch}" prefix="/musl-${musl_arch}" install -j4
+)
+rm -rf kernel-headers-4.4.2-2
```
mati865: It would make sense to move all those 3 `4.4.2-2` to a variable.

gnzlbg:
I think the current way things are structured is quite weird, but maybe that's life. 

For example, we have linux as a platform, and then `musl`, `!musl+mips`, `!musl+s390x`, and `!musl` (called `other`) as sub-modules. Where AFAICT `!musl` is just `gnu`. From the code `!musl` could also be `uclib`, `newlib`, etc. but since none of those are tested chances are none of those is working anyways.

So i'm tending towards just splitting the linux module into two modules: musl, and gnu, moving mips and s390x into the gnu submodule. If `target_env` is not `musl` or `gnu` for linux, we should error, and if there is someone using a different `target_env` on linux, they can just add it in parallel, instead of adding all sorts of workarounds to the `gnu` module.

mati865:
Seems reasonable, chances for any other Linux libc to be fully compatible with `gnu` are close to zero.

gnzlbg:
cc @asomers @alexcrichton

apmanol:
nix-0.14 broke too.

```
info: component 'rust-std' for target 'mips-unknown-linux-gnu' is up to date
   Compiling nix v0.14.0
error[E0425]: cannot find value `MADV_SOFT_OFFLINE` in module `libc`
   --> /.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.14.0/src/sys/mman.rs:143:9
    |
143 |         MADV_SOFT_OFFLINE,
    |         ^^^^^^^^^^^^^^^^^ not found in `libc`
help: possible candidate is found in another module, you can import it into scope
    |
1   | use sys::mman::MmapAdvise::MADV_SOFT_OFFLINE;
    |

error[E0081]: discriminant value `14` already exists
   --> /.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.14.0/src/macros.rs:197:26
    |
197 |                   $entry = libc::$entry,
    |                            ^^^^
    |                            |
    |                            first use of `14`
    |                            enum already has `14`
    |
   ::: /.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.14.0/src/sys/mman.rs:105:1
    |
105 | / libc_enum!{
106 | |     /// Usage information for a range of memory to allow for performance optimizations by the kernel.
107 | |     ///
108 | |     /// Used by [`madvise`](./fn.madvise.html).
...   |
187 | |     }
188 | | }

error: aborting due to 2 previous errors


```

gnzlbg:
> MADV_SOFT_OFFLINE not found

That's a nix error, that constant is not available on MIPS on any released Linux kernel.

PR #1371, opened 2019-05-27
Manually implement extra traits for `mq_attr` and `sockaddr_nl`

### LINKED ISSUE #1302: mq_attr checks __reserved and pad fields in PartialEq comparison
We run tests on mq_attr, and moving from our `PartialEq` implementation to `libc`'s broke those tests because these fields are compared. They are only used for alignment on some platforms and can contain junk data on some platforms. These fields should be skipped when doing a comparison of these structs.

Susurrus:
Avoid including padding fields in extra trait implementations as these fields aren't guaranteed to be 0 or some other sensible value.

Closes #1302

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -366,13 +366,6 @@ s! {
         pub ai_next: *mut addrinfo,
     }
 
-    pub struct sockaddr_nl {
-        pub nl_family: ::sa_family_t,
-        nl_pad: ::c_ushort,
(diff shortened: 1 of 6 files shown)


Susurrus:
cc @asomers

gnzlbg:
@bors: r+ 

---

We should probably just use "special" padding types that derive trivial implementations of the traits (e.g. `PartialEq` always returns true, etc.).

PR #1372, opened 2019-05-28
Add F_SEAL_FUTURE_WRITE on Linux/Android

tklauser:
This was added in Linux 5.1 and will only show up in the next glibc
release, thus skip in tests.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2266,6 +2266,8 @@ fn test_linux(target: &str) {
             // Require Linux kernel 5.x:
             | "MSG_COPY"
                if musl  => true,
+            // Require Linux kernel 5.1:
+            "F_SEAL_FUTURE_WRITE" => true,
 
             // The musl version 1.0.22 used in CI does not
             // contain these glibc constants yet:
--- a/src/unix/notbsd/linux/mod.rs
+++ b/src/unix/notbsd/linux/mod.rs
@@ -1050,6 +1050,8 @@ pub const F_TEST: ::c_int = 3;
 pub const F_TLOCK: ::c_int = 2;
 pub const F_ULOCK: ::c_int = 0;
 
+pub const F_SEAL_FUTURE_WRITE: ::c_int = 0x0010;
+
 pub const IFF_LOWER_UP: ::c_int = 0x10000;
 pub const IFF_DORMANT: ::c_int = 0x20000;
 pub const IFF_ECHO: ::c_int = 0x40000;


gnzlbg:
Thank you for the PR. This LGTM. I prefer to merge this after #1365 lands, so after that's merged this will need to be rebased on top of it.

src/unix/notbsd/mod.rs
```diff
@@ -465,6 +465,7 @@ pub const F_SEAL_SEAL: ::c_int = 0x0001;
 pub const F_SEAL_SHRINK: ::c_int = 0x0002;
 pub const F_SEAL_GROW: ::c_int = 0x0004;
 pub const F_SEAL_WRITE: ::c_int = 0x0008;
+pub const F_SEAL_FUTURE_WRITE: ::c_int = 0x0010;
```
gnzlbg: Is this available in emscripten ?

tklauser: Not yet, I think. It also seems to be missing from musl currently. What is the proper place to put this in that case? `src/unix/notbsd/linux/other/mod.rs` and a copy to `src/unix/notbsd/android/mod.rs`?

gnzlbg: So one copy goes into `android/mod.rs`, another copy goes into `linux/mips/mod.rs`/`linux/s390x/mod.rs`, `linux/other/mod.rs` - basically everywhere in `linux` except `linux/musl/mod.rs`.

gnzlbg:
The build.rs has been substantially changed. There is a section in `skip_const` for Linux where we document what's not available in the current kernel headers used in CI. You probably want to add there a new "section" for the 5.1 kernel, adding this constant to the list.

libc-test/build.rs
```diff
@@ -1378,6 +1378,8 @@ fn test_android(target: &str) {
             "SIG_DFL" | "SIG_ERR" | "SIG_IGN" => true, // sighandler_t weirdness
             // FIXME: still necessary?
             "SIGUNUSED" => true, // removed in glibc 2.26
+            // FIXME: remove once Android uses kernel headers >= 5.1 which expose this constant
+            "F_SEAL_FUTURE_WRITE" => true,
```
gnzlbg: I would prefer to just not expose this on Android for now.

tklauser: Done

tklauser:
Thanks @gnzlbg. Now pushed a rebased version and integrated your review comments.

libc-test/build.rs
```diff
@@ -2283,6 +2283,7 @@ fn test_linux(target: &str) {
 
             // FIXME: conflicts with glibc headers and is tested in
             // `linux_termios.rs` below:
+
```
gnzlbg: spurious new line

libc-test/build.rs
```diff
@@ -2293,6 +2294,10 @@ fn test_linux(target: &str) {
             t if mips32_musl && t.starts_with("IFF") => true,
             "MFD_HUGETLB" | "AF_XDP" | "PF_XDP" if mips32_musl => true,
 
+            // FIXME: added in Linux 5.1 but not yet in glibc or musl. Enable in linux_fcntl tests
+            // once available.
+            "F_SEAL_FUTURE_WRITE" => true,
```
gnzlbg: Search for "// Require Linux kernel 5.x:" above. Could you add this there, e.g. adding a new "// Require Linux kernel 5.1" followed by this constant ?

libc-test/build.rs
tklauser: done

libc-test/build.rs
tklauser: done

gnzlbg:
Thank you!

libc-test/build.rs
```diff
@@ -2266,6 +2266,8 @@ fn test_linux(target: &str) {
             // Require Linux kernel 5.x:
             | "MSG_COPY"
                if musl  => true,
+            // Require Linux kernel 5.1:
+            "F_SEAL_FUTURE_WRITE" => true,
```
tklauser: Sorry, I have a bit of a hard time understanding the match logic here (I'm not a very proficient in rust, yet). Should it be:

`"F_SEAL_FUTURE_WRITE" => true,`

or

`| "F_SEAL_FUTURE_WRITE" => true,`

?

gnzlbg: The first `|` is optional: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3007c8d7c8a3b7d425ce1580c8cc51ed

tklauser: I see, thanks for the explanation.

PR #1378, opened 2019-05-29
Add statx on Linux

### LINKED ISSUE #1178: Add statx linux system call so we can implement fs::Metadata::created on Linux

gnzlbg:
Closes #1178.


ariasuni:
Thanks a lot ^^

PR #1379, opened 2019-05-29
Deprecate fixed width integer type aliases

### LINKED ISSUE #1304: Deprecate the fixed-width integer aliases

gnzlbg:
cc @emilio - I think it makes sense to ensure that the latest released version of bindgen works properly with this change. That is, that even when asked to use C types from, e.g., `libc::`, it does not use aliases for the fixed-width integer C types (e.g. `libc::int64_t`) but uses Rust primitive types instead (e.g. `u64`).

Closes #1304 .


emilio:
Bindgen should never generate these by default.

PR #1384, opened 2019-06-02
Upgrade rlimit consts

gnzlbg:
cc @aidanhs - this should fix the error you are seeing.


gnzlbg:
@aidanhs 0.2.58 has been released with this fix. It would be nice to hear whether this fixes your issue.

PR #1386, opened 2019-06-02
Fix libstd compilation for arm-uclibc target

red75prime:


gnzlbg:
Is this target in ci/build.sh? If not, it should be.

red75prime:
> Is this target in ci/build.sh? If not, it should be.

It is for arm*-unknown-linux-uclibcgnueabi set of targets and no, they aren't in ci/build.sh. The changes in my PR is pretty benign and I verified them manually. But I'm still struggling to get libc-test running.

gnzlbg:
So I think this PR should at least add these to the “NOCORE” part of
ci/build.sh to verify that things at least build.

For libc-test, it would help if you can upload the output to a gist.

>

red75prime:
> I think this PR should at least add these to the “NOCORE” part of ci/build.sh to verify that things at least build.

Where target definition files (armv5te-unknown-linux-uclibcgnueabi.json and armv7-[...].json) should be placed? Or is there another procedure for adding targets?

gnzlbg:
Is this target in rust-lang/rust/src/rustc_target/spec? If it is, adding it
to the list here should be enough.

If it isn’t then you can add whatever you need to the ci/ directory, and
for example check how the Nintendo Switch target does things.

src/unix/uclibc/arm/mod.rs
```diff
@@ -611,6 +618,358 @@ pub const _SC_V6_LP64_OFF64: ::c_int = 0xb2;
 pub const _SC_V6_LPBIG_OFFBIG: ::c_int = 0xb3;
 pub const _SC_XOPEN_STREAMS: ::c_int = 0xf6;
 
+// Syscall table is copied from src/unix/notbsd/linux/musl/b32/arm.rs
+pub const SYS_restart_syscall: ::c_long = 0;
```
gnzlbg: These are Linux syscalls. Are the uclibc targets only intended for use on Linux? @alexcrichton @malbarbo ? If so, these should be in a `notbsd/linux/uclibc` folder probably, and not here.

alexcrichton: AFAIK uclibc has no official support nor ever been built in rust-lang/rust. These targets I think were just contributed by community porters, I don't know how accurate and/or the status of them are.

gnzlbg:
@bors: r+

---

There are no build tests for this, so we can't really guarantee that the next changes to libc won't break building for this target. These aren't hard to add, but that can be done in a subsequent PR. If this works for you, and doesn't break anything, I think it is ok to add it as is. This particular target doesn't have too many stakeholders.

PR #1388, opened 2019-06-03
Add more posix message queue functions for Linux and all for solarish

tormol:
I'm not able to run the tests locally (likely due to too old headers on Ubuntu 18.04), but I've tested my [posixmq crate](https://github.com/tormol/posixmq/tree/0decb7c82dd472024c7e90fd8121671881a81ae2) with these functions on Linux and OmniOS.

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -2294,10 +2294,20 @@ extern {
                       msg_ptr: *mut ::c_char,
                       msg_len: ::size_t,
                       msq_prio: *mut ::c_uint) -> ::ssize_t;
+    pub fn mq_timedreceive(mqd: ::mqd_t,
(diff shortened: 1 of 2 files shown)


gnzlbg:
Linux test are passing, but we don't test Illuminos, Solaris, etc. on CI yet. @jasonbking would you be able to see if these tests pass for you there as well ?

jasonbking:
`cargo test` was successful on illumos -- was there any other specific testing you were interested in?

gnzlbg:
Did that run `libc-test` tests? Otherwise you might need to do `cargo test --all`, or `cargo test --manifest-path libc-test/Cargo.toml`.

jasonbking:
I'll look more this evening, but the ctest crate is throwing errors when building (because of LLVM issues, I'm still on 1.33, so not sure if that makes a difference).

PR #1389, opened 2019-06-03
Rename the notbsd module to linux_like

gnzlbg:


kleimkuhler:
@gnzlbg Hey I'm running into an error where `notbsd::...` is expected, but `linux_like::...` is found. I've replicated it in [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=97367c7afa15efc9bc6d5b0994b6a31c).

Any idea if this is an error in trying to make a `SocketAddr` or some version conflict?

gnzlbg:
@kleimkuhler you should open an issue in rust-lang/rust since that looks like a bug in libstd. 

Essentially, the types exposed by `libc` are not the same types exposed by `libstd`. Often, `libstd` has newtypes for them (e.g. `pthread_t`), but sometimes someone made a bug somewhere.

gnzlbg:
Looking at the source, it appears that libstd does properly have a newtype here, the fields of that type are private, so you should never see a type error here, but a privacy error instead. 

I think this is worth reporting as an error to the Rust compiler. The error message is confusing.

gnzlbg:
I've reported it here: https://github.com/rust-lang/rust/issues/64684

kleimkuhler:
@gnzlbg Oh okay thanks for the explanation on that--and reporting it!

PR #1393, opened 2019-06-07
Re-add PTRACE_DETACH for mips64 GNU targets

Susurrus:
During the refactor in 4bd419eb, `PTRACE_DETACH` was accidentally removed on
these targets. This re-adds it.

Fixes brokenness encountered during nix-rust/nix#1035

CC @asomers


gnzlbg:
@bors: r+ 

Thanks!

PR #1394, opened 2019-06-09
Switch to manual trait impls for sigevent

Susurrus:
`sigevent` on most platforms have padding or unused fields. Rather
than display those in the `Debug` impl by deriving it, manually implement
all `extra_traits` instead ignoring those fields.

I do worry that my `PartialEq` implementations for this for some platforms (like Linux) is not correct due to ignoring bytes that shouldn't be ignored because these structs don't have a proper union set up.

cc @asomers 
Part of nix-rust/nix#1035

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -432,15 +432,6 @@ s! {
         pub int_n_sign_posn: ::c_char,
     }
 
-    pub struct sigevent {
-        pub sigev_value: ::sigval,
-        pub sigev_signo: ::c_int,
-        pub sigev_notify: ::c_int,
-        pub sigev_notify_function: fn(::sigval),
-        pub sigev_notify_attributes: *mut pthread_attr_t,
-        pub __pad: [::c_char; 56 - 3 * 8 /* 8 == sizeof(long) */],
-    }
-
     pub struct rlimit64 {
         pub rlim_cur: rlim64_t,
         pub rlim_max: rlim64_t,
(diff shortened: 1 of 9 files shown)


gnzlbg:
FWIW it might be that `sigevent` is broken anyways (there are few types that we skip testing everywhere), but `sigevent`, `sig_t`/`sighandler_t`, etc. are usually amongst them. 

Looking at the type, does the function pointer field need to always be a non-null pointer ? If one can set that field to NULL in C, then this needs to be an Option in Rust, etc.

Susurrus:
@gnzlbg Are you referring to the `sigev_notify_function` field that's only exposed on Fuchsia? The function pointer fields for all other platforms are private members, so can't be called even if they're initialized properly. I can change all of these fields to be `Option`'s if that's what you want, however then I think we need to use `NonNull` if we want to preserve the layout guarantees of these structs.

gnzlbg:
> Are you referring to the sigev_notify_function field that's only exposed on Fuchsia?

Not only, in other platforms there are fields with type `*mut c_void` that are actually function pointers.

> I can change all of these fields to be Option's if that's what you want, however then I think we need to use NonNull if we want to preserve the layout guarantees of these structs.

Function pointers (`fn()->()`) in Rust are always NonNull. That is, initializing one to zero invokes undefined behavior instantaneously. an `Option<fn()->()>` has the same layout of a pointer, and `Option::None` represents the null pointer. 

---

Another issue I am seeing is that these function pointers are not `extern "C"`. So they should actually probably be of type `Option<extern "C" unsafe fn(...)->...>`.
Also, these function pointers are not `extern "C"

PR #1395, opened 2019-06-11
Mark msghdr.{_pad1, pad2} as pub

jiangliu:
Mark msghdr.{pad1, pad2} as pub, so we could initialize a msghdr object.


gnzlbg:
You can initialize this object by using `mem::zeroed`.

jiangliu:
Thanks, I have solved my issue with the suggestion.
BTW, I don't know how to fix the failed test case, should I close this PR?

PR #1399, opened 2019-06-13
Add getrandom() bindings on Linux

### LINKED ISSUE #659: getrandom(2) is missing from the bindings
We have `getentropy`, but not `getrandom`

josephlr:
Closes #659

### DIFF
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -2195,11 +2195,6 @@ extern {
     pub fn inotify_add_watch(fd: ::c_int,
                              path: *const ::c_char,
                              mask: u32) -> ::c_int;
-    pub fn getrandom(
-        buf: *mut ::c_void,
-        buflen: ::size_t,
-        flags: ::c_uint,
-    ) -> ::ssize_t;
 }
 
 cfg_if! {
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -2196,7 +2196,7 @@ extern {
                              path: *const ::c_char,
                              mask: u32) -> ::c_int;
     pub fn getrandom(
-        buf: *mut u8,
+        buf: *mut ::c_void,
         buflen: ::size_t,
         flags: ::c_uint,
     ) -> ::ssize_t;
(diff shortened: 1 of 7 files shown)


josephlr:
@gnzlbg I initially also added bindings for Android, but it looks like their libc doesn't have the `getrandom()` *function* even though it has the `getrandom(2)` syscall. See [this LWN article](https://lwn.net/Articles/711013/) on why this split exists.

josephlr:
@gnzlbg this seems to be working on glibc and musl. However, I'm getting failures, but only on 32 MIPS musl targets ([here](https://travis-ci.com/rust-lang/libc/jobs/208000795) and [here](https://travis-ci.com/rust-lang/libc/jobs/208000798)). Any idea why this is happening? Are we using an older version of musl for mips32?

josephlr:
So I decided to just remove musl support for the time being. While `getrandom()` [was added to glibc](https://lwn.net/Articles/711013/) in [version 2.25](https://sourceware.org/glibc/wiki/Release/2.25) (Feb 2017), musl [only added such an interface](https://github.com/ifduyue/musl/commit/e20658209177667e490c661dfd35b976749ef3f7) in [version 1.1.20](https://www.musl-libc.org/oldversions.html) (Sept 2018). So it might be too new for us to support it.

> However, I'm getting failures, but only on 32 MIPS musl targets (here and here). 

This was happening because the the libc CI uses a new musl for all targets except for mips/mipsel, where it uses [an older OpenWRT toolchain hosted in an AWS bucket](https://github.com/rust-lang/libc/blob/e0ff1e68b9e34173e9c4c3217d1b0fc81a7d352d/ci/docker/mips-unknown-linux-musl/Dockerfile#L9-L12). However, [OpenWRT's archive only seems to have musl 1.1.16](https://archive.openwrt.org/snapshots/trunk/ar71xx/generic/). So unless there's a way to get a new musl toolchain for mips, we won't be able to support `getrandom()` for musl on mips.

gnzlbg:
> We won't be able to support getrandom() for musl on mips.

We should support this for glibc, and for the musl targets that currently use a newer musl version. The OpenWRT toolchains for the mips targets don't get updated very often, but it does not make sense to penalize other musl users because of this.

gnzlbg:
> but it does not make sense to penalize other musl users because of this.

To be more specific, could you add this to each of the `musl/b32/...` and `musl/b64/...` modules, and remove it only from those for which CI fails due to an old musl version ?

josephlr:
> To be more specific, could you add this to each of the `musl/b32/...` and `musl/b64/...` modules, and remove it only from those for which CI fails due to an old musl version ?

Done

josephlr:
@gnzlbg Everything is green

josephlr:
@gnzlbg It looks like @bors didn't hear you.

Issue #1410, opened 2019-06-26
Linux: ucontext_t definition is only correct for glibc 2.28 and above


acfoltzer:
This commit introduced a shadow stack `__ssp` field to the definition of `ucontext_t`: https://github.com/rust-lang/libc/commit/4497a7867d56ef61194e4cbb63ef4c4fe146e4c6

However, that field is only present in glibc 2.28 and above, so this struct doesn't work correctly on distros with earlier versions, including Ubuntu 18.04 where I encountered memory corruption after updating from 0.2.47 to 0.2.58.

gnzlbg:
How is code compiled for glibc < 2.28 ABI compatible with >= 2.28 ?

acfoltzer:
Sorry, I must not be understanding your question. I'm on a < 2.28 distro, compiling and running against < 2.28; I currently don't use any >= 2.28 distros.

gnzlbg:
If you compile a binary against glibc < 2.28 headers, ucontext_t has a smaller size than if one uses glibc >= 2.28. AFAIK, you can compile against glibc < 2.28 and then use that binary in a system where glibc >= 2.28 is dynamically linked. I was wondering how does this work here?

E.g., `getcontext` on glibc >= 2.28 might try to write to the extra field of `ucontext_t`, but if your library used older headers, it wouldn't have the field, so that would be a write out-of-bounds. 

Maybe there are new different symbols for newer `getcontext`, so that if you compile your binary against an older libc, and dynamically link against a new version, the older symbols get picked, and these do not write to the extra field ? If so, then we could fix this by using `link_name` to make these APIs link to their older symbols.

gnzlbg:
Relevant commit: https://github.com/bminor/glibc/commit/25123a1c5c96429d70e75b85a9749b405909d7f2

So apparently that's allowed, see: https://github.com/bminor/glibc/commit/25123a1c5c96429d70e75b85a9749b405909d7f2

gnzlbg:
Do you have a minimal working example showing the issue ?

gnzlbg:
(This issue might be related: https://github.com/nix-rust/nix/issues/1092)

acfoltzer:
So, this is coming up in a case where a Rust program is writing a `libc::ucontext_t` to a pointer to a `ucontext_t` provided by C. Since the size of the struct is smaller on the C side, when we write to it using the Rust type we write off the end into memory we shouldn't have access to.

[...]

PR #1411, opened 2019-06-26
Remove new field from ucontext_t for compatibility with earlier glibc versions

acfoltzer:
Per discussion in #1410 with @gnzlbg, this is necessary to avoid struct size mismatches between Rust and C on systems with glibc < 2.28.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -1950,6 +1950,7 @@ fn test_linux(target: &str) {
                "syslog.h",
                "termios.h",
                "time.h",
+               "ucontext.h",
                "unistd.h",
                "utime.h",
                "utmp.h",
@@ -1967,10 +1968,6 @@ fn test_linux(target: &str) {
                // <execinfo.h> is not supported by musl:
                // https://www.openwall.com/lists/musl/2015/04/09/3
                [!musl]: "execinfo.h",
-               // ucontext_t added a new field as of glibc 2.28; our struct definition is
-               // conservative and omits the field, but that means the size doesn't match for newer
-               // glibcs
-               [!gnu]: "ucontext.h",
     }
 
     // Include linux headers at the end:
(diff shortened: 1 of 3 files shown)


libc-test/build.rs
```diff
@@ -2101,6 +2098,11 @@ fn test_linux(target: &str) {
             // FIXME: musl version using by mips build jobs 1.0.15 is ancient:
             "ifmap" | "ifreq" | "ifconf" if mips32_musl => true,
 
+            // ucontext_t added a new field as of glibc 2.28; our struct definition is
+            // conservative and omits the field, but that means the size doesn't match for newer
+            // glibcs (see https://github.com/rust-lang/libc/issues/1410)
+            "ucontext_t" if gnu => true,
```
gnzlbg: Could you add the following FIXME here? 

> FIXME: remove once Ubuntu 20.04 LTS is released, somewhere in 2020.

src/unix/linux_like/linux/gnu/b64/x86_64/mod.rs
```diff
@@ -285,7 +285,6 @@ s_no_extra_traits! {
         pub uc_mcontext: mcontext_t,
         pub uc_sigmask: ::sigset_t,
         __private: [u8; 512],
-        __ssp: [::c_ulonglong; 4],
```
gnzlbg: Instead of removing the field, can you comment it out, and add a comment about why it is commented out ? For example:

> // FIXME: the shadow stack field requires glibc >= 2.28. 
> // Re-add once we drop compatibility with glibc versions older than 2.28.

libc-test/build.rs
acfoltzer: Done, though I'm a little concerned about the timeline given that 18.04 LTS is supported until 2023. Is there a stated policy for how long this crate supports a given glibc version?

gnzlbg: Not really. We tend to support a "sufficiently" old version, adding features of newer glibc versions on top. The glibc version supported differs per target, it's all a huge mess.

acfoltzer: Hah, fair enough. It would be nice if we could trigger a load-time failure for too-old versions of glibc, as the original bug that triggered my report was pretty difficult to track down. I've seen programs fail to dynamically load before, asking for a particular glibc version, but I don't know how we could introduce that behavior. Either way, something for another PR probably.

Issue #1415, opened 2019-06-28
Support future riscv64-unknown-linux-elf target


fintelia:
Rust currently has support for bare metal RISC-V targets, but cannot compile for RISC-V Linux ones because it lacks standard library support. This is partially blocked because the libc crate does not recognize `target_arch = "riscv64"` and thus does not export symbols like `c_int`, `c_void`, and `uintptr_t` for it.

gnzlbg:
A good place to start would be by adding a docker container for the target, e.g., that install gcc-riscv64-linux-gnu libc6-dev-riscv64-cross qemu-system-riscv64, and compiles libc for the target. That should not be blocked by anything. 

If compilation fails, that container can be added to `allow_failures:` in the .travis.yml.

tgross35:
It seems like this may have been addressed by https://github.com/rust-lang/libc/pull/1684. Please feel free to reopen or refile if that isn't the case.

PR #1416, opened 2019-06-28
Add Dockerfile for riscv64-unknown-linux-gnu

fintelia:
See #1415

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -195,6 +195,8 @@ matrix:
       stage: tier2
     - env: TARGET=powerpc64le-unknown-linux-gnu
       stage: tier2
+    - env: TARGET=riscv64gc-unknown-linux-gnu
+      stage: tier2
     - env: TARGET=s390x-unknown-linux-gnu
       stage: tier2
     - env: TARGET=sparc64-unknown-linux-gnu
--- a/ci/docker/riscv64-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/riscv64-unknown-linux-gnu/Dockerfile
@@ -1,0 +1,11 @@
+FROM ubuntu:19.04
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        gcc libc6-dev qemu-user ca-certificates \
+        gcc-riscv64-linux-gnu libc6-dev-riscv64-cross \
+        qemu-system-riscv64 linux-headers-generic
+
+ENV CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_LINKER=riscv64-linux-gnu-gcc \
+    CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_RUNNER="qemu-riscv64 -L /usr/riscv64-linux-gnu" \
+    CC_riscv64_unknown_linux_gnu=riscv64-linux-gnu-gcc \
+PATH=$PATH:/rust/bin


gnzlbg:
Note that this is not tested on CI, that would require adding a build job to .travis.yml. Is that intended ?

fintelia:
Not intended so much as I just wasn't sure how to add the job to CI. I'll make that change in a bit

gnzlbg:
I've restarted this particular build: https://travis-ci.com/rust-lang/libc/jobs/212378258

fintelia:
That runs into a sort of circular issue: there isn't a riscv64gc-unknown-linux-gnu target yet because the standard library can't be compiled for 64-bit RISC-V. And the standard library can't be build for it because libc doesn't yet support it.

Perhaps there'd be a way to temporarily use the risc64gc-unknown-none-elf target?

mati865:
@fintelia to add new CPU and target at same time you have to fix and test everything locally using by building Rust with modified dependencies (like patched libc), get dependencies fixes merged, get new target merged and finally add tests everywhere once the target works.

fintelia:
@mati865 That makes more sense

gnzlbg:
@fintelia my recommendation is to add it as a libcore target only first to rust-lang/rust . That requires adding a target specification to rust-lang/rust/src/librustc_target, and not much more. Once that is done, libc is no_std, so it can be built against that. Once you have libcore and libc working, getting support for liballoc shouldn't be hard, you can just shim out to libc::malloc. The largest chunk of work is getting libstd support, depending on how similar this is to other unix targets, might be a breeze or not.

mati865:
Good point, I didn't know you can make something as big as `*-unknown-linux-gnu` std-less.

gnzlbg:
> Good point, I didn't know you can make something as big as *-unknown-linux-gnu std-less.

If the target isn't tier2, nothing is built, and then you have to build it with xargo, which you can use to build libcore only. So... it kind of works by omission 😆

fintelia:
Where is the tier listing actually specified in the rustc source? It seems that just adding a target spec file and including it in the [supported_targets!](https://github.com/rust-lang/rust/blob/6ea4036cd20a4b1ab1d56e555108c21735ff6946/src/librustc_target/spec/mod.rs#L330) macro is enough to cause `./x.py build` to try building the standard library for that target

gnzlbg:
@fintelia you need to set `no_std = true` for your target here: https://github.com/rust-lang/rust/blob/03bd2f653f0d0adb69d862fbeec64663157e71e1/src/bootstrap/sanity.rs#L171

gnzlbg:
This needs rebasing. What was previously in the `.travis.yml` needs to be moved to `ci/azure.yml` (check out how the other targets are added there - should be pretty similar).

gnzlbg:
ping

fintelia:
Sorry for the delay on this. I don't think I'm going to have a chance to go through all of the steps to get this together in the near term so if anyone else is interested feel free to take over

gnzlbg:
@msizanoen1 please feel free to fork this branch and update it as you see fit and submit a PR.

msizanoen1:
Actually getting std to compile for riscv64 is pretty simple: https://github.com/rust-lang/rust/compare/master...msizanoen1:riscv-std

JohnTitor:
The dockerfile is added by #1684 and I'm closing this in favor of it, thanks for contributing!

PR #1417, opened 2019-06-28
Fix link in README.md

fintelia:


gnzlbg:
This PR should be filled against the readme in master. The one in gh-pages is overwritten on each commit.

PR #1419, opened 2019-07-03
Update ctest version

gnzlbg:
The latest ctest version enabled the ABI roundtrip test by default, in which we initialize all types in Rust by default to some random bit-pattern, pass them to C, verify, modify, pass back to Rust, and verify. 

This catches issues in the call ABI / calling convention.

This PR will silence those here for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -235,6 +235,7 @@ matrix:
   allow_failures:
       - name: "Semver Linux"
       - name: "Semver MacOSX"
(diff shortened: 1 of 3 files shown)


gnzlbg:
@sunfishcode @alexcrichton wasi is failing to build because the memoffset dependency couldn't be resolved, could you point me to commit hashes that are known to build ? I'll include the fix here.

PR #1421, opened 2019-07-04
Expose signal value of siginfo_t

DoumanAsh:
Currently exposes it for following platforms:

- Linux has it as enum _timer and it is exposed alongside sigval
- On FreeBSD like systems si_value follows after si_addr (according to headers)
- Darwin is similar to FreeBSD
- NetBSD and FreeBSD uses struct that contains pid, uid, and si_value. Exposed via it

P.s. I'd like to take a look at other platforms too, but these are what I know so far.
FreeBSD needs some testing though

cc @gnzlbg

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -144,6 +144,8 @@ s! {
         pub si_uid: ::uid_t,
         pub si_status: ::c_int,
         pub si_addr: *mut ::c_void,
+        //Requires it to be union for tests
+        //pub si_value: ::sigval,
         _pad: [usize; 9],
     }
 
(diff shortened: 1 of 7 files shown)


gnzlbg:
Thank you for the PR. Our CI is currently down because we are migrating to Azure Pipelines, but it looks good, I'll try to merge it this weekend.

DoumanAsh:
Updated with other unix platforms.
I used publicly available repositories with source code as reference.
But I'm not sure how correct it is per se

gnzlbg:
@asomers @semarie this might need tweaking for OpenBSD, but it shouldn't break anything to land this - i'm not sure about NetBSD, but with this in master it should be easier to try this out, and improve this iteratively.

asomers:
You need to double check the size of your pad fields.  They aren't right for 64-bit FreeBSD.

DoumanAsh:
Ok, I'll need to check it.
I was using [signal.h](https://github.com/freebsd/freebsd/blob/master/sys/sys/signal.h#L229) as reference but i didn't exactly check every type inside.
Wonder if it is `sigval` union behaving

gnzlbg:
It appears that the freebsd jobs are still failing.

DoumanAsh:
@gnzlbg Yeah, before my traveling I actually was trying to fix MacOS tests, which are still failing for some reason.
But as for freebsd I'm not exactly sure even where it fails, so I'd need to take a look at how this error ouput gets generated

gnzlbg:
you might just need to rebase on top of master, those error shouldn't be there, and don't appear related to this PR

DoumanAsh:
Ok, done, I'll check it out once CI is finished

DoumanAsh:
@gnzlbg CI passed

PR #1403, opened 2019-06-21
Redox CI

jackpot51:
This does the following:

- Fix compilation on Redox
- Add Redox (x86_64-unknown-redox) to the tier2 CI targets
- Bump version to 0.2.59, so that the broken version 0.2.58 is superseded

PR #1422, opened 2019-07-05
Redox target fix

### LINKED ISSUE #1420: Building for x86_64-unknown-redox results in an empty crate
See for example generated [docs](https://rust-lang.github.io/libc/x86_64-unknown-redox/libc/index.html). I think it was overlooked in  #1332 and until rust-lang/rust#60139 lands we should use `#[cfg(any(unix, target_os="redox"))]` [here](https://github.com/rust-lang/libc/blob/master/src/lib.rs#L116).

cc @jackpot51

newpavlov:
Closes #1420

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -113,7 +113,7 @@ cfg_if! {
 
         mod switch;
         pub use switch::*;
-    } else if #[cfg(unix)] {
+    } else if #[cfg(any(unix, target_os="redox"))] {
         mod fixed_width_ints;
         pub use fixed_width_ints::*;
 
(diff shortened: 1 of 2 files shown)


gnzlbg:
@newpavlov please implement these like #1403 does.

newpavlov:
Looks like some more involved fixes are needed. @jackpot51 can you handle `cfg` fix as part of #1403?

gnzlbg:
@newpavlov I would prefer if that happened in #1403 as well, could you send @jackpot51 a PR with your fix ?

newpavlov:
Done! Closing in favor of the linked PR.

PR #1376, opened 2019-05-28
Setup Azure Pipelines

gnzlbg:
cc @alexcrichton this needs enabling Azure Pipelines for this repo

PR #1428, opened 2019-07-05
Bump to 0.2.59

jbaublitz:
I've added some code to libc for my library and to be able to release the next version of my library, I need the new constants exposed by my PR to be available in crates.io. Would a new release soon be possible?


gnzlbg:
Will be done as part of #1376 .

PR #1429, opened 2019-07-06
Reduce appveyor churn

gnzlbg:
This should reduce the Appveyorn churn for other projects a bit more.

Please let me know if there is anything else that can be done. Maybe the infra team could discuss whether to document some general guidelines for rust-lang and rust-lang-nursery projects, and write them down somewhere. I would be willing to help.

cc @RalfJung @Mark-Simulacrum

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,6 +9,7 @@ environment:
   - TARGET: x86_64-pc-windows-msvc
   - TARGET: i686-pc-windows-msvc
 install:
+  - if NOT defined APPVEYOR_PULL_REQUEST_NUMBER if "%APPVEYOR_REPO_BRANCH%" == "master" appveyor exit
   - if defined MSYS_BITS set PATH=C:\msys64\mingw%MSYS_BITS%\bin;C:\msys64\usr\bin;%PATH%;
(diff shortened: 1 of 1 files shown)


RalfJung:
Thanks a lot!

Another thing you could do is not run the full AppVeyor test suite on every push to every PR. Currently that's about 15min each time. Maybe running one of the 4 jobs is representative enough that the rest can be covered by bors (including try builds if you want pre-landing testing)?

gnzlbg:
> Maybe running one of the 4 jobs is representative enough that the rest can be covered by bors (including try builds if you want pre-landing testing)?

Sadly this isn't the case - the only thing we test here is the system ABI, and that's completely different for each of the targets running on appveyor. 

What we could do is move those jobs to Azure Pipelines, once we get a green light to land #1376 .

RalfJung:
I was thinking the same for Miri (re: moving to Azure). ;)

So, how long until we'll have congestion there? :D Do you know how many runners we collectively have there?

gnzlbg:
> So, how long until we'll have congestion there? :D Do you know how many runners we collectively have there?

No idea, but if rust-lang/rust ends up completely switching to Azure, I suppose the infra team will make an announcement about what their plan is, and what to do with the other repos in the org.

PR #1432, opened 2019-07-09
Add __errno_location on Redox, DragonFlyBSD, Haiku

josephlr:
Currently [`libstd`](https://github.com/rust-lang/rust/blob/909f5a049415a815b23502a5498de9a99bbf276b/src/libstd/sys/unix/os.rs#L29-L49) and the [`getrandom` crate](https://github.com/rust-random/getrandom/pull/54/files#diff-027a56068e2bf3d31dc4273ee6e07034R12) have to use external declarations in order to implement `errno_location` across all UNIX platforms. This change adds bindings for the remaining UNIX platforms, allowing everyone to just use normal `libc` bindings.

This also removes the need for a seperate [`errno-dragonfly`](https://crates.io/crates/errno-dragonfly) crate.

Links to the relevant header files:

- Redox: [`relibc`'s `bits/errno.h`](https://github.com/redox-os/relibc/blob/master/include/bits/errno.h) uses `__errno_location`
- Haiku: [`posix/errno.h`](https://github.com/luciang/haiku/blob/master/headers/posix/errno.h) uses `_errnop`
- DragonFlyBSD: [`sys/errno.h`](https://github.com/DragonFlyBSD/DragonFlyBSD/blob/master/sys/sys/errno.h) uses `__error` just like FreeBSD (which makes sense)
  - Note that the `__error` function is inlined. I don't think this causes a problem.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,9 +26,8 @@ rustc-std-workspace-core = { version = "1.0.0", optional = true }
 default = ["std"]
 std = []
 align = []
-rustc-dep-of-std = ['align', 'rustc-std-workspace-core', 'unstable']
+rustc-dep-of-std = ['align', 'rustc-std-workspace-core']
 extra_traits = []
-unstable = []
 # use_std is deprecated, use `std` instead
 use_std = [ 'std' ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,8 +26,9 @@ rustc-std-workspace-core = { version = "1.0.0", optional = true }
 default = ["std"]
 std = []
 align = []
-rustc-dep-of-std = ['align', 'rustc-std-workspace-core']
+rustc-dep-of-std = ['align', 'rustc-std-workspace-core', 'unstable']
 extra_traits = []
+unstable = []
 # use_std is deprecated, use `std` instead
 use_std = [ 'std' ]
 
--- a/README.md
+++ b/README.md
@@ -35,8 +35,10 @@ libc = "0.2"
 * `extra_traits`: all `struct`s implemented in `libc` are `Copy` and `Clone`.
   This feature derives `Debug`, `Eq`, `Hash`, and `PartialEq`.
 
-* `unstable`: enable currently unstable bindings. Right now, this just allows
-  bindings to `#[thread_local]` statics on certain platforms. Requires nightly.
+* `unstable`: This feature enables `libc` bindings that are only possible with
+  unstable Rust features. Right now, this just for
+  [`extern` `#[thread_local]` statics](https://github.com/rust-lang/rust/issues/29594)
+  on certain platforms. Requires nightly.
 
 * **deprecated**: `use_std` is deprecated, and is equivalent to `std`.
 
--- a/README.md
+++ b/README.md
@@ -35,11 +35,6 @@ libc = "0.2"
 * `extra_traits`: all `struct`s implemented in `libc` are `Copy` and `Clone`.
   This feature derives `Debug`, `Eq`, `Hash`, and `PartialEq`.
 
-* `unstable`: This feature enables `libc` bindings that are only possible with
-  unstable Rust features. Right now, this just for
-  [`extern` `#[thread_local]` statics](https://github.com/rust-lang/rust/issues/29594)
-  on certain platforms. Requires nightly.
-
 * **deprecated**: `use_std` is deprecated, and is equivalent to `std`.
 
 ## Rust version support
--- a/README.md
+++ b/README.md
@@ -35,6 +35,9 @@ libc = "0.2"
 * `extra_traits`: all `struct`s implemented in `libc` are `Copy` and `Clone`.
   This feature derives `Debug`, `Eq`, `Hash`, and `PartialEq`.
 
+* `unstable`: enable currently unstable bindings. Right now, this just allows
+  bindings to `#[thread_local]` statics on certain platforms. Requires nightly.
+
 * **deprecated**: `use_std` is deprecated, and is equivalent to `std`.
 
 ## Rust version support
--- a/build.rs
+++ b/build.rs
@@ -7,7 +7,6 @@ fn main() {
         rustc_minor_version().expect("Failed to get rustc version");
     let rustc_dep_of_std = env::var("CARGO_FEATURE_RUSTC_DEP_OF_STD").is_ok();
     let align_cargo_feature = env::var("CARGO_FEATURE_ALIGN").is_ok();
-    let unstable_cargo_feature = env::var("CARGO_FEATURE_UNSTABLE").is_ok();
 
     if env::var("CARGO_FEATURE_USE_STD").is_ok() {
         println!(
@@ -55,7 +54,7 @@ fn main() {
     }
 
     // #[thread_local] is currently unstable
-    if unstable_cargo_feature || rustc_dep_of_std {
+    if rustc_dep_of_std {
         println!("cargo:rustc-cfg=libc_thread_local");
     }
 }
--- a/build.rs
+++ b/build.rs
@@ -5,18 +5,18 @@ use std::str;
 fn main() {
     let rustc_minor_ver =
         rustc_minor_version().expect("Failed to get rustc version");
-    let rustc_dep_of_std =
-        std::env::var("CARGO_FEATURE_RUSTC_DEP_OF_STD").is_ok();
-    let align_cargo_feature = std::env::var("CARGO_FEATURE_ALIGN").is_ok();
+    let rustc_dep_of_std = env::var("CARGO_FEATURE_RUSTC_DEP_OF_STD").is_ok();
+    let align_cargo_feature = env::var("CARGO_FEATURE_ALIGN").is_ok();
+    let unstable_cargo_feature = env::var("CARGO_FEATURE_UNSTABLE").is_ok();
 
-    if std::env::var("CARGO_FEATURE_USE_STD").is_ok() {
+    if env::var("CARGO_FEATURE_USE_STD").is_ok() {
         println!(
             "cargo:warning=\"libc's use_std cargo feature is deprecated since libc 0.2.55; \
              please consider using the `std` cargo feature instead\""
         );
     }
 
-    if std::env::var("LIBC_CI").is_ok() {
+    if env::var("LIBC_CI").is_ok() {
         if let Some(12) = which_freebsd() {
             println!("cargo:rustc-cfg=freebsd12");
         }
@@ -52,6 +52,11 @@ fn main() {
     // Rust >= 1.33 supports repr(packed(N))
     if rustc_minor_ver >= 33 || rustc_dep_of_std {
         println!("cargo:rustc-cfg=libc_packedN");
+    }
+
+    // #[thread_local] is currently unstable
+    if unstable_cargo_feature || rustc_dep_of_std {
+        println!("cargo:rustc-cfg=libc_thread_local");
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,6 +21,7 @@
     feature = "rustc-dep-of-std",
     feature(cfg_target_vendor, link_cfg, no_core)
 )]
+#![cfg_attr(libc_thread_local, feature(thread_local))]
 // Enable extra lints:
 #![cfg_attr(feature = "extra_traits", deny(missing_debug_implementations))]
 #![deny(missing_copy_implementations, safe_packed_borrows)]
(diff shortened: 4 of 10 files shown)


gnzlbg:
@bors: r+

Thanks!

gnzlbg:
@bors: r-

The paths need fixing.

josephlr:
> The paths need fixing.

Copy paste bites again, should be fixed now.

gnzlbg:
Thanks! @bors: r+

BurntSushi:
Note that both [std](https://github.com/rust-lang/rust/blob/909f5a049415a815b23502a5498de9a99bbf276b/src/libstd/sys/unix/os.rs#L68-L76) and [errno-dragonfly](https://github.com/mneumann/errno-dragonfly-rs/blob/52512bf37a7d3a161ee9054dc7e4b38d887d01c6/src/errno.c#L3) seem to implement this differently than what's in this PR for Dragonfly. Was that intentional?

gnzlbg:
cc @strangelittlemonkey

gnzlbg:
cc @sikmir could you review this for haiku? 
cc @jackpot51 could you review this for redox ?

josephlr:
> Note that both [std](https://github.com/rust-lang/rust/blob/909f5a049415a815b23502a5498de9a99bbf276b/src/libstd/sys/unix/os.rs#L68-L76) and [errno-dragonfly](https://github.com/mneumann/errno-dragonfly-rs/blob/52512bf37a7d3a161ee9054dc7e4b38d887d01c6/src/errno.c#L3) seem to implement this differently than what's in this PR for Dragonfly. Was that intentional?

This was intentional. I looks like implementing it with `#[thread_local]` would be slightly more efficient (as you don't have a function call). However, that's not possible on stable Rust, so using the function instead seems like the best bet.

BurntSushi:
Interesting. The `errno-dragonfly` crate actually uses [custom C code to return `errno`](https://github.com/mneumann/errno-dragonfly-rs/blob/52512bf37a7d3a161ee9054dc7e4b38d887d01c6/src/errno.c#L3) and then [links to that](https://github.com/mneumann/errno-dragonfly-rs/blob/52512bf37a7d3a161ee9054dc7e4b38d887d01c6/src/lib.rs#L5). Is that just an oversight on their part?

N.B. I am speaking from ignorance here. I'm only asking these questions because I'm curious. :-)

gnzlbg:
> I think Rust can handle such a function OK

The `__error` function is static (TU-private) and inline, so it probably isn't part of the generated binary, and we can't link against that. That is, we would need to implement it in this crate.

DragonflyBSD defines `errno` [here](https://github.com/DragonFlyBSD/DragonFlyBSD/blob/master/sys/sys/errno.h#L45) as :

```c
extern __thread int	errno;
static __inline int* __error() { return &errno; }
```

This:

```c
extern __thread int	errno;
int* foo() { return &errno; }
```

emits:

```llvm
define dso_local i32* @foo() #0 {
  ret i32* @errno
}
```

Notice the `dso_local` in there.

---

_I think_ that in Rust we would need: `extern "C" { #[thread_local] static errno: i32; }` (using `feature(thread_local)`), but I don't know whether this currently works when used on `extern static`s. Right now we don't emit `dso_local`, and I don't know how important that is. cc @eddyb @rkruppe

gnzlbg:
Also cc @joshtriplett: there is no stable Rust way to link against `extern #[thread_local] static`s, that should probably be added to the list of missing features for C FFI.

gnzlbg:
@BurntSushi 

> Is that just an oversight on their part?

I don't think so. They can't crate a pointer to the `static __thread int errno` from Rust, so they have to do that from C. See above.

eddyb:
`dso_local` is a function attribute and likely the `static` on `foo`.
You need to compare the declarations of the `@errno` global to check if the `#[thread_local]` works (it should, since we lower it pretty straightforwardly).

gnzlbg:
@eddyb so we lower the `extern "C" { #[thread_local] static errno: i32; }` to 

```llvm
@errno = external thread_local global i32
```

clang lowers this to:

```llvm
@errno = external thread_local global i32, align 4
```

So I think `#[thread_local]` should work.

gnzlbg:
@josephlr we do have a cargo feature `feature = "unstable"` that you could use to only provide the `#[thread_local]`s for nightly users that enable that feature via a `#[cfg()]`.

I don't think there is a pure Rust solution available for stable Rust users. Linking to a C snippet that returns a pointer to the thread local like the `errno-dragonfly` crate does feels like the simplest solution to this problem.

eddyb:
Could that platform's `errno` be gated by `#[cfg(feature = "unstable")]` to land this, and have an issue for stable support?

gnzlbg:
@eddyb sure, and libstd enables the "unstable" feature, so it can make use of this to simplify the implementation. The `getrandom` crate would need to wait a bit.

sikmir:
LGTM

josephlr:
> @josephlr we do have a cargo feature `feature = "unstable"` that you could use to only provide the `#[thread_local]`s for nightly users that enable that feature via a `#[cfg()]`.

@gnzlbg that works, I’ll make the change to add `__error` and `errno` behind this guard. For `getrandom`, we might just be OK requiring nightly for DragonFlyBSD.

gnzlbg:
Also, I don't know how good will `ctest` be with verifying the ABI of thread-locals. If there are any issues, feel free to just use `cfg.skip_static` in `libc-test/build.rs` to skip the tests, and fill an issue in ctest upstream. We can solve those later.

josephlr:
@gnzlbg this crate doesn't seem to have an `unstable` feature in `Cargo.toml`. Would you want me to add it, or am I missing something?

gnzlbg:
Yes, please add it. I think we might have removed it when we started enabling features depending on the Rust-versions.

josephlr:
> Yes, please add it. I think we might have removed it when we started enabling features depending on the Rust-versions.

Would you want it to be called `nightly` or `unstable`? Using `nightly` seems slightly more common among crates.

gnzlbg:
So, call the cargo feature `unstable`, and in the `libc/build.rs`, if that feature or `rustc_dep_of_std` are enabled, pass a `--cfg libc_thread_local` when building the library. Inside `libc/src`, use `#[cfg_attr(libc_thread_local, feature(thread_local))]` to enable the `feature(thread_local)` and `#[cfg(libc_thread_local)]` to gate its uses.

gnzlbg:
The reason I prefer not to use nightly, is that a lot of people compile libc with nightly, but don't want to use unstable features when they do so. So whether a nightly toolchain is used is orthogonal to whether unstable features are used.

strangelittlemonkey:
DragonFly BSD has no ability at present to build nightly so if you move this to requiring nightly then there’s no point in making any rules in the setup that account for DragonFly.

josephlr:
> So, call the cargo feature `unstable`, and in the `libc/build.rs`, if that feature or `rustc_dep_of_std` are enabled, pass a `--cfg libc_thread_local` when building the library. Inside `libc/src`, use `#[cfg_attr(libc_thread_local, feature(thread_local))]` to enable the `feature(thread_local)` and `#[cfg(libc_thread_local)]` to gate its uses.

@gnzlbg this is done. I had to use `cfg_if` and a sub-module to add the errno stuff, otherwise the style checker yelled at me. Also, there seem to be a bunch of CI failures related to `serde_derive`, any idea why those are happening?

> DragonFly BSD has no ability at present to build nightly so if you move this to requiring nightly then there’s no point in making any rules in the setup that account for DragonFly.

@strangelittlemonkey I'm not sure what you mean here. Nightly is required to build `libstd` and we have Dragonfly related code in `libstd`.

josephlr:
It turns out the CI is failing due to a transitive `serde_derive` dependency, which is breaking because of https://github.com/rust-lang/rust/issues/62562

All the errors (as far as I can tell) are related to this bug, so the breakage is not caused by this PR.

strangelittlemonkey:
You’re fine to write DragonFly code for something needing nightly but it’s dead code because we have no nightly build and I’m not sure we’ll ever have nightly support or a nightly build even. If you’re requiring nightly you should take DragonFly off the support list for the crate or any other like that because it would just be misleading.

eddyb:
@strangelittlemonkey "nightly" is misleading - what's important here is "unstable", which is allowed for any dependency of `std`/`rustc`.

Even if users of `libc` from `crates.io` might not use the unstable feature on DragonFlyBSD, to build even `stable`/`beta`, `std` needs `errno` from `libc`, for which it can enable some "unstable" flag to provide `errno`.

gnzlbg:
@strangelittlemonkey or in other words, when the compiler is compiling itself, which is something that has to be done to build a compiler for DragonflyBSD, it can use unstable features that are generally only available on nightly.

strangelittlemonkey:
I’m just saying requiring nightly to build it at all would make DragonFly a moot point, but allowing extra features if building with nightly would be different and fine, but still dead code if it applied to DragonFly specifically.

eddyb:
> I’m just saying requiring nightly to build it at all would make DragonFly a moot point, but allowing extra features if building with nightly would be different and fine

It's the latter, we're talking about an opt-in way to enable `errno` on DragonFly that requires unstable. This is *either* for nightly *or* building Rust itself.

Like I said before, that opt-in would be enabled by `std`, which depends on this crate for its FFI to implement `std::io`, `std::fs` and other such functionality.

So when you build any future release of Rust (including stable ones) for DragonFlyBSD, the `errno` added here would be used, *even if* a use of that build of Rust wouldn't have that `errno` available.

So, yes, it's dead code for *normal users* of this crate, but would still be needed and used for the standard library (which does its OS FFI through this very same crate).

josephlr:
@gnzlbg https://github.com/rust-lang/rust/issues/62562 has been fixed, so the CI is passing again

@strangelittlemonkey Nightly build issues aside, the code added in this PR is required to build `libstd` _for_ DragonFlyBSD (not necessarily building _on_ DragonFlyBSD). Will this code work for DragonFlyBSD? and if not, what should we do instead?

I updated the docs to point out _why_ we need `unstable` at all

gnzlbg:
Just to triple check, is there any crate that wants to use this from crates.io ? 

If not, then we don't need the `unstable` feature, and can just feature gate this using the "dep-of-std" feature.

josephlr:
> Just to triple check, is there any crate that wants to use this from crates.io ?

Initially the thought was to also use this for `getrandom`, but due to https://github.com/rust-lang/cargo/issues/4866 we can't reasonably turn on the `unstable` feature for only one target, so I don't know of any crates that currently want it. 

> If not, then we don't need the `unstable` feature, and can just feature gate this using the "dep-of-std" feature.

Sounds good, I'll make the change unless anyone else want's to use this.

gnzlbg:
@josephlr you can chime in in the `thread_local` tracking issue in rust-lang/rust, and mention that getrandom wants to use it. I don't know if anybody has needed to use `extern` thread-locals before, but that's something required for being able to FFI with C..

BurntSushi:
> Just to triple check, is there any crate that wants to use this from crates.io ?

Yes. `walkdir` will eventually want this. I'm currently using the same hack as the `errno-dragonfly` crate in my branch: https://github.com/BurntSushi/walkdir/blob/ag/sys/src/os/unix/errno.rs

josephlr:
> > Just to triple check, is there any crate that wants to use this from crates.io ?
> 
> Yes. `walkdir` will eventually want this. I'm currently using the same hack as the `errno-dragonfly` crate in my branch: https://github.com/BurntSushi/walkdir/blob/ag/sys/src/os/unix/errno.rs (Which I'll have to continue using if `errno` on DragonFly is only available with nightly Rust.)

OK, as `walkdir` and `getrandom` can’t use this until it’s on stable, we don’t need the `unstable` feature. I’ll mention this on the main Rust tracking issue.

gnzlbg:
Looking at the tracking issue, and the `thread_local!` discussions, it appears that 1) `thread_local!` solves the problem for most people, and 2) the main uncertainty is that `#[thread_local]` isn't portable.

Maybe we could reduce the scope of an initial version of `#[thread_local]` to `extern static` _declarations_. If the target does not support `#[thread_local]`, well then the `extern static` _declaration_ is incorrect since there cannot be a _definition_ anywhere, and using it would already be UB. We could add on top of this a "best effort" compilation error, e.g., if the compiler knows that the target doesn't support it.

josephlr:
@gnzlbg I removed the `unstable` feature, and commented on the appropriate [`thread_local` tracking issue](https://github.com/rust-lang/rust/issues/29594).

gnzlbg:
Thanks!

@bors: r+

gnzlbg:
libc 0.2.60 has been published with this - I'll update libc on rust-lang/rust

PR #1433, opened 2019-07-09
Link to Windows debug CRTs when crt-debug target_feature is present.

jdm:
Relies on https://github.com/rust-lang/rust/pull/62534 in order to fix https://github.com/rust-lang/rust/issues/39016.

### DIFF
--- a/src/windows/mod.rs
+++ b/src/windows/mod.rs
@@ -193,8 +193,10 @@ pub const SIG_ERR: ::c_int = -1;
 
 // inline comment below appeases style checker
 #[cfg(all(target_env = "msvc", feature = "rustc-dep-of-std"))] // " if "
-#[link(name = "msvcrt", cfg(not(target_feature = "crt-static")))]
-#[link(name = "libcmt", cfg(target_feature = "crt-static"))]
+#[link(name = "msvcrt", cfg(all(not(target_feature = "crt-static"), not(target_feature = "crt-debug"))))]
+#[link(name = "msvcrtd", cfg(all(not(target_feature = "crt-static"), target_feature = "crt-debug")))]
+#[link(name = "libcmt", cfg(all(target_feature = "crt-static", not(target_feature = "crt-debug"))))]
+#[link(name = "libcmtd", cfg(all(target_feature = "crt-static", target_feature = "crt-debug")))]
 extern {}
 
 #[cfg_attr(feature = "extra_traits", derive(Debug))]


jdm:
This makes it possible to link a Rust binding crate against a C library that was compiled with /MTd or /MDd.

gnzlbg:
What happens if the standard library is built to link to one, but this crate from crates.io is set to link to the other ?

What happens if a `#[no_std]` binary only uses this crate without using the standard library.

Also, this needs tests that show that this works.

jdm:
@gnzlbg What happens in those cases when crt-static is used?

jdm:
It's true; this change may not be enough because of the issue of building the standard library. I tried building https://github.com/servo/mozjs with a Cargo.toml patch for libc that replaced msvcrt with msvcrtd, and I still got the same error that I'm used to seeing:
```
 = note: lld-link: error: /failifmismatch: mismatch detected: 0 and 2 for key _ITERATOR_DEBUG_LEVEL
```

gnzlbg:
@jdm note that libc is built as part of `libstd`, and linked there, applying certain link flags, and that when using it from crates.io, how things actually should get linked will depend, on whether the binary is `#[no_std]` or not. If the binary links libstd, trying to override link flags here will probably cause incompatibilities. I'd recommend you to open an issue in rust-lang/rust to discuss whether this is the right way to enable this feature - or does such an issue already exist ?

jdm:
I mentioned it in the original PR description: https://github.com/rust-lang/rust/issues/39016

ollie27:
@gnzlbg when using this crate from crates.io nothing get linked: https://github.com/rust-lang/libc/blob/6307a0bfce8b3fd980b9cb514e17d524de0a666d/src/windows/mod.rs#L195

gnzlbg:
@ollie27 so if one builds a `#[no_std]` binary, and adds the libc crate as a dependency, no C standard library will be linked ? That sounds like a bug.

ollie27:
> so if one builds a `#[no_std]` binary, and adds the libc crate as a dependency, no C standard library will be linked ?

Correct. I don't know if it's a bug or not but it looks like the same is true of musl: https://github.com/rust-lang/libc/blob/6307a0bfce8b3fd980b9cb514e17d524de0a666d/src/unix/mod.rs#L299-L305

I guess this is just how `crt-static` works at the moment.

alexcrichton:
It looks like this was realized in the last few comments, but replying to [this cc](https://github.com/rust-lang/rust/issues/39016#issuecomment-509723742) this PR won't have the desired effect until something like https://github.com/rust-lang/rust/pull/62534 is also added.

Bromeon:
Has there been progress towards linking to MSVCRT**D** (debug version), as described in https://github.com/rust-lang/rust/issues/39016?

If not, are there meanwhile any known workarounds?

jdm:
No.

jdm:
I don't have access to a Windows development environment any more, so I won't be pushing forward on this work.

PR #1435, opened 2019-07-09
Add missing syscalls for aarch64-unknown-linux-musl

dianpopa:
The list of syscalls from [rust-lang-aarch64-musl-syscalls](../blob/master/src/unix/linux_like/linux/musl/b64/aarch64.rs) misses some of the syscalls that are defined in [musl-libc-aarch64-syscalls](https://git.musl-libc.org/cgit/musl/tree/arch/aarch64/bits/syscall.h.in).
We need this so that we get rid of hardcoding them inside [our project](https://github.com/firecracker-microvm/firecracker/blob/master/vmm/src/default_syscalls/filters.rs#L16).

Signed-off-by: Diana Popa <dpopa@amazon.com>


gnzlbg:
@bors: r+ 

Looks good, if bors agrees this is in, thanks!

Issue #1424, opened 2019-07-05
Re-enable WASI CI


gnzlbg:
cc @sunfishcode

tgross35:
Done in https://github.com/rust-lang/libc/pull/3869.

PR #1437, opened 2019-07-11
Update wasmtime to the latest master.

sunfishcode:
The previous wasmtime revision is broken because one of its dependencies, `memoffset` 0.3, was yanked, which appears to make it unavailable. This was fixed by downgrading to `memoffset` 0.2.

### DIFF
--- a/ci/docker/wasm32-wasi/Dockerfile
+++ b/ci/docker/wasm32-wasi/Dockerfile
@@ -60,7 +60,7 @@ ENV PATH=/root/.cargo/bin:$PATH
 RUN apt-get install -y --no-install-recommends python
 RUN git clone --recursive https://github.com/CraneStation/wasmtime wasmtime && \
   cd wasmtime && \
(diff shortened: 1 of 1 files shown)


sunfishcode:
I should mention; this PR fixes https://github.com/rust-lang/libc/issues/1424.

sunfishcode:
The Travis failure here looks like the same failure happening [on master](https://travis-ci.com/rust-lang/libc/jobs/218149470).

gnzlbg:
Sorry that it took so long, it was unclear what was causing the master failure and it turned out to be the LLVM 9 upgrade... fun times...

@bors: r+

gnzlbg:
@sunfishcode the wasi build job is still failing on master (notice that it is allowed to fail). Would you mind taking a look?

sunfishcode:
Looks like a compiler problem compiling the new linux_termios test for wasm. I'll investigate further when I get a chance.

PR #1438, opened 2019-07-12
Increase Redox & Relibc support

AdminXVII:
- Add a lot of constants from relibc
- Fix the timezone not found error found previously on Redox
- Wrap WIFEXITED _et al._ in an unsafe block to match the rest of the API
- Add support for the extra_traits feature and Redox
- Fmt

cc @jackpot51

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -609,6 +609,7 @@ extern {
                link_name = "fdopendir$INODE64")]
     #[cfg_attr(all(target_os = "macos", target_arch = "x86"),
                link_name = "fdopendir$INODE64$UNIX2003")]
+    #[cfg(not(target_os = "redox"))]
     pub fn fdopendir(fd: ::c_int) -> *mut ::DIR;
 
     #[cfg_attr(target_os = "macos", link_name = "readdir$INODE64")]
@@ -630,6 +631,7 @@ extern {
     /// https://illumos.org/man/3lib/libc
     /// https://docs.oracle.com/cd/E36784_01/html/E36873/libc-3lib.html
     /// https://www.unix.com/man-page/opensolaris/3LIB/libc/
+    #[cfg(not(target_os = "redox"))]
     pub fn readdir_r(dirp: *mut ::DIR, entry: *mut ::dirent,
                      result: *mut *mut ::dirent) -> ::c_int;
     #[cfg_attr(all(target_os = "macos", target_arch = "x86"),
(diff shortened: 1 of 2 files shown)


src/unix/mod.rs
```diff
@@ -1131,6 +1101,46 @@ extern {
         stream: *mut FILE) -> ssize_t;
 }
 
+cfg_if! {
+    if #[cfg(not(target_os = "redox"))] {
+        extern {
+            pub fn getsid(pid: pid_t) -> pid_t;
+            pub fn truncate(path: *const c_char, length: off_t) -> ::c_int;
```
jackpot51: This should definitely be implemented for Redox in relibc, if it is not already

src/unix/mod.rs
```diff
@@ -1131,6 +1101,46 @@ extern {
...
+            pub fn pause() -> ::c_int;
```
jackpot51: This should definitely be implemented for Redox in relibc, if it is not already

src/unix/mod.rs
```diff
@@ -1131,6 +1101,46 @@ extern {
...
+            pub fn mkdirat(dirfd: ::c_int, pathname: *const ::c_char,
+                           mode: ::mode_t) -> ::c_int;
+            pub fn openat(dirfd: ::c_int, pathname: *const ::c_char,
+                          flags: ::c_int, ...) -> ::c_int;
+
+            #[cfg_attr(all(target_os = "macos", target_arch = "x86_64"),
+                       link_name = "fdopendir$INODE64")]
+            #[cfg_attr(all(target_os = "macos", target_arch = "x86"),
+                       link_name = "fdopendir$INODE64$UNIX2003")]
+            pub fn fdopendir(fd: ::c_int) -> *mut ::DIR;
```
jackpot51: This should definitely be implemented for Redox in relibc, if it is not already

AdminXVII:
Note to the reviewer: The changes requested by jackpot51 are to be done after merging this PR.

AdminXVII:
Ok two builds failed with the same error, so my changes are likely responsible for the break. The error log prints
```
bad pipe function pointer: rust: 2805304260 (0xa7358bc4) != c 2805910536 (0xa73ecc08)
bad pause function pointer: rust: 2805324985 (0xa735dcb9) != c 2805304260 (0xa7358bc4)
```
for target i686-apple-darwin. Do you have any pointers for places to look for?

PR #1440, opened 2019-07-21
Add support for FreeBSD CURRENT (aka freebsd13)

pizzamig:
Using the last FreeBSD-CURRENT (development snapshot) the libc build, but tests fail.
This patch detects and supports FreeBSD CURRENT as freebsd13, and reworks the conditional compilation to use the `freebsd11` attribute instead of `not(freebsd12)`
For now, freebsd13 is reusing all freebsd12 definitions, except for `ELAST`
While here, add a new `errno`introduced in freebsd13


gnzlbg:
So this LGTM, @asomers what do you think?

pizzamig:
This patch has the only goal to allow me to have a libc up and running on FreeBSD 13 and I thought it was a good way to start contributing to this crate.
My following goal is to make `cpuset_{get|set}affinity()` functions usable from Rust.
However I'm willing to help to improve FreeBSD support in general.

`ELAST` was already there and I see it changed because of `EINTEGRITY` (already added with this patch). In general I agree to remove it, but I'm not sure that it can be easily done without breaking the crate retro-compatibility (removing is considered a breaking change)

I can work to add the other FreeBSD13 symbols in a later patch.

For the cross-compile problem, I've not a clear idea of how that could be solved. Maybe adding a feature to the crate, managed by `build.rs`, to generate appropriate rustc-cfg, with default to freebsd11 ?

gnzlbg:
@bors: r+

---

I think that, as long as freebsd11 and 12 are not broken, if this works for you, it is ok to merge. We should try setting up at least a build job for freebsd 13 to check that things don't stop building, but that can be done later (should be the next step though). At some point once there are freebsd 13 images available on cirrus-ci, we can start running the tests there.

gnzlbg:
This needs to be rebased

pizzamig:
rebased

gnzlbg:
It appears that some ctest dependency is now failing to build on cirrus ci for some reason.

pizzamig:
I ran some tests and it seems that the extprim crate is broken using nightly

gnzlbg:
Which dependency is pulling it? Maybe we can just remove that dependency somehow (this would be a ctest issue), it is weird that it only triggers on freebsd12 cirrus build job.

pizzamig:
extprim is needed by syntex_syntax2 that's needed by ctest.
Theoretically, extprim is not needed anymore, because rust 1.16 provides u128 and i128

The error is of this form:
```
error: `X...` range patterns are not supported
   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/extprim-1.7.0/src/traits.rs:94:21
    |
94  |                     1 ... $e => u128::new(mantissa) << exp,
    |                     ^^^^^ help: try using the maximum value for the type: `1...MAX`
...
117 | impl_to_extra_primitive_for_float!(f32, 23, 104, 103);
    | ------------------------------------------------------ in this macro invocation

error: expected one of `=>`, `if`, or `|`, found `104`
   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/extprim-1.7.0/src/traits.rs:94:27
    |
94  |                     1 ... $e => u128::new(mantissa) << exp,
    |                           ^^ expected one of `=>`, `if`, or `|` here
...
117 | impl_to_extra_primitive_for_float!(f32, 23, 104, 103);
    | ------------------------------------------------------ in this macro invocation
```
and it seems something related to macro, the `$e` is not recognized as upper limit.

gnzlbg:
So syntex_syntax2 requires extprim 1.0, https://github.com/gnzlbg/syntex/blob/master/syntex_syntax/Cargo.toml#L15

gnzlbg:
I suppose we could patch that to a `extprim = "=1.0"` to fix it, or maybe `extprim` 1.6 did not have this issue ?

pizzamig:
1.0 and 1.6 fails to build, they all have the same macro...

pizzamig:
Is there something I can do to fix it?

mati865:
Isn't git history a bit messed up here?

PR #1443, opened 2019-07-27
Add support for hexagon-unknown-linux-musl

androm3da:


gnzlbg:
@bors: r+ 

LGTM. Do you know if qemu supports `hexagon` ? If so, it might be worth it to try to set up a docker container that runs libc tests for this target under qemu-user like we do for the other targets.

androm3da:
> @bors: r+
> 
> LGTM. Do you know if qemu supports `hexagon` ? If so, it might be worth it to try to set up a docker container that runs libc tests for this target under qemu-user like we do for the other targets.

No upstream support yet but we'll be working on that soon.  Once it's present there we can add it.

Do you know how the qemu package(s) are installed for the CI?  Does it just use the distro release of qemu?  I ask because it would probably take quite a while after upstreaming before we saw it in an Ubuntu/Debian release.

gnzlbg:
> Do you know how the qemu package(s) are installed for the CI? Does it just use the distro release of qemu? I ask because it would probably take quite a while after upstreaming before we saw it in an Ubuntu/Debian release.

I think we could just install the latest released qemu from a third-party repository, at least, until the required qemu version starts being shipped by the distro.

PR #1447, opened 2019-07-30
Add ioctl() constants for BPF of Apple

equal-l2:


equal-l2:
I'll fix everything and force-push soon!

PR #1449, opened 2019-07-31
Add musl support for MIPS64 & bump to 0.2.63

xen0n:
Tested with patched stage2; both static and dynamic binaries confirmed working.

Initial CI support in the form of no-core targets are added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "libc"
-version = "0.2.62"
+version = "0.2.63"
 authors = ["The Rust Project Developers"]
 license = "MIT OR Apache-2.0"
 readme = "README.md"
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -204,6 +204,8 @@ i686-unknown-netbsd \
 i686-unknown-openbsd \
 mips-unknown-linux-uclibc \
 mipsel-unknown-linux-uclibc \
+mips64-unknown-linux-muslabi64 \
+mips64el-unknown-linux-muslabi64 \
 nvptx64-nvidia-cuda \
 powerpc-unknown-linux-gnuspe \
 powerpc-unknown-netbsd \
--- a/ci/docker/mips64-unknown-linux-muslabi64/Dockerfile
+++ b/ci/docker/mips64-unknown-linux-muslabi64/Dockerfile
@@ -1,0 +1,15 @@
+FROM ubuntu:19.04
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  gcc make libc6-dev git curl ca-certificates \
+  gcc-mips64-linux-gnuabi64 qemu-user
+
+COPY install-musl.sh /
+RUN sh /install-musl.sh mips64
+
+# FIXME: shouldn't need the `-lgcc` here, shouldn't that be in libstd?
+ENV PATH=$PATH:/musl-mips64/bin:/rust/bin \
+    CC_mips64_unknown_linux_muslabi64=musl-gcc \
+    RUSTFLAGS='-Clink-args=-lgcc' \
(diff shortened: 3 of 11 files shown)


xen0n:
failure is the same as #1450

xen0n:
The i686 mingw failure seems unrelated.

ci/azure.yml
```diff
@@ -58,8 +58,12 @@ jobs:
           TARGET: mips-unknown-linux-musl
         mips64-unknown-linux-gnuabi64:
           TARGET: mips64-unknown-linux-gnuabi64
+        #mips64-unknown-linux-musl:
+        #  TARGET: mips64-unknown-linux-musl
```
gnzlbg: These tests need to be re-enabled

xen0n: Aren't `rustc` support necessary before the tests can even pass, forming a bootstrap dependency cycle?

xen0n: Well, seems like `libc` changes should go after the `librustc_target` enablement. I'll ready the `rustc` bits in a few hours.

gnzlbg: > Aren't rustc support necessary before the tests can even pass, forming a bootstrap dependency cycle?

Yes. First add the target to librustc_target. Then add support here, and add the targets to ci/build.sh (there is a nightly_nocore list there). Then update libc upstream in rust-lang/rust, and then you can start working on liballoc and libstd support while using libc. 

At some point in between you might want to enable these, but these might require libstd support before they are enabled, you might need to tune which tests are run for this target.

xen0n:
For the record, I for some unknown reason included CI changes in *both* of the PRs, so neither would build... 🤦‍♂

gnzlbg:
You can remove the changes to azure.yml here. If you don't want to wait for the target to be available, you can add a `.json` file here if you want, but since there is a PR upstream already, I'll just wait on this till that lands.

gnzlbg:
This needs to be updated (e.g. remove the comments in the azure.yml file, or enable those jobs), also, these targets need to be added to the ci/build.sh file (probably to the NIGHTLY_NOCORE list until we start shipping core and/or libstd for these targets).

xen0n:
Updated. The `librustc_target` enablement should land today; this should go in afterwards. (And with a version bump.)

xen0n:
Ah. The `rustc` change didn't make it in, in fact it's still waiting.

Once rust-lang/rust#63165 lands this can be retried; merging of this should be avoided before that.

xen0n:
rust-lang/rust#63165 just landed! Can we retry this?

gnzlbg:
@xen0n while the PR to rust-lang/rust landed, new nightly toolchains are only created approx. once per day, so usually you have to wait until tomorrow. Don't despair ;)

xen0n:
Turns out I misunderstood the new CI pipeline...

https://github.com/rust-lang/libc/blob/4bd0c0002631b09ca0f15d93c3101701dea83826/ci/azure-install-rust.yml#L24-L30

It seems any target in the build matrix needs to have `std` already built. Hence I'm taking the new targets out of the build matrix just like other `NO_CORE` targets.

gnzlbg:
> It seems any target in the build matrix needs to have std already built.

Yes, you only can do that once you have enabled `std` support in rust-lang/rust, which needs to happen after the initial changes here (where we only check that the library builds correctly).

xen0n:
We're blocked by #1473. Should we just cherry-pick the build-fixing commit to at least unblock things?

xen0n:
Could this version bump be pushed to crates.io, so I can continue work on upstreaming the corresponding `std` support?

gnzlbg:
Sure

gnzlbg:
Released, sorry for the delay, we had too many time outs the last couple of days.

PR #1450, opened 2019-07-31
Fix redox

mati865:
Revival of https://github.com/rust-lang/libc/pull/1422

All the missing impls were added in https://github.com/rust-lang/libc/pull/1438

Should unblock https://github.com/rust-lang/rust/pull/61393

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -114,7 +114,7 @@ cfg_if! {
 
         mod switch;
         pub use switch::*;
-    } else if #[cfg(unix)] {
+    } else if #[cfg(any(unix, target_os="redox"))] {
         mod fixed_width_ints;
         pub use fixed_width_ints::*;
 


mati865:
Successful build: https://travis-ci.com/mati865/libc/builds/121390368

newpavlov:
You will need to add [`Clone` impls](https://github.com/rust-lang/libc/pull/1422/files and define `timezone`) and probably remove derive of traits (or implement them manually) which cause failures [here](https://travis-ci.com/rust-lang/libc/jobs/213715301).

gnzlbg:
The derive of traits (except for Copy and Clone) aren't really necessary for anything, so you can just leave them all out.

mati865:
It was fixed in https://github.com/rust-lang/libc/pull/1438

The build fails because of https://github.com/rust-lang/rust/issues/63115 which will be fixed by https://github.com/rust-lang/rust/pull/63122

newpavlov:
@gnzlbg
Can you re-run CI to see if the issue got fixed?

mati865:
CI is being migrated to Azure and will not succeed without https://github.com/rust-lang/libc/pull/1376

jackpot51:
This is no longer required

mati865:
It won't be necessary for new nightly builds but libc is still empty when building for Redox from old nightly, beta or stable although I don't even know if they work after all those changes.

jackpot51:
I don't really see a use case for building for Redox prior to the 2019-08-08 nightly

jackpot51:
Anything doing that will need to be updated anyways

PR #1452, opened 2019-08-02
WIP: Support all possible netlink constants in gnu/mod.rs on musl

jbaublitz:
This PR is meant to discover which netlink constants need to be gated in `gnu/mod.rs` and which ones are okay in `mod.rs` for Linux. It seems many of the early constants added were defaulted to GNU only which is not necessarily needed. This will probably fail CI a few times before I am able to remove WIP.


jbaublitz:
I'm going to close this out and make a new PR. I set up my own musl-based CI that seems to work with this and the new pipelines that you mentioned in #1460 don't seem to be showing up.

PR #1458, opened 2019-08-06
Bump version to 0.2.61

jackpot51:
Releasing a version now will allow Redox compilation to work again, as it will include #1438


gnzlbg:
FYI 0.2.61 has been published

PR #1461, opened 2019-08-09
Remove WASI Core API

### LINKED ISSUE #1434: Status of WASI Core API
Initially I've asked this questions relevant to `libc` [here](https://www.reddit.com/r/rust/comments/cap8sy/rust_136_stabilized_the_wasm32wasi_target/). I am not sure if i should ask them in WebAssembly/WASI repository, so I'll duplicate them here:
- WASI is defined in terms of [Core API](https://github.com/WebAssembly/WASI/blob/master/design/WASI-core.md), but I don't quite understand the part about "wasi_unstable". Does it mean that it's not recommended to link to functions directly like it's [done](https://github.com/rust-lang/libc/blob/master/src/wasi.rs#L1054) in `libc`?
- Why Core API is defined in `libc` instead of a separate crate, like `cloudabi` or Fuchsia crates? It's a bit strange to use `libc` for a target which, well, is not defined in terms of libc.

cc @sunfishcode @alexcrichton

newpavlov:
Closes #1434

This change does not break the backwards compatibility promise since WASI Core API is unstable right now. If applications or libraries want to use Core API directly they should use [`wasi`](https://crates.io/crates/wasi) instead of `libc`.

Blocked by: rust-lang/rust#63676

cc @sunfishcode


gnzlbg:
@sunfishcode this needs your review
cc @alexcrichton maybe?

alexcrichton:
I do not agree with this change and would not remove this API. Instability inherent with a platform will of course cause instability in the libraries for the platform, and no one's expecting the `libc` crate to make a breaking change when the wasi api changes.

newpavlov:
@alexcrichton 
Do you believe that `libc` is a correct place for exposing (unstable) WASI Core API, even considering the existence of the dedicated `wasi` crate? Or are you hesitant to introduce a breaking change with such removal?

If it's the latter, I guess we could mark those functions `deprecated` for the time being.

alexcrichton:
Yes, I believe `libc` is the right place for these APIs. They're in the header files in the C implementation, so `libc` seems like the place to put them. 

I would also not go as far as marking them deprecated. We'll just follow what wasi does in C, and if that breaks then the Rust target will break. Again no one is expecting the `libc` crate to have a breaking version bump for the wasi target. Everything about wasi is unstable and breaking and changing.

sunfishcode:
This path is growing on me. The reason these things on the C side are declared in WASI libc is mainly that we don't have a package manager for C so it's convenient to have everything live in one place. But the WASI interfaces are logically a layer below the rest of libc.

@newpavlov This patch should probably also remove the various `__wasi_*` types. And, what would you think about removing all the various `__WASI_*` constants too, and importing them from the wasi crate, as needed, instead?

newpavlov:
@sunfishcode 
Ah, yes, I have indeed overlooked `__wasi_*` types. I am not sure about importing constants from `wasi` though. Can't we simply duplicate them? For example for Fuchsia `libc` duplicates constants instead of importing them from Fuchsia crates.

newpavlov:
I have removed `__wasi_*` types and `__WASI_*` constants, but left `__wasilibc_*` functions untouched.

sunfishcode:
Overall this looks good.

I'm wondering if there's anything else we can do about the `__WASI_*` constants though. WASI is less stable than Fuchsia at this point, and some of these constants could very well change at the WASI level. Besides the awkwardness for maintainers to keep everything in sync, it's awkward for end users if eg `__WASI_O_CREAT` ever changes, because then their `open(..., O_CREAT)` code would break. It doesn't say which WASI version it's using. But, I'm not sure if there's anything we *can* do about that.

alexcrichton:
I'm fine deferring to @sunfishcode's judgement here, but to avoid another debacle of updating `libc` in rust-lang/rust I'd request that the reliance of libstd on these symbols is removed before this PR is merged.

newpavlov:
@sunfishcode 
Can you add removed types and constants to `wasi`? After that I can help with migrating rust-lang/rust to `wasi` if you want.

sunfishcode:
@newpavlov The current wasi package has all the types and constants in it.

newpavlov:
@sunfishcode What should I do with `__wasilibc_find_relpath`? It uses `__wasi_rights_t`, should I duplicate it in `libc`?

sunfishcode:
Yeah, I think it's fine to duplicate the definition of `__wasi_rights_t` here for now. That said, [it may change](https://github.com/WebAssembly/WASI/issues/63), but that's just another instance of a [more general problem](https://github.com/rust-lang/libc/pull/1461#issuecomment-521052628).

newpavlov:
rust-lang/rust#63676 has landed, so this PR is not blocked anymore.

PR #1464, opened 2019-08-12
Update README Azure badge to rust-lang2 org

gnzlbg:


gnzlbg:
@pietroalbini  azure doesn't run when checking PRs, even though the `yml` has a `pr = ["master"]`. Is that incorrect ?

pietroalbini:
That seems correct... I don't have much time to investigate it though.

PR #1467, opened 2019-08-14
Fix FreeBSD

### LINKED ISSUE #1466: 0.2.61 regression
The 0.2.61 release causes the Tokio CI to fail on FreeBSD:

Failure: https://cirrus-ci.com/task/5089534616797184

I tested the CI run with 0.2.60 and it passes: https://github.com/tokio-rs/tokio/pull/1434

I checked the diff, but did not see anything that would obviously cause this. I will continue to investigate, but any advise would be appreciated.

gnzlbg:
#1440 broke FreeBSD by changing `libc` FreeBSD targets to require a `cfg(freebsdXX)` to be defined, but not updating `build.rs` to define `freebsd11` when `LIBC_CI` is not available. Since `LIBC_CI` is always defined on CI, this issue went undetected.

This PR fixes that issue in the `build.rs` and introduces a build task that tests FreeBSD without `LIBC_CI` on FreeBSD11, although I'm  not sure this would have caught the issue in #1466 .

Closes #1466 .

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,5 +1,5 @@
 task:
-  name: stable x86_64-unknown-freebsd-11
+  name: stable x86_64-unknown-freebsd
   freebsd_instance:
     image: freebsd-11-3-stable-amd64-v20190801
   setup_script:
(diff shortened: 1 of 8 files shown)


gnzlbg:
@carllerche thank you for reporting this issue. Would you be able to try the libc branch used by this PR and let us know whether the issue is resolved ?

You would need to amend the libc dependency in your Cargo.toml with:

```toml
libc = { version = "0.2", git = "https://github.com/gnzlbg/libc/", branch = "fix_freebsd" }
```

pizzamig:
Removing the `LIBC_CI` detection, the `freebsd11` is not the default one anymore.
I guess, the `LIBC_CI` detection should be there, adding a `else` branch that forcibly define `freebsd11`

gnzlbg:
I'm not sure what you mean by: 

> Removing the LIBC_CI detection, the freebsd11 is not the default one anymore.
> I guess, the LIBC_CI detection should be there, adding a else branch that forcibly define freebsd11

pizzamig:
I mean, that I misread the patch, sorry.

The patch is explicitly forcing `freebsd11` when no `LIBC_CI` is defined.
It looks good to me.

PR #1470, opened 2019-08-15
add support for sparc64 on OpenBSD

semarie:


gnzlbg:
@bors delegate=semarie 

cc @asomers in case you want to review - otherwise, @semarie if this doesn't break anything on our end, and it works for OpenBSD, i'm fine with it.

My only request is for you to add this particular target to `ci/build.sh` at some point to make sure that this target at least builds correctly.

semarie:
@gnzlbg hey. I missed the addition in `ci/build.sh`. It should be fine now.

semarie:
@gnzlbg I see `style-and-doc` test failed because the target is unknown. should I postpone the inclusion of `sparc64-unknown-openbsd` for after support was added to `rustc` ?

gnzlbg:
You can comment it here for now, and uncomment it once you have added a rustc_target specification for the target. Ideally there would at least be a PR for that before merging this.

semarie:
I removed the commit adding `openbsd-unknown-sparc64` in `ci/build.sh`. Commenting wasn't a solution as it was a part of a shell variable. It will be readded later.

Waiting the PR in rustc is possible, but I expected to do *one* PR in rustc, and not one for the target and another for updating libc. I am preparing a PR for rustc.

gnzlbg:
@bors: r+

I trust you on this, you have done this before.

PR #1468, opened 2019-08-14
Remove sendfile() from Fuchsia

Draphar:
#849 moved Fuchsia into its own module, but it was not checked whether it actually supports all these functions.
[This commit](https://fuchsia.googlesource.com/fuchsia/+/4de4800d4aa442f38cb6be2876b7a6d9f49e219b%5E%21/#F0) shows that the Linux `sendfile()` is not available on Fuchsia.
Furthermore, it is not listed in the [Zircon System Calls](https://fuchsia.dev/fuchsia-src/zircon/syscalls.md) page.

PR #1472, opened 2019-08-15
Remove invalid Fuchsia functions

josephlr:
Extension of #1468 

These functions are not present in Fuchsia's libc, so they cannot be
used in a program today. If Fuchsia ever decides to add these functions
to their libc, they can be added back as necessary.

These files were generated in the following manner:

1. Create a list of the current functions in Fuchsia's libc, put in `rust-functions.txt`
2. Change directory into [`fuchsia/zircon/third_party/ulib/musl/include`](https://fuchsia.googlesource.com/fuchsia/+/master/zircon/third_party/ulib/musl/include/)
3. Run a script to determine which symbols in `rust-functions.txt` are not in the include dir:
    ```bash
    for i in $(cat rust-functions.txt);
        if ! rg $i > /dev/null; then
            echo $i
        fi
    done
    ```
4. This produces [a list of 99 functions](https://github.com/rust-lang/libc/pull/1468#issuecomment-521777743), which are then removed from this crate.

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -3460,7 +3460,6 @@ extern {
     pub fn access(path: *const c_char, amode: ::c_int) -> ::c_int;
     pub fn alarm(seconds: ::c_uint) -> ::c_uint;
     pub fn chdir(dir: *const c_char) -> ::c_int;
-    pub fn fchdir(dirfd: ::c_int) -> ::c_int;
     pub fn chown(path: *const c_char, uid: uid_t,
                  gid: gid_t) -> ::c_int;
     pub fn lchown(path: *const c_char, uid: uid_t,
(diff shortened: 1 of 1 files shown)


josephlr:
@cramertj is what I'm doing here reasonable

cramertj:
Seems plausible to me! I'm curious if you've checked outside of the musl include directory to see if any of these functions are defined by other Fuchsia libraries. I wouldn't think so, and a cursory check on my end didn't reveal anything, but I figured I should ask anyhow.

josephlr:
> Seems plausible to me! I'm curious if you've checked outside of the musl include directory to see if any of these functions are defined by other Fuchsia libraries. I wouldn't think so, and a cursory check on my end didn't reveal anything, but I figured I should ask anyhow.

I just grepped the whole source tree, looks like the functions I removed are not defined anywhere else, so this patch looks good.

cramertj:
Not sure if I have rights on this repo, but I can try!

@bors: r+

cramertj:
Shucks, have to wait for @gnzlbg then.

gnzlbg:
Thank you both. It looks good to me. 

The latest nightly broke libc build, but there is a PR in the queue that fixes that. I'll merge this as soon as that one land (its unclear whether it might need some other fixes). 

Once this lands, would it be possible for any of you to run a `TARGET=<target name> ci/run.sh` of libc in an actual Fuchsia host and report the results ?

PR #1477, opened 2019-08-17
Update newlib socket types to FreeBSD

leo60228:


gnzlbg:
These are breaking changes to all newlib targets, not just FreeBSD right ?

leo60228:
@gnzlbg As far as I'm aware, only me and @FenrirWolf are actually libc newlib.

gnzlbg:
ok, @bors: r+

FenrirWolf:
@gnzlbg yeah, the newlib bindings were added to support a custom nintendo 3ds homebrew target, and now @leo60228 is using them for a similar toolchain for the nintendo switch. No idea if anyone else has ever tried using these bindings.

Technically these changes will break networking for my target, but one way around that would be to put the old definitions in the `arm` subfolder and have the new ones under `aarch64`

gnzlbg:
How are you compiling with these? Do you think it would be possible for you to try to set up some docker containers so that we can run run-time tests for these?

leo60228:
@gnzlbg I have a Docker image at `rustyhorizon/docker`, but I haven't been able to get binaries to run in an emulator. I think somebody was trying to set up a CI farm out of broken Switches.

PR #1473, opened 2019-08-15
Some update for vxWorks

BaoshanPang:
This is to address issues:
rust-lang#1469 and rust-random/getrandom#86 (comment)

   
What have been changed in this PR:

 1. adding randBytes(), randABytes(), randUBytes(), randSecure() and taskDelay()
 2. change armv7-wrs-vxworks to armv7-wrs-vxworks-eabihf
 3. code cleanup

PR #1479, opened 2019-08-17
Update minor patch version to 0.2.63

semarie:
it will permit to me to [update several projects](https://github.com/openbsd/ports/blob/5ff04b132c6718b4ca7401ce289a6f5db801e2ee/lang/rust/Makefile#L5) on OpenBSD ports tree to build on `sparc64-unknown-openbsd` target.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "libc"
-version = "0.2.62"
+version = "0.2.63"
 authors = ["The Rust Project Developers"]
 license = "MIT OR Apache-2.0"
 readme = "README.md"


semarie:
CI is failing due to the use of `armv7-wrs-vxworks` target in "style & doc" test.

#1473 has code in `ci/build/sh` to rename it. I dunno if it should be landed first, or if an independent PR should be done just for renaming the target, as it will be blocking for homu until it is corrected.

xen0n:
This collides with #1449 in version bumping. We should keep only one...

semarie:
@xen0n as you rebased your PR to include latest changes (the one I need), and @gnzlbg  has already `r+` your PR, it is fine for me.

PR #1481, opened 2019-08-17
Define newlib socket types by target arch

FenrirWolf:
This is a follow-up to https://github.com/rust-lang/libc/pull/1477. As mentioned in that PR, the primary (only?) users of these newlib definitions are myself and @leo60228. This PR makes it so that my arm-based 3DS target keeps using the old socket definitions while leo's aarch64-based Switch target uses the BSD versions.

The 3DS also has inverted definitions for `POLLHUP` and `POLLOUT` for whatever reason, so I figured I'd take care of that here too.


gnzlbg:
Style check is failing.

leo60228:
Looks like that was a timeout, not an actual error.

PR #1448, opened 2019-07-30
Add ioctl() constants for BPF

equal-l2:

PR #1482, opened 2019-08-18
several const ioctl() BPF aren't defined on OpenBSD

semarie:
unbreak OpenBSD after #1448

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -264,6 +264,7 @@ fn test_openbsd(target: &str) {
         "netinet/ip.h",
         "netinet/tcp.h",
         "netinet/udp.h",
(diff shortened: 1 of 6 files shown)


semarie:
I could do better if I include `net/bpf.h` in `libc-test`

semarie:
@gnzlbg I looked deeper to include `net/bpf.h` in openbsd ci, and redefine what's need.

ioctl not defined on OpenBSD:
- BIOCGSEESENT
- BIOCSSEESENT

ioctl with different value on OpenBSD:
- BIOCGRSIG
- BIOCSRSIG
- BIOCSDLT

equal-l2:
@semarie Thank you for the follow-up!

PR #1487, opened 2019-08-21
Adding UTIME_NOW and UTIME_OMIT to dragonfly

lpetre:
Verified in both repos:
 - [musl](https://git.musl-libc.org/cgit/musl/tree/include/sys/stat.h#n70)
 - [dragonfly](https://github.com/DragonFlyBSD/DragonFlyBSD/blob/e7ab884bd49753f8884eb597d10d6569a08fa0df/sys/sys/stat.h#L250)


lpetre:
Doing this makes me wish there was some tooling to generate constants from a given body of C code (ie a git url)

gnzlbg:
> Doing this makes me wish there was some tooling to generate constants from a given body of C code (ie a git url)

You can probably use rust-bindgen for that.

@bors: r+

PR #1488, opened 2019-08-22
Copy structs from bits/user.h for musl x86_64

loganwendholt:
While statically compiling a binary with `musl`, I ran into the following error regarding a missing struct:

```
error[E0412]: cannot find type `user_regs_struct` in crate `libc`
  --> src/debug.rs:37:32
   |
37 | fn show_user_regs(regs: &libc::user_regs_struct) -> String {
   |                                ^^^^^^^^^^^^^^^^ not found in `libc`
```

This struct was previously added for `glibc` in #599, but was never added to `musl`, despite the data format being the same in both.

This fix simply copies `user_regs_struct` into the proper location within the `musl` files.


gnzlbg:
Looks like the type definition does not match the C one for musl on some targets.

loganwendholt:
Type definitions should all be correct now.

loganwendholt:
I'm a bit stumped on this one. Looks like the `homu` test has two build scripts that fail: `rand_chacha v0.1.1` and `rand_pcg v0.1.2`. This is strange to me for two reasons:

1) This test is using `nightly x86_64-unknown-freebsd-12`, while the code change in the PR is specific to `musl` targets. It's not clear to me why the change should impact any of the FreeBSD targets at all.

2) In the _other_ `nightly x86_64-unknown-freebsd-12` test, these same two packages built successfully. So why are they failing in the `homu` test?

gnzlbg:
It's not this PR fault, see #1489

loganwendholt:
@gnzlbg since the current build failure is unrelated to this change, are we okay to move forward?

PR #1490, opened 2019-08-24
[WIP] Test FreeBSD failure

gnzlbg:


.cirrus.yml
```diff
@@ -24,4 +10,4 @@ task:
     - rustup default nightly
   test_script:
     - . $HOME/.cargo/env
-    - cargo test
+    - gdb --args cargo build
```
mati865: You have to run it in batch mode like here https://stackoverflow.com/a/20181664/5940067

PR #1491, opened 2019-08-24
Add FreeBSD10 build job

gnzlbg:


libc-test/build.rs
```diff
@@ -1609,12 +1641,28 @@ fn test_freebsd(target: &str) {
...
+    });
+
     cfg.skip_fn(move |name| {
         // skip those that are manually verified
         match name {
             // FIXME: https://github.com/rust-lang/libc/issues/1272
             "execv" | "execve" | "execvp" | "execvpe" | "fexecve" => true,
 
+            // These functions were added in FreeBSD 11:
+            "fdatasync" | "aio_waitcomplete" | "mq_getfd_np"
```
asomers: Actually, `aio_waitcomplete` has been present since FreeBSD 4.  But the return type changed in 11.0 (SVN r297167).

gnzlbg: Thanks, fixed.

libc-test/build.rs
```diff
@@ -1466,7 +1467,10 @@ fn test_freebsd(target: &str) {
     // Required for `getline`:
     cfg.define("_WITH_GETLINE", None);
     // Required for making freebsd11_stat available in the headers
-    cfg.define("_WANT_FREEBSD11_STAT", None);
+    match freebsd_ver {
+        Some(10) => &mut cfg,
+        _ => cfg.define("_WANT_FREEBSD11_STAT", None),
```
asomers: You shouldn't need this symbol . This is only needed on 12+ to expose the backwards-compatible definition of `struct stat` under a different name, and libc doesn't appear to be doing anything with it.

libc-test/build.rs
```diff
@@ -1594,6 +1598,34 @@ fn test_freebsd(target: &str) {
...
+            | "PD_CLOEXEC"
+            | "PD_ALLOWED_AT_FORK"
+            | "IP_RSS_LISTEN_BUCKET"
+                if Some(10) == freebsd_ver =>
+            {
+                true
+            }
+
+            // FIXME: This constant has a different value in FreeBSD 10:
+            "RLIM_NLIMITS" if Some(10) == freebsd_ver => true,
```
asomers: This constant shouldn't even be defined by libc.  I'm going to open a new PR to deprecate it.

PR #1495, opened 2019-09-01
Fix the link_name for fstat and fstatfs on FreeBSD

asomers:


gnzlbg:
I've merged this manually because its required to fix CI.

@asomers would you mind taking #1491 and pushing it through the finish line so that we can land it as well ?

asomers:
Actually, I'm pretty sure that it isn't required to fix CI.  `statfs` is rarely used because it's non-portable.  Portable software uses `statvfs` instead.

PR #1496, opened 2019-09-02
linux/gnu: add utmp(5) default paths

lucab:
This adds `utmp(5)` default paths on Linux with GNU libc.

Ref: https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/paths.h;h=1342ab3a96ab12065311b718e57e7235e07587f8;hb=HEAD

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -499,6 +499,10 @@ pub const O_ACCMODE: ::c_int = 3;
 pub const ST_RELATIME: ::c_ulong = 4096;
 pub const NI_MAXHOST: ::socklen_t = 1025;
 
+// paths.h
+pub const _PATH_UTMP: &'static [u8; 14usize] = b"/var/run/utmp\0";
+pub const _PATH_WTMP: &'static [u8; 14usize] = b"/var/log/wtmp\0";
+
 pub const ADFS_SUPER_MAGIC: ::c_long = 0x0000adf5;
 pub const AFFS_SUPER_MAGIC: ::c_long = 0x0000adff;
 pub const CODA_SUPER_MAGIC: ::c_long = 0x73757245;


gnzlbg:
These should be extern statics without a default value.

lucab:
@gnzlbg thanks for the reviews!

I have a couple of doubts when trying to apply your suggestion above (and I didn't spot any similar example in `libc`):
 * what's the actual type signature? I mean, is it a pointer or an actual array of bytes? If the latter, is the size harcoded based on the current observed value?
 * for the final linkage, where would the symbols actually be defined?

I tried defining a bare extern static, but then consumers fail to link with `undefined reference to `_PATH_UTMP'`.
I ran an `objdump -Ttx` on my local glibc, and indeed I cannot see the symbol in there.

That seems consistent to me with the fact that those are C pre-processor definitions and not actual constants, so they may not end up anywhere in the (public) ABI.

I also run latest `bindgen` on the header and its suggestion seems to agree on re-defining them:
```
pub const _PATH_UTMP: &'static [u8; 14usize] = b"/var/run/utmp\0";
```

gnzlbg:
What bindgen suggests sounds correct.

lucab:
Ack. I've updated the PR with that.

gnzlbg:
There is an assertion triggering in ctest, which I believe is maybe due to ctest not supporting byte strings, e.g., you probably want to add a test for byte strings here: https://github.com/gnzlbg/ctest/blob/master/testcrate/src/t1.rs#L6 and then work up your way to adding support to ctest for these. Shouldn't be too hard since &'static str is already supported.

joshtriplett:
Because these are preprocessor definitions, they don't really have a precise type. There isn't a symbol being bound to here, and nothing will break if we give them a better type for Rust. As a result, I think it'd make more sense to define these as `&'static str`, and adapt ctest to match.

(Also, as a side note, whatever's depending on utmp/wtmp/btmp paths, please make it cope with those files not existing, as they aren't *required* to exist.)

I don't think we should add all of these, unless someone specifically needs them. Adding the ones people specifically ask for, that aren't historical relics, seems reasonable.

JohnTitor:
Closing as this is inactive for years and in favor of Josh's comment. Feel free to re-open if you really want them, we can re-consider it anytime.

PR #1497, opened 2019-09-02
Deprecate RLIM_NLIMITS

asomers:
This constant is not stable across OS versions, so it cannot be used in
any backwards- or forwards- compatible way.  It's typically used to size
arrays in the kernel and in debugging utilities that are closely tied to
the OS version.  Since libc is ignorant about OS versions, we shouldn't
even be defining it.


semarie:
@asomers I am not sure to understand the rational here. lot of things are not stable across OS versions. do you want to deprecate them all ?

asomers:
@semarie Pretty much.  libc's versioning model is basically "adding new stuff is ok; but don't change anything already defined".  That's also more or less what glibc does.  Right now, libc has no way to guarantee that `RLIM_NLIMITS` will have the correct value.  If you try to use that symbol, you're likely to experience errors at runtime.

PR #1498, opened 2019-09-02
Deprecate AIO_LISTIO_MAX

asomers:
This value can vary at runtime.  Applications should instead use
sysconf(3) with _SC_AIO_LISTIO_MAX.


asomers:
BTW, POSIX says that this constant shouldn't even be defined on systems (like FreeBSD and Apple) that publish its value through sysconf:
https://pubs.opengroup.org/onlinepubs/009695399/basedefs/limits.h.html

asomers:
Looks like the same libc bug that affects the nightly Rust stdlib.

gnzlbg:
@bors: retry

> Looks like the same libc bug that affects the nightly Rust stdlib.

The failure is in azure.

PR #554, opened 2017-03-23
add renameat2 and flags

Mic92:

PR #1508, opened 2019-09-11
add renameat2 and flags

phi-gamma:
This used to be PR #554.

Adapted to match the current state of *libc*. Also, adapt for the
fact that in Linux, the *flags* argument is of type *unsigned*,
not *int*.

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -978,6 +978,9 @@ extern "C" {
         needle: *const ::c_void,
         needlelen: ::size_t,
     ) -> *mut ::c_void;
+    pub fn renameat2(olddirfd: ::c_int, oldpath: *const ::c_char,
+                     newdirfd: ::c_int, newpath: *const ::c_char,
+                     flags: ::c_uint) -> ::c_int;
 }
 
 #[link(name = "util")]
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -1226,9 +1226,9 @@ pub const PTHREAD_PROCESS_PRIVATE: ::c_int = 0;
 pub const PTHREAD_PROCESS_SHARED: ::c_int = 1;
 pub const __SIZEOF_PTHREAD_COND_T: usize = 48;
 
-pub const RENAME_NOREPLACE: ::c_int = 1;
-pub const RENAME_EXCHANGE: ::c_int = 2;
-pub const RENAME_WHITEOUT: ::c_int = 4;
+pub const RENAME_NOREPLACE: ::c_uint = 1;
+pub const RENAME_EXCHANGE: ::c_uint = 2;
+pub const RENAME_WHITEOUT: ::c_uint = 4;
 
 pub const SCHED_OTHER: ::c_int = 0;
 pub const SCHED_FIFO: ::c_int = 1;


phi-gamma:
Note that I changed the definition of the flags to match the
signature of the syscall argument. IMO this is the right thing
to do since ``renameat2()`` is the sole syscall the macros
are used with.

This means the type now diverges from those in
``src/fuchsia/mod.rs``. Not sure whether this is an issue
for you guys.

gnzlbg:
> This means the type now diverges from those in
src/fuchsia/mod.rs. Not sure whether this is an issue
for you guys.

Not an issue, but @josephlr might want to take a look at those.

Looking at the CI errors, these are apparently broken on musl. You might want to move these to the `gnu` module. Other than that, this looks good to merge to me.

josephlr:
> Not an issue, but @josephlr might want to take a look at those.

Seems fine. This divergence is expected and shouldn't break anything.

phi-gamma:
Regarding the CI issues: I have rustfmt 1.3 (stable) here which doesn’t catch
that formatting issue, so I applied the changes as the version in CI suggests.

Can’t comment on that linker error on sparc or what seems to be an update
problem on arm.

gnzlbg:
`cargo +nightly fmt --all` should use the nightly `rustfmt` if you have a nightly toolchain installed, even if it is not your default.

phi-gamma:
Sorry, there is no nightly toolchain on stock Fedora 30 and rustup
is allergic against the system toolchain. I’m going to have to do that
at home later.

gnzlbg:
@cuviper might be able to give some advice.

I don't think this problem has a great solution. Every stable rustfmt point release produces different formatting as well, due to bugs being fixed, etc. So picking one single rustfmt version isn't really a great solution to this problem either :/ I'm open to suggestions. We definitely have some freedom about which rustfmt version we pick. 

We could also just use rustfmt in an advisory position, and just reformat the library before each release. But sadly rustfmt isn't always idempotent (e.g. `cargo fmt --all` might produce output that fails a subsequent `cargo fmt --all -- --check`), so I worry that would create a higher maintanence burden. I obviously find it easier if we only merge PRs that are properly formatted.

mati865:
Rustup can be made to live alongside system Rust toolchain: https://github.com/rust-lang/rustup.rs#working-with-distribution-rust-packages

cuviper:
Yeah, I also use rustup that way myself.

Picking a rustfmt version for CI is hard, but I don't think tracking nightly is the best. If nothing else, it's annoying for contributors if they have to fix unrelated formatting in their PR, just because it changed in nightly. Implicitly tracking stable is a little better because that changes less frequently. But I think it's best to choose an explicit stable version, even if as a maintainer you always bump it to latest stable -- that act of bumping can catch up on any formatting changes at the same time.

gnzlbg:
FWIW the main reason libc uses nightly rustfmt is that you need nightly rust to run `libc` tests. As in, if you are developing a PR, without nightly rust you can at best guess if its going to work on CI.

cuviper:
Sure, that's reasonable.

phi-gamma:
@mati865: Those instructions proved incomplete for me; rustup outright
refuses to work unless I pass ``RUSTUP_INIT_SKIP_PATH_CHECK=yes``.

gnzlbg:
@bors: retry

---

There is a failure in sparc64, and it appears legit, but lets try one more time.

fbernier:
Hey, seems like there was an issue merging this in but the build is now gone. Can we try again and I'll check out the failure if there's one?

PR #1517, opened 2019-09-13
updates for vxworks

BaoshanPang:
fix problems found in testing
VxWorks socket and epoll definitions needed by mio and related crates; epoll only supported in VxWorks kernel; just provide stub definitions for user mode for now
Fixed usage of isize
libc VxWorks cleanups and additions
remove copyright notice

### DIFF
--- a/src/vxworks/mod.rs
+++ b/src/vxworks/mod.rs
@@ -551,50 +551,35 @@ enum nfsstat {
     NFSERR_BAD_COOKIE = 10003,
     NFSERR_NOTSUPP = 45,
     NFSERR_TOOSMALL = 10005,
-//    NFSERR_SERVERFAULT = 10006,
     NFSERR_BADTYPE = 10007,
     NFSERR_JUKEBOX = 10008,
 }
 
 pub const S_nfsLib_NFS_OK: ::c_int = nfsstat::NFS_OK as ::c_int;
-pub const S_nfsLib_NFSERR_PERM: ::c_int =
-    nfsstat::NFSERR_PERM as ::c_int;
-pub const S_nfsLib_NFSERR_NOENT: ::c_int =
-    nfsstat::NFSERR_NOENT as ::c_int;
-pub const S_nfsLib_NFSERR_IO: ::c_int =
-    nfsstat::NFSERR_IO as ::c_int;
-pub const S_nfsLib_NFSERR_NXIO: ::c_int =
-    nfsstat::NFSERR_NXIO as ::c_int;
-pub const S_nfsLib_NFSERR_ACCESS: ::c_int =
-    nfsstat::NFSERR_ACCESS as ::c_int;
-pub const S_nfsLib_NFSERR_EXIST: ::c_int =
-    nfsstat::NFSERR_EXIST as ::c_int;
+pub const S_nfsLib_NFSERR_PERM: ::c_int = nfsstat::NFSERR_PERM as ::c_int;
+pub const S_nfsLib_NFSERR_NOENT: ::c_int = nfsstat::NFSERR_NOENT as ::c_int;
+pub const S_nfsLib_NFSERR_IO: ::c_int = nfsstat::NFSERR_IO as ::c_int;
+pub const S_nfsLib_NFSERR_NXIO: ::c_int = nfsstat::NFSERR_NXIO as ::c_int;
+pub const S_nfsLib_NFSERR_ACCESS: ::c_int = nfsstat::NFSERR_ACCESS as ::c_int;
+pub const S_nfsLib_NFSERR_EXIST: ::c_int = nfsstat::NFSERR_EXIST as ::c_int;
 pub const S_nfsLib_NFSERR_XDEV: ::c_int =
     M_nfsStat | nfsstat::NFSERR_XDEV as ::c_int;
 pub const S_nfsLib_NFSERR_NODEV: ::c_int =
     M_nfsStat | nfsstat::NFSERR_NODEV as ::c_int;
-pub const S_nfsLib_NFSERR_NOTDIR: ::c_int =
-    nfsstat::NFSERR_NOTDIR as ::c_int;
-pub const S_nfsLib_NFSERR_ISDIR: ::c_int =
-    nfsstat::NFSERR_ISDIR as ::c_int;
-pub const S_nfsLib_NFSERR_INVAL: ::c_int =
-    nfsstat::NFSERR_INVAL as ::c_int;
-pub const S_nfsLib_NFSERR_FBIG: ::c_int =
-    nfsstat::NFSERR_FBIG as ::c_int;
-pub const S_nfsLib_NFSERR_NOSPC: ::c_int =
-    nfsstat::NFSERR_NOSPC as ::c_int;
-pub const S_nfsLib_NFSERR_ROFS: ::c_int =
-    nfsstat::NFSERR_ROFS as ::c_int;
+pub const S_nfsLib_NFSERR_NOTDIR: ::c_int = nfsstat::NFSERR_NOTDIR as ::c_int;
+pub const S_nfsLib_NFSERR_ISDIR: ::c_int = nfsstat::NFSERR_ISDIR as ::c_int;
+pub const S_nfsLib_NFSERR_INVAL: ::c_int = nfsstat::NFSERR_INVAL as ::c_int;
+pub const S_nfsLib_NFSERR_FBIG: ::c_int = nfsstat::NFSERR_FBIG as ::c_int;
+pub const S_nfsLib_NFSERR_NOSPC: ::c_int = nfsstat::NFSERR_NOSPC as ::c_int;
+pub const S_nfsLib_NFSERR_ROFS: ::c_int = nfsstat::NFSERR_ROFS as ::c_int;
 pub const S_nfsLib_NFSERR_MLINK: ::c_int =
     M_nfsStat | nfsstat::NFSERR_MLINK as ::c_int;
 pub const S_nfsLib_NFSERR_NAMETOOLONG: ::c_int =
     nfsstat::NFSERR_NAMETOOLONG as ::c_int;
 pub const S_nfsLib_NFSERR_NOTEMPTY: ::c_int =
     nfsstat::NFSERR_NOTEMPTY as ::c_int;
-pub const S_nfsLib_NFSERR_DQUOT: ::c_int =
-    nfsstat::NFSERR_DQUOT as ::c_int;
-pub const S_nfsLib_NFSERR_STALE: ::c_int =
-    nfsstat::NFSERR_STALE as ::c_int;
+pub const S_nfsLib_NFSERR_DQUOT: ::c_int = nfsstat::NFSERR_DQUOT as ::c_int;
+pub const S_nfsLib_NFSERR_STALE: ::c_int = nfsstat::NFSERR_STALE as ::c_int;
 pub const S_nfsLib_NFSERR_WFLUSH: ::c_int =
     M_nfsStat | nfsstat::NFSERR_WFLUSH as ::c_int;
 pub const S_nfsLib_NFSERR_REMOTE: ::c_int =
@@ -609,8 +594,7 @@ pub const S_nfsLib_NFSERR_NOTSUPP: ::c_int =
     nfsstat::NFSERR_NOTSUPP as ::c_int;
 pub const S_nfsLib_NFSERR_TOOSMALL: ::c_int =
     M_nfsStat | nfsstat::NFSERR_TOOSMALL as ::c_int;
-pub const S_nfsLib_NFSERR_SERVERFAULT: ::c_int =
-    nfsstat::NFSERR_IO as ::c_int;
+pub const S_nfsLib_NFSERR_SERVERFAULT: ::c_int = nfsstat::NFSERR_IO as ::c_int;
 pub const S_nfsLib_NFSERR_BADTYPE: ::c_int =
     M_nfsStat | nfsstat::NFSERR_BADTYPE as ::c_int;
 pub const S_nfsLib_NFSERR_JUKEBOX: ::c_int =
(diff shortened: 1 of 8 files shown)


BaoshanPang:
I am going to fix the issue reported by rustfmt.

BaoshanPang:
@gnzlbg 
cc @n-salim 

I made some change to fix the erros reported by rustfmt.

BaoshanPang:
> Why does it fail? It shows an error
Sorry for my broken English, I meant it shows an error.

BaoshanPang:
> Note that the const above the enum is not public either, so no user of libc can use any of these two. If they are not used somewhere else inside the library, then.. they are not used ever.

Thanks for the explanation.

BaoshanPang:
@gnzlbg 
Can you help to explain what this error means:
 Running `CARGO_MANIFEST_DIR=/home/vsts/work/1/s CARGO_PKG_VERSION_PRE= CARGO_PKG_REPOSITORY='***' CARGO_PKG_VERSION_MINOR=2 CARGO_PKG_AUTHORS='The Rust Project Developers' OUT_DIR=/home/vsts/work/1/s/target/armv7-wrs-vxworks-eabihf/debug/build/libc-194b416859615cc0/out CARGO_PKG_VERSION_MAJOR=0 LD_LIBRARY_PATH='/home/vsts/work/1/s/target/debug/deps:/usr/share/rust/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib:/home/vsts/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib' CARGO_PKG_DESCRIPTION='Raw FFI bindings to platform libraries like libc.
' CARGO=/usr/share/rust/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/cargo CARGO_PKG_VERSION=0.2.64 CARGO_PKG_HOMEPAGE='***' CARGO_PKG_NAME=libc CARGO_PKG_VERSION_PATCH=64 rustc --crate-name libc src/lib.rs --color never --crate-type lib --emit=dep-info,metadata,link -C debuginfo=2 --cfg 'feature="extra_traits"' -C metadata=d1551dc5066b0da9 -C extra-filename=-d1551dc5066b0da9 --out-dir /home/vsts/work/1/s/target/armv7-wrs-vxworks-eabihf/debug/deps --target armv7-wrs-vxworks-eabihf -C incremental=/home/vsts/work/1/s/target/armv7-wrs-vxworks-eabihf/debug/incremental -L dependency=/home/vsts/work/1/s/target/armv7-wrs-vxworks-eabihf/debug/deps -L dependency=/home/vsts/work/1/s/target/debug/deps --sysroot /home/vsts/work/1/s/target/sysroot --cfg freebsd11 --cfg libc_deny_warnings --cfg libc_priv_mod_use --cfg libc_union --cfg libc_const_size_of --cfg libc_align --cfg libc_core_cvoid --cfg libc_packedN`
error[E0277]: arrays only have std trait implementations for lengths 0..=32
   --> src/vxworks/mod.rs:170:9
    |
107 | / s! {
108 | |     pub struct _Vx_semaphore {}
109 | |
110 | |     // b_pthread_condattr_t.h
...   |
170 | |         pub __ss_pad2  : [::c_char; _SS_PAD2SIZE],
    | |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `core::array::LengthAtMost32` is not implemented for `[u8; 120]`
...   |
432 | |     }
433 | | }
    |

src/vxworks/mod.rs
```diff
@@ -65,24 +66,19 @@ pub type uid_t = ::c_ushort;
...
 
 pub type pthread_key_t = ::c_ulong;
 
 // From b_off_t.h
 pub type off64_t = ::c_longlong;
-pub type off_t64 = ::c_longlong;
 
 // From b_BOOL.h
-pub type BOOL = ::c_int; // excuse me what
+pub type BOOL = ::c_int;
```
gnzlbg: What is the layout of `_Bool` in C and `bool` in C++ for this target ? 

What does `sizeof(_Bool)` and `sizeof(bool)` report ?

src/vxworks/mod.rs
```diff
@@ -434,33 +385,58 @@ s! {
...
+        pub dli_fname: *const ::c_char,
+        pub dli_fbase: *mut ::c_void,
+        pub dli_sname: *const ::c_char,
+        pub dli_saddr: *mut ::c_void,
+    }
+}
+
+s_no_extra_traits! {
+    // dirent.h
+    #[allow(missing_debug_implementations)]
```
gnzlbg: You probably need to manually implement `Debug` for these. This is quite common; you can see how other modules do it.

src/vxworks/mod.rs
```diff
@@ -12,6 +12,9 @@ impl ::Clone for DIR {
     }
 }
 
+// Throughout we use usize / isize for types that are
+// (unsigned) int in ILP32 and (unsigned) long in ILP64
```
gnzlbg: `usize == uintptr_t` and `isize == intptr_t` 

`unsigned int` is `c_uint` and `unsigned long` is `c_ulong`.

src/vxworks/mod.rs
```diff
@@ -100,43 +96,22 @@ pub type SD_ID = ::OBJ_HANDLE;
...
-        pub magic  : ::c_uint,
-        pub semType: ::c_uint,
-        pub options: ::c_uint,
-        pub recurse: ::c_uint,
-        pub owned_k: ::c_uint, // owned_k is volatile
-        pub semId_k: ::_Vx_SEM_ID_KERNEL,
-        pub state  : ::c_uint, //state is union of _Vx_UINT and _Vx_UINT
-        pub events : ::_Vx_EVENTS_RSRC,
-    }
+    pub struct _Vx_semaphore {}
```
gnzlbg: What's the definition of this C type ?

src/vxworks/mod.rs
gnzlbg: You can use `cfg(target_pointer_width = "...")` to use different types for different pointer widths.

src/vxworks/mod.rs
```diff
@@ -2094,6 +2138,43 @@ pub fn posix_memalign(
     }
 }
 
+// epoll.h
+// Unfortunately epoll is currently only supported in the VxWorks kernel
```
gnzlbg: Why are these here then?

gnzlbg:
I've left some comments. I hope they help. Let me know if you get stuck.

src/vxworks/mod.rs
BaoshanPang: We are going to support it in user land.

gnzlbg: How? IIUC you'd need to provide an implementation in user-land for those, and this crate wouldn't be the place to do that (this crate only provides FFI bindings).

BaoshanPang: I will remove this part.

src/vxworks/mod.rs
BaoshanPang: Do we have to add the definition here? We want to hide the internal details for this struct if possible.

src/vxworks/mod.rs
BaoshanPang: I see in many files, the big size arrary is ignored so far in the debug function, for example:
        impl ::fmt::Debug for sockaddr_un {
                f.debug_struct("sockaddr_un")
                    .field("sun_family", &self.sun_family)
                // FIXME: .field("sun_path", &self.sun_path)                                                                         
                    .finish()
            }
        }

Should I do it in the same way? Do we have a workable solution for these big size arrary?

gnzlbg: IIRC there are a couple of examples that use large arrays there.

src/vxworks/mod.rs
gnzlbg: I don't know. Why are you adding it here ? (that's why I asked for the C definition, if this is a public type that's used on some public API, then it might make sense adding it here, but since C does not have types of zero-size, the looks of this type are quite suspicious).

src/vxworks/mod.rs
BaoshanPang: BOOL is a type we used in vxWorks, it is different than _Bool or bool.

src/vxworks/mod.rs
BaoshanPang: Can you give an examaple? I can't find any of them so far.

gnzlbg: Sure: https://github.com/rust-lang/libc/blob/9b936e30116da305b799a74491c82ee2d2e8f678/src/unix/linux_like/linux/mod.rs#L749

You just print it as a slice `&[..]`

src/vxworks/mod.rs
BaoshanPang: I will remove "pub" for the struct. We only want o make the type of pointer to this struct be public which is defined at line 90:

pub type _Vx_SEM_ID = *mut ::_Vx_semaphore;

src/vxworks/mod.rs
BaoshanPang: Thanks.

src/vxworks/mod.rs
BaoshanPang: Removing "pub" would result compiling error.

BaoshanPang: I find a example in hermit/mod.rs which is using the similiary code:
    pub struct utsname {}
...

Is such code not recommented any more?

gnzlbg: Hermit is not a supported platform, in fact it didn't even compile for a long time and nobody noticed. 

The only properly supported platforms are those for which `libc-test` runs and passes correctly on the platform.

gnzlbg: Running libc-test on your platform should catch all of these issues.

src/vxworks/mod.rs
BaoshanPang: I am adding the following code:

cfg_if! {
    if #[cfg(target_pointer_width = "32")] {
        pub type size_t = u32;
        pub type ssize_t = i32;
        pub type ptrdiff_t = i32;
        pub type intptr_t = i32;
        pub type uintptr_t = u32;
    } else if #[cfg(target_pointer_width = "64")] {
        pub type size_t = u64;
        pub type ssize_t = i64;
        pub type ptrdiff_t = i64;
        pub type intptr_t = i64;
        pub type uintptr_t = u64;
    } else {
        // Unknown target_pointer_width                                                                                              
    }
}

But when run the build script (ci/build.sh), I get such error:

    --> src/vxworks/mod.rs:2156:45
     |
2156 |     if (align & align - 1 != 0) || (align % size_of::<::size_t>() != 0) {
     |                                             ^^^^^^^^^^^^^^^^^^^^^ expected u32, found usize

error[E0277]: cannot mod `u32` by `usize

I think this should be the expected result as size_t was defined as usize and right now it is defined as u32.

n-salim: > `usize == uintptr_t` and `isize == intptr_t`
> 
> `unsigned int` is `c_uint` and `unsigned long` is `c_ulong`.

Understood but maybe we just need to clarify the comment rather than changing the definitions of `size_t `et al.

`size_t = usize`, `ssize_t = isize`, ec, are both layout and "in spirit" compatible with VxWorks. I think it will create a lot of unneeded headaches if we change the definitions of these types.

We have recently got the libc-test working on VxWorks and from what I'm told, this "discrepancy" did not show up as an issue.

Furthermore, the [Unix libc](https://github.com/rust-lang/libc/blob/master/src/unix/mod.rs#L19) provides a similar definition of `size_t` etc. in terms of `usize/isize`, even though e.g. on my Linux box the system header definitions of these types are similar to those in VxWorks.

"Morally" these C types are supposed to act like `usize/isize`. E.g. in Rust `mem::size_of` returns something that  should act like C's `size_t,` right? So it makes sense for them to have the same type (assuming of course that the "real" `size_t` is layout compatible with `usize`, which for our targets it is).

src/vxworks/mod.rs
n-salim: @gnzlbg As I mentioned above, we did recently get libc-test working (thanks to your suggestion in an earlier review), and apparently (with the changes under review here) it passes, i.e. it does not complain about this.

As @BaoshanPang mentioned, the intent here is to provide an "opaque" type, i.e. clients of this interface only need to deal with pointers to this type, and should not (nor need to) know the inner workings.

n-salim: However let me double check, because we may be "skipping" some tests.

src/vxworks/mod.rs
n-salim: Just to confirm, from our libc-test logs with the previous "`size_t = usize`" code:

```
[58] size_t size => Passed
[59] size_t align => Passed
[60] size_t signed => Passed
```

(same result on both 32-bit and 64-bit targets).

src/vxworks/mod.rs
n-salim: I think we are skipping `_Vx_semaphore` itself (the type we want to keep opaque), but the types used in the APIs that are defined in terms of this type (e.g. _Vx_SEM_ID), pass the libc-test:

```
[172] _Vx_SEM_ID size => Passed
[173] _Vx_SEM_ID align => Passed

```

src/vxworks/mod.rs
gnzlbg: > align % size_of::<::size_t>()

The problem is that in that line, `align` and `size_of::<...>()` have different types (u32 vs usize), but `%` only works when they have the same size. There are many ways to fix this error, depending on what exactly does the C code do. One way is just adding a cast `align as usize % size_of::<::size_t>()`.

gnzlbg: > Understood but maybe we just need to clarify the comment rather than changing the definitions of size_t et al.

Yes, my original comment was exclusively about the comment. I was just warning you that `usize` and `isize` do not in general map to `unsigned int`/`int` in C - we use `libc::c_uint` and `libc::c_int` for those throughout the code instead.

gnzlbg: So if you have a C struct that uses `int`, you should write here `libc::c_int`, instead of using `isize` for that.

src/vxworks/mod.rs
gnzlbg: > As @BaoshanPang mentioned, the intent here is to provide an "opaque" type, i.e. clients of this interface only need to deal with pointers to this type, and should not (nor need to) know the inner workings.

The way this is done is by using an enum, e.g., `pub enum _Vx_semaphore{}` - users cannot create a value of that type because it has no variants.

n-salim: Thanks, we'll make that change.

src/vxworks/mod.rs
n-salim: > I was just warning you that usize and isize do not in general map to unsigned int/int in C - we use libc::c_uint and libc::c_int for those throughout the code instead.

Right, I think we're in agreement, but the comment was poorly written (my fault). In fact it really only applies to types like `size_t` where (I think) it really does make sense to map them to `usize/isize` (as argued above).

We are correctly using `libc::c_int` etc. for fields whose type really is supposed to be a C `int`.

I suggest moving the comment above the block of code that defines size_t et al, and changing it to:

```
// size_t and similar size types are defined in VxWorks as (unsigned) int
// in ILP32 and (unsigned) long in ILP64, but it is both layout compatible
// and "in the spirit" of the VxWorks definitions and C semantics to define them in
// terms of usize/isize here.
```

gnzlbg: FWIW my recommendation here is try to stick to using the same types as C. Sometimes, using an u32 when C uses an int will work, but is much clearer for everybody to just translate the C sources to Rust "as is".

So if C defines `size_t` as unsigned int, just write:

```rust
type size_t = c_uint;
```

if `c_uint` is the same size of a pointer, you can then also write:

```rust
type c_uint = uintptr_t;
```

gnzlbg: That way, it all just works out nicely, and you don't need a documentation comment.

BaoshanPang: @gnzlbg 
@n-salim

Please help to check my latest update and let me if that's what we want.

src/vxworks/mod.rs
```diff
@@ -100,43 +105,29 @@ pub type SD_ID = ::OBJ_HANDLE;
 pub type CONDVAR_ID = ::OBJ_HANDLE;
 
 // From vxTypes.h
-pub type _Vx_usr_arg_t = ::ssize_t; // c_int for LP32
-pub type _Vx_exit_code_t = ::ssize_t; // c_int for LP32
+pub type _Vx_usr_arg_t = isize;
+pub type _Vx_exit_code_t = isize;
```
gnzlbg: Shouldn't this just be `c_int` ?

gnzlbg:
This LGTM. @bors: r+ 

I would recommend that you commit your changes to libc-test at some point as well.

BaoshanPang:
> I would recommend that you commit your changes to libc-test at some point as well.
Sure, I will do it later.

BaoshanPang:
>  Test failed - [checks-cirrus-freebsd-11](https://cirrus-ci.com/task/5194592124665856)
>>Instance failed to start

Is there some issue in the QA system?

PR #1518, opened 2019-09-16
Add FreeBSD10 support

gnzlbg:
This adds libc-test support for Freebsd10 and a CI build job that tests
FreeBSD10 with LIBC_CI only.

cc @asomers this is a follow up to #1491 .

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,4 +1,18 @@
 task:
+  name: nightly x86_64-unknown-freebsd-10
+  freebsd_instance:
+    image: freebsd-10-4-release-amd64 
+  setup_script:
+    - pkg install -y curl
+    - curl https://sh.rustup.rs -sSf --output rustup.sh
+    - sh rustup.sh --default-toolchain nightly -y
+    - . $HOME/.cargo/env
+    - rustup default nightly
+  test_script:
+    - . $HOME/.cargo/env
+    - LIBC_CI=1 sh ci/run.sh x86_64-unknown-freebsd
+ 
+task:
   name: stable x86_64-unknown-freebsd-11
   freebsd_instance:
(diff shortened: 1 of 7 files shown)


asomers:
I still disagree with this PR on general principle: FreeBSD 10 is EOL so we shouldn't bother to support it anymore.  I assume that reason to do it is that libstd still uses a FreeBSD 10 build host?  If so, can we get them to upgrade their builder instead?

gnzlbg:
> I assume that reason to do it is that libstd still uses a FreeBSD 10 build host?

libstd uses a FreeBSD10 build host, yes.

> If so, can we get them to upgrade their builder instead?

Maybe someone can, but the only thing I can do is try to prevent libc from being broken on FreeBSD10 hosts until libstd migrates away from it so that broken libc's don't end up in libstd, and as a result all FreeBSD hosts end up broken.

asomers:
Does the current situation result in libstd that is broken on FreeBSD 11 and 12 hosts, or only broken on FreeBSD 10 hosts?

gnzlbg:
> Does the current situation result in libstd that is broken on FreeBSD 11 and 12 hosts, or only broken on FreeBSD 10 hosts?

The situation that happened a couple of weeks ago broke rustc on FreeBSD10 and 11 hosts. I think it was a consequence of us not testing FreeBSD10, because the `libc/build.rs` was broken for that host, and ended up picking up freebsd12 symbols, and that resulted in the broken rustc.

gnzlbg:
So I'm not sure what to do about this. If there isn't consensus then I think I'll just close it.

gnzlbg:
I'm merging this. If someone wants to put in the work of upgrading FreeBSD in rust-lang/rust upstream, they can do so, and let us know, and then I'll remove all of this.

@bors: r+

PR #1519, opened 2019-09-17
Implement max_align_t

gnzlbg:
Implement `max_align_t` for windows and unix targets.

WIP: at this stage I just want to see what breaks.


gnzlbg:
cc @ralfjung this will need some ironing out but after this it should be available on a couple of platforms at least.

gnzlbg:
@emilio I'm having some trouble here and maybe you have already run into this. 

I'm trying to wrap `max_align_t`, but that's failing on 32-bit targets. I think I'm getting the call ABI of `max_align_t` wrong. Looking at the C sources, this is often defined as a type alias to `__float128` or __Decimal128`. On x86_64 i'm able to just use a [f64; 2] array, but on 32-bit that's failing, [f32; 4] is failing as well.

retep998:
On Windows the CRT defines `max_align_t` as `double` unconditionally.

gnzlbg:
Thanks peter. Do you know in which headerfile it is? It isn’t in the ones
we are including, and I searched for it and learned that max_align_t is a
C11 constant and MSVC only supports C99 so I have up :/

retep998:
`cstddef` which is from the VC++ headers, not the UCRT headers.
`C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\include\cstddef`
`using max_align_t = double; // most aligned type`

gnzlbg:
> cstddef which is from the VC++ headers, not the UCRT headers.

IIRC libc aimed to only expose access to the UCRT for msvc targets.

retep998:
>max_align_t is a C11 constant and MSVC only supports C99 so I have up :/

Well, `cstddef` is actually a C++ header so apparently MSVC supports `max_align_t` only in C++ and not C.

gnzlbg:
cc @joshtriplett - this would be another item for the FFI WG. It is currently impossible to directly interface with C functions that use `__Float128`, `__Decimal128`, and similar types from Rust, because there is no way to match the call ABI of these types from Rust (e.g. using `[f64; 2]` to interface with `__Float128` will use the wrong registers). 

I suppose @emilio has the same problem. A workaround that bindgen could use is generating and compiling shims on the C side that pass these types by memory (e.g. using `__Float128*)`.

emilio:
Yeah, this may be also related to https://github.com/rust-lang/rust/issues/54341

gnzlbg:
> Yeah, this may be also related to rust-lang/rust#54341

I'm not sure. I wanted to add `__int128` to libc but I haven't done that yet. I was hoping that `#[repr(align(N), C)] struct __int128(i128);` where `N` has the right value would have the right call ABI.

PR #1521, opened 2019-09-17
Fix building error at backtrace

BaoshanPang:
I get this error when builind backtrace:

  --> /folk/prj-rust-dev/bpang/ala-diab-pb19l/fix_bugs/latest/.cargo/git/checkouts/backtrace-rs-fb1f822361417489/e745627/src/symboli\
ze/dladdr.rs:38:73
   |
38 |                         Some(SymbolName::new(slice::from_raw_parts(ptr, len)))
   |                                                                         ^^^ expected usize, found u64
help: you can convert an `u64` to `usize` and panic if the converted value wouldn't fit

It shows the 'size_t' is expected to be defined as "usize".

### DIFF
--- a/src/vxworks/mod.rs
+++ b/src/vxworks/mod.rs
@@ -25,23 +25,11 @@ pub type c_ulonglong = u64;
 pub type intmax_t = i64;
 pub type uintmax_t = u64;
 
-cfg_if! {
-    if #[cfg(target_pointer_width = "32")] {
-        pub type size_t = c_uint;
-        pub type ssize_t = c_int;
-        pub type ptrdiff_t = c_int;
-        pub type intptr_t = c_int;
-        pub type uintptr_t = c_uint;
-    } else if #[cfg(target_pointer_width = "64")] {
-        pub type size_t = c_ulonglong;
-        pub type ssize_t = c_longlong;
-        pub type ptrdiff_t = c_longlong;
-        pub type intptr_t = c_longlong;
-        pub type uintptr_t = c_ulonglong;
-    } else {
-        // Unknown target_pointer_width
-    }
-}
+pub type uintptr_t = usize;
+pub type intptr_t = isize;
+pub type ptrdiff_t = isize;
+pub type size_t = ::uintptr_t;
+pub type ssize_t = ::intptr_t;
 
 pub type pid_t = i32;
 pub type in_addr_t = u32;


gnzlbg:
There is no libc function in that error message.

src/vxworks/mod.rs
```diff
@@ -25,23 +25,11 @@ pub type c_ulonglong = u64;
...
-        pub type size_t = c_ulonglong;
-        pub type ssize_t = c_longlong;
-        pub type ptrdiff_t = c_longlong;
-        pub type intptr_t = c_longlong;
-        pub type uintptr_t = c_ulonglong;
-    } else {
-        // Unknown target_pointer_width
-    }
-}
+pub type size_t = usize;
```
gnzlbg: How is `size_t` defined in the VxWorks headers ? Can you copy that part of the header ?

gnzlbg:
Depending on how the things are defined in the C header, this line might have a bug:

> https://github.com/rust-lang/backtrace-rs/blob/master/src/symbolize/dladdr.rs#L38

Instead of `len`, it should be `len as _`.

BaoshanPang:
> There is no libc function in that error message.

Sorry, I should paste more code:

                 unsafe {
                        let len = libc::strlen(self.inner.dli_sname);
                        Some(SymbolName::new(slice::from_raw_parts(ptr, len)))
                    }

src/vxworks/mod.rs
BaoshanPang: It is defined like this:
#ifdef _WRS_CONFIG_LP64
typedef unsigned long long size_t;
#else
typedef unsigned int size_t;
#endif

gnzlbg: How are uintptr_t and intptr_t defined ?

BaoshanPang:
> Depending on how the things are defined in the C header, this line might have a bug:
> 
> > https://github.com/rust-lang/backtrace-rs/blob/master/src/symbolize/dladdr.rs#L38
> 
> Instead of `len`, it might need to be `len as _`. Hard to tell without seeing the headers.


I see the size_t is defined as usize almost for every targets:

$ grep "pub type size_t" * -rn
cloudabi/mod.rs:14:pub type size_t = usize;
fuchsia/mod.rs:23:pub type size_t = usize;
hermit/mod.rs:29:pub type size_t = usize;
sgx.rs:16:pub type size_t = usize;
switch.rs:16:pub type size_t = usize;
unix/mod.rs:19:pub type size_t = usize;
vxworks/mod.rs:28:pub type size_t = usize;
wasi.rs:16:pub type size_t = usize;
windows/mod.rs:16:pub type size_t = usize;

src/vxworks/mod.rs
BaoshanPang: #ifdef _WRS_CONFIG_LP64
typedef long long intptr_t;
typedef unsigned long long uintptr_t;
#else
typedef int intptr_t;
typedef unsigned int uintptr_t;
#endif

gnzlbg: So I suggest defining `uintptr_t = usize` and `intptr_t = isize`, and using `uintptr_t` to define `size_t`, and `intptr_t` to define `ssize_t`.

BaoshanPang: How about ptrdiff_t?

gnzlbg: It's ok to make it `isize` I think.

BaoshanPang:
It seems the fail is caused by some network issue.

>>E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?

Issue #1520, opened 2019-09-17
Return type of pthread_exit


newpavlov:
If I am not mistaken correct signature for it should be `fn pthread_exit(value: *mut ::c_void) -> !`, while now it returns `()`. [Documentation](https://linux.die.net/man/3/pthread_exit) clearly states that:
>This function does not return to the caller.

gnzlbg:
Agreed, the return type should be `!`.

PR #1525, opened 2019-09-20
pthread_exit return type fix

newpavlov:
Closes: #1520

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -3732,7 +3732,7 @@ extern "C" {
         native: ::pthread_t,
         value: *mut *mut ::c_void,
     ) -> ::c_int;
-    pub fn pthread_exit(value: *mut ::c_void);
+    pub fn pthread_exit(value: *mut ::c_void) -> !;
     pub fn pthread_attr_init(attr: *mut ::pthread_attr_t) -> ::c_int;
     pub fn pthread_attr_destroy(attr: *mut ::pthread_attr_t) -> ::c_int;
     pub fn pthread_attr_setstacksize(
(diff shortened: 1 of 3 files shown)


gnzlbg:
Thank you. I think we should do this, but I want to be careful that we can revert this easily if this breaks too much code. 

I hope that nobody is using the return value of `pthread_exit` in any way, but who knows, so I'm going to land this in the following way:

* [ ] I'm going to do a libc release (0.2.64):
   * [ ] this is currently blocked on deciding what to do with #1518 
* [ ] I'm going to then update libc on rust-lang rust to 0.2.64
   * [ ] and wait a couple of nightlies to see if that goes smoothly
* [ ] At this point we have a 0.2.64 version that's known to work
* [ ] Merge this change, and release 0.2.65 only with this change
   * if people complain too much, we can yank 0.2.65
* [ ] Update rust-lang/rust to 0.2.65.
    * if something breaks, we can yank, and then revert the PR
* [ ] Done, this change is good. 

This is sadly a lot of steps, and will take ~1 week for the 0.2.64 release, and ~1 week for the 0.2.65 release, but such is life. 

If after doing the 0.2.64 thing something high-priority must land upstream (new target, critical target upgrade), I'll prioritize that, and do this in the next release.

newpavlov:
I don't think it should break any code. This snippet works without any issues:
```rust
fn foo() -> ! { panic!(); }
fn main() { let a: u64 = foo(); }
```
And during type inference right now Rust defaults to `()`, e.g. this code:
```
fn foo() -> ! { panic!(); }
fn main() { let a = foo(); a + 1; }
```
Returns "cannot add `i32` to `()`" compilation error.

gnzlbg:
Ah awesome, I thought that `let x = !` would fail to compile.

@bors: r+ then :)

PR #1526, opened 2019-09-20
VxWorks libc cleanups and fixes for some libc-test issues

BaoshanPang:


BaoshanPang:
cc @n-salim

src/vxworks/mod.rs
```diff
@@ -253,11 +252,8 @@ s! {
     }
 
     // signal.h
-    pub struct sigaction { // pulled from kernel side,
-        pub sa_u     : ::size_t,
-        // This is a union of two function pointers.
-        // Under the assumption that function pointers are the same
-        // size as other pointers, we can replace the union with size_t
+    pub struct sigaction {
+        pub sa_u     : ::size_t, // actually union of two function pointers
```
gnzlbg: Why not use an `union` then ?

src/vxworks/mod.rs
```diff
@@ -273,9 +269,7 @@ s! {
     pub struct siginfo_t {
         pub si_signo : ::c_int,
         pub si_code  : ::c_int,
-        // This field is a union of int and void * in vxworks
-        // The size has been set to the larger of the two
-        pub si_value : ::size_t,
+        pub si_value : ::size_t, // actually union of int and void *
```
gnzlbg: Same, you should use unions for unions.

src/vxworks/mod.rs
BaoshanPang: I see on another targets, they are using sighandler_t for similiar case, and sighandler_t is defined as:
unix/mod.rs:30:pub type sighandler_t = ::size_t; 

Can we use the same way on vxWorks? i.e, change size_t to sighandler_t.

gnzlbg: Those other targets where added when Rust did not had `union`s, so they couldn't use them.

src/vxworks/mod.rs
```diff
@@ -431,6 +415,16 @@ s_no_extra_traits! {
         pub __ss_pad2  : [::c_char; _SS_PAD2SIZE],
     }
 
+    pub union sa_u_t {
+        pub sa_handler : extern "C" fn(::c_int) -> !,
```
gnzlbg: These function types need to be `unsafe fn`, and since they can probably be null, need to be behind an `Option<...>`, so 
```suggestion
        pub sa_handler : Option<unsafe extern "C" fn(::c_int) -> !>,
```

src/vxworks/mod.rs
```diff
@@ -480,6 +474,46 @@ cfg_if! {
                     .finish()
             }
         }
+
+        impl PartialEq for sa_u_t {
```
gnzlbg: I don't think the traits make sense for the union. You don't know which field is "used" so one can't really do anything meaningful for it. 

What one needs is manual implementations for `pub struct sigaction`, which just treat the union as an `usize` for the purpose of the traits.

src/vxworks/mod.rs
```diff
@@ -480,6 +474,46 @@ cfg_if! {
...
+                    .finish()
+            }
+        }
+        impl ::hash::Hash for sa_u_t {
+            fn hash<H: ::hash::Hasher>(&self, state: &mut H) {
+                unsafe { self.sa_handler.hash(state) };
+            }
+        }
+
+        impl PartialEq for sigval {
```
gnzlbg: Same for sigval - the trait impls don't make sense for these.

src/vxworks/mod.rs
```diff
@@ -431,6 +415,16 @@ s_no_extra_traits! {
         pub __ss_pad2  : [::c_char; _SS_PAD2SIZE],
     }
 
+    pub union sa_u_t {
```
gnzlbg: The vxworks/mod.rs should be guarded in the lib/mod.rs file with a `libc_union` gate - otherwise the older rust build jobs will try to compile these and fail.

src/vxworks/mod.rs
n-salim: >  don't think the traits make sense for the union

True, but by implementing them, we avoid "infecting" the structs that use the union. It seems wasteful to have to manually derive these traits for those structs just because they happen to use this type.

Effectively we *are* treating the union as a usize for the purpose of implementing these traits.

src/vxworks/mod.rs
BaoshanPang: But then the type would be undefined, so the older rust build jobs would still fail?

src/vxworks/mod.rs
n-salim: However if what you suggested is the "standard" approach to using unions, then of course we should follow it for consistency.

gnzlbg: If these are only used for type punning and accessing both fields is always valid then I think this is probably ok, just please make it explicit by casting to `as usize` in the trait implementations. For the `Debug` impls it does not make much sense to print both variants, I think it is better to just print the contents as an `usize` instead.

Otherwise, I'm not sure if these are really totally-ordered (e.g. if the `Eq` impl is correct).

src/vxworks/mod.rs
gnzlbg: Ah sorry, my bad! We only try to compile these targets with nightly in the build.rs! The older toolchains don't know these targets at all, so that would always fail.

src/vxworks/mod.rs
gnzlbg: > However if what you suggested is the "standard" approach to using unions, then of course we should follow it for consistency.

The problem with unions is that there is no standard approach: one needs to look at each type and figure out what makes sense for the type. 

Sometimes, the types containing the unions have a field that's used as a discriminant, to denote which field of the union is "active" (Rust does not have such a thing, but C does). So when that happens, one needs to derive the impls manually for those types. For example, for `PartialEq`, only compare the contents of the union fields if the discriminant of both types are the same. For `Debug` only print the active union field according to the discriminant, etc. 

In this case, it appears that the types containing the unions do not have any discriminants, but one should triple check. So if this is right, implementing the traits like an `usize` assumption seems ok.

BaoshanPang: I have updated the code, please let me know if my understanding is wrong.

gnzlbg:
cc @pietroalbini if one r+ bors in a review comment, bors does not start ?

pietroalbini:
I think that's a know issue, there might be an issue on the homu repo already on that. If there isn't feel free to open one.

BaoshanPang:
@gnzlbg 
@n-salim 

I am seeing this error:
error[E0412]: cannot find type `Option` in this scope
   --> /folk/prj-rust-dev/bpang/bugs/dev/.cargo/git/checkouts/libc-4ba4dd73077262d7/b618fdc/src/vxworks/mod.rs:474:34
    |
474 |                 pub sa_handler : Option<unsafe extern "C" fn(::c_int) -> !>,
    |                                  ^^^^^^ not found in this scope
help: possible candidates are found in other modules, you can import them into scope
    |

The 'Option' hase been used in this file, but it has never been used in union before, I also searched it and pretty sure it is the first time we use the 'Option' in union in libc. Is there something missing that cause the problem?

n-salim:
Option is provided by std. Is libc allowed to rely on std? I'm not clear on the dependency order, but I suspect not.

gnzlbg:
libc requires libcore. When building in tree, you need to enable the `rustc-dep-of-std` feature, could you check if this is being done in your build?

BaoshanPang:
The feature is enabled in the command line:

LD_LIBRARY_PATH="/folk/prj-rust-dev/bpang/bugs/dev/rust/build/x86_64-unknown-linux-gnu/stage1/lib:/folk/prj-rust-dev/bpang/bugs/de\
v/rust/build/x86_64-unknown-linux-gnu/stage1-std/release/deps:/folk/prj-llvm-dev/rust/cached-bootstrap/2019-08-13/beta/lib:/folk/p\
rj-rust-dev/bpang/bugs/vxworks/helix/guests/vxworks-7/host/x86_64-linux/lib:/folk/prj-rust-dev/bpang/bugs/vxworks/helix/guests/vxw\
orks-7/host/x86-linux2/lib:/folk/prj-rust-dev/bpang/bugs/vxworks/workbench-4/wrsysviewer/host/x86_64-linux2/lib:/folk/prj-rust-dev\
/bpang/bugs/vxworks/workbench-4/x86_64-linux2/lib:/folk/prj-rust-dev/bpang/bugs/vxworks/license/lmapi-5/x86-linux2/lib" "/folk/prj\
-rust-dev/bpang/bugs/dev/rust/build/x86_64-unknown-linux-gnu/stage1/bin/rustc" "--crate-name" "libc" "/folk/prj-rust-dev/bpang/bug\
s/dev/.cargo/git/checkouts/libc-2d0b80a096c2fb1e/16ddbc2/src/lib.rs"  "--crate-type" "lib" "--emit=dep-info,metadata,link" "-C" "o\
pt-level=2" "--cfg" "feature=\"align\"" **"--cfg" "feature=\"rustc-dep-of-std\""** "--cfg" "feature=\"rustc-std-workspace-core\"" "-C"\
 "metadata=256da7e9aa559704-rustc" "-C" "extra-filename=-256da7e9aa559704" "--out-dir" "/folk/prj-rust-dev/bpang/bugs/dev/rust/bui\
ld/x86_64-unknown-linux-gnu/stage1-std/x86_64-wrs-vxworks/release/deps" "--target" "x86_64-wrs-vxworks" "-L" "dependency=/folk/prj\
-rust-dev/bpang/bugs/dev/rust/build/x86_64-unknown-linux-gnu/stage1-std/x86_64-wrs-vxworks/release/deps" "-L" "dependency=/folk/pr\
j-rust-dev/bpang/bugs/dev/rust/build/x86_64-unknown-linux-gnu/stage1-std/release/deps" "--extern" "rustc_std_workspace_core=/folk/\
prj-rust-dev/bpang/bugs/dev/rust/build/x86_64-unknown-linux-gnu/stage1-std/x86_64-wrs-vxworks/release/deps/librustc_std_workspace_\
core-f544e248473e4569.rmeta" "--cap-lints" "allow" "-Zbinary-dep-depinfo" "--cfg" "freebsd11" "--cfg" "libc_priv_mod_use" "--cfg" \
"libc_union" "--cfg" "libc_const_size_of" "--cfg" "libc_align" "--cfg" "libc_core_cvoid" "--cfg" "libc_packedN" "--cfg" "libc_thre\
ad_local" "-Cdebuginfo=0" "-Wrust_2018_idioms" "-Wunused_lifetimes" "-Dwarnings" "--sysroot" "/folk/prj-rust-dev/bpang/bugs/dev/ru\
st/build/x86_64-unknown-linux-gnu/stage1" "-Zexternal-macro-backtrace" "-Cprefer-dynamic" "-Clinker=wr-c++" "-C" "debug-assertions\
=n" "-C" "link-args=-Wl,-rpath,$ORIGIN/../lib" "-Z" "force-unstable-if-unmarked"

BaoshanPang:
Adding use core::option::Option; at the head of file src/vxworks/mod.rs would make the error disappear, but I don't understand why we don't need to add it when "Option" is used in another places.

n-salim:
If you add` -Zunstable-options --pretty=expanded ` to the rustc line it will show the code after macro expansion. Does that provide any insight?

I also noticed that in other places `Option` appears as `::Option`. Does that make a difference? (I have to admit I don't know why that would help). The [documentation](https://doc.rust-lang.org/stable/reference/paths.html) says 

> Paths starting with :: are considered to be global paths where the segments of the path start being resolved from the crate root. Each identifier in the path must resolve to an item.
> 
>     Edition Differences: In the 2015 Edition, the crate root contains a variety of different items, including external crates, default crates such as std and core, and items in the top level of the crate (including use imports).
> 
>     Beginning with the 2018 Edition, paths starting with :: can only reference crates.
>

BaoshanPang:
>>I also noticed that in other places Option appears as ::Option.

Great catch. we should use ::Option.

BaoshanPang:
Created PR https://github.com/rust-lang/libc/pull/1532 to address this issue and update rtpSpawn prototype.

gnzlbg:
See https://github.com/rust-lang/libc/blob/70d1f16e4ad4f5b51c77876191553fabd3b04253/src/lib.rs#L41 and the cfg if below.

When building with libcore or libstd, `Option` is in the prelude, so it is not necessary to import it, and will "just work" at least, here on CI. When building rust-lang/rust, `libc` is built before `libcore` is built, so the library is built with `no_core`, and there is no prelude, and therefore no `Option` in the prelude. We import a different `Option` type from a different crate (the `rust-std-workspace-core` crate which is like a hacky mini-libcore) and put it in the crate root. The way to access the crate root is to do a`::X`, so `::Option` uses that option. To make that work when libcore is available, we bring its `Option` type into the crate root as well (that's what the cfg_if below that one does).

So.. yes, one needs to use `::Option` for the rust-lang/rust build to work.

n-salim:
@gnzlbg Thank you for the detailed explanation! It makes sense now.

PR #1531, opened 2019-09-25
Fix max_align_t

gnzlbg:


gnzlbg:
@BaoshanPang so this fix is now on master, please do check that it fixes your build issues - sadly we have no easy way to test this from the libc repo, these things are usually tested when updating libc on rust-lang/rust

BaoshanPang:
@gnzlbg 
Thanks for the quick fix for this issue. I verified the building issue has been fixed in the master branch.

PR #1537, opened 2019-10-04
Haiku: Fix constants for S_IFIFO and S_IFCHR

nielx:


nielx:
The check fails because a download on the build server was too slow, not because of the change. It is very unlikely that this change will cause a build failure in any case (barring syntax errors), since the build servers will never pick up this code path.

PR #1539, opened 2019-10-06
Add extended IFLA_ consts

gluxon:
Only Netlink interface link attributes up `IFLA_STATS` are currently available. This commit adds `IFLA_COST` to `IFLA_MAX_MTU`.

See https://github.com/torvalds/linux/blob/54ecb8f7/include/uapi/linux/if_link.h#L106

Thanks!


gluxon:
@bors Could I get another r+? I removed some extra values that got introduced after the Linux kernel versions used on CI. That got the Linux tests passing.

There's some failures on wasm32, asmjs, and windows that look like build flakes.

PR #1548, opened 2019-10-13
Add settimeofday for openbsd

### LINKED ISSUE #1546: OpenBSD does not expose settimeofday
OpenBSD have a [man page](https://man.openbsd.org/settimeofday.2), describing both get and set version, but currently only the get one is exposed

x86_64-unknown-openbsd (but probably all OpenBSD target triple

GrayJack:
Closes #1546


gnzlbg:
cc @semarie

semarie:
Just wonder about the use case. But from code point of vue, it should be fine.

PR #1549, opened 2019-10-13
Add settimeofday for solaris/illumos

### LINKED ISSUE #1547: Solaris does not expose settimeofday
Solaris/Illumos have a [man page](https://illumos.org/man/3c/settimeofday), describing both get and set version, but currently only the get one is exposed

x86_64-sun-solaris (but probably all Solaris target triple)

GrayJack:
Closes #1547

### DIFF
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -1889,6 +1889,7 @@ extern "C" {
     pub fn srand(seed: ::c_uint);
 
     pub fn gettimeofday(tp: *mut ::timeval, tz: *mut ::c_void) -> ::c_int;
+    pub fn settimeofday(tp: *const ::timeval, tz: *const ::c_void) -> ::c_int;
     pub fn getifaddrs(ifap: *mut *mut ::ifaddrs) -> ::c_int;
     pub fn freeifaddrs(ifa: *mut ::ifaddrs);
 


gnzlbg:
@mati865 all musl targets are broken on nightly, is this known ?

gnzlbg:
So the PR that broke this has been reverted and the next nightly shouldn't have this issue.

mati865:
Yeah all targets packaging start objects are broken. 
Revert PR has failed multiple times (very unlucky) https://github.com/rust-lang/rust/pull/65342 so it will be fixed tomorrow.

GrayJack:
Is it fixed now?

gnzlbg:
@GrayJack thanks for the ping, yes, it was fixed a couple of hours ago. 

@bors: r+

gnzlbg:
cc @asomers @kinnison - so apparently rustup fails on FreeBSD11 because of some components not being available.

mati865:
^ opened https://github.com/rust-lang/libc/pull/1550

kinnison:
> EDIT: @kinnison this looks like a rustup bug, where it tries to install components that are not available for a target. Can it be that the new profiles do not know that certain targets never have some components ? (I'm not sure if FreeBSD11 sometimes has clippy and rustfmt).

I'm aware of something like this on lower tier platforms yes.  I've a fix in-train which might go out tomorrow for that part.

asomers:
> cc @asomers @kinnison - so apparently rustup fails on FreeBSD11 because of some components not being available.
> 
> EDIT: @kinnison this looks like a rustup bug, where it tries to install components that are not available for a target. Can it be that the new profiles do not know that certain targets never have some components ? (I'm not sure if FreeBSD11 sometimes has clippy and rustfmt).

FreeBSD11 is supposed to have clippy and rustfmt, as should most platforms.  But it hasn't since late May due to this bug: https://github.com/rust-lang/rust/issues/62558 .

GrayJack:
Doesn't seem to have something to do with my changes, only linux-android failed

PR #1550, opened 2019-10-15
Use minimal profile for rustup

mati865:
Haven't tested it but should fix issue from https://github.com/rust-lang/libc/pull/1549#issuecomment-542203074


mati865:
So it should be good to go, Azure image still have outdated rustup but once they upgrade it you can just uncomment lines I just added. That will make prepare step faster (less downloading and installing) and fix error if it also shows there.

PR #1554, opened 2019-10-16
Refine definition of mcontext_t on x86_64-unknown-linux-musl

### LINKED ISSUE #1553: Export mcontext fields on x86_64-unknown-linux-musl
[lucet](https://github.com/fastly/lucet) uses some of the fields of `mcontext` that were added recently for `x86_64-unknown-linux-gnu`. It'd be great to port those to `x86_64-unknown-linux-musl`. In particular:

https://github.com/rust-lang/libc/blob/7b31ef34ede6860ce2a4cfeeef01b02f8120a607/src/unix/linux_like/linux/gnu/b64/x86_64/mod.rs#L221-L225

and 

https://github.com/rust-lang/libc/blob/7b31ef34ede6860ce2a4cfeeef01b02f8120a607/src/unix/linux_like/linux/gnu/b64/x86_64/mod.rs#L871

which were ported in https://github.com/rust-lang/libc/commit/ae06c2b2ae6938c9162b9deadae442194a816ad2 and https://github.com/rust-lang/libc/commit/ac7f0fac2f34d4eed1f05e9d20f236149711ff87. Lucet only need `mcontext.gregs` and `REG_RIP` to be defined (see [here](https://github.com/fastly/lucet/blob/6af51b4235b7c5e09ca30d9bea7b68f4f70483cb/lucet-runtime/lucet-runtime-internals/src/sysdeps/linux.rs#L16)).

These are defined in musl with `GNU_SOURCE`:
* https://github.com/ifduyue/musl/blob/b4b1e10364c8737a632be61582e05a8d3acf5690/arch/x86_64/bits/signal.h#L42
* https://github.com/ifduyue/musl/blob/b4b1e10364c8737a632be61582e05a8d3acf5690/arch/x86_64/bits/signal.h#L81

nmattia:
Fixes #1553 

I've inlined the links I used for defining constants and fields, although they pretty much map to the changes introduced in ac7f0fac2f34d4eed1f05e9d20f236149711ff87 and ae06c2b2ae6938c9162b9deadae442194a816ad2.

### DIFF
--- a/src/unix/linux_like/linux/musl/b64/x86_64/mod.rs
+++ b/src/unix/linux_like/linux/musl/b64/x86_64/mod.rs
@@ -100,7 +100,9 @@ s! {
         pub u_debugreg: [::c_ulong; 8],
     }
 
-    // https://github.com/ifduyue/musl/blob/b4b1e10364c8737a632be61582e05a8d3acf5690/arch/x86_64/bits/signal.h#L80-L84
+    // GitHub repo: ifduyue/musl/
+    // commit: b4b1e10364c8737a632be61582e05a8d3acf5690
+    // file: arch/x86_64/bits/signal.h#L80-L84
     pub struct mcontext_t {
         pub gregs: [greg_t; 23],
         __private: [u64; 9],
@@ -607,7 +609,9 @@ pub const FS: ::c_int = 25;
 pub const GS: ::c_int = 26;
 
 // offsets in mcontext_t.gregs from bits/signal.h
-// https://github.com/ifduyue/musl/blob/b4b1e10364c8737a632be61582e05a8d3acf5690/arch/x86_64/bits/signal.h#L9-L56
+// GitHub repo: ifduyue/musl/
+// commit: b4b1e10364c8737a632be61582e05a8d3acf5690
+// file: arch/x86_64/bits/signal.h#L9-L56
 pub const REG_R8: ::c_int = 0;
 pub const REG_R9: ::c_int = 1;
 pub const REG_R10: ::c_int = 2;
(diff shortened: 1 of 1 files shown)


gnzlbg:
@bors: r+

Thanks!

nmattia:
@gnzlbg thanks for the quick response! When should we expect a release on crates.io? This would help us tremendously!

gnzlbg:
I personally would prefer to do so after https://github.com/rust-lang/rust/pull/65469 is merged. That PR is green but apparently blocked by https://github.com/rust-lang/rust/pull/65251 which is being merged right now. The risk of producing new releases of libc in the meantime is that we might have to yank them if upgrading rust discovers newer issues.

I'd say let's wait 24h and see if the situation resolves by itself, which it often does, and if it doesn't, then please ping me again here and I'll re-evaluate whether we maybe just do a new release to unblock work on your side or not. If it resolves by itself and I do a new release I will ping you here instead.

In the meantime, please use a git dependency on libc master if you want to already use this. The cargo book explains how to `patch` dependencies if you don't want to modify your Cargo.toml.

nmattia:
here's the ping @gnzlbg ! https://github.com/rust-lang/rust/pull/65251 is in and it looks like https://github.com/rust-lang/rust/pull/65469 is ready as well, though bors didn't perform the merge.

Unfortunately we can't `patch`, as the dependency is then treated as a local crate and we treat warnings as errors.

gnzlbg:
Thanks for the ping, yes everything looks green on their side, so I've released 0.2.65. If something unlikely were to fail on their side, we can always release 0.2.66 with a fix.

nmattia:
Thanks a bunch!

PR #1555, opened 2019-10-16
Add struct ip_mreqn: ip multicast req with intf id

flocknetworks:
The ip_mreqn struct has an additional interface id parameter. This
allows programming IPv4 multicast groups using the interface id
alongside the src IPv4 address.
Linux has supported the ip_mreqn struct since Linux 2.2
ref: man ip.7

### DIFF
--- a/src/unix/linux_like/linux/gnu/b64/x86_64/mod.rs
+++ b/src/unix/linux_like/linux/gnu/b64/x86_64/mod.rs
@@ -261,6 +261,12 @@ s! {
         pub c_ispeed: ::speed_t,
         pub c_ospeed: ::speed_t,
     }
+
+    pub struct ip_mreqn {
+        pub imr_multiaddr: ::in_addr,
+        pub imr_address: ::in_addr,
+        pub imr_ifindex: ::c_int,
+    }
 }
 
 s_no_extra_traits! {


flocknetworks:
rust-lang.libc spar64 + mips64 tests failed but I cant see anything in the logs specific to this diff set ?

gnzlbg:
There are sparc and mips linux targets, and the ABI of the type you added does not match the C ABI on those targets. Why, I don't know. Maybe the struct definition that you have added does not match the struct definition on those targets ?

src/unix/linux_like/mod.rs
```diff
@@ -24,6 +24,7 @@ s! {
         pub imr_interface: in_addr,
     }
 
+    #[cfg(not(any(target_arch = "mips64", target_arch = "sparc64")))]
     pub struct ip_mreqn {
```
gnzlbg: This won't work. If you don't want to add this to all linux targets, you need to manually add this to the targets you want. I do however recommend finding the source code for this and checking out what mips and sparc do.

flocknetworks:
in glibc I cant find any ip_mreqn struct specific to an architecture.  I assume this struct is just not in the sparc64 / mips64 builds ?

gnzlbg:
Which header file does expose it?

flocknetworks:
struct ip_mreq is defined in inet/netinet/in.h
struct ip_mreqn is defined in sysdeps/unix/sysv/linux/bits/in.h

gnzlbg:
So that's in this header: https://elixir.bootlin.com/linux/v5.4-rc2/source/include/uapi/linux/in.h#L173 

> I assume this struct is just not in the sparc64 / mips64 builds ?

It is available - we know because the C tests for it on sparc64 and mips64 use the type and they compile successfully, so a type definition must be available.

gnzlbg:
So looking at the error, the test that's failing is the following. 

This struct has 3 fields, where each field is 4 bytes wide. The first two fields are a `struct { u32 }`, while the last fields is an `i32`. 

The test initializes the struct bytes in Rust to the following values: [0, 1, 2, 3, ..., 11]. It then passes the struct by value to a C function expecting it. This function then reads the bytes of the struct and verifies that they match the values set on the Rust side. This fails, reporting that the last 4 bytes (the ones of the `i32`) are read as `[0, 0, 0, 0]` instead if `[8, 9, 10, 11]`. 

That's weird. In particular, this is apparently the only test that fails there. That is, the struct sizes must match, so there can't be any padding between the `in_addr` and the `c_int` (that would explain the zeroes in the C side).

Maybe you can just modify the `test_linux` function in the `libc-test/build.rs` file to skip the `roundtrip` test for this type ? (there is a cfg.skip_roundtrip there where you can specify that). It would be interesting to know if all other tests pass.

flocknetworks:
thanks for looking into this, weird indeed. I will try modifying the build.rs

gnzlbg:
If you need these now and don't have time to investigate that's ok, just add these structs to the platforms you need them, e.g., instead of to linux/, maybe only linux/gnu/x86_64 - copy pasting them isn't a problem. 

What we try to avoid is adding types that might be subtly broken on some platforms. In this case, it would cost a user quite a bit of time to figure out why C is crashing when passing it these types without actually writing some code that tries to print their value in C, because on the Rust side, everything looks fine.

flocknetworks:
all the other tests pass (when the roundtrip test is skipped)

gnzlbg:
> all the other tests pass (when the roundtrip test is skipped)

Damn, so I don't know what might be going on. Maybe our ABI implementation on mips64 and sparc64 is borked somehow? cc @nagisa @nikic 

I'm going to open an issue to track investigating what's going on here. In the meantime, we can land this for x86_64, thanks @flocknetworks 

@bors: r+

flocknetworks:
I'm assuming these failures are due to something in the CI infra rather than my diffs ?

PR #1556, opened 2019-10-16
Sort targets in documentation. Fixes #1524.

### LINKED ISSUE #1524: Sort targets in the docs
The current list of targets in the auto-generated docs is unsorted.

senden9:
Should be a simple but effective fix for #1524. Assume `sort` command is available.

### DIFF
--- a/ci/dox.sh
+++ b/ci/dox.sh
@@ -25,7 +25,7 @@ cargo +nightly install cargo-xbuild -Z install-upgrade
 # shellcheck disable=SC1003
 grep '[\d|\w|-]* \\' ci/build.sh > targets
 sed -i.bak 's/ \\//g' targets
-grep '^[_a-zA-Z0-9-]*$' targets > tmp && mv tmp targets
+grep '^[_a-zA-Z0-9-]*$' targets | sort > tmp && mv tmp targets
 
 # Create a markdown list of supported platforms in $PLATFORM_SUPPORT
 rm $PLATFORM_SUPPORT || true


gnzlbg:
Thanks, this is exactly what I had in mind.

senden9:
@gnzlbg I also don't like unsorted random lists.

But the list at https://rust-lang.github.io/libc/#platform-specific-documentation seems not quite sorted. Does it take some time till the new documentation get pushed?
Because as far as i understand this list is generated form the master Branch. My PR is now merged into master. So this list should be now sorted or?

gnzlbg:
It should just work, but apparently it stopped working 2 months ago: https://github.com/rust-lang/libc/tree/gh-pages (last time that was updated). 

The job for this is indeed running: https://github.com/rust-lang/libc/blob/master/ci/azure.yml#L132 

But apparently the deploy to github pages is being skipped: https://dev.azure.com/rust-lang2/libc/_build/results?buildId=501

@pietroalbini do you have any idea of what might be going on ?

gnzlbg:
> But apparently the deploy to github pages is being skipped: https://dev.azure.com/rust-lang2/libc/_build/results?buildId=501

Ok, so the problem is that these run in the `auto-libc` branch, but the deployment only happens for `master`.

gnzlbg:
I've added a new pipeline that should build master: https://dev.azure.com/rust-lang2/libc/_build?definitionId=7

gnzlbg:
@senden9 thanks, look: https://rust-lang.github.io/libc/#platform-specific-documentation now everything shows up sorted!

senden9:
Nice :D.

I see `mipsel-unknown-linux-gnu` is two times in the list. See screenshot.
image
Should I fix that i another PR by checking for doubles in the list or do you want to fix that problem on another level?

gnzlbg:
#1557 should fix it

senden9:
Good. Found just in that moment the same lines that you edited.

PR #1561, opened 2019-10-18
Add flock64 to linux_like platforms

elichai:
Hi,
I'm not sure about what to make of the ifdef in glibc `__USE_LARGEFILE64` that covers it.
but I copied the impl from glibc's headers.
everything is the same except for sparc which has a reserved short int.
and in musl it's just `#define flock64 flock`

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2230,6 +2230,9 @@ fn test_linux(target: &str) {
 
             t if t.ends_with("_t") => t.to_string(),
 
+            // This is either a struct or a typedef to a struct
+            "flock64" if musl => format!("struct {}", ty),
+
             // put `struct` in front of all structs:.
             t if is_struct => format!("struct {}", t),
 
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2230,7 +2230,7 @@ fn test_linux(target: &str) {
 
             t if t.ends_with("_t") => t.to_string(),
 
-            // This is either a struct or a typedef to a struct
+            // In MUSL `flock64` is a typedef to `flock`.
             "flock64" if musl => format!("struct {}", ty),
 
             // put `struct` in front of all structs:.
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -22,6 +22,8 @@ pub type ino_t = ::c_ulong;
 pub type __CPU_BITTYPE = ::c_ulong;
 pub type idtype_t = ::c_int;
 pub type loff_t = ::c_longlong;
+pub type __kernel_loff_t = ::c_longlong;
+pub type __kernel_pid_t = ::c_int;
 
 s! {
     pub struct stack_t {
@@ -76,6 +78,14 @@ s! {
         pub l_start: ::off_t,
         pub l_len: ::off_t,
         pub l_pid: ::pid_t,
+    }
+
+    pub struct flock64 {
+        pub l_type: ::c_short,
+        pub l_whence: ::c_short,
+        pub l_start: ::__kernel_loff_t,
+        pub l_len: ::__kernel_loff_t,
+        pub l_pid: ::__kernel_pid_t,
     }
 
     pub struct cpu_set_t {
--- a/src/unix/linux_like/linux/gnu/b64/sparc64/mod.rs
+++ b/src/unix/linux_like/linux/gnu/b64/sparc64/mod.rs
@@ -60,6 +60,15 @@ s! {
         pub l_pid: ::pid_t,
     }
 
+    pub struct flock64 {
+        pub l_type: ::c_short,
+        pub l_whence: ::c_short,
+        pub l_start: ::off64_t,
+        pub l_len: ::off64_t,
+        pub l_pid: ::pid_t,
+        __reserved:  ::c_short,
+    }
+
     pub struct stack_t {
         pub ss_sp: *mut ::c_void,
         pub ss_flags: ::c_int,
--- a/src/unix/linux_like/emscripten/mod.rs
+++ b/src/unix/linux_like/emscripten/mod.rs
@@ -223,6 +223,14 @@ s! {
         pub l_pid: ::pid_t,
     }
 
+    pub struct flock64 {
+        pub l_type: ::c_short,
+        pub l_whence: ::c_short,
+        pub l_start: ::off64_t,
+        pub l_len: ::off64_t,
+        pub l_pid: ::pid_t,
+    }
+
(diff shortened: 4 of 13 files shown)


src/unix/linux_like/linux/gnu/b64/sparc64/mod.rs
```diff
@@ -60,6 +60,14 @@ s! {
         pub l_pid: ::pid_t,
     }
 
+    pub struct flock64 {
+        pub l_type: ::c_short,
+        pub l_whence: ::c_short,
+        pub l_start: ::off64_t,
+        pub l_len: ::off64_t,
+        pub l_pid: ::pid_t,
+    }
```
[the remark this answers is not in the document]
elichai: Apparently I fixed it locally already but forgot to push. good catch anyway.
fixed now.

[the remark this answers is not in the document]
glaubitz: Thanks!

glaubitz:
If this gets merged before my sparc PR, I should probably add flock64 to sparc as well.

gnzlbg:
From the error messages it appears that the type is defined differently on different targets. You need to fix those.

elichai:
> From the error messages it appears that the type is defined differently on different targets. You need to fix those.

First of all awesome test suite :) I need to read how it works.

Second, seems that the problem is the musl targets, which makes sense. i'll fix it.

gnzlbg:
Notice that there are generic files for `linux/gnu/mod.rs` and `linux/musl/mod.rs`, so you might get away with just having two structs (I'm not sure).

elichai:
> Notice that there are generic files for `linux/gnu/mod.rs` and `linux/musl/mod.rs`, so you might get away with just having two structs (I'm not sure).

the glibc once are fine (they match glibc correctly and pass the tests) on the other hand, musl have 
`#define flock64 flock` and even that it's under `#if defined(_LARGEFILE64_SOURCE) || defined(_GNU_SOURCE)` so maybe there's just no `flock64` in the test framework, although the errors are weird. I'm trying to reproduce them locally

gnzlbg:
The test framework always defines `_GNU_SOURCE` for all linux targets: https://github.com/rust-lang/libc/blob/master/libc-test/build.rs#L2074

glaubitz:
Be careful though as flock64 is different on sparc* targets so you need arch-specific definitions.

elichai:
> Be careful though as flock64 is different on sparc* targets so you need arch-specific definitions.

I know, did that already, the weird thing is actually musl targets which should be flock==flock64

elichai:
@gnzlbg Ok, I got it down to this example:  https://ideone.com/rgk7OZ
the problem seems to be line 22:
```
  typedef struct {
    unsigned char c;
    flock v;
  } type;
```

it should be `struct flock v;`

The question is why this happens

elichai:
Update, that's definitely because I defined `pub type flock64 = ::flock;`. it seems like the translation unit to C doesn't understand it properly. 
i'll try to reproduce on `ctest` and open an issue there.

Should I replace the typedef with a duplicate struct?

gnzlbg:
There are methods in ctest to configure how to name the type aliases, you can use `cfg.type_name` to tell ctest to expand flock64 on that particular target to `struct flock64` - there should be many examples of this in the libc-test/build.rs.

libc-test/build.rs
```diff
@@ -2230,6 +2230,9 @@ fn test_linux(target: &str) {
 
             t if t.ends_with("_t") => t.to_string(),
 
+            // This is either a struct or a typedef to a struct
+            "flock64" => format!("struct {}", ty),
```
gnzlbg: You want to add an `if musl && ... =>` to do this only for those targets for which this is required.

elichai: even in gnu it's a struct, but I can ass the if if you prefer.

gnzlbg: For the `gnu` targets it is defined as a `struct`, not as a `type` alias in Rust though.

gnzlbg: This is only needed on `musl` because on musl, it is a macro that expands to the name of a struct, and we use a type alias, which we need to tell ctest to expand as a struct _on those targets only_.

elichai: Well the end result is the same.
On gnu the is_struct will be true and will resolve to the same string (`struct NAME`)

src/unix/linux_like/android/mod.rs
```diff
@@ -78,6 +78,14 @@ s! {
         pub l_pid: ::pid_t,
     }
 
+    pub struct flock64 {
+        pub l_type: ::c_short,
+        pub l_whence: ::c_short,
+        pub l_start: ::off64_t,
+        pub l_len: ::off64_t,
+        pub l_pid: ::pid_t,
+    }
```
gnzlbg: `l_start` and `l_end` have the wrong types on some android targets.

gnzlbg: Should be `__kernel_loff_t` and not `off64_t`.

elichai: You're right, I looked at glibc here instead of google's code. https://android.googlesource.com/platform/bionic/+/refs/heads/master/libc/kernel/uapi/asm-generic/fcntl.h#161

elichai: BTW, not really relevant but it seems like the sizes for `loff_t` etc. can be different per 32/64 bits depends on if the macro is enabled, so i'm not sure what to make of it 
https://android.googlesource.com/platform/bionic/+/refs/heads/master/libc/include/sys/types.h#100
https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md

gnzlbg: You can just define `loff_t` differently for the different android targets.

libc-test/build.rs
gnzlbg: The targets for which this must be enabled are still missing here. Otherwise this LGTM.

elichai: Done

libc-test/build.rs
```diff
@@ -2230,6 +2230,9 @@ fn test_linux(target: &str) {
 
             t if t.ends_with("_t") => t.to_string(),
 
+            // This is either a struct or a typedef to a struct
```
gnzlbg:
```suggestion
            // In MUSL `flock64` is a typedef to `flock`.
```

gnzlbg:
Thanks! @bors: r+

glaubitz:
Oh, is the CI working again? I just re-pushed my PR.

PR #1563, opened 2019-10-20
add getlogin_r, re-entrant getlogin

FauxFaux:
`getlogin_r()`: `_POSIX_C_SOURCE >= 199506L`, according to `glibc`, so I believe it will be widely deployed. Let's see what CI says!

Prototype copied from `gethostname`, which appears identical; see (closed/abandoned) https://github.com/nix-rust/nix/pull/789

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -837,6 +837,7 @@ extern "C" {
     pub fn getgid() -> gid_t;
     pub fn getgroups(ngroups_max: ::c_int, groups: *mut gid_t) -> ::c_int;
     pub fn getlogin() -> *mut c_char;
+    pub fn getlogin_r(name: *mut ::c_char, len: ::size_t) -> ::c_int;
     #[cfg_attr(
         all(target_os = "macos", target_arch = "x86"),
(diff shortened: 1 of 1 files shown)


FauxFaux:
Apparently not.

I'm going to abandon this, because `getlogin_r` is probably worse than `getpwuid_r`, which is already exposed, and just use that.

May this be a lesson to any future digger.

> BUGS
>       Unfortunately, it is often rather easy to fool getlogin().  Sometimes it does not work at all, because some program messed up the utmp file.  Often, it gives only the first 8 characters of the login name.  The user currently logged in on the controlling terminal of our program need not be the user who started it.  Avoid getlogin() for security-related purposes.
>       Note that glibc does not follow the POSIX specification and uses stdin instead of /dev/tty.  A bug.  (Other recent systems, like SunOS 5.8 and HP-UX 11.11 and FreeBSD 4.8 all return the login name also when stdin is redirected.)
>       Nobody knows precisely what cuserid() does; avoid it in portable programs.  Or avoid it altogether: use getpwuid(geteuid()) instead, if that is what you meant.  Do not use cuserid().

PR #1564, opened 2019-10-22
ci: switch mirrors to use our CDN

pietroalbini:
We recently added a CDN in front of our CI mirrors as it's faster and
cheaper for us. This switches libc's CI to use it instead of accessing
the underlying bucket directly.

r? @gnzlbg


gnzlbg:
@bors: r+ 

Thanks!

pietroalbini:
Seems spurious?

gnzlbg:
🤷‍♂ 

@bors: retry

pietroalbini:
This can probably be retried, it shouldn't cause timeouts.

PR #1566, opened 2019-10-23
Add a few functions & POSIX error codes to Windows API

mikehoyle:
This change adds a few simple additional functions to the windows apis, as well as some additional error codes.

Tests and style pass, as per contributing guidelines.

### DIFF
--- a/src/windows/mod.rs
+++ b/src/windows/mod.rs
@@ -183,6 +183,49 @@ pub const ENOTEMPTY: ::c_int = 41;
 pub const EILSEQ: ::c_int = 42;
 pub const STRUNCATE: ::c_int = 80;
 
+// POSIX Supplement (from errno.h)
+pub const EADDRINUSE: ::c_int = 100;
+pub const EADDRNOTAVAIL: ::c_int = 101;
+pub const EAFNOSUPPORT: ::c_int = 102;
+pub const EALREADY: ::c_int = 103;
+pub const EBADMSG: ::c_int = 104;
+pub const ECANCELED: ::c_int = 105;
+pub const ECONNABORTED: ::c_int = 106;
+pub const ECONNREFUSED: ::c_int = 107;
+pub const ECONNRESET: ::c_int = 108;
+pub const EDESTADDRREQ: ::c_int = 109;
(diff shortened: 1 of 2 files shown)


retep998:
`ucrt\errno.h` does have those error constants, and notably they are *different* than the WSA constants of the same name. For example `EADDRINUSE` from ucrt has a value of 100 but `EADDRINUSE` from winsock2 has a value of 10048. Obviously a massive footgun but it is what it is.

`time` and `_lseeki64` are both from ucrt as well. Notably however `time` is actually an inline wrapper around `_time64`. The signatures look correct otherwise.

UdamSaini:
Inside winsock.h, they say to not use EADDRINUSE and to use WSAEADDRINUSE due to the conflict for socket related errors. 
/*
* Windows Sockets errors redefined as regular Berkeley error constants.
* These are commented out in Windows NT to avoid conflicts with errno.h.
* Use the WSA constants instead.
*/
#if 0
#define EWOULDBLOCK  WSAEWOULDBLOCK
......

I think we should probably use the errno constants as that is what windows uses, and have the WSA constants be separate or in the win32 library? I wonder if we should add a comment to that effect, similar to the windows comment.

mikehoyle:
In my most recent commit I updated the time funtion to link to _time64 (as time is just a wrapper around that anyways). It also seems that the gnu toolchain uses a different (probably outdated) version of errno.h that doesn't contain EOTHER, so I have moved that specific constant to the msvc-specific file.

src/windows/mod.rs
```diff
@@ -183,6 +183,48 @@ pub const ENOTEMPTY: ::c_int = 41;
 pub const EILSEQ: ::c_int = 42;
 pub const STRUNCATE: ::c_int = 80;
 
+// POSIX Supplement (from errno.h)
+pub const EADDRINUSE: ::c_int = 100;
```
gnzlbg: Per the comment below from @UdamSaini , does it make sense to add this one here ? (or maybe we can add it as `#[deprecated]` with the same warning as the windows headers?

retep998: Those warnings are on the winsock definitions that just redefine to the WSA constants, not the errno definitions.

PR #1567, opened 2019-10-24
Add initial support for sparc-unknown-linux-gnu

glaubitz:
While the Rust compiler already supports the sparc-unknown-linux-gnu target, the libc crate is missing support for it. Let's add it so that Rust can built for this target as well.

### DIFF
--- a/src/unix/linux_like/linux/gnu/b32/sparc/mod.rs
+++ b/src/unix/linux_like/linux/gnu/b32/sparc/mod.rs
@@ -1,0 +1,978 @@
+//! SPARC-specific definitions for 32-bit linux-like values
+
+pub type c_char = u8;
+pub type wchar_t = i32;
+
+s! {
+    pub struct sigaction {
+        pub sa_sigaction: ::sighandler_t,
+        pub sa_mask: ::sigset_t,
+        pub sa_flags: ::c_int,
+        pub sa_restorer: ::Option<extern fn()>,
+    }
+
+    pub struct statfs {
+        pub f_type: ::__fsword_t,
+        pub f_bsize: ::__fsword_t,
+        pub f_blocks: ::fsblkcnt_t,
+        pub f_bfree: ::fsblkcnt_t,
+        pub f_bavail: ::fsblkcnt_t,
+
+        pub f_files: ::fsfilcnt_t,
+        pub f_ffree: ::fsfilcnt_t,
+        pub f_fsid: ::fsid_t,
+
+        pub f_namelen: ::__fsword_t,
+        pub f_frsize: ::__fsword_t,
+        f_spare: [::__fsword_t; 5],
+    }
+
+    pub struct siginfo_t {
+        pub si_signo: ::c_int,
+        pub si_errno: ::c_int,
+        pub si_code: ::c_int,
+        _pad: [::c_int; 29],
+        _align: [usize; 0],
+    }
+
+    pub struct flock {
+        pub l_type: ::c_short,
+        pub l_whence: ::c_short,
+        pub l_start: ::off_t,
+        pub l_len: ::off_t,
+        pub l_pid: ::pid_t,
+    }
+
+    pub struct flock64 {
+        pub l_type: ::c_short,
+        pub l_whence: ::c_short,
+        pub l_start: ::off64_t,
+        pub l_len: ::off64_t,
+        pub l_pid: ::pid_t,
+        __reserved:  ::c_short,
+    }
+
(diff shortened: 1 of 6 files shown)


glaubitz:
Can we try to re-run the rust-lang.libc test? The failure seems unrelated.

src/unix/linux_like/linux/gnu/b32/sparc/mod.rs
```diff
@@ -0,0 +1,976 @@
...
+        pub si_errno: ::c_int,
+        pub si_code: ::c_int,
+        #[doc(hidden)]
+        #[deprecated(
+            since="0.2.54",
+            note="Please leave a comment on \
+                  https://github.com/rust-lang/libc/pull/1316 if you're using \
+                  this field"
+        )]
+        pub _pad: [::c_int; 29],
```
gnzlbg: This field should be private, and the `doc(hidden)` and deprecation not necessary since no code is using this.

glaubitz: This struct is verbatim identical to several other architectures in this source, so I would rather commit this as-is and then make the requested changes later for all architectures at the same time.

I already spotted multiple smaller bugs in the sparc64 code which I wanted to fix after this has been merged and I'm happy to add your requested change to the list of fixes.

Would that be okay?

gnzlbg: > This struct is verbatim identical to several other architectures in this source, so I would rather commit this as-is and then make the requested changes later for all architectures at the same time.

Doing this change for other architectures would be a breaking change.

glaubitz: Then I don't understand why it should be changed here now. It would be confusing for anyone reading the code.

gnzlbg: A bug was found that's hard to fix because it's an API breaking change.

If you want to argue that new APIs should also introduce the bug, please explain what is the value and cost of doing that, and while introducing the bug for newer APIs is a trade-off that is worth it.

glaubitz: Well, if you had mentioned that there was a bug in the first place, I would have actually understood the argument.

I'll push something in a minute.

gnzlbg: Sorry, I'm very low on time. 

After this the rest looked good to me. I'll try to merge it ASAP but we are having network reliability problems.

glaubitz: No worries. Pushed. Also fixed the names of two private "reserved" fields to match the actual glibc headers.

glaubitz:
Looks like there are unrelated test failures for some tests. Would a manual merge be possible?

gnzlbg:
> Looks like there are unrelated test failures for some tests

Yes, looks like nightly Rust broke a couple of libc jobs. Need to fix that upstream.

> Would a manual merge be possible?

Sadly no, all PRs are blocked on the build being green, too many crates depend on libc.

glaubitz:
@gnzlbg Any idea when the CI will be working again?

glaubitz:
Just rebased and force-pushed, should be good for merge after CI has finished.

glaubitz:
Failure is unrelated, the CI process for "DockerLinuxTier2 aarch64-unknown-linux-android" failed to install build dependencies due to a time out.

glaubitz:
Same failure: same target and package that failed to download.

gnzlbg:
@bors: retry

Let's try one more time.

glaubitz:
Rebased and pushed. Let's see if the CI is happier today.

glaubitz:
CI failed again for unrelated reasons :(.

glaubitz:
@gnzlbg Can you approve again?

gnzlbg:
Every time the source code is modified I need to review, and with the re-bases it is unclear to me what changed from the previous time I reviewed, so I kind of need to review all of it, which takes time. It would be simpler if new commits were independent commits.

glaubitz:
@gnzlbg I'm just rebasing every time and then force-pushing to retrigger the testsuite. No changes.

gnzlbg:
Ah, please don't do that. That re-triggers the test suite here, blocking the queue and preventing from bors from trying to merge it, and requires re-reviewing the PR instead of just retrying it. 

It's better just to ping me :)

PR #1577, opened 2019-10-31
Provide SYS_statx for more target_arch/target_env

### LINKED ISSUE #1545: `statx` should be available on more platform/architecture
Including `struct statx`, `fn statx`, `const SYS_statx`, sub-structs and related flags used by them.

Currently, `SYS_statx` is missing on `gnu` with arch `mips`, `aarch64`, `mips64`, `s390x`.
And all of them are missing on `musl`, while it seems to support them already.

Syscall id is only architecture related. Maybe be can simply get them from [here](http://git.etalabs.net/cgit/musl/commit/?id=9864f60e929100e253fc813bd4105d6dd7652787) ?

oxalica:
Fixes #1545 

The syscall `statx` is provided by Linux Kernel, so it should be available on any `target_env`/`target_arch`, not only some arch on `gnu`.

Syscall ids are got from [this commit of musl](http://git.etalabs.net/cgit/musl/commit/?id=9864f60e929100e253fc813bd4105d6dd7652787), except for hexagon, which is got from Linux Kernel (`make O=build ARCH=hexagon headers`).

**Not tested yet**

### DIFF
--- a/src/unix/linux_like/linux/gnu/b32/mips/mod.rs
+++ b/src/unix/linux_like/linux/gnu/b32/mips/mod.rs
@@ -526,6 +526,7 @@ pub const SYS_pwritev2: ::c_long = 4000 + 362;
 pub const SYS_pkey_mprotect: ::c_long = 4000 + 363;
 pub const SYS_pkey_alloc: ::c_long = 4000 + 364;
 pub const SYS_pkey_free: ::c_long = 4000 + 365;
+pub const SYS_statx: ::c_long = 4000 + 366;
 
 pub const O_DIRECT: ::c_int = 0x8000;
 pub const O_DIRECTORY: ::c_int = 0x10000;
--- a/src/unix/linux_like/linux/gnu/b64/aarch64/mod.rs
+++ b/src/unix/linux_like/linux/gnu/b64/aarch64/mod.rs
@@ -924,6 +924,7 @@ pub const SYS_pwritev2: ::c_long = 287;
 pub const SYS_pkey_mprotect: ::c_long = 288;
 pub const SYS_pkey_alloc: ::c_long = 289;
 pub const SYS_pkey_free: ::c_long = 290;
+pub const SYS_statx: ::c_long = 291;
 
 #[link(name = "util")]
 extern "C" {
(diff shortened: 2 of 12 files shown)


gnzlbg:
> error: invalid application of 'sizeof' to incomplete type 'struct statx'

You probably need to update the musl headers or the musl-sanitized linux kernel headers for this to work. The current kernel headers used by `libc` do not contain `statx`.

oxalica:
@gnzlbg 
The constant `SYS_statx` is available on latest musl, and seems first introduced in v1.1.19, while the ci use musl v1.1.15.

Maybe need to upgrade?

gnzlbg:
> Maybe need to upgrade?

The current musl version used by libc on most targets is here: https://github.com/rust-lang/libc/blob/master/ci/install-musl.sh#L8  

and the linux kernel that these targets support is here:

https://github.com/rust-lang/libc/blob/master/ci/install-musl.sh#L75

However, for some targets (e.g. the mips ones), the latest available musl version in their toolchains is older, e.g., the mips 32-bit target only support musl 1.1.15: https://github.com/rust-lang/libc/blob/master/ci/docker/mips-unknown-linux-musl/Dockerfile#L14

oxalica:
@gnzlbg 

Why we use OpenWrt snapshots instead of building from souce **only** in `mips{el,}`?
Can we install them in the same way as others?

gnzlbg:
> Why we use OpenWrt snapshots instead of building from souce only in mips{el,}?

Probably because building the toolchain from source takes a long time. You can try to change those to build from source though, but given everything that OpenWrt provides it might take a while.

gnzlbg:
IIUC this is blocked by #1588 , right?

oxalica:
Yes ------------------ 原始邮件 ------------------
发件人: "gnzlbg"<notifications@github.com>
发送时间: 2019年11月22日(星期五) 凌晨2:30
收件人: "rust-lang/libc"<libc@noreply.github.com>;
抄送: "oxalica"<hooccooh@qq.com>;"Author"<author@noreply.github.com>;
主题: Re: [rust-lang/libc] Provide SYS_statx for moretarget_arch/target_env (#1577)



IIUC this is blocked by #1588 , right?
 
&mdash;

oxalica:
Rebased

gnzlbg:
Thanks! @bors: r+

PR #1578, opened 2019-10-31
- add ssm struct and setsockopt constants

fatpelt:
Signed-off-by: patrick felt <patrick.felt@sling.com>

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2471,6 +2471,7 @@ fn test_linux(target: &str) {
             true
         }
         "ipv6_mreq"
+        | "ip_mreq_source"
         | "sockaddr_in6"
         | "sockaddr_ll"
         | "in_pktinfo"
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -24,6 +24,12 @@ s! {
         pub imr_interface: in_addr,
     }
 
+    pub struct ip_mreq_source {
+        pub imr_multiaddr: in_addr,
+        pub imr_interface: in_addr,
+        pub imr_sourceaddr: in_addr,
+    }
+
     pub struct sockaddr {
         pub sa_family: sa_family_t,
         pub sa_data: [::c_char; 14],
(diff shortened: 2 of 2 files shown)


gnzlbg:
The ABI of `ip_mreq_src` appears to differ in the failing platforms.

fatpelt:
i dug through the output of the tests for + run mips64-unknown-linux-gnuabi64
 and i see 0 failures; with no lines "error: "


?

gnzlbg:
The error message is there:

```
RUNNING ALL TESTS
rust[8] = 0 != 8 (C): Rust "ip_mreq_source" -> C
rust[9] = 0 != 9 (C): Rust "ip_mreq_source" -> C
rust[10] = 0 != 10 (C): Rust "ip_mreq_source" -> C
rust[11] = 0 != 11 (C): Rust "ip_mreq_source" -> C
thread 'main' panicked at 'some tests failed', /checkout/target/mips64-unknown-linux-gnuabi64/debug/build/libc-test-14c84cd81e668114/out/main.rs:12:21
```

The struct contains `(in_addr, in_addr, in_addr)`, where `in_addr is just a `(in_addr_t)` and `in_addrt_t` is just an `u32`. So `ip_mreq_source` is at least 12 bytes long.

However, when creating a `ip_mreq_source` with the byte patter `[0, 1, 2, ..., 11]` and passing it to C, the C code is apparently reading `[0, 1, 2, ..., 7, 0, 0, 0, 0]`, that is, the value of the last `u32` is apparently being lost. 

You can try skipping the roundtrip test for this type in `libc-test/build.rs` by using `cfg.skip_roundtrip` and see if any other run-time test fails, but it appears all others are passing, which means the struct size and the field offsets match those of C, and the only thing that might be different is the calling convention.

It might also be that, for some reason, the alignment of `ip_mreq_source` is higher than `4`, such that the type might end up being 16 bytes wide, containing 4 bytes of trailing padding. But if that were the case i would expect the test to succeed till the 12-th byte, and then maybe only fail in the last 4 bytes of trailing padding.

fatpelt:
@gnzlbg your suggestion worked

PR #1584, opened 2019-11-06
Add support for shared memory operations for solaris/illumos

psumbera:
This is needed because Firefox now uses slice-deque rust crate.

### DIFF
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -1010,6 +1010,7 @@ pub const MAP_PRIVATE: ::c_int = 0x0002;
 pub const MAP_FIXED: ::c_int = 0x0010;
 pub const MAP_NORESERVE: ::c_int = 0x40;
 pub const MAP_ANON: ::c_int = 0x0100;
+pub const MAP_ANONYMOUS: ::c_int = 0x0100;
 pub const MAP_RENAME: ::c_int = 0x20;
 pub const MAP_ALIGN: ::c_int = 0x200;
 pub const MAP_TEXT: ::c_int = 0x400;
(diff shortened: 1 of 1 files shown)


gnzlbg:
cc @bgermann @jasonbking 

@bors: r+

src/unix/solarish/mod.rs
```diff
@@ -206,6 +217,24 @@ s! {
         pub ai_next: *mut addrinfo,
     }
 
+    pub struct shmid_ds {
+        pub shm_perm: ipc_perm,
+        pub shm_segsz: ::size_t,
+        pub shm_flags: ::uintptr_t,
```
jasonbking: There is no shm_flags in illumos.  This is also at the location where shm_amp would be for illumos. In addition, shm_gransize and shm_allocated are not defined on illumos, and the padding appears to be incorrect for illumos.

jasonbking: https://github.com/illumos/illumos-gate/blob/master/usr/src/uts/common/sys/shm.h#L88-L115 shows the illumos definition (as @bgermann said, currently there is no 32-bit support for rust on illumos, so it's probably not necessary to add the 32-bit versions -- if that were to ever happen, it'd probably better be done as its own change since many more things would need to be updated).

jasonbking: The IPC_\* values are ok -- for likely historical reasons lost to time, the values are defined in octal on illumos (but appear to match what you have)

bgermann: Oh, yes, so forget about that.

gnzlbg:
Is there anything that @psumbera needs to update here or is this ready?

jasonbking:
The shmid_ds struct is wrong for illumos.  I cannot say if it's correct for Solaris or not.  At minimum it either needs to be guarded so it's only used with the solaris target, or updated so it's correct on both.

jasonbking:
Looks good now. Thanks for the update!

PR #1588, opened 2019-11-10
Upgrade to musl 1.1.24 in CI

oxalica:
Required by #1577 

Note that in musl 1.1.24, `struct sched_param` from `sched.h` has changed and some fields became reserved. So [these fields](https://github.com/rust-lang/libc/blob/13d4a5da2eafd82d3ebb2acb729d8b8c9148fb1f/src/unix/linux_like/mod.rs#L97) are outdated. I'm not sure if we should rename them, since they are in public API.

I simply skip `struct sched_param` from the test now.

Here's the diff between musl 1.1.23 and 1.1.24
```
diff --git a/include/sched.h b/include/sched.h
index 05d40b1e..7e470d3a 100644
--- a/include/sched.h
+++ b/include/sched.h
@@ -18,10 +18,12 @@ extern "C" {
 
 struct sched_param {
        int sched_priority;
-       int sched_ss_low_priority;
-       struct timespec sched_ss_repl_period;
-       struct timespec sched_ss_init_budget;
-       int sched_ss_max_repl;
+       int __reserved1;
+       struct {
+               time_t __reserved1;
+               long __reserved2;
+       } __reserved2[2];
+       int __reserved3;
 };
```

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2061,7 +2061,6 @@ fn test_linux(target: &str) {
     let i686 = target.contains("i686");
     let mips = target.contains("mips");
     let mips32 = mips && !target.contains("64");
-    let mips32_musl = mips32 && musl;
     let mips64 = mips && target.contains("64");
     let ppc64 = target.contains("powerpc64");
     let s390x = target.contains("s390x");
@@ -2069,6 +2068,7 @@ fn test_linux(target: &str) {
     let x32 = target.contains("x32");
     let x86_32 = target.contains("i686");
     let x86_64 = target.contains("x86_64");
+    let aarch64_musl = target.contains("aarch64") && musl;
 
     let mut cfg = ctest_cfg();
     cfg.define("_GNU_SOURCE", None);
@@ -2132,8 +2132,7 @@ fn test_linux(target: &str) {
                "sys/prctl.h",
                "sys/ptrace.h",
                "sys/quota.h",
-               // FIXME: the mips-musl CI build jobs use ancient musl 1.0.15:
-               [!mips32_musl]: "sys/random.h",
+               "sys/random.h",
                "sys/reboot.h",
                "sys/resource.h",
                "sys/sem.h",
@@ -2187,8 +2186,7 @@ fn test_linux(target: &str) {
         "linux/fs.h",
         "linux/futex.h",
         "linux/genetlink.h",
-        // FIXME: musl version 1.0.15 used by mips build jobs is ancient
-        [!mips32_musl]: "linux/if.h",
+        "linux/if.h",
         "linux/if_addr.h",
         "linux/if_alg.h",
         "linux/if_ether.h",
@@ -2303,9 +2301,6 @@ fn test_linux(target: &str) {
             // structs.
             "termios2" => true,
 
-            // FIXME: musl version using by mips build jobs 1.0.15 is ancient:
-            "ifmap" | "ifreq" | "ifconf" if mips32_musl => true,
-
             // FIXME: remove once Ubuntu 20.04 LTS is released, somewhere in 2020.
             // ucontext_t added a new field as of glibc 2.28; our struct definition is
             // conservative and omits the field, but that means the size doesn't match for newer
@@ -2349,7 +2344,7 @@ fn test_linux(target: &str) {
             // Require Linux kernel 5.1:
             "F_SEAL_FUTURE_WRITE" => true,
 
-            // The musl version 1.0.22 used in CI does not
+            // The musl version 1.1.24 used in CI does not
             // contain these glibc constants yet:
             | "RLIMIT_RTTIME" // should be in `resource.h`
             | "TCP_COOKIE_TRANSACTIONS"  // should be in the `netinet/tcp.h` header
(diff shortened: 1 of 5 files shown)


oxalica:
There's another change that [field `__seq`](https://github.com/rust-lang/libc/blob/13d4a5da2eafd82d3ebb2acb729d8b8c9148fb1f/src/unix/linux_like/linux/musl/b64/aarch64/mod.rs#L59) of `struct ipc_perm` on musl-aarch64 becomes `int` from `unsigned short` after musl 1.1.24
It's also part of public API.

oxalica:
ping? @gnzlbg

libc-test/build.rs
```diff
@@ -2458,7 +2449,17 @@ fn test_linux(target: &str) {
...
+                                           field == "ssi_arch")) ||
+        // FIXME: After musl 1.1.24, it have only one field `sched_priority`,
+        // while other fields become reserved.
+        (struct_ == "sched_param" && [
+            "sched_ss_low_priority",
+            "sched_ss_repl_period",
+            "sched_ss_init_budget",
+            "sched_ss_max_repl",
+        ].contains(&field) && musl) ||
+        // FIXME: After musl 1.1.24, the type becomes `int` instead of `unsigned short`.
```
gnzlbg: This means that musl 1.1 has ABI breaking changes with respect to musl 1.0.x ?

libc-test/build.rs
```diff
@@ -2458,7 +2449,17 @@ fn test_linux(target: &str) {
                                            field == "_pad2" ||
                                            field == "ssi_syscall" ||
                                            field == "ssi_call_addr" ||
-                                           field == "ssi_arch"))
+                                           field == "ssi_arch")) ||
+        // FIXME: After musl 1.1.24, it have only one field `sched_priority`,
+        // while other fields become reserved.
```
gnzlbg: Can you elaborate on what this means?

libc-test/build.rs
oxalica: Yes

libc-test/build.rs
oxalica: I placed the diff of musl in [the initial comment](https://github.com/rust-lang/libc/pull/1588#issue-339146716)

libc-test/build.rs
oxalica: @gnzlbg Oh, sorry. It may not be an **ABI breaking** change. The struct layout is kept and just some fields are renamed to reserved.

libc-test/build.rs
gnzlbg: Thanks, there are functions for mapping Rust field names to C field names in ctest, e.g., https://docs.rs/ctest/0.2.22/ctest/struct.TestGenerator.html#method.field_name . So maybe we can just translate the Rust names to the new __reservedX names for now ?

libc-test/build.rs
gnzlbg: Does that refer to the change in type of `ipc_perm.__seq` field from `unsigned short` to `int` on aarch64 ? Do you have a musl diff for this change?

mati865: IIUC it was unsupported https://github.com/bminor/musl/commit/827aa8fbcac89a63c6efb986871663861500cd13
But still it's breaking change between 1.1.23 and 1.1.24

oxalica:
@gnzlbg 

The `struct icp_perm` on aarch64 is in `arch/aarch64/bits/ipc.h` of musl 1.1.23.
```c
struct ipc_perm {
        key_t __ipc_perm_key;
        uid_t uid;
        gid_t gid;
        uid_t cuid;
        gid_t cgid;
        mode_t mode;
        unsigned short __ipc_perm_seq;

        unsigned long __pad1;
        unsigned long __pad2;
};

#define IPC_64 0
```

But the file is removed in musl 1.1.24. It seems to fallback to `arch/generic/bits/ipc.h`.
```c
struct ipc_perm {
        key_t __ipc_perm_key;
        uid_t uid;
        gid_t gid;
        uid_t cuid;
        gid_t cgid;
        mode_t mode;
        int __ipc_perm_seq;
        long __pad1;
        long __pad2;
};
```

Note that only the field `__ipc_perm_seq` changed to `int` from `unsigned short`.
Due to the padding, the size and field offsets of the struct are unchanged.
But it will still lead to problem due to incompatible type (interpret the padding as a part of `int`).

libc-test/build.rs
mati865: Relevant commit for this change https://github.com/bminor/musl/commit/319b2d02e7799107c0090eb088be48d17df0825c

libc-test/build.rs
oxalica: @gnzlbg 
>  So maybe we can just translate the Rust names to the new __reservedX names for now ?

This does not work for `struct { time_t __reserved1; long __reserved2; } __reserved2[2];`.
The anonymous struct does not match `struct timespec`.  [Build log](https://travis-ci.org/oxalica/libc/jobs/614414350#L5198)

We still need to skip these fields at least.

gnzlbg: Yes, you are right.

alexcrichton:
This seems fine to me to land, and it should also be fine to land an update in rust-lang/rust

oxalica:
Do we need to update the public API?

PR #1598, opened 2019-11-18
Provide definition for tcp_info

NeoLegends:
`struct tcp_info` contains two bitfields. I was unsure how to implement them. Could anybody point me to the right direction?

### DIFF
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -193,6 +193,57 @@ s! {
         pub msg_hdr: ::msghdr,
         pub msg_len: ::c_uint,
     }
+}
+
+s_no_extra_traits! {
+    #[cfg_attr(
+        any(
+            all(
+                target_arch = "x86",
+                not(target_env = "musl"),
+                not(target_os = "android")),
+            target_arch = "x86_64"),
+        repr(packed))]
+    pub struct epoll_event {
+        pub events: u32,
+        pub u64: u64,
+    }
+
+    pub struct sockaddr_un {
+        pub sun_family: sa_family_t,
+        pub sun_path: [::c_char; 108]
+    }
+
+    pub struct sockaddr_storage {
+        pub ss_family: sa_family_t,
+        __ss_align: ::size_t,
+        #[cfg(target_pointer_width = "32")]
+        __ss_pad2: [u8; 128 - 2 * 4],
+        #[cfg(target_pointer_width = "64")]
+        __ss_pad2: [u8; 128 - 2 * 8],
+    }
+
+    pub struct utsname {
+        pub sysname: [::c_char; 65],
+        pub nodename: [::c_char; 65],
+        pub release: [::c_char; 65],
+        pub version: [::c_char; 65],
+        pub machine: [::c_char; 65],
+        pub domainname: [::c_char; 65]
+    }
+
+    pub struct sigevent {
+        pub sigev_value: ::sigval,
+        pub sigev_signo: ::c_int,
+        pub sigev_notify: ::c_int,
+        // Actually a union.  We only expose sigev_notify_thread_id because it's
+        // the most useful member
+        pub sigev_notify_thread_id: ::c_int,
+        #[cfg(target_pointer_width = "64")]
+        __unused1: [::c_int; 11],
+        #[cfg(target_pointer_width = "32")]
+        __unused1: [::c_int; 12]
+    }
 
     pub struct tcp_info {
         pub tcpi_state: u8,
@@ -283,57 +334,6 @@ s! {
 
         /// peer's advertised receive window after scaling in bytes
         pub tcpi_snd_wnd: u32,
-    }
-}
-
-s_no_extra_traits! {
-    #[cfg_attr(
-        any(
-            all(
-                target_arch = "x86",
-                not(target_env = "musl"),
-                not(target_os = "android")),
-            target_arch = "x86_64"),
-        repr(packed))]
-    pub struct epoll_event {
-        pub events: u32,
-        pub u64: u64,
-    }
-
-    pub struct sockaddr_un {
-        pub sun_family: sa_family_t,
-        pub sun_path: [::c_char; 108]
-    }
-
-    pub struct sockaddr_storage {
-        pub ss_family: sa_family_t,
-        __ss_align: ::size_t,
-        #[cfg(target_pointer_width = "32")]
-        __ss_pad2: [u8; 128 - 2 * 4],
-        #[cfg(target_pointer_width = "64")]
-        __ss_pad2: [u8; 128 - 2 * 8],
-    }
-
-    pub struct utsname {
-        pub sysname: [::c_char; 65],
-        pub nodename: [::c_char; 65],
-        pub release: [::c_char; 65],
-        pub version: [::c_char; 65],
-        pub machine: [::c_char; 65],
-        pub domainname: [::c_char; 65]
-    }
-
-    pub struct sigevent {
-        pub sigev_value: ::sigval,
-        pub sigev_signo: ::c_int,
-        pub sigev_notify: ::c_int,
-        // Actually a union.  We only expose sigev_notify_thread_id because it's
-        // the most useful member
-        pub sigev_notify_thread_id: ::c_int,
-        #[cfg(target_pointer_width = "64")]
-        __unused1: [::c_int; 11],
-        #[cfg(target_pointer_width = "32")]
-        __unused1: [::c_int; 12]
     }
 }
 
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -585,7 +585,7 @@ cfg_if! {
                     .field("tcpi_wscale", &self.tcpi_wscale)
                     .field(
                         "tcpi_delivery_rate_app_limited",
-                        &self.tcpi_delivery_rate_app_limited,
+                        &(self.tcpi_delivery_rate_app_limited & 1),
                     )
                     .field("tcpi_rto", &self.tcpi_rto)
                     .field("tcpi_ato", &self.tcpi_ato)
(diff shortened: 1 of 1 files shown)


gnzlbg:
> struct tcp_info contains two bitfields. I was unsure how to implement them. 

Rust does not support bitfields, you probably need to just match the struct layout using private variables, and then in some other crate expose the bit fields on top.

NeoLegends:
> Rust does not support bitfields, you probably need to just match the struct layout using private variables, and then in some other crate expose the bit fields on top.

Makes sense, thank you.

In theory this should be good now, but one test seems to be failing in that there are fields missing / wrongly named. In particular the generated test fails because I exposed two 4 bit bitfields as one u8, which messes up the naming (because the name of the u8 is invented by me and not found in the native struct). How do I make the tests pass in that case? By appending private members with the right name (but wrong size?) to the end of the struct? I presume this would cause issues with correctly initializing a `tcp_info`, because the size would differ from the kernel one.

In other failing test cases, there seem to be members fully missing from the native struct. I copied the definition directly from the current kernel source tree, which was last modified September 16th (the part where I copied it from). Do you think it could be a header version mismatch that's causing the errors? How would we resolve these (if ever?)? It does seem like `tcp_info`s definition is not very rigid and can change rather frequently.

gnzlbg:
> How do I make the tests pass in that case?

You probably need to skip the particular field name that does not exist in C in the `libc-test/build.rs` using a `cfg.skip_...` call.

> It does seem like tcp_infos definition is not very rigid and can change rather frequently.

This might need some more thought. If we need to change the internals of `tcp_info`, it might make sense to make them private. Also, the kernel does in general evolve its API in a backward compatible. It would be interesting to investigate how they handle these upgrades without breaking clients. Maybe the `tcp_info` type is only exposed behind a pointer, or it is an unsized type (e.g. containing a VLA field at the end), or similar.

NeoLegends:
> You probably need to skip the particular field name that does not exist in C in the libc-test/build.rs using a cfg.skip_... call.

Will do, thanks!

> Maybe the tcp_info type is only exposed behind a pointer, or it is an unsized type (e.g. containing a VLA field at the end), or similar.

The kernel gives you a `tcp_info` as the result of `getsockopt`, so you have to allocate memory for it, and `getsockopt` will fill it out and afterwards tell you how many bytes it wrote. Given that you have to know how much memory to allocate (as long as you're trying to be reasonable and allocate just the right amount), I have no idea how this would be evolved in practice without breaking clients. As far as I can see, it's a regular, sized C-struct. The last commit that affected that file https://github.com/torvalds/linux/commit/8f7baad7f03543451af27f5380fc816b008aa1f2 actually added a member.

-----

Update: Just read some manpages and found ([here](http://man7.org/linux/man-pages/man7/tcp.7.html)):

> TCP_INFO (since Linux 2.4)
> Used to collect information about this socket. The kernel returns a struct tcp_info as defined in the file /usr/include/linux/tcp.h. This option should not be used in code intended to be portable.

I guess "portable" refers to the kernel version / being tied to a specific machine and installation, right?

NeoLegends:
Update: In my own code I'm now using bindgen to generate the rust definitions at compile time, and the machines I run the code on (Ubuntu 18.04 LTS) have a significantly different / older definition of `tcp_info` than what is currently found in kernel sources. I think this rules out the possibility of exposing this in libc. Thanks for your guidance @gnzlbg !

gnzlbg:
The kernel sources and the libc sources might be different. In general, the definition that `rust-bindgen` generates should be forward compatible, e.g., future kernel changes cannot break code that uses it. 

> I have no idea how this would be evolved in practice without breaking clients. As far as I can see, it's a regular, sized C-struct. The last commit that affected that file torvalds/linux@8f7baad actually added a member.

IIUC, the struct had padding at the end, and adding that field uses that padding, without changing the actual size of the struct. But I've never used this type, so no idea. Do you know of any C libraries that work on this type? Changes to the kernel that need these libraries to evolve would be reflected there.

NeoLegends:
I think I confused two files. I thought I saw a commit that added a member in the middle of the struct (which would break backwards compatibility), but I now think that was in a different file. Then you're right. We could pick the version of `tcp_info` from the minimum kernel version libc supports and expose that to Rust. I'm not sure about the usefulness of that, however. WDYT?

gnzlbg:
I think that would be a good start, the minimum supported kernel version is 2.6.18. 

Looking at the kernel headers ([5.4rc8](https://elixir.bootlin.com/linux/v5.4-rc8/source/include/uapi/linux/tcp.h#L206)) and doing a diff with 2.6.18: 

```diff
@@ -7,6 +7,7 @@ struct tcp_info
        __u8    tcpi_backoff;
        __u8    tcpi_options;
        __u8    tcpi_snd_wscale : 4, tcpi_rcv_wscale : 4;
+       __u8    tcpi_delivery_rate_app_limited:1;
 
        __u32   tcpi_rto;
        __u32   tcpi_ato;
@@ -39,4 +40,36 @@ struct tcp_info
        __u32   tcpi_rcv_space;
 
        __u32   tcpi_total_retrans;
+
+       __u64   tcpi_pacing_rate;
+       __u64   tcpi_max_pacing_rate;
+       __u64   tcpi_bytes_acked;    /* RFC4898 tcpEStatsAppHCThruOctetsAcked */
+       __u64   tcpi_bytes_received; /* RFC4898 tcpEStatsAppHCThruOctetsReceived */
+       __u32   tcpi_segs_out;       /* RFC4898 tcpEStatsPerfSegsOut */
+       __u32   tcpi_segs_in;        /* RFC4898 tcpEStatsPerfSegsIn */
+
+       __u32   tcpi_notsent_bytes;
+       __u32   tcpi_min_rtt;
+       __u32   tcpi_data_segs_in;      /* RFC4898 tcpEStatsDataSegsIn */
+       __u32   tcpi_data_segs_out;     /* RFC4898 tcpEStatsDataSegsOut */
+
+       __u64   tcpi_delivery_rate;
+
+       __u64   tcpi_busy_time;      /* Time (usec) busy sending data */
+       __u64   tcpi_rwnd_limited;   /* Time (usec) limited by receive window */
+       __u64   tcpi_sndbuf_limited; /* Time (usec) limited by send buffer */
+
+       __u32   tcpi_delivered;
+       __u32   tcpi_delivered_ce;
+
+       __u64   tcpi_bytes_sent;     /* RFC4898 tcpEStatsPerfHCDataOctetsOut */
+       __u64   tcpi_bytes_retrans;  /* RFC4898 tcpEStatsPerfOctetsRetrans */
+       __u32   tcpi_dsack_dups;     /* RFC4898 tcpEStatsStackDSACKDups */
+       __u32   tcpi_reord_seen;     /* reordering events seen */
+
+       __u32   tcpi_rcv_ooopack;    /* Out-of-order packets received */
+
+       __u32   tcpi_snd_wnd;        /* peer's advertised receive window after
+                                     * scaling (bytes)
+                                     */
 };
```

it appears that the file has grown quite a bit. 

cc @cuviper - I'm not an expert in these particular APIs, but AFAICT one needs to pass the `sizeof(tcp_info)` to all kernel APIs using it, right ? If so, then that's how the kernel is able to do the right thing for different "versions" of the struct (e.g. not try to set a field that's out-of-bounds because a new kernel version defines it but your code uses the old headers).

NeoLegends:
@gnzlbg You tell the kernel how large the buffer you allocated is, yes (cf. `getsockopt`). Reopen this?

gnzlbg:
So @NeoLegends then I think that might be enough to get this to work. You can just implement the 5.x kernel header version, and it should "just work". Looking at the definition, you need to make the bitsets a single private type that fills in the padding between the last field before the bitset and the first field after it. I think you probably want 4 bytes of padding with an alignment of 1, so something like `/*private*/ __pad: [u8; 4], ` might work.

NeoLegends:
But this will leave some fields uninitialized / zero-initialized, if one is using the 5.x definitions on older kernels. Is that acceptable since `getsockopt` is unsafe anyway?

gnzlbg:
If you wanted to let the kernel initialize the struct, you'd need to be able to query which fields did the kernel initialize, so that you can initialize the others - it would surprise me if the kernel provides such an API. 

So the only meaningful thing you can do is zero-initialize the value on construction in Rust, such that if the kernel only writes to some fields afterwards, the remaining fields are still properly initialized. For example, something in this spirit:

```rust
let x: tcp_info = mem::zeroed(); // or MaybeUninit::zeroed().assume_init()
getsockopt(&x, size_of::<tcp_info>())
```

Does that make sense?

NeoLegends:
Sure it would, I was just wondering whether that was acceptable. :) Will implement this ASAP.

gnzlbg:
Check out how we run our CI (e.g. ci/run_docker.sh and maybe the azure.yml script), so that you can run it for x86_64-unknown-linux-gnu locally at least. The process is not very well documented, but the scripts should "just run". That might simplify debugging. Let me know if you run into any issues.

NeoLegends:
In regards to your https://github.com/rust-lang/libc/pull/1598#issuecomment-558089893, do you really think we need to get the padding right there? Don't you think including two private `u8` should be enough to let `#[repr(C)]` to figure out the rest?

cuviper:
> cc @cuviper - I'm not an expert in these particular APIs, but AFAICT one needs to pass the `sizeof(tcp_info)` to all kernel APIs using it, right ? If so, then that's how the kernel is able to do the right thing for different "versions" of the struct (e.g. not try to set a field that's out-of-bounds because a new kernel version defines it but your code uses the old headers).

> @gnzlbg You tell the kernel how large the buffer you allocated is, yes (cf. `getsockopt`). Reopen this?

Right, in `getsockopt` you pass `socklen_t *optlen` with the size of your buffer, and the kernel will then set it to the length actually used, the minimum of your length and its own `tcp_info`. So if your struct is older/smaller, only your part will be filled; if the kernel's struct is older/smaller, the remaining part of your memory will be untouched. You can see how much was written by `optlen` afterward.

PR #1599, opened 2019-11-19
add pthread_getname_np for mac os

YangKeao:
In macos's `pthread.h`

```c
/*SPI to set and get pthread name*/
__API_AVAILABLE(macos(10.6), ios(3.2))
int	pthread_getname_np(pthread_t,char*,size_t);

__API_AVAILABLE(macos(10.6), ios(3.2))
int	pthread_setname_np(const char*);
```

I believe `pthread_getname_np` is usable in macos after 10.6.

Signed-off-by: Yang Keao <keao.yang@yahoo.com>


gnzlbg:
@bors: r+

Thanks!

gnzlbg:
@bors: retry

@pietroalbini I've had to retry each PR like 3-4 times today. I'm not sure what's going on, but it appears that Azure connections to download the android and mips tools, as well as rust toolchains, are constantly timing out.

PR #1603, opened 2019-11-22
Implement utmp for solarish targets

GrayJack:
I followed these references:

https://github.com/illumos/illumos-gate/blob/4e0c5eff9af325c80994e9527b7cb8b3a1ffd1d4/usr/src/head/utmpx.h
https://github.com/illumos/illumos-gate/blob/4e0c5eff9af325c80994e9527b7cb8b3a1ffd1d4/usr/src/head/utmp.h

### DIFF
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -353,6 +353,16 @@ s! {
         e_termination: ::c_short,
         e_exit: ::c_short,
     }
+
+    pub struct utmp {
+        pub ut_user: [::c_char; 8],
+        pub ut_id: [::c_char; 4],
+        pub ut_line: [::c_char; 12],
+        pub ut_pid: ::c_short,
+        pub ut_type: ::c_short,
+        pub ut_exit: exit_status,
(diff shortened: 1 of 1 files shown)


glaubitz:
@GrayJack Can you fix the spelling error "solarish" => "solaris"?

GrayJack:
@glaubitz I did used that intentionally, since internally the module is called `solarish`, probably because there more than one platform that have the same base (Oracle Solaris and Illumos).

I think there was a discussion to have a illumos target for Rust, probably cause there is enough differences between both nowadays.

But if that is really gonna be a problem, sure, I can change

gnzlbg:
cc @jasonbking @bgermann

jasonbking:
Another community member coined the term -- while there's a common heritage between illumos and Solaris, they've been separate projects for well over a decade now and have diverged in some areas so it's similar to the BSDs in that sense (a lot of things in common, and a number of things that are different). As to the change itself, it looks good to me.

bgermann:
Looks good to me, too.

PR #1500, opened 2019-09-04
Support target mipsel-unknown-linux-uclibc compile

kolapapa:
issue: https://github.com/rust-lang-nursery/net2-rs/issues/88

PR #1605, opened 2019-11-26
Add ucred and socket related types to uclibc-mips32

zvirja:
Fix compilation of the [tokio](https://github.com/tokio-rs) rust for MIPS architecture when using uClibc. Added missed types, which are defined in library headers:
https://github.com/kraj/uClibc/blob/ca1c74d67dd115d059a875150e10b8560a9c35a8/libc/sysdeps/linux/common/bits/socket.h#L320
https://github.com/kraj/uClibc/blob/ca1c74d67dd115d059a875150e10b8560a9c35a8/libc/sysdeps/linux/common/bits/in.h#L35-L36
https://github.com/kraj/uClibc/blob/ca1c74d67dd115d059a875150e10b8560a9c35a8/libc/sysdeps/linux/common/bits/in.h#L140-L143

Should also solve https://github.com/rust-lang-nursery/net2-rs/issues/88

Revisit of #1500

### DIFF
--- a/src/unix/uclibc/mod.rs
+++ b/src/unix/uclibc/mod.rs
@@ -288,6 +288,12 @@ s! {
         pub msgtql: ::c_int,
         pub msgseg: ::c_ushort,
     }
+
+    pub struct ucred {
+        pub pid: ::pid_t,
+        pub uid: ::uid_t,
+        pub gid: ::gid_t,
+    }
 }
 
 s_no_extra_traits! {


zvirja:
All builds passed except one, which I believe is not related to this PR at all 😟

zvirja:
Same results, but there is one spontaneous failure due to unavailable resources 😟

I'm happy to fix one with unused constants, could you please point me to right direction what to do?

gnzlbg:
> I'm happy to fix one with unused constants, could you please point me to right direction what to do?

Sure. Some of the constants that you added are "unused" which means that they are not exported by the library. This happens, for example, if a parent module e.g. `.../mips/mod.rs` already adds them, since these will shadow the ones that get imported from `.../mips/mips32/mod.rs`. So the best way to fix this is figure out if any of the parent modules exposes them, and if the values are correct, then just remove the ones that are duplicated. If the values are incorrect, then you need to move the constant from the parent module to the "upper" (more specific) modules for which they are correct.

zvirja:
@gnzlbg Thanks for the hint! Those were indeed duplicated. Actually, after I made more precise look, I realized that the `ucred` type is common for all architectures and is not mips32 specific. So altered PR to define type on common layer and make everybody happy 😎

zvirja:
Fair enough, but worth’s trying ;-)

@gnzlbg It looks like it stuck and didn’t run tests again to merge the PR :thinking:

gnzlbg:
@zvirja there were other PRs in the queue, so bors couldn't start testing it immediately: https://buildbot2.rust-lang.org/homu/queue/libc

zvirja:
@gnzlbg Thanks for your help and review! 🤝 
Do you plan to make a new patch release in a nearby future? So I could consume the changes via crate? 😊

gnzlbg:
There are a couple of PRs in the queue with good chances of merging as well. As soon as they merge in the next couple of hours I'll do a new release.

PR #1606, opened 2019-11-27
Support for RISC-V 64-bit GNU/Linux

msizanoen1:
Add support for RISC-V 64-bit GNU/Linux.

Follow up to rust-lang/rust#66661.

r? @alexcrichton


msizanoen1:
The target specification is not in the latest nightly, only master - even when using `rustup update --force`.
It looks like we will have to wait for next nightly.

PR #1607, opened 2019-11-27
expose futimens() for illumos systems

jclulow:
illumos has an implementation of `futimens()` which we should expose.  I'm working on a broader set of [fixes for illumos](https://github.com/rust-lang/libc/compare/master...jclulow:illumos_fixes) in general, but that's going to take a lot more work to be ready so the test suite doesn't currently seem to function.


jclulow:
Apologies for missing the style check.  I'll fix it up.

PR #1608, opened 2019-11-28
use *const pointer for NetBSD's pthread_setname_np

lzutao:


gnzlbg:
@lzutao did you run the libc-test suite on NetBSD ?

lzutao:
@gnzlbg I didn't because I don't have NetBSD machines. 
On NetBSD the second argument is a printf-like string, which can refer to the value provided by the third argument.
Here is how pthread_setname_np implemented in NetBSD: https://github.com/NetBSD/src/blob/465ec24dbfce4dd9b06234d657a05da1d9777865/lib/libpthread/pthread.c#L825
```c
int
pthread_setname_np(pthread_t thread, const char *name, void *arg)
{
	char *oldname, *cp, newname[PTHREAD_MAX_NAMELEN_NP];
	int namelen;

	if (pthread__find(thread) != 0)
		return ESRCH;

	if (thread->pt_magic != PT_MAGIC)
		return EINVAL;

	namelen = snprintf(newname, sizeof(newname), name, arg);
	if (namelen >= PTHREAD_MAX_NAMELEN_NP)
		return EINVAL;

	cp = strdup(newname);
	if (cp == NULL)
		return ENOMEM;

	pthread_mutex_lock(&thread->pt_lock);
	oldname = thread->pt_name;
	thread->pt_name = cp;
	(void)_lwp_setname(thread->pt_lid, cp);
	pthread_mutex_unlock(&thread->pt_lock);

	if (oldname != NULL)
		free(oldname);

	return 0;
}
```

gnzlbg:
Thanks @lzutao , that checks out. I'm generally a bit more wary of API breaking changes like this, but for NetBSD we only guarantee that `libc` "builds".

@bors: r+

PR #1612, opened 2019-11-30
Solarish: Fix utmx Debug impl and re-add support for extra traits

GrayJack:
Fix a bug where the compilation failed cause `utmpx.host` is bigger than 32 so we cannot print it.

Due to a break of compilation (only with extra trait enabled) on a Tier 2 platform, I do think it would be better to do another release as soon as proven fixing compilation

### DIFF
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -536,7 +536,7 @@ cfg_if! {
                     .field("ut_session", &self.ut_session)
                     .field("ut_pad", &self.ut_pad)
                     .field("ut_syslen", &self.ut_syslen)
-                    .field("ut_host", &self.ut_host)
+                // .field("ut_host", &self.ut_host)
                     .finish()
             }
         }
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -561,8 +561,8 @@ cfg_if! {
             fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
                 unsafe {
                     match self {
-                        Self { d_desc } => f.debug_struct("door_desc_t__d_data").field("d_desc", &self.d_desc),
-                        Self { d_resv } => f.debug_struct("door_desc_t__d_data").field("d_resv", &self.d_resv),
+                        Self { d_desc } => f.debug_struct("door_desc_t__d_data").field("d_desc", &self.d_desc).finish(),
+                        Self { d_resv } => f.debug_struct("door_desc_t__d_data").field("d_resv", &self.d_resv).finish(),
                     }
                 }
             }
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -557,6 +557,17 @@ cfg_if! {
             }
         }
 
+        impl ::fmt::Debug for door_desc_t__d_data {
+            fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                unsafe {
+                    match self {
+                        Self { d_desc } => f.debug_struct("door_desc_t__d_data").field("d_desc", &self.d_desc),
+                        Self { d_resv } => f.debug_struct("door_desc_t__d_data").field("d_resv", &self.d_resv),
+                    }
+                }
+            }
+        }
+
         impl PartialEq for epoll_event {
             fn eq(&self, other: &epoll_event) -> bool {
                 self.events == other.events
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -479,18 +479,19 @@ s_no_extra_traits! {
         __sigev_pad2: ::c_int,
     }
 
+    #[cfg_attr(feature = "extra_traits", allow(missing_debug_implementations))]
     pub union door_desc_t__d_data {
         pub d_desc: door_desc_t__d_data__d_desc,
         d_resv: [::c_int; 5], /* Check out /usr/include/sys/door.h */
     }
 
-    #[derive(Debug)]
+    #[cfg_attr(feature = "extra_traits", allow(missing_debug_implementations))]
     pub struct door_desc_t {
         pub d_attributes: door_attr_t,
         pub d_data: door_desc_t__d_data,
     }
 
-    #[derive(Debug)]
+    #[cfg_attr(feature = "extra_traits", allow(missing_debug_implementations))]
     pub struct door_arg_t {
         pub data_ptr: *const ::c_char,
         pub data_size: ::size_t,
@@ -556,21 +557,6 @@ cfg_if! {
                 self.ut_tv.hash(state);
                 self.ut_syslen.hash(state);
                 self.ut_pad.hash(state);
-            }
-        }
-
-        impl ::fmt::Debug for door_desc_t__d_data {
-            fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
-                unsafe {
-                    match self {
-                        Self { d_desc } => f.debug_struct("door_desc_t__d_data")
-                            .field("d_desc", &self.d_desc)
-                            .finish(),
-                        Self { d_resv } => f.debug_struct("door_desc_t__d_data")
-                            .field("d_resv", &self.d_resv)
-                            .finish(),
-                    }
-                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/unix/solarish/mod.rs
```diff
@@ -536,7 +536,7 @@ cfg_if! {
                     .field("ut_session", &self.ut_session)
                     .field("ut_pad", &self.ut_pad)
                     .field("ut_syslen", &self.ut_syslen)
-                    .field("ut_host", &self.ut_host)
+                // .field("ut_host", &self.ut_host)
```
gnzlbg: The solution is to just coerce the array as a slice, and debug print that instead.

GrayJack: I was just following the pattern in linux-gnu utmpx

GrayJack: Also when I try to coerce to slice it complains about not knowing that size at compilation time

gnzlbg: I'm not sure I follow. What error do you get when you write `&self.ut_host as &[_]` ?

GrayJack: error E0277:
```sh
error[E0277]: the size for values of type `[_]` cannot be known at compilation time
   --> /home/grayjack/MySources/RustProjects/libc/src/unix/solarish/mod.rs:539:39
    |
539 |                     .field("ut_host", &self.ut_host as &[_])
    |                                       ^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `core::marker::Sized` is not implemented for `[_]`
    = note: to learn more, visit <https://doc.rust-lang.org/book/ch19-04-advanced-types.html#dynamically-sized-types-and-the-sized-trait>
    = note: required for the cast to the object type `dyn core::fmt::Debug`
```

gnzlbg: I think you need to split that up:

```rust
let s = &self.ut_host as &[_];
...
.field("ut_host", s);
```

gnzlbg: Wrong. Can you try: `.field("ut_host", &&self.ut_host)` ?

GrayJack: Well, this give me a huge error on target `x86_64-sun-solaris`, but I tested on the host field of target linux-gnu and worked fine

The error:
```sh
Checking libc v0.2.66 (/home/grayjack/MySources/RustProjects/libc)
error: type does not implement `fmt::Debug`; consider adding `#[derive(Debug)]` or a manual implementation
   --> /home/grayjack/MySources/RustProjects/libc/src/macros.rs:101:21
    |
101 |                       pub union $i { $($field)* }
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
   ::: /home/grayjack/MySources/RustProjects/libc/src/unix/solarish/mod.rs:406:1
    |
406 | / s_no_extra_traits! {
407 | |     #[cfg_attr(any(target_arch = "x86", target_arch = "x86_64"), repr(packed))]
408 | |     pub struct epoll_event {
409 | |         pub events: u32,
...   |
499 | |     }
500 | | }
    |
note: lint level defined here
   --> /home/grayjack/MySources/RustProjects/libc/src/lib.rs:32:44
    |
32  | #![cfg_attr(feature = "extra_traits", deny(missing_debug_implementations))]
    |                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: type does not implement `fmt::Debug`; consider adding `#[derive(Debug)]` or a manual implementation
   --> /home/grayjack/MySources/RustProjects/libc/src/macros.rs:115:13
    |
115 |               pub struct $i { $($field)* }
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
   ::: /home/grayjack/MySources/RustProjects/libc/src/unix/solarish/mod.rs:406:1
    |
406 | / s_no_extra_traits! {
407 | |     #[cfg_attr(any(target_arch = "x86", target_arch = "x86_64"), repr(packed))]
408 | |     pub struct epoll_event {
409 | |         pub events: u32,
...   |
499 | |     }
500 | | }
error: type does not implement `fmt::Debug`; consider adding `#[derive(Debug)]` or a manual implementation
   --> /home/grayjack/MySources/RustProjects/libc/src/macros.rs:115:13
    |
115 |               pub struct $i { $($field)* }
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
   ::: /home/grayjack/MySources/RustProjects/libc/src/unix/solarish/mod.rs:406:1
    |
406 | / s_no_extra_traits! {
407 | |     #[cfg_attr(any(target_arch = "x86", target_arch = "x86_64"), repr(packed))]
408 | |     pub struct epoll_event {
409 | |         pub events: u32,
...   |
499 | |     }
500 | | }
```

GrayJack: With further investigation, looks like we have 2 structs and 1 union inside `s_no_extra_traits!` invocation, don't implement manually the debug traits

gnzlbg: If you want to build with the `extra_traits` feature for this target, you need to manually implement the traits for all types inside `s_no_extra_traits!`, or add an `allow(...)` to disable the lint for a particular type.

GrayJack: I know, but that's the thing I wasn't me who added these 3 structures inside `s_no_extra_traits!`

gnzlbg: We only guarantee that `extra_traits` works for the platforms for which it is tested (tier-1 and some tier-2). We do not support `extra_traits` for other platforms, and it is not tested. The only reason is, nobody has added support for it.

If you want to add support for `extra_traits` for these solaris platforms, you'd have to add the impls. If you don't want to add support for `extra_traits` for this platform, then, the simplest thing would be to just don't use the `extra_traits` feature on the platform.

The main reason to add an `allow(...)` is if, for some reason, it is impossible to implement a particular trait for a type, but that's super rare.

GrayJack: It's impossible for union types to implement any of the extra traits? What about structs that have union fields?

gnzlbg: > Is it impossible for union types to implement any of the extra traits?

Not in general, you just have to manually implement them - but for many unions, you don't have enough information to provide a meaningful implementation.

> What about structs that have union fields?

Same, but for structs with union fields you often have more information to provide a meaningful implementation (e.g. a discriminant of some sort somewhere).

GrayJack: Thanks, I'll have a look!! :smile: 
I do want solarish to keep supporting extra_traits cause I used it a lot when debugging ('keep' cause it worked till 0.2.65)

gnzlbg: FWIW I've open an issue to start testing in `ci/build.sh` that `extra_traits` at least builds on all targets in which we currently are able to build libc, but that will take some work, since I expect many targets to be broken.

GrayJack:
@gnzlbg Can you have a look in the way that I implemented the `Debug` trait for that union and give me feedback for that.

Thanks

src/unix/solarish/mod.rs
```diff
@@ -557,6 +559,21 @@ cfg_if! {
...
+                        Self { d_desc } => f.debug_struct("door_desc_t__d_data")
+                            .field("d_desc", &self.d_desc)
+                            .finish(),
+                        Self { d_resv } => f.debug_struct("door_desc_t__d_data")
+                            .field("d_resv", &self.d_resv)
+                            .finish(),
+                    }
+                }
+            }
+        }
```
gnzlbg: The second match arm is unreachable (it will never get called), and it should be producing an "unreachable code" warning which should fail to build when warnings are turned into errors.

GrayJack: Yeap, I see now, but I can't see where in the pattern match that I did it wrong. I tried to follow the `pattern-matching-on-unions` on the Reference book, but it looks like it didn't worked as expected.

gnzlbg: The first pattern will always match because all variants of unions always overlap - notice that, unlike enums, unions do not have a discriminant that tells you which variant is active. 

I don't think there is a meaningful way of implementing `Debug` for this union right now. The union members are public, and it is currently unclear whether it would be ok for Rust unions to be uninitialized, so  you might need extra context information to know what to print.

GrayJack: So in that case, is it ok to allow `missing_debug_implementations` for that union?

gnzlbg: Yes!

GrayJack:
There some CI build was canceled

GrayJack:
Can we test it again?

PR #1613, opened 2019-12-02
Add sock_extended_err and associated constants from errqueue.h

mcpherrinm:
IP_RECVERR is set, when recvmsg has the MSG_ERRQUEUE flag set.  IP_RECVERR and


mcpherrinm:
Closing this for now; will try to fix libc-test offline, and reopen later, rather than spamming this PR.

PR #1614, opened 2019-12-02
Add sock_extended_err and associated constants from errqueue.h

mcpherrinm:
IP_RECVERR is set, when recvmsg has the MSG_ERRQUEUE flag set.  IP_RECVERR and

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -484,6 +484,16 @@ s! {
         pub svm_cid: ::c_uint,
         pub svm_zero: [u8; 4]
     }
+
+    pub struct sock_extended_err {
+        pub ee_errno: u32,
+        pub ee_origin: u8,
+        pub ee_type: u8,
+        pub ee_code: u8,
+        pub ee_pad: u8,
+        pub ee_info: u32,
+        pub ee_data: u32
+    }
 }
 
 s_no_extra_traits! {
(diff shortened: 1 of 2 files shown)


mcpherrinm:
Note: I put this in linux/ but I believe it should work on Android.  However, the libc-test was failing, and I think it may be related to header ordering.  It's not relevant on emscripten, so I'm not sure it belongs at the linux_like/mod.rs level.

mcpherrinm:
#1616 is nearly the same as this, but with the code up a level, at linux_like.

src/unix/linux_like/linux/mod.rs
```diff
@@ -2404,6 +2421,10 @@ f! {
     pub fn RT_LOCALADDR(flags: u32) -> bool {
         (flags & RTF_ADDRCLASSMASK) == (RTF_LOCAL | RTF_INTERFACE)
     }
+
+    pub fn SO_EE_OFFENDER(ee: *const ::sock_extended_err) -> *mut ::sockaddr {
+        ee.offset(1) as *mut ::sockaddr
+    }
```
gnzlbg: This would need a test.

mcpherrinm: Would you suggest I add a wrapper around the C macro and test equivalence, like is done for `cmsg`, eg in `test_cmsg_data`?

gnzlbg: Yes, that would be great!

mcpherrinm:
Looks like reordered headers works for android, so I'm closing this PR in favor of #1616.  I've added the test there as requested here.

PR #1616, opened 2019-12-02
Add sock_extended_err and associated constants from errqueue.h to linux_like

mcpherrinm:
This is an alternate version of #1614 that attempts to fix issues with the android ctests by rearranging the headers.  Due to long CI cycle time, I've opened it as an alternate PR.  Only one or the other should merge, depending if linux_like or linux is the correct place for this.

IP_RECVERR is set, when recvmsg has the MSG_ERRQUEUE flag set. IP_RECVERR and


mcpherrinm:
CI failure was in:
The command '/bin/sh -c curl --retry 5 -L https://downloads.openwrt.org/releases/19.07.0-rc1/targets/ar71xx/generic/openwrt-sdk-19.07.0-rc1-ar71xx-generic_gcc-7.4.0_musl.Linux-x86_64.tar.xz |     tar xJf - -C /toolchain --strip-components=1' returned a non-zero code: 2
 
Can we re-run that assuming it was an intermittent outage/error?

mcpherrinm:
Actually, I just force-pushed to make CI re-run

mcpherrinm:
@gnzlbg I added a test per the other PR. Anything else I need here?

mcpherrinm:
Hi, anything I can do to get this merged?

mcpherrinm:
thank you!

PR #1621, opened 2019-12-08
Add Redox FD_* function-macros and Android AI_* constants

coolreader18:
I'm not sure if I'm allowed to add multiple, separate APIs in the same PR, but I guess they're both for uncommon/exotic platforms?

The `AI_*` constants were taken from `usr/include/netdb.h` on my Termux install; the `FD_` functions were taken from `linux_like/mod.rs`.

### DIFF
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -1157,6 +1157,21 @@ pub const RTLD_NOLOAD: ::c_int = 0x4;
 
 pub const SEM_FAILED: *mut sem_t = 0 as *mut sem_t;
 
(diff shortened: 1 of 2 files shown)


coolreader18:
The force push was just to fix a type for not properly qualifying `c_int`.

coolreader18:
I'm not really sure why CI is failing; it just says that it's been cancelled because it ran too long but I'm not sure why that is.

coolreader18:
@gnzlbg do you have any idea why CI might be failing?

gnzlbg:
Closing and re-opening to re-run CI.

Issue #1604, opened 2019-11-26
Question: Why aren't some constans/structs in ucontext.h for x86_64-android?


igrep:
* the target triple: x86_64-linux-android, aarch64-linux-android, etc.
* link to the documentation of the API showing the type signatures, how to use
  it, etc.:
    - https://pubs.opengroup.org/onlinepubs/7908799/xsh/ucontext.h.html
    - https://android.googlesource.com/platform/bionic/+/refs/tags/ndk-r20/libc/include/sys/ucontext.h

I confirmed some symbols (e.g. `_libc_xmmreg`, `ucontext_t`, `REG_R10` etc.) are available also in Android (see the source in googlesource.com above).
So I think symbols in ucontext.h should be added under linux_like/android/.
Am I missing something?

gnzlbg:
99% of the time the answer is "because nobody has sent a PR to implement them", so if you want these, PRs welcome.

PR #1622, opened 2019-12-12
Add definitions in ucontext.h in Android for ARM, x86, and x86_64

igrep:
Ref: https://android.googlesource.com/platform/bionic/+/refs/tags/ndk-r20/libc/include/sys/ucontext.h

Will close https://github.com/rust-lang/libc/issues/1604

### DIFF
--- a/src/unix/linux_like/android/b32/arm.rs
+++ b/src/unix/linux_like/android/b32/arm.rs
@@ -1,5 +1,151 @@
 pub type c_char = u8;
 pub type wchar_t = u32;
+pub type greg_t = i32;
+pub type mcontext_t = sigcontext;
+
+s! {
+    pub struct sigcontext {
+        pub trap_no: ::c_ulong,
+        pub error_code: ::c_ulong,
+        pub oldmask: ::c_ulong,
+        pub arm_r0: ::c_ulong,
+        pub arm_r1: ::c_ulong,
+        pub arm_r2: ::c_ulong,
+        pub arm_r3: ::c_ulong,
+        pub arm_r4: ::c_ulong,
+        pub arm_r5: ::c_ulong,
+        pub arm_r6: ::c_ulong,
+        pub arm_r7: ::c_ulong,
+        pub arm_r8: ::c_ulong,
+        pub arm_r9: ::c_ulong,
+        pub arm_r10: ::c_ulong,
+        pub arm_fp: ::c_ulong,
+        pub arm_ip: ::c_ulong,
+        pub arm_sp: ::c_ulong,
+        pub arm_lr: ::c_ulong,
+        pub arm_pc: ::c_ulong,
+        pub arm_cpsr: ::c_ulong,
+        pub fault_address: ::c_ulong,
+    }
+}
+
+cfg_if! {
+    if #[cfg(libc_union)] {
+        s_no_extra_traits! {
+            pub struct __c_anonymous_uc_sigmask_with_padding {
+                pub uc_sigmask: ::sigset_t,
+                /* Android has a wrong (smaller) sigset_t on x86. */
+                __padding_rt_sigset: u32,
+            }
+
+            pub union __c_anonymous_uc_sigmask {
+                uc_sigmask: __c_anonymous_uc_sigmask_with_padding,
+                uc_sigmask64: ::sigset64_t,
+            }
+
+            pub struct ucontext_t {
+                pub uc_flags: ::c_ulong,
+                pub uc_link: *mut ucontext_t,
+                pub uc_stack: ::stack_t,
+                pub uc_mcontext: mcontext_t,
+                pub uc_sigmask__c_anonymous_union: __c_anonymous_uc_sigmask,
+                /* The kernel adds extra padding after uc_sigmask to match
+                 * glibc sigset_t on ARM. */
+                __padding: [c_char; 120],
+                __align: [::c_longlong; 0],
+                uc_regspace: [::c_ulong; 128],
+            }
+        }
+
+        cfg_if! {
+            if #[cfg(feature = "extra_traits")] {
+                impl PartialEq for __c_anonymous_uc_sigmask_with_padding {
+                    fn eq(
+                        &self, other: &__c_anonymous_uc_sigmask_with_padding
+                    ) -> bool {
+                        self.uc_sigmask == other.uc_sigmask
+                            // Ignore padding
+                    }
+                }
+                impl Eq for __c_anonymous_uc_sigmask_with_padding {}
+                impl ::fmt::Debug for __c_anonymous_uc_sigmask_with_padding {
+                    fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                        f.debug_struct("uc_sigmask_with_padding")
+                            .field("uc_sigmask_with_padding", &self.uc_sigmask)
+                            // Ignore padding
+                            .finish()
+                    }
+                }
+                impl ::hash::Hash for __c_anonymous_uc_sigmask_with_padding {
+                    fn hash<H: ::hash::Hasher>(&self, state: &mut H) {
+                        self.uc_sigmask.hash(state)
+                            // Ignore padding
+                    }
+                }
+
+                impl PartialEq for __c_anonymous_uc_sigmask {
+                    fn eq(&self, other: &__c_anonymous_uc_sigmask) -> bool {
+                        unsafe { self.uc_sigmask == other.uc_sigmask }
+                    }
+                }
+                impl Eq for __c_anonymous_uc_sigmask {}
+                impl ::fmt::Debug for __c_anonymous_uc_sigmask {
+                    fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                        f.debug_struct("uc_sigmask")
+                            .field("uc_sigmask", unsafe { &self.uc_sigmask })
+                            .finish()
+                    }
+                }
+                impl ::hash::Hash for __c_anonymous_uc_sigmask {
+                    fn hash<H: ::hash::Hasher>(&self, state: &mut H) {
+                        unsafe { self.uc_sigmask.hash(state) }
+                    }
+                }
+
+                impl PartialEq for ucontext_t {
+                    fn eq(&self, other: &Self) -> bool {
+                        self.uc_flags == other.uc_flags
+                            && self.uc_link == other.uc_link
+                            && self.uc_stack == other.uc_stack
+                            && self.uc_mcontext == other.uc_mcontext
+                            && self.uc_sigmask__c_anonymous_union
+                                == other.uc_sigmask__c_anonymous_union
+                            && &self.uc_regspace[..] == &other.uc_regspace[..]
+                            // Ignore padding field
+                    }
+                }
+                impl Eq for ucontext_t {}
+                impl ::fmt::Debug for ucontext_t {
+                    fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                        f.debug_struct("ucontext_t")
+                            .field("uc_flags", &self.uc_flags)
+                            .field("uc_link", &self.uc_link)
+                            .field("uc_stack", &self.uc_stack)
+                            .field("uc_mcontext", &self.uc_mcontext)
+                            .field(
+                                "uc_sigmask__c_anonymous_union",
+                                &self.uc_sigmask__c_anonymous_union
+                            )
+                            .field("uc_regspace", &&self.uc_regspace[..])
+                            // Ignore padding field
+                            .finish()
+                    }
+                }
+                impl ::hash::Hash for ucontext_t {
+                    fn hash<H: ::hash::Hasher>(&self, state: &mut H) {
+                        self.uc_flags.hash(state);
+                        self.uc_link.hash(state);
+                        self.uc_stack.hash(state);
+                        self.uc_mcontext.hash(state);
+                        self.uc_sigmask__c_anonymous_union.hash(state);
+                        &self.uc_regspace[..].hash(state);
+                        // Ignore padding field
+                    }
+                }
+            }
+        }
+    }
+}
 
 pub const O_DIRECT: ::c_int = 0x10000;
 pub const O_DIRECTORY: ::c_int = 0x4000;
@@ -355,3 +501,23 @@ pub const SYS_pwritev2: ::c_long = 393;
 pub const SYS_pkey_mprotect: ::c_long = 394;
 pub const SYS_pkey_alloc: ::c_long = 395;
 pub const SYS_pkey_free: ::c_long = 396;
+
+// offsets in mcontext_t.gregs from sys/ucontext.h
+pub const REG_R0: ::c_int = 0;
+pub const REG_R1: ::c_int = 1;
+pub const REG_R2: ::c_int = 2;
+pub const REG_R3: ::c_int = 3;
+pub const REG_R4: ::c_int = 4;
+pub const REG_R5: ::c_int = 5;
+pub const REG_R6: ::c_int = 6;
+pub const REG_R7: ::c_int = 7;
+pub const REG_R8: ::c_int = 8;
+pub const REG_R9: ::c_int = 9;
+pub const REG_R10: ::c_int = 10;
+pub const REG_R11: ::c_int = 11;
+pub const REG_R12: ::c_int = 12;
+pub const REG_R13: ::c_int = 13;
+pub const REG_R14: ::c_int = 14;
+pub const REG_R15: ::c_int = 15;
+
+pub const NGREG: ::c_int = 18;
--- a/src/unix/linux_like/android/b32/x86/mod.rs
+++ b/src/unix/linux_like/android/b32/x86/mod.rs
@@ -1,5 +1,144 @@
 pub type c_char = i8;
 pub type wchar_t = i32;
+pub type greg_t = i32;
+
+s! {
+    pub struct _libc_fpreg {
+        pub significand: [u16; 4],
+        pub exponent: u16,
+    }
+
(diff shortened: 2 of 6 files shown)


src/unix/linux_like/android/b32/arm.rs
```diff
@@ -1,5 +1,128 @@
...
+s_no_extra_traits! {
+    pub struct ucontext_t {
+        pub uc_flags: ::c_ulong,
+        pub uc_link: *mut ucontext_t,
+        pub uc_stack: ::stack_t,
+        // NOTE This field's type is actually mcontext_t,
+        //      which is a synonym for sigcontext.
+        //      But to avoid the error reported by style checker,
+        //      it must be sigcontext.
+        pub uc_mcontext: sigcontext,
```
gnzlbg: What error is this? Using `mcontext_t` should work just fine?

src/unix/linux_like/android/b32/arm.rs
```diff
@@ -1,5 +1,128 @@
...
+        // FIXME this and __padding_rt_sigset are actually a union
+        pub uc_sigmask: ::sigset_t,
```
gnzlbg: Then use an `union` type.

src/unix/linux_like/android/b32/arm.rs
```diff
@@ -1,5 +1,128 @@
...
+        }
+
+        impl PartialEq for ucontext_t {
+            fn eq(&self, other: &Self) -> bool {
+                self.uc_flags == other.uc_flags
+                    && self.uc_link == other.uc_link
+                    && self.uc_stack == other.uc_stack
+                    && self.uc_mcontext == other.uc_mcontext
+                    && self.uc_sigmask == other.uc_sigmask
+                    && Array128(self.uc_regspace)
```
gnzlbg: You can write `&self.uc_regspace[..] == &other.uc_regspace[..]` instead.

src/unix/linux_like/android/b32/arm.rs
```diff
@@ -1,5 +1,128 @@
...
+        impl Eq for ucontext_t {}
+        impl ::fmt::Debug for ucontext_t {
+            fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                f.debug_struct("ucontext_t")
+                    .field("uc_flags", &self.uc_flags)
+                    .field("uc_link", &self.uc_link)
+                    .field("uc_stack", &self.uc_stack)
+                    .field("uc_mcontext", &self.uc_mcontext)
+                    .field("uc_sigmask", &self.uc_sigmask)
+                    .field("uc_regspace", &Array128(self.uc_regspace))
```
gnzlbg: Same here: `&self.uc_regspace[..]` should work.

src/unix/linux_like/android/b32/arm.rs
```diff
@@ -1,5 +1,128 @@
...
+            }
+        }
+        impl ::hash::Hash for ucontext_t {
+            fn hash<H: ::hash::Hasher>(&self, state: &mut H) {
+                self.uc_flags.hash(state);
+                self.uc_link.hash(state);
+                self.uc_stack.hash(state);
+                self.uc_mcontext.hash(state);
+                self.uc_sigmask.hash(state);
+                Array128(self.uc_regspace).hash(state);
```
gnzlbg: Same here: `&self.uc_regspace[..]` should work.

src/unix/linux_like/android/b32/x86/mod.rs
```diff
@@ -1,5 +1,43 @@
...
+        pub cr2: ::c_ulong,
+    }
+
+    pub struct ucontext_t {
+        pub uc_flags: ::c_ulong,
+        pub uc_link: *mut ucontext_t,
+        pub uc_stack: ::stack_t,
+        pub uc_mcontext: mcontext_t,
+        // FIXME this and __padding_rt_sigset are actually a union
+        pub uc_sigmask: ::sigset_t,
```
gnzlbg: Use an `union` then.

src/unix/linux_like/android/b64/x86_64/mod.rs
```diff
@@ -41,6 +42,46 @@ s! {
         pub st_ctime_nsec: ::c_long,
         __unused: [::c_long; 3],
     }
+
+    pub struct _libc_fpxreg {
+        pub significand: [u16; 4],
+        pub exponent: u16,
+        __padding: [u16; 3], // TODO: ignore padding
```
gnzlbg: What is exactly what remains to be done here ?

The padding will participate in the Debug, Hash, etc. impls, so this type needs to be defined in `s_no_extra_traits!`.

src/unix/linux_like/android/b64/mod.rs
```diff
@@ -105,6 +105,10 @@ s! {
         pub f_namemax: ::c_ulong,
         __f_spare: [::c_int; 6],
     }
+
+    pub struct sigset64_t {
+        __bits: [::c_ulong; 1]
+    }
```
gnzlbg: It makes more sense to move this to `s_no_extra_traits!` and implement the traits in a meaningful way (if there is one), or not implement them at all.

src/unix/linux_like/android/b64/x86_64/mod.rs
```diff
@@ -41,6 +42,46 @@ s! {
...
+        pub swd: u16,
+        pub ftw: u16,
+        pub fop: u16,
+        pub rip: u64,
+        pub rdp: u64,
+        pub mxcsr: u32,
+        pub mxcr_mask: u32,
+        pub _st: [_libc_fpxreg; 8],
+        pub _xmm: [_libc_xmmreg; 16],
+        __private: [u32; 24], // TODO: ignore padding
```
gnzlbg: Should be in `s_no_extra_traits!` due to padding.

src/unix/linux_like/android/b64/x86_64/mod.rs
```diff
@@ -41,6 +42,46 @@ s! {
...
+    }
+
+    pub struct mcontext_t {
+        pub gregs: [greg_t; 23],
+        pub fpregs: *mut _libc_fpstate,
+        __private: [u64; 8], // TODO: ignore padding
```
gnzlbg: Should be in `s_no_extra_traits!` due to padding.

src/unix/linux_like/android/b64/x86_64/mod.rs
```diff
@@ -41,6 +42,46 @@ s! {
...
+
+    pub struct ucontext_t {
+        pub uc_flags: ::c_ulong,
+        pub uc_link: *mut ucontext_t,
+        pub uc_stack: ::stack_t,
+        pub uc_mcontext: mcontext_t,
+        // FIXME this is actually a union
+        pub uc_sigmask64: ::sigset_t,
+        __fpregs_mem: _libc_fpstate,
+    }
```
gnzlbg: Should be in `s_no_extra_traits!` due to padding.

src/unix/linux_like/android/b64/x86_64/mod.rs
```diff
@@ -41,6 +42,46 @@ s! {
...
+        __private: [u64; 8], // TODO: ignore padding
+    }
+
+    pub struct ucontext_t {
+        pub uc_flags: ::c_ulong,
+        pub uc_link: *mut ucontext_t,
+        pub uc_stack: ::stack_t,
+        pub uc_mcontext: mcontext_t,
+        // FIXME this is actually a union
+        pub uc_sigmask64: ::sigset_t,
```
gnzlbg: Should use an `union`.

igrep:
All of your comments don't sound minor to me... 😂 
Anyway, I'll fix as many as possible.
Thanks.

src/unix/linux_like/android/b32/arm.rs
igrep: At first I wrote [like this](https://github.com/rust-lang/libc/pull/1622/commits/546a690047f01e85dd50f5b0967fa03be5294a3b#diff-8c71e51ad9afb09c731c4ec6ac99ed44R31-R41):

```
pub type mcontext_t = sigcontext;

cfg_if! {
    if #[cfg(libc_union)] {
        s_no_extra_traits! {

            pub struct ucontext_t {
...
                pub uc_mcontext: mcontext_t,
...
```

But I got this error:

```
src/unix/linux_like/android/b32/arm.rs:31 - typedef found after struct when it belongs before
```

from https://dev.azure.com/rust-lang2/libc/_build/results?buildId=949&view=logs&j=c5397555-4577-58d8-d518-821f0664b4e0&t=01cef1e2-802e-5c67-cfe4-12bd866a84b0&l=16

I guess this is because the macros are expanded and evaluated before `pub type mcontext_t = sigcontext;` is evaluated.

So I tried to move `pub type mcontext_t = sigcontext;` inside `s_no_extra_traits!` block.
But that wasn't compilable because `s_no_extra_traits!` can't match `pub type ...`.
Maybe I should fix `s_no_extra_traits!`, but I'm not sure about Rust's macros.

src/unix/linux_like/android/b32/arm.rs
igrep: How can I write `union`s like this in Rust?

```c
typedef struct ucontext {
  ...
  union {
    sigset_t uc_sigmask;
    sigset64_t uc_sigmask64;
  };
  ...
} ucontext_t;
```

(from https://android.googlesource.com/platform/bionic/+/refs/tags/ndk-r20/libc/include/sys/ucontext.h#381)

At first I made a separate types to represent the anonymous `union`s ([here](https://github.com/rust-lang/libc/pull/1622/commits/546a690047f01e85dd50f5b0967fa03be5294a3b#diff-8c71e51ad9afb09c731c4ec6ac99ed44R55-R58)):

```rust
            pub struct ucontext_t {
...
                pub uc_sigmask: uc_sigmask,
...
            }

            pub struct uc_sigmask_with_padding {
                pub uc_sigmask: ::sigset_t,
                /* Android has a wrong (smaller) sigset_t on x86. */
                __padding_rt_sigset: u32,
            }

            pub union uc_sigmask {
                uc_sigmask: uc_sigmask_with_padding,
                uc_sigmask64: ::sigset64_t,
            }
        }
```

But this was rejected by `libc-test` (Sorry I can't retrieve the log from CI).

As far as I know this isn't possible in current Rust.
I guessed that is why there are several comments such as `// FIXME this is actually a union` in this repository.

src/unix/linux_like/android/b32/x86/mod.rs
igrep: Same here: https://github.com/rust-lang/libc/pull/1622#discussion_r361042076

src/unix/linux_like/android/b64/x86_64/mod.rs
igrep: Same here https://github.com/rust-lang/libc/pull/1622#discussion_r361042076

src/unix/linux_like/android/b64/mod.rs
igrep: Oh, okay. I'm not actually sure how those instances should be implemented. So I'll move to `s_no_extra_traits!`.

src/unix/linux_like/android/b32/arm.rs
gnzlbg: > But I got this error:

That just means that you need to group the `type = ...` with all other `type = ...` at the beginning of the file.

gnzlbg: That error could be arguably much better, sorry about that 😆

src/unix/linux_like/android/b32/arm.rs
gnzlbg: > At first I made a separate types to represent the anonymous unions

That's how you do it.

> But this was rejected by libc-test (Sorry I can't retrieve the log from CI).

You probably just need to skip the union in the `libc-test/build.rs`, since it does not exist as a named union in C.

src/unix/linux_like/android/b64/mod.rs
gnzlbg: Yes, I don't see a way to meaningful implement much for this type either, so don't worry about that.

gnzlbg:
Thank you for fixing all of these issues so quickly. I'll be a bit unavailable over christmas, but I'll try to take a look every now and then if you make progress with the types using an `union`. 

Basically it' what you originally suggested: you need to create a named union, and then use it as a field. However, there is no union with that name in the C file, so you need to skip trying to generate tests for that in the `libc-test/build.rs` (e.g. by adding a match arm like: `cfg.skip_struct(|name, is_struct, is_union| match name { ..., "your_union_name" if is_union => true, }` ).  This might not suffice, since then the struct has a field of the form `field: your_union_name` and there is no type `your_union_name` in C either, so we need to skip testing this struct field as well (there are examples in the `libc-test/build.rs` about how to do this). 

I think it would be nice to have a way to test fields of anonymous unions by mapping them to some Rust union field in the `ctest` crate, but we don't have that functionality yet - so it's ok to skip these tests for now.

src/unix/linux_like/android/b32/arm.rs
```diff
@@ -1,5 +1,167 @@
...
+            pub struct ucontext_t {
+                pub uc_flags: ::c_ulong,
+                pub uc_link: *mut ucontext_t,
+                pub uc_stack: ::stack_t,
+                pub uc_mcontext: mcontext_t,
+                pub uc_sigmask__c_anonymous_union: __c_anonymous_uc_sigmask,
+                /* The kernel adds extra padding after uc_sigmask to match
+                 * glibc sigset_t on ARM. */
+                __padding: [c_char; 120],
+                __align: [::c_longlong; 0],
```
igrep: Oh, I forgot to ask about alignment.
In `ucontext_t` for Android ARM32, `uc_regspace` is defined as follows:
(from https://android.googlesource.com/platform/bionic/+/refs/tags/ndk-r20/libc/include/sys/ucontext.h#100)

```c
typedef struct ucontext {
  ...
  unsigned long uc_regspace[128] __attribute__((__aligned__(8)));
} ucontext_t;
```

To specify the alignment, I defined as follows at first:

```rust
pub struct ucontext_t {
    ...
    #[repr(align(8))]
    uc_regspace: [::c_ulong; 128],
}
```

(from https://github.com/rust-lang/libc/pull/1622/commits/546a690047f01e85dd50f5b0967fa03be5294a3b#diff-8c71e51ad9afb09c731c4ec6ac99ed44R45)

But this is invalid as the style checker warns.

So I [defined the separate type for `uc_regspace`](https://github.com/rust-lang/libc/pull/1622/commits/37d06c76a55c4b723dd2c1fb4364cb87de8f7b2b), but that was rejected by the test because the separate type doesn't exist in C, and I didn't know the way to ignore in `libc-test/build.rs`.
But now you taught me the how, now I have several questions:

- (1) In the first place, are this line (`__align: [::c_longlong; 0]`) and [this commit](https://github.com/rust-lang/libc/pull/1622/commits/37d06c76a55c4b723dd2c1fb4364cb87de8f7b2b) the right way to set alignment for `uc_regspace: [::c_ulong; 128]`? I'm new to memory alignment...
- (2) Should I recreate the new separate type for `uc_regspace` in a `if #[cfg(align)]` block?
- (3) If yes, should I support for older Rusts like `src/unix/uclibc/x86_64/align.rs` and `src/unix/uclibc/x86_64/no_align.rs`?

igrep: @gnzlbg ping.

gnzlbg: > In the first place, are this line (__align: [::c_longlong; 0]) and this commit the right way to set alignment for uc_regspace: [::c_ulong; 128]? I'm new to memory alignment...

I think that's the simplest way to achieve that, simpler than by defining a new type, so if that works, I'd prefer this over defining a different type with the `repr(align(N))` attribute, since it avoids the need for workarounds.

igrep: > I think that's the simplest way to achieve that, simpler than by defining a new type

Um... 😕  I asked you because the first way (attaching `#[repr(align(8))]` to the field of `struct`) doesn't seem to work, as the style checker warns that `#[repr(align(8))]` is unused.

igrep: Ah, okay, sorry I got it wrong. So I leave this code as is.

igrep:
@gnzlbg Can you review again and merge?
The both of the last failures in CI are not related to my change (`i686-apple-darwin` and `armv7-apple-ios`)

https://dev.azure.com/rust-lang2/libc/_build/results?buildId=1075&view=logs&j=c7464085-50dc-55b0-d49a-ca2fc4e4d69e&t=7417e154-0538-5077-3a89-29444dec6e1c

https://dev.azure.com/rust-lang2/libc/_build/results?buildId=1075&view=logs&j=a0d58cc1-afb1-556c-24c8-109083aad81d&t=3a3b8099-ea38-5855-a2d4-9817ff96dcb5

igrep:
@gnzlbg ping.

igrep:
@JohnTitor Can I ask you to review this instead of @gnzlbg?
This pull request has been stalled since I've fixed all problems pointed out...

src/unix/linux_like/android/b32/arm.rs
```diff
@@ -1,5 +1,167 @@
...
+                            .field("uc_mcontext", &self.uc_mcontext)
+                            .field(
+                                "uc_sigmask__c_anonymous_union",
+                                &self.uc_sigmask__c_anonymous_union
+                            )
+                            // &self.uc_regspace[..] doesn't work due
+                            //   to error[E0277]:
+                            //   the size for values of type `[u32]` cannot be
+                            //   known at compilation time
+                            .field("uc_regspace", &Array128(self.uc_regspace))
```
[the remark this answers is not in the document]
igrep: It worked! thanks!

Issue #1572, opened 2019-10-28
[API request] Expose dlmopen and dlinfo on linux gnu 


afinch7:
[dlmopen](http://man7.org/linux/man-pages/man3/dlopen.3.html)
[dlinfo](http://man7.org/linux/man-pages/man3/dlinfo.3.html)

I already have a basic working example for this that compiles and runs on 
1.38.0 `x86_64-unknown-linux-gnu`:
``` rust
use std::ffi::{CStr, OsStr};
use libc::{c_int, c_long, c_char, c_void, dlopen, RTLD_LOCAL, RTLD_LAZY};
use std::os::unix::ffi::OsStrExt;

pub const RTLD_DI_LMID: c_int = 0x1;
pub const LM_ID_NEWLM: c_long = -1;

type Lmid_t = c_long;

extern "C" {
    pub fn dlmopen(lmid: Lmid_t, filename: *const c_char, flag: c_int) -> *mut c_void;
    pub fn dlinfo(handle: *mut c_void, request: c_int, info: *mut c_void) -> c_int;
}

const DEFAULT_FLAGS: c_int = RTLD_LOCAL | RTLD_LAZY;

// This code was borrowed and modified from https://github.com/szymonwieloch/rust-dlopen/blob/master/src/raw/unix.rs
pub unsafe fn open_lib<S>(lmid: Lmid_t, name: S) -> *mut c_void
where
    S: AsRef<OsStr> {
    let mut v: Vec<u8> = Vec::new();
    let name = name.as_ref();
    //as_bytes i a unix-specific extension
    let cstr = if name.len() > 0 && name.as_bytes()[name.len() - 1] == 0 {
        //don't need to convert
        CStr::from_bytes_with_nul_unchecked(name.as_bytes())
    } else {
        //need to convert
        v.extend_from_slice(name.as_bytes());
        v.push(0);
        CStr::from_bytes_with_nul_unchecked(v.as_slice())
    };
    dlmopen(lmid, cstr.as_ptr(), DEFAULT_FLAGS)
}

fn main() {
    let self_handle = unsafe { dlopen(std::ptr::null(), DEFAULT_FLAGS) };
    let mut lmid_self: Lmid_t = unsafe { std::mem::uninitialized() };
    let result = unsafe { dlinfo(self_handle, RTLD_DI_LMID, &mut lmid_self as *mut _ as *mut c_void) };
    assert_eq!(result, 0);

[...]

PR #1624, opened 2019-12-15
Add dlmopen dlinfo and related consts.

noam93k:
I actually have no idea if the constants I copied from my computer are cross-platform.


noam93k:
Closes #1572

PR #1625, opened 2019-12-16
Update bindings for the wasm32-wasi target

alexcrichton:
This commit performs a number of updates for libc with the `wasm32-wasi`
target:

* Updates the `wasi-libc` repository commit used (previously known as
  `wasi-sysroot`)
* Updates the container to Ubuntu 19.10 which has Clang 9 packaged which
  is all we need.
* Avoids building `wasmtime` and instead downloads a precompiled binary.
* Updates bindings in `src/wasi.rs` to match the current upstream state.


gnzlbg:
@alexcrichton this LGTM - feel free to merge when you think is ready. Keep in mind that the `wasi` build job is disabled: https://github.com/rust-lang/libc/blob/master/ci/azure.yml#L75

alexcrichton:
👍

PR #1626, opened 2019-12-16
Signal controlling constants for windows

### LINKED ISSUE #1600: SIG_DFL and SIG_IGN on windows
Hello

The `SIG_DFL` and `SIG_IGN` constants seem to be missing on windows targets (`*-pc-windows-*` tripples). It seems that according to the documentation they should exist

https://docs.microsoft.com/en-us/previous-versions/xdkz3x12(v=vs.140)

I'd try adding them, but I don't have any Windows at hand so I can't find the value. Some tests on CI suggest that `SIG_DFL` might be 0, though, but I can't really be sure.

Is it possible to add them?

Thank you

vorner:
SIG_DFL and similar.

Closes #1600.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -465,7 +465,9 @@ fn test_windows(target: &str) {
         match name {
             // FIXME: API error:
             // SIG_ERR type is "void (*)(int)", not "int"
-            "SIG_ERR" => true,
+            "SIG_ERR" |
+            // Similar for SIG_DFL/IGN/GET/SGE/ACK
+            "SIG_DFL" | "SIG_IGN" | "SIG_GET" | "SIG_SGE" | "SIG_ACK" => true,
             _ => false,
         }
     });
--- a/src/windows/mod.rs
+++ b/src/windows/mod.rs
@@ -233,7 +233,13 @@ pub const SIGSEGV: ::c_int = 11;
 pub const SIGTERM: ::c_int = 15;
 pub const SIGABRT: ::c_int = 22;
 pub const NSIG: ::c_int = 23;
+
 pub const SIG_ERR: ::c_int = -1;
+pub const SIG_DFL: ::sighandler_t = 0;
+pub const SIG_IGN: ::sighandler_t = 1;
+pub const SIG_GET: ::sighandler_t = 2;
+pub const SIG_SGE: ::sighandler_t = 3;
+pub const SIG_ACK: ::sighandler_t = 4;
 
 // inline comment below appeases style checker
 #[cfg(all(target_env = "msvc", feature = "rustc-dep-of-std"))] // " if "


libc-test/build.rs
```diff
@@ -465,7 +465,9 @@ fn test_windows(target: &str) {
         match name {
             // FIXME: API error:
             // SIG_ERR type is "void (*)(int)", not "int"
-            "SIG_ERR" => true,
+            "SIG_ERR" |
+            // Similar for SIG_DFL/IGN/GET/SGE/ACK
```
gnzlbg: Can we add them with the right type ?

src/windows/mod.rs
```diff
@@ -233,7 +233,13 @@ pub const SIGSEGV: ::c_int = 11;
 pub const SIGTERM: ::c_int = 15;
 pub const SIGABRT: ::c_int = 22;
 pub const NSIG: ::c_int = 23;
+
 pub const SIG_ERR: ::c_int = -1;
```
retep998: `#define SIG_ERR ((_crt_signal_t)-1)`

Also `libc` defines `sighandler_t` as `usize` when in reality it is:
`typedef void (__CRTDECL* _crt_signal_t)(int);`

src/windows/mod.rs
```diff
@@ -233,7 +233,13 @@ pub const SIGSEGV: ::c_int = 11;
...
+pub const SIG_DFL: ::sighandler_t = 0;
+pub const SIG_IGN: ::sighandler_t = 1;
+pub const SIG_GET: ::sighandler_t = 2;
+pub const SIG_SGE: ::sighandler_t = 3;
+pub const SIG_ACK: ::sighandler_t = 4;
```
retep998: These values are correct.

src/windows/mod.rs
gnzlbg: > Also libc defines sighandler_t as usize when in reality it is:
typedef void (__CRTDECL* _crt_signal_t)(int);

I'll take a PR to fix this, but I'll prefer if this change were in a different PR, and not here. We probably need to do a crater run while rolling in this change.

retep998: Just keep in mind we can't naively use `extern "C" fn(c_int)` as the type because  `SIG_DFL` has a value of 0.

gnzlbg: Yes, we need `Option<extern "C" fn(c_int) -> ()>` or similar (I'm not sure what to do about the `__CRTDECL` or whether `extern "system"` or some other ABI has to be used for windows.

retep998: `__CRTDECL` is `__cdecl` which is `extern "cdecl"` or `extern "C"`. Most of the CRT uses `__CRTDECL`.

vorner:
Hello

To be honest, I'm a bit lost here what the next steps from me should be. I can go and send a pull request, changing the type. However, there are few issues.

* I don't have an easy access to windows machine, so I won't be able to do even basic testing.
* The `Option` solves the problem of `0`. Is it however OK to assign values like `3` to it? Doesn't it mandate the „reference“ to function is valid?

gnzlbg:
> The Option solves the problem of 0. Is it however OK to assign values like 3 to it? Doesn't it mandate the „reference“ to function is valid?

cc @rust-lang/wg-unsafe-code-guidelines

retep998:
If only Rust had raw function pointers.

RalfJung:
Function pointers have to be non-null and aligned. While it seems unlikely that we'll ever have a platform where the alignment requirement is non-trivial, that's also not excluded currently. Cc https://github.com/rust-lang/unsafe-code-guidelines/pull/197

vorner:
Reading the code and considering this is for Windows platform only, which has quite limited support for HW platforms it runs on, I think this should be safe then. I'll prepare the other PR soon.

retep998:
Probably saner to use a raw pointer to an uninhabited type, since it has the same ABI as function pointers but we don't actually care about it being a *callable* function pointer.

vorner:
But then it wouldn't really match the *correct* type… and then, is it worth changing at all, instead of staying as `usize`? I mean, there'll be some work and churn and crater runs… and if the result will be a different type that doesn't match the reality, then it seems like little benefit.

gnzlbg:
How does this work in C? 

IIRC, C requires function pointers to be properly aligned as well, at least when you call them, right ?

> But then it wouldn't really match the correct type… and then, is it worth changing at all, instead of staying as usize? 

I think we should move the discussion about how to "fix" the function pointer type to a different issue. This PR is not the place to do this (CI is broken right now, but once it works again, I'll merge this).

nikomatsakis:
This is ready to merge, right, apart from rebase?

vorner:
I think so…

> CI is broken right now, but once it works again, I'll merge this

Except I must have deleted my fork of the repo in the meantime during some cleanup :-|. I can re-create it, but I'm afraid that'll mean a new PR :-( ‒ I don't expect github UI to handle it reappearing well.

vorner:
Indeed, it's as I expected, github didn't pick it up even if the branch has the same name. It's in #2201 now, with the same content.

And apologies for the confusion :-|

PR #1571, opened 2019-10-28
Nfnetfilter log

vorner:
This adds the constants from linux/netfilter/nfnetlink.h and nfnetlink_log.h. These are the files I need for https://github.com/jbaublitz/neli/pull/48. After this gets in, I'd like to follow-up with the other nfnetlink_*.h files too, as I'd like to extend neli with further protocols in the future, but I want to do a smaller PR first to see if there are some things to tweak.

I've noticed similar netfilter constants are also in the android subfolder, therefore I'm adding them there too (I don't like the copy-pasting, but it seems the other ones are already copy-pasted). I assume the test will catch it if anything is different on that platform.

Issue #1562, opened 2019-10-18
nfnetlink constants


vorner:
Hello

I'd like to ask, would it be OK to add netfilter netlink constants. They are in the linux headers (`/usr/include/linux/netfilter/nfnetlink*.h`). I'm trying to extend neli with nfnetlink support (currently NFLOG, https://github.com/jbaublitz/neli/pull/48).

I'll send a pull request, I just wanted to ask first if including these would be OK.

gnzlbg:
Sure.

JohnTitor:
I assume this is done by #1628 and am going to close but feel free to re-open if I'm missing something.

vorner:
I think there are some few more constants in the header, but I guess a lazy approach (once they are actually needed) is good and there's not much use for keeping this issue open :-). Sorry for letting it rot in here.

PR #1628, opened 2019-12-28
Nfnetfilter queue constants

vorner:
Constants for the linux nfqueue netlink protocol (userspace firewall).

Continuation of #1562 and #1571.


vorner:
I somehow missed the build failed and now it seems to be already deleted so I can't have a look why :-(.

Is there a way I can re-run it?

JohnTitor:
Hey @vorner could you rebase to make CI happy?
r? @JohnTitor

vorner:
Sure, that makes sense. Even if the failure would be my fault, I'd at least get the logs.

vorner:
So it seems the CI failures are of the same nature as the ones in #1571 ‒ some of the added constants are too new and are not available in the musl sanitized kernel headers yet. I think the options are:

* Remove these from the pull request (I've included them for completeness, but I don't need these specific ones).
* Keep them, but somehow omit them from the test on the musl-based targets (I'm not sure how exactly to do that, but I think I'd find a way by studying already existing code). As these are kernel-constants, checking them on glibc should give enough confidence that they are correct.

What would you prefer me to do?

JohnTitor:
> Remove these from the pull request

I'd prefer this if you don't need them. But could you not remove them but comment out and add FIXME comment so that we don't forget them?

vorner:
I've added the change as a fixup so you can review the changes incrementally. But let me squash it before merging.

JohnTitor:
Looks good! CI failure is unrelated to this and fixed in master, could you rebase again?

vorner:
Thanks. I've also done the squashing.

vorner:
OK, fixed, thanks. It's a bit unusual to find a rust project with 80-column policy, the majority is 120.

vorner:
Hmm. This failure seems to be a network issue when setting things up. Is there a way for me to re-run it?

PR #1414, opened 2019-06-28
[Draft] Add __int128 to unix targets

gnzlbg:
This PR is for testing purposes only for now.

AFAICT, the ABI is wrong (AGGREGATE due to repr(C) vs SCALAR for __int128), but on my laptop ctest does not report any errors.

PR #1630, opened 2019-12-30
FreeBSD: ucontext/mcontext, aarch64 fixes, librt linking, etc

myfreeweb:
- adds mcontext/ucontext structs, needed for lucet
  - on amd64 and aarch64 because that's what I have… (well I technically have armv7 but it's not set up properly)
- fixes all tests on aarch64 (because I wanted to test mcontext)
  - https://github.com/gnzlbg/ctest/pull/89 necessary to handle u128
    - I haven't noticed #1414… hmm?? is that the better approach?


myfreeweb:
ping. this is really needed for Lucet, which is in pretty good shape now.

we could leave FreeBSD/aarch64 untested because the tests currently don't work in master anyway. or I could split the aarch64 mcontext into a separate PR that would wait for proper 128 bit support in tests or whatever is needed.

JohnTitor:
Sorry for the delay, I'll check it later.
r? @JohnTitor

JohnTitor:
Hm, CI fails: nightly x86_64-unknown-freebsd-10. Also needs to fix style issues.

myfreeweb:
> CI fails: nightly x86_64-unknown-freebsd-10

Added `ucontext.h` to libc-test headers. (On newer versions it was picked up transitively).

Now docs for i686 fail to generate??

> cannot find type `ucontext_t` in this scope

but.. I see there's plenty of `target_arch` usage and I think this particular CI didn't fail on the previous push o_0

myfreeweb:
- Fixed style
- Moved the `cfg` outside of the `s_no_extra_traits!`, now docs don't fail
- Added new fix: `aio_*` and `mq_*` functions need linking to `librt`! How did no one notice this o_0

Now all CI failures seem unrelated

myfreeweb:
ping

JohnTitor:
Oh, sorry for the delay! Unfortunately, the build log has been gone, re-triggering CI.

JohnTitor:
> or I could split the aarch64 mcontext into a separate PR that would wait for proper 128 bit support in tests or whatever is needed.

I'd go this way if you don't need it right now.
Other stuff looks good, I think it's ready to merge once we resolve the above and the style issue that CI reported.

myfreeweb:
- dropped aarch64 mcontext
- rebased
- fixed style
- added changes to make tests pass on -CURRENT (hello new `rand()` ABI LOL)

src/unix/bsd/freebsdlike/freebsd/x86_64/mod.rs
```diff
@@ -78,6 +80,48 @@ s_no_extra_traits! {
         pub xmm_reg: [[u8; 16]; 8],
         pub xmm_pad: [u8; 224],
     }
+
+    #[repr(align(16))]
```
JohnTitor, line 84: This should be in `align.rs`, see what other targets do.

libc-test/build.rs
```diff
@@ -1809,6 +1810,22 @@ fn test_freebsd(target: &str) {
...
+            "CTL_SYSCTL_DEBUG"
+            | "CTL_SYSCTL_NAME"
+            | "CTL_SYSCTL_NEXT"
+            | "CTL_SYSCTL_NAME2OID"
+            | "CTL_SYSCTL_OIDFMT"
+            | "CTL_SYSCTL_OIDDESCR"
+            | "CTL_SYSCTL_OIDLABEL" => true,
+
+            // This was renamed in FreeBSD 12 and 13 (r350749).
+            "IPPROTO_SEP" | "IPPROTO_DCCP" => true,
```
JohnTitor, line 1827: I think we could use `freebsd_ver`.

JohnTitor:
Do you want us to ship this ASAP? If so, I'll hold #1807 off for a while, until this gets merged.

libc-test/build.rs
myfreeweb: Not really: this would change between minor versions (12.1 and the upcoming 12.2)

myfreeweb:
> Do you want us to ship this ASAP?

Yes, that would be very good.

libc-test/build.rs
JohnTitor: Ah, I see. Then it'd be great if the comment mentions that.

JohnTitor:
CI still fails with importing and `Copy` trait issues.

myfreeweb:
That looks like some really ancient rustcs not being able to apply `cfg` to a macro invocation >_<

I guess I can move the whole `ucontext_t` to `x86_64` for now, and deal with making it multi platform later..

src/unix/bsd/freebsdlike/freebsd/x86_64/mod.rs
```diff
@@ -78,6 +80,16 @@ s_no_extra_traits! {
         pub xmm_reg: [[u8; 16]; 8],
         pub xmm_pad: [u8; 224],
     }
+
+    #[cfg_attr(feature = "extra_traits", derive(Debug))]
```
JohnTitor, line 84: This is unnecessary since `s!` macro already derives it.

src/unix/bsd/freebsdlike/freebsd/x86_64/mod.rs
```diff
@@ -78,6 +80,16 @@ s_no_extra_traits! {
...
+    pub struct ucontext_t {
+        pub uc_sigmask: ::sigset_t,
+        pub uc_mcontext: ::mcontext_t,
```
JohnTitor, line 87: This causes importing failure:
```
error[E0412]: cannot find type `mcontext_t` in the crate root
  --> src/unix/bsd/freebsdlike/freebsd/x86_64/mod.rs:87:28
   |
87 |         pub uc_mcontext: ::mcontext_t,
   |                            ^^^^^^^^^^ did you mean `ucontext_t`?
```

src/unix/bsd/freebsdlike/freebsd/x86_64/mod.rs
myfreeweb: No, this is `s_no_extra_traits`!

I guess since we implement everything on `mcontext`, `ucontext` could go into `s`..

src/unix/bsd/freebsdlike/freebsd/x86_64/mod.rs
myfreeweb: I guess these ancient rustcs were single pass and the `pub use self::align::*` at the bottom doesn't bring it into this scope?!

Screw it, moving `ucontext_t` into the align module too.

JohnTitor:
Okay, now looks fine to me!
Remaining CI failures are spurious and #1812 fixed one of them, could you rebase?

myfreeweb:
Done!

asomers:
For the case of aio_ functions, nobody noticed the lack of librt before because it _isn't_ required.  Linking to librt is only required if you wish to use `SIGEV_THREAD` completion notifications. Everything else requires only libc.  That requirement is documented in the sigevent(3) man page.

Using `SIGEV_THREAD` is very rare in practice, because the performance is not so good.  In fact, due to Rust's originally poor support for C unions, libc's definition for `struct sigevent` doesn't even allow you to set the fields needed by `SIGEV_THREAD`!  So we shouldn't link those functions to librt.

Your change was of course correct as regards mq_ functions.

PR #1639, opened 2020-01-06
Add memmem

vickenty:
memmem is a non-standard extension, first added in GNU libc and later
ported to other systems. Support for it is non-uniform, thus it was only
added to platforms that seem to support it.


JohnTitor:
Thanks!

PR #1646, opened 2020-01-24
Update mcontext_t definition on musl, add fpstate struct

MarkMcCaskey:
Hello!  This PR updates definitions for x86_64 in musl to match the GNU ones and the current musl headers:

### Relevant links for review:
- [fpstate in musl header](https://git.musl-libc.org/cgit/musl/tree/arch/x86_64/bits/signal.h#n60)
- [mcontext in musl header](https://git.musl-libc.org/cgit/musl/tree/arch/x86_64/bits/signal.h#n80)
- [fpstate in Rust libc on GNU](https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/linux/gnu/b64/x86_64/mod.rs#L163-L175)
- [mcontext in Rust libc on GNU](https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/linux/gnu/b64/x86_64/mod.rs#L229-L233)

### DIFF
--- a/src/unix/linux_like/linux/musl/b64/x86_64/mod.rs
+++ b/src/unix/linux_like/linux/musl/b64/x86_64/mod.rs
@@ -66,6 +66,30 @@ s! {
         pub nla_type: u16,
     }
 
+    pub struct _libc_fpxreg {
+        pub significand: [u16; 4],
+        pub exponent: u16,
+        __private: [u16; 3],
+    }
+
+    pub struct _libc_xmmreg {
+        pub element: [u32; 4],
+    }
+
+    pub struct _libc_fpstate {
+        pub cwd: u16,
+        pub swd: u16,
+        pub ftw: u16,
+        pub fop: u16,
+        pub rip: u64,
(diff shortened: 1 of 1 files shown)


MarkMcCaskey:
Hmm, I'm not sure why the tests are failing,

also the comment on `mcontext` on the musl side,
```
    // GitHub repo: ifduyue/musl/
    // commit: b4b1e10364c8737a632be61582e05a8d3acf5690
    // file: arch/x86_64/bits/signal.h#L80-L84
```

is probably out of date now.  Is that comment being parsed and used for testing? 🤔

JohnTitor:
@MarkMcCaskey Could you rebase to check CI?
r? @JohnTitor

JohnTitor:
I'm not sure if we could apply the same way as GNU since musl does a different way. At least, I think `_libc_fpstate` should be `fpregset_t` here.

JohnTitor:
I'm going to close this as inactive. Feel free to re-open if you have some time to resolve the above comment, thanks!

PR #1647, opened 2020-01-25
Fix CI

vickenty:
`i686-apple-darwin` was demoted to tier-3 platform due to lack of support from the vendor,
and is no longer available from rustup for nightly, shortly beta.

This PR disables demoted target for nightly and beta, and in 7 weeks, before 1.42 release, stable should be gated as well.

Upd: I also included small patches that should also take care of other failures. One caused by a new warning in nightly breaking due `#[deny(warnings)]`. Second due to native `find` being used on windows instead of a gnu version.


JohnTitor:
Sorry for the delay! Overall looks good, the remaining thing is to make sure builds with demoted targets, I think. Could you also add `RUST_OSX_NO_CORE_TARGETS` like `RUST_LINUX_NO_CORE_TARGETS`?

r? @JohnTitor

PR #1649, opened 2020-01-29
structs nlmsghdr, nlmsgerr and nlattr added to linux mips musl target

rfael:


xen0n:
The additions look good; while you're at it, would you mind adding the structs for the `mips64` targets too? This way things are more consistent.

rfael:
I added that struct to mips64 musl target too

rfael:
I filled x86 and arm musl targets too

JohnTitor:
Thanks!

PR #1652, opened 2020-02-02
Introducing libcap

Isawan:
libcap is the stable library for interacting with linux capabilities. The upstream source can be found here.

https://git.kernel.org/pub/scm/linux/kernel/git/morgan/libcap.git

### DIFF
--- a/ci/docker/aarch64-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/aarch64-unknown-linux-gnu/Dockerfile
@@ -1,6 +1,6 @@
 FROM ubuntu:19.10
 RUN apt-get update && apt-get install -y --no-install-recommends \
-  gcc libc6-dev ca-certificates \
+  gcc libc6-dev libcap-dev ca-certificates \
   gcc-aarch64-linux-gnu libc6-dev-arm64-cross qemu-user
 ENV CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc \
     CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUNNER="qemu-aarch64 -L /usr/aarch64-linux-gnu" \
--- a/ci/docker/i686-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/i686-unknown-linux-gnu/Dockerfile
@@ -1,5 +1,5 @@
 FROM ubuntu:19.10
 RUN apt-get update
 RUN apt-get install -y --no-install-recommends \
-  gcc-multilib libc6-dev ca-certificates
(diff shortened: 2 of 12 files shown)


Isawan:
@gnzlbg @malbarbo hi. I've ported this feature onto x86_64-gnu with successful tests. Before I continue porting to the other architectures; I just wanted to check with you that the implementation of libcap is within the scope of this crate. libcap is the only stable and recommended way of interacting with linux capabilities as documented in the manpage `man capget(2)`.

>  Since  Linux  2.2, the power of the superuser (root) has been partitioned into a set of discrete capabilities.  Each thread has a set of effective capabilities identifying which capabilities (if any)  it may  currently  exercise.   Each  thread also has a set of inheritable capabilities that may be passed through an execve(2) call, and a set of permitted capabilities that it can make effective or inheritable.

> These two system calls are the raw kernel interface for getting and setting thread capabilities.  Not only are these system calls specific to Linux, but the kernel API is likely to change and use of these system  calls  (in  particular the format of the cap_user_*_t types) is subject to extension with each kernel revision, but old programs will keep working.

> The portable interfaces are cap_set_proc(3) and cap_get_proc(3); if possible,  you  should  use  those interfaces  in  applications.  If you wish to use the Linux extensions in applications, you should use the easier-to-use interfaces capsetp(3) and capgetp(3).

I ask this as on many linux distributions, the development libraries are generally required to be installed seperately from the libc-dev, for example on ubuntu the package libcap-dev is required to be installed. This will require some modifications to the CI for cross compilation.

alexcrichton:
Thanks for the PR here, but [right now there's not many maintainers to help merge this](https://github.com/rust-lang/libc/issues/1657), so it may take awhile.

JohnTitor:
@Isawan could you rebase to check current CI status?
r? @JohnTitor

Isawan:
Hi, a rebase has been carried out.

joshtriplett:
@Isawan 
> I've ported this feature onto x86_64-gnu with successful tests. Before I continue porting to the other architectures; I just wanted to check with you that the implementation of libcap is within the scope of this crate.
...
> I ask this as on many linux distributions, the development libraries are generally required to be installed seperately from the libc-dev, for example on ubuntu the package libcap-dev is required to be installed.

I personally don't think this is something that needs to happen under the auspices of the libc crate, for a combination of reasons:
- It requires a library other than libc.
- That library doesn't normally ship with libc.
- In practice, it's Linux-specific. This alone isn't a problem, there are plenty of Linux-specific things in libc, but rather, it means that there's no portable interface we're trying to abstract over here
- As a result of all of those things, there doesn't seem to be any advantage to having this in libc compared to a separate crate.

I think it would make sense to put this interface in a `libcap-sys` package, and then provide more Rustic bindings via a `cap` or `libcap` crate.

PR #1658, opened 2020-02-11
Unify mmap flags

connorkuehl:
`MAP_SHARED_VALIDATE` and `MAP_FIXED_NOREPLACE` can be exported for both gnu and musl.

Add `MAP_SYNC`.

I checked these values for both gnu and musl and they are the same.

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -2310,6 +2310,7 @@ pub const MAP_SHARED_VALIDATE: ::c_int = 0x3;
 // include/uapi/asm-generic/mman-common.h
 #[cfg(not(any(target_arch = "mips", target_arch = "mips64")))]
 pub const MAP_SYNC : ::c_int = 0x080000;
+pub const MAP_FIXED_NOREPLACE: ::c_int = 0x100000;
 
 // uapi/linux/vm_sockets.h
 pub const VMADDR_CID_ANY: ::c_uint = 0xFFFFFFFF;
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -2307,6 +2307,10 @@ pub const ALG_OP_ENCRYPT: ::c_int = 1;
 // include/uapi/linux/mman.h
 pub const MAP_SHARED_VALIDATE: ::c_int = 0x3;
 
+// include/uapi/asm-generic/mman-common.h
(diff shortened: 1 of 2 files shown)


JohnTitor:
@connorkuehl Could you rebase to check current CI status?
r? @JohnTitor

connorkuehl:
Thanks @JohnTitor! The CI's been cleaned up a lot since I last checked.

It looks like I still have some work to do on this PR. I'll leave a note once I've resolved this.

src/unix/linux_like/linux/mod.rs
```diff
@@ -2304,6 +2304,14 @@ pub const ALG_SET_AEAD_AUTHSIZE: ::c_int = 5;
 pub const ALG_OP_DECRYPT: ::c_int = 0;
 pub const ALG_OP_ENCRYPT: ::c_int = 1;
 
+// include/uapi/linux/mman.h
+pub const MAP_SHARED_VALIDATE: ::c_int = 0x3;
+
+// include/uapi/asm-generic/mman-common.h
+#[cfg(not(any(target_arch = "mips", target_arch = "mips64")))]
+pub const MAP_SYNC : ::c_int = 0x080000;
```
connorkuehl: @JohnTitor It looks like C code generation was failing on mips, mips64. I'm not sure if MAP_SYNC is used or defined for mips. Is this an acceptable way of turning it off?

connorkuehl:
@JohnTitor I think I've resolved the CI issues with my branch. The only failure I see in the CI looks unrelated to this change.

src/unix/linux_like/linux/mod.rs
JohnTitor: I'll take a look but it should be fine if it isn't defined.

JohnTitor:
> I think I've resolved the CI issues with my branch. The only failure I see in the CI looks unrelated to this change.

Yeah, Rust 1.42.0 that demotes apple 32-bit targets is released so we should tweak CI, one sec..

connorkuehl:
@JohnTitor It's fixed! Thank you. Please let me know if you'd like me to make any changes to this PR.

src/unix/linux_like/linux/mod.rs
JohnTitor: I investigated a bit, perhaps we need CI tweaks for mips64. I'll investigate more but it's out of scope here. I'll accept it at this point and follow-up later.

PR #1659, opened 2020-02-12
Appease the angry linter which causes many CI tasks to fail

connorkuehl:
This is a draft PR while I use the CI as a feedback loop. It appears a large number of pull requests to this repo are not passing CI due to checked-in code failing the linting process (and possibly other reasons).


connorkuehl:
Oops, sorry, I missed #1647 when I was looking for work done in this area.

PR #1661, opened 2020-02-16
Move linux/reboot.h constants into linux module

Richard-W:
The `LINUX_REBOOT_*` constants were defined in the gnu module although they are just as valid for musl (and any other linux libc). The `reboot` function was already in the right module although it does not work without these constants.


JohnTitor:
@Richard-W Could you rebase to check current CI status?
r? @JohnTitor

Richard-W:
Done. CI passes.

PR #1662, opened 2020-02-16
Add ioctl generation macros for linux

sm-Fifteen:
The `_IOC`, `_IO`, `_IOR`, `_IOW` and `_IOWR` macros, which are exposed by both musl (which redefines them per architecture) and glibc (which includes them from the linux headers) don't currently have equivalents in Rust's libc. Given [there are a lot of ioctls out there](https://www.kernel.org/doc/Documentation/ioctl/ioctl-number.txt) and that [the format is known to be architecture-specific](https://www.kernel.org/doc/html/latest/userspace-api/ioctl/ioctl-decoding.html), not having access to those macros to generate ioctl request numbers means having to duplicate that platform-specific logic elsewhere [like the `nix` crate currently does](https://github.com/nix-rust/nix/blob/master/src/sys/ioctl/linux.rs). Moving these into the libc crate alongside all of the other musl/glibc would (as far as I can tell) be more in line with the goals of this crate.

With that said, I'm currently having trouble figuring out how those should be tested. Most of the ioctl codes currently defined in libc follow the short 16 bits format, and the few of them using the longer format are architecture-specific redefinitions, [like `TIOCGPGRP`](https://github.com/torvalds/linux/search?q=TIOCGPGRP+filename%3Aioctls.h&unscoped_q=TIOCGPGRP+filename%3Aioctls.h) which is is defined as `0x540F` in asm-generic but as `_IOR('t', 119, int)` for mips and powerpc and as `_IOR('t', 131, int)` for sparc, so testing by ensuring that the manual ioctl definitions match the macro's output most likely wouldn't work. We could always use a list of driver-specific IOCTLs with known type letters and sequence numbers and testing that those match with the kernel headers ones, but I'm not familiar enough with the way the tests are setup to get that to work.

Another thing I'm not sure how to do is declaring those as const_fn, especially since I would need `size_of` to be const_fn as well for that to work. I know there is a `libc_const_size_of` config switch for that and I have noticed that macros are available for this as well, but the macros are undocumented and thier usage is unclear, so I've left those as plain functions for now.

The reason why the definitions don't follow the same format between musl and glibc is that I was attempting to mimic how [musl defines its own macros per architecture](https://git.musl-libc.org/cgit/musl/tree/arch/generic/bits/ioctl.h) while [glibc imports them from the kernel header files](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/bits/ioctls.h) where platform-specific constants [tend to be defined in the platform-specific headers](https://github.com/torvalds/linux/blob/master/arch/powerpc/include/uapi/asm/ioctl.h) which then [re-use the generic macro](https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/ioctl.h). I'm not entirely sure whether this is preferable to sticking to one way or the other.

I'm open to suggestions on what the best way of handling this is, or if someone thinks this may not actually belong in libc.

### DIFF
--- a/src/unix/linux_like/linux/gnu/b32/arm/mod.rs
+++ b/src/unix/linux_like/linux/gnu/b32/arm/mod.rs
@@ -868,6 +868,16 @@ pub const SYS_pkey_alloc: ::c_long = 395;
 pub const SYS_pkey_free: ::c_long = 396;
 pub const SYS_statx: ::c_long = 397;
 
+// from uapi/asm-generic
+pub const _IOC_NRBITS: u32 = 8;
+pub const _IOC_TYPEBITS: u32 = 8;
+pub const _IOC_SIZEBITS: u32 = 14;
+pub const _IOC_DIRBITS: u32 = 2;
+
+pub const _IOC_NONE: ::c_ulong = 0;
+pub const _IOC_READ: ::c_ulong = 1;
+pub const _IOC_WRITE: ::c_ulong = 2;
+
 cfg_if! {
     if #[cfg(libc_align)] {
         mod align;
--- a/src/unix/linux_like/linux/gnu/b32/mips/mod.rs
+++ b/src/unix/linux_like/linux/gnu/b32/mips/mod.rs
@@ -893,6 +893,17 @@ pub const TIOCM_DSR: ::c_int = 0x400;
 
 pub const EHWPOISON: ::c_int = 168;
 
+// from uapi/asm-generic
+pub const _IOC_NRBITS: u32 = 8;
+pub const _IOC_TYPEBITS: u32 = 8;
+// from arch/mips/include/uapi/asm
+pub const _IOC_SIZEBITS: u32 = 13;
+pub const _IOC_DIRBITS: u32 = 3;
+
+pub const _IOC_NONE: ::c_ulong = 1;
+pub const _IOC_READ: ::c_ulong = 2;
+pub const _IOC_WRITE: ::c_ulong = 4;
+
 cfg_if! {
     if #[cfg(libc_align)] {
         mod align;
--- a/src/unix/linux_like/linux/gnu/b32/powerpc.rs
+++ b/src/unix/linux_like/linux/gnu/b32/powerpc.rs
@@ -871,3 +871,14 @@ pub const SYS_preadv2: ::c_long = 380;
 pub const SYS_pwritev2: ::c_long = 381;
 pub const SYS_kexec_file_load: ::c_long = 382;
 pub const SYS_statx: ::c_long = 383;
+
+// from uapi/asm-generic
+pub const _IOC_NRBITS: u32 = 8;
+pub const _IOC_TYPEBITS: u32 = 8;
+// from arch/powerpc/include/uapi/asm
+pub const _IOC_SIZEBITS: u32 = 13;
+pub const _IOC_DIRBITS: u32 = 3;
+
+pub const _IOC_NONE: ::c_ulong = 1;
+pub const _IOC_READ: ::c_ulong = 2;
+pub const _IOC_WRITE: ::c_ulong = 4;
--- a/src/unix/linux_like/linux/gnu/b32/sparc/mod.rs
+++ b/src/unix/linux_like/linux/gnu/b32/sparc/mod.rs
@@ -958,6 +958,17 @@ pub const SYS_preadv2: ::c_long = 358;
 pub const SYS_pwritev2: ::c_long = 359;
 pub const SYS_statx: ::c_long = 360;
 
+// from uapi/asm-generic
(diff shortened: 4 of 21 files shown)


sm-Fifteen:
I think I'm doing something wrong with musl, I'm getting syntax errors from the powerpc module (whoops) in the CI results for other architectures. Could that have something to do with the `f!` macro?

workingjubilee:
It's a pity this languished because it would make the libc code more structured and significantly help with musl support, which has recently been getting more attention (including from er, me, I guess). Needs a rebase though at this point.

sm-Fifteen:
> It's a pity this languished because it would make the libc code more structured and significantly help with musl support, which has recently been getting more attention (including from er, me, I guess). 

Yeah, not having support in the lower level libraries for ioctls of all things is very strange to me, given it's one of the basic file descriptor interactions you can have on Unix platforms. Not having them in libc (which otherwise has all the platform-specific code by design, that's kind of what it's here for) means it has to be implemented and maintained at a higher level, like [the nix crate](https://github.com/nix-rust/nix/blob/master/src/sys/ioctl/linux.rs) or use a bit of rust-bindgen for the ioctl codes and hand-rolled macros for file descriptor interaction, like with [Mozilla's own CTAP HID module](https://github.com/mozilla/authenticator-rs/tree/main/src/linux).

If there's interest, I can try rebasing the conflicting GNU libc files so this is reviewable and mergable.

workingjubilee:
Please do! I can't r+ things myself but I have a stick and I can poke people with it who can. :^)

joshtriplett:
:+1: for including these in libc.

I do think they want to be `const fn` if at all possible.

JohnTitor:
r? @joshtriplett 

@sm-Fifteen Could you rebase onto master or open a new PR if you're still interested in this change?

sm-Fifteen:
> @sm-Fifteen Could you rebase onto master or open a new PR if you're still interested in this change?

@JohnTitor: Right, I'll get to that ASAP. Should I also write some unit tests to make sure that the values generated here match what the respective C header macros generate? It's my first PR here, and given how this is unsafe-adjacent code, that makes me a bit nervous, especially for the architectures like sparc and ppc, where the ioctl structure diverges from what's used for x86.


> I do think they want to be `const fn` if at all possible.

@joshtriplett: Is it ok by now to mark these as `const fn` directly? Making all of these const was my intention initially, but I couldn't figure out how to get the "conditional const where supported" macros to work.

JohnTitor:
> Should I also write some unit tests to make sure that the values generated here match what the respective C header macros generate?

I guess it should? How to add unit tests, see https://github.com/rust-lang/libc/blob/master/libc-test/test/makedev.rs for example.


> Is it ok by now to mark these as `const fn` directly? Making all of these const was my intention initially, but I couldn't figure out how to get the "conditional const where supported" macros to work.

Yes, it's totally fine, it's conditional by default: https://github.com/rust-lang/libc/blob/c20064f1fcba81b365ef3c06d39183687737ef34/build.rs#L100-L113

JohnTitor:
Closing as inactive, feel free to reopen/resubmit if you're still interested in this change. Thank you!

habnabit:
if this is revived.. could those be `const fn _IO*` ?

sm-Fifteen:
> if this is revived.. could those be `const fn _IO*` ?

I'm honestly not too sure anymore whether ioctls should belong in there or in libc anymore, but I believe [the nix crate](https://docs.rs/nix/0.29.0/nix/sys/ioctl/index.html) has since gained this feature (with support for [all the architectures defined in the Linux uapi](https://github.com/nix-rust/nix/blob/15db1fd97bb296c58c34de8344d079abcb8f18b3/src/sys/ioctl/linux.rs)), if you need something you can use right now. If the parameters can all be resolved into a constant value at compile time, it looks like that's what the macro would do for the wrapper function (it wraps the `ioctl(fd)` call as well) that it creates.

habnabit:
> > if this is revived.. could those be `const fn _IO*` ?
> 
> I'm honestly not too sure anymore whether ioctls should belong in there or in libc anymore, but I believe [the nix crate](https://docs.rs/nix/0.29.0/nix/sys/ioctl/index.html) has since gained this feature (with support for [all the architectures defined in the Linux uapi](https://github.com/nix-rust/nix/blob/15db1fd97bb296c58c34de8344d079abcb8f18b3/src/sys/ioctl/linux.rs)), if you need something you can use right now. If the parameters can all be resolved into a constant value at compile time, it looks like that's what the macro would do for the wrapper function (it wraps the `ioctl(fd)` call as well) that it creates.

i'd use whatever for raymanfx/libv4l-rs#118 if nix is a better option. i used rustix after i saw this PR wasn't merged.

PR #1664, opened 2020-02-19
Switch macOS images to 10.14

JohnTitor:
[Pipelines will remove macOS-10.13 builder](https://devblogs.microsoft.com/devops/removing-older-images-in-azure-pipelines-hosted-pools/), let's switch to 10.14.


JohnTitor:
Hm, I just want to remove 10.13 so let's stay on 10.14.

JohnTitor:
It seems fine, merging.

PR #1667, opened 2020-02-20
Re-enable semver check

JohnTitor:
Let's see if they work well now...

r? @ghost


JohnTitor:
Waiting on upstream fix.

JohnTitor:
Should be fixed in latest semververfork release. And we need some CI tweaks included in this, so let's merge.

PR #1678, opened 2020-03-05
Add constants from linux/keyctl.h

### LINKED ISSUE #1674: keyctl syscall constants missing
As a follow up to #1663, I've done more research into the matter and see that there is a `keyctl` raw syscall as outlined [here](http://man7.org/linux/man-pages/man2/keyctl.2.html) and the syscall number is even already supported by the libc crate. It does seem like many of the constant values are missing from the libc crate however. I would like to add the constants referenced in `/usr/include/linux/keyctl.h`. Would a PR to add these be of interest?

jbaublitz:
Closes #1674

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2237,6 +2237,7 @@ fn test_linux(target: &str) {
         "linux/if_ether.h",
         "linux/if_tun.h",
         "linux/input.h",
+        "linux/keyctl.h",
         "linux/magic.h",
         "linux/memfd.h",
         "linux/module.h",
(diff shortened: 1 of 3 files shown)


JohnTitor:
CI had some unrelated failures and they are fixed in master, could you rebase?

JohnTitor:
Also it'd be great if you could squash :)

jbaublitz:
Sure thing!

jbaublitz:
@JohnTitor It looks like I need to add the header to `libc-test/build.rs`. I'll update and squash again.

jbaublitz:
CI seems to be giving me a hard time. Going to throw up some commits and I'll squash once CI passes.

JohnTitor:
:+1:, you can ignore semver check failures, I'll take a look later.

jbaublitz:
Okay great. It looks like the failure is unrelated to my changes and is just in the android test. This seems to happen inconsistently because the last change I made did not have this failure. Does that seem consistent with what you've seen?

JohnTitor:
Hm, it seems weird. Anyway the change looks good to me, could you squash again so that we can re-run CI?

jbaublitz:
@JohnTitor Okay, squashed and pushed again. I'll keep an eye on the CI.

PR #1684, opened 2020-03-08
Add `riscv64gc-unknown-linux-gnu` image

JohnTitor:
r? @ghost

### DIFF
--- a/ci/docker/riscv64gc-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/riscv64gc-unknown-linux-gnu/Dockerfile
@@ -1,0 +1,12 @@
+FROM ubuntu:20.04
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        gcc libc6-dev qemu-user ca-certificates \
+        gcc-riscv64-linux-gnu libc6-dev-riscv64-cross \
+        qemu-system-riscv64 linux-headers-generic
+
+ENV CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_LINKER=riscv64-linux-gnu-gcc \
+    CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_RUNNER="qemu-riscv64 -L /usr/riscv64-linux-gnu" \
+    CC_riscv64gc_unknown_linux_gnu=riscv64-linux-gnu-gcc \
+    CFLAGS_riscv64gc_unknown_linux_gnu="-mabi=lp64d -fPIC" \
+    PATH=$PATH:/rust/bin
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2231,6 +2231,7 @@ fn test_linux(target: &str) {
     let aarch64_musl = target.contains("aarch64") && musl;
     let gnuabihf = target.contains("gnueabihf");
     let x86_64_gnux32 = target.contains("gnux32") && x86_64;
+    let riscv64 = target.contains("riscv64");
 
     let mut cfg = ctest_cfg();
     cfg.define("_GNU_SOURCE", None);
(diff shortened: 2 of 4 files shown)


JohnTitor:
Seems ctest needs to add that target:

```
--- stderr
thread 'main' panicked at 'unknown arch/pointer width: riscv64gc-unknown-linux-gnu', /cargo/registry/src/github.com-1ecc6299db9ec823/ctest-0.2.22/src/lib.rs:1080:9
```

Hey @gnzlbg, would you mind inviting me to ctest repo as a collaborator? Or if you have time to review, I'm happy to submit a PR!

JohnTitor:
Okay, CI is now green.
cc @msizanoen1 and @andreas-schwab, could you take a look? Looks like you're familiar with RISC-V than me :)

libc-test/build.rs
```diff
@@ -2657,7 +2658,7 @@ fn test_linux(target: &str) {
         (struct_ == "timex" && field.starts_with("__unused")) ||
         // FIXME: It now takes mode_t since glibc 2.31 on some targets.
         (struct_ == "ipc_perm" && field == "mode"
-            && ((x86_64 || i686 || arm) && gnu || x86_64_gnux32)
+            && ((x86_64 || i686 || arm || riscv64) && gnu || x86_64_gnux32)
```
andreas-schwab, line 2661: ipc_perm now uses mode_t everywhere.

andreas-schwab:
I have been building rust for riscv64 sucessfully for some time now, see <https://build.opensuse.org/package/show/openSUSE:Factory:RISCV/rust> and <https://build.opensuse.org/package/show/home:Andreas_Schwab:riscv:rust/rust>.

ci/docker/riscv64gc-unknown-linux-gnu/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        gcc libc6-dev qemu-user ca-certificates \
+        gcc-riscv64-linux-gnu libc6-dev-riscv64-cross \
+        qemu-system-riscv64 linux-headers-generic
+
+ENV CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_LINKER=riscv64-linux-gnu-gcc \
+    CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_RUNNER="qemu-riscv64 -L /usr/riscv64-linux-gnu" \
+    CC_riscv64gc_unknown_linux_gnu=riscv64-linux-gnu-gcc \
+    CFLAGS_riscv64gc_unknown_linux_gnu="-mabi=lp64d -fPIC" \
```
andreas-schwab, line 11: I think this is only needed because rust is still using llvm9, with llvm10 the riscv64gc-unknown-linux-gnu target should use the proper defaults now.

libc-test/build.rs
JohnTitor: Skipping this because it was introduced in the newest version of glibc and I'd like to wait for a while. On riscv, isn't it newer change?

JohnTitor:
This PR makes some fields private that shouldn't be public so I believe the build shouldn't be broken unless someone depends on that accidentally.

ci/docker/riscv64gc-unknown-linux-gnu/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ubuntu:20.04
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        gcc libc6-dev qemu-user ca-certificates \
+        gcc-riscv64-linux-gnu libc6-dev-riscv64-cross \
+        qemu-system-riscv64 linux-headers-generic
+
+ENV CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_LINKER=riscv64-linux-gnu-gcc \
+    CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_RUNNER="qemu-riscv64 -L /usr/riscv64-linux-gnu" \
```
tblah, line 9: Sometimes I've seen rust test suites hang on risc-v while using qemu user emulation, but the same tests work correctly under full system emulation. If it works okay here then great, but it is something to watch out for.

I believe this effects other targets too as it is mentioned in the [cross readme](https://github.com/rust-embedded/cross/blob/master/README.md) (just above the supported targets table).

JohnTitor: Hm, I have never seen that failure on other targets enabled here. Let's see whether it works well for a while.

JohnTitor:
So, I'd like to merge this since it seems not to affect current behavior. Thanks all for reviewing!

PR #1686, opened 2020-03-10
Add F_OFD_* constants

andrenth:
This PR adds the `F_OFD_*` Linux-specific constants for open file descriptor locks.


andrenth:
Thank you!

PR #1690, opened 2020-03-12
add macOS mount flags

### LINKED ISSUE #1688: mount flags for MacOS (BSD)
Hi All,

I'm using `libc::mount` under MacOS, and it seems `libc` only defines limited number of MacOS (BSD) `MNT_*` flags. Whereas, `libc` defines much more `MS_*` flags for Linux.

However, the mount(2) flags w.r.t. Linux and MacOS (BSD) are different, not only flag names differ, but also flag values.
Linux mount(2) flags:
```
#define MS_RDONLY	 1	/* Mount read-only */
#define MS_NOSUID	 2	/* Ignore suid and sgid bits */
#define MS_NODEV	 4	/* Disallow access to device special files */
...
```
MacOS (BSD) mount(2) flags:
```
#define	MNT_RDONLY	0x00000001	/* read only filesystem */
#define	MNT_SYNCHRONOUS	0x00000002	/* file system written synchronously */
#define	MNT_NOEXEC	0x00000004	/* can't exec from filesystem */
...
```

I'd like to add more MacOS (BSD) mount flags to libc, is [src/unix/bsd/mod.rs](https://github.com/rust-lang/libc/blob/master/src/unix/bsd/mod.rs#L326) the right place to start with?

Thanks!

pwang7:
#1688 add macOS mount flags according to Apple [sys/mount.h](https://github.com/apple/darwin-xnu/blob/master/bsd/sys/mount.h#L288)

Closes #1688

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -3037,6 +3037,43 @@ pub const DLT_LOOP: ::c_uint = 108;
 // sizeof(i32)
 pub const BPF_ALIGNMENT: ::c_int = 4;
 
+// sys/mount.h
+pub const MNT_RDONLY: ::c_int = 0x00000001;
+pub const MNT_SYNCHRONOUS: ::c_int = 0x00000002;
+pub const MNT_NOEXEC: ::c_int = 0x00000004;
+pub const MNT_NOSUID: ::c_int = 0x00000008;
+pub const MNT_NODEV: ::c_int = 0x00000010;
+pub const MNT_UNION: ::c_int = 0x00000020;
+pub const MNT_ASYNC: ::c_int = 0x00000040;
+pub const MNT_CPROTECT: ::c_int = 0x00000080;
+
+// NFS export related mount flags.
+pub const MNT_EXPORTED: ::c_int = 0x00000100;
+
+// MAC labeled / "quarantined" flag
+pub const MNT_QUARANTINE: ::c_int = 0x00000400;
+
+// Flags set by internal operations.
+pub const MNT_LOCAL: ::c_int = 0x00001000;
+pub const MNT_QUOTA: ::c_int = 0x00002000;
+pub const MNT_ROOTFS: ::c_int = 0x00004000;
+pub const MNT_DOVOLFS: ::c_int = 0x00008000;
+
+pub const MNT_DONTBROWSE: ::c_int = 0x00100000;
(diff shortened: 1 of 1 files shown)


JohnTitor:
> Since I just added bunch of flags, do these flags have to be used?

Because `MNT_FORCE` is already defined in:
https://github.com/rust-lang/libc/blob/8ba9ba2bc6055ec123e56864173b557e68278a32/src/unix/bsd/mod.rs#L326

You can just remove that const from your defined.
Hm, we may be able to move the consts to bsd/mod.rs (I thought they were defined as `c_ulonglong`)... But it's out of scope, I can follow-up later if you don't.

And you should comment on this PR, not the issue so that we don't have to come and go :)

JohnTitor:
Also, you should resolve style failure.

pwang7:
Should I move these flags to [libc/src/unix/bsd/mod.rs](https://github.com/rust-lang/libc/blob/8ba9ba2bc6055ec123e56864173b557e68278a32/src/unix/bsd/mod.rs#L326)?

JohnTitor:
> Should I move these flags to libc/src/unix/bsd/mod.rs?

I guess we can but we may need additional investigation so the answer is "if you like" :)
Also, CI will fail because of apple 32-bit targets, I fixed CI on master so could you rebase? And it'd be great if you could squash commits into one after finishing the work.

pwang7:
I did rebase and squash, I also moved the flags to libc/src/unix/bsd/mod.rs, but still some errors.

JohnTitor:
Thanks!

> I also moved the flags to libc/src/unix/bsd/mod.rs, but still some errors.

Heh, that's "we may need additional investigation" point. It seems some BSDs don't have some flags (therefore `use of undeclared identifier` error occurs). We should move them to under /bsd/apple. And `MNT_AUTOMOUNTED` should be defined as ulonglong in some BSDs.

pwang7:
BTW, should I move `MNT_FORCE` from [src/unix/bsd/mod.rs](https://github.com/rust-lang/libc/blob/8ba9ba2bc6055ec123e56864173b557e68278a32/src/unix/bsd/mod.rs#L326) to src/unix/bsd/apple/mod.rs?

JohnTitor:
> BTW, should I move MNT_FORCE from src/unix/bsd/mod.rs to src/unix/bsd/apple/mod.rs?

You don't have to.

JohnTitor:
Thanks!

pwang7:
Thanks @JohnTitor! This is my first commit to Rust~

PR #1692, opened 2020-03-12
Reduce checks for demoted apple 32-bit targets

JohnTitor:
Apple 32-bit targets are demoted in latest stable, we should reduce support level also here.

r? @ghost


JohnTitor:
Merging since this blocks other PRs.

PR #1694, opened 2020-03-14
Tweak `MAP_SYNC`

JohnTitor:
Investigate why mips64 is missing `MAP_SYNC`.
r? @ghost


JohnTitor:
Hm, seems [glibc doesn't have it yet](https://github.com/bminor/glibc/blob/5f72f9800b250410cad3abfeeb09469ef12b2438/sysdeps/unix/sysv/linux/mips/bits/mman.h). Okay let's define with usual way and remove `cfg`.

JohnTitor:
This should be the same as the current definition, merging.

PR #1695, opened 2020-03-14
Enable `s390x-unknown-linux-gnu` on CI

JohnTitor:
Let's check if this target works well, we should add FIXME comment at least if it doesn't.
r? @ghost


JohnTitor:
Yay, we now run `s390x-unknown-linux-gnu` target successfully! Merging.

PR #1698, opened 2020-03-18
Remove MSVC's wmemchr declaration

lzutao:


lzutao:
Style (lint) check failed
```
+ ./style src
src/windows/mod.rs:383 - use cfg_if! and submodules instead of #[cfg]
```

What should I do? There are no way to link to wmemchr on MSVC.

igrep:
I guess you should move the function you modified to https://github.com/rust-lang/libc/blob/master/src/windows/gnu/mod.rs

lzutao:
Thanks. Guess it should be fine now.

src/windows/gnu/mod.rs
```diff
@@ -13,6 +13,10 @@ extern "C" {
         s2: *const ::c_char,
         n: ::size_t,
     ) -> ::c_int;
+
+    // NOTE: For MSVC target, `wmemchr` is only a inline function in `<wchar.h>`
+    //      header file. We cannot find a way to link to that symbol from Rust.
+    pub fn wmemchr(cx: *const wchar_t, c: wchar_t, n: size_t) -> *mut wchar_t;
```
JohnTitor:
```suggestion
    pub fn wmemchr(cx: *const ::wchar_t, c: ::wchar_t, n: ::size_t) -> *mut ::wchar_t;
```

retep998:
I'm surprised CI hadn't detected this. The change looks correct to me.

JohnTitor:
Thanks, @retep998!

> I'm surprised CI hadn't detected this.

Yeah, I'll take a look later.

@lzutao r=me once code is formatted, line length limit here is tight a bit.
Also, thanks @igrep for the comment.

lzutao:
> CI could detect this absence

I think that the `extern` block relates to linkage, which happen during link stage and after codegen.
IMO we cannot test this case unless we have tests for every libc function.

JohnTitor:
> IMO we cannot test this case unless we have tests for every libc function.

Well CI should do that with ctest crate iirc?

lzutao:
Just curious: So we don't use bors anymore.

JohnTitor:
> Just curious: So we don't use bors anymore.

Yeah, Alex wants to abandon the use of bors here, I think we'll do so if there are no issues for a while.

PR #1702, opened 2020-03-19
Add definition of ucontext_t for macOS

alexcrichton:


alexcrichton:
Ok got it all green now too!

Issue #1501, opened 2019-09-04
af_alg_iv::as_slice makes a slice of the struct data, but the fields are public.


Lokathor:
[af_alg_iv::as_slice](https://github.com/rust-lang/libc/blob/6307a0bfce8b3fd980b9cb514e17d524de0a666d/src/unix/linux_like/android/mod.rs#L548-L557) makes a slice, but the [fields are public](https://github.com/rust-lang/libc/blob/6307a0bfce8b3fd980b9cb514e17d524de0a666d/src/unix/linux_like/android/mod.rs#L263-L266) and it's a public type, so you could construct a value, set the length to be some invalid value, and then call `as_slice` to make a slice with an improper length, pointing into invalid memory.

Shnatsel:
This is a read/write buffer overflow, which is most likely an exploitable security vulnerability. Please prioritize fixing this.

semarie:
@gnzlbg the `as_slice` code is present in both `src/unix/linux_like/android/mod.rs` and `src/unix/linux_like/linux/mod.rs`.

Removing it, means the type `af_alg_iv` will lost `PartialEq`, `Eq`, `Debug`, and `Hash` traits, and I doubt it will be possible to implement them without trusting `ivlen` for knowing `iv` array size.

But removing it is also a breaking change. It isn't used internally by `libc` but do it exist ways to know which (public) crates are using it ?

Shnatsel:
Alternatively you could make the fields private and create `unsafe fn` accessors to mutate them.

semarie:
@Shnatsel libc crate intent is to export the exact binding of struct definition "as it".

I think it would be against RFC1291 to make fields private and functions to modify them.

gnzlbg:
@semarie this type is borked. 

It should be a dynamically-sized type (e.g. using `[c_uchar]` as its last field), but instead its a sized type (using a zero-sized array). 

Then its fields are public, but should be private, because how to actually emulate dynamically sized types is an implementation detail (e.g. ideally we would just make this a type alias to `[c_uchar]` if `usize == u32`). 

So I'd be fine with:

* adding a Deref impl of &af_alg_iv to &[c_uchar], and a DerefMut impl of &mut af_alg_iv to &mut [c_uchar]
* deprecating the fields of these type, telling people to use the deref to slice instead
* one release later, making the types private, and changing the zero-sized array to actually be a [c_uchar]

[...]

PR #1703, opened 2020-03-19
Add deprecation notice to `af_alg_iv::as_slice` and trait implementations that depend on it

eduardosm:
These trait implementations exposed an unsound API (see https://github.com/rust-lang/libc/issues/1501).

### DIFF
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -591,36 +591,11 @@ cfg_if! {
             }
         }
 
-        impl af_alg_iv {
-            fn as_slice(&self) -> &[u8] {
-                unsafe {
-                    ::core::slice::from_raw_parts(
-                        self.iv.as_ptr(),
-                        self.ivlen as usize
-                    )
-                }
-            }
-        }
-
-        impl PartialEq for af_alg_iv {
-            fn eq(&self, other: &af_alg_iv) -> bool {
-                *self.as_slice() == *other.as_slice()
-           }
-        }
-
-        impl Eq for af_alg_iv {}
-
         impl ::fmt::Debug for af_alg_iv {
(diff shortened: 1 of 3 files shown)


eduardosm:
@JohnTitor ping

JohnTitor:
Given from https://github.com/rust-lang/libc/issues/1501#issuecomment-529094428 and https://github.com/rust-lang/libc/issues/1501#issuecomment-529136006, I'd be in favor of removing that implementation but I'd like to take some period for other users to notice. We could add deprecated attr at first and remove stuff after some releases.

eduardosm:
Looking at what I did with the `Debug` implementation (print `ivlen` and ignore `iv`), would it be acceptable to do the same with the other traits (e.g., `PartialEq` would only compare `ivlen`)? This would allow to keep the `PartialEq` and `Hash` implementations, but their behavior would change.

JohnTitor:
So, as discussed in #1892, let's just remove the impls for it. But as I said before, I'd add a deprecation note to make the breaking change less-painful. r=me once it's added instead of removing the impls here.

eduardosm:
Done, I have re-added the implementations and marked them as deprecated.

src/unix/linux_like/android/mod.rs
```diff
@@ -602,22 +604,31 @@ cfg_if! {
             }
         }
 
+        #[deprecated(
+            note = "this trait implementation is unsound and will be removed"
+        )]
```
JohnTitor, line 609: The attr should be applied to struct, not impl. Could you also add `since = "0.2.80"`?

eduardosm: Done. I also had to add some `#[allow(deprecated)]` in `src/macros.rs` to avoid warnings.

PR #1704, opened 2020-03-20
Add copy_file_range to linux_like

slp:
It was already present as a syscall, and both glibc and musl support
it, so add its function definition to export it to crate users.

Signed-off-by: Sergio Lopez <slp@redhat.com>

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -1198,6 +1198,14 @@ extern "C" {
     pub fn ntp_adjtime(buf: *mut timex) -> ::c_int;
     #[link_name = "ntp_gettimex"]
     pub fn ntp_gettime(buf: *mut ntptimeval) -> ::c_int;
+    pub fn copy_file_range(
+        fd_in: ::c_int,
+        off_in: *mut ::off64_t,
+        fd_out: ::c_int,
+        off_out: *mut ::off64_t,
+        len: ::size_t,
+        flags: ::c_uint,
+    ) -> ::ssize_t;
 }
 
 #[link(name = "util")]


JohnTitor:
Hm, I think Android should have that but CI failed, maybe SDK version or something else is wrong? If you don't need that in Android or emscripten right now, I'd suggest putting that in `linux_like/linux/`.
Also, musl targets failed to link. We use musl 1.1.24 on our CI, does that version have `copy_file_range`?

slp:
> Hm, I think Android should have that but CI failed, maybe SDK version or something else is wrong? If you don't need that in Android or emscripten right now, I'd suggest putting that in `linux_like/linux/`.

Yes, Linux would be enough for now, thanks.

> Also, musl targets failed to link. We use musl 1.1.24 on our CI, does that version have `copy_file_range`?

This is weird. According to musl's WHATSNEW, it was added on 1.1.24:

```
1.1.24 release notes

new features:
- GLOB_TILDE extension to glob
- non-stub catgets localization API, using netbsd binary catalog format
- posix_spawn file actions for [f]chdir (extension, pending future standard)
- secure_getenv function (extension)
- copy_file_range syscall wrapper (Linux extension)
- header-level support for new linux features in 5.2
```

slp:
I've reproduced the issue in a VM. I suspect the problem is that, despite installing musl 1.1.24 on the container, Rust is using the libc bundled with the x86_64-unknown-linux-musl.

Should I add it to `src/unix/linux_like/linux/gnu/`, or just wait for an update of the CI platform?

JohnTitor:
> Should I add it to src/unix/linux_like/linux/gnu/, or just wait for an update of the CI platform?

I'm on a break now and unsure that I have time to investigate/tweak CI atm. So the former is a better option for me.

slp:
> > Should I add it to src/unix/linux_like/linux/gnu/, or just wait for an update of the CI platform?
> 
> I'm on a break now and unsure that I have time to investigate/tweak CI atm. So the former is a better option for me.

Works for me :+1:

JohnTitor:
The failures are unrelated, closing/re-opening to re-trigger CI...

PR #1706, opened 2020-03-21
ci: allow overriding run-docker.sh CARGO_HOME.

cpu:
The `ci/run-docker.sh` utility script adds a `--volume` argument to the `docker` command to mount the Cargo home directory of the host machine into the container at `/cargo`.

Prior to this patch the host's Cargo home directory is assumed to be the `dirname` of the `dirname` of the `cargo` command's path. That works in most cases where the host machine installed rust with vanilla `rustup`. It may fail if the host machine used a different method.

As one concrete example if the host machine used the [Archlinux rustup package](https://wiki.archlinux.org/index.php/rust#Arch_Linux_package) then `cargo` is installed to `/usr/bin/cargo` and the `run-docker.sh` script incorrectly mounts `/usr` to the `/cargo` directory of the test container.

This patch allows specifying an explicit `CARGO_HOME` to the `ci/run-docker.sh` script so that users with a non-standard cargo dir can use the utility without modification. By default if no `CARGO_HOME` is set then the legacy behaviour is used and `CARGO_HOME` defaults to the `dirname` of the `dirname` of the `cargo` command.

### DIFF
--- a/ci/run-docker.sh
+++ b/ci/run-docker.sh
@@ -5,6 +5,13 @@
 
 set -ex
 
+# Default to assuming the CARGO_HOME is one directory up (to account for a `bin`
+# subdir) from where the `cargo` binary in `$PATH` lives.
+DEFAULT_CARGO_HOME="$(dirname "$(dirname "$(command -v cargo)")")"
+# If the CARGO_HOME env var is already set, use that. If it isn't set use the
+# default.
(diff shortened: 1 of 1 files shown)


cpu:
> It's mainly for CI but we could use it for debugging and this patch is certainly an improvement. Thanks!

Yup! This was exactly my use-case: local debugging. Thanks for the :mag:!

PR #1707, opened 2020-03-25
Add lockf

### LINKED ISSUE #1631: Missing POSIX lockf
* Linux documentation says it is POSIX, so I _think_ every unix system supported by Rust has this API, but I do not know if there is an exception. What I can tell is that Linux target triples should support it. Probably `#[cfg(unix)]` should be used.
* [link to the linux documentation on this function (which seems to be POSIX compliant)](https://linux.die.net/man/3/lockf). 

* [Example on playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c6bf45c5d557505480f18249f866920f):
```rust
use libc;

fn main() {
    let name = b"foo" as *const u8 as *const i8;
    let file = libc::open(name, libc::O_WRONLY);
    libc::lockf(file, libc::F_TLOCK, 0);
    libc::lockf(file, libc::F_ULOCK, 0);
    libc::close(file);
}
```

* the target triple - x86_64-unknown-linux-gnu
* Try to compile the following code in the given target triple. The following error appears:
```rust
   Compiling playground v0.0.1 (/playground)
error[E0425]: cannot find function `lockf` in crate `libc`
 --> src/main.rs:6:11
  |
6 |     libc::lockf(file, libc::F_TLOCK, 0);
  |           ^^^^^ not found in `libc`

error[E0425]: cannot find function `lockf` in crate `libc`
 --> src/main.rs:7:11
  |
7 |     libc::lockf(file, libc::F_ULOCK, 0);
  |           ^^^^^ not found in `libc`

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0425`.
error: could not compile `playground`.

To learn more, run the command again with --verbose.
```

pickfire:
Fix #1631

### DIFF
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -1209,6 +1209,11 @@ pub const SO_EE_ORIGIN_ICMP6: u8 = 3;
 pub const SO_EE_ORIGIN_TXSTATUS: u8 = 4;
 pub const SO_EE_ORIGIN_TIMESTAMPING: u8 = SO_EE_ORIGIN_TXSTATUS;
 
+pub const F_ULOCK: ::c_int = 0;
+pub const F_LOCK: ::c_int = 1;
+pub const F_TLOCK: ::c_int = 2;
+pub const F_TEST: ::c_int = 3;
+
 const_fn! {
     {const} fn CMSG_ALIGN(len: usize) -> usize {
         len + ::mem::size_of::<usize>() - 1 & !(::mem::size_of::<usize>() - 1)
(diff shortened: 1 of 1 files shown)


semarie:
as `lockf()` is POSIX function, I am suspecting that it is existing on several unix too (if not all). could you try to move your definition from `linux_like` to `unix` directory in order to see what CI will do ? (it should cover most of the supported platform)

Note I have checked that OpenBSD has it, and have same constants too.

pickfire:
@semarie Isn't `lockf` not a POSIX function? The man page I found is `3` not `2p`.

Still I will send another patch.

semarie:
the POSIX definition is [here](https://pubs.opengroup.org/onlinepubs/9699919799/functions/lockf.html).

regarding man page section, it is really distribution specific, so I can't really comment :-)

pickfire:
Superseded by https://github.com/rust-lang/libc/pull/1708

PR #1715, opened 2020-03-31
Exposes gregs from Fuchsia's libc

benbrittain:


benbrittain:
cc/ @tmandry & @leecam

JohnTitor:
Any reference for that? This is all I could find, the definition is different: https://fuchsia.googlesource.com/fuchsia/+/master/zircon/third_party/ulib/musl/include/bits/x86_64/signal.h#57

benbrittain:
I only added a subset of that definition. However, after going further down this path I think I've decided to not upstream this and use native syscalls to get the `pc`. Thanks for taking a look!

PR #1716, opened 2020-04-03
Split up Solaris and illumos targets

pfmooney:
While they share a common lineage, Solaris and illumos are distinct (and diverging) platforms which should be handled differently in libc.  This PR splits them up and addresses some all of the `libc-test` failures which appeared on illumos.  I have a separate branch for the `libc-test` updates, but given that `ctest` does not support illumos today, and the maintainer appears to be absent, those updates only function with that dependency overridden.

This is one of the last remaining pre-req updates required for rust-lang/rust#55553

### DIFF
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -1,6 +1,7 @@
 pub type c_char = i8;
 pub type c_long = i64;
 pub type c_ulong = u64;
+pub type caddr_t = *mut ::c_char;
 
 pub type clockid_t = ::c_int;
 pub type blkcnt_t = ::c_long;
@@ -35,9 +36,6 @@ pub type mqd_t = *mut ::c_void;
 pub type id_t = ::c_int;
 pub type idtype_t = ::c_uint;
 pub type shmatt_t = ::c_ulong;
-
-pub type door_attr_t = ::c_uint;
-pub type door_id_t = ::c_ulonglong;
 
 #[cfg_attr(feature = "extra_traits", derive(Debug))]
 pub enum timezone {}
@@ -217,33 +215,6 @@ s! {
         pub ai_next: *mut addrinfo,
     }
 
-    pub struct shmid_ds {
-        pub shm_perm: ipc_perm,
-        pub shm_segsz: ::size_t,
-        #[cfg(target_os = "illumos")]
-        pub shm_amp: *mut ::c_void,
-        #[cfg(target_os = "solaris")]
-        pub shm_flags: ::uintptr_t,
-        pub shm_lkcnt: ::c_ushort,
-        pub shm_lpid: ::pid_t,
-        pub shm_cpid: ::pid_t,
-        pub shm_nattch: ::shmatt_t,
-        pub shm_cnattch: ::c_ulong,
-        pub shm_atime: ::time_t,
-        pub shm_dtime: ::time_t,
-        pub shm_ctime: ::time_t,
-        #[cfg(target_os = "illumos")]
-        pub shm_pad4: [i64; 4],
-        #[cfg(target_os = "solaris")]
-        pub shm_amp: *mut ::c_void,
-        #[cfg(target_os = "solaris")]
-        pub shm_gransize: u64,
-        #[cfg(target_os = "solaris")]
-        pub shm_allocated: u64,
-        #[cfg(target_os = "solaris")]
-        pub shm_pad4: [i64; 1],
-    }
-
     pub struct sigset_t {
         bits: [u32; 4],
     }
@@ -371,7 +342,7 @@ s! {
         pub mq_maxmsg: ::c_long,
         pub mq_msgsize: ::c_long,
         pub mq_curmsgs: ::c_long,
-        _pad: [::c_int; 4]
+        _pad: [::c_int; 12]
     }
 
     pub struct port_event {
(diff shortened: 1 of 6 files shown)


src/unix/solarish/mod.rs
```diff
@@ -1361,7 +1312,7 @@ pub const RLIMIT_AS: ::c_int = RLIMIT_VMEM;
 
 #[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
 pub const RLIM_NLIMITS: rlim_t = 7;
-pub const RLIM_INFINITY: rlim_t = 0x7fffffff;
+pub const RLIM_INFINITY: rlim_t = 0xfffffffffffffffd;
```
pfmooney, line 1315: This was previously using the 32-bit value.  Given that all of the illumos/Solaris targets are 64-bit today, it made sense to use that value.

src/unix/solarish/mod.rs
```diff
@@ -1018,7 +969,7 @@ pub const O_CREAT: ::c_int = 256;
 pub const O_EXCL: ::c_int = 1024;
 pub const O_NOCTTY: ::c_int = 2048;
 pub const O_TRUNC: ::c_int = 512;
-pub const O_NOFOLLOW: ::c_int = 0x200000;
+pub const O_NOFOLLOW: ::c_int = 0x20000;
```
pfmooney, line 972: This appears to have been a typo during the initial import.

src/unix/solarish/mod.rs
```diff
@@ -1920,9 +1869,9 @@ pub const B230400: speed_t = 20;
 pub const B307200: speed_t = 21;
 pub const B460800: speed_t = 22;
 pub const B921600: speed_t = 23;
-pub const CSTART: ::tcflag_t = 021;
-pub const CSTOP: ::tcflag_t = 023;
-pub const CSWTCH: ::tcflag_t = 032;
+pub const CSTART: ::tcflag_t = 0o21;
+pub const CSTOP: ::tcflag_t = 0o23;
+pub const CSWTCH: ::tcflag_t = 0o32;
```
pfmooney, line 1874: These were transcription errors, given they were 0-prefixed octal values in the source header

src/unix/solarish/solaris.rs
```diff
@@ -0,0 +1,95 @@
...
+    #[cfg_attr(feature = "extra_traits", allow(missing_debug_implementations))]
+    pub struct door_arg_t {
+        pub data_ptr: *const ::c_char,
+        pub data_size: ::size_t,
+        pub desc_ptr: *const door_desc_t,
+        pub dec_num: ::c_uint,
+        pub rbuf: *const ::c_char,
+        pub rsize: ::size_t,
+    }
+}
```
pfmooney, line 95: Being not thrilled with the type naming mess, I've hidden the doors API from the new illumos target for now.  We can circle back in the future to address it.  At least one error was made when it was imported (`dec_num` should be `desc_num` in `door_arg_t`, for example)

JohnTitor:
Thanks for the PR!

> I have a separate branch for the libc-test updates, but given that ctest does not support illumos today, and the maintainer appears to be absent, those updates only function with that dependency overridden.

I'd like to [fork ctest repo](https://github.com/JohnTitor/ctest) and use that repo until @gnzlbg gets back. Would you mind sending PR(s) to the repo?

pfmooney:
> Thanks for the PR!
> 
> > I have a separate branch for the libc-test updates, but given that ctest does not support illumos today, and the maintainer appears to be absent, those updates only function with that dependency overridden.
> 
> I'd like to [fork ctest repo](https://github.com/JohnTitor/ctest) and use that repo until @gnzlbg gets back since I have no permission on the original repo. Would you mind sending PR(s) to the repo?

Yes, of course!  Some of these issues were frustrating to find, as they would have been caught by the tests (had they actually been run).

pfmooney:
The additional `ctest`-related commit fixes the test suite for illumos, and presumably improves the situation for Solaris.  I have no Solaris(tm) environment to test in.  It does require a local `ctest` dependency with JohnTitor/ctest#1 merged (or equivalent).

pfmooney:
If/when this is merged, what are your thoughts about a patch version bump for the package?  Do you have some preferred criteria for what's necessary to meet that threshold?

JohnTitor:
> If/when this is merged, what are your thoughts about a patch version bump for the package? Do you have some preferred criteria for what's necessary to meet that threshold?

Usually, I release a new version if requested as mentioned in [contributing guide](https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md#releasing-your-change-to-cratesio). So if you want me to release it once this PR gets merged, I'm happy to do so :)

pfmooney:
> Usually, I release a new version if requested as mentioned in [contributing guide](https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md#releasing-your-change-to-cratesio). So if you want me to release it once this PR gets merged, I'm happy to do so :)

Once this is merged, I'll circle back to do another round of testing with the rust-lang/rust#55553 wad and then open a ticket here requesting the version bump.

Thanks!

src/unix/solarish/mod.rs
GuillaumeDIDIER: Wouldn’t it be a good idea to make this definition different for 32 and 64 bit targets ?

pfmooney: The existing builds for rust on both Solaris and illumos target only x86_64.  If/when someone is inclined to set up a 32-bit build, they'll need to check all of the libc interfaces, this constant included, for conformance.

pfmooney:
With the tests added, are there any other changes you'd like to see made to this before it goes in?

pfmooney:
> Looks good to me, sorry for the delay!

No worries! Thanks.

PR #1717, opened 2020-04-03
fix cfmakeraw() for illumos and Solaris

jclulow:
This change fixes two issues.  First, the current `cfmakeraw()` implementation in this crate appears to be making a stack copy of the input `struct termios` before modifying it, rather than correctly modifying the original through the pointer.  Before this modification the routine did not, thus, set the flags for raw mode.

Second, we address the default settings of the **MIN** and **TIME** terminal options.  On at least FreeBSD and Linux systems, the modern default value for **MIN** appears to be 1; i.e., block and wait for at least one input byte.  On most Solaris and illumos systems, the stock default value is 4.  This breaks at least the examples in the **termion** crate, and probably quite a lot of other software written first and foremost for Linux systems.  In addition, the value is not even a hard-coded default on illumos systems; it can be overridden in `/kernel/drv/options.conf`, so forcing the default here seems wise.

I'd like to get a version bump after this integrates, so that I can fix **termion** to depend on it.  Thanks!

### DIFF
--- a/src/unix/solarish/compat.rs
+++ b/src/unix/solarish/compat.rs
@@ -4,8 +4,7 @@
 use unix::solarish::*;
 
 pub unsafe fn cfmakeraw(termios: *mut ::termios) {
-    let mut t = *termios as ::termios;
-    t.c_iflag &= !(IMAXBEL
+    (*termios).c_iflag &= !(IMAXBEL
         | IGNBRK
         | BRKINT
         | PARMRK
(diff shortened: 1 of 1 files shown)


jclulow:
With my apologies, the original comment was not quite accurate.  Upon further research the **MIN** default isn't 4 for reasons of capricious platform differences, but rather for unfortunate historical ABI compatibility reasons.  I've updated it to reflect my findings.  The actual code hasn't changed.

See also, [illumos bug 12470](https://www.illumos.org/issues/12470), which I am now less confident we can just fix!

PR #1721, opened 2020-04-06
Add strsignal(3) to unix

brian-gavin:
This does not add sys_siglist because the docs specify that the function
should be used instead, whenever possible.

Note: Draft PR until the CI is finished.


JohnTitor:
The change looks good, could you rebase to make CI happy and mark as ready?

PR #1722, opened 2020-04-07
Update freebsd 12 image

JohnTitor:
I guess it fixes recent freebsd failure.


JohnTitor:
Okay, seems it's fixed, merging.

PR #1725, opened 2020-04-08
FD_ISSET: take a *const fd_set instead of *mut fd_set

asomers:
FD_ISSET does not modify its fd_set argument, so it may as well take a
const pointer.  AFAICT the only reason to take a *mut pointer is because
the Linux man page documents it that way (though since glibc implements
it as a macro, the constedness is undefined).  But since libc implements
it directly rather than calling a (nonexistent on most platforms) C
function, we're defining the API ourselves.

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -3158,7 +3158,7 @@ f! {
         return
     }
 
-    pub fn FD_ISSET(fd: ::c_int, set: *mut fd_set) -> bool {
+    pub fn FD_ISSET(fd: ::c_int, set: *const fd_set) -> bool {
         let fd = fd as usize;
         let size = ::mem::size_of_val(&(*set).fds_bits[0]) * 8;
         return ((*set).fds_bits[fd / size] & (1 << (fd % size))) != 0
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -541,7 +541,7 @@ f! {
         return
     }
 
-    pub fn FD_ISSET(fd: ::c_int, set: *mut fd_set) -> bool {
+    pub fn FD_ISSET(fd: ::c_int, set: *const fd_set) -> bool {
         let bits = ::mem::size_of_val(&(*set).fds_bits[0]) * 8;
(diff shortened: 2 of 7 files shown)


zombiezen:
While I was researching my nix PR, I came across [this page](https://www.austingroupbugs.net/view.php?id=220&nbn=11) about the const-ness of that argument. I'm not enough of an OS expert to opine on whether the arguments are convincing, but thought I'd pass along in case it's useful.

JohnTitor:
Uhm, it's reasonable indeed but I'd prefer as-is since it may cause some confusion by difference with the original definition.

asomers:
By "original definition" do you mean existing versions of libc, or what's documented in C libraries?  AFAIK the only place in C that describes the pointer as mutable is in some man pages.  But that's a misnomer since the function is really a macro, and its arguments don't have constedness.  And some operating systems document `FD_ISSET` as it really is: as a macro.  So I don't think we should feel bound by any constedness described in some OS's man pages.

JohnTitor:
Fair enough, but on the other hand, it makes the current signature stricter, I think we shouldn't do that without any strong reason or announcement.

asomers:
Apparently libc does not object to changing the signatures of other functions, even when that causes downstream breakage.  Since that's the case, why not merge this PR?
https://github.com/rust-lang/libc/pull/1891

JohnTitor:
Triage: Re-visiting, I still, however, don't see much motivation for this.
@rust-lang/libc Do y'all have any thoughts on it?

asomers:
The motivation is const-correctness.  Right now for example, Nix forces users to use a mutable reference in a place where an immutable reference should suffice, just because libc defines `FD_ISSET` wrongly.
https://docs.rs/nix/0.22.0/nix/sys/select/struct.FdSet.html#method.contains

Amanieu:
A `*mut` will coerce to a `*const`, so this will most likely not break anything. Also I would agree with @asomers that `*const` is the correct type to use here.

JohnTitor:
Thanks for clarifying both!
@Amanieu I'm unfamiliar with this, so feel free to r=you if it makes sense to you.

Issue #1592, opened 2019-11-13
i686-pc-windows-gnu broken on CI


gnzlbg:
The i686-pc-windows-gnu target has been disabled on CI because it fails with the following segmentation fault: 

```
error: test failed, to rerun pass '--test main'

Caused by:
  process didn't exit successfully: `D:\a\1\s\target\i686-pc-windows-gnu\debug\deps\main-65866843958a68e7.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
./ci/run.sh: line 93:  1846 Segmentation fault      cargo test -vv $opt --no-default-features --manifest-path libc-test/Cargo.toml --target "${TARGET}"
```

For example, see: https://dev.azure.com/rust-lang2/libc/_build/results?buildId=608 

This issue tracks re-enabling CI for this target. Even though this is a tier-1 target, it has been broken for a sufficiently long time that new releases of libc won't be tested against it, and probably rust-lang/rust libc version will be updated to one of those sooner or later. Without a debugger, there is not much I can do here.

cc @mati865 @retep998 (feel free to ping windows users that might be able to investigate)

mati865:
Latest nightly has fix for i686 miscompilation: https://github.com/rust-lang/rust/pull/66318
Can you retry the build?

mati865:
Hopefully useful:
```
(lldb) r
Process 11012 launched: 'D:\msys64\tmp\libc\target\i686-pc-windows-gnu\debug\deps\main-fc7d8e49f1073eac.exe' (i686)
RUNNING ALL TESTS
Process 11012 stopped
* thread #1, stop reason = Exception 0xc0000005 encountered at address 0x4adab0: Access violation writing location 0x2ef71bd1
    frame #0: 0x004adab0 main-fc7d8e49f1073eac.exe`__test_roundtrip_max_align_t(value=max_align_t @ 0x009bfa90, rust_size=1996201872, error=0x2ef71bd1, pad="") at main.c:4583:28
   4580                         "size of max_align_t is %d in C and %d in Rust\n",
   4581                         (int)size, (int)rust_size
   4582                     );
-> 4583                     *error = 1;
                                   ^
   4584                     return value;
   4585                 }
   4586                 int i = 0;
(lldb) bt
(lldb) bt
* thread #1, stop reason = Exception 0xc0000005 encountered at address 0x4adab0: Access violation writing location 0x2ef71bd1
[...]

PR #1726, opened 2020-04-08
Fix i686-pc-windows-gnu CI

mati865:
Fixes https://github.com/rust-lang/libc/issues/1592

### DIFF
--- a/ci/azure.yml
+++ b/ci/azure.yml
@@ -114,12 +114,10 @@ jobs:
           ARCH: x86_64
         x86_64-pc-windows-msvc:
           TARGET: x86_64-pc-windows-msvc
-        # Disabled because broken:
-        #  https://github.com/rust-lang/libc/issues/1592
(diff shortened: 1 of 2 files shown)


mati865:
Huh?
```
bad max_align_t size: rust: 24 (0x18) != c 48 (0x30)
bad max_align_t align: rust: 8 (0x8) != c 16 (0x10)
```
Seems like MinGW on the CI is seriously old.

retep998:
Why does `time` on 32-bit mingw take a 64-bit value while `time_t` is 32-bit?

mati865:
> Why does time on 32-bit mingw take a 64-bit value while time_t is 32-bit?

Because of `#[link_name = "_gmtime64_s"]`, on second thought it's not correct and should be `#[link_name = "_gmtime32_s"]` for i686.

Anyway it still crashes with GCC so probably that's GCC fault.

PR #1729, opened 2020-04-11
Remove license headers

JohnTitor:
Some projects on the rust-lang org removed the license headers, I think we can do so here.
r? @alexcrichton


JohnTitor:
Thanks!

PR #1730, opened 2020-04-11
unix: add strndup()

NeoRaider:
Add strndup(), as specified by POSIX.1-2008.


JohnTitor:
The failure is unrelated but some jobs were skipped. Could you rebase to make sure things?

PR #1738, opened 2020-04-14
Added support for linux sockaddr_pppox.

alvardes:


JohnTitor:
Hm, CI fails. I think the definition is wrong.

JohnTitor:
Note: It seems you failed to rebase (given from CI).

alvardes:
Hmm... 

@JohnTitor Can you help me?

Why does this fail on mips and sparc? It seems that the layout is the same. The roundtrip test however fails :\ Any idea?

JohnTitor:
So, at a glance, `#[repr(packed)]` causes the failure, I guess?

alvardes:
@JohnTitor Could you take a look at this:

https://dev.azure.com/rust-lang2/libc/_build/results?buildId=1917&view=logs&j=c8d582de-107e-5fbd-122b-5384eb107855&t=7bb3b962-ee23-58e7-68da-769fcc260c0d&l=2371

```
o you wish to create a custom hardware profile? [no] 
Error processing tar file(exit status 1): write /root/.android/avd/aarch64.avd/userdata.img: no space left on device

##[error]Bash exited with code '1'.

```

JohnTitor:
I guess it's temporary, I saw that failure on rust-lang/rust as well.
Let me close/re-open to re-trigger CI..

JohnTitor:
Ugh, it still appears. Then my another guess is that our cache storage is full.
cc @pietroalbini thoughts --^?

alvardes:
@JohnTitor 

Now it passes all the test except `aarch64-unknown-linux-android`. 

What now?

(Because 1.13 doesn't support unions I only added support for `sockaddr_pppox` above 1.13. `pppoe_addr` and `pptp_addr` which are simple structs are added for all versions. Is that a good or should I cfg them too so that they are only present above 1.13?)

libc-test/build.rs
```diff
@@ -2231,6 +2231,9 @@ fn test_linux(target: &str) {
     let aarch64_musl = target.contains("aarch64") && musl;
     let gnuabihf = target.contains("gnueabihf");
 
+    let supports_unaligned_memory_access =
+        !(mips32 || mips64 || sparc64 || s390x);
```
JohnTitor, line 2235: Should we really skip tests on their targets?

src/unix/linux_like/linux/mod.rs
```diff
@@ -504,6 +504,33 @@ s! {
...
+        pub dev: [u8; 16]
+    }
+
+    pub struct pptp_addr {
+        pub call_id: u16,
+        pub sin_addr: ::in_addr
+    }
+}
+
+#[cfg(libc_union)]
```
JohnTitor, line 520: We could use `cfg_if!` macro.

src/unix/linux_like/linux/mod.rs
```diff
@@ -597,6 +624,97 @@ cfg_if! {
             }
         }
 
+        #[cfg(libc_union)]
+        #[allow(unused_unsafe)]
```
JohnTitor, line 628: I don't think we should allow it here.

src/unix/linux_like/linux/mod.rs
```diff
@@ -597,6 +624,97 @@ cfg_if! {
...
+                    /* PX_PROTO_PPTP */
+                    2  => unsafe { self.sa_addr.pptp == other.sa_addr.pptp },
+
+                    _ => {
+                        debug_assert!(
+                            false,
+                            "The sockaddr_pppox only supports PX_PROTO_OE \
+                            and PX_PROTO_PPTP. sa_protocol was: {}",
+                            unsafe { self.sa_protocol.clone() }
+                        );
```
JohnTitor, line 646: This could be removed once it's ready to land.

src/unix/linux_like/linux/mod.rs
```diff
@@ -597,6 +624,97 @@ cfg_if! {
...
+                    /* PX_PROTO_PPTP */
+                    2 => unsafe { self.sa_addr.pptp.hash(s) },
+
+                    _ => {
+                        debug_assert!(
+                            false,
+                            "The sockaddr_pppox only supports PX_PROTO_OE \
+                            and PX_PROTO_PPTP. sa_protocol was: {}",
+                            unsafe { self.sa_protocol.clone() }
+                        );
```
JohnTitor, line 677: Same here.

src/unix/linux_like/linux/mod.rs
```diff
@@ -597,6 +624,97 @@ cfg_if! {
...
+                        (self.sa_family,
+                        self.sa_protocol,
+                        self.sa_addr.pptp).fmt(f)
+                    } else {
+                        debug_assert!(
+                            false,
+                            "The sockaddr_pppox only supports PX_PROTO_OE \
+                            and PX_PROTO_PPTP. sa_protocol was: {}",
+                            self.sa_protocol.clone()
+                        );
```
JohnTitor, line 709: Same here.

src/unix/linux_like/linux/mod.rs
```diff
@@ -597,6 +624,97 @@ cfg_if! {
             }
         }
 
+        #[cfg(libc_union)]
```
JohnTitor, line 627: We could use `cfg_if!` macro here as well.

src/unix/linux_like/linux/mod.rs
```diff
@@ -597,6 +624,97 @@ cfg_if! {
...
+            }
+        }
+
+        #[cfg(libc_union)]
+        impl ::fmt::Debug for sockaddr_pppox_sa_addr_t {
+            fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                assert_eq!(24, ::mem::size_of::<Self>());
+                unsafe { ::mem::transmute::<_, [u8; 24]> (*self).fmt(f) }
+            }
+        }
```
JohnTitor, line 689: Hm, I think we should the same way as well as other places.

JohnTitor:
Also, CI issue is now fixed on master.

libc-test/build.rs
alvardes: The problem is that reading from packed fields should take into account that the field is unaligned. I don't think the rust side of the CTest generated code doesn't this, thats why the roundrip tests fail for the packed struct on these targets.

src/unix/linux_like/linux/mod.rs
alvardes: fixed

src/unix/linux_like/linux/mod.rs
alvardes: fixed

src/unix/linux_like/linux/mod.rs
alvardes: fixed

src/unix/linux_like/linux/mod.rs
alvardes: fixed

src/unix/linux_like/linux/mod.rs
alvardes: fixed

src/unix/linux_like/linux/mod.rs
alvardes: fixed

src/unix/linux_like/linux/mod.rs
alvardes: fixed

libc-test/build.rs
JohnTitor: So, I think it's better to have different definitions. Did you confirm that items work well on skipped targets?

JohnTitor:
Oh sorry, I somehow overlooked this one! If you're still interested, it'd be great if you could resolve merge conflicts. I'd like to see the CI result and take a look at the changes once it's done.

JohnTitor:
I'm going to close as inactive, feel free to rebase and re-open when you have some time to work on.

PR #1740, opened 2020-04-14
freebsdlike: add SCM_CREDS, struct sockcred

myfreeweb:
- `SCM_CREDS` is present on both FreeBSD and DragonFly, is used for sending `struct cmsgcred` control messages
- `struct sockcred` is present on FreeBSD, is used with `LOCAL_CREDS` socket option

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -107,6 +107,15 @@ s! {
         pub msg_hdr: ::msghdr,
         pub msg_len: ::ssize_t,
     }
+
+    pub struct sockcred {
+        pub sc_uid: ::uid_t,
+        pub sc_euid: ::uid_t,
+        pub sc_gid: ::gid_t,
+        pub sc_egid: ::gid_t,
+        pub sc_ngroups: ::c_int,
+        pub sc_groups: [::gid_t; 1],
+    }
 }
 
 s_no_extra_traits! {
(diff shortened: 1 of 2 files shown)


asomers:
This change looks good, but you should also define that constant for NetBSD (0x4), OpenBSD (0x2), and OSX (0x3).

myfreeweb:
It's already defined for Apple and NetBSD — but it's not 0x4 on the latter:

```
src/unix/bsd/netbsdlike/netbsd/mod.rs
1101:pub const SCM_CREDS: ::c_int = 0x10;
```

myfreeweb:
And it hasn't been available in OpenBSD since 2012 (I wonder what these "newer APIs" are lol)

https://github.com/openbsd/src/commit/8dd28dc23a49f9c516f4e7bb07c53b7cda3349ae

asomers:
> It's already defined for Apple and NetBSD — but it's ([correctly](https://github.com/NetBSD/src/blob/969b2c00707e6f10354f0012087cec3b74756f64/sys/sys/socket.h#L581)) not 0x4 on the latter:
> 
> ```
> src/unix/bsd/netbsdlike/netbsd/mod.rs
> 1101:pub const SCM_CREDS: ::c_int = 0x10;
> ```

Ahh, good catch.  It looks like NetBSD changed it 4 years ago, and wisely added a backwards-compatibility version: https://github.com/NetBSD/src/commit/8e79650ae14caa358c84daccf3ca3c99c5faaa3e .  I was looking at NetBSD 6 source.

PR #1741, opened 2020-04-14
Add CPU_ALLOC_SIZE

kubo39:


src/unix/linux_like/linux/mod.rs
```diff
@@ -2620,6 +2620,13 @@ f! {
         }
     }
 
+    pub fn CPU_ALLOC_SIZE(count: ::size_t) -> ::c_int {
```
JohnTitor, line 2623: I think it should be?

```suggestion
    pub fn CPU_ALLOC_SIZE(count: ::c_int) -> ::size_t {
```

JohnTitor:
r=me with squashing and fixing style issue (the failure on i686-unknown-linux-gnu is spurious).

PR #1745, opened 2020-04-25
Add RLIMIT_* constants for RISC-V 64-bit GNU/Linux and minor fixes for type definitions

msizanoen1:


andreas-schwab:
That doesn't make sense.  PTRACE_GETREGS, PTRACE_SETREGS, PTRACE_GETFPREGS, PTRACE_SETFPREGS, PTRACE_GETFPXREGS, PTRACE_SETFPXREGS don't exist on riscv64, and PTRACE_PEEKSIGINFO_SHARED is not architecture dependent.

msizanoen1:
I checked and it seems like those constants are not in the Linux headers for riscv64 but it is still in the glibc headers and therefore is mistakenly included in this port.

msizanoen1:
PTRACE_PEEKSIGINFO_SHARED is arch-independent but is not present in arch-independent files so this should be fixed in another PR.

msizanoen1:
There is still another part of this PR for minor type definition fixes by removing redundant type definitions that is already in the platform-independent type defintions.

PR #1751, opened 2020-05-06
Add timerfd bindings on android

### LINKED ISSUE #1589: timerfd API on android (aarch64-linux-android && arm-linux-androideabi)
`timerfd` APIs are only available as `SYS_*` calls
As Android is linux, these should be available too.

Should they be added?

List of APIs:

- `timerfd_create`
- `timerfd_settime`
- `timerfd_gettime`

That's and various definitions like `TFD_NONBLOCK`

Reference http://man7.org/linux/man-pages/man2/timerfd_create.2.html

wusyong:
Close #1589

### DIFF
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -147,6 +147,11 @@ s! {
         _pad: [u8; 28],
     }
 
+    pub struct itimerspec {
+        pub it_interval: ::timespec,
+        pub it_value: ::timespec,
+    }
+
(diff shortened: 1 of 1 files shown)


wusyong:
Sorry, didn't realize it's deprecated.

saurik:
I don't understand; these aren't deprecated? These APIs have been available since Android 4.4 (which is very old at this point) and should definitely be exposed on Android if they are exposed on Linux at all. Assuming this patch still works, can this be resurrected?

PR #1752, opened 2020-05-06
Update Ubuntu image to 20.04

JohnTitor:


JohnTitor:
> ##[error]An error occurred while provisioning resources (Error Type: Disconnect).
,##[error]The remote provider was unable to process the request.

Well, that failure should be unrelated to this PR but occurs twice :P

JohnTitor:
Hmm, the failure I saw doesn't occur anymore. Anyway, let's merge this and see whether it goes well.

PR #1753, opened 2020-05-06
Add c typedefs for the Sony PSP

overdrivenpotato:
This adds support for PSP targets to build important libraries like `libpanic_unwind`.

### DIFF
--- a/src/psp.rs
+++ b/src/psp.rs
@@ -1,0 +1,47 @@
+//! PSP C type definitions
+
+pub type c_schar = i8;
+pub type c_uchar = u8;
+pub type c_short = i16;
+pub type c_ushort = u16;
+pub type c_int = i32;
+pub type c_uint = u32;
+pub type c_float = f32;
+pub type c_double = f64;
+pub type c_longlong = i64;
+pub type c_ulonglong = u64;
+pub type intmax_t = i64;
+pub type uintmax_t = u64;
+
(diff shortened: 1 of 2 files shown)


overdrivenpotato:
The CI failure seems to be spurious, given that this PR does not touch android:

```
+ ./sdk/tools/bin/avdmanager create+ grep -v =
 avd --name arm --package system-images;android-24;google_apis;armeabi-v7a
Error: Package path is not valid. Valid system image paths are:
null
```

JohnTitor:
The failure should be fixed by #1754, could you rebase?

overdrivenpotato:
Sure, just rebased

overdrivenpotato:
Sure, I’ll give it a shot soon. Going to try and get a basic PSP target into rustc first.

src/psp.rs
```diff
@@ -0,0 +1,47 @@
...
+pub type intptr_t = isize;
+pub type uintptr_t = usize;
+pub type ssize_t = isize;
+
+pub type c_char = u8;
+pub type c_long = i64;
+pub type c_ulong = u64;
+
+pub const INT_MIN: c_int = -2147483648;
+pub const INT_MAX: c_int = 2147483647;
```
lzutao, line 27: I wonder those constants is needed as you can use `c_int::MAX`.

PR #1756, opened 2020-05-10
Add and rearrange a few fcntl pieces

khuey:
Add f_owner_ex, F_GET/SETOWN_EX, F_GET/SETSIG, and move F_OWNER_* to the toplevel linux-like module because they're libc and platform agnostic.


joshtriplett:
Closing and reopening to re-run CI.

src/unix/linux_like/linux/mod.rs
```diff
@@ -504,6 +504,11 @@ s! {
         pub ee_info: u32,
         pub ee_data: u32,
     }
+
+    pub struct f_owner_ex {
+        pub type_: ::c_int,
```
JohnTitor, line 509: We should add `f_owner_ex` here:

https://github.com/rust-lang/libc/blob/0d374002b8d4972e97e04b0a7fb1a210e22368ed/libc-test/build.rs#L2513-L2522

JohnTitor (changes requested):
The definition itself looks good to me but we need to tweak the test to pass CI.

khuey:
I'm still not sure how to get this test to pass unfortunately. There appears to be something going wrong inside ctest, perhaps?

@JohnTitor

JohnTitor:
I'm going to close as inactive, feel free to rebase and re-open when you get some time :)

PR #1759, opened 2020-05-12
Generate linux syscall constants

mahkoh:
Tl;dr: Many syscall constants were missing so I've written a small script that generates the constants from kernel sources.

As a first step I've reduced the duplication between glibc and musl by creating a new sibling module which is supposed to contain kernel structures and constants. There are many of those that could be moved out of the libc modules but for now I've only moved the syscall constants.

This violates the bottom-up design of the library but this design is not suitable for a system with multiple libcs because a significant part of their surface will be identical reexports of kernel structures and constants.


JohnTitor:
Hey, thanks for the PR! But this is a major change to our process, so I think we need to discuss it before reviewing.

mahkoh:
Just to be clear: The syscalls are generated manually and not during the build. So the process stays the same unless you want to add new syscalls which you should then do by using the script.

mahkoh:
@JohnTitor What's the status of this?

JohnTitor:
That's great that we could cover many syscalls here but I'm reluctant to this because this will affect too many targets at once.

mahkoh:
Adding new linux syscall numbers affects all linux targets? This is your objection?

mahkoh:
@JohnTitor Can I assume that this PR will stay in limbo indefinitely? You've said that it needs to be "discussed" but I've seen no such discussion nor any concerns raised beyond "it affects all linux targets". Please close the PR if you have no interest in the changes.

JohnTitor:
Sorry for my absence! So, as I said this would affect the targets broadly, and I concerned that this may introduce some inconsistencies. Since the current workflow is fine to maintain, I'm going to close this.
If you still want to add such a change, you could open an issue to make a decision at first.
Thanks for contributing anyway!

kpcyrd:
In case this is getting revived, the script I'm using to generate the syscall list for the syscallz crate is located here: https://github.com/kpcyrd/syscallz-rs/blob/main/gen.py

This is lightly tested on the major debian architectures: https://buildd.debian.org/status/package.php?p=rust-syscallz

PR #1762, opened 2020-05-15
Add extra functions & constants from sys/mman.h for Redox

AdminXVII:
This should add support for the mmap crate for Redox.

cc @jackpot51


AdminXVII:
I don't think the CI failure is due to this PR, as it only changes the redox's definitions.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "libc"
-version = "0.2.70"
+version = "0.2.71"
```
JohnTitor, line 3: I'd like to wait for a while to make a release since we recently did.

AdminXVII: Sure

AdminXVII:
Done

PR #1768, opened 2020-05-20
[FreeBSD] Add missing getnameinfo() flag values.

Erk-:
This patchs adds missing flag values for getnameinfo()
on FreeBSD, the following flags have been added from
the FreeBSD tree.

```c
/*
 * Flag values for getnameinfo()
 */
#define	NI_NOFQDN	0x00000001
#define	NI_NUMERICHOST	0x00000002
#define	NI_NAMEREQD	0x00000004
#define	NI_NUMERICSERV	0x00000008
#define	NI_DGRAM	0x00000010
#define	NI_NUMERICSCOPE	0x00000020
```

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -349,6 +349,13 @@ pub const RLIMIT_UMTXP: ::c_int = 14;
 #[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
 pub const RLIM_NLIMITS: ::rlim_t = 15;
 
+pub const NI_NOFQDN: ::c_int = 0x00000001;
+pub const NI_NUMERICHOST: ::c_int = 0x00000002;
+pub const NI_NAMEREQD: ::c_int = 0x00000004;
+pub const NI_NUMERICSERV: ::c_int = 0x00000008;
+pub const NI_DGRAM: ::c_int = 0x00000010;
+pub const NI_NUMERICSCOPE: ::c_int = 0x00000020;
+
 pub const Q_GETQUOTA: ::c_int = 0x700;
 pub const Q_SETQUOTA: ::c_int = 0x800;
 
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -1740,6 +1740,7 @@ fn test_freebsd(target: &str) {
             | "IP_RECVORIGDSTADDR"
             | "IPV6_ORIGDSTADDR"
             | "IPV6_RECVORIGDSTADDR"
+            | "NI_NUMERICSCOPE"
                 if Some(11) == freebsd_ver =>
             {
                 true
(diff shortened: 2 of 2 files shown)


JohnTitor:
Re-triggering CI.

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -349,6 +349,13 @@ pub const RLIMIT_UMTXP: ::c_int = 14;
 #[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
 pub const RLIM_NLIMITS: ::rlim_t = 15;
 
+pub const NI_NOFQDN: ::c_int = 0x00000001;
+pub const NI_NUMERICHOST: ::c_int = 0x00000002;
+pub const NI_NAMEREQD: ::c_int = 0x00000004;
+pub const NI_NUMERICSERV: ::c_int = 0x00000008;
+pub const NI_DGRAM: ::c_int = 0x00000010;
+pub const NI_NUMERICSCOPE: ::c_int = 0x00000020;
```
JohnTitor, line 357: This causes CI error on FreeBSD 10 and 11, it seems they don't have?: https://github.com/freebsd/freebsd/commit/43835a137db2c11db762c613be4c1be506f8cfb8
If so, then could you also tweak libc-test? You should touch here:

https://github.com/rust-lang/libc/blob/0dc26a9f290e683fa109a1bbcccced08cc85ee41/libc-test/build.rs#L1733

Erk-: This should be fixed now

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -354,7 +354,6 @@ pub const NI_NUMERICHOST: ::c_int = 0x00000002;
 pub const NI_NAMEREQD: ::c_int = 0x00000004;
 pub const NI_NUMERICSERV: ::c_int = 0x00000008;
 pub const NI_DGRAM: ::c_int = 0x00000010;
-pub const NI_NUMERICSCOPE: ::c_int = 0x00000020;
```
JohnTitor: You don't need to move it to the `freebsd12` dir since it was newly added.

Erk-: Do I then need to put it behind a `freebsd12` feature gate in `freebsd/mod.rs`? Or is that not needed here?

JohnTitor: `cfg`s e.g. `freebsd12` are for libc CI only and newly added constants are placed in `freebsd/mod.rs` usually (see other constants). CI will skip them by `skip_const` regardless of where they are (the `freebsd` dir vs. the `freebsdN` dir) but I'd like to keep them in the same place.

JohnTitor: Also, it'd be great if you could squash commits.

Erk-: I think that should be all

PR #1772, opened 2020-05-22
Update FIXME comment

JohnTitor:


JohnTitor:
So, the problem is that currently, it's _undefined_ whether we support glibc 2.27 or lower, or not. I'm not sure how many users depend on glibc < 2.28, so let's leave it for now.

PR #1774, opened 2020-05-23
linux: ptrace: Add definition for PTRACE_SYSEMU/_SINGLESTEP

jabedude:
This patch adds definitions for the PTRACE_SYSEMU and PTRACE_SYSEMU_SINGLESTEP constants. These ptrace requests are available on Linux x86 and x86_64.


JohnTitor:
Re-triggering CI...

jabedude:
Hmm, I don't understand the CI failures. It doesn't look like it's because of my commit?

JohnTitor:
> Hmm, I don't understand the CI failures. It doesn't look like it's because of my commit?

CI failures were unrelated and re-triggering resolved it.

Issue #1558, opened 2019-10-17
ip_mreqn roundtrip test fails for mips64 and sparc64 


gnzlbg:
See #1555 .

nikic:
Based on your godbolt link, looks like rustc is passing `{ i64, i32 }`, while C expects `i64, i32`. Not sure what the effect of that is though.

gnzlbg:
Not sure either :/

tgross35:
Resolved via sparc ABI fixes, and we no longer test mips in CI

PR #1786, opened 2020-05-28
Add `ip_mreqn` on musl targets

### LINKED ISSUE #1783: mreqn missing on musl
`struct mreqn` is available in the gnu env but not in the musl env.

JohnTitor:
Fixes #1783


JohnTitor:
So, some gnu targets have the roundtrip issue (see #1558) but musl ones don't, it seems. Hopefully, this won't cause any fallouts.

PR #1790, opened 2020-06-09
Add constants and eventfd to illumos

pfmooney:
A few constants as well as eventfd were noted as missing for the illumos target.  This change adds them both, as well as an update to the test harness to properly verify the eventfd parts.

I've verified that the test suite runs clean on a vanilla illumos (OpenIndiana) host:
```
RUNNING ALL TESTS
PASSED 6303 tests
```


pfmooney:
> Sorry for the delay, looks good!

No worries!  Thanks for taking a look.

PR #1791, opened 2020-06-11
FreeBSD: machine register structs

SWW13:
Add the CPU register structs for FreeBSD (x64) `reg32`, `reg`, `fpreg32`, `fpreg`, `xmmreg` from [machine/reg.h](https://github.com/freebsd/freebsd/blob/releng/12.1/sys/x86/include/reg.h)

The 32bit structs should also be available in x86 under a different name (without `32`), but I have no clue how to nicely add them without code duplication or complex `[cfg]`, happy to hear suggestions.

I skipped the debug register structs `dbreg32`, `dbreg`, their defines and deprecated register indices to simplify this PR (and because I don't need them).

Other BSD variants seem to have similar structs, but not identical one:
- dragonflybsd [machine/reg.h (x86)](https://gitweb.dragonflybsd.org/dragonfly.git/blob/master:/sys/cpu/x86_64/include/reg.h)
- openbsd [machine/reg.h (x86)](https://github.com/openbsd/src/blob/master/sys/arch/i386/include/reg.h)

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -1644,6 +1644,7 @@ fn test_freebsd(target: &str) {
                 "libutil.h",
                 "limits.h",
                 "locale.h",
+                "machine/reg.h",
                 "mqueue.h",
                 "net/bpf.h",
                 "net/if.h",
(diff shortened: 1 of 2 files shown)


JohnTitor:
Looks good! r=me once their commits are squashed.

PR #1792, opened 2020-06-16
Add ancillary socket data accessor functions for solarish OSes

jasonbking:


jclulow:
I did some work on this too, though I hadn't gotten around to ensuring mine was 100% correct: https://github.com/jclulow/rust-libc/blob/illumos_fixes/src/unix/solarish/mod.rs#L1920-L1978

It'd be good to make sure we do the right thing for SPARC systems here; i.e., something like:

```rust
// As per sys/socket.h, header alignment must be 8 bytes on SPARC
// and 4 bytes everywhere else:
#[cfg(target_arch = "sparc64")]
const _CMSG_HDR_ALIGNMENT: usize = 8;
#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
const _CMSG_HDR_ALIGNMENT: usize = 4;
```

pfmooney:
@jasonbking It would be good to re-enable the CMSG-related bits of libc-test (for illumos and solaris) as part of this.

jasonbking:
I have no idea of I did the building of `cmsg.c` the 'right' way -- it appears the build rule for it doesn't use any defines from the `cfg` instance, but it needs the `-D_XOPEN_SOURCE=700` define to correctly compile -- if there's a better way, please let me know and I'll update accordingly.

I also altered the `CMSG_DATA` definition to better match the spirit of the C macro (where it uses `((struct cmsghdr *)(c) + 1)` for the argument to `_CMSG_DATA_ALIGN`. The cmsg tests pass on illumos with this change.

JohnTitor:
Thanks all for working on this! The change looks good, r=me once their commits are squashed.
@pfmooney @jclulow Let me know if something is wrong as I'm sure you're familiar with this target than me :)

Issue #468, opened 2016-12-12
AF_INET and AF_INET6 should be of type sa_family_t


Yamakaky:
It's a define in the includes, but used as a `sa_family_t`. See `sockaddr` for example.

alexcrichton:
Unfortunately these probably can't change at this point, but we could consider change for the next major version bump of libc.

gnzlbg:
PRs welcome.

sum12:
https://github.com/sum12/libc/commit/93580add0dc2d66213400d9cc5d2ad19f7c2c436

I tried to fix it. but am not sure the fix is complete.

sum12:
Should I create a PR ? I would like to contribute.

safinaskar:
Breaking release of libc will be released soon, so we can now change type for `AF_INET`, etc. But I think its type should not be changed, because first argument of `socket` is `c_int`. :) There is no perfect solution here. Let's keep everything as is

safinaskar:
> Based on this comment, should we close this issue then?

According to this poll: [#t-libs > Questions and polls about &#96;libc&#96; 1.0 @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/Questions.20and.20polls.20about.20.60libc.60.201.2E0/near/511527554) , we should make it `sa_family_t`, because `sa_family_t` is smaller than `c_int`

PR #1794, opened 2020-06-23
type change for AF_INET, PF_INET  AF_INET6 PF_INET6

sum12:
I am not sure if this change is enough or more is needed fo #468


JohnTitor:
Hey, thanks for the PR! But this is a breaking change, as Alex said, and I'd hold off this as we currently have no plan to make a release 0.3 or 1.0.

sum12:
ok. Thanks for taking a look. 
Any other ways I could contribute   ?

JohnTitor:
> Any other ways I could contribute ?

Basically we add items when we need them. There're some requests on the issue tracker that someone wants to use constants or functions, I think they're good stuff to contribute (btw, if you want to contribute to Rust, there're some repos like rust-lang/rust that have more newcomer welcome issues).

joshtriplett:
Even if we were making breaking changes, I don't think we'd want to change the types of these particular constants and not others in the same category.

In C, these are preprocessor symbols, and they don't have a type. I don't think this kind of change is worth the breakage it would cause.

safinaskar:
Now this can be resent to libc 0.3

Issue #1765, opened 2020-05-18
[CI] CI for i686-linux-android is broken


JohnTitor:
Currently, aarch64-unknown-linux-android and i686-linux-android targets are broken.
The logs:
- [aarch64-unknown-linux-android](https://gist.github.com/JohnTitor/f38a8cf2a3ae7fd8159a59460866b7ef): it's now resolved but some it seems API version is old somehow.
- [i686-linux-android](https://gist.github.com/JohnTitor/f0b471790736babd0c6bb0b29c048348)

We should investigate the cause and land a fix.

rupansh:
Consider using a newer toolchain for android.
Current CI uses API 21's toolchain. These functions were added in later versions of android to the TC.
According to headers in `sysroot/usr/include` of the latest ndk

API 23 Functions:
`openpty`, `forkpty`, `login_tty`, `sethostname`, `sigtimedwait`, `fmemopen`, `open_memstream`, `open_wmemstream`, `clock_getcpuclockid`

API 24 Functions:
`lockf`,`preadv64`,`pwritev64`, `getifaddrs`, `freeifaddrs`, `getgrgid_r`, `getgrnam_r`

JohnTitor:
> Current CI uses API 21's toolchain. 

Hm, it's weird. IIUC we use API 24 for arm/aarch64 and API 28 for i686/x86_64, am I missing something?

rupansh:
> > Current CI uses API 21's toolchain.
> 
> Hm, it's weird. IIUC we use API 24 for arm/aarch64 and API 28 for i686/x86_64, am I missing something?

Did you recently change the CI? It was using android21 with azure pipelines 12 days ago.

JohnTitor:
No. Here's our config for Android NDK:
https://github.com/rust-lang/libc/blob/2d044d98ad1694ecdaa947974212b8f0202f570d/ci/android-install-ndk.sh#L5-L34

> It was using android21 with azure pipelines 12 days ago.

You mean `r21d` is API 21?

rupansh:
> No. Here's our config for Android NDK:
> 
> https://github.com/rust-lang/libc/blob/2d044d98ad1694ecdaa947974212b8f0202f570d/ci/android-install-ndk.sh#L5-L34
> 
> > It was using android21 with azure pipelines 12 days ago.
> 
> You mean `r21d` is API 21?

The old build is no longer opening on azure pipelines but I remember it was using  aarch64-linux-android21-clang
[...]

PR #1795, opened 2020-06-24
Add a bunch of constants and functions which were missing on Android

qwandor-google:
These are all based on the latest corresponding Bionic headers, from https://cs.android.com/android/platform/superproject/+/master:bionic/.

### DIFF
--- a/ci/android-install-ndk.sh
+++ b/ci/android-install-ndk.sh
@@ -2,7 +2,7 @@
 
 set -ex
 
-NDK=android-ndk-r19c
+NDK=android-ndk-r21d
 curl --retry 20 -O https://dl.google.com/android/repository/${NDK}-linux-x86_64.zip
 unzip -q ${NDK}-linux-x86_64.zip
 
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -1427,6 +1427,7 @@ fn test_android(target: &str) {
                "sys/sysinfo.h",
                "sys/time.h",
                "sys/times.h",
+               "sys/timerfd.h",
                "sys/types.h",
                "sys/ucontext.h",
                "sys/uio.h",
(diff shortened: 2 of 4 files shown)


JohnTitor:
Thanks for the PR! According to CI, it seems that we should also tweak libc-test to import header files (semver check can be ignored, we should fix that check but I have no bandwidth to do so currently, sorry!).

qwandor-google:
Thanks. I've added the missing headers, though some of the errors from CI don't seem to be related to that so there may still be some issues. Unfortunately I haven't been able to get the tests to run locally, sorry for the spam.

qwandor-google:
There is one remaining issue I haven't figured out yet, with sigtimedwait on aarch64 only. I can't see why this arch would be different. The symbol is in signal.h, so shouldn't be arch-specific. It was introduced in API version 23, and android-install-ndk.sh sets version 24 on all Arm architectures or 28 on x86 architectures. Any idea what's going wrong here?

JohnTitor:
> There is one remaining issue I haven't figured out yet, with sigtimedwait on aarch64 only. I can't see why this arch would be different.

Possibly related to #1765, I'd skip test for aarch64 at this point as it's obviously safe to ship, I think. Other stuff looks good, r=me once that point is resolved and you squash commits.

qwandor-google:
Okay, skipping the `sigtimedwait` test for aarch64. What would you like squashed? Commits all have descriptions and should build at each step of the way, and CONTRIBUTING.md doesn't say anything about a policy on squashing. I can squash everything into one commit if you like, but I figured the individual descriptions would be more useful.

JohnTitor:
> I figured the individual descriptions would be more useful.

I'd prefer to avoid history-noise than to remain each description except for CI stuff.

PR #1797, opened 2020-06-27
Add PSP OS API and add PSP CI support

overdrivenpotato:
This PR declares `extern` functions in a `"C"` ABI-compatible way, which is not necessarily the same ABI as required by the PSP OS (LLVM does not support the one we need). Dynamic linking on the PSP is also unfortunately non-standard, so simply including `libc` as a dependency and calling functions won't work.

Both of the problems here are resolved by using a crate like [rust-psp](https://github.com/overdrivenpotato/rust-psp), which patches over ABI incompatibility and resolves functions correctly. It only needs to be included in the crate graph once, as all OS functions are marked `extern` and exported via `#[no_mangle]`.

Having these functions in `libc` allows external crates to avoid depending on `rust-psp` while adding PSP support, and allows us to e.g. implement Rust `std` support for the PSP upstream.


JohnTitor:
I'll _rubbar-stamp_ this as the diff is too large and I think you're the primary user of this target. However I want to keep the file naming conventions the same as for other targets, for maintainability.

overdrivenpotato:
I’ve merged all the modules into a single one

src/psp.rs
```diff
@@ -45,3 +46,5078 @@ cfg_if! {
...
+        username: *mut u8,
+        password: *mut u8,
+        need_entity: i32,
+        entity_body: *mut *mut u8,
+        entity_size: *mut usize,
+        save: *mut i32,
+    ) -> i32,
+>;
+
+#[allow(non_camel_case_types)]
```
JohnTitor, line 136: I think this is unnecessary.

sajattack:
@overdrivenpotato asked me to take over this PR.

sajattack:
Looks like that build failed spuriously.

src/psp.rs
JohnTitor: This isn't removed yet.

src/psp.rs
```diff
@@ -45,3 +46,5267 @@ cfg_if! {
...
+
+extern "C" {
+    pub fn sceAudioChReserve(
+        channel: i32,
+        sample_count: i32,
+        format: AudioFormat,
+    ) -> i32;
+    pub fn sceAudioChRelease(channel: i32) -> i32;
+    pub fn sceAudioOutput(channel: i32, vol: i32, buf: *mut c_void) -> i32;
+    pub fn sceAudioOutputBlocking(
```
JohnTitor, line 388: I think we could use `f!` macro here.

src/psp.rs
```diff
@@ -45,3 +46,5267 @@ cfg_if! {
...
+pub struct Atrac3BufferInfo {
+    pub puc_write_position_first_buf: *mut u8,
+    pub ui_writable_byte_first_buf: u32,
+    pub ui_min_write_byte_first_buf: u32,
+    pub ui_read_position_first_buf: u32,
+    pub puc_write_position_second_buf: *mut u8,
+    pub ui_writable_byte_second_buf: u32,
+    pub ui_min_write_byte_second_buf: u32,
+    pub ui_read_position_second_buf: u32,
+}
```
JohnTitor, line 458: You should use `s!` macro here.

sajattack:
Can you explain more about what the macros do and how to use them?

JohnTitor:
> Can you explain more about what the macros do and how to use them?

Their macros can define repr, derives, and impls easily. For example, `Atrac3BufferInfo` can be included in the `s!` macro, I think.
You can refer to the definitions and existing implementations to see how each macro works.

sajattack:
It looks like f! is for non-extern functions?

sajattack:
@JohnTitor I think this is ready for another review.

JohnTitor:
> It looks like f! is for non-extern functions?

Ah yep!


@sajattack Sorry for the delay! Somehow this slipped from my notifications. Now I think it's ready to merge, my last concern is this has some style/format fix commits that should be squashed and merge commits that should be rebased instead. I'd like to squash their commits to make it easier to follow commit history. r=me once it's done. Thanks for your contribution!

JohnTitor:
@sajattack If you want us to ship this in the next release, I'll hold #1825 off for a while.

sajattack:
r=@JohnTitor

Rebase complete.

JohnTitor:
> In terms of squashing the style/format commits, do you want to just squash the whole thing?

Nope, it looks great! I just want to make it easy to follow history and blame, style fix commits are noisy :P

PR #1807, opened 2020-07-02
Bump version ready to import into Android tree.

qwandor-google:
I need to bump the version number so that I can get the changes from https://github.com/rust-lang/libc/pull/1795 into the version of the libc crate vendored into the Android tree at https://android.googlesource.com/platform/external/rust/crates/libc/.


JohnTitor:
Published 0.2.72, thanks!

PR #1810, opened 2020-07-05
Add some missing fcntl flags for freebsd

nagisa:


JohnTitor:
It seems some are newly added and CI image doesn't have them:
- https://github.com/freebsd/freebsd/commit/13d4dfe478f82d11e0c7d467bf4cd4e03a0112b6#diff-c1260f62516c033d5093931c9ab68797
- https://github.com/freebsd/freebsd/commit/fa60d0fc5121c0f566b8f83ab65f428d635cd687#diff-c1260f62516c033d5093931c9ab68797

You can skip tests for them by tweaking here:
https://github.com/rust-lang/libc/blob/51acd7f488f63c6f4c60a38b44b8439bb9d04821/libc-test/build.rs#L1756

nagisa:
I went ahead and:

a) Moved the constants to the `freebsd12` module as these constants are to be available starting with freebsd12.2 (I don’t think we have granularity finer than that)
b) Ignored the constants _still_ (freebsd12.2 is not yet a stable release).

Hopefully that works out.

JohnTitor:
Yeah, it should work! ...But it also raises some style issues, r=me once they're fixed.

PR #1812, opened 2020-07-05
Disable tests for the switch target, for now

JohnTitor:


WinterMute:
You shouldn't use pacman for CI, we provide docker images for this purpose. https://hub.docker.com/u/devkitpro

Issue #1802, opened 2020-07-01
README link to platform specific documentation is broken


jbaublitz:
It appears that the documentation is still accessible using the links in my history (for example [this page](https://rust-lang.github.io/libc/x86_64-unknown-linux-gnu/libc/index.html) is still accessible), but the link to the platform specific documentation in the README throws a 404.

PR #1816, opened 2020-07-07
Put index.html using unstable rustdoc features

JohnTitor:
Investigate the cause of #1802.


JohnTitor:
Okay, now this should fix #1802.

PR #1824, opened 2020-07-17
Expose EPOLLEXCLUSIVE on illumos platform

pfmooney:
Initially the `EPOLLEXCLUSIVE` definition was hidden on the illumos platform as it lacked explicit support.  After further review, it was concluded that `EPOLLEXCLUSIVE` can safely be considered a no-op, when not fully implemented by the OS, making it safe for use on illumos.

I've confirmed that the `ci/style` check is clean, and the `libc-test` suite runs successfully on an illumos system.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -794,6 +794,11 @@ fn test_solarish(target: &str) {
         // This evaluates to a sysconf() call rather than a constant
         "PTHREAD_STACK_MIN" => true,
 
+        // EPOLLEXCLUSIVE is a relatively recent addition to the epoll interface and may not be
+        // defined on older systems.  It is, however, safe to use on systems which do not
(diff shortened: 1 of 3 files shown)


libc-test/build.rs
```diff
@@ -794,6 +794,11 @@ fn test_solarish(target: &str) {
         // This evaluates to a sysconf() call rather than a constant
         "PTHREAD_STACK_MIN" => true,
 
+        // EPOLLEXCLUSIVE is a relatively recent addition to the epoll interface and may not be
+        // defined on older systems.  It is, however, safe to use on systems which do not
+        // explicitly support it. (A no-op is an acceptable implementation of EPOLLEXCLUSIVE.)
```
JohnTitor, line 799: 👍

PR #1828, opened 2020-07-20
linux: replace VMADDR_CID_RESERVED with VMADDR_CID_LOCAL

stefano-garzarella:
In Linux we replaced VMADDR_CID_RESERVED with VMADDR_CID_LOCAL
in commit ef343b35d46667668a099655fca4a5b2e43a5dfe.
It is available since Linux v5.6, and it can be used to do
local communication if supported.

Signed-off-by: Stefano Garzarella <sgarzare@redhat.com>

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -2355,7 +2355,13 @@ pub const MAP_FIXED_NOREPLACE: ::c_int = 0x100000;
 // uapi/linux/vm_sockets.h
 pub const VMADDR_CID_ANY: ::c_uint = 0xFFFFFFFF;
 pub const VMADDR_CID_HYPERVISOR: ::c_uint = 0;
+#[deprecated(
+    since = "0.2.74",
+    note = "VMADDR_CID_RESERVED is removed since Linux v5.6 and \
+            replaced with VMADDR_CID_LOCAL"
+)]
 pub const VMADDR_CID_RESERVED: ::c_uint = 1;
+pub const VMADDR_CID_LOCAL: ::c_uint = 1;
 pub const VMADDR_CID_HOST: ::c_uint = 2;
 pub const VMADDR_PORT_ANY: ::c_uint = 0xFFFFFFFF;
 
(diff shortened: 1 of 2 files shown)


src/unix/linux_like/linux/mod.rs
```diff
@@ -2355,7 +2355,7 @@ pub const MAP_FIXED_NOREPLACE: ::c_int = 0x100000;
 // uapi/linux/vm_sockets.h
 pub const VMADDR_CID_ANY: ::c_uint = 0xFFFFFFFF;
 pub const VMADDR_CID_HYPERVISOR: ::c_uint = 0;
-pub const VMADDR_CID_RESERVED: ::c_uint = 1;
+pub const VMADDR_CID_LOCAL: ::c_uint = 1;
```
JohnTitor, line 2358: We should add `#[deprecated]` attribute instead of removing for backward compatibility (even if it's unused).

JohnTitor: And CI images don't have `VMADDR_CID_LOCAL` as they're old a bit so we should also tweak the test to skip this.

stefano-garzarella: Sure! I'll do it!
Thanks!

stefano-garzarella: > And CI images don't have `VMADDR_CID_LOCAL` as they're old a bit so we should also tweak the test to skip this.

Where can I find the tests?

JohnTitor: Here's the test file: https://github.com/rust-lang/libc/blob/master/libc-test/build.rs
And you should tweak here:
https://github.com/rust-lang/libc/blob/f3bc8363593e6e62a5229c7fe22192961330e0e7/libc-test/build.rs#L2558

stefano-garzarella: Should I skip also VMADDR_CID_RESERVED?

E.g. on my host the test fails because it is not defined anymore.

JohnTitor: Yes good catch, we should certainly (actually the test fails on my local as well :P).

JohnTitor:
CI says:
```
+ rustc ci/style.rs
+ ./style src
src/unix/linux_like/linux/mod.rs:2360 - line longer than 80 chars
thread 'main' panicked at 'found some lint errors', ci/style.rs:54:9
```

r=me once it's fixed.

PR #1831, opened 2020-07-22
Added clock_getcpuclockid to linux

xonatius:
https://man7.org/linux/man-pages/man3/clock_getcpuclockid.3.html

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -2759,6 +2759,11 @@ extern "C" {
         sevp: *mut ::sigevent,
     ) -> ::c_int;
 
+    pub fn clock_getcpuclockid(
+        pid: ::pid_t,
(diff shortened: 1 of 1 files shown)


JohnTitor:
Note that a part of CI is broken currently due to upstream bug, it should be fixed in next nightly.

xonatius:
@JohnTitor got it, thanks!

src/unix/linux_like/linux/mod.rs
```diff
@@ -2759,6 +2759,11 @@ extern "C" {
         sevp: *mut ::sigevent,
     ) -> ::c_int;
 
+    pub fn clock_getcpuclockid(
+        pid: ::pid_t,
+        clk_id: *mut ::clockid_t
```
JohnTitor, line 2764: Add a comma for style.

```suggestion
        clk_id: *mut ::clockid_t,
```

PR #1832, opened 2020-07-23
Android: Add preadv and pwritev.

qwandor-google:
From sys/uio.h. Note that preadv64/pwritev64 are already included in
src/unix/linux_like/mod.rs.

### DIFF
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -2399,20 +2399,32 @@ extern "C" {
         sevlen: ::size_t,
         flags: ::c_int,
     ) -> ::c_int;
+    pub fn preadv(
+        fd: ::c_int,
+        iov: *const ::iovec,
+        count: ::c_int,
+        offset: ::off_t,
+    ) -> ::ssize_t;
+    pub fn pwritev(
+        fd: ::c_int,
+        iov: *const ::iovec,
+        count: ::c_int,
+        offset: ::off_t,
+    ) -> ::ssize_t;
     pub fn process_vm_readv(
         pid: ::pid_t,
         local_iov: *const ::iovec,
-        liovcnt: ::c_ulong,
+        local_iov_count: ::c_ulong,
         remote_iov: *const ::iovec,
-        riovcnt: ::c_ulong,
+        remote_iov_count: ::c_ulong,
         flags: ::c_ulong,
     ) -> ::ssize_t;
     pub fn process_vm_writev(
         pid: ::pid_t,
         local_iov: *const ::iovec,
-        liovcnt: ::c_ulong,
+        local_iov_count: ::c_ulong,
         remote_iov: *const ::iovec,
-        riovcnt: ::c_ulong,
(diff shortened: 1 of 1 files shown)


JohnTitor:
Hm, aarh64 test fails on CI.

qwandor-google:
Sorry, I was away for a bit and now can't see the CI results. How do I get it to re-run?

JohnTitor:
> 
> 
> Sorry, I was away for a bit and now can't see the CI results. How do I get it to re-run?

No worries! I also doesn't have the permission to re-run on Pipelines and usually close/re-open PRs to do it.

qwandor-google:
This looks like it might be #1765 again.

I do notice that the aarch64 Android build is using aarch64-linux-android21-clang, while the others use x86_64-linux-android28-clang and armv7a-linux-androideabi24-clang (note the different versions). The functions in question were introduced with API level 24, so this could be related. Where do these toolchains come from?

qwandor-google:
I might be wrong about that, digging some more. It looks like it comes from ci/android-install-ndk.sh calling build/tools/make_standalone_toolchain.py, but it always passes at least API 24, including on aarch64, so that shouldn't be the difference.

JohnTitor:
Hm, seems Pipelines couldn't find 07dbd45cdf621e964d1e06c42483719cd3565196, it has just three checks on Cirrus CI.
image

JohnTitor:
This AArch64 issue is painful and I have no idea how to fix TBH, sorry. The only way we can do here is to skip the test for it. I could r=me if you drop 07dbd45 and skip the test.

qwandor-google:
Done, PTAL.

src/unix/linux_like/android/mod.rs
```diff
@@ -2513,6 +2513,34 @@ extern "C" {
...
+        flags: ::c_ulong,
+    ) -> ::ssize_t;
+    pub fn process_vm_writev(
+        pid: ::pid_t,
+        local_iov: *const ::iovec,
+        local_iov_count: ::c_ulong,
+        remote_iov: *const ::iovec,
+        remote_iov_count: ::c_ulong,
+        flags: ::c_ulong,
+    ) -> ::ssize_t;
```
JohnTitor, line 2543: Oops sorry, they're added by 582c5cf0c04429b379ed73904437f52368eeece7 already.

qwandor-google: I guess you mean #1878. Fixed.

JohnTitor: Yeah, you're right :)

JohnTitor (approved):
Thanks! I'll r+ once CI is fixed.

JohnTitor:
@bors retry
```
warning: spurious network error (2 tries remaining): error inflating zlib stream; class=Zlib (5)
warning: spurious network error (1 tries remaining): error inflating zlib stream; class=Zlib (5)
error: failed to get `cc` as a dependency of package `libc-test v0.1.0 (D:\a\libc\libc\libc-test)`
```

Issue #1412, opened 2019-06-27
Minimum Supported glibc version


gnzlbg:
IIRC, the current minimum supported glibc version on `x86_64-unknown-linux-gnu` is 2.26. We offer APIs that are only available in glibc 2.27 and 2.28 though. When used, those API would crash on glibc 2.26, but if they are not used, everything is ok. 

That is, we support glibc >= 2.26, where some APIs are "forward-compatible" extensions. 

However, some new features in newer glibc versions (e.g. shadow stack in glibc 2.28, see #1410 ) are ABI incompatible with older glibc versions. That is, adding support for these features introduces undefined behavior in Rust programs that are dynamically linked against older glibc versions. 

We should:

* [ ] document the minimum glibc version
* [ ] have a policy for bumping the minimum supported glibc version (cc @rust-lang/libs)

Most crates use libc via crates.io, and not via the Rust toolchain, so we can't use crater to test breaking changes AFAIK. Upgrading the minimum glibc version is bad, because it introduces undefined behavior _silently_ into working Rust programs. Not upgrading the minimum glibc version is bad, because we can't expose newer features to users, and because code that uses newer backward-compatible features has undefined behavior if the minimum supported version is dynamically linked. It also has the downside that when the distros that use that particular glibc version go end-of-life, testing libc becomes harder (e.g. CI needs to work around this, and force Rust programs to be linked to the minimum supported glibc version even if the system has a newer one).

I don't know of any solution that does not have serious downsides. Right now, the policy for upgrading the minimum glibc version is "try it out, and if nobody complains, then the upgrade was ok". But since the UB is introduced silently, it takes a while for people to notice when it happens.

elichai:
Adding to the docs the minimum supported linux kernel could also help.
currently glib 2.26 implies minimum of Linux kernel 3.2.
https://sourceware.org/ml/libc-alpha/2017-08/msg00010.html

xhebox:
Anything new? I would like to mention a similar case like #1412: recently musl may introduce `pthread_getname_np`. Then rust can have that api on musl, too, getting tikv/pprof-rs compilable on musl.

But yes, it will break the old builds on older musl.

namazso:
Was this issue solved in the meantime? [This recent post on the rust blog](https://blog.rust-lang.org/2022/08/01/Increasing-glibc-kernel-requirements.html) implies it was.

dybucc:
For some time now, the libc-test build script has had version
[...]

PR #1833, opened 2020-07-24
FreeBSDLike: Remove obsolete constants in version 1.0.0

### LINKED ISSUE #454: Remove obsolete constants in version 1.0.0
This issue tracks constants that should be removed from libc for release 1.0.0.  

The following are absent in FreeBSD 11, never had stable values across versions, and should never have been used by third party software.  They should be removed from FreeBSD.
```
CTL_MAXID
KERN_MAXID
HW_MAXID
USER_MAXID
```
The following is still present in FreeBSD 11, but like the previous four it should never have been used by third party software, and its value isn't guaranteed to be stable across versions.  It should be removed from FreeBSD
`CTL_P1003_1B_MAXID`

kellda:
Fixes #454

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -3857,26 +3857,6 @@ pub const SHM_STAT: ::c_int = 13;
 pub const SHM_INFO: ::c_int = 14;
 pub const SHM_ANON: *mut ::c_char = 1 as *mut ::c_char;
 
-// The *_MAXID constants never should've been used outside of the
-// FreeBSD base system.  And with the exception of CTL_P1003_1B_MAXID,
-// they were all removed in svn r262489.  They remain here for backwards
-// compatibility only, and are scheduled to be removed in libc 1.0.0.
-#[doc(hidden)]
-#[deprecated(since = "0.2.54", note = "Removed in FreeBSD 11")]
-pub const CTL_MAXID: ::c_int = 10;
-#[doc(hidden)]
-#[deprecated(since = "0.2.54", note = "Removed in FreeBSD 11")]
-pub const KERN_MAXID: ::c_int = 38;
-#[doc(hidden)]
-#[deprecated(since = "0.2.54", note = "Removed in FreeBSD 11")]
-pub const HW_MAXID: ::c_int = 13;
-#[doc(hidden)]
-#[deprecated(since = "0.2.54", note = "Removed in FreeBSD 11")]
-pub const USER_MAXID: ::c_int = 21;
-#[doc(hidden)]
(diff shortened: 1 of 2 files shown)


src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1039,25 +1039,6 @@ pub const SHM_STAT: ::c_int = 13;
...
-#[deprecated(since = "0.2.54", note = "Removed in FreeBSD 11")]
-pub const KERN_MAXID: ::c_int = 38;
-#[doc(hidden)]
-#[deprecated(since = "0.2.54", note = "Removed in FreeBSD 11")]
-pub const HW_MAXID: ::c_int = 13;
-#[doc(hidden)]
-#[deprecated(since = "0.2.54", note = "Removed in FreeBSD 11")]
-pub const USER_MAXID: ::c_int = 21;
-#[doc(hidden)]
-pub const CTL_P1003_1B_MAXID: ::c_int = 26;
```
JohnTitor, line 1059: This constant isn't deprecated.

asomers: But it should be.  And in fact, it has been removed in FreeBSD 13.  How about we remove it in this PR, and I submit another that deprecates it immediately (a non breaking change)?

JohnTitor: Generally I'd like to avoid a breaking change unless it breaks the build. So, the problem here is that we don't have any policies for supported FreeBSD versions like #1412 IIRC. We have hidden the items and added deprecation notes, I feel it's enough here, isn't it? One possible way is to tweak the deprecation notes like "We plan to remove this in 0.2.7x" so that users notice the removal.
Deprecating `CTL_P1003_1B_MAXID` is welcome, of course.

asomers: In this case, relying on those constants could cause misbehavior at runtime, which IMHO is worse than a broken build, because it's harder to detect.  There is precedent for removing similar constants which vary from OS release to OS release: 215f095601637ee7b5c73f2d9d5fc67d9a3f843e

JohnTitor: In 215f095601637ee7b5c73f2d9d5fc67d9a3f843e, we told the user to open an issue in the deprecation notes as I suggested above. But we don't do it yet for `CTL_MAXID` and others.

tgross35:
@kellda `main` is now intended for 1.0. The changes look good to me, are you able to rebase this?

@rustbot author

tgross35 (approved):
Thanks! Could you squash the commits? LGTM after that

tgross35:
I should be able to merge this after CI gets fixed.

@asomers could you ack the change?

asomers (approved):

PR #1834, opened 2020-07-24
Add `PTRACE_EVENT_STOP`

GabrielMajeri:
Exposes this constant, which I presume was not enabled because the `glibc` version it required was too new?

### DIFF
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -981,8 +981,6 @@ pub const PTRACE_EVENT_EXEC: ::c_int = 4;
 pub const PTRACE_EVENT_VFORK_DONE: ::c_int = 5;
 pub const PTRACE_EVENT_EXIT: ::c_int = 6;
 pub const PTRACE_EVENT_SECCOMP: ::c_int = 7;
-// PTRACE_EVENT_STOP was added to glibc in 2.26
-// pub const PTRACE_EVENT_STOP: ::c_int = 128;
 
 pub const __WNOTHREAD: ::c_int = 0x20000000;
 pub const __WALL: ::c_int = 0x40000000;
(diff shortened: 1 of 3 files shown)


src/unix/linux_like/mod.rs
```diff
@@ -982,7 +982,7 @@ pub const PTRACE_EVENT_VFORK_DONE: ::c_int = 5;
 pub const PTRACE_EVENT_EXIT: ::c_int = 6;
 pub const PTRACE_EVENT_SECCOMP: ::c_int = 7;
 // PTRACE_EVENT_STOP was added to glibc in 2.26
-// pub const PTRACE_EVENT_STOP: ::c_int = 128;
+pub const PTRACE_EVENT_STOP: ::c_int = 128;
```
JohnTitor, line 985: Did you check emscripten has this? It seems it doesn't.

GabrielMajeri: I've added a `cfg` to disable it on emscripten (their version of `musl` hasn't been updated to have this yet).

JohnTitor: You should move the item to `/android` and `/linux` instead of using `cfg`.

GabrielMajeri: Thanks for the suggestion, I've moved them to the respective files.

src/unix/linux_like/mod.rs
```diff
@@ -982,7 +982,8 @@ pub const PTRACE_EVENT_VFORK_DONE: ::c_int = 5;
 pub const PTRACE_EVENT_EXIT: ::c_int = 6;
 pub const PTRACE_EVENT_SECCOMP: ::c_int = 7;
 // PTRACE_EVENT_STOP was added to glibc in 2.26
-// pub const PTRACE_EVENT_STOP: ::c_int = 128;
+// emscripten doesn't have this yet
+//pub const PTRACE_EVENT_STOP: ::c_int = 128;
```
JohnTitor, line 986: This should be removed now.

GabrielMajeri: Done!

PR #1836, opened 2020-07-24
Declare `faccessat` for Emscripten.

sunfishcode:
This declaration is the same as for other platforms.


JohnTitor:
Some of CI failures should be spurious but emscripten's one seems real failure.

sunfishcode:
Looks like I was mistaken, Emscripten doesn't currently export `faccessat`.

PR #1838, opened 2020-07-25
Provide support for `least` and `fast` integer types

### LINKED ISSUE #631: Provide support for `least` and `fast` integer types
libc currently provides a number of convenient integer type aliases. However, the `least` and `fast` integer types (e.g. `int_fast16_t`, `uint_least8_t`, [etc.](http://en.cppreference.com/w/c/types/integer)) aren't currently included. Is there a reason for their exclusion, or is this just awaiting an implementation?

kellda:
Fixes #631 

For now only `least` is implemented, I'm not sure how to implement `fast`

### DIFF
--- a/src/fixed_width_ints.rs
+++ b/src/fixed_width_ints.rs
@@ -18,3 +18,12 @@ pub type uint16_t = u16;
 pub type uint32_t = u32;
 #[deprecated(since = "0.2.55", note = "Use u64 instead.")]
 pub type uint64_t = u64;
+
+pub type int_least8_t = i8;
+pub type int_least16_t = i16;
+pub type int_least32_t = i32;
+pub type int_least64_t = i64;
+pub type uint_least8_t = u8;
+pub type uint_least16_t = u16;
+pub type uint_least32_t = u32;
+pub type uint_least64_t = u64;


src/fixed_width_ints.rs
```diff
@@ -18,3 +18,12 @@ pub type uint16_t = u16;
...
 pub type uint64_t = u64;
+
+pub type int_least8_t = i8;
+pub type int_least16_t = i16;
+pub type int_least32_t = i32;
+pub type int_least64_t = i64;
+pub type uint_least8_t = u8;
+pub type uint_least16_t = u16;
+pub type uint_least32_t = u32;
+pub type uint_least64_t = u64;
```
JohnTitor, line 29: This should be in platfrom-dependent module.

kellda: I don't think it makes sense to duplicate it in platform-dependent modules because:
- It is the same for every supported target.
- The typedef being different than what is above would mean that some primitive type isn't available at all on a platform, which is very unlikely if ever it exists.
- If such a platform existed, it would be really unlikely to be targeted by Rust, as not beign able to use e.g. u8 would be very hard to handle.
- In any case it can always be moved later if ever a platform need other typedefs

JohnTitor: Hm, then I don't think we need this with the same reason as #1304.

kellda: @alexcrichton [said](https://github.com/rust-lang/libc/issues/631#issuecomment-311832151) it was awaiting an implementation, but I'm fine with whatever decision is taken

PR #1842, opened 2020-07-27
Create macros for defining Copy/Clone on enums and struct(), use for psp

MrAwesome:
I tried to keep the macros simple and just define the basic things needed, happy to modify them in whatever way feels best.

Tested this by compiling and running a psp EBOOT.PBP including stdlib, so I'm pretty certain this at least isn't introducing anything catastrophic. I noticed we don't have any entries in libc-test, and neither does switch. Is that something I should add? Not sure what it would look like but I'm happy to crib it from another type if it makes sense.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -121,6 +121,36 @@ macro_rules! s_no_extra_traits {
     );
 }
 
+#[allow(unused_macros)]
+macro_rules! e {
+    ($($(#[$attr:meta])* pub enum $i:ident { $($field:tt)* })*) => ($(
+        __item! {
+            #[cfg_attr(feature = "extra_traits", derive(Debug, Eq, Hash, PartialEq))]
+            $(#[$attr])*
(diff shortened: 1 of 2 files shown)


JohnTitor:
Overall looks good to me! Could you fix the style issue that CI reports? It blocks the test for this target.

>  I noticed we don't have any entries in libc-test, and neither does switch. Is that something I should add? Not sure what it would look like but I'm happy to crib it from another type if it makes sense.

Certainly we don't have it but we do test the build here: https://github.com/rust-lang/libc/blob/master/ci/build.sh

We don't because libc-test (i.e. `run.sh`) needs a virtual environment (see other platforms for an example) and I'm not familiar with the targets. If you want to go ahead, I'm happy to review and accept it :)

JohnTitor:
Okay, now CI is green! Could you squash commits? r=me once it's done.

PR #1844, opened 2020-07-27
Deprecate CTL_P1003_1B_MAXID

asomers:
It's been removed in FreeBSD 13 (svn r363622), and never had any
legitimate use outside of the base system anyway.


asomers:
cc @kellda

PR #1846, opened 2020-07-28
Update musl version to 1.2.0

JohnTitor:


JohnTitor:
So, we need to change the `time_t` definition on musl 32-bit targets according to "time64": https://musl.libc.org/time64.html
I think we should do it at some point but should announce it via deprecation notes or something else before changing.

JohnTitor:
I'm going to close this since the change for `time_t` needs some time.

PR #1849, opened 2020-07-28
Add sys/personality.h constants for Linux

nbaksalyar:
These consts are used for the `personality(2)` syscall and can be found in `/usr/include/sys/personality.h`

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -771,6 +771,10 @@ pub const NDA_MASTER: ::c_ushort = 9;
 pub const NDA_LINK_NETNSID: ::c_ushort = 10;
 pub const NDA_SRC_VNI: ::c_ushort = 11;
 
+// linux/personality.h
+pub const UNAME26: ::c_int = 0x0020000;
+pub const FDPIC_FUNCPTRS: ::c_int = 0x0080000;
+
 // linux/if_addr.h
 pub const IFA_FLAGS: ::c_ushort = 8;
 
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -961,6 +961,17 @@ pub const WEXITED: ::c_int = 0x00000004;
 pub const WCONTINUED: ::c_int = 0x00000008;
 pub const WNOWAIT: ::c_int = 0x01000000;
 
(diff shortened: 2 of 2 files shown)


JohnTitor:
It seems musl and emscripten don't have `UNAME26` and `FDPIC_FUNCPTRS`.

https://github.com/bminor/musl/blob/b4b1e10364c8737a632be61582e05a8d3acf5690/include/sys/personality.h#L8-L16

https://github.com/emscripten-core/emscripten/blob/2bca083cbbd5a4133db61fbd74d04f7feecfa907/system/include/libc/sys/personality.h#L8-L16

You can move them to the platforms that you want to use.

nbaksalyar:
> You can move them to the platforms that you want to use.

Thanks! Will do

nbaksalyar:
Hmm. CI fails at the `StyleAndDocs` step with the following error:

```
+ cargo doc --target aarch64-unknown-cloudabi --no-default-features --features extra_traits
   Compiling libc v0.2.74 (/home/vsts/work/1/s)
 Documenting libc v0.2.74 (/home/vsts/work/1/s)
error[E0463]: can't find crate for `core`
  |
  = note: the `aarch64-unknown-cloudabi` target may not be installed
```

Any ideas why this might happen?

JohnTitor:
The problem here is that `rust-src` for that target is missing in today's nightly. I'll fix it soon, one sec.

JohnTitor:
Well, I noticed today's nightly is missing `rust-src` for a lot of targets, testing in #1850, it's... painful :/ Could you wait for fixing this issue on upstream?

nbaksalyar:
Sure, thanks for your help!

JohnTitor:
So, it turned out the cargo failure, not component's. I missed this...:
```
+ cargo doc --target aarch64-unknown-cloudabi -Z build-std=core,alloc --no-default-features --features extra_traits
error: failed to read `/usr/share/rust/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

This should be fixed by https://github.com/rust-lang/cargo/pull/8558 and I confirmed the things work well in today's nightly.
Let's re-trigger CI to make it green.

nbaksalyar:
Awesome, thanks :)

PR #1851, opened 2020-07-29
Add DevkitPPC support

DarkKirb:
DevkitPPC is the de-facto Homebrew development kit for the Nintendo Gamecube, Wii and Wii U consoles. It is based on a gcc+binutils+newlib toolchain.

This pull request adds initial support for these 3 consoles.

DevkitPPC does not support unix-style sockets natively, meaning that bindings
to these functions was removed for powerpc targets with "nintendo" as
vendor

Suggested target json files:

Nintendo Gamecube:
```
{
  "arch": "powerpc",
  "data-layout": "E-m:e-p:32:32-i64:64-n32",
  "dynamic-linking": false,
  "env": "newlib",
  "executables": true,
  "has-elf-tls": false,
  "has-rpath": true,
  "linker-flavor": "gcc",
  "llvm-target": "powerpc-eabi",
  "max-atomic-width": 32,
  "os": "dolphin",
  "target-c-int-width": "32",
  "target-endian": "big",
  "target-family": "unix",
  "target-mcount": "_mcount",
  "target-pointer-width": "32",
  "vendor": "nintendo"
}
```

Nintendo Wii:
```
{
  "arch": "powerpc",
  "data-layout": "E-m:e-p:32:32-i64:64-n32",
  "dynamic-linking": false,
  "env": "newlib",
  "executables": true,
  "has-elf-tls": false,
  "has-rpath": true,
  "linker-flavor": "gcc",
  "llvm-target": "powerpc-eabi",
  "max-atomic-width": 32,
  "os": "revolution",
  "target-c-int-width": "32",
  "target-endian": "big",
  "target-family": "unix",
  "target-mcount": "_mcount",
  "target-pointer-width": "32",
  "vendor": "nintendo"
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ newer Rust features are only available on newer Rust toolchains:
 | `extra_traits`       |  1.25.0 |
 | `core::ffi::c_void`  |  1.30.0 |
 | `repr(packed(N))`    |  1.33.0 |
+| `cfg(target_vendor)` |  1.33.0 |
 
 ## Platform support
 
--- a/build.rs
+++ b/build.rs
@@ -65,9 +65,10 @@ fn main() {
         println!("cargo:rustc-cfg=libc_core_cvoid");
     }
 
-    // Rust >= 1.33 supports repr(packed(N))
+    // Rust >= 1.33 supports repr(packed(N)) and cfg(target_vendor).
     if rustc_minor_ver >= 33 || rustc_dep_of_std {
         println!("cargo:rustc-cfg=libc_packedN");
+        println!("cargo:rustc-cfg=libc_cfg_target_vendor");
     }
 
     // #[thread_local] is currently unstable
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -596,9 +596,13 @@ extern "C" {
     pub fn getchar_unlocked() -> ::c_int;
     pub fn putchar_unlocked(c: ::c_int) -> ::c_int;
 
+    #[cfg(not(all(libc_cfg_target_vendor, target_arch = "powerpc",
+          target_vendor = "nintendo")))]
     #[cfg_attr(target_os = "netbsd", link_name = "__socket30")]
     #[cfg_attr(target_os = "illumos", link_name = "__xnet_socket")]
     pub fn socket(domain: ::c_int, ty: ::c_int, protocol: ::c_int) -> ::c_int;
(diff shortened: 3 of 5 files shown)


JohnTitor:
Looks good, let's re-run CI...

JohnTitor:
It seems `target_vendor` is the newer change so we should gate it with the feature like `const-extern-fn`.

DarkKirb:
I added another commit to gate the cfg checks behind a feature as well

JohnTitor:
I think this would be better: https://github.com/JohnTitor/libc/commit/70e1129fc700776a9e36588bf16936dbb2c1d807
This doesn't require the feature flag and it only applies `cfg` if Rust > 1.33.0.
You could tweak the condition and make sure `rustc ci/style.rs && ./style src` passes (I think it suggests using `cfg_if!` macros).

DarkKirb:
it doesn't appear to do that, probably because i'm only using one layer of conditionals inside the cfg

also i added another commit that removes the addrinfo struct outright on devkitpro, because it's not defined anywhere in the header files

JohnTitor:
The changes look good to me. Does this work fine on your environment? If so, r=me once the commits are squashed.

DarkKirb:
works fine on my end
r=@JohnTitor

DarkKirb:
works fine on my end
r? @JohnTitor

JohnTitor:
Hmm, it seems spurious:
```
+ cargo test --manifest-path libc-test/Cargo.toml --target aarch64-linux-android
   Compiling libc v0.2.74 (/checkout)
   Compiling libc-test v0.1.0 (/checkout/libc-test)
    Finished test [unoptimized + debuginfo] target(s) in 56.76s
     Running target/aarch64-linux-android/debug/deps/cmsg-8b6337b17e5d316d
/checkout/target/aarch64-linux-android/debug/deps/cmsg-8b6337b17e5d316d: 1 file pushed, 0 skipped. 10.8 MB/s (5735248 bytes in 0.505s)
status: exit code: 101
stdout ---

stderr ---
thread 'main' panicked at 'No option 'h' defined', /cargo/registry/src/github.com-1ecc6299db9ec823/getopts-0.2.21/src/lib.rs:799:21

thread 'main' panicked at 'failed to find successful test run', /tmp/runtest.rs:45:9
error: test failed, to rerun pass '--test cmsg'
```

Re-triggering CI...

PR #1852, opened 2020-08-02
add ucred(3C) support for illumos and Solaris systems

jclulow:
This series of routines allows the caller to determine the credentials
of another process by pid, or of the process on the remote end of a UNIX
domain socket.  The ucred_t is an opaque object with accessor routines,
and must be freed through ucred_free(3C) after use.


jclulow:
`cd libc-test && cargo test` passes all tests on an illumos system.  For reference, the functions here are exposed in this header:

https://code.illumos.org/plugins/gitiles/illumos-gate/+/refs/heads/master/usr/src/head/ucred.h

... and documented in this manual page:

https://illumos.org/man/3C/ucred

PR #1854, opened 2020-08-03
add openpty and forkpty implementation for illumos systems

jclulow:
At time of writing, illumos systems do not provide an implementation of
the openpty() and forkpty() wrappers provided on some other UNIX
systems.  While we expect to grow an implementation, it seems prudent to
provide a compatibility routine here first to unblock illumos support in
the popular nix crate.

### DIFF
--- a/src/unix/solarish/compat.rs
+++ b/src/unix/solarish/compat.rs
@@ -3,6 +3,9 @@
 
 use unix::solarish::*;
 
+const PTEM: &[u8] = b"ptem\0";
+const LDTERM: &[u8] = b"ldterm\0";
+
 pub unsafe fn cfmakeraw(termios: *mut ::termios) {
     (*termios).c_iflag &= !(IMAXBEL
         | IGNBRK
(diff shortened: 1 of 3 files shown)


jclulow:
Inspiration drawn from:

* [Advanced Programming in the UNIX® Environment, Third Edition](http://www.apuebook.com/apue3e.html); Section 19.3 Opening Pseudo-Terminal Devices
* [compat/forkpty-sunos.c](https://github.com/tmux/tmux/blob/master/compat/forkpty-sunos.c) from the tmux source
* manual pages: [pts(7D)](https://illumos.org/man/7D/pts), [ptm(7D)](https://illumos.org/man/7D/ptm), etc

The local libc crate tests and style check pass.

I have been hacking away at the nix crate, and with many substantial changes have managed to get it to build.  Some tests are passing -- most critically, the [tests which exercise `openpty()` and `forkpty()`](https://github.com/nix-rust/nix/blob/96054b6933404b4c1d5dbd49db7ec7c7a26e44de/test/test_pty.rs#L163-L275) pass with this change.

cc: @jasonbking @pfmooney

src/unix/solarish/compat.rs
```diff
@@ -45,3 +48,136 @@ pub unsafe fn cfsetspeed(
...
+    if ::setsid() < 0
+        || ::ioctl(fds, TIOCSCTTY, 0) < 0
+        || ::dup2(fds, 0) < 0
+        || ::dup2(fds, 1) < 0
+        || ::dup2(fds, 2) < 0
+    {
+        // At this stage there are no particularly good ways to handle failure.
+        // Exit as abruptly as possible, using _exit() to avoid messing with any
+        // state still shared with the parent process.
+        ::_exit(1);
```
pfmooney, line 174: Could perhaps use `EXIT_FAILURE` here, rather than the numeric value. *shrug*

JohnTitor:
Thanks both for reviewing! r=me once the nit is resolved.

PR #1858, opened 2020-08-09
Expose si_pid and si_uid from siginfo_t as functions

joshtriplett:
On Linux, siginfo_t cannot expose these fields directly due to
https://github.com/rust-lang/libc/issues/716 , so expose them as
functions, just like si_addr and si_value.

Provide the same functions on other platforms that have these fields
declared, for consistency.

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -322,6 +322,8 @@ cfg_if! {
             si_pid: ::pid_t,
             si_uid: ::uid_t,
             si_status: ::c_int,
+            si_utime: ::c_long,
+            si_stime: ::c_long,
         }
 
         // Internal, for casts to access union fields
@@ -355,6 +357,14 @@ cfg_if! {
 
             pub unsafe fn si_status(&self) -> ::c_int {
                 self.sifields().sigchld.si_status
+            }
+
+            pub unsafe fn si_utime(&self) -> ::c_long {
+                self.sifields().sigchld.si_utime
+            }
+
+            pub unsafe fn si_stime(&self) -> ::c_long {
+                self.sifields().sigchld.si_stime
             }
         }
     }
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -313,6 +313,55 @@ impl siginfo_t {
     }
 }
 
+cfg_if! {
+    if #[cfg(libc_union)] {
+        // Internal, for casts to access union fields
+        #[repr(C)]
+        #[derive(Copy,Clone)]
+        struct sifields_sigchld {
+            si_pid: ::pid_t,
+            si_uid: ::uid_t,
+            si_status: ::c_int,
+        }
+
+        // Internal, for casts to access union fields
+        #[repr(C)]
+        union sifields {
+            _align_pointer: *mut ::c_void,
+            sigchld: sifields_sigchld,
+        }
+
+        // Internal, for casts to access union fields. Note that some variants
+        // of sifields start with a pointer, which makes the alignment of
+        // sifields vary on 32-bit and 64-bit architectures.
+        #[repr(C)]
+        struct siginfo_f {
+            _si_signo: ::c_int,
+            _si_errno: ::c_int,
+            _si_code: ::c_int,
+            sifields: sifields,
+        }
+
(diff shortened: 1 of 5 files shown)


joshtriplett:
Pushed a new version to fix structure alignment.

joshtriplett:
Currently reworking this a bit more, to handle the alignment correctly on both 64-bit and 32-bit platforms.

joshtriplett:
Re-pushed with style issue (line length) fixed. Simplified in the process.

joshtriplett:
Alright, looks like just two builds failed here this time.

mipsel-unknown-linux-musl appears spurious: it failed to download something. Needs a retry.

And the builder running rust 1.13 failed, because rust 1.13 didn't support `union`. I'll switch the pre-existing functions back to not use `union`, and make the new functions conditional on `#[cfg(libc_union)]`.

joshtriplett:
Fixed style for conditional compilation.

src/unix/linux_like/linux/gnu/mod.rs
```diff
@@ -313,6 +313,55 @@ impl siginfo_t {
...
 
+cfg_if! {
+    if #[cfg(libc_union)] {
+        // Internal, for casts to access union fields
+        #[repr(C)]
+        #[derive(Copy,Clone)]
+        struct sifields_sigchld {
+            si_pid: ::pid_t,
+            si_uid: ::uid_t,
+            si_status: ::c_int,
```
pickfire, line 324: What about the timings?
```c
	struct {
		__kernel_pid_t _pid;	/* which child */
		__kernel_uid32_t _uid;	/* sender's uid */
		int _status;		/* exit code */
		__ARCH_SI_CLOCK_T _utime;
		__ARCH_SI_CLOCK_T _stime;
	} _sigchld;
```

src/unix/linux_like/linux/gnu/mod.rs
```diff
@@ -313,6 +313,55 @@ impl siginfo_t {
...
+        }
+
+        // Internal, for casts to access union fields. Note that some variants
+        // of sifields start with a pointer, which makes the alignment of
+        // sifields vary on 32-bit and 64-bit architectures.
+        #[repr(C)]
+        struct siginfo_f {
+            _si_signo: ::c_int,
+            _si_errno: ::c_int,
+            _si_code: ::c_int,
```
pickfire, line 341: These two fields may be swapped based on the original source code.
```c
#ifndef __ARCH_HAS_SWAPPED_SIGINFO
#define __SIGINFO 			\
struct {				\
	int si_signo;			\
	int si_errno;			\
	int si_code;			\
	union __sifields _sifields;	\
}
#else
#define __SIGINFO 			\
struct {				\
	int si_signo;			\
	int si_code;			\
	int si_errno;			\
	union __sifields _sifields;	\
}
#endif /* __ARCH_HAS_SWAPPED_SIGINFO */
```
Should we ignore that?

joshtriplett: They're the same size, so for the purposes of this cast it doesn't matter. If you'd like, I could change it to a slice of 3 c_int values to make that more explicit.

src/unix/linux_like/linux/gnu/mod.rs
pickfire: @joshtriplett I think you missed this.

src/unix/linux_like/linux/gnu/mod.rs
pickfire: Yes, they have the same size. But I am not talking about the size, I mean the C code have different orderings for different architectures but we don't have it. Shouldn't we have it or is it automatically solved?

joshtriplett: @pickfire We already have the different ordering for the actual base structure (prior to sifields); only MIPS sets `__ARCH_HAS_SWAPPED_SIGINFO`, and the MIPS targets already define the structure with the fields flipped. The structure defined *here* is never used to access the base fields, and just needs to skip past them, so their order doesn't matter; I'm suggesting that I can change them to `_base: [c_int; 3]` to make it explicit that the names don't matter.

src/unix/linux_like/linux/gnu/mod.rs
joshtriplett: I saw; just hadn't responded yet. I can add these fields as well.

PR #1859, opened 2020-08-09
Add W_EXITCODE to construct an exit code

joshtriplett:
On Linux, `sys/wait.h` defines a `W_EXITCODE` macro to construct an exit
code from a return value and a signal number. Provide an equivalent
function.

### DIFF
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -1258,6 +1258,10 @@ f! {
         (ret << 8) | sig
     }
 
+    pub fn W_STOPCODE(sig: ::c_int) -> ::c_int {
+        (sig << 8) | 0x7f
+    }
+
     pub fn QCMD(cmd: ::c_int, type_: ::c_int) -> ::c_int {
         (cmd << 8) | (type_ & 0x00ff)
     }
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -1254,6 +1254,10 @@ f! {
         (status & 0x80) != 0
     }
 
+    pub fn W_EXITCODE(ret: ::c_int, sig: ::c_int) -> ::c_int {
+        (ret << 8) | sig
+    }
+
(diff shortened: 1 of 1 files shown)


src/unix/linux_like/mod.rs
```diff
@@ -1254,6 +1254,10 @@ f! {
         (status & 0x80) != 0
     }
 
+    pub fn W_EXITCODE(ret: ::c_int, sig: ::c_int) -> ::c_int {
+        (ret << 8) | sig
+    }
+
```
pickfire, line 1260: While doing this, maybe you could also add `W_STOPCODE`?
```suggestion

    pub fn W_STOPCODE(sig: ::c_int) -> ::c_int {
        (sig << 8) | 0x7f
    }

```

joshtriplett: Done.

JohnTitor:
These stuff are really part of musl? I cannot find any reference for musl and emscripten.

mati865:
AFAIK those two are non standard extensions.
Definitely not available in musl.

pickfire:
Oh, I just realized it isn't in musl.

joshtriplett:
These are ultimately designed to match kernel ABI: they construct exit codes in the form that kernel syscalls (like `waitpid` or `wait4`) would return. They're the reverse of things like `WIFEXITED` and `WEXITSTATUS`. They're Linux-specific; other platforms that have them would need to define them consistently with define exit codes and `WIFEXITED` and `WEXITSTATUS` and similar.

joshtriplett:
Any thoughts on this? Does it seem reasonable to merge, now that it includes `W_STOPCODE` as well?

JohnTitor:
> These are ultimately designed to match the Linux kernel ABI

So, this should be fine to use in musl or emscripten, right? If so, I don't have objections here and would be happy to merge.

joshtriplett:
It should be fine to use on any Linux libc, including musl.

PR #1861, opened 2020-08-10
Define P_PIDFD, used with waitid

joshtriplett:
Linux defines a waitid type `P_PIDFD`, for use with process file
descriptors (`pidfd`). Add that constant.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -1486,6 +1486,7 @@ fn test_android(target: &str) {
                 "linux/seccomp.h",
                 "linux/sched.h",
                 "linux/sockios.h",
+                "linux/wait.h",
 
     }
 
@@ -1555,6 +1556,9 @@ fn test_android(target: &str) {
             "SIG_DFL" | "SIG_ERR" | "SIG_IGN" => true, // sighandler_t weirdness
             // FIXME: deprecated - removed in glibc 2.26
             "SIGUNUSED" => true,
+
+            // Needs a newer Android SDK for the definition
+            "P_PIDFD" => true,
 
             _ => false,
         }
@@ -2442,6 +2446,7 @@ fn test_linux(target: &str) {
         "linux/seccomp.h",
         "linux/sockios.h",
         "linux/vm_sockets.h",
+        "linux/wait.h",
         "sys/auxv.h",
         "sys/fanotify.h",
     }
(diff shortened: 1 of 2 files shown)


joshtriplett:
libc-test passed here; looks like the libc headers are outdated on the test server? What's the process for updating those?

joshtriplett:
Re-pushed with a testsuite fix. Add `linux/wait.h` to the C headers used in libc-test, so that it can find the definition of `P_PIDFD`.

joshtriplett:
Looks like Android, musl, and empscripten are unhappy.

I pushed a new change, which should fix Android by adding `linux/wait.h` to that separate header list, and which should fix emscripten by omitting `P_PIDFD` there.

I'm not sure why musl is unhappy.

joshtriplett:
Pushed a style fix to use `cfg_if`.

I could use some help understanding why this is still failing libc-test on some platforms.

JohnTitor:
> I could use some help understanding why this is still failing libc-test on some platforms.

On Android, it's newly added and CI images don't know it: https://android.googlesource.com/platform/bionic/+/9584fa46e9dd7071f8ad870341e7a0ab5ab918e8
musl is the same as Android, we use musl 1.1.24 on CI but it's added in 1.2.0: https://github.com/bminor/musl/commit/f296be74d5ab94f6ccac26e129d96e08f1c20538

I'd recommend to add to the environment that you want to use if you don't know much.

joshtriplett:
I'd like to have this consistently available on Linux platforms. It's a Linux kernel thing, not a C library thing, and I try not to make things gratuitously glibc-specific when they're not glibc extensions (which this isn't). It's defined in `linux/wait.h`, which is a Linux kernel header, not a libc-specific header.

What's the process for updating the CI images? (In the case of musl, that wouldn't *necessarily* require updating musl itself, just the kernel headers.)

JohnTitor:
> What's the process for updating the CI images?

It needs to update CI config but I'm sure that it's quite hard since musl 1.2.0 has a _breaking change_ called "time64" and android images also have some breakages for our CI iirc.
It's an easiest way just to skip tests for the item with a comment if it's available on the newer platforms.

joshtriplett:
Looks like nightly had an unrelated spurious failure (something missing in rustup), and everything else passes now.

JohnTitor (approved):
The failure is due to rust-lang/rust#75568, indeed unrelated. I'm going to merge this and confirm that CI is green in #1865. Thanks for working on it!

PR #1863, opened 2020-08-16
Add ucontext_t for aarch64-unknown-linux-musl

ifreund:
See: https://github.com/bytecodealliance/wasmtime/issues/2133

### DIFF
--- a/src/unix/linux_like/linux/musl/b64/aarch64/align.rs
+++ b/src/unix/linux_like/linux/musl/b64/aarch64/align.rs
@@ -5,3 +5,21 @@ s_no_extra_traits! {
         priv_: [f32; 8]
     }
 }
+
+s!{
+    pub struct ucontext_t {
+        pub uc_flags: ::c_ulong,
+        pub uc_link: *mut ucontext_t,
+        pub uc_stack: ::stack_t,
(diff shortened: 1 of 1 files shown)


JohnTitor:
> Use no_extra_traits due to array longer than 32

Did you test with this workaround? I guess we could auto-derive it here as well...

https://github.com/rust-lang/libc/blob/035a4809428d319e72ff1aaae5c5335f18623836/src/unix/linux_like/linux/gnu/b64/aarch64/align.rs#L25-L27

ifreund:
> Did you test with this workaround? I guess we could auto-derive it here as well...

I tried it but the math doesn't work out in this case. The prime factorization of `36 + 512` is `2 * 2 * 137`.

JohnTitor:
Ah, makes sense. I'd still prefer to auto-derive it as this will make the difference between arch/env and users will be confused.

JohnTitor:
(Note that CI will fail anyway until #1865 gets merged.)

JohnTitor:
Looks good except for spurious failures. Could you rebase and squash commits into one?

ifreund:
@JohnTitor squashed!

JohnTitor (approved):
Thanks a lot!

PR #1855, opened 2020-08-07
add fmemopen, open_memstream, and open_wmemstream for POSIX.1-2008

jedbrown:
This is useful for interfacing with C libraries that can write output to `FILE` objects, for example.

PR #1867, opened 2020-08-19
Added clock_getcpuclockid to more targets

xonatius:
Draft for now - will make a normal pull request once build passes.


src/unix/bsd/apple/mod.rs
```diff
@@ -3297,6 +3297,10 @@ extern "C" {
     pub fn fchflags(fd: ::c_int, flags: ::c_uint) -> ::c_int;
     pub fn clock_getres(clk_id: ::clockid_t, tp: *mut ::timespec) -> ::c_int;
     pub fn clock_gettime(clk_id: ::clockid_t, tp: *mut ::timespec) -> ::c_int;
+    pub fn clock_getcpuclockid(
+        pid: ::pid_t,
+        clk_id: *mut ::clockid_t,
+    ) -> ::c_int;
```
JohnTitor, line 3303: I couldn't find any reference for `clock_getcpuclockid` on macOS. Did you confirm if it exists?

src/unix/bsd/freebsdlike/dragonfly/mod.rs
```diff
@@ -1071,6 +1071,10 @@ extern "C" {
         clk_id: ::clockid_t,
         tp: *const ::timespec,
     ) -> ::c_int;
+    pub fn clock_getcpuclockid(
+        pid: ::pid_t,
+        clk_id: *mut ::clockid_t,
+    ) -> ::c_int;
```
JohnTitor, line 1077: I think you can just place it on `freebsdlike/mod.rs`?

src/unix/linux_like/emscripten/mod.rs
```diff
@@ -1784,6 +1784,10 @@ extern "C" {
     pub fn srand(seed: ::c_uint);
 
     pub fn gettimeofday(tp: *mut ::timeval, tz: *mut ::c_void) -> ::c_int;
+    pub fn clock_getcpuclockid(
+        pid: ::pid_t,
+        clk_id: *mut ::clockid_t,
+    ) -> ::c_int;
```
JohnTitor, line 1790: It should be available on Android as well, let's move it to `linux_like/mod.rs`.

xonatius: I originally tried to put it in `linux_like/mod.rs`, but the build for android failed due to `clock_getcpuclockid` being missing.

src/unix/bsd/freebsdlike/dragonfly/mod.rs
xonatius: Sure, let me try that.

src/unix/bsd/apple/mod.rs
xonatius: Indeed, looks like macOS doesn't support it. Removed.

src/unix/linux_like/emscripten/mod.rs
JohnTitor: If it's on aarch64 only, then the cause is not absence, but #1765 likely.

src/unix/bsd/freebsdlike/dragonfly/mod.rs
xonatius: Should all other `clock_*` functions be in `freebsdlike/mod.rs` too then? Looks like the only difference between freebsd and dragonfly is that one uses `c_int` for `clockid_t` and the other `c_ulong`. Does it make sense to have functions in  `freebsdlike/mod.rs`, while `clockid_t` in more platform specific place?

src/unix/linux_like/emscripten/mod.rs
xonatius: Ah, yes, it was [aarch64](https://dev.azure.com/rust-lang2/libc/_build/results?buildId=2966&view=logs&j=c8d582de-107e-5fbd-122b-5384eb107855&t=7bb3b962-ee23-58e7-68da-769fcc260c0d) . In that case - should I keep it as is for now and move it to `linux_like/mod.rs` once the issue is resolved?

src/unix/bsd/freebsdlike/dragonfly/mod.rs
JohnTitor: > Should all other clock_* functions be in freebsdlike/mod.rs too then?

Yeah, sounds good.

> Does it make sense to have functions in freebsdlike/mod.rs, while clockid_t in more platform specific place?

It should be fine as we `pub use` all the items.

src/unix/linux_like/emscripten/mod.rs
JohnTitor: I'd like to skip the tests for now, could you tweak `libc-test/build.rs` to do so? (see https://github.com/rust-lang/libc/pull/1855/files#diff-747c41e91bd79d9208c2953b72047176 for example.)

xonatius: Thanks for the example, done!

src/unix/bsd/freebsdlike/dragonfly/mod.rs
xonatius: Done!

PR #1869, opened 2020-08-20
libc can't use derive(Copy,Clone) because it doesn't work in rustc

joshtriplett:


joshtriplett:
Needed to unbreak rustc builds.

JohnTitor:
Thanks! I'll merge it and make a new release once CI passes.

JohnTitor (approved):
The remaining builders are the semver checks, I think we could just ignore here as it's urgent.

PR #1870, opened 2020-08-20
Make some inline functions like WIFEXITED and WEXITSTATUS const and safe

joshtriplett:


JohnTitor:
I'm quite not sure if we could allow the safe things here; @alexcrichton if you have some time, could you review this? I assume you're familiar with that.

alexcrichton:
AFAIK this should be fine to do, there's nothing technically preventing this. I believe the original decision for these functions was uniformity in `libc`, but that's not necessarily the best thing to optimize for here.

JohnTitor (approved):
Thanks, Alex! The macro itself looks good to me, let's go ahead :) And thanks to Josh for implementing it!

joshtriplett:
Awesome, thanks @JohnTitor and @alexcrichton.

PR #1875, opened 2020-08-23
Dragonfly: expose __errno_location() rather than define __error().

cmusser:
This takes advantage of a recent addition to DragonFly that makes it possible to access errno (which is thread-local) without any dependence on the thread_local Rust feature. I tested it out by using it in the Nix crate. Nix had a C language shim, which I removed and replaced by using this modified version of libc.


JohnTitor:
I think this will break the backward compatibility, right? Then I'd like to avoid just removing it but add a deprecation note for a while then drop it.

cmusser:
Good point. I suspect not many people use it, but no easy way to tell. I added the errno module back, along with its reference. The function has a deprecation warning now.

cmusser:
Do these changes look acceptable?

JohnTitor (approved):
Sorry for the delay, it looks good now! Thanks <3

cmusser:
awesome thanks!

PR #1878, opened 2020-08-25
process_vm_readv, process_vm_writev should be available on android

rupansh:
process_vm_readv and process_vm_writev generally return -EPERM(-1) when used in android, however the syscalls work fine when running under privileged shell.
The following was tested(with modified nix) on Android 9/10 running Linux Kernel 4.9 on a SDM 845 phone:

```
fn read_mem_real(pid: i32, addr: usize, vsize: usize) -> Result<Vec<u8>, String> {
    let mut res = vec![0; vsize];
    let riovec: [RemoteIoVec; 1] = [RemoteIoVec {base: addr, len: vsize}];

    match process_vm_readv(Pid::from_raw(pid), &[IoVec::from_mut_slice(&mut res)], &riovec) {
        Err(e) => {
            log::info!("ERR {}", e);
            return Err(e.to_string());
        },
        Ok(_) => {
            return Ok(res);
        }
    }
}
```

This will work fine when run under su shell.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -1610,6 +1610,8 @@ fn test_android(target: &str) {
             | "open_memstream"
             | "open_wmemstream"
             | "clock_getcpuclockid"
+            | "process_vm_readv"
+            | "process_vm_writev"
                 if aarch64 =>
             {
                 true
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -2394,7 +2394,7 @@ extern "C" {
         remote_iov: *const ::iovec,
         riovcnt: ::c_ulong,
         flags: ::c_ulong,
-    ) -> isize;
+    ) -> ::ssize_t;
     pub fn process_vm_writev(
         pid: ::pid_t,
         local_iov: *const ::iovec,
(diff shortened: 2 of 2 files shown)


joshtriplett:
Closing and reopening to re-run CI.

rupansh:
@joshtriplett I am not sure why its failing. I used android-ndk's aarch64-linux-android27-clang

joshtriplett:
On Fri, Oct 02, 2020 at 07:46:05PM -0700, Rupansh wrote:
>  @joshtriplett I am not sure why its failing. I used android-ndk's aarch64-linux-android27-clang

Since it succeeds on other Android targets, and fails on aarch64, is it
possible that the Android SDK for aarch64 doesn't have that syscall in
its headers yet?

rupansh:
> On Fri, Oct 02, 2020 at 07:46:05PM -0700, Rupansh wrote: @joshtriplett I am not sure why its failing. I used android-ndk's aarch64-linux-android27-clang
> Since it succeeds on other Android targets, and fails on aarch64, is it possible that the Android SDK for aarch64 doesn't have that syscall in its headers yet?

sysroot/usr/include/sys/uio.h has it in my case for android27. It was added in a later version. android21(Lollipop) is pretty old now!

JohnTitor (commented):
The failure on AArch64 should be related to #1765. Could you skip the test for it like https://github.com/rust-lang/libc/pull/1867/files#diff-aedce4e8d381f53dae77653609e3174da7ee2a184308251ad33f7a47bebabe14R1593-R1608?

JohnTitor:
Thanks!
@bors r+

PR #1880, opened 2020-08-28
Add IOV_MAX and UIO_MAXIOV constants

Thomasdezeeuw:
These constant can be used to determine the maximum number of iovecs can
be passed to functions like readv/writev.

Linux like uses UIO_MAXIOV, while the BSD family uses IOV_MAX.


Thomasdezeeuw:
I don't know how to add tests for these constants. I also couldn't run the tests locally as it failed with a huge list of errors.

JohnTitor (approved):
Looks good, thanks!

PR #1881, opened 2020-08-30
Add support for Apple's O_SYMLINK flag

JohnColanduoni:
Adds support for macOS/iOS's `O_SYMLINK` flag, which allows opening symlinks themselves (instead of their targets).


JohnTitor (approved):
Looks good, thanks!

PR #1882, opened 2020-08-31
Add missing syscall numbers for aarch64-musl

alindima:
The crate is missing some syscall numbers for linux aarch64 with musl.
I added the corresponding numbers.


JohnTitor (approved):
Looks good, thanks!

PR #1886, opened 2020-09-04
Define several more constants for WASI.

sunfishcode:
Define `O_NOCTTY`, `R_OK`, `POLLIN`, and related constants using the
current values from WASI libc.


JohnTitor (approved):
Looks good, thanks!

PR #1889, opened 2020-09-11
Add more constants for Linux like OS

LinkTed:
I added more constants for Linux like OS. Most of the new constants are `IP_` and `IPV6_` related.


LinkTed:
I have to check if these constants exits in Linux musl.

joshtriplett:
@LinkTed CI seems happy with it, which means any constants that this PR adds to musl targets are also available to C on musl in a matching way.

Do you have a specific concern that CI wouldn't have caught here?

LinkTed:
Oh, I did not know that. Okay if the Linux `musl` is also tested then it should be fine.

PR #1891, opened 2020-09-13
Use safe_f! consistently across platforms

xonatius:
The pr #1870 introduced safe_f! macro, which made some functions like WIFEXITED and WEXITSTATUS const and safe on linux_like platform only, which causes inconsistency when trying to use those functions in crates compiled across multiple platforms, as using unsafe on those functions will generate unused_unsafe warning on linux platforms and lack of unsafe block will fail compilation on non-linux platforms.

To avoid the inconsistency, this commit applies the same macro for all the same functions on other platforms too.

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -3158,42 +3158,6 @@ f! {
         }
     }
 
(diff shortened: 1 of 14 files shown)


JohnTitor (approved):
That's a reasonable point, thanks!

PR #1892, opened 2020-09-13
Remove `af_alg_iv::as_slice` (which is unsound) and modify trait implementations that depended on it.

eduardosm:
Fixes https://github.com/rust-lang/libc/issues/1501.

This PR is an alternative to https://github.com/rust-lang/libc/pull/1703. In this case, no trait implementation is removed.

### DIFF
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -583,20 +583,9 @@ cfg_if! {
             }
         }
 
-        impl af_alg_iv {
-            fn as_slice(&self) -> &[u8] {
-                unsafe {
-                    ::core::slice::from_raw_parts(
-                        self.iv.as_ptr(),
-                        self.ivlen as usize
-                    )
-                }
-            }
-        }
-
         impl PartialEq for af_alg_iv {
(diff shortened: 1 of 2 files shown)


joshtriplett:
I think it would be dangerous to ignore iv for things like equality checking. The rest seems fine, but that seems potentially dangerous to me. Two different structure values shouldn't compare equal if they aren't equal.

eduardosm:
`af_alg_iv` is defined as
```rust
#[repr(C)]
pub struct af_alg_iv {
    pub ivlen: u32,
    pub iv: [c_uchar; 0],
}
```

`iv` has length zero, so it does not contain any data that to compare. The current implementation reads up to `ivlen` bytes beyond the struct, but that allows to easily do out-of-bounds reads without `unsafe`. For example:

```rust
let a = af_alg_iv {
    ivlen: 100,
    iv: [],
};
let b = af_alg_iv {
    ivlen: 100,
    iv: [],
};

a == b // Undefined behavior without unsafe keyword
```

joshtriplett:
I realize that, but given that, I think we shouldn't implement `PartialEq` for this structure at all.

JohnTitor:
Closing in favor of #1703.

PR #1893, opened 2020-09-13
Define some `sysconf` constants for WASI.

sunfishcode:
This defines the `_SC_` constants for `sysconf` which the code I'm looking at happens to need.


JohnTitor (approved):
Looks good, thanks!

PR #1894, opened 2020-09-15
linux: riscv32: Add support for the 64-bit time_t RV32 glibc port

alistair23:


JohnTitor:
I'd rubber-stamp this as it seems like copy-pasted. And could you add the `riscv32gc-unknown-linux-gnu` there? I think it's worth to run CI for that target as well.

https://github.com/rust-lang/libc/blob/9c17cad7f0b03e3515f0cd5c56b5592b384cf9d4/ci/build.sh#L225-L227

alistair23:
Done!

src/unix/linux_like/linux/gnu/b32/mod.rs
```diff
@@ -278,6 +278,9 @@ cfg_if! {
     } else if #[cfg(target_arch = "sparc")] {
         mod sparc;
         pub use self::sparc::*;
+    } else if #[cfg(any(target_arch = "riscv32"))] {
```
JohnTitor, line 281:
```suggestion
    } else if #[cfg(target_arch = "riscv32")] {
```

JohnTitor:
Oh, CI found an issue, seems some items are re-defined in the 32-bit space and CI complains about it. We should remove/tweak them following the actual implementation.

JohnTitor:
Ugh, Pipelines is lazy, re-run it...

alistair23:
Thanks, sorry I can't get the CI to run locally. I'll ping you here when I have the tests passing.

alistair23:
Argh! Now I'm getting errors like this:

```rust
error: type alias is never used: `ino_t`
  --> src/unix/linux_like/linux/gnu/b32/riscv32/mod.rs:18:1
   |
18 | pub type ino_t = u64;
   | ^^^^^^^^^^^^^^^^^^^^^
```

I need to overwrite the `ino_t` type though. Any ideas?

JohnTitor:
The easiest way is to use `cfg_if` macro, like:
```rust
cfg_if! {
    if #[cfg(target_arch = "riscv32")] {
        pub type ino_t = u64;
    } else {
        // ...
    }
}
```
at gnu/b32/mod.rs.

alistair23:
Woo! CI is passing.

JohnTitor (approved):
Awesome, thanks!

PR #1895, opened 2020-09-17
Add UIO_MAXIOV constant to uclibc

schultetwin1:
#1880 added UIO_MAXIOV to linux-like libc, but did not add it
to uclibc. This causes build breaks for std against uclibc.

### DIFF
--- a/src/unix/uclibc/mod.rs
+++ b/src/unix/uclibc/mod.rs
@@ -932,6 +932,8 @@ pub const SS_DISABLE: ::c_int = 2;
 
 pub const PATH_MAX: ::c_int = 4096;
 
+pub const UIO_MAXIOV: ::c_int = 1024;
+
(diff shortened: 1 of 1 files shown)


JohnTitor:
Looks good, but CI didn't agree with it. Let's re-run it to see if it still happens.

JohnTitor:
Found out that rust-lang/rust#74532 causes the failure on macOS nightly CI (s390x's is spurious and semver check's is unrelated to this). I'd skip the build check for that target if the upstream doesn't get a fix in a few days.

schultetwin1:
Sounds good! Thanks for looking into that.

JohnTitor (approved):
Okay, it's now green. Thanks!

schultetwin1:
Thank you!!

PR #1899, opened 2020-09-19
Add support for building with static glibc

joshtriplett:
This will need corresponding changes in rust-lang/rust to activate, but
this will make it possible to make those changes.

Note that despite the apparent redundancy in config directives, the link
directives cannot be simplified any further. Attempting to factor out
the checks for `target_feature = "crt-static"` does not work.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,7 +19,7 @@
 // Attributes needed when building as part of the standard library
 #![cfg_attr(
     feature = "rustc-dep-of-std",
-    feature(cfg_target_vendor, link_cfg, no_core)
+    feature(cfg_target_vendor, link_cfg, no_core, static_nobundle)
 )]
 #![cfg_attr(libc_thread_local, feature(thread_local))]
 // Enable extra lints:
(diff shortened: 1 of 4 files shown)


joshtriplett:
Currently working on the corresponding changes in rust-lang/rust. It would really help to have a libc release with this PR included.

joshtriplett:
Still WIP, need to do something similar for the link lines for `dl` and `util` buried inside Linux-specific code.

joshtriplett:
OK, this should work now.

joshtriplett:
Update: had to reverse the order of the libraries so that symbol resolution worked correctly when linking statically.

joshtriplett:
Both of the failures seem entirely unrelated to this PR.

joshtriplett:
Also, as far as I can tell, this change (together with some corresponding changes in rust-lang/rust) will noticeably speed up linking of all Rust binaries on `linux-gnu` targets, by reducing the amount of work the linker needs to do, since it won't need to process some libraries twice. I've tested this by manually re-invoking the linker and timing it, and overall, eliminating repeated libraries seems to provide a ~5% improvement in link time for a "hello world" Rust program.

JohnTitor:
Thanks for the work and sorry for the delay! Looks great to me, let's re-trigger CI to make sure it.

JohnTitor (approved):
The failure is due to timeout and it was green previously so I'd ignore here. Thanks!

PR #1903, opened 2020-09-25
Add struct ip_mreqn to more Linux targets

bartelsielski:
Support was previously added to GNU x86_64 and all musl targets but others were not included because of a [roundtrip issue](https://github.com/rust-lang/libc/issues/1558).

This commit adds support for the ip_mreqn struct on all Linux GNU targets which did not have the roundtrip issue.


JohnTitor:
Looks good, let's re-trigger CI as the current is spurious.

JohnTitor (approved):
It's now green, thanks!

PR #1904, opened 2020-09-27
Fix type definition for c_char on sparc-unknown-linux-gnu

glaubitz:
On sparc-unknown-linux-gnu, char is signed, not unsigned.


JohnTitor (approved):
I assume this won't break anything as no one has complained about it since you introduced it. Thanks!

PR #1906, opened 2020-09-30
Bump version ready to import into Android tree.

qwandor-google:
I need to bump the version number so that I can get the changes from #1905 into the version of the libc crate vendored into the Android tree at https://android.googlesource.com/platform/external/rust/crates/libc/.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "libc"
-version = "0.2.77"
+version = "0.2.78"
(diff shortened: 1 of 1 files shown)


qwandor-google:
The test failure looks like a flake, can we re-run?

JohnTitor:
Josh released 0.2.78 at e3312f5a5ccb6e37f08a60ccc32a7e1cad02eeac so that version should be available now. Thanks anyways!

PR #1910, opened 2020-10-02
Clean-up no core riscv targets

JohnTitor:
Moves `riscv64gc-unknown-linux-gnu` to `RUST_NIGHTLY_LINUX_TARGETS` and adds some other `riscv*` targets to `RUST_LINUX_NO_CORE_TARGETS`.


JohnTitor:
Oh, should fix to make it green.

JohnTitor:
Windows failure should be spurious and can be ignored, let's merge this.

PR #1911, opened 2020-10-04
Fix bootstrap on redox

joshtriplett:
Avoid attempting to enable the static_nobundle feature twice, which
results in rustc error E0636.


JohnTitor (approved):
The only failure is spurious and this shouldn't affect that target (s390x) so let's just merge it, thanks!

PR #1913, opened 2020-10-06
Android: Add VMADDR_ constants.

qwandor-google:


joshtriplett:
Could you please move these up a level to `src/unix/linux_like/mod.rs`, so that they can be shared among targets? (Assuming that doing so builds on all linux_like targets.) If necessary, you could put them under `cfg_if!`.

JohnTitor:
Also, seems `VMADDR_CID_LOCAL` is newly added (https://github.com/torvalds/linux/commit/ef343b35d46667668a099655fca4a5b2e43a5dfe) so we should skip the tests for it.

qwandor-google:
There are many things which are the same between `linux_like/android` and `linux_like/linux`, as they are both actually Linux. The trouble is that `emscripten` is also under `linux_like`, and of course doesn't support all these sorts of things. If we moved `emscripten` up a level to `unix` instead then we could factor out a lot of common code from `android` and `linux`. Or alternatively, `android` could be moved under `linux`, as it is really just Linux with Bionic (rather than glibc or musl, which are both included under `linux` already).

This would be a fairly major refactor though. For now, it seems more consistent with the rest of the code to put things in both `android` and `linux` rather than having lots of `cfg_if!` in `linux_like`.

joshtriplett:
@qwandor-google I don't know if emscripten is getting much benefit from being under linux_like rather than just unix. On the one hand it might implement some things that Linux does but unix doesn't, in which case it'd be nice to avoid duplicating that. On the other hand, it does regularly seem like emscripten needs excluding like this from APIs which otherwise exist anywhere the Linux kernel does.

Separately from that, moving android under the `linux` directory (alongside `gnu`) seems like a great plan, and I'd love to see that. That seems like it'd avoid a lot of duplication, and it would also allow moving anything in linux_like that excludes emscripten down a level to `linux` to remove the corresponding `cfg_if`.

That's obviously not a prerequisite for *this* change. But if you'd be willing to work on that, I think that'd make it much easier to align APIs between the `linux` and `android` cases. @JohnTitor, what do you think?

qwandor-google:
@joshtriplett I don't know enough about emscripten to know what the right way forward is on that. I guess we should open an issue to track the refactoring in any case, and discuss it there rather than on this pull request.

The test failure above is only on s390x, and looks like a timeout unrelated to this change. I suggest re-running.

JohnTitor:
> I don't know if emscripten is getting much benefit from being under linux_like rather than just unix. On the one hand it might implement some things that Linux does but unix doesn't, in which case it'd be nice to avoid duplicating that. On the other hand, it does regularly seem like emscripten needs excluding like this from APIs which otherwise exist anywhere the Linux kernel does.

Fair enough, but we should do that carefully not to break anything.

> I don't know enough about emscripten to know what the right way forward is on that. I guess we should open an issue to track the refactoring in any case, and discuss it there rather than on this pull request.

Sounds good, opened #1915, and just added a brief explanation. I'd appreciate any comments :)

JohnTitor (approved):
> The test failure above is only on s390x, and looks like a timeout unrelated to this change. I suggest re-running.

I'm going to approve as-is since it passed at the previous testing. Thanks!

PR #1916, opened 2020-10-08
Retry N times on s390x CI to avoid timeout failure

JohnTitor:
This target often fails to test due to timeout so let's retry N times like rustup installing.


JohnTitor:
Hmm, failed in {aarch64,arm}-android targets and it's definitely unrelated, but we should fix it as it doesn't seem spurious.

JohnTitor:
Yay, it works fine. Let's see if it works as expected in some other PRs!

PR #1917, opened 2020-10-09
Fix {aarch64,arm}-android targets' CI

JohnTitor:
Let's see if it still happens.


JohnTitor:
Oh well, s390x is failing here... But the android issue is fixed, let's merge this.

PR #1919, opened 2020-10-10
Add missing NetBSD RLIMIT_ constants

rbtcollins:
I only really need RLIMIT_AS for my project, but the other two that are missing are easy to add so :shrug:


JohnTitor (approved):
The change itself looks good, could you also fix the style issue?

JohnTitor:
Thanks!

PR #1922, opened 2020-10-13
Add clock_nanosleep to freebsd and netbsd

### LINKED ISSUE #1921: Add clock_nanosleep to most unix targets
`clock_nanosleep` is defined by libc only for Linux, Fuchsia, Android and Solaris, despite being supported by most Unix OSs (not macos).

maxbla:
fixes #1921

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/freebsd12/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/freebsd12/mod.rs
@@ -221,6 +221,12 @@ extern "C" {
         msgtyp: ::c_long,
         msgflg: ::c_int,
     ) -> ::ssize_t;
+    pub fn clock_nanosleep(
+        clk_id: ::clockid_t,
+        flags: ::c_int,
+        rqtp: *const ::timespec,
+        rmtp: *mut ::timespec,
+    ) -> ::c_int;
 }
 
 cfg_if! {
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -1743,6 +1743,12 @@ safe_f! {
 extern "C" {
     pub fn ntp_adjtime(buf: *mut timex) -> ::c_int;
     pub fn ntp_gettime(buf: *mut ntptimeval) -> ::c_int;
+    pub fn clock_nanosleep(
+        clk_id: ::clockid_t,
+        flags: ::c_int,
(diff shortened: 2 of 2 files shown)


maxbla:
`clock_nanosleep` was added in freeBSD 11.1 release! time to add another cfg.

JohnTitor:
Overall it looks good to me but I'm not sure this is available on redox, is it @jackpot51?

maxbla:
Thanks for the review. After submitting this PR, I have a couple more concerns
1) Freebsd 11.0 doesn't have `clock_nanosleep` - it was added in 11.1. If we're still trying to support 11.0, I would need to remove `clock_nanosleep` from our freebsd 11 target.
2) Contributing.md makes it look like I should add `clock_nanosleep` to every supported platform's mod.rs instead of unix's mod.rs with a cfg.

regarding redox OS, I searched for `clock_nanosleep` in redox and relibc and only got results from llvm, so I think its safe to say redox doesn't implement it.

semarie:
Please note that OpenBSD doesn't support `clock_nanosleep'. For others BSD, I dunno.

JohnTitor:
So, I'm happy to extend the support for `clock_nanosleep` but we should add them to the only platforms that we're sure it's available. Given the above comments, I don't think `src/unix/mod.rs` is the best place atm?

maxbla:
Yep, totally agree. I'll force push the change I should have made.

maxbla:
Should I add clock_nanosleep for freebsd 11? clock_nanosleep appears to have been added in 11.1, so it wouldn't work for 11.0

src/unix/bsd/freebsdlike/freebsd/freebsd12/mod.rs
```diff
@@ -229,4 +235,4 @@ cfg_if! {
         mod b64;
         pub use self::b64::*;
     }
-}
+}
```
JohnTitor, line 238: nit: We usually put a line break on the last line, could you re-add a line break? Otherwise we will change it back-and-forth :)

JohnTitor:
The change itself looks good to me, r=me once the above nit is resolved.

src/unix/bsd/freebsdlike/freebsd/freebsd12/mod.rs
maxbla: whoops. not sure how that happened.

JohnTitor:
Thanks!
@bors r+

PR #1927, opened 2020-10-16
Fix aarch64-linux-android target CI

JohnTitor:


JohnTitor:
Finally, we've got AArch64-Android CI to work, with a newer NDK version. I think we could also fix i686-android, but I'd do it on another PR.
@bors r+

PR #1929, opened 2020-10-16
Testing Android CI

qwandor-google:
Trying to debug #1765 .


JohnTitor:
Note that we don't have Android builders on PRs, but on bors.

JohnTitor:
You cannot do because bors permission is restricted to the maintainers. I could delegate you but I think the ways I mentioned in https://github.com/rust-lang/libc/issues/1765#issuecomment-710249540 are more suitable.

JohnTitor:
Since the issue is fixed now, I'm going to close this. Thanks for contributing anyway!

PR #1932, opened 2020-10-17
Add some more targets for build testing

JohnTitor:
- `x86_64-unknown-illumos`: We now have the illumos module and it'd be great if we could test it actually.
- `aarch64-apple-darwin`: This target will be used more widely in the future and it should make sense to run CI for it as well.

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -154,11 +154,14 @@ x86_64-unknown-illumos \
 x86_64-unknown-linux-gnux32 \
 "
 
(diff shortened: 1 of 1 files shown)


JohnTitor:
Hm, I'm not sure about this failure. Anyway, FreeBSD 10 reached EoL a while ago so I think it's a good time to drop it here.

JohnTitor:
@bors retry
Seems spurious.

JohnTitor:
Hmm, @bors r-

PR #1934, opened 2020-10-17
Run build test for `x86_64-unknown-redox`

### LINKED ISSUE #1457: Temporarily disable redox target (tracking issue)
cc @jackpot51 I'm disabling the build tests for the redox target as part of the migration to Azure. It should be re-enabled in the redox PRs that migrate to the unix target.

JohnTitor:
Fix #1457


JohnTitor:
Okay, we've found that it seems some items in `src/unix/mod.rs` aren't defined in redox actually.
I'd downgrade CI to build testing to avoid breakage for now but we should address at some point.

pietroalbini:
@bors r-

Bors mistakenly thought this didn't fail after a synchronize.

PR #1938, opened 2020-10-18
Adding nmount call to support FreeBSD disk mounting

DarcInc:
Adding disk mounting support for FreeBSD, which prefers nmount over mount.


DarcInc:
Closing to resubmit with format fix.

PR #1939, opened 2020-10-18
Add freebsd nmount

DarcInc:
Adding FreeBSD's nmount call to mount devices on filesystems.  nmount is preferred over mount.


JohnTitor:
Note that CI for FreeBSD 10 is currently broken, it should be fixed in a few days. The change itself looks good, r=me once you squash commits into one or two (`nmount` and `fdatasync`) and CI works fine again.

DarcInc:
Rgr.  @JohnTitor Squashed and redelivered.

JohnTitor (approved):
Thanks!

PR #1943, opened 2020-10-20
Add definitions for Linux debug related MAGIC

KentaTada:
This commit adds definitions for Linux debug related MAGIC to confirm if the filesystem for debug is mounted.

Signed-off-by: Kenta Tada <Kenta.Tada@sony.com>


JohnTitor (approved):
Thanks!

JohnTitor:
Ugh, MIPS64 uses too many resources in normal testing as well. I'll disable it shortly, one sec.

PR #1947, opened 2020-10-21
Add lockf() operations to Android

fornwall:
Reference:
https://github.com/aosp-mirror/platform_bionic/blob/master/libc/include/bits/lockf.h#L39-L46


JohnTitor (approved):
Looks good, thanks!

PR #1958, opened 2020-10-25
Allow `deprecated` attribute on `f!` macros to fix dragonfly build

### LINKED ISSUE #1957: Fails to build on dragonfly target with a macro matcher failure
```
error: no rules expected the token `#`
Error:    --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.79/src/unix/bsd/freebsdlike/dragonfly/errno.rs:4:5
    |
4   |     #[deprecated(since = "0.2.77", "Use `__errno_location()` instead")]
    |     ^ no rules expected this token in macro call
    | 
   ::: /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.79/src/macros.rs:230:9
    |
230 |         macro_rules! f {
    |         -------------- when calling this macro
```

From a brief look at the code it seems like `f!` has not been written in a way that would make it expect an item with attributes, making building of libc fail when this attribute was added.

JohnTitor:
Fixes #1957
cc @nagisa I think this should resolve the build :)

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -40,12 +40,12 @@ macro_rules! cfg_if {
 
     // Internal and recursive macro to emit all the items
     //
-    // Collects all the negated cfgs in a list at the beginning and after the
+    // Collects all the negated `cfg`s in a list at the beginning and after the
     // semicolon is all the remaining items
     (@__items ($($not:meta,)*) ; ) => {};
     (@__items ($($not:meta,)*) ; ( ($($m:meta),*) ($($it:item)*) ),
      $($rest:tt)*) => {
-        // Emit all items within one block, applying an approprate #[cfg]. The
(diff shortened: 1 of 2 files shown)


nagisa:
I was going to test it but could not because my test involves `-Zbuild-std` (and I don't have the time to set up anything else); libstd depends on a specific libc crate version I wasn't able to patch out, so I'll just have to trust my gut and say it LGTM.

nagisa (approved):

nagisa:
Can I have a release with this included so that I could update libstd's libc dependency?

JohnTitor:
Sure! Opened #1959

JohnTitor:
Released 0.2.80 :)

PR #1964, opened 2020-10-30
Add more TCP options for solaris systems.

jasonbking:
This adds the `TCP_KEEPCNT` and `TCP_KEEPINTVL` tcp values to the libc crate. It also corrects the value of `TCP_KEEPIDLE` for Solaris systems. As seen in the diff, the value is different between illumos and Solaris (as the feature was added after the two code bases forked, and each ended up choosing a different value).

Adding these to libc will allow the nix crate to utilize them when used on solarish systems.


JohnTitor (approved):
Looks good!

PR #1967, opened 2020-10-31
Adding dl_iterate_phdr function to uclibc

tsidea:
Adding dl_iterate_phdr function and related structs and types to the
uclibc module as per the file include/link.h found in the uClibc-ng
repository as of version 1.0.36. This is necessary in order for rust to
work with a new target armv7-unknown-linux-uclibceabihf.

### DIFF
--- a/src/unix/uclibc/mod.rs
+++ b/src/unix/uclibc/mod.rs
@@ -21,6 +21,16 @@ pub type msglen_t = ::c_ulong;
 pub type nfds_t = ::c_ulong;
 pub type nl_item = ::c_int;
 pub type idtype_t = ::c_uint;
+pub type Elf32_Half = u16;
+pub type Elf32_Word = u32;
+pub type Elf32_Off = u32;
+pub type Elf32_Addr = u32;
+pub type Elf64_Half = u16;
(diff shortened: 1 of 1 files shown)


tsidea:
The only problem I have is that I couldn't figure out how to cross compile libc-test for the new target `armv7-unknown-linux-uclibceabihf` I have created locally. I managed however to compile std and my own project for this new target. I even compiled my project's tests for this new target, but not libc-test. Actually, I tried cross compiling libc-test for `armv7-unknown-linux-gnueabihf` as well, which is an existing target and I couldn't do that either.

JohnTitor (approved):
Looks good!

JohnTitor:
> The only problem I have is that I couldn't figure out how to cross compile libc-test for the new target armv7-unknown-linux-uclibceabihf I have created locally. I managed however to compile std and my own project for this new target. I even compiled my project's tests for this new target, but not libc-test. Actually, I tried cross compiling libc-test for armv7-unknown-linux-gnueabihf as well, which is an existing target and I couldn't do that either.

libc-test for ARM targets needs to be run on qemu via Docker IIRC. You could see some other targets using qemu and Docker.

PR #1968, opened 2020-10-31
Implement accept4 on Android as raw syscall.

de-vri-es:
This PR implements `accept4()` on Android using `syscall()` instead of the `accept4()` wrapper.

This is done because the `accept4()` wrapper wasn't added to Androids `libc` until version 5.0 (API level 21). By using `syscall` directly, `accept4` will also work on older  Android versions.

The kernel itself has supported it since Linux 2.6.28, which was shipped with stock Android since version 1.6 already: https://en.wikipedia.org/wiki/Android_version_history#Android_1.6_Donut_(API_4)

At the moment, the CI for the rust repo still uses API level 14. Although I also opened a PR to bump that too: https://github.com/rust-lang/rust/pull/78601. However, it might make sense to keep the old API level in CI if this is merged.

### DIFF
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -2349,6 +2349,20 @@ f! {
     pub fn SO_EE_OFFENDER(ee: *const ::sock_extended_err) -> *mut ::sockaddr {
         ee.offset(1) as *mut ::sockaddr
     }
+
+    // Sadly, Android before 5.0 (API level 21), the accept4 syscall is not
+    // exposed by the libc. As work-around, we implement it through `syscall`
+    // directly. This workaround can be removed if the minimum version of
+    // Android is bumped. When the workaround is removed, `accept4` can be
+    // moved back to `linux_like/mod.rs`
+    pub fn accept4(
+        fd: ::c_int,
+        addr: *mut ::sockaddr,
+        len: *mut ::socklen_t,
+        flg: ::c_int
+    ) -> ::c_int {
+        syscall(SYS_accept4, fd, addr, len, flg) as ::c_int
+    }
 }
 
 extern "C" {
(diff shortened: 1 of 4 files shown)


JohnTitor (approved):
Looks fine to me, thanks!

src/unix/linux_like/android/mod.rs
```diff
@@ -2349,6 +2349,20 @@ f! {
     pub fn SO_EE_OFFENDER(ee: *const ::sock_extended_err) -> *mut ::sockaddr {
         ee.offset(1) as *mut ::sockaddr
     }
+
+    // Sadly, Android before 5.0 (API level 21), the accept4 syscall is not
+    // exposed by the libc. As work-around, we implement it through `syscall`
+    // directly. This workaround can be removed if the minimum version of
+    // Android is bumped. When the workaround is removed, `accept4` can be
+    // moved back to `linux_like/mod.rs`
+    pub unsafe fn accept4(
```
JohnTitor, line 2358:
```suggestion
    pub fn accept4(
```

de-vri-es: Is that the right solution? All the other syscall wrappers are implicitly unsafe because they're extern "C" declarations.

I imagine it would cause warnings about unnecessary `unsafe` blocks for users of `accept4()` when compiling for Android platforms.

de-vri-es: Oh, the `f` macro already takes care of that, it seems. Applying the suggestion :)

de-vri-es:
Applied the suggestion :)

JohnTitor:
> Should I squash the commits?

Yeah, r=me once it's done.

de-vri-es:
Squashed :) 

@bors r=JohnTitor

de-vri-es:
Awh, bors didn't like that :(

JohnTitor:
Thanks!
@bors r+

PR #1971, opened 2020-11-05
Add ENOTSUP constant for uclibc

skrap:
In the vein of other linux-like c libs, just define `ENOTSUP` as equal to `EOPNOTSUPP`.


skrap:
Actually I'm not sure that this is correct.  Seems like ENOTSUP is defined in libc/src/unix/uclibc/mips/mod.rs already.  I must not understand when something belongs in an arch-specific sub-module.  

Some of this must have been generated programmatically.  If anyone knows how that was done, I'd love a heads up.

PR #1973, opened 2020-11-10
mips some definition migration

kolapapa:
```config
[target.mipsel-unknown-linux-uclibc]
ar = "mipsel-openwrt-linux-ar"
linker = "mipsel-openwrt-linux-gcc"
```

Some definition migration


kolapapa:
https://github.com/rust-lang/libc/pull/1967
It seems that some of my commits conflict with this pr

PR #1974, opened 2020-11-10
fix(personality): Fix types for personality constants

woodruffw:
First of all, thanks for these bindings! 

I noticed this while working on some higher-level (nix) bindings for `personality(2)`: these constants are passed to `personality(2)`, which takes a `c_ulong`. Thus, these should also be `c_ulong`s, rather than `c_int`.

### DIFF
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -1091,15 +1091,15 @@ pub const WCONTINUED: ::c_int = 0x00000008;
 pub const WNOWAIT: ::c_int = 0x01000000;
 
 // Options for personality(2).
-pub const ADDR_NO_RANDOMIZE: ::c_int = 0x0040000;
-pub const MMAP_PAGE_ZERO: ::c_int = 0x0100000;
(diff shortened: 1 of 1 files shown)


JohnTitor:
You're right, sorry for the trouble!
I'd follow this but it's a breaking change, let's make some period before changing actually:
1. Add a deprecation notice like `#[deprecated(note = "This type is defined as c_int by mistake. We will change it to c_ulong in the future release")]` so that the users know it.
1. After some releases, change the type.

cc @nbaksalyar as you introduced these consts.

woodruffw:
No worries! On closer inspection, it might actually be reasonable: `personality(2)` takes a `c_ulong` but *returns* a `c_int`. Moreover the return value of `personality(2)` (when it isn't an error) is specified to be the *previous* persona, meaning that all personality values need to fit into `c_int` and that it's an equally valid type choice. 

I'm interested in hearing what @nbaksalyar thinks, but I'd personally be okay with the current signatures in light of the above 🙂

woodruffw:
I'm gonna go ahead and close this, since I've convinced myself that the current types are correct.

PR #1975, opened 2020-11-12
Move uclibc under linux

skrap:
This is a first cut at moving uclibc under linux, alongside `musl` and `gnu`.

All tests pass on a vexpress a9 running in qemu.  I am working on testing the other tier 3 uclibc targets: `mips-unknown-linux-uclibc` and `mipsel-unknown-linux-uclibc`.  ~Not having access to these platforms, I am working on getting them running in qemu, but it may take a bit.~

The style check doesn't pass.  I would appreciate some direction here.  Many of these constants are defined under 2-of-3 out of musl/glibc/uclibc, so I'm not sure whether I should transform:

```rust
#[cfg(not(target_env = "uclibc"))]
pub fn foo();
```
into 2 separate declarations, one each in `musl/mod.rs` and `gnu/mod.rs` or use a `cfg_if` block for each one (which explodes 2 lines into 5).

- [x] Help me choose which fix to use for the items defined in musl and gnu but not uclibc.

It's my guess that exploding into the `cfg_if` block is better for long-term maintainability, but I'd really appreciate opinions from the maintainers.


skrap:
as of the last round of changes, the `mipsel-unknown-linux-uclibc` target works.  The code doesn't seem to distinguish between that and `mips-unknown-linux-uclibc` so I am going to declare victory over mips+uclibc and await review comments!

skrap:
@kolapapa I see you have other PRs related to this uclibc.  if you have a real mips or mipsel uclibc system, and don't mind trying to compile this branch for your use case, I would love any feedback.  I definitely don't want to break anyone's code!

kolapapa:
> @kolapapa I see you have other PRs related to uclibc. if you have a real mips or mipsel uclibc system, and don't mind trying to compile this branch for your use case, I would love any feedback. I definitely don't want to break anyone's code!

Of course, how can I cooperate with you? I have the real system of mips and the target uses `mipsel-unknown-linux-uclibc`.

skrap:
> Of course, how can I cooperate with you? I have the real system of mips and the target uses `mipsel-unknown-linux-uclibc`.

Would you mind running your code against this branch?  I'd be interested in any compilation issues, and definitely any runtime issues you see.

kolapapa:
> Would you mind running your code against this branch? I'd be interested in any compilation issues, and definitely any runtime issues you see.

Currently encountering a problem:
`master` branch
```rust
pub fn getnameinfo(
        sa: *const ::sockaddr,
        salen: ::socklen_t,
        host: *mut ::c_char,
        hostlen: ::socklen_t,
        serv: *mut ::c_char,
        sevlen: ::socklen_t,
        flags: ::c_int,
    ) -> ::c_int;
```

`feature/move-uclibc-under-linux` branch
```rust
pub fn getnameinfo(
        sa: *const ::sockaddr,
        salen: ::socklen_t,
        host: *mut ::c_char,
        hostlen: ::socklen_t,
        serv: *mut ::c_char,
        sevlen: ::socklen_t,
        flags: ::c_uint,
    ) -> ::c_int;
```
`flags` type looks wrong，i use `dns-lookup`(v1.0.5) crates.(https://github.com/keeperofdakeys/dns-lookup/blob/master/src/nameinfo.rs#L27). It is also of type int when it is defined

skrap:
Hmmm, so uclibc defines flags as a uint in <netdb.h>, like so:

```extern int getnameinfo (const struct sockaddr *__restrict __sa,
			socklen_t __salen, char *__restrict __host,
			socklen_t __hostlen, char *__restrict __serv,
			socklen_t __servlen, unsigned int __flags);
```

However this seems to make uclibc the odd one out in terms of defining this parameter, and since there are dependent crates which require the type to be signed, we can just make libc pretend that it's a signed int.  Since it's a flags field it won't really matter.

I pushed a change to this branch to address this.  Would you mind checking again?

kolapapa:
> Hmmm, so uclibc defines flags as a uint in <netdb.h>, like so:
> 
> ```
> extern int getnameinfo (const struct sockaddr *__restrict __sa,
> 			socklen_t __salen, char *__restrict __host,
> 			socklen_t __hostlen, char *__restrict __serv,
> 			socklen_t __servlen, unsigned int __flags);
> ```
> 
> However this seems to make uclibc the odd one out in terms of defining this parameter, and since there are dependent crates which require the type to be signed, we can just make libc pretend that it's a signed int. Since it's a flags field it won't really matter.
> 
> I pushed a change to this branch to address this. Would you mind checking again?


So far everything is normal.
I used these crates
```toml
clap = { version = "2.33.0", features = ["yaml"] }
tokio = { version = "0.2", features = ["full"] }
reqwest = { version = "0.10.4", features = ["json", "blocking"] }
serde = { version = "1.0", features = ["derive"] }
serde_json = "1.0"
regex = "1.3"
telnet = "0.1.4"
glob = "0.3.0"
lazy_static = "1.4.0"
log = "0.4"
pretty_env_logger = "0.4.0"
anyhow = "1.0.27"
crossbeam-queue = "0.2.1"
config = "0.10.1"
futures = "0.3.4"
trust-dns-resolver = "0.19.4"
url = "2.1.1"
dns-lookup = "1.0.5"
```

skrap:
@kolapapa Great news.  That's super helpful, thanks!

jyn514:
Since this is mostly moving code around, this should make the PR a lot easier to review: https://github.com/rust-lang/rust/issues/76268#mentoring-instructions-for-reviewers

JohnTitor:
(Note that we haven't built any consensus on moving yet.)

skrap:
> (Note that we haven't built any consensus on moving yet.)

Hello @JohnTitor!  I realize you are quite busy.  I wanted to reach out and ask if there is a way that I can help move this conversation forward.  I know it's a big change.  It's almost entirely out of the tier-1/2 space, and I will have the ability to maintain this area of the code long-term.  It would allow crates like `nix` to operate atop uclibc, which isn't really possible today.  I would also be happy to do the work to build some CI tests around this work as well.

JohnTitor:
So, generally, I'd avoid moving the code without a strong reason to keep git-history consistent. If this PR can simplify our codebase and structure, however, I'm happy to accept it. In this case, most changes look fine but I'd like to:
- Use `cfg_if!` as possible to avoid adding `#[cfg(not(target_env = "uclibc"))]` to each item.
- I'd see CI stuff first if you could set up for it, to ensure everything isn't broken.
- Squash commits into appropriate numbers (~5 or so) as the current will hurt our history.

skrap:
Thanks much for the reply!  That all makes sense, and I'll work on what you suggested.

For CI, there are no official builds of `uclibc` targets.  I think this means that CI will have to build a full rust toolchain plus spin up an image for use within qemu.  Do I have that right?  Is that amount of CI CPU acceptable?

jyn514:
@skrap instead of building a self-hosted rustc (which like you mentioned will take forever), can you cross-compile with -Z build-std? That should have reasonable compile times.

skrap:
I will try that!

skrap:
> @skrap instead of building a self-hosted rustc (which like you mentioned will take forever), can you cross-compile with -Z build-std? That should have reasonable compile times.

@jyn514 Sadly I don't think that's going to work.  There's a chicken & egg problem here:

```
$ cargo +nightly run -Z build-std --target mips-unknown-linux-uclibc
[...]
error[E0425]: cannot find function `dl_iterate_phdr` in crate `libc`
   --> /home/skrap/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/../../backtrace/src/symbolize/gimli.rs:378:23
    |
378 |                 libc::dl_iterate_phdr(Some(callback), &mut ret as *mut Vec<_> as *mut _);
    |                       ^^^^^^^^^^^^^^^ not found in `libc`
```

So, nightly rust can't `build-std` against uclibc because libc doesn't build against uclibc.  This commit would fix that, but only once nightly rust starts using some libc build which includes this change.

There's seemingly no way to get `build-std` to use a patch-like mechanism, sadly.

jyn514:
Hmm, you shouldn't need all of libstd, just core - and actually, looking at https://github.com/rust-lang/libc/blob/master/src/lib.rs#L19-L23 you might not even need core. If you remove `-Z build-std` and add `--features rustc-dep-of-std` does it work?

skrap:
Sadly no, but I suspect I'm doing it wrong:

```
skrap@pop-os:/tmp/foo$ cargo +nightly run --features rustc-dep-of-std --target mips-unknown-linux-uclibc
error: Package `foo v0.1.0 (/tmp/foo)` does not have the feature `rustc-dep-of-std`
```

jyn514:
@skrap I would rather not debug this on github, the 10 minute back and forth is pretty painful. Feel free to hop on [the discord](https://discord.gg/rust-lang) in #contribute or [Zulip](https://rust-lang.zulipchat.com/) and I'm happy to help out.

skrap:
So it seems that since rustup does not currently provide nightlies for any of the uclibc targets, I don't see any way to get this into CI without building a full rust in CI.  `-Z build-std` doesn't help, as it does not allow patching of the deps tree, and ends up trying to build libc 0.2.79, which doesn't compile for any uclibc target.

I'm out of answers for how to proceed on CI, so I will work on the other tasks.  Once this change lands in nightly, and libc can once again build against uclibc, I will take up the CI issue again.

(Unless someone has ideas for a different workaround.)

JohnTitor:
So, is it difficult to fix a libc build (and bump its version on rust-lang/rust) first then move items? If it's some simple work, I think it's one of the good ways. If not, then I'm fine to review/merge this one.

skrap:
Ok, after some consulting with @jyn514 and others I was informed that I *could* build libc via build-std!  So I'll work on using that to have CI at least build libc.  I think it's probably worth it to have the tests run in CI as well but that will have to come later, once libc-test can build against nightly rust.

skrap:
Todo still:

- [ ]  Use cfg_if! as possible to avoid adding #[cfg(not(target_env = "uclibc"))] to each item.
- [ ] Separate commit for CI first, via `cargo +nightly build --target mipsel-unknown-linux-uclibc -Z build-std=core --no-default-features` or similar
- [ ] Squash commits into appropriate numbers (~5 or so)

skrap:
```src/unix/linux_like/linux/mod.rs:507 - multiple s! macros in one module```

This style failure is tricky: I need to conditionally define things in a `s!` block, and that will necessarily cause a style failure.  I could create a whole new module for just those two functions, but that seems likely obfuscate the code a bit.

Other than that, the style failures are all outside of the changes made here.

**Comment by skrap** [comment]
@rustbot modify labels: +S-waiting-on-review -S-waiting-on-author

@JohnTitor I have completed the requested changes to this PR.

As for the request to add CI for a uclibc: I am working on this but it doesn't appear to be simple.  I do not have a public source for a prebuilt uclibc toolchain at the moment.  I'm looking, but without that I think CI will be out of scope.

**Comment by skrap** [comment]
Fixed merge conflicts.  FWIW the constants added by these conflicts were already present in the linux_like subtree; it's a good indication of the benefits of moving uclibc under there.

File: src/unix/mod.rs [hunk 1]
```diff
@@ -1315,32 +1345,40 @@ extern "C" {
...
     pub fn localtime(time_p: *const time_t) -> *mut tm;
     #[cfg_attr(target_os = "netbsd", link_name = "__difftime50")]
-    #[cfg_attr(target_env = "musl", allow(deprecated))] // FIXME: for `time_t`
+    #[cfg_attr(target_env = "musl", allow(deprecated))]
+    // FIXME: for `time_t`
     pub fn difftime(time1: time_t, time0: time_t) -> ::c_double;
     #[cfg_attr(target_os = "netbsd", link_name = "__timegm50")]
-    #[cfg_attr(target_env = "musl", allow(deprecated))] // FIXME: for `time_t`
+    #[cfg_attr(target_env = "musl", allow(deprecated))]
+    // FIXME: for `time_t`
```
JohnTitor, line 1381: I don't think these formatting changes are necessary?

JohnTitor (commented):
It still fails on the style check. Also, could you add uclibc targets to semver.sh so that we could run semver check for it (or running it locally is also fine)? Once they're addressed, I'd like to kick off a try run.

src/unix/mod.rs
skrap: I agree and I don't know where they came from.  Maybe rustfmt.  I'll investigate.

skrap: I think it was rustfmt. I manually removed the unrelated formatting changes from my patches and resumbitted.

skrap:
> It still fails on the style check. Also, could you add uclibc targets to semver.sh so that we could run semver check for it (or running it locally is also fine)? Once they're addressed, I'd like to kick off a try run.

All set with the above requests.  I also applied a fix to the libc-test breakage from #2019.  Let me know how your run goes!

JohnTitor:
Failed on `wasm32-unknown-emscripten`:
```
error[E0425]: cannot find value `LC_CTYPE` in this scope
   --> src/unix/linux_like/mod.rs:561:41
    |
561 | pub const LC_CTYPE_MASK: ::c_int = 1 << LC_CTYPE;
    |                                         ^^^^^^^^ not found in this scope

error[E0425]: cannot find value `LC_NUMERIC` in this scope
   --> src/unix/linux_like/mod.rs:562:43
    |
562 | pub const LC_NUMERIC_MASK: ::c_int = 1 << LC_NUMERIC;
    |                                           ^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `LC_TIME` in this scope
   --> src/unix/linux_like/mod.rs:563:40
    |
563 | pub const LC_TIME_MASK: ::c_int = 1 << LC_TIME;
    |                                        ^^^^^^^ not found in this scope

error[E0425]: cannot find value `LC_COLLATE` in this scope
   --> src/unix/linux_like/mod.rs:564:43
    |
564 | pub const LC_COLLATE_MASK: ::c_int = 1 << LC_COLLATE;
    |                                           ^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `LC_MONETARY` in this scope
   --> src/unix/linux_like/mod.rs:565:44
    |
565 | pub const LC_MONETARY_MASK: ::c_int = 1 << LC_MONETARY;
    |                                            ^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `LC_MESSAGES` in this scope
   --> src/unix/linux_like/mod.rs:566:44
    |
566 | pub const LC_MESSAGES_MASK: ::c_int = 1 << LC_MESSAGES;
    |                                            ^^^^^^^^^^^ not found in this scope

error: aborting due to 6 previous errors
```

You're using `#[cfg(any(target_env = "gnu", target_env = "musl"))]` for their declaration but it should be `not(uclibc)`, I suppose?

skrap:
I know nearly nothing about wasm but at least from a high level structural point of view I think your suggestion is correct.  I don't have the toolchain to fully test emscripten but I have made the suggested change and verified that emscripten compiles, and x86-64 linux and ucilbc on armv7 both pass the full test suite.

JohnTitor:
Awesome, let's try again.
@bors try

skrap:
Doh!  Seems like debugging will be required.  I will work on getting an appropriate target set up in qemu.

skrap:
I cannot reproduce the crash - either by compiling on my own or by running the CI script for full containerization.  Here's the tail of the output (lots of docker setup omitted, but the test results shown):

```+ cargo test --no-default-features --manifest-path libc-test/Cargo.toml --target mips-unknown-linux-musl
   Compiling proc-macro2 v1.0.24
   Compiling unicode-xid v0.2.1
   Compiling syn v1.0.48
   Compiling serde_derive v1.0.117
   Compiling libc v0.2.80
   Compiling serde v1.0.117
   Compiling cfg-if v0.1.10
   Compiling ryu v1.0.5
   Compiling log v0.4.11
   Compiling serde_json v1.0.59
   Compiling bitflags v1.2.1
   Compiling itoa v0.4.6
   Compiling semver-parser v0.7.0
   Compiling cc v1.0.61
   Compiling libc v0.2.82 (/checkout)
   Compiling semver v0.9.0
   Compiling rustc_version v0.2.3
   Compiling quote v1.0.7
   Compiling dirs-sys v0.3.5
   Compiling dirs v2.0.2
   Compiling term v0.6.1
   Compiling garando_pos v0.1.0
   Compiling garando_errors v0.1.0
   Compiling garando_syntax v0.1.0
   Compiling ctest2 v0.3.0
   Compiling libc-test v0.1.0 (/checkout/libc-test)
    Finished test [unoptimized + debuginfo] target(s) in 1m 33s
     Running target/mips-unknown-linux-musl/debug/deps/cmsg-f4cc850e8484f5a8

running 5 tests
test t::test_cmsg_data ... ok
test t::test_cmsg_len ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 6.01s

     Running target/mips-unknown-linux-musl/debug/deps/errqueue-c756e023f297c431

running 1 test
test t::test_cmsg_data ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s

     Running target/mips-unknown-linux-musl/debug/deps/linux_elf-1bf086251f3b19e5
RUNNING ALL TESTS
PASSED 52 tests
     Running target/mips-unknown-linux-musl/debug/deps/linux_fcntl-b3538a28aeb25c5d
RUNNING ALL TESTS
PASSED 28 tests
     Running target/mips-unknown-linux-musl/debug/deps/linux_ipv6-66c2d571c2979be6
RUNNING ALL TESTS
PASSED 20 tests
     Running target/mips-unknown-linux-musl/debug/deps/linux_strerror_r-d0554e1c62e2c94a
RUNNING ALL TESTS
PASSED 1 tests
     Running target/mips-unknown-linux-musl/debug/deps/linux_termios-0cff8d72a084b605
RUNNING ALL TESTS
PASSED 0 tests
     Running target/mips-unknown-linux-musl/debug/deps/main-dd39c4f6b607422b
RUNNING ALL TESTS
PASSED 14609 tests
+ cargo test --manifest-path libc-test/Cargo.toml --target mips-unknown-linux-musl
   Compiling libc v0.2.82 (/checkout)
   Compiling libc-test v0.1.0 (/checkout/libc-test)
    Finished test [unoptimized + debuginfo] target(s) in 1m 04s
     Running target/mips-unknown-linux-musl/debug/deps/cmsg-52f54c2e7c50c929

running 5 tests
test t::test_cmsg_data ... ok
test t::test_cmsg_len ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 6.01s

     Running target/mips-unknown-linux-musl/debug/deps/errqueue-1da432e41cfc77d4

running 1 test
test t::test_cmsg_data ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s

     Running target/mips-unknown-linux-musl/debug/deps/linux_elf-1b37094435405878
RUNNING ALL TESTS
PASSED 52 tests
     Running target/mips-unknown-linux-musl/debug/deps/linux_fcntl-e6a5b926d259f1cb
RUNNING ALL TESTS
PASSED 28 tests
     Running target/mips-unknown-linux-musl/debug/deps/linux_ipv6-49f2e4ea8bbba686
RUNNING ALL TESTS
PASSED 20 tests
     Running target/mips-unknown-linux-musl/debug/deps/linux_strerror_r-dcd01c8e988978c8
RUNNING ALL TESTS
PASSED 1 tests
     Running target/mips-unknown-linux-musl/debug/deps/linux_termios-c5cdaccf4ce84023
RUNNING ALL TESTS
PASSED 0 tests
     Running target/mips-unknown-linux-musl/debug/deps/main-56f7f0a86ea7d56a
RUNNING ALL TESTS
PASSED 14609 tests
+ cargo test --features extra_traits --manifest-path libc-test/Cargo.toml --target mips-unknown-linux-musl
   Compiling libc-test v0.1.0 (/checkout/libc-test)
    Finished test [unoptimized + debuginfo] target(s) in 58.20s
     Running target/mips-unknown-linux-musl/debug/deps/cmsg-4e802d0d57f7dbb4

running 5 tests
test t::test_cmsg_data ... ok
test t::test_cmsg_len ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 6.80s

     Running target/mips-unknown-linux-musl/debug/deps/errqueue-05ff178d42d16565

running 1 test
test t::test_cmsg_data ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s

     Running target/mips-unknown-linux-musl/debug/deps/linux_elf-fa47e5e8a2ad8fa4
RUNNING ALL TESTS
PASSED 52 tests
     Running target/mips-unknown-linux-musl/debug/deps/linux_fcntl-ccece0b6111e3f06
RUNNING ALL TESTS
PASSED 28 tests
     Running target/mips-unknown-linux-musl/debug/deps/linux_ipv6-80efedb14eb22ea9
RUNNING ALL TESTS
PASSED 20 tests
     Running target/mips-unknown-linux-musl/debug/deps/linux_strerror_r-4414d41747f6ac1b
RUNNING ALL TESTS
PASSED 1 tests
     Running target/mips-unknown-linux-musl/debug/deps/linux_termios-162953cff325e036
RUNNING ALL TESTS
PASSED 0 tests
     Running target/mips-unknown-linux-musl/debug/deps/main-585b2d3a4bf7beff
RUNNING ALL TESTS
PASSED 14609 tests
```

skrap:
I wonder if bors will let me issue a try:
@bors try

JohnTitor:
> I cannot reproduce the crash - either by compiling on my own or by running the CI script for full containerization. Here's the tail of the output (lots of docker setup omitted, but the test results shown):

Are you using the newest nightly?

@bors delegate+ for try run.

skrap:
Hmm, I'm using
```
$ rustc --version
rustc 1.51.0-nightly (44e3daf5e 2020-12-31)
```

That seems to be what I'm given when I `rustup update nightly`.  Should I do something differently?

skrap:
Ok, I force-updated nightly to latest and it still passes:
```
$ rustc --version
rustc 1.51.0-nightly (7a9b552cb 2021-01-12)
```

I ran both directly on the host and in the docker container via the CI script.  I'm gonna have bors give it another shot just to see...

@bors try

skrap:
A very different failure, and mips-unknown-linux-musl passed this time.  I'd bet it's a bug introduced in rustc or lower.

JohnTitor:
Yeah, I think it was a spurious failure.

skrap:
OK there's a new nightly, so let's try again and see what happens:
@bors try

skrap:
I think those are spurious errors as well.  What's the next steps here?  I'm comfortable with the change given that the code passes all tests on stable rust, but if you'd prefer to wait until we can see it pass on nightly as well, I'm happy to just kick bors for each new nightly until we get a pass.  

Thanks again for all your careful attention here!

jyn514:
```
error[E0412]: cannot find type `c_char` in the crate root
  --> src/unix/mod.rs:45:29
   |
6  | pub type c_schar = i8;
   | ---------------------- similarly named type alias `c_schar` defined here
...
45 |         pub gr_name: *mut ::c_char,
   |                             ^^^^^^ help: a type alias with a similar name exists: `c_schar`
```
@skrap why do you think these errors are spurious?

skrap:
@jyn514  Only that it passes on stable and beta toolchains.  But I do see now that these failures are for the `x86_64-unknown-l4re-uclibc` target, which makes me suspicious that it could have something to do with this pull request.  I will read about l4re and see how I could try to reproduce this failure.

skrap:
@TobiasSchaffner @humenda After reading a bit about the L4RE environment, I wanted to reach out and see if you had opinions about this PR and perhaps some moments to help me make sure it works with the L4RE uclibc.

skrap:
I have made a commit to fix the l4re build, but I am not able to test it.  It's correct to the best of my knowledge, but hopefully someone from L4Re will speak up.  The L4Re target is tier 2, and unfortunately there's no CI set up for running the libc-test tests.

@bors try

humenda:
Thanks for reaching out. I will have a llok at the weekend, but unfortunately
I won't make it earlier.

humenda:
I took some time and read the changes. Unfortunately, I am not in a position
to test the changes, ATM. L4Re uses a fork of Uclibc, tha has, to my
knowledge, not been updated for a while. There's a high chance that
assumptions on your uclibc version do not hold. I only have an old snapshot
around (2 years old) and lack the time to check out what L4Re is doing these
days. I would suggest that you just continue and we hand in a PR in case
something breaks in a newer version of L4Re.

skrap:
Thanks @humenda.  @JohnTitor sounds like the work here is complete!

JohnTitor:
Oh sorry, missed the ping. Seems uclibc target doesn't have std component and we cannot run semver check for it (https://github.com/rust-lang/libc/runs/1697567233). So 9aad8ac is now meaningless, r=me once it's dropped.

JohnTitor:
Thanks!
@bors r+

skrap:
Thank you!  All of your time reviewing and shepherding this change was very much appreciated, and I am looking forward to seeing this change in production rust!  :partying_face:

PR #1976, opened 2020-11-15
Fix bug in `CMSG_NXTHDR` for android.

LinkTed:
`CMSG_NXTHDR` should return NULL if it is the last control message in the chain.


JohnTitor (commented):
Looks good, thanks!

LinkTed:
It seems that the android [implementation](https://android.googlesource.com/platform/bionic/+/master/libc/bionic/__cmsg_nxthdr.cpp#31) (bionic) is wrong. Should I close the PR?

JohnTitor:
Hm, if it's wrong and/or we don't need a fix for rust-lang/rust#69864, then yes.

PR #1979, opened 2020-11-23
add `getnameinfo` && `EAI_NODATA` in uclibc

kolapapa:
when compiling `dns-lookup v1.0.5` in target `mipsel-unknown-linux-uclibc`
```shell
error[E0432]: unresolved import `libc::getnameinfo`
error[E0531]: cannot find unit struct, unit variant or constant `EAI_NODATA` in crate `c`
```

### DIFF
--- a/src/unix/uclibc/mod.rs
+++ b/src/unix/uclibc/mod.rs
@@ -2326,7 +2326,7 @@ extern "C" {
         hostlen: ::socklen_t,
         serv: *mut ::c_char,
         sevlen: ::socklen_t,
-        flags: ::c_int,
+        flags: ::c_uint,
     ) -> ::c_int;
 }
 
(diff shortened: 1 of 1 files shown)


src/unix/uclibc/mod.rs
```diff
@@ -2318,6 +2319,15 @@ extern "C" {
...
     pub fn popen(command: *const c_char, mode: *const c_char) -> *mut ::FILE;
     pub fn uname(buf: *mut ::utsname) -> ::c_int;
+    pub fn getnameinfo(
+        sa: *const ::sockaddr,
+        salen: ::socklen_t,
+        host: *mut ::c_char,
+        hostlen: ::socklen_t,
+        serv: *mut ::c_char,
+        sevlen: ::socklen_t,
+        flags: ::c_int,
```
JohnTitor, line 2329:
```suggestion
        flags: ::c_uint,
```

Should be?

kolapapa: emmm..It should be `uint` on uclibc.
https://docs.rs/dns-lookup/1.0.5/dns_lookup/fn.getnameinfo.html
It seems that conditional compilation is needed here.

JohnTitor:
Thanks!
@bors r+

PR #1981, opened 2020-11-24
Use the newer nightly to fix the build

JohnTitor:


JohnTitor:
Let's try again! @bors r+

PR #1982, opened 2020-11-24
Add getrandom to FreeBSD

Thomasdezeeuw:
Introduced in FreeBSD 12.0.

Manual page: https://www.freebsd.org/cgi/man.cgi?query=getrandom.

Not sure if the constants should be `c_int`, just matching the `flags` argument in the function, in c they're macros.


cemeyer (approved):
LGTM, thanks.

JohnTitor (approved):
Looks good, thanks!

Issue #423, opened 2016-10-09
Generating libc from C headers


Zoxc:
It would be nice if there was a way to generate `libc` using just a standard C library. That would make porting to less common targets way easier and also ensures that the headers are correct.

alexcrichton:
I don't think there's necessarily a large amount of motivation for this in the sense that this entire library is already verified as correct and header -> Rust generators I've seen tend to have lots of things that need to be fixed by hand. It's certainly plausible to generate headers and then commit them, but I don't think there's any particular reason to rewrite the crate to use generation now.

Zoxc:
It doesn't support newlib currently (or more obscure C libraries), which makes porting `libstd` more difficult.

kristate:
Trying to clean-up issues: is this still an issue @Zoxc ?

Zoxc:
@kristate It is.

gnzlbg:
Platforms break ABI compatibility all the time due to changes in the values of constants, enums, sizes and alignment of structs, even signature of functions. 

As long as libc does not use a binding generation, every single libc version targets a specific version of an API, and has undefined behavior on all others.

gnzlbg:
@Zoxc This is not a bad idea. The main problems are that generating libc from headers requires:

* re-compiling standard, so this is only (at least for now) relevant for those using xargo
* using rust-bindgen requires a particular clang toolchain (which, at least for now, rustup is not bundling)

If one needs to re-compile libstd anyways, one can `[patch]`/`[source]` it to use a different "libc", e.g., "libc-bindgen", which uses rust-bindgen to generate Rust FFI from headers. 

I think it would be better to create such a "libc-bindgen" first, add minimal support to `xargo` to patch libstd (and all other crates) to use it in place of libc, and then explore solutions for the problems that will arise (like improving bindgen), e.g., due to differences in API between libc and the result of `bindgen`. 

Once we are there, we could start considering whether it is worth integrating such a thing with libc somehow. But I think that right now it would be too soon to discuss such a thing because we lack information about the problems / issues that will need to be solved for that to happen.

Another approach that such a different `libc-bindgen` library could support is using that to generate Rust headers for an environment like `newlib`, and then statically import those into the `libc` repo, as a way of quickly gaining comprehensive support for new platforms. 

[...]

PR #1985, opened 2020-12-03
Followup to #1649.

dvc94ch:
When adding `nlmsghdr` for `musl` targets the `aarch64` arch was singled out, not getting the definition. This remedies the problem so that `if-watch` compiles on `aarch64-unknown-linux-musl`.

- [0] https://github.com/dvc94ch/if-watch/issues/3

### DIFF
--- a/src/unix/linux_like/linux/musl/b32/arm/mod.rs
+++ b/src/unix/linux_like/linux/musl/b32/arm/mod.rs
@@ -150,24 +150,6 @@ s! {
         pub f_namemax: ::c_ulong,
         __f_spare: [::c_int; 6],
     }
-
-    pub struct nlmsghdr {
-        pub nlmsg_len: u32,
-        pub nlmsg_type: u16,
-        pub nlmsg_flags: u16,
-        pub nlmsg_seq: u32,
-        pub nlmsg_pid: u32,
-    }
-
-    pub struct nlmsgerr {
-        pub error: ::c_int,
-        pub msg: nlmsghdr,
-    }
-
-    pub struct nlattr {
-        pub nla_len: u16,
-        pub nla_type: u16,
-    }
 }
 
 pub const SIGSTKSZ: ::size_t = 8192;
--- a/src/unix/linux_like/linux/musl/b32/mips/mod.rs
+++ b/src/unix/linux_like/linux/musl/b32/mips/mod.rs
@@ -161,24 +161,6 @@ s! {
         pub f_namemax: ::c_ulong,
         __f_spare: [::c_int; 6],
     }
-
-    pub struct nlmsghdr {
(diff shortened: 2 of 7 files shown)


JohnTitor:
IIUC these items are available across all the musl targets? If so, I now wonder we could place them in `unix/linux_like/linux/musl/mod.rs`.

dvc94ch:
I guess. But I don't even understand why `libc` doesn't use something like bindgen instead of manually curating structs. It seems like it would be more reliable and complete?

JohnTitor:
> 
> 
> I guess. But I don't even understand why `libc` doesn't use something like bindgen instead of manually curating structs. It seems like it would be more reliable and complete?

See https://github.com/rust-lang/libc/issues/423. Some are changed, e.g., cargo now has the `build-std` feature (unstable) but most are still valid and should be considered.

dvc94ch:
Seems to be that `glibc` reexports `nlmsghdr` from `linux/netlink` while `musl` does not. And in `include/uapi/linux/netlink.h` `nlmsghdr` is defined as:
```c
struct nlmsghdr {
	__u32		nlmsg_len;	/* Length of message including header */
	__u16		nlmsg_type;	/* Message content */
	__u16		nlmsg_flags;	/* Additional flags */
	__u32		nlmsg_seq;	/* Sequence number */
	__u32		nlmsg_pid;	/* Sending process port ID */
};
```
meaning it should be identical on all archs.

I moved it to `musl/mod.rs`.

JohnTitor (approved):
Thanks!

JohnTitor:
```
qemu-system-sparc64: terminating on signal 15 from pid 2145 (timeout)
OpenBIOS for Sparc64
Configuration device id QEMU version 1 machine id 0
kernel addr 404000 size b267e0
kernel cmdline init=/run_prog.sh
CPUs: 1 x SUNW,UltraSPARC-IIi
UUID: 00000000-0000-0000-0000-000000000000
Welcome to OpenBIOS v1.1 built on Mar 12 2020 14:03
  Type 'help' for detailed information

RUNNING ALL TESTS[   12.031764] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000000
error: test failed, to rerun pass '--test linux-strerror_r'
```

Spurious, @bors retry

dvc94ch:
Looks like there aren't any open PR's and that there have been quite a few commits since the last release. Is it maybe possible to make a new release?

JohnTitor:
As mentioned in the contributing guide (https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md#releasing-your-change-to-cratesio), you can send a PR to make a release anytime! This time I opened #1988.

dvc94ch:
Thanks, hadn't seen that

PR #1990, opened 2020-12-07
Fix `thread_state64` on `aarch64-apple-darwin`

### LINKED ISSUE #1977: Apple/Darwin layout for thread state doesn't match in Apple Silicon (ARM)
When working on making [Wasmer and work in the Apple Silicon](https://github.com/wasmerio/wasmer/pull/1831), I realized there were some issues on the layout that rust libc thinks the thread state has, which are different in `x86_64` and `arm64`.

Currently rust libc uses `__darwin_x86_thread_state64` when accessing to `ucontext_t.uc_mcontext.__ss`, which has an invalid layout in the Apple ARM M1 context.

https://github.com/rust-lang/libc/blob/d2963c22c609eb427d6fd3ea13307276117d1fa8/src/unix/bsd/apple/b64/mod.rs#L75-L97

The structure in macOS Big Sur is defined as follows (`mach/arm/_structs.h`):

```c
_STRUCT_ARM_THREAD_STATE64
{
	__uint64_t x[29];       /* General purpose registers x0-x28 */
	void*      __opaque_fp; /* Frame pointer x29 */
	void*      __opaque_lr; /* Link register x30 */
	void*      __opaque_sp; /* Stack pointer x31 */
	void*      __opaque_pc; /* Program counter */
	__uint32_t cpsr;        /* Current program status register */
	__uint32_t __opaque_flags; /* Flags describing structure format */
};
```

Which translates roughly to:
```rust
                    pub struct __darwin_arm_thread_state64 {
                        pub __x: [u64; 29], /* General purpose registers x0-x28 */
                        pub __fp: u64,    /* Frame pointer x29 */
                        pub __lr: u64,    /* Link register x30 */
                        pub __sp: u64,    /* Stack pointer x31 */
                        pub __pc: u64,   /* Program counter */
                        pub __cpsr: u32,  /* Current program status register */
                        pub __pad: u32,   /* Same size for 32-bit or 64-bit clients */
                    };
```

The codebase should be updated to use the new layout for the Apple Silicon

JohnTitor:
Fixes #1977

### DIFF
--- a/.github/workflows/bors.yml
+++ b/.github/workflows/bors.yml
@@ -226,12 +226,10 @@ jobs:
     name: Semver Linux
     needs: build_channels_linux
     runs-on: ubuntu-20.04
+    continue-on-error: true
     strategy:
       fail-fast: true
     steps:
-      - uses: rust-lang/simpleinfra/github-actions/cancel-outdated-builds@master
-        with:
-          github_token: "${{ secrets.GITHUB_TOKEN }}"
       - uses: actions/checkout@v2
       - name: Setup Rust toolchain
         # Should update the semverver revision in semver.sh if we touch nightly ver.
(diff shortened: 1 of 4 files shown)


JohnTitor:
@syrusakbary Could you confirm if this patch works on your end? I cannot do by myself as not having Apple Silicon Mac.

And cc @shepmaster could you also review this as you're familiar with this target?

JohnTitor:
@bors r+
Let's give it a try.

PR #1991, opened 2020-12-07
Add getauxval for 64-bit Android

### LINKED ISSUE #1987: Add getauxval for Android
`getauxval` is useful for detecting CPU specific features such as availability of the ARMv8 AES instructions.

Currently in this crate, `getauxval` is available on Linux but not on Android. This function was added to Android in API Level 18 (https://developer.android.com/ndk/guides/cpu-features). So it might be a problem for someone building to an older API level. However on `aarch64` and `x86_64` the earliest API Level is 21 and thus for these architectures `getauxval` is always available.

Happy to contribute a patch, just wanted to check if enabling `getauxval` only for `aarch64-linux-android` and `x86_64-linux-android` was ok or if you had suggestions on how to handle the possibility of it not being included when built for an old Android.

jack-signal:
Fixes #1987 

This adds `getauxval` declaration on all 64-bit Android targets. This function was not declared in Bionic until API level 18. 64-bit support was added to Android in 5.0 (API level 21) thus all 64-bit targets (including AFAICT MIPS64 though I am not able to confirm this) should have `getauxval` declared.

The `AT_` constants should be the same on all platforms.


JohnTitor (approved):
Thanks!

JohnTitor:
```
Caused by:
  process didn't exit successfully: `qemu-mips -L /toolchain/staging_dir/toolchain-mips_24kc_gcc-7.4.0_musl /checkout/target/mips-unknown-linux-musl/debug/deps/cmsg-b71a9b7d7603f07d` (signal: 11, SIGSEGV: invalid memory reference)
```

Spurious? @bors retry

JohnTitor:
Ah, you should import header files here:
https://github.com/rust-lang/libc/blob/70962e3943d22d94141954db18eb7295b5edec5d/libc-test/build.rs#L1473

jack-signal:
Should be ready for a retry

JohnTitor:
Let's try with a new bors feature.

@bors squash r+

PR #1992, opened 2020-12-07
Android: make statfs64.f_fsid public

victorhsieh:
This has been the case since the initial commit
a36da11fb9cfcafcee6cb263bf1dc2c84371730a. But there is no reason to hide
this struct member specifically.

### DIFF
--- a/src/unix/linux_like/android/b32/mod.rs
+++ b/src/unix/linux_like/android/b32/mod.rs
@@ -74,7 +74,7 @@ s! {
         pub f_bavail: u64,
         pub f_files: u64,
         pub f_ffree: u64,
-        f_fsid: [u32; 2],
+        pub f_fsid: ::__fsid_t,
         pub f_namelen: u32,
         pub f_frsize: u32,
         pub f_flags: u32,
--- a/src/unix/linux_like/android/b64/mod.rs
+++ b/src/unix/linux_like/android/b64/mod.rs
@@ -84,7 +84,7 @@ s! {
         pub f_bavail: u64,
         pub f_files: u64,
         pub f_ffree: u64,
-        f_fsid: [u32; 2],
+        pub f_fsid: ::__fsid_t,
         pub f_namelen: u64,
         pub f_frsize: u64,
         pub f_flags: u64,


JohnTitor:
> But there is no reason to hide
this struct member specifically.

Hmm, I don't think so. The `f_fsid` has some security concerns according to the man page, and `fsid_t` also hides their value. Therefore I don't think it makes sense to expose it at all.

victorhsieh:
libc is not the right place to deal with this "security concern".  It'll limit what programs can do.  This specific syscall wrapper should be a transparent interface to fstatfs(2), like what's provided in other programming languages, other platforms (note this one is only not public for Android in this crate), or even other syscall struct (i.e. statvfs64 in the same file).

My understanding to this "security concern" is really about confidentiality. Using statfs/fstatfs against files on a remote filesystem would expose the server's backing filesystem types.

JohnTitor:
> note this one is only not public for Android in this crate

This isn't right, other `statfs64` exposes `f_fsid` but `fsid_t` doesn't actually.

victorhsieh:
Please take a look at the latest patch.  I fixed the type as well to be consistent with Android's libc.

JohnTitor:
This will be a breaking change, though.

Before considering accepting this, I'd also like to hear the initial thoughts from @alexcrichton as I'm not sure if my understanding is correct.

alexcrichton:
Ah unfortunately I don't remember the initial context around this, but I would agree with the point of view that `libc` isn't positioned to nudge towards best practices, but rather it's intended to give a faithful and accurate view of  the underlying system. In that senses I would agree that this shouldn't be hidden for confidentiality reasons, but I'm not sure if that was the only reason it was hidden.

victorhsieh:
> This will be a breaking change, though.

Probably not, since this field wasn't public before?

> ... I'm not sure if that was the only reason it was hidden.

If I had to make a guess, maybe it was trying to be cautious about endian?  I saw some condition for mips in the original change.

But anyway, Android kernel nowadays is pretty much just Linux, so it's nice to keep the structs consistent.  In fact, I don't see any good reason to maintain duplicated definitions, but de-duplication would be a bigger change and is out of the current scope :)

JohnTitor (approved):
> Probably not, since this field wasn't public before?

Ah my bad, you're right.

So, we haven't seen any issues here and I think it's reasonable to de-duplicate items.
We can revert it anytime if we find an issue.

victorhsieh:
Thanks!

PR #1996, opened 2020-12-12
Declare `seekdir` and `telldir` for Android

whitequark:
These declarations are the same as those for other platforms.

They are available [since API level 23](https://github.com/aosp-mirror/platform_bionic/blob/29cff99e08746f8060f32eba03df3b95acf1163d/libc/include/dirent.h#L145-L163).


JohnTitor (approved):
Looks good!

PR #1998, opened 2020-12-17
Make si_status method more compatible

dylni:
Now that [this PR](https://github.com/rust-lang/libc/pull/1858/files) has been merged, `siginfo_t::si_status` is a method for some targets and a field on others. This PR adds the method on targets that already expose the field, which lets the method be used without target detection.

I think this change is backward compatible, but it would be better if someone else could confirm that. The struct would now have a method and a field with the same name on some targets. If that's incompatible, another method could be added that accesses the field or calls the method depending on the target, but this is cleaner.

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -690,6 +690,10 @@ impl siginfo_t {
     pub unsafe fn si_uid(&self) -> ::uid_t {
         self.si_uid
     }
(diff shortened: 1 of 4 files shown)


JohnTitor (approved):

JohnTitor:
Thanks!

@bors r+

dylni:
Thank you too!

PR #2000, opened 2020-12-18
Populate mcontext_t on aarch64-linux-musl

lu-zero:
It is used by wasmtime.

Should address https://github.com/bytecodealliance/wasmtime/issues/2133

### DIFF
--- a/src/unix/linux_like/linux/musl/b64/aarch64/align.rs
+++ b/src/unix/linux_like/linux/musl/b64/aarch64/align.rs
@@ -6,7 +6,7 @@ s_no_extra_traits! {
     }
 }
 
-s!{
+s! {
     pub struct ucontext_t {
         pub uc_flags: ::c_ulong,
         pub uc_link: *mut ucontext_t,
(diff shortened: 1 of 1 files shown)


src/unix/linux_like/linux/musl/b64/aarch64/align.rs
```diff
@@ -17,9 +17,11 @@ s!{
...
-        // What we want here is a single [u64; 36 + 512], but splitting things
-        // up allows Debug to be auto-derived.
-        __regs1: [[u64; 18]; 2], // 36
-        __regs2: [[u64; 32]; 16], // 512
+        pub fault_address: ::c_ulong,
+        pub regs: [::c_ulong; 31],
+        pub sp: ::c_ulong,
+        pub pc: ::c_ulong,
+        pub pstate: ::c_ulong,
+        pub __reserved: [u128; 256],
```
JohnTitor, line 25: - We cannot use `u128` here to support old Rust version
- Should be nested to auto-derive
- Should not be `pub` as `__reserved`

lu-zero: You are right, updated :)

JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

JohnTitor:
Seems musl-libc.org is down, I'll retry once it gets back.

JohnTitor:
```
PING musl-libc.org (45.63.0.111) 56(84) bytes of data.
64 bytes from libc.org (45.63.0.111): icmp_seq=1 ttl=52 time=195 ms
64 bytes from libc.org (45.63.0.111): icmp_seq=2 ttl=51 time=195 ms
```

@bors retry

PR #2004, opened 2020-12-22
Define CLD_ constants for more targets

dylni:
This is an extension of #1860 to support more targets. The values for these constants should be correct, but I'm not familiar with most of the targets, so they should be verified.

These files should define the constants too, but the targets are closed source, so I don't know their values:
- [src/unix/solarish/mod.rs](https://github.com/rust-lang/libc/blob/master/src/unix/solarish/mod.rs)
- [src/vxworks/mod.rs](https://github.com/rust-lang/libc/blob/master/src/vxworks/mod.rs)


JohnTitor (approved):
Looks good!

dylni:
Thanks!

PR #2007, opened 2020-12-23
Fix symbols names on macOS aarch64

bsteinb:
libSystem on macOS Big Sur on aarch64 exports these symbols without the `$suffix`es. See also: https://github.com/GuillaumeGomez/sysinfo/issues/372


JohnTitor (approved):
Looks good, thanks!

PR #2008, opened 2020-12-23
Provide complete definition of __darwin_mcontext64 on aarch64

bsteinb:
While working on #2007 I tried to run `cargo test` in `libc-test`, which failed, because the definition of `__darwin_mcontext64` was incomplete (see #1990). This adds definitions for the exception state and the floating point state as well.

`libc-test` still does not pass, because I use the type `[u128; 32]` for the `__v` field of `__darwin_arm_neon_state64` (in C it is `__uint128_t __v[32]`. `ctest2` does not translate `u128` to `__uint128_t` and the generated C code does not compile. Any ideas for working around this?


JohnTitor:
Thanks!

> libc-test still does not pass, because I use the type [u128; 32] for the __v field of __darwin_arm_neon_state64 (in C it is __uint128_t __v[32]. ctest2 does not translate u128 to __uint128_t and the generated C code does not compile. Any ideas for working around this?

I think we could use the same workaround in #2000.

bsteinb:
> I think we could use the same workaround in #2000.

I am not sure which workaround in #2000 you are referring to.

JohnTitor:
> I am not sure which workaround in #2000 you are referring to.

This one: https://github.com/rust-lang/libc/pull/2000#discussion_r546194181
I guess you could nest the array to auto-derive and not to use `u128`.

bsteinb:
I can do that, but it does not pass the tests either. It also does not have the correct alignment.

JohnTitor:
> I can do that, but it does not pass the tests either. It also does not have the correct alignment.

Hmm, then could you tweak the alignment with `#[repr(align)]`? The problem here is I don't have the target device and cannot debug at all :(

Thomasdezeeuw:
This passes on macOS 10.15.7 (Catalina), what errors are you seeing @bsteinb?

JohnTitor:
@Thomasdezeeuw The error should occur on Apple Silicon.

tamird:
I have an Apple Silicon device; how can I help? How do I reproduce the failure?

bsteinb:
Right now, if I run `cargo test` in `libc-test`, `ctest2` generates a function as follows:

```c
u128(*__test_field_type___darwin_arm_neon_state64___v(struct __darwin_arm_neon_state64* b))[32] {                  
  return &b->__v;                                                                                                 
}                        
```

Which the C compiler refuses to compile, saying:

```
libc/target/debug/build/libc-test-377e25964c16162d/out/main.c:32704:24: error: type specifier missing, defaults to 'int' [-Werror,-Wimplicit-int]
                  u128(*__test_field_type___darwin_arm_neon_state64___v(struct __darwin_arm_neon_state64* b))[32] {
                        ^
```

It does not recognize `u128` as a type. If I change that `u128` to `__uint128_t` by hand, I can compile the generated file just fine. Fixing this for good would require teaching `ctest2` to translate Rust's `u128` into C's `__uint128_t`, which I am not sure how good of an idea that is, given that `__uint128_t` is (I think) a compiler internal type.

If I change the pull request to have `__v: [[u64; 2]; 32]` instead of `__v: [u128; 32]` the generated function is:

```c
uint64_t(*__test_field_type___darwin_arm_neon_state64___v(struct __darwin_arm_neon_state64* b))[32][2] {           
  return &b->__v;                                                                                                 
}
```

About which the compiler complains:

```
libc/target/debug/build/libc-test-377e25964c16162d/out/main.c:32705:28: error: incompatible pointer types returning '__uint128_t (*)[32]' from a function with result type 'uint64_t (*)[32][2]' [-Werror,-Wincompatible-pointer-types]
                    return &b->__v;
                           ^~~~~~~
```

Also, as I mentioned above, `std::mem::align_of::<[u64; 2]>() == 8` whereas `std::mem::align_of::<u128>() == 16`.

So, two possible solutions:
- use `u128` and either teach `ctest2` how to translate it to C, or accept that the tests do not pass,
- use `[u64; 2]`, manually set the correct `#[repr(align(16))]` and accept that the tests do not pass.

tamird:
Translating to `__uint128_t` seems fine, since that's the type that is used in the apple header: https://github.com/opensource-apple/cctools/blob/fdb4825f303fd5c0751be524babd32958181b3ed/include/mach/arm/_structs.h#L61.

bsteinb:
I have pushed a change to use `[u64; 2]` with manual alignment instead of `u128`, since the same compatibility concerns that applied in #2000 should apply here. Also, there still seem to be concerns about using `u128` in FFI https://github.com/rust-lang/rust/issues/54341 (although on this particular target, `u128` and `__uint128_t` seem to match).

This does not match the declaration in the C headers 100%, but is closer than what exists now and probably as close as possible without using `u128`.

Once `u128` becomes viable (w.r.t. compatibility and FFI safety), this definition can be changed and with a translation for `u128` implemented in `ctest2`, the test suite should then pass. There is an open pull request to make the original ctest support `u128` etc.: https://github.com/gnzlbg/ctest/pull/89

tamird:
What about #1414 ? Would that approach work?

src/unix/bsd/apple/b64/aarch64.rs
```diff
@@ -14,4 +20,11 @@ s! {
         pub __cpsr: u32,
         pub __pad: u32,
     }
+
+    #[repr(align(16))]
```
JohnTitor, line 24: This should be in `align.rs`, see other definitions.

JohnTitor:
So, we should make it "at least it works" in this PR as `u128` issue won't be resolved right away. For `libc-test` failure, we could skip the check with a comment.

src/unix/bsd/apple/b64/aarch64.rs
bsteinb: Should I then move everything in `aarch64.rs` over to `align.rs` and guard it with a `cfg_if ... target_arch = "aarch64"`?

JohnTitor: I think:
- b64/aarch64/mod.rs: just contain `cfg_if` stuff for `align.rs` for now. Having mod.rs makes sense when we add more stuff for AArch64 target (i.e. Apple Silicon) in the future.
- b64/aarch64/align.rs: Should have everything in the current `aarch64.rs`, as you said.

mod.rs should be:
```rust
cfg_if! {
    if #[cfg(libc_align)] {
        mod align;
        pub use self::align::*;
    }
}
```

bsteinb:
With the tests ignoring the `__v` field and the changes from #2016, the test suite now builds and I get more fallout:

```
bad HW_MAXID value at byte 0: rust: 26 (0x1a) != c 28 (0x1c)
bad SHMLBA value at byte 1: rust: 16 (0x10) != c 64 (0x40)
bad boolean_t signed: rust: 0 (0x0) != c 1 (0x1)
bad max_align_t size: rust: 16 (0x10) != c 8 (0x8)
bad max_align_t align: rust: 16 (0x10) != c 8 (0x8)
size of max_align_t is 8 in C and 16 in Rust
bad PTHREAD_STACK_MIN value at byte 1: rust: 32 (0x20) != c 64 (0x40)
```

Would you like the fixes for those as separate pull requests or should I include them here?

JohnTitor:
> Would you like the fixes for those as separate pull requests or should I include them here?

Either way is fine to me if you separate commits!

bsteinb:
The test suite now fails on x86_64, because of the changes to the `HW_...` constants. These seem to be specific to macOS 11 (Big Sur) and the tests still run on 10.15 (Catalina).

JohnTitor:
Let's skip the test for now, you should tweak here:
https://github.com/rust-lang/libc/blob/7e7452fa28de2e4447f5f9f31357a497459e48d5/libc-test/build.rs#L193-L204

Big Sur is available as a preview on GHA but it may cause another issue and I'd avoid it.

JohnTitor (approved):
Looks good, thanks!

src/unix/bsd/apple/b64/mod.rs
```diff
@@ -2,7 +2,6 @@
 
 pub type c_long = i64;
 pub type c_ulong = u64;
-pub type boolean_t = ::c_uint;
 pub type mcontext_t = *mut __darwin_mcontext64;
```
JohnTitor, line 5: I'm not sure about `aarch64-apple-ios` but if it's the same as macOS, we could move this to align.rs on AArch64 (and mod.rs on x86_64).

JohnTitor:
Let's try again, @bors r+

PR #2016, opened 2021-01-01
fix glob symbol name on macOS aarch64

bsteinb:
This is another symbol like the ones #2007. Missed this one earlier, because the test suite did not get this far, because of #2008 .


bsteinb:
I have included this in #2008.

PR #2017, opened 2021-01-01
fix type of __priority_which_t/__rlimit_resource_t

Mic92:
This has the incorrect type as shown here:
https://github.com/bminor/glibc/blob/be9b0b9a012780a403a266c90878efffb9a5f3ca/resource/sys/resource.h#L44

According to posix this is also defined as int.
This is causing compatibility problems when compiling against glibc vs musl.


Mic92:
Whatever. These compile tests break posix!

PR #2018, opened 2021-01-03
aio functions do not require librt on FreeBSD

asomers:
On FreeBSD, the aio_ functions require librt _only_ if they use
SIGEV_THREAD completion notification.  However, due to Rust's originally
poor support for C unions, libc doesn't even expose some of the fields of
struct sigevent that SIGEV_THREAD requires.  Accordingly, there is no
need to link librt to programs using aio via libc.

This change partially reverts 8c23f77704d4749f5f137c92a048e22fea9d385e
from PR #1630 .

While I'm here, fix the linkage of lio_listio on DragonflyBSD.  It
_does_ require librt.

### DIFF
--- a/src/unix/bsd/freebsdlike/dragonfly/mod.rs
+++ b/src/unix/bsd/freebsdlike/dragonfly/mod.rs
@@ -1081,6 +1081,27 @@ extern "C" {
     ) -> *mut ::c_void;
 }
 
+#[link(name = "rt")]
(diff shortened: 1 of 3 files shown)


asomers:
cc @myfreeweb

JohnTitor (approved):
Looks reasonable to me, thanks!

asomers:
Thanks John!

PR #2023, opened 2021-01-07
Add mount ID to statx

XanClic:
This mirrors the modifications to `include/uapi/linux/stat.h` by [Linux commit fa2fcf4f1df1559a0a4ee0f46915b496cc2ebf60 (“statx: add mount ID”)](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=fa2fcf4f1df1559a0a4ee0f46915b496cc2ebf60).

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -26,7 +26,9 @@ s! {
         pub stx_rdev_minor: u32,
         pub stx_dev_major: u32,
         pub stx_dev_minor: u32,
-        pub __statx_pad2: [u64; 14],
+        pub stx_mnt_id: u64,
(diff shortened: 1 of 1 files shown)


XanClic:
I see the i686-unknown-linux-gnu test failed (and the x86_64 test was subsequently cancelled). As far as I can tell that is because Ubuntu 20.04.1 doesn’t have `STATX_MNT_ID` yet. I’m not sure how to proceed then.

Would it be OK to skip it by adding it to `test_linux()`’s `skip_const()` in `libc-test/build.rs`?

XanClic:
> Would it be OK to skip it by adding it to `test_linux()`’s `skip_const()` in `libc-test/build.rs`?

On second thought, I’ll just do that and close this pull request.

PR #2024, opened 2021-01-07
Add mount ID to statx

XanClic:
This mirrors the modifications to `include/uapi/linux/stat.h` by [Linux commit fa2fcf4f1df1559a0a4ee0f46915b496cc2ebf60 (“statx: add mount ID”)](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=fa2fcf4f1df1559a0a4ee0f46915b496cc2ebf60).

What changed in v2 (v1: #2023):
* Skip testing the new `STATX_MNT_ID` constant, because it’s only part of Linux as of 5.8. That’s too new e.g. for Ubuntu 20.04.1 (with a kernel derived from 5.4), which is what’s used by the CI.

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -13,7 +13,7 @@ s! {
         pub stx_uid: u32,
         pub stx_gid: u32,
         pub stx_mode: u16,
-        pub __statx_pad1: [u16; 1],
+        __statx_pad1: [u16; 1],
         pub stx_ino: u64,
         pub stx_size: u64,
         pub stx_blocks: u64,
@@ -26,7 +26,7 @@ s! {
         pub stx_rdev_minor: u32,
         pub stx_dev_major: u32,
         pub stx_dev_minor: u32,
-        pub __statx_pad2: [u64; 14],
+        __statx_pad2: [u64; 14],
     }
 
     pub struct statx_timestamp {
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -26,7 +26,9 @@ s! {
         pub stx_rdev_minor: u32,
         pub stx_dev_major: u32,
         pub stx_dev_minor: u32,
-        __statx_pad2: [u64; 14],
+        pub stx_mnt_id: u64,
+        __statx_pad2: u64,
+        __statx_pad3: [u64; 12],
     }
 
     pub struct statx_timestamp {
@@ -1171,6 +1173,7 @@ pub const STATX_SIZE: ::c_uint = 0x0200;
 pub const STATX_BLOCKS: ::c_uint = 0x0400;
 pub const STATX_BASIC_STATS: ::c_uint = 0x07ff;
 pub const STATX_BTIME: ::c_uint = 0x0800;
+pub const STATX_MNT_ID: ::c_uint = 0x1000;
 pub const STATX_ALL: ::c_uint = 0x0fff;
 pub const STATX__RESERVED: ::c_int = 0x80000000;
 pub const STATX_ATTR_COMPRESSED: ::c_int = 0x0004;
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2656,6 +2656,9 @@ fn test_linux(target: &str) {
             | "CAN_RAW_FILTER_MAX"
             | "CAN_NPROTO" => true,
 
+            // FIXME: Requires recent kernel headers (5.8):
+            "STATX_MNT_ID" => true,
+
             _ => false,
         }
     });


JohnTitor:
This should break the current use of statx, shouldn't it?

src/unix/linux_like/linux/gnu/mod.rs
```diff
@@ -26,7 +26,9 @@ s! {
         pub stx_rdev_minor: u32,
         pub stx_dev_major: u32,
         pub stx_dev_minor: u32,
-        pub __statx_pad2: [u64; 14],
```
JohnTitor, line 29: ...and found the padding field is public, it shouldn't be generally.

XanClic:
Thanks for the quick reply!

> This should break the current use of statx, shouldn't it?
> ...and found the padding field is public, it shouldn't be generally.

While I was under the impression that padding fields (or more generally, fields beginning with underscores) were not supposed to be used, I understand that guarantees nothing when the field is public. So perhaps there are users who indeed use this padding field, and then this change would break their code.

(Am I correct in assuming that if the fields were not public, this wouldn’t break the current use of statx?)

Do you have a recommendation on how to proceed? With a full deprecation cycle to make the padding fields private first?

joshtriplett:
@XanClic In other cases, libc has introduced methods to parse out the new fields from the padding, to avoid incompatible changes.

That said, this depends greatly on if anyone is actually using it. A code search that turns up zero references to `__statx_pad2` *might* be a sufficiently good argument to make this change.

XanClic:
Thanks, @joshtriplett. I couldn’t find any references that actually use rust-libc’s `statx.__statx_pad2`[1]. So from what I can tell, I think making the fields private should be feasible.

As for implementing a getter, can you point me to an example in rust-libc for a method that parses out a padding/reserved field? The only thing I found by grepping for `'fn.*(.*self'` that looked like such a case are the `siginfo_t::si_*` functions, which look more like something to implement access to various union fields. (Which I think is a different case, because you should be able to access all union fields. But I don’t think reserved/padding fields need to be accessible, especially after they’ve been given meaning.)

For me, the choices look like this:
1. Making the fields private would be the right thing to have done from the start. The “most correct” definition of `struct statx` would have a `stx_mnt_id` field. There’s no way to tell for sure whether this breaks existing users of rust-libc, but those users would have had to access a field with a double underscore prefix, so I would consider them warned. (Though I come from a C background originally, so I’m heavily biased in treating everything that begins with an underscore as effectively private anyway.)
2. Adding a getter would be simple and safe. But I wouldn’t be happy with the result. I expect the structs provided by rust-libc to match the original libc structs (or the one from the kernel header in this case). Of course it’s entirely possible that rust-libc never intended to meet that expectation, though.
3. We could do both; deprecate the padding fields, add a getter, and at some later point make the padding fields private (and add `stx_mnt_id`). This would be safer than (1), and still get us the “right” result eventually.

I can’t make the choice, because I don’t know how the rust-libc project weighs compatibility vs. closeness to the original libc definition.

(I’m kind of leaning towards (3), but I’d like to have an example for an existing method that parses out a reserved/padding field before adding a getter myself.)

___

[1] On github at least, looking for `__statx_pad2` in Rust code (https://github.com/search?l=Rust&q=__statx_pad2&type=Code) yields nearly only results which have `src/unix/linux_like/linux/gnu/mod.rs` in their filename and are just the `struct` definition, i.e. the field isn’t used there. (I.e., they are simply copies of rust-libc, but don’t really refence the field.)

The results that don’t look like copies of `rust-libc` are:
* [lemonrock/linux-support’s workspace/linux-support/src/file_descriptors/directory/c/statx.rs#L47](https://github.com/lemonrock/linux-support/blob/1ca2c2aae74a7cded627d7c693f8f6270a07155f/workspace/linux-support/src/file_descriptors/directory/c/statx.rs#L47) is the only place that actually uses this field. It defines its own `struct statx` and then has a function to set its padding to 0. So this `statx` has nothing to do with rust-libc’s `statx`.
* [passcod/defnew’s src/bin/libc-def.rs#L325](https://github.com/passcod/defnew/blob/3367a55869c7dd0fb6423a6c14378690ac09268c/src/bin/libc-def.rs#L325), which as far as I can tell deals with the C definition of the struct. So again, unrelated to rust-libc’s `statx`.
* [dubiosjim/nix-extra’s src/stat.rs#L258](https://github.com/dubiousjim/nix-extra/blob/967cc07a6c57eafa9a1eaaafb1ad511561299556/src/stat.rs#L258), which just defines its own `struct statx` (and doesn’t use the field).
* [mnussbaum/Launcher’s vendor/libc/src/unix/notbsd/linux/other/mod.rs#L27](https://github.com/mnussbaum/Launcher/blob/c6761d7d00cec9c1c019c029e174a749270322e4/vendor/libc/src/unix/notbsd/linux/other/mod.rs#L27), which looks like just an older copy out of rust-libc, when the path was still different. (Again, the field isn’t used.)

FWIW, on Google, I can only see the Rust struct definitions, some C results, and the rust-libc doc.

Gitlab doesn’t seem to have a global code search, so I can’t vouch for that.

joshtriplett:
It seems *reasonably* safe to make this change, then.

JohnTitor:
Hey thanks for searching @XanClic! It seems reasonable to change the fields.
I think we can merge this with hiding padding fields (`__statx_pad2` and `__statx_pad3`) and follow-up if it breaks something.

XanClic:
v2:
* I’ve added a new patch (patch 1), which makes both of `struct statx`’s padding fields (`__statx_pad1` and `__statx_pad2`) private. (I’ve looked for users of `__statx_pad1` and found the exact same result as for `__statx_pad2`, which I described in https://github.com/rust-lang/libc/pull/2024#issuecomment-760893218. That is, there don’t seem to be any users.)
* Patch 2 then keeps `__statx_pad2` and `__statx_pad3` private.

JohnTitor:
Thanks!
@bors r+

XanClic:
Thanks everyone!

PR #2026, opened 2021-01-08
Upgrade macOS version to 11.0 on CI

JohnTitor:
It's a preview builder but let's check if it's usable.


JohnTitor:
Okay, at least it works without tweaking on x86_64 but the issue is its booting time (seems an internal error occurs). It'd be great if we could remove some tweaks to skip tests by this but it isn't urgent, closing until the issue is resolved.

PR #2034, opened 2021-01-13
Haiku: add definitions for the Haiku's native sytem API.

nielx:
On the Haiku platform, the POSIX (and BSD) API coexists with the native API,
that has its origins on the BeOS platform. Unlike other UNIX-like platforms,
the native API is not an extension of the POSIX API, but instead exists sui
generis, and many of the POSIX concepts have their own native variety, with
relatively limited interoperability.

Nonetheless, the native API coexists in the same library as the standard C and
POSIX functions, namely libroot.so, and therefore this crate is a good place
to add bindings to it.

This commit implements most of Haiku's support kit, the most important parts
of the kernel kit, and a part of the storage kit.

### DIFF
--- a/libc-test/Cargo.toml
+++ b/libc-test/Cargo.toml
@@ -11,7 +11,7 @@ default-features = false
 [build-dependencies]
 cc = "1.0.61"
 # FIXME: Use fork ctest until the maintainer gets back.
-ctest2 = "0.3"
+ctest2 = "0.4"
 
 [features]
 default = [ "std" ]
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -32,6 +32,7 @@ fn do_ctest() {
         t if t.contains("dragonfly") => return test_dragonflybsd(t),
         t if t.contains("emscripten") => return test_emscripten(t),
         t if t.contains("freebsd") => return test_freebsd(t),
+        t if t.contains("haiku") => return test_haiku(t),
         t if t.contains("linux") => return test_linux(t),
         t if t.contains("netbsd") => return test_netbsd(t),
         t if t.contains("openbsd") => return test_openbsd(t),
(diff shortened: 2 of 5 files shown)


nielx:
@kallisti5 @jessicah @ahwayakchih Since you have contributed to the Haiku bindings before, could you have a look and perhaps add some comments?

nielx:
I'm working on the style and semantic issues pointed out by the CI.

kallisti5:
This patch looks like an awesome start to Haiku having improved rust support.   Our rust toolchain is fairly up to date, and critical rust crates [have](https://github.com/rust-lang/socket2/pull/136) been receiving [patches](https://github.com/deprecrated/net2-rs/pull/109) improving support.

nielx:
Changes since previous push:

- reorder the file a bit more to be consistent with the other files
- added some more bindings to the native API (kernel/fs_*.h)
- checked the validity using the libc-test process. I will send a PR to enable libc-test for Haiku soon.

This change is ready for review.

kallisti5:
@JohnTitor  This one's looking pretty solid and fills in a lot of missing base libc stuff from Haiku.  Thoughts?

JohnTitor:
Hey, thanks for PR!

@nielx Do you want me to make a new relase for `ctest2` and run libc-test here?

nielx:
@JohnTitor That would be appreciated. I have some follow-up work to this commit that adds support for libc-test and that applies some fixes that it pointed out, so if there is a ctest2 package that supports Haiku, it will allow others to reproduce the tests and build on that work.

JohnTitor:
Okay, published ctest2 0.4.0.

nielx:
I added support for Haiku in the testing framework, and the fixes for the issues that it pointed out.

nielx:
@JohnTitor Anything I can do to help you progress this?

JohnTitor:
Hey, thanks for the ping!
Could we add a Docker image and tweak the CI config so that we can run CI for Haiku?
One question, it seems this has some breaking changes, does it mean the old definitions are incorrect and won't work anyway? If so, I think this is acceptable. If not (e.g. it has been recently changed), I wonder this breaks some uses.

nielx:
> Could we add a Docker image and tweak the CI config so that we can run CI for Haiku?

It should be possible to set up a cross-compiling CI configuration to build libc for Haiku. I don't think it is possible to build and run the tests that way. Haiku is not (yet) virtualizable friendly, so I there's not such a thing as a Haiku VM image yet. I do wonder what the relative value of a continuous CI build would be, but we can certainly try to set it up.

> One question, it seems this has some breaking changes, does it mean the old definitions are incorrect and won't work anyway? If so, I think this is acceptable. If not (e.g. it has been recently changed), I wonder this breaks some uses.

The old definitions have always been incorrect.

JohnTitor (approved):
Okay, thanks for clarifying! Looks good to me now.

PR #2035, opened 2021-01-16
openbsd: make WIFCONTINUED() safe

### LINKED ISSUE #2001: WIFCONTINUED is unsafe on OpenBSD
For no apparent reason.

semarie:
it seems it was a missed part of #1891

Found it because rustc -nightly started to use `libc::WIFCONTINUED()` recently in a safe context and the build fail to build on OpenBSD.

Fixes #2001


JohnTitor (approved):
Thanks for catching this!

PR #2039, opened 2021-01-20
Add support for the ILP32 variant of AArch64

Amanieu:
This also fixes mutex initializers on big-endian AArch64.


JohnTitor (approved):
LGTM, this contains some breaking changes but I assume they currently doesn't work correctly.

Amanieu:
Failures seem to be unrelated to this PR.

PR #2042, opened 2021-01-22
Add port_notify struct for illumos and Solaris

pfmooney:
This adds a missing struct for configuring event ports notifications.  The `libc-test` suite passes on illumos (OmniOSCE) and is assumed to pass on Solaris, given that the interface has been present since Oracle forked from OpenSolaris.

### DIFF
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -364,6 +364,11 @@ s! {
         pub portev_user: *mut ::c_void,
     }
 
+    pub struct port_notify {
+        pub portnfy_port: ::c_int,
+        pub portnfy_user: *mut ::c_void,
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):
LGTM

> The libc-test suite passes on illumos (OmniOSCE) and is assumed to pass on Solaris, given that the interface has been present since Oracle forked from OpenSolaris.

👍, thanks for clarifying!

pfmooney:
> LGTM
> 
> > The libc-test suite passes on illumos (OmniOSCE) and is assumed to pass on Solaris, given that the interface has been present since Oracle forked from OpenSolaris.
> 
> +1, thanks for clarifying!

Thank you!

PR #2044, opened 2021-01-26
Fix CI on `i686-unknown-linux-musl`

JohnTitor:
I suspect `focal-20210119` breaks the build, uses an older image to fix the `Could not configure 'libc6:i386'` failure.


Amanieu:
I believe ubuntu has dropped 32-bit x86 support.

JohnTitor:
Hm then some other i686 targets may fail as well, let's check...

PR #2048, opened 2021-01-28
ci/semver.sh: Pass --locked when installing semverver

joshtriplett:
We don't, in general, want our CI to be the testbed for building
semverver with newer versions of its dependencies. Pass --locked to use
the checked-in Cargo.lock instead.

This works around https://github.com/rust-lang/cargo/issues/9101 .


joshtriplett:
Failure looks spurious.

@bors retry

PR #2051, opened 2021-01-29
Add aio_readv and aio_writev

asomers:
They are new in FreeBSD 13.

While I'm here, refactor freebsd to add a new FreeBSD 13 module.


asomers:
cc @gnzlbg and @pizzamig .

PR #2066, opened 2021-02-11
Linux: Add `preadv2` and `pwritev2` and associated constants

wmanley:
These functions are the same as `preadv` and `pwritev` but have a flags
parameter.  `preadv2()` and `pwritev2()` first appeared in Linux 4.6.
Library support was added in glibc 2.26.

See the definition of the constants in [linux/fs.h](https://github.com/torvalds/linux/blob/fcadab740480e0e0e9fa9bd272acd409884d431a/tools/include/uapi/linux/fs.h#L288-L301).

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -860,6 +860,15 @@ pub const EPOLLWAKEUP: ::c_int = 0x20000000;
 pub const SEEK_DATA: ::c_int = 3;
 pub const SEEK_HOLE: ::c_int = 4;
 
+// linux/fs.h
+
+// Flags for preadv2/pwritev2
+pub const RWF_HIPRI: ::c_int = 0x00000001;
+pub const RWF_DSYNC: ::c_int = 0x00000002;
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):
Looks good, thanks!

JohnTitor:
Hmm, I think the failure is due to our musl libc version (v1.1.24), the recent version should have them, right?

wmanley:
> Hmm, I think the failure is due to our musl libc version (v1.1.24), the recent musl should have them, right?

Looks like musl doesn't have `preadv2` at all. I suspect the same is true for bionic.  I've moved the existing `preadv2` and `pwritev2` into gnu.

I've also got a fallback generic implementation using `syscall(SYS_preadv2, ...)`, but I'll raise that as a separate PR once this one's gone in as I imagine it will require more discussion.

JohnTitor:
Okay, thanks! Let's give it another try.
@bors r+

JohnTitor:
Ah, consts should be also moved.

wmanley:
Ok, I've moved the consts too now.

PR #2069, opened 2021-02-12
Add memfd_create to linux and android

jwellhofer:
Android only added the wrapper in API level 30, i.e. Android 11. Should this have libc::syscall implementation for android ?

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -3914,6 +3914,8 @@ extern "C" {
         needlelen: ::size_t,
     ) -> *mut ::c_void;
     pub fn sched_getcpu() -> ::c_int;
+
+    pub fn memfd_create(name: *const ::c_char, flags: ::c_uint) -> ::c_int;
 }
 
 cfg_if! {


Amanieu:
I'm personally in favor of just exposing everything and letting the user deal with linker errors if their toolchain is too old. The only limitation is whether our CI accepts it.

Amanieu:
Maybe try upgrading the Android NDK in the CI scripts?

jwellhofer:
I tried to figure out the CI and get it to use Android API level 30, but couldn't figure out how to fiddle linker/linker64 out of the new sdk emulator image. 
I gave up and just added a syscall implementation for android.

Amanieu:
I'm not sure how I feel about this. Traditionally libc only contains bindings to existing functions rather than actual implementations.

Maybe just drop the Android part for now? Or is Android support something that you actually need?

jwellhofer:
I've found a way to get the newer android binaries, but it'll take me a while to figure out how to update the CI. Once i've got the CI working, i'll make a PR for that. I'll come back here when that is done.

Ralith:
Could we proceed with just the Linux version in the mean time? e.g. https://github.com/tokio-rs/tracing/issues/1698 calls for it.

jwellhofer:
Updating the android CI was such a pain that i gave up and just used a patched rust-libc.
I've put the linux only into the PR branch. Couldn't run the tests on my machine though because of some sockaddr_vm padding size mismatches. Hopefully it'll work on here.

Ralith:
Thanks!

PR #2070, opened 2021-02-13
Haiku: use raw pointers instead of references in convenience functions

nielx:
The Haiku API has some convenience macros to make it easier to call certain
functions. In the libc implementation, these are implemented as unsafe
functions. The previous choice was to take certain pointer parameters as
references, and do the conversion to raw pointers when the actual external
function was called.

However, this causes issues with the image_info struct, which needs to be
initialized in Rust, before a native API call is used to enter data. Since
part of this structure consists of function pointers, mem::zeroed() cannot be
used, since in Rust function pointers cannot be NULL. Thus one needs to use the
MaybeUnit<T> API to properly initialize it. This then makes it problematic to
use the convenience functions, as a MaybeUnit<image_info> cannot be converted
into an &mut image_info before it is marked as initialized with valid data. It can
be converted into a raw *mut image_info, so if the function accepts this as a
parameter it can be used.

For consistency, all convenience functions have been converted from using
references to using raw pointers.

### DIFF
--- a/src/unix/haiku/native.rs
+++ b/src/unix/haiku/native.rs
@@ -1023,14 +1023,14 @@ pub unsafe fn get_next_area_info(
     )
 }
 
-pub unsafe fn get_port_info(port: port_id, buf: &mut port_info) -> status_t {
+pub unsafe fn get_port_info(port: port_id, buf: *mut port_info) -> status_t {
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):

JohnTitor:
Thanks for clarifying, seems reasonable to me.
@bors r+

PR #2071, opened 2021-02-15
add definitions for s390x musl targets

kaniini:
Add support for s390x musl targets to libc.

I haven't added CI because I am not familiar with the pipelines, but would be glad to do so if somebody outlines what needs to be done.


kaniini:
CI is mostly blocking on rust-lang/rust#82166 being merged, which adds the s390x-unknown-linux-musl target.

Amanieu:
LGTM. Let's wait on rust-lang/rust#82166 so we can have CI set up.

I'm not too familiar with the libc CI, but basically just copy what we're already doing for other architectures on musl (e.g. mips, arm). The relevant files are all in `ci/`.

kaniini:
Yeah that's why I decided to go with #82166 first.  Then handle the libc bits and any std bits afterward.

workingjubilee (commented):
These all are destined to go into fn ioctl which is expecting a c_int.

kaniini:
Based on conversation in Zulip, I wonder if it makes sense to move a lot of these constants to the common part of the musl support code.

workingjubilee:
Yes, I think so, that's kind of how the libc crate is "supposed" to be designed (according to https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md anyways!). Probably using a similar structure as the "generic bits" vs. "arch-specific bits" pattern that musl uses for its own organization.

kaniini:
@Amanieu this should be good to go now.  rust#82166 has landed.

Amanieu:
Is ci/s390x-linux-musl.json still needed?

kaniini:
git, how does it even work?

Amanieu:
You need to add the target to .github/workflows/bors.yml so it gets picked up by CI.

kaniini:
oh, technically the s390x-unknown-linux-musl target is tier 3 atm.  can we just skip the CI?  or how do you want to tackle it?

Amanieu:
Ah good point, let's just skip it for now.

PR #2072, opened 2021-02-16
Add O_DIRECTORY to solarish

smklein:
FYI @pfmooney . I was trying to port [nix](https://crates.io/crates/nix) to illumos, and noticed this in the illumos-gate [headers](https://github.com/illumos/illumos-gate/blob/221e47fb90c5fcfe7add9a33f6c915ee5253ece9/usr/src/uts/common/sys/fcntl.h), but not in libc.

### DIFF
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -994,6 +994,7 @@ pub const O_EXCL: ::c_int = 1024;
 pub const O_NOCTTY: ::c_int = 2048;
 pub const O_TRUNC: ::c_int = 512;
 pub const O_NOFOLLOW: ::c_int = 0x20000;
+pub const O_DIRECTORY: ::c_int = 0x1000000;
 pub const O_SEARCH: ::c_int = 0x200000;
 pub const O_EXEC: ::c_int = 0x400000;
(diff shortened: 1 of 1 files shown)


src/unix/solarish/mod.rs
```diff
@@ -994,6 +994,7 @@ pub const O_EXCL: ::c_int = 1024;
 pub const O_NOCTTY: ::c_int = 2048;
 pub const O_TRUNC: ::c_int = 512;
 pub const O_NOFOLLOW: ::c_int = 0x20000;
+pub const O_DIRECTORY: ::c_int = 0x1000000;
```
pfmooney, line 997: It would be good to get confirmation from the Solaris folks that `O_DIRECTORY` exists there, and with the same value, before adding it at the top level like this (rather than the illumos-specific module)

Amanieu: Seems to exist according to https://docs.oracle.com/cd/E26502_01/html/E29032/open-2.html. I'm not sure about the value though.

pfmooney: I received confirmation that it's the same, so this should be fine.

PR #2074, opened 2021-02-18
Make extra_traits work with rustc-dep-of-std

### LINKED ISSUE #2064: Features 'rustc-dep-of-std' and 'extra_traits' don't work together
These features individually work, but if I enable both I get hundreds of build errors.

The repro I have is: clone rust repo, update `library/test/Cargo.toml` to add `features = ['rustc-dep-of-std', 'extra_traits']` to `libc`. If you add only `rustc-dep-of-std` the crate builds fine. But if you add both you get a few hundred build errors, most in the form "X not found, try core::blah::X". Diff for the repro:

```diff
diff --git a/library/test/Cargo.toml b/library/test/Cargo.toml
index 226557430df..80e964e84f5 100644
--- a/library/test/Cargo.toml
+++ b/library/test/Cargo.toml
@@ -13,7 +13,7 @@ getopts = { version = "0.2.21", features = ['rustc-dep-of-std'] }
 term = { path = "../term" }
 std = { path = "../std" }
 core = { path = "../core" }
-libc = { version = "0.2", default-features = false }
+libc = { version = "0.2", default-features = false, features = ['rustc-dep-of-std', 'extra_traits'] }
 panic_unwind = { path = "../panic_unwind" }
 panic_abort = { path = "../panic_abort" }
```

I was able to most of the build errors for my platform by just prefixing `Debug` with `core::fmt:Debug`, and the same for other symbols from core (`Eq`, `PartialEq`, ...). I also tried tweaking `cfg_attr`s in `lib.rs` but couldn't make this work with just `cfg_attr` tweaks.

Remaining issues are fixed by adding a few anonymous lifetime annotations, like `&core::fmt::Formatter<'_>` instead of `&Formatter`.

## Why is this needed?

I was hoping to use the `ctrlc` crate in rustc's test runner, which uses `nix`, which uses `libc` with the 'extra_traits' feature. Without 'extra_traits' it's possible to use libc in rustc test runner without the 'rustc-dep-of-std' feature, but with 'extra_traits' we need 'rustc-dep-of-std' too.

---

I'm happy to submit a PR to add prefixes to symbols like `Debug`, `Formatter`, `PartialEq`, etc. but I'm not sure if that's the best way to fix this. It's easily possiblewith a combination of `fd` and `sed`, but the diff is large.

osa1:
Fixes #2064

---

Files for other platforms also need to be updated.


src/unix/linux_like/mod.rs
```diff
@@ -1,3 +1,10 @@
+#[cfg(feature = "extra_traits")]
+use core::{
+    cmp::{Eq, PartialEq},
+    fmt::Debug,
+    iter::Iterator,
+};
```
JohnTitor, line 6: Seems this import style won't work on an old Rust.

src/unix/linux_like/mod.rs
```diff
@@ -264,7 +271,7 @@ cfg_if! {
         }
         impl Eq for epoll_event {}
         impl ::fmt::Debug for epoll_event {
-            fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+            fn fmt(&self, f: &mut ::fmt::Formatter<'_>) -> ::fmt::Result {
```
JohnTitor, line 274: And I think this also won't as I mentioned on the issue.

src/unix/linux_like/mod.rs
osa1: Should be fixed now.

src/unix/linux_like/mod.rs
osa1: I added explicit lifetimes to `Formatter` args.

src/macros.rs
```diff
@@ -73,7 +73,7 @@ macro_rules! s {
     (it: $(#[$attr:meta])* pub struct $i:ident { $($field:tt)* }) => (
         __item! {
             #[repr(C)]
-            #[cfg_attr(feature = "extra_traits", derive(Debug, Eq, Hash, PartialEq))]
+            #[cfg_attr(feature = "extra_traits", derive(core::fmt::Debug, core::cmp::Eq, core::hash::Hash, core::cmp::PartialEq))]
```
JohnTitor, line 76: According to CI this doesn't work on Rust 1.13.

osa1: Hm, that's weird. I don't know what the problem is with this syntax. I guess I'll have to install rustc 1.13 and experiment with it.

osa1: These are the errors:

```
error: expected `,`, found `::`
  --> src/macros.rs:76:61
   |
76 |             #[cfg_attr(feature = "extra_traits", derive(core::fmt::Debug, core::cmp::Eq, core::hash::Hash, core::cmp::PartialEq))]
   |                                                             ^^

error: expected `,`, found `::`
  --> src/macros.rs:76:61
   |
76 |             #[cfg_attr(feature = "extra_traits", derive(core::fmt::Debug, core::cmp::Eq, core::hash::Hash, core::cmp::PartialEq))]
   |                                                             ^^

error: expected `]`, found `::`
  --> src/macros.rs:76:61
   |
76 |             #[cfg_attr(feature = "extra_traits", derive(core::fmt::Debug, core::cmp::Eq, core::hash::Hash, core::cmp::PartialEq))]
   |                                                             ^^

error: Could not compile `libc`.
```

I have no idea what the problem is and how to work around this. Anyone have any ideas?

osa1: I guess qualified paths were not allowed in `derive` arguments in 1.13.

When building libc for rustc we know we're using a more recent version, so ideally I wouldn't have to deal with this.

JohnTitor: We can use `cfg(feature = "rustc-dep-of-std")` for it.

osa1: I don't understand how that would work. The `core::fmt::...` syntax will still be there and rustc 1.13 will try to parse it, no?

JohnTitor: That syntax won't work on 1.13: https://rust.godbolt.org/z/f9nf9rTTG

So, we want to make it work under `rustc-dep-of-std`, right? Then we can hide them behind `rustc-dep-of-std`, it's fine because we suppose it's used on rust-lang/rust.

JohnTitor:
I'm going to close as inactive, feel free to re-open when you get some time!

PR #2079, opened 2021-02-24
Implement accept4 on x86 android with `socketcall` syscall.

de-vri-es:
Linux x86 kernels before 4.3 only support the `socketcall` syscall rather than individual syscalls for socket operations. Since `libc` does a raw syscall for `accept4` on Android, it doesn't work on x86 systems.

This PR instead implements `accept4` for x86 android using `socketcall`. The value for `SYS_ACCEPT4` (in contrast to `SYS_accept4` :eyes:) is taken from the `linux/net.h` header.

Also note that the `socketcall` syscall takes all arguments as array of long ints. I've double checked with `glibc` to check how they pass arguments, since the Linux man page only says this: "args points to a block containing the actual arguments" and "only standard library implementors and kernel hackers need to know about socketcall()".

This should fix https://github.com/rust-lang/rust/issues/82400

### DIFF
--- a/src/unix/linux_like/android/b32/x86/mod.rs
+++ b/src/unix/linux_like/android/b32/x86/mod.rs
@@ -574,6 +574,35 @@ pub const REG_EFL: ::c_int = 16;
 pub const REG_UESP: ::c_int = 17;
 pub const REG_SS: ::c_int = 18;
 
+// socketcall values from linux/net.h (only the needed ones, and not public)
+const SYS_ACCEPT4: ::c_int = 18;
+
+f! {
+    // Sadly, Android before 5.0 (API level 21), the accept4 syscall is not
+    // exposed by the libc. As work-around, we implement it as raw syscall.
+    // Note that for x86, the `accept4` syscall is not available either,
+    // and we must use the `socketcall` syscall instead.
+    // This workaround can be removed if the minimum Android version is bumped.
+    // When the workaround is removed, `accept4` can be moved back
+    // to `linux_like/mod.rs`
+    pub fn accept4(
+        fd: ::c_int,
+        addr: *mut ::sockaddr,
+        len: *mut ::socklen_t,
+        flg: ::c_int
+    ) -> ::c_int {
+        // Arguments are passed as array of `long int`
+        // (which is big enough on x86 for a pointer).
+        let mut args = [
+            fd as ::c_long,
+            addr as ::c_long,
+            len as ::c_long,
+            flg as ::c_long,
+        ];
+        ::syscall(SYS_socketcall, SYS_ACCEPT4, args[..].as_mut_ptr())
+    }
+}
+
 cfg_if! {
     if #[cfg(libc_align)] {
         mod align;
--- a/src/unix/linux_like/android/b64/mod.rs
+++ b/src/unix/linux_like/android/b64/mod.rs
@@ -305,6 +305,22 @@ pub const UT_LINESIZE: usize = 32;
 pub const UT_NAMESIZE: usize = 32;
 pub const UT_HOSTSIZE: usize = 256;
 
+f! {
+    // Sadly, Android before 5.0 (API level 21), the accept4 syscall is not
+    // exposed by the libc. As work-around, we implement it through `syscall`
+    // directly. This workaround can be removed if the minimum version of
+    // Android is bumped. When the workaround is removed, `accept4` can be
+    // moved back to `linux_like/mod.rs`
+    pub fn accept4(
+        fd: ::c_int,
+        addr: *mut ::sockaddr,
+        len: *mut ::socklen_t,
+        flg: ::c_int
+    ) -> ::c_int {
+        syscall(SYS_accept4, fd, addr, len, flg) as ::c_int
+    }
+}
+
 extern "C" {
     pub fn getauxval(type_: ::c_ulong) -> ::c_ulong;
 }
--- a/src/unix/linux_like/android/b64/mod.rs
+++ b/src/unix/linux_like/android/b64/mod.rs
@@ -317,7 +317,7 @@ f! {
         len: *mut ::socklen_t,
         flg: ::c_int
     ) -> ::c_int {
-        syscall(SYS_accept4, fd, addr, len, flg) as ::c_int
+        ::syscall(SYS_accept4, fd, addr, len, flg) as ::c_int
     }
 }
 
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -2350,20 +2350,6 @@ f! {
     pub fn SO_EE_OFFENDER(ee: *const ::sock_extended_err) -> *mut ::sockaddr {
         ee.offset(1) as *mut ::sockaddr
     }
-
-    // Sadly, Android before 5.0 (API level 21), the accept4 syscall is not
-    // exposed by the libc. As work-around, we implement it through `syscall`
-    // directly. This workaround can be removed if the minimum version of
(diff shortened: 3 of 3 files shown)


src/unix/linux_like/android/b32/x86/mod.rs
```diff
@@ -574,6 +574,40 @@ pub const REG_EFL: ::c_int = 16;
...
 
+// socketcall values from linux/net.h (only the needed ones, and not public)
+const SYS_ACCEPT4: ::c_int = 18;
+
+f! {
+    // Sadly, Android before 5.0 (API level 21), the accept4 syscall is not
+    // exposed by the libc. As work-around, we implement it as raw syscall.
+    // Note that for x86, the `accept4` syscall is not available either,
+    // and we must use the `socketcall` syscall instead.
+    // This workaround can be removed if the minimum Andoird version is bumped.
```
sdroege, line 585:
```suggestion
    // This workaround can be removed if the minimum Android version is bumped.
```

de-vri-es: Whoops, thanks :)

src/unix/linux_like/android/b32/x86/mod.rs
```diff
@@ -574,6 +574,40 @@ pub const REG_EFL: ::c_int = 16;
...
+const SYS_ACCEPT4: ::c_int = 18;
+
+f! {
+    // Sadly, Android before 5.0 (API level 21), the accept4 syscall is not
+    // exposed by the libc. As work-around, we implement it as raw syscall.
+    // Note that for x86, the `accept4` syscall is not available either,
+    // and we must use the `socketcall` syscall instead.
+    // This workaround can be removed if the minimum Android version is bumped.
+    // When the workaround is removed, `accept4` can be moved back
+    // to `linux_like/mod.rs`
```
wngr, line 587: From my understanding, the  `socketcall` syscall needs to be made on x86 Android even on higher API versions, so that comment doesn't seem to be correct.

Further, a reference to the seccomp profile at https://android.googlesource.com/platform/bionic.git/+/master/libc/SYSCALLS.TXT might also be helpful?

de-vri-es: For higher API levels, libc already exposes `accept4`. I would expect the system libc to implement that correctly.

de-vri-es: To clarify: rather than a function in Rust `libc`, it would just be an extern fn calling into the system libc, which should work out of the box.

wngr: Maybe I'm a bit confused, but isn't that what I tried on https://github.com/rust-lang/rust/issues/82400#issuecomment-783405484, using `libc::accept4` directly on an Android 6.0 x86 emulator?

de-vri-es: I don't think so, because that was calling the function that this PR replaces for x86.

If you want to call `accept4` directly from the system libc, you have to declare it in an `extern "C"` block and call that :)

wngr: I can see clearly now the rain is gone :-)

src/unix/linux_like/android/b32/x86/mod.rs
```diff
@@ -574,6 +574,40 @@ pub const REG_EFL: ::c_int = 16;
...
+            len as ::c_long,
+            flg as ::c_long,
+        ];
+        socketcall(SYS_ACCEPT4, &args)
+    }
+}
+
+extern "C" {
+    // used to implement `accept4`, but not made public
+    fn socketcall(call: c_int, args: *mut c_long) -> c_int;
```
joshtriplett, line 608: I think libc exposes socketcall for some targets, and I thought Android was one of them.

joshtriplett:
I made one comment with something to double-check. Otherwise looks good to me.

joshtriplett:
r=me whenever bors starts responding again.

de-vri-es:
Pushed a fix, passes a local build for i686-android (can't run the tests though).

JohnTitor:
Failure:
```
   cargo:warning=/checkout/target/i686-linux-android/debug/build/libc-test-0ad756840b253b95/out/main.c:34804:24: error: use of undeclared identifier 'socketcall'; did you mean 'socketpair'?
  cargo:warning=                return socketcall;
  cargo:warning=                       ^~~~~~~~~~
  cargo:warning=                       socketpair
  cargo:warning=/android/ndk-i686/bin/../sysroot/usr/include/sys/socket.h:329:18: note: 'socketpair' declared here
  cargo:warning=__socketcall int socketpair(int __af, int __type, int __protocol, int __fds[2]);
  cargo:warning=                 ^
  cargo:warning=/checkout/target/i686-linux-android/debug/build/libc-test-0ad756840b253b95/out/main.c:34804:24: error: incompatible pointer types returning 'int (int, int, int, int *) __attribute__((cdecl))' from a function with result type 'int (*)(int, long *)' [-Werror,-Wincompatible-pointer-types]
  cargo:warning=                return socketcall;
  cargo:warning=                       ^~~~~~~~~~
  cargo:warning=2 errors generated.
  exit code: 1

  --- stderr
  rust version: 1.52.0-nightly
```

We should tweak the test.

de-vri-es:
This was probably a missing "#include <linux/net.h>" in the C tests. Added that now, I think.

src/unix/linux_like/android/b32/x86/mod.rs
```diff
@@ -574,6 +574,40 @@ pub const REG_EFL: ::c_int = 16;
...
+            len as ::c_long,
+            flg as ::c_long,
+        ];
+        socketcall(SYS_ACCEPT4, args[..].as_mut_ptr())
+    }
+}
+
+extern "C" {
+    // used to implement `accept4`, but not made public
+    fn socketcall(call: ::c_int, args: *mut ::c_long) -> ::c_int;
```
JohnTitor, line 608: I guess this is actually `__socketcall` on upstream?

de-vri-es: It seems there is no wrapper function for `socketcall` at all. Kinda makes sense for a syscall that is only an implementation detail for libc implementors. Looking back, including a Linux header to try and get a libc function declared doesn't make much sense :)

I changed the patch to use `::syscall(SYS_socketcall, SYS_ACCEPT4, ...)`, reverted the change to the tests and squashed it all.

JohnTitor:
Thanks, let's try again :)

@bors r=joshtriplett,JohnTitor

src/unix/linux_like/android/b64/mod.rs
```diff
@@ -305,6 +305,22 @@ pub const UT_LINESIZE: usize = 32;
...
+    // directly. This workaround can be removed if the minimum version of
+    // Android is bumped. When the workaround is removed, `accept4` can be
+    // moved back to `linux_like/mod.rs`
+    pub fn accept4(
+        fd: ::c_int,
+        addr: *mut ::sockaddr,
+        len: *mut ::socklen_t,
+        flg: ::c_int
+    ) -> ::c_int {
+        syscall(SYS_accept4, fd, addr, len, flg) as ::c_int
```
JohnTitor, line 320:
```suggestion
        ::syscall(SYS_accept4, fd, addr, len, flg) as ::c_int
```

de-vri-es: Doh, the other one now :) Thanks for the fix.

JohnTitor:
This is merged on master now but bors doesn't close it, manually closing.

de-vri-es:
Awesome, thanks for the reviews! :heart:

PR #2080, opened 2021-02-25
Add FreeBSD BPF structures

jbit:
This PR adds BPF structures that are common between [FreeBSD 11](https://github.com/freebsd/freebsd-src/blob/release/11.0.0/sys/net/bpf.h), [FreeBSD 12](https://github.com/freebsd/freebsd-src/blob/release/12.0.0/sys/net/bpf.h), [FreeBSD mainline/13](https://github.com/freebsd/freebsd-src/blob/main/sys/net/bpf.h) and [DragonFlyBSD 5.9](https://github.com/DragonFlyBSD/DragonFlyBSD/blob/v5.9.0/sys/net/bpf.h).

It also fixes the definition of `BPF_ALIGNMENT`, which should be equal to `sizeof(long)`, which is `4` on 32bit FreeBSD.

https://www.freebsd.org/cgi/man.cgi?query=arch&sektion=7
> All supported ABIs can be	divided	into two groups:
> * ILP32: `int`, `long`, `void *` types machine representations all have 4-byte size.
> * LP64: `int` type machine representation uses 4 bytes, while `long` and `void *` are 8 bytes.

I introduced the private `SIZEOF_LONG` const, since I didn't want to introduce a dependency on rust 1.24+ by depending on `libc_const_size_of`.

These changes allow my experimental crate to build on FreeBSD: https://github.com/jbit/powerline

### DIFF
--- a/src/unix/bsd/freebsdlike/mod.rs
+++ b/src/unix/bsd/freebsdlike/mod.rs
@@ -359,6 +359,14 @@ cfg_if! {
     }
 }
 
+// Non-public helper constant
+#[cfg(all(not(libc_const_size_of), target_pointer_width = "32"))]
+const SIZEOF_LONG: usize = 4;
+#[cfg(all(not(libc_const_size_of), target_pointer_width = "64"))]
+const SIZEOF_LONG: usize = 8;
+#[cfg(libc_const_size_of)]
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):
Looks good, thanks!

PR #2082, opened 2021-02-26
use GNU/Linux siginfo_t on Android

landfillbaby:
Bionic uses an unchanged siginfo.h from the upstream linux so why not? :)
I literally just copied everything with siginfo_t from unix/linux_like/linux/gnu/mod.rs to unix/linux_like/android/mod.rs

### DIFF
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -2824,6 +2824,17 @@ cfg_if! {
 }
 
 impl siginfo_t {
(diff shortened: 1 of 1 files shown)


Amanieu:
It seems like you removed the definition of the `siginfo_t` struct?

landfillbaby:
opps! fixed

landfillbaby:
i forgot, i added that back in to my test build, but forgot to commit it here!

landfillbaby:
ok fixed for real now

Issue #2061, opened 2021-02-08
The which_freebsd detection is cross-compilation unfriendly


nagisa:
The detection of `freebsd` version invokes a `freebsd-version` executable:

https://github.com/rust-lang/libc/blob/5f7c6c1ef93ff4fcfa3d767aac93c2b261c8f140/build.rs#L122-L123

however that only meaningfully works when building on a freebsd host. When building on any other host and targeting a freebsd, `libc` will always default to targeting a freebsd11.0:

https://github.com/rust-lang/libc/blob/5f7c6c1ef93ff4fcfa3d767aac93c2b261c8f140/build.rs#L26-L34

Thus making it difficult to e.g.

* Cross-compile a library or binary that relies on freebsd12.0 or freebsd13.0 APIs;
* CI test other libraries that rely on freebsd12.0 or freebsd13.0 APIs.

Note that adding a `freebsd-version` executable to the path is not a very viable solution – some builds running in parallel may target other versions of freebsd too.

JohnTitor:
Hm any suggestions here? I'm not familiar with FreeBSD, would it help that we could override the FreeBSD version by user flag?

nagisa:
I recently just realised that this detection logic only works in CI anyway (`if libc_ci`) so even when compiling on a `freebsd` host you will only ever get freebsd11 APIs at most.

This kind of makes the freebsd12 and freebsd13 modules dead currently.

(I don't really have any good suggestions, on how to approach this, alas)

nagisa:
Related: https://github.com/rust-lang/rfcs/pull/3036

raphaelcohn:
The problem of versioning of the underlying libc is a hard problem that probably needs addressing via the compilation target 'tuple' - not that I'm a fan of these (we copied this *very* broken concept from GNU's autotools build system cruft - have you ever seen how hard it is for them to maintain their `config` shell scripts). This issue is going to come up time and again as Rust matures. A recent minor example was the transition to 64-bit time with musl 1.20.

[...]

Issue #1848, opened 2020-07-28
musl: Change `time_t` definition on 32-bit targets according to "time64"


JohnTitor:
`time_t` is defined as `c_long` on musl:
https://github.com/rust-lang/libc/blob/b1268e43cd34ebc95d184928296fc73740cd5162/src/unix/linux_like/linux/musl/mod.rs#L3
and `c_long` is `i32` on 32-bit targets:
https://github.com/rust-lang/libc/blob/b1268e43cd34ebc95d184928296fc73740cd5162/src/unix/linux_like/linux/musl/b32/mod.rs#L1
But the `time_t` definition has been changed to 64-bit by https://github.com/bminor/musl/commit/38143339646a4ccce8afe298c34467767c899f51, since musl 1.2.0.
We will change our type to i64 on 32-bit targets as well at some point but we should announce it before changing.
cc #1846

mahkoh:
Isn't this broken either way as soon as C libraries are involved or am I missing something?

As documented in time64.html, C code compiled against >=1.2 assumes time_t = i64 whereas code compiled against <1.2 assumes time_t = i32 on 32 bit systems. If you use a C library that uses time_t in its interface (e.g. openssl), the time_t used by rust code must have the same size as in the musl headers.

abcfy2:
Any updates? I'm facing the same issue when building openssl:

```
   Compiling shadowsocks-rust v1.8.19 (/project)
error: linking with `arm-linux-musleabi-gcc` failed: exit code: 1
  |
[...]

PR #2088, opened 2021-02-28
musl 1.2.x support

kaniini:
Switch the CI to use musl 1.2.2 and fix any CI problems.

### DIFF
--- a/ci/install-musl.sh
+++ b/ci/install-musl.sh
@@ -5,7 +5,7 @@
 
 set -ex
 
-MUSL_VERSION=1.1.24
+MUSL_VERSION=1.2.2
 MUSL="musl-${MUSL_VERSION}"
 
 # Download, configure, build, and install musl:
--- a/src/unix/linux_like/linux/musl/mod.rs
+++ b/src/unix/linux_like/linux/musl/mod.rs
@@ -260,14 +260,9 @@ s_no_extra_traits! {
         pub __reserved: [::c_char; 256],
     }
 
-    // FIXME: musl added paddings and adjusted
-    // layout in 1.2.0 but our CI is still 1.1.24.
-    // So, I'm leaving some fields as comments for now.
-    // ref. https://github.com/bminor/musl/commit/
-    // 1e7f0fcd7ff2096904fd93a2ee6d12a2392be392
     pub struct utmpx {
         pub ut_type: ::c_short,
-        //__ut_pad1: ::c_short,
+        __ut_pad1: ::c_short,
         pub ut_pid: ::pid_t,
         pub ut_line: [::c_char; 32],
         pub ut_id: [::c_char; 4],
@@ -275,15 +270,15 @@ s_no_extra_traits! {
         pub ut_host: [::c_char; 256],
         pub ut_exit: __exit_status,
 
-        //#[cfg(target_endian = "little")]
-        pub ut_session: ::c_long,
-        //#[cfg(target_endian = "little")]
-        //__ut_pad2: ::c_long,
+        #[cfg(target_endian = "little")]
+        pub ut_session: ::c_int,
+        #[cfg(target_endian = "little")]
+        __ut_pad2: ::c_int,
 
-        //#[cfg(not(target_endian = "little"))]
-        //__ut_pad2: ::c_int,
-        //#[cfg(not(target_endian = "little"))]
-        //pub ut_session: ::c_int,
+        #[cfg(not(target_endian = "little"))]
+        __ut_pad2: ::c_int,
+        #[cfg(not(target_endian = "little"))]
+        pub ut_session: ::c_int,
 
         pub ut_tv: ::timeval,
         pub ut_addr_v6: [::c_uint; 4],
--- a/src/unix/linux_like/linux/musl/mod.rs
+++ b/src/unix/linux_like/linux/musl/mod.rs
@@ -1,15 +1,6 @@
 pub type pthread_t = *mut ::c_void;
 pub type clock_t = c_long;
-#[cfg_attr(
-    not(feature = "rustc-dep-of-std"),
-    deprecated(
-        since = "0.2.80",
-        note = "This type is changed to 64-bit in musl 1.2.0, \
-                we'll follow that change in the future release. \
-                See #1848 for more info."
-    )
-)]
-pub type time_t = c_long;
+pub type time_t = i64;
 pub type suseconds_t = c_long;
 pub type ino_t = u64;
 pub type off_t = i64;


JohnTitor:
Note that rust-lang/rust's CI also uses 1.1.24 currently (https://github.com/rust-lang/rust/blob/7e9a36fa8a4ec06daec581e23f390389e05f25e4/src/ci/docker/scripts/musl.sh#L28) and I think we should keep in sync with it.

kaniini:
@JohnTitor it is a chicken + egg issue.  we have to update both.  i intend to update both.

JohnTitor:
Yeah I just left a comment just in case.

ericonr:
Switching to musl 1.2.x only works despite of https://github.com/rust-lang/libc/issues/1848 because musl doesn't support symbol versioning, so the linker allows linking against the time32 compat symbols without issue. I have no idea how to properly solve this.

The switch to 1.2.x will also make `rustup` provided binaries incompatible with Void Linux (since we haven't updated yet), I think? Or well, potentially incompatible, if new functions start being used.

The ancient glibc used for builds is also unsupported (I think?), but it's used because it allows all sorts of systems to use the "universal" binaries without issue. Because of that, I'm not sure it's already a good time to update the libc used for *builds* (which, if all goes well in https://github.com/rust-lang/rust/pull/82556, should be different from the runtime libc).

ericonr:
Hmm, I believe this also needs to change the function names... The redirects to time64 functions in musl are implemented via headers, but linking to `clock_gettime` will get you the time32 impl, while the time64 impl is: `__clock_gettime64`.

richfelker:
If Rust code is going to pass the modern time-related data structures with 64-bit fields, it needs to honor (ideally derive somehow rather than hard-coding, but they're not going to change so you could just ship the table) the symbol name redirects from the libc headers. But if it's still passing 32-bit versions, it's fine, and will continue to be fine, to reference the legacy symbol names. As such I don't think anything is broken with regard to targeting musl 1.2.x right now. But there may be problems mixing Rust code with third-party C libraries that have already been built using the new types.

kaniini:
> Hmm, I believe this also needs to change the function names... The redirects to time64 functions in musl are implemented via headers, but linking to clock_gettime will get you the time32 impl, while the time64 impl is: __clock_gettime64.

Yes, the redirection is not yet implemented.  That allows for musl 1.1 to still be supported.  But Void really should upgrade to 1.2 already.

ericonr:
> Yes, the redirection is not yet implemented.

So changing `time_t` definition, as this PR does, is wrong.

> But Void really should upgrade to 1.2 already.

I don't disagree :)

Amanieu:
Could we keep `time_t` as 32-bit and expose a separate set of functions which use `time64_t`?

ericonr:
> Could we keep time_t as 32-bit and expose a separate set of functions which use time64_t?

I don't see how this would help the situation any. Pushing the choice of using time64 symbols on programmers will make them almost entirely useless, and people will have to find some way of deciding which symbols to use, at which point it'd just be better if `libc` could give them the appropriate version directly.

This is somewhat related to #2061... Could the `libc` crate try to build binaries (for the *target*, since host information is independent) in `build.rs` which link to symbols introduced in new versions of a certain library, in order to determine its version? So in this case, something like `__clock_gettime64`. And then create a file which the rest of the headers use to determine what `musl` version should be targeted?

Ideally `libc` would be a wrapper around `bindgen`, IMHO, but that would require shipping headers for platforms whose information is instead contained in `libc` in hardcoded form.

joshtriplett:
Given that this seems to be a necessarily backwards-incompatible change, would it potentially make sense to coordinate this with a potential switch of the `musl` target from static to dynamic?

ericonr:
It isn't necessarily backwards incompatible; I believe the BSDs have already requested some mechanism to support ABI incompatible versions of their own platforms. If such a thing was implemented, it'd be possible to carry the two versions "just fine". Given that the rust ecosystem seems to dislike feature tests (which would be the usual way of implementing such a mechanism), it will probably end up being backwards-incompatible.

Anyway, I think both of these incompatible changes are very different in what they touch (default linkage vs minimum supported version), so I don't see much of a reason to coordinate them. Could be convinced differently.

joshtriplett:
Then, rather than risking runtime incompatibility and difficult-to-debug brokenness (e.g. a -sys crate that's building Rust bindings to a C library that uses `time_t` in its interface, such as `openssl-sys`), I think it'd make sense to just start *requiring* musl 1.2 at compile-time and runtime. The target tier policy has a provision for how to upgrade the minimum requirements for a target; I think we should use that to bump the baseline requirement for the -musl targets to 1.2.

ericonr:
Requiring musl 1.2 would make sense for (old - the eventual official riscv32 target should be legacy free in this sense) 32bit targets, but not for 64bit ones.

joshtriplett:
> Requiring musl 1.2 would make sense for (old - the eventual official riscv32 target should be legacy free in this sense) 32bit targets, but not for 64bit ones.

That's a good point; the ABI hasn't changed for 64-bit targets, so the baseline wouldn't need to change.

> That said, how would you detect the version? Or would it be simply a contract?

We wouldn't need to detect the version. We would bind time-related symbols to the new 64-bit versions that don't exist in older versions of musl, and then we'd fail to link with older versions of musl.

richfelker:
Exactly. That seems very reasonable to me, and it automatically does the right thing for 64-bit targets without encoding more than the minimal needed knowledge (the symbol redirections).

BTW is Rust already encoding that sort of redirection for glibc targets, where it's needed to use most of the standard unix and stdio file functions (since the legacy symbols on 32-bit archs have 32-bit `off_t`)?

wesleywiser:
I've read through this thread and I'm trying to get a sense for what needs to be done to move forward the upgrade to musl 1.2.2. 

I believe this is the current state of this thread:

- We will upgrade to musl 1.2.2 and require that version at runtime. Per [RFC 2803](https://rust-lang.github.io/rfcs/2803-target-tier-policy.html) this requires approval from the compiler and release teams.
    - This is a breaking change for the `i686-unknown-linux-musl` target but not the `x86_64-unknown-linux-musl` target.

- There was a suggestion to bundle this breaking change with a change for the musl targets from static to dynamic but consensus seems to be that we don't need to do these at the same time. 

- There was a suggestion to expose a `time64_t` type but consensus seems to be that we should not go down that route. 

- There is the question of handling the redirects musl provides for 32-bit compatibility. Is this still a concern if we've decided to make this a breaking change to the target? 

I'm sure I've missed something so please feel free to correct me. I'd like to help push this along if possible 🙂

kaniini:
> There is the question of handling the redirects musl provides for 32-bit compatibility. Is this still a concern if we've decided to make this a breaking change to the target?

In general, legacy binaries will continue to run as expected.  From a source-compilation POV, musl only provides the 64-bit versions by default.  So, you should just be able to upgrade and make time_t 64-bit.

joshtriplett:
If we currently defaulted to dynamically linking musl, I would want to have more of a transition plan for requiring a newer version.

But since *currently* people would have to go out of their way to dynamically link musl, it seems unlikely that bumping our requirement forward will cause much grief.

Amanieu:
In any case, this PR is incomplete as it is: if we're changing the type definitions for the libc types then we must also change the functions to link to the `_time64` symbols instead of the normal ones. In the musl C headers this is done using the `__REDIR` macro, the Rust equivalent is the `link_name` attribute.

wesleywiser:
Thanks for the responses @kaniini, @joshtriplett and @Amanieu! It seems like the next step then is to change our functions to link with the new 64-bit symbols where applicable. After that, I guess we'll start an FCP to get approval to make this change to the target?

Amanieu:
Sounds good!

12101111:
Will it cause any problem if I use rust (stable) with musl 1.2.2 without this code change?

Amanieu:
Rust will work fine, the only issue is that it will use 32-bit `time_t` on 32-bit platforms.

danielframpton:
I had a look at adding the link attributes to redirect things to `time64` symbols for musl+32bit, and I think that part seems fairly straightforward (although there are fair number of changes).

However, I did run into an issue that is not related to the rename. When building the generated `main.c` file, I see a failure: `struct input_event` has no member named `time`. We define that as a `timeval` field, which is evidently not correct for musl with time64 on 32bit.

Not sure what the correct definition here is as we make this change? It seems like an interesting 32 vs 64bit difference and so I am curious how libc should manage the transition?

richfelker:
The `time` member of `struct input_event` is deprecated because it was wrongly declared with type `struct timeval` but can't match the type, since it always has `__kernel_ulong_t` for seconds. Instead applications should use `input_event_sec` and `input_event_usec`. It's still provided for backwards compatibility with old sources on 64-bit systems and legacy time32 systems, but programs using the old name won't work on time64, so they need to be fixed.

danielframpton:
Thanks @richfelker for the details.

@Amanieu, any suggestions from the libc maintainer perspective on the right approach here? 

In terms of managing the breaking change I don't see any great options, the best I have are one of:
1. introducing methods on `input_event` to access/set the `sec` and `usec` parts, dropping `time` on time64 (since we can't use `timeval`) and (optionally) marking `time` as deprecated elsewhere, or
2. removing `time` everywhere and adding `input_event_sec` and `input_event_usec` fields instead, dealing with the fallout.

I was also looking for inspiration from previous libc changes, but it looks like when we added 64-bit file offsets we did so by adding new functions (not redirecting to 64-bit symbols as proposed here) and I don't see that we have added time64 support for other linux targets.

Amanieu:
I think the best way forward here for the libc crate is to try to stay compatible with the glibc bindings. In practice, most Rust code is developed on the `x86_64-unknown-linux-gnu` target so we want the musl target to be compatible with that.

Currently the glibc bindings in this crate use a 32-bit `time_t`. The first step would be to add 64-bit time functions to the glibc bindings (e.g. `__localtime64`). See [this](https://www.gnu.org/software/libc/manual/html_node/64_002dbit-time-symbol-handling.html) for more information on 64-bit time in glibc.

Once that it implemented, we can expose 64-bit time in musl using the same types and symbols as the glibc bindings for source compatibility.

Finally, we can switch the Rust standard library to use the libc functions with 64-bit time (perhaps with some run-time detection for backwards-compatibility). In practice most Rust code doesn't use the libc bindings to get the current time, it will use `Instant` and `SystemTime` from the standard library.

danielframpton:
Thanks @Amanieu. I think this means that we should instead be moving musl forward 1.2.x without enabling time64 for 32 bit targets right now. I thought the earlier consensus was to try and couple multiple potentially disruptive changes but I think the update is worthwhile even if is a longer term project to get time64.

My other investigations also seem to confirm this is the right approach; updating the targets that use the in-box musl will require the musl 1.2.x update in rust to link so doing it first is helpful, and it appears we have multiple sources of musl in libc testing as well, including at least one target (a mips target for openwrt) that is currently musl 1.1.x until a future stable (22.xx) release. So we can’t (yet) require a time64 enabled musl here.

richfelker:
I would really encourage rethinking that and moving to the 64-bit `time_t` now while it's easy rather than getting stuck with something that's already at EOL (many users need to represent times at now+16 years). The time32 support in musl >1.2 is "ABI-compat only"; it's not intended for new programs, and continuing to use it will make it hard to interoperate with C and C++ libraries which will necessarily be built with 64-bit `time_t`. This is unlike glibc, where there are 2 build options and 32-bit is still the default. We've even been asked about options to build a libc without the legacy time32 symbols, which would be helpful to users who want to make sure their binaries are all Y2038-ready and not using legacy EOL'd interfaces; presently the main thing stopping this is autoconf (where many broken link-only tests check the symbol without including the headers, and would thereby get wrong results). Of course this would not be suitable for a systemwide shared libc that's intended to be able to run existing binaries, but I bring it up to highlight that time32 is not intended as supported API, just ABI.

Amanieu:
Since musl 1.2 is ABI-compatible with musl 1.1 we should be able to just upgrade it without any problems.

I feel that the correct place to make the transition to 64-bit time is in the Rust crates using libc (in particular `std`) rather than the libc crate itself. Essentially we would just have those crates switch to using `time64_t`-based APIs instead.

This is similar to how we currently handle `off_t` vs `off64_t`: https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/fs.rs

Amanieu:
I can see 2 ways of making the transition to 64-bit `time_t`.

The first option is to make a breaking change in the libc crate and simply change `time_t` directly.  This is likely to break existing crates in practice since Rust is much less forgiving than C with regards to implicit conversions between integer types.

The second option is what I described above. We would keep `time_t` as a 32-bit type in the `libc` crate and add a separate `time64_t` type with associated functions. The original functions would be wired to musl's 32-bit time_t ABI and the `time64_t` functions would be wired to musl's 64-bit time_t ABI. `#[deprecated]` attributes on the 32-bit time_t function can be used to encourage people to migrate to the 64-bit time_t API.

I personally favor the second option: the deprecation warnings will appear every time any code uses the 32-bit time APIs in libc, even in a transitive dependency, so there is little risk they would be ignored.

richfelker:
>  This is likely to break existing crates in practice since Rust is much less forgiving than C with regards to implicit conversions between integer types.

I don't see how you can claim it could break anything at the source level. Any program that builds on 64-bit archs already has `time_t` as a 64-bit type. The *only* possible breakage I see is ABI between dynamic-linked code using the old ABI with `time_t` in public interfaces, and Rust is not even encouraging the use of such dynamic linking at present as far as I can tell. Nothing else should break.

musl has always been very intentional about not having `foo64_t` types, and I would really **really** like for Rust to not try to impose that on us.

Amanieu:
A crater run might be tricky since `libc` is published as a normal crate on crates.io rather than shipped as part of the standard library. It's probably possible but I'm not familiar enough with crater.

richfelker:
> but if we change time_t to be 64-bit (pub type time_t = i64;) then it will fail to compile:

That code *already fails to compile* on any 64-bit target, so nobody is going to have written that. If they did they already got a bug report that it breaks on 64-bit targets.

Amanieu:
> Oh, right, duh! Perhaps we could audit some of the top reverse dependencies on crates.io or something to see what kind of impact this might have?

`libc` has 4470 reverse dependencies... I don't really fell comfortable making this change without some sort of crater run, but otherwise I'm happy to go ahead with a 64-bit time_t if there aren't too many regressions.

Perhaps we can configure crater to run with a `[patch.crates-io]` in the `Cargo.toml` of the crates being tested? I'm not too familiar with crater and don't really know who to ping for this.

danielframpton:
I took some time to look at this and wanted to see if we could get things moving along again. I have a branch [here](https://github.com/rust-lang/libc/compare/master...danielframpton:libc:musl1.2.x) that includes some additional commits that I would be happy to have pulled in here.

Those commits include redirects for methods with time parameters, including functions related to the `stat` structure which contains times (and for which I also had to perform updates).

Unfortunately, some of the (tier 3) platforms with musl support (hexagon and riscv32) do not have that musl support upstreamed and so we will need to decide how to proceed with getting those updated (assuming that a musl 1.2 port has been performed for them).

In CI today we also use a toolset from OpenWRT, which has a release candidate with musl 1.2 so we would need to update to that (or the final release when it is made).

Amanieu:
Thanks @danielframpton! I think that with your changes, we should be able to make the migration to musl 1.2.

Could you explain what the situation is with riscv32 and hexagon? Are the musl ports for those platforms still stuck on musl 1.1?

wesleywiser:
> Could you explain what the situation is with riscv32 and hexagon? Are the musl ports for those platforms still stuck on musl 1.1?

I've been working with @danielframpton on this and the problem we ran into was that we haven't been able to test any changes for these Tier 3 platforms because they have no CI or bors coverage. hexagon seems to have a musl 1.2.2 fork [here](https://github.com/quic/musl/tree/hexagon) and riscv32 support hasn't landed in upstream musl either as far as we can tell. Currently, the branch linked above doesn't really make changes to either target since we just weren't sure what to with them. Given that these are Tier 3 targets, we thought this was probably ok and they can be fixed in later PRs by people who can validate their changes. 

The other thing worth mentioning is that `mips-unknown-linux-musl` and `mipsel-unknown-linux-musl` use the OpenWRT distro in CI which doesn't currently have a stable release with musl 1.2. There is an [upcoming release](https://github.com/openwrt/openwrt/releases/tag/v22.03.0-rc6) which includes musl 1.2 so we could choose to use a pre-release version of OpenWRT for those Tier 2 targets for now.

richfelker:
> Could you explain what the situation is with riscv32 and hexagon? Are the musl ports for those platforms still stuck on musl 1.1?

No, there is no old time ABI for new/future archs and no symbol redirections. All new archs are natively time64 from the beginning with the unadorned symbol names. This also applies to any old archs being newly added to musl later, like for example sparc32.

Amanieu:
Since the condition is more complex than just `cfg(all(target_env = "musl", target_pointer_width = "32"))` then we should instead have a `musl_time64_abi` cfg set by the build script which is enabled when we need to link to the time64 symbols instead of the normal ones.

richfelker:
The condition is basically "target is musl and `__time64` symbol exists in libc".

Amanieu:
I would actually prefer a hard-coded list of archs. We don't have an easy way to check for the existence of a symbol in a library. We could attempt to probe for it like autoconf does but this would significantly impact the build time of the `libc` crate and would require adding additional build-time dependencies (`cc`).

richfelker:
OK. Since the list of affected archs is non-expanding (just those existent at the time time64 was introduced) it's not really a big deal to hard-code it.

wesleywiser:
Hey everyone, just wanted to provide an update on where we're at. 

I've been working on implementing @Amanieu's feedback about introducing a `musl_time64_abi` cfg flag to control when we link to the `time64` family of symbols (in [my musl-1.2 branch](https://github.com/rust-lang/libc/compare/master...wesleywiser:libc:musl-1.2)). That's been implemented in 6468d63bfdbe064a27b36fbfc63208d052be77d6. 

Along the way, I discovered that when `libc-test` runs in CI, it doesn't actually link with the version of musl CI is testing, it uses the musl bundled with rustc's target. I've opened #2893 to fix that (it's the first commit in my branch). 

After fixing CI, that uncovered a number of other changes we needed to make for various targets (d76bd2f47c9ffdc74f9b2ffdf3e2c4cbbdcf0f99...94d7b3a97bbe5428eb2eea19e868129e849c7130): mostly just adjusting constants and field orderings. 

Finally, I also updated the mips CI runners to use OpenWRT 22.03 rc6 artifacts so they can be tested with mips 1.2.3 (ede8be23f05e01ff890cbf00f727c3d590564b26...b79bf16c06637c8aad277b63d8b9aacc52dd12a8).

At this point, I believe all of the prior discussion points have been addressed and the next step is to figure out how we can do a crater run to see what breaks in the ecosystem. I hope to have some time later in the week to work on that.

wesleywiser:
A first [crater run of these changes](https://github.com/rust-lang/rust/pull/102891#issuecomment-1317145761) has been completed for `i686-unknown-linux-musl`. I'm still analyzing the full results but to give a brief summary:

- 5749 regressions out of 248247 packages tested.
- 104 root regressions.
- 56 root regressions appear to be unrelated to this change (flaky tests, a few nightly compiler ICEs, etc) and I believe can be ignored.
- Of the remaining 48 regressions, 27 of these are because `timespec` now contains private padding fields on `i686-unknown-linux-musl` where it previously didn't. As such, a `timespec` struct can no longer be created using struct literal syntax. 

@Amanieu how does libc typically handle this kind of change? I'm thinking it would make sense to target some of the most used crates in this category and update them [like this](https://github.com/rust-lang/rust/pull/102891/files#diff-3183738c82f747fd92bb573712073aa384b16e14d154bf2df18f41077edc31cb) so that they can build with both older and newer versions of libc.

Amanieu:
Where are these padding fields coming from? I don't see them in our definition of `struct timespec`.

wesleywiser:
[Here](https://github.com/rust-lang/libc/compare/master...wesleywiser:libc:musl-1.2#diff-a828924ffbc30c564e49bc3652ffa5375a123671b228fc99b5775a29cff6eea5) on the Rust side which corresponds to [here](https://git.musl-libc.org/cgit/musl/tree/include/alltypes.h.in?h=v1.2.3#n44) on the musl side.

wesleywiser:
If we don't include the trailing padding on little-endian, then the size of `timespec` will shrink from 24 bytes to 16. Is there a different way you're thinking of to keep the size of the struct the same?

Amanieu:
That seems wrong: the size should be 16 bytes on all platforms. Looking at the C code, the padding is only added when `sizeof(time_t) != sizeof(long)` (i.e. on 32-bit targets only).

richfelker:
> If we don't include the trailing padding on little-endian, then the size of `timespec` will shrink from [24 bytes](https://rust.godbolt.org/z/ne4ezfPTK) to [16 bytes](https://rust.godbolt.org/z/zWs8GjPMK). Is there a different way you're thinking of to keep the size of the struct the same?

I'm confused. The size of `struct timespec` is 16 bytes on all musl archs (with time64). The form is 8 bytes of `tv_sec` followed by 8 bytes of padding together with `tv_nsec`. On 64-bit archs there is no padding because `tv_nsec` (`long`) is already 8 bytes. On 32-bit archs, `tv_nsec` (`long`) is only 4 bytes, and the 4 bytes of padding appear on whichever side makes the location of the significant bits line up with where they would be in a 64-bit word with the arch's endianness. If you're getting it as 24 bytes somehow, then Rust has a wrong definition for the type.

richfelker:
> * Of the remaining 48 regressions, 27 of these are because `timespec` now contains private padding fields on `i686-unknown-linux-musl` where it previously didn't. As such, a `timespec` struct can no longer be created using struct literal syntax.

In C, creation of a struct literal `timespec` is not valid without designated initializers. I don't know what the Rust equivalent is, but it's fundamentally wrong to assume an ordering of the members or absence of any additional members. Code doing so is buggy. The spec is:

> The <time.h> header shall declare the timespec structure, which shall include **at least** the following members:
> 
>     time_t  tv_sec    Seconds. 
>     long    tv_nsec   Nanoseconds. 

(Emphasis mine.)

Amanieu:
> In C, creation of a struct literal `timespec` is not valid without designated initializers. I don't know what the Rust equivalent is, but it's fundamentally wrong to assume an ordering of the members or absence of any additional members. Code doing so is buggy. The spec is:

Rust only has one kind of initialization for structs, which is similar to C designated initializers. Except it requires *all* fields to be specified, whereas C will zero-initialized any omitted fields.

The "proper" solution would be to zero-initialize the struct and then fill in its fields one-by-one as done [here](https://github.com/rust-lang/rust/pull/102891/files#diff-3183738c82f747fd92bb573712073aa384b16e14d154bf2df18f41077edc31cb). The downside is that this is much more verbose and requires some unsafe code.

wesleywiser:
Sorry, that's my bad. I was looking at x86_64 in godbolt and messed up the code while trying to get it to compile there. I see what you're saying now and that should work fine. I'll try that and see what the crater results look like.

wesleywiser:
> I see what you're saying now and that should work fine.

So that actually does not work fine. I thought `i64` on `i686` has 8 byte alignment and therefore the overall structure would have 8 byte alignment, forcing the final, trailing padding bytes, but it only has 4 byte alignment. However, without the explicit trailing padding, `timespec` only ends up taking [12 bytes instead of 16](https://github.com/wesleywiser/libc/actions/runs/3517900390/jobs/5896299534#step:5:3384). 

@Amanieu I think that means we have to include the private padding field which will prevent use of the literal struct initialization syntax unless you have another idea how we can force `sizeof(timespec) == 16` without private fields.

richfelker:
Indeed, setting alignment to 8 does not work for archs without a full alignment requirement for 64-bit types, and of course doesn't solve the big endian case where the padding is before `tv_nsec` not after.

Maybe this is a sign that Rust should have [an extension to do?] default initialization like C does for unmentioned members. I think the core issue here is that the Rust feature is incompatible with structures that may have implementation-defined or internal/private fields that are not part of the API, and thus that application code is not allowed to mention by name.

wesleywiser:
> ~We can add `#[repr(align = 8)]` which will round the size of the structure up to 16 bytes.~
> 
> EDIT: Actually that would still be wrong since the alignment of the struct in C is still 4. In that case I think we have no choice but to leave the padding in.

Yeah I gave that a try earlier today and the CI flagged the mismatched alignments right away. I don't think we want to do that. 

> Maybe this is a sign that Rust should have [an extension to do?] default initialization like C does for unmentioned members. I think the core issue here is that the Rust feature is incompatible with structures that may have implementation-defined or internal/private fields that are not part of the API, and thus that application code is not allowed to mention by name.

I think usually what we'd do for that case is something like this:

```rust
pub struct Padding<T> {
    _: T // this field is not pub so it can't be used outside of the module
}

pub struct timespec {
    pub tv_sec: time_t,
    pub tv_nsec: ::c_long,
    pub __pad0: Padding<u32>, // this field is pub to allow struct initialization syntax to work
}

pub fn zeroed() -> timespec {
    timespec {
        // 0 init all fields
    }
}
```

You would then do something like `timespec { tv_sec: 1234, tv_nsec: 5678, ..zeroed() }` when you want to literally initialize a `timespec` value. `..zeroed()` copies any unmentioned field values from the return value of `zeroed()`

@Amanieu would there be interest in trying to explore some kind of solution like that (perhaps using `Default` instead of dedicated zero-initialization functions)? We would still have to help migrate dependent crates to support that but it could be done backwards compatibly. 

The other option I see would be to pick some of the most used crates impacted by this and switch them to the `MemUninit::zeroed()` approach like I did with libstd. That can also be done backwards compatibly. 

A second crater run revealed that at least 97% of the 5749 regressions from the initial run are soley due to `timespec` having private fields. I suspect most of this comes down to a few key crates that are widely used on top of libc like `nix`. So migrating a few of those to either of those approaches could resolve most of the breakage.

Amanieu:
I don't mind making the padding fields public to allow struct initialization. For the `zeroed` function, we would ideally have some variant of a `Pod` trait to indicate that the type can be safely zero-initialized. However this feels more like a long-term plan for libc 2.0.

In the short term I think it's best to do what you suggested:

> The other option I see would be to pick some of the most used crates impacted by this and switch them to the MemUninit::zeroed() approach like I did with libstd. That can also be done backwards compatibly.

Just a slight nit: I prefer `mem::zeroed()` over `MaybeUninit::zeroed().assume_init()` since there's no uninitialized memory involved.

thomcc:
> `..zeroed()` copies any unmentioned field values from the return value of `zeroed()`.

Note that they'd still have to be public, and `#[non_exhaustive]` doesn't help either (due to how FRU syntax is desugared). But we don't *strictly* have to match names with what musl calls its unmentionable fields and what we call them, so I guess that's fine.

wesleywiser:
@kaniini Could you pull the changes from my [`musl-1.2` branch](https://github.com/rust-lang/libc/compare/master...wesleywiser:libc:musl-1.2) into this PR? I've rebased and verified that a bors run will be green. 

After that, I think we're ready to start an FCP! 🙂

JohnTitor:
Closing in favor of https://github.com/rust-lang/libc/pull/3068, thanks for the PR!

PR #2100, opened 2021-03-04
Add gettid() for Linux

### LINKED ISSUE #2076: Add gettid() for Linux
I'm using x86_64-unknown-linux-gnu. I need to use `gettid()` in Rust, and it's not in this crate yet. It's got a simple signature:

https://man7.org/linux/man-pages/man2/gettid.2.html

It is Linux-specific. I could probably add it if it would help, but I'm not positive where to add it. From searching, it seems gnu, musl, uclibc, and Android include it, but I don't think Emscripten does (not sure).

I suppose it needs added in either

https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/mod.rs

or in both

https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/android/mod.rs
https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/linux/mod.rs

And it seems like the tests are automatic so they wouldn't need changed?

TheDoctor314:
Fixes #2076

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2737,6 +2737,9 @@ fn test_linux(target: &str) {
             // assume it's a int instead.
             "getnameinfo" if uclibc => true,
 
+            // FIXME: This needs musl 1.2.2 or later.
+            "gettid" if musl => true,
+
             _ => false,
         }
     });
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -2809,6 +2809,8 @@ extern "C" {
     pub fn regfree(preg: *mut ::regex_t);
 
     pub fn android_set_abort_message(msg: *const ::c_char);
+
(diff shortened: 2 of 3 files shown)


JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

JohnTitor:
Failed on musl, we should skip tests for now as it requires a recent kernel.

TheDoctor314:
Hello,
To be honest, I am a bit lost here what the next steps from me should be.
I apologize for any inconvenience.

JohnTitor:
Okay, no worries! Could I push some commits to your branch to tweak tests?

TheDoctor314:
Yeah, sure! Go for it!

JohnTitor:
Emscripten doesn't seem to have it actually, moved the declaration under Linux and Android.
@bors r+

PR #2105, opened 2021-03-07
Add more unix domain socket peer information to more targets

tormol:
* Add NetBSD peer credentials struct `unpcbid` and socket option `SO_PEEREID` (similar to Linux struct `ucred`).
  (and some other socket options too along the way)
* Add `struct xucred`, `LOCAL_PEERCRED` and related to DragonFly BSD too.
* Add new `cr_pid` field to FreeBSD `struct xucred` (added in FreeBSD 13).
  It is added through an union as in the C header to make sure the alignment is correct, but could probably be a simple field with padding too.
* Add `SO_PEERSEC` and `SO_PASSEC` to Android and all Linux architectures that they hadn't already been added to.


JohnTitor (approved):
Looks good, thanks!

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -225,6 +234,73 @@ cfg_if! {
...
+        impl ::hash::Hash for xucred {
+            fn hash<H: ::hash::Hasher>(&self, state: &mut H) {
+                self.cr_version.hash(state);
+                self.cr_uid.hash(state);
+                self.cr_ngroups.hash(state);
+                self.cr_groups.hash(state);
+                #[cfg(libc_union)]
+                self.cr_pid__c_anonymous_union.hash(state);
+                #[cfg(not(libc_union))]
+                self.__cr_unused1.hash(self.state);
```
JohnTitor, line 300:
```suggestion
                self.__cr_unused1.hash(state);
```

tormol: Fixed, I had just started looking into the failure.
man I had a lot of errors in the initial code for those trait impls. (that's what happens when i started on it a year ago and didn't think to check with that feature enabled :))
I haven't tested the trait impls, and only [tested the new field on FreeBSD 12](https://cirrus-ci.com/task/5849412262952960) (it is zeroed there). I've tested the NetBSD and DragonFlyBSD structs in VMs and `SO_PEERSEC` locally on x86.

Do we actually want to compare and hash unused fields? I added it for maximum backwards compatibility with the previous derives, but it doesn't seem useful.
With the union only the bytes of the pid are compared/hashed.

JohnTitor: Yeah, actually, I'm not sure if it makes sense to hash unused or reserved fields. I wonder we could clean-up those things at some point but for now I'm going to merge as-is.

PR #2106, opened 2021-03-11
Add structs from linux/uinput.h + a couple of input-related constants

coolreader18:
I could also add the full constant list from `input-event-codes.h`, but that's about 1000 lines of constants and it's not necessary to interact with the ioctls, just to interpret the values they return/accept. Also, I made the `_CNT` constants usize instead of u16 since they're intended to be buffer lengths, not actual `input_event.code` (or whatever else) values.

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -284,6 +284,40 @@ s! {
         pub u: [u32; 7],
     }
 
+    pub struct uinput_ff_upload {
+        pub request_id: ::__u32,
+        pub retval: ::__s32,
+        pub effect: ff_effect,
+        pub old: ff_effect,
+    }
+
+    pub struct uinput_ff_erase {
+        pub request_id: ::__u32,
+        pub retval: ::__s32,
+        pub effect_id: ::__u32,
+    }
+
+    pub struct uinput_setup {
+        pub id: input_id,
+        pub name: [::c_char; UINPUT_MAX_NAME_SIZE],
+        pub ff_effects_max: ::__u32,
+    }
+
+    pub struct uinput_abs_setup {
+        pub code: ::__u16,
+        pub absinfo: input_absinfo,
+    }
+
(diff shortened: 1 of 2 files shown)


src/unix/linux_like/linux/mod.rs
```diff
@@ -282,6 +282,40 @@ s! {
         pub u: [u32; 7],
     }
 
+    pub struct uinput_ff_upload {
+        request_id: ::__u32,
+        retval: ::__s32,
+        effect: ff_effect,
+        old: ff_effect,
+    }
```
JohnTitor, line 290: We should make these fields public?

coolreader18: Oh, whoops, yeah 😅

JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

JohnTitor:
Does this require musl 1.2.0 or higher? If so, we should skip tests for musl as we use musl 1.1.24 on CI.

JohnTitor:
Failures on `mips-unknown-linux-gnu`:
```
bad SW_MAX value at byte 1: rust: 16 (0x10) != c 15 (0xf)
bad SW_CNT value at byte 3: rust: 17 (0x11) != c 16 (0x10)
```
Looks like mips CI has the old value for `SW_MAX`, I'm not sure though. Could you also skip on mips?

coolreader18:
Huh, I think github was down for a little bit but the tests seem to be passing.

JohnTitor:
Thank you for the quick fix, let's give it another try.
@bors r+

JohnTitor:
@bors delegate+ feel free to r=me (r=JohnTitor) when you fix the test.
Or, if you want to merge this quickly, you could also skip them unconditionally and I'll follow-up later.

coolreader18:
whoops, didn't realize the build had already retried... @bors cancel ?

coolreader18:
Jeez, finally 😅

Issue #1359, opened 2019-05-23
sighandler_t has an incorrect type on most unix targets


gnzlbg:
For example, it is always defined as `size_t`, but on Android it should be of type `extern "C" fn(c_int, *mut siginfo, *mut c_void)`.

PR #2107, opened 2021-03-11
Creating proper type for sighandler_t using anonymous union on Unix

landhb:
Solves #1359

This PR allows you to use the sigaction struct on Unix properly. For instance you'll be able to create a new sigaction struct passing all three types of handlers, below demonstrates both the creation of a handler of type `void (*sa_handler)(int)` and for `SIG_IGN` which is a const `size_t`:

```rust
extern "C" fn signal_handler(signal: i32) {todo!()}

let mut act = libc::sigaction {
    sa_sigaction: libc::sighandler_t {
        sa_handler: Some(signal_handler),
    },
    sa_mask: newmask,
    sa_flags: 0,
    sa_restorer: None,
};

// Demonstrating use of SIG_IGN constant 
let _ignore_action = libc::sigaction {
    sa_sigaction: libc::SIG_IGN,
    sa_mask: newmask,
    sa_flags: 0,
    sa_restorer: None,
};

unsafe {
    libc::sigaction(libc::SIGUSR1, &mut act, core::ptr::null_mut());
}
```

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -43,9 +43,13 @@ pub type locale_t = *mut ::c_void;
 s_no_extra_traits! {
     pub union _c_anonymous_sigaction_handler {
         pub sa_handler: Option<extern "C" fn(c_int) -> ()>,
-        pub sa_sigaction: Option<extern "C" fn(c_int, *mut siginfo_t, *mut c_void) -> ()>,
+        pub sa_sigaction: Option<extern "C" fn(
+            c_int,
+            *mut siginfo_t,
+            *mut c_void
+        ) -> ()>,
         pub default: size_t,
-    } 
+    }
 }
 
 s! {
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -43,7 +43,7 @@ pub type locale_t = *mut ::c_void;
 s_no_extra_traits! {
     pub union _c_anonymous_sigaction_handler {
         pub sa_handler: Option<extern "C" fn(c_int) -> ()>,
-        pub sa_sigaction: Option<extern "C" fn(c_int, *mut siginfo_t, *const c_void) -> ()>,
+        pub sa_sigaction: Option<extern "C" fn(c_int, *mut siginfo_t, *mut c_void) -> ()>,
         pub default: size_t,
     } 
 }
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -27,7 +27,7 @@ pub type uid_t = u32;
 pub type gid_t = u32;
 pub type in_addr_t = u32;
 pub type in_port_t = u16;
-pub type sighandler_t = ::size_t;
+pub type sighandler_t = _c_anonymous_sigaction_handler;
 pub type cc_t = ::c_uchar;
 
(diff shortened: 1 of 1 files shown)


landhb:
Working on fixing the test suite errors.

landhb:
All checks are passing now.

src/unix/mod.rs
```diff
@@ -1712,3 +1724,26 @@ cfg_if! {
...
+                    .finish()
+            }
+        }
+        impl ::hash::Hash for __c_anonymous_sigaction_handler {
+            fn hash<H: ::hash::Hasher>(&self, state: &mut H) {
+                unsafe { self.default.hash(state) };
+            }
+        }
+    }
+}
```
JohnTitor, line 1749: Could you move this before the consts declaration?

src/unix/mod.rs
```diff
@@ -27,7 +27,7 @@ pub type uid_t = u32;
 pub type gid_t = u32;
 pub type in_addr_t = u32;
 pub type in_port_t = u16;
-pub type sighandler_t = ::size_t;
+pub type sighandler_t = __c_anonymous_sigaction_handler;
```
JohnTitor, line 30: Is the current type _wrong_ on all the UNIX platforms?

JohnTitor (commented):
Thanks for the PR!
But this should be a breaking change and I'd hold off unless it's a serious issue.

JohnTitor:
Also cc @Amanieu as I'm unfamiliar with this.

src/unix/mod.rs
landhb: I believe so. Here's what I can dig up so far:

BSD/XNU uses one of the function pointer variants `void (*sig_t)(int)`, and calls it `sig_t`:

https://github.com/apple/darwin-xnu/blob/main/bsd/sys/signal.h#L500
https://github.com/NetBSD/src/blob/trunk/sys/sys/signal.h#L172

Haiku has both function pointer variants:

https://github.com/haiku/haiku/blob/d0e006f5569f24e73a4d83f5c9766f1132160d7a/headers/posix/signal.h#L61-L62

Hermit has the `void (*signal_handler_t)(int)`:

https://github.com/hermitcore/libhermit/blob/63e67e1af1e56fbb51f14c89997bbaac11a37cfa/include/hermit/signal.h#L46

Redox as well:

https://github.com/redox-os/relibc/blob/29e2f2923170a7a52c5645ab002564bfa4b0c43b/src/header/signal/mod.rs#L35

src/unix/mod.rs
landhb: Yep! Moved in 6913b31. Let me know if that's where you meant.

Amanieu:
The new version is more correct but I'm afraid it will be a significant breaking change so we should hold off on it until the next major release.

safinaskar:
@landhb , breaking release of libc will be released soon. So, consider to continue this work

tgross35:
@landhb if you are interested in continuing this work, it would be great to have for 1.0 (the current `main` branch).

@rustbot author

landhb:
Sorry for the delay @tgross35 @safinaskar 

I'll take a look and rebase.

tgross35:
@landhb gentle ping, are you still interested in working on this?

PR #2108, opened 2021-03-12
Haiku: some type fixes for ioctl() arguments and readv()/writev(), and some missing constants

nielx:
These two commits fix some issues that became visible while building libstd.


JohnTitor:
So does this mean the current declaration won't work properly anyway?

nielx:
> So does this mean the current declaration won't work properly anyway?

I assume you are talking about readv() and writev()

With the caveat that I am not an expert what goes on at the linker level/CPU register level, we have indications that both varieties will work. The valid input of the readv() size parameter is between 0 and 1024 anyway, and in the libstd it was even maximized to 16. These values fit both in an 8 byte unsigned integer, as well as a 4 byte signed integer. Furthermore, on 32 bit systems size_t and int are both 4 bytes, so there is not even a mismatch of the size. My [fellow Haiku contributors](https://review.haiku-os.org/c/haiku/+/3770) think that it won't matter anyway; on 64 bit systems the 4 byte int will be put in an 8 byte register anyway, so if we pass a 4 byte value from Rust to the 8 byte value to Haiku, it will not break. I am doing some extra tests right now to make sure that this is correct, but so far it seems like this is correct.

There are unwanted downstream effects when not conforming to the standards, meaning that we will have to put specific cases in libstd for Haiku. It seems easier to fix the API, both in Rust's `libc` and on Haiku itself.

JohnTitor (approved):
Alright, thanks for clarifying!

PR #2109, opened 2021-03-13
Add regression test infrastructure

### LINKED ISSUE #2104: Idea for regression test for accidental removal of definitions
I've been thinking about how to add regression tests for issues like #2101, wherein between versions 0.2.86 and 0.2.87 a system call was accidentally removed for a specific target. The structure for libc is quite complex with many `#[cfg]`s throughout the code base (for good reasons), this makes it quite easy to mistakenly remove an item for a target and not catch it in review.

Currently there is libc-test, which tests if a given definition in Rust is the same as in C. But this doesn't catch if a definition is removed between libc version.

So, now my idea. Keep around a list of definitions (functions, constants etc.) that are exported by libc in a text file (separate from the code base) for each OS/architecture combination. For example such a file for Android:

```text
# linux-android.txt
accept
accept4
SOCK_NONBLOCK
SOCK_CLOEXEC
# Etc...
```

To make architecture specific changes we can use a diff-like format: `+` for additional definitions, `-` for unsupported definitions. 

```text
# arm-linux-androideabi.txt
-accept4 # Android on Arm doesn't support `accept4(2)` (Not true, just an example).
+accept6 # A system call only supported on Android on Arm (again just an example).
```

A build script could read these files for a given target, i.e. for the `arm-linux-androideabi` it would include both `linux-android.txt` and `arm-linux-androideabi.txt`, and convert it to a Rust test file. Such a test file would look something like the following:

```rust
#![allow(unused_imports)]

// Used files:
// linux-android.txt
// arm-linux-androideabi.txt

use libc::{
    accept, // From linux-android.txt.
    SOCK_NONBLOCK, // From linux-android.txt.
    SOCK_CLOEXEC, // From linux-android.txt.
    accept6, // From arm-linux-androideabi.txt.
    // Etc...
};
```

If a definition is removed between versions the test file wouldn't build, and thus fail the test/CI.

This is just an idea I had, it might be too much work to maintain such lists of definitions. I can put in the work for the setup/build script, but I realise that that is not were most of the work is in.

Thomasdezeeuw:
Please the commit messages for details.

I still need to add lists for the following targets, but I got the major ones I think.

TODO:
* aarch64-unknown-hermit
* x86_64-unknown-hermit
* x86_64-pc-solaris
* x86_64-sun-solaris
* sparcv9-sun-solaris
* x86_64-fortanix-unknown-sgx
* x86_64-unknown-illumos
* asmjs-unknown-emscripten
* wasm32-unknown-emscripten
* wasm32-unknown-unknown
* wasm32-wasi
* Check symbols added after commit ed45c2649b848bc2df48ba10d68194d45da03b75.

TODO: add a bit to the contributing guide about adding to these lists.

Closes #2104.


Thomasdezeeuw:
@JohnTitor before I rebase could you review this first?

JohnTitor:
One question is how we deal with the FreeBSD version difference but otherwise looks good to me!

Thomasdezeeuw:
I think it's best to merge this as is and I'll try to find some time to add lists for the other targets.

For FreeBSD can make lists like `freebsd-12.txt`, `freebsd-13.txt`, etc where `freebsd-13.txt` includes `freebsd-12.txt`, etc.

JohnTitor:
Seems the TODO in the OP isn't addressed, I'd like to see it in this PR.

Thomasdezeeuw:
I don't really have the time to add all the lists for the targets any time soon, so it will take a while.

JohnTitor:
Ah I mean the second TODO, `add a bit to the contributing guide about adding to these lists.` (sorry, I didn't notice we have two TODOs). It should be helpful for contributors. Adding all the targets isn't necessary for now.

Thomasdezeeuw:
> Ah I mean the second TODO, `add a bit to the contributing guide about adding to these lists.` (sorry, I didn't notice we have two TODOs). It should be helpful for contributors. Adding all the targets isn't necessary for now.

Done.

JohnTitor (approved):

JohnTitor:
Thank you for working on it!
@bors r+

JohnTitor:
Looks like the check isn't working well on `i686-unknown-linux-musl`.

Thomasdezeeuw:
Is these are a couple more problems with the lists, is there any way to not cancel the checks if one fails? That might be a bit quicker than fixing them one by one.

JohnTitor:
You can set `fail-fast` to `false` on `.github/workflows/bors.yml`. Feel free to do testing with a try build.

@bors delegate+

JohnTitor:
@Thomasdezeeuw If you're busy, may I push commits to your branch to get this merged?

Thomasdezeeuw:
> @Thomasdezeeuw If you're busy, may I push commits to your branch to get this merged?

Yes and yes 👍

Thomasdezeeuw:
@JohnTitor are you actively working on this? I have some time for this this afternoon (~2 from now).

JohnTitor:
@Thomasdezeeuw Currently no, feel free to work on it :)

Thomasdezeeuw:
@JohnTitor I've created some TODO lists to keep track of some of the symbols that currently aren't available on all combinations of arch/env (e.g. gnu vs. musl) for an OS. For example as far as I know all those linux-*.txt lists (e.g. linux-aarch64.txt) indicate that the symbols in those files aren't properly defined for all architectures of Linux.

But hopefully even with these TODO list this is at least ready to be merged.

JohnTitor:
Sounds nice, let's try: @bors try

Thomasdezeeuw:
The failure for arm-linux-androideabi seems unrelated: `thread 'main' panicked at 'failed to find successful test run', /tmp/runtest.rs:46:9` (https://github.com/rust-lang/libc/runs/2253378077?check_suite_focus=true).

JohnTitor:
Hmm, but this happens twice (https://github.com/rust-lang/libc/runs/2245677639?check_suite_focus=true) on semver check so I guess we need some work there.

Thomasdezeeuw:
Looking at https://github.com/rust-lang/libc/blob/bd7e0522f7631a0be106fa3ea3f6f6d8ad514b66/ci/runtest-android.rs#L42-L47, does it need to output something on Android?

Thomasdezeeuw:
I can't run it locally so hopefully 418c481 fixes it.

Thomasdezeeuw:
Thank you @bors, @JohnTitor CI is green 👍

JohnTitor:
Thanks for working on it! @bors r+

semarie:
I have the following error on OpenBSD:

```
+ cargo test --no-default-features --manifest-path libc-test/Cargo.toml --target x86_64-unknown-openbsd               
   Compiling libc-test v0.2.92 (/data/semarie/repos/rust/libc-rs/libc-test)
error[E0432]: unresolved imports `libc::EMULTIHOP`, `libc::ENOLINK`, `libc::WEXITED`, `libc::WNOWAIT`, `libc::WSTOPPED
`                                                           
   --> /data/semarie/repos/rust/build-libc/build/x86_64-unknown-openbsd/debug/build/libc-test-f95bb6991c27a551/out/sem
ver.rs:87:5                                                                                                           
    |            
87  |     EMULTIHOP,                                        
    |     ^^^^^^^^^ no `EMULTIHOP` in the root              
...                                                                                                                   
98  |     ENOLINK,                                                                                                    
    |     ^^^^^^^ no `ENOLINK` in the root                                                                            
...                                                         
431 |     WEXITED,                                                                                                    
    |     ^^^^^^^ no `WEXITED` in the root                                                                            
...                                                         
438 |     WNOWAIT,                                          
    |     ^^^^^^^ no `WNOWAIT` in the root                                                                            
439 |     WSTOPPED,                                                                                                   
    |     ^^^^^^^^ no `WSTOPPED` in the root                                                                          
    |                                                      
help: a similar name exists in the module                                                                             
    |                                                      
98  |     EMLINK,                                          
    |     ^^^^^^                                           
help: a similar name exists in the module                                                                             
    |                                                      
431 |     WIFEXITED,                                       
    |     ^^^^^^^^^                                        
help: a similar name exists in the module                                                                             
    |                                                      
439 |     WIFSTOPPED,                                      
    |     ^^^^^^^^^^                                       

error: aborting due to previous error                                                                                 

For more information about this error, try `rustc --explain E0432`.
error: could not compile `libc-test`                                                                                  

To learn more, run the command again with --verbose.                                                                  
warning: build failed, waiting for other jobs to finish...                                                            
error: build failed                                        
```

it seems it is because the mentioned symbols are in `unix.txt` whereas OpenBSD (which is unix) doesn't have them. I checked that we really don't have them (it isn't just rust-libc missing them).

how to correct it ? move the symbols from `unix.txt` to target specific file ?

Thomasdezeeuw:
Sorry for the late response @semarie, but I see you've already fixed the problem, thanks.

PR #2112, opened 2021-03-15
Bump to 0.2.89

joshtriplett:


JohnTitor:
Could you also bump the libc-test version?

joshtriplett:
@JohnTitor Ah, good catch, thank you.

Perhaps we should add a quick test to CI that makes sure those two version numbers match.

JohnTitor (approved):

JohnTitor:
Thanks :)
@bors r+

PR #2113, opened 2021-03-15
CI: Consolidate and blanket-ignore kernel definitions on non-glibc

joshtriplett:
Skip definitions from the kernel on non-glibc Linux targets.
They're libc-independent, so we only need to check them on one
libc. We don't want to break CI if musl or another libc doesn't
have the definitions yet. (We do still want to check them on
every glibc target, though, as some of them can vary by
architecture.)

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2561,6 +2561,38 @@ fn test_linux(target: &str) {
     });
 
     cfg.skip_const(move |name| {
+        if !gnu {
+            // Skip definitions from the kernel on non-glibc Linux targets.
+            // They're libc-independent, so we only need to check them on one
+            // libc. We don't want to break CI if musl or another libc doesn't
+            // have the definitions yet. (We do still want to check them on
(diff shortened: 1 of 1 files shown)


joshtriplett:
This also experimentally removes `P_PIDFD` from the ignores on glibc. Let's see if that builds now; if not, I can re-add it.

JohnTitor:
FYI: https://www.githubstatus.com/incidents/s654n76c1bwr
Let's wait until it's resolved.

JohnTitor (approved):

JohnTitor:
Looks reasonable to me.

> 
> 
> This also experimentally removes `P_PIDFD` from the ignores on glibc. Let's see if that builds now; if not, I can re-add it.

Let's check it, @bors r+

JohnTitor:
Actually, failed on `mips-unknown-linux-gnu`.

joshtriplett:
Fixed.

@bors r=JohnTitor

PR #2114, opened 2021-03-15
Add more syscall constansts for Linux

voidc:
In https://github.com/rust-lang/libc/pull/1897#issuecomment-695889272, there was interest in adding constants for recently added Linux syscalls. I went ahead and added consts for syscalls up until (the currently unreleased) Linux 5.12. Here is an overview:

| Syscall           | NR  | since |
| ----------------- | --- | ----- |
| pidfd_send_signal | 424 | 5.1   |
| io_uring_setup    | 425 | 5.1   |
| io_uring_enter    | 426 | 5.1   |
| io_uring_register | 427 | 5.1   |
| open_tree         | 428 | 5.2   |
| move_mount        | 429 | 5.2   |
| fsopen            | 430 | 5.2   |
| fsconfig          | 431 | 5.2   |
| fsmount           | 432 | 5.2   |
| fspick            | 433 | 5.2   |
| pidfd_open        | 434 | 5.3   |
| clone3            | 435 | 5.3   |
| close_range       | 436 | 5.9   |
| openat2           | 437 | 5.9   |
| pidfd_getfd       | 438 | 5.9   |
| faccessat2        | 439 | 5.9   |
| process_madvise   | 440 | 5.10  |
| epoll_pwait2      | 441 | 5.11  |
| mount_setattr     | 442 | 5.12  |

([source](https://github.com/torvalds/linux/blob/1e28eed17697bcf343c6743f0028cc3b5dd88bf0/include/uapi/asm-generic/unistd.h#L825-L865))

The constants are added for `-gnu` and `-musl` targets.
I'm unsure to what extend they are actually present in the headers of these libraries.

Support is added for all architectures except hexagon.


JohnTitor:
Thanks! I'd like to wait until #2113 gets merged.

voidc:
Sure! It seems like glibc doesn't contain `SYS_*` constants for syscalls which were added in 5.9 or later. Should I add exceptions for those to the build script?

JohnTitor:
> 
> 
> It seems like glibc version used by CI doesn't contain `SYS_*` constants for syscalls which were added in 5.9 or later. Should I add exceptions for those to the build script?

Yeah, let's do so with a comment. But the above PR may cause a merge conflict so I'd wait anyway.

JohnTitor (approved):

src/unix/linux_like/linux/gnu/b32/sparc/mod.rs
```diff
@@ -909,8 +909,26 @@ pub const SYS_preadv2: ::c_long = 358;
...
+pub const SYS_io_uring_setup: ::c_long = 425;
+pub const SYS_io_uring_enter: ::c_long = 426;
+pub const SYS_io_uring_register: ::c_long = 427;
+pub const SYS_open_tree: ::c_long = 428;
+pub const SYS_move_mount: ::c_long = 429;
+pub const SYS_fsopen: ::c_long = 430;
+pub const SYS_fsconfig: ::c_long = 431;
+pub const SYS_fsmount: ::c_long = 432;
+pub const SYS_fspick: ::c_long = 433;
+pub const SYS_pidfd_open: ::c_long = 434;
```
JohnTitor, line 922: We re-defined it (L911).

voidc: oops, sorry :)

PR #2116, opened 2021-03-17
Add `renameat2` and change their flags' type to `c_uint`

JohnTitor:
Taking over #1508. This changes the type of some flags but it should've been `c_uint`.

Closes #1508

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -1477,6 +1477,13 @@ extern "C" {
         offset: ::off_t,
         flags: ::c_int,
     ) -> ::ssize_t;
+    pub fn renameat2(
+        olddirfd: ::c_int,
+        oldpath: *const ::c_char,
+        newdirfd: ::c_int,
+        newpath: *const ::c_char,
+        flags: ::c_uint,
+    ) -> ::c_int;
 }
 
 extern "C" {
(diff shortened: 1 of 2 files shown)


bgilbert:
The type change causes a build failure for [openat](https://crates.io/crates/openat) (https://github.com/tailhook/openat/issues/35).  Should this have gone through a deprecation period per the [breaking change policy](https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md#breaking-change-policy)?

phi-gamma:
Thanks for merging this!

JohnTitor:
> 
> 
> The type change causes a build failure for [openat](https://crates.io/crates/openat) ([tailhook/openat#35](https://github.com/tailhook/openat/issues/35)). Should this have gone through a deprecation period per the [breaking change policy](https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md?rgh-link-date=2021-03-19T00%3A46%3A03Z#breaking-change-policy)?

@bgilbert I'd consider it as bugfix since `renameat2` needs `c_uint` as their flag and the initial definition (i.e. `c_int`) should be wrong in that context. But hmm is there any use of the flags as `c_int`?

bgilbert:
Well, it can be a bugfix and still be a breaking change. :slightly_smiling_face:  But meanwhile, openat has been updated with a workaround in https://github.com/tailhook/openat/pull/36.

PR #2119, opened 2021-03-19
Define SO_INCOMING_CPU for {aarch64,x86_64}-unknown-linux-musl

oherrala:
This continues where https://github.com/rust-lang/libc/pull/2115 left off and hopefully fixes `socket2` crate on these two targets.

PR #2123, opened 2021-03-24
Define SO_INCOMING_CPU for i686-unknown-linux-gnu

Heliozoa:
Hi! Following the example from #2119, I've tried adding the constant to i686-unknown-linux-gnu as well to fix a similar issue.


JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

PR #2124, opened 2021-03-26
Define SO_INCOMING_CPU for i686-unknown-linux-musl

fbecart:
Following the examples from https://github.com/rust-lang/libc/pull/2119 and https://github.com/rust-lang/libc/pull/2123, to fix the same issue with `socket2`: https://github.com/rust-lang/socket2/issues/213


JohnTitor (approved):
It'd be great if we could clean-up this at some point but I'm fine to add it to each arch + bits for now.

PR #2141, opened 2021-04-08
Unified Linux TCP socket options

zonyitoo:
https://github.com/torvalds/linux/blob/master/include/uapi/linux/tcp.h

### DIFF
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -957,6 +957,43 @@ pub const TCP_WINDOW_CLAMP: ::c_int = 10;
 pub const TCP_INFO: ::c_int = 11;
 pub const TCP_QUICKACK: ::c_int = 12;
 pub const TCP_CONGESTION: ::c_int = 13;
+pub const TCP_MD5SIG: ::c_int = 14;
+cfg_if! {
+    if #[cfg(all(target_os = "linux", any(target_env = "gnu", target_env = "musl")))] {
+        // WARN: deprecated
+        pub const TCP_COOKIE_TRANSACTIONS: ::c_int = 15;
+    }
+}
+pub const TCP_THIN_LINEAR_TIMEOUTS: ::c_int = 16;
+pub const TCP_THIN_DUPACK: ::c_int = 17;
+pub const TCP_USER_TIMEOUT: ::c_int = 18;
+pub const TCP_REPAIR: ::c_int = 19;
+pub const TCP_REPAIR_QUEUE: ::c_int = 20;
+pub const TCP_QUEUE_SEQ: ::c_int = 21;
+pub const TCP_REPAIR_OPTIONS: ::c_int = 22;
+pub const TCP_FASTOPEN: ::c_int = 23;
+pub const TCP_TIMESTAMP: ::c_int = 24;
+pub const TCP_NOTSENT_LOWAT: ::c_int = 25;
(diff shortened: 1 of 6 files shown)


src/unix/linux_like/mod.rs
```diff
@@ -957,6 +957,31 @@ pub const TCP_WINDOW_CLAMP: ::c_int = 10;
...
+pub const TCP_REPAIR_WINDOW: ::c_int = 29;
+pub const TCP_FASTOPEN_CONNECT: ::c_int = 30;
+pub const TCP_ULP: ::c_int = 31;
+pub const TCP_MD5SIG_EXT: ::c_int = 32;
+pub const TCP_FASTOPEN_KEY: ::c_int = 33;
+pub const TCP_FASTOPEN_NO_COOKIE: ::c_int = 34;
+pub const TCP_ZEROCOPY_RECEIVE: ::c_int = 35;
+pub const TCP_INQ: ::c_int = 36;
+pub const TCP_CM_INQ: ::c_int = TCP_INQ;
+pub const TCP_TX_DELAY: ::c_int = 37;
```
JohnTitor, line 984: They don't seem to be declared on the emscripten.

zonyitoo: Could you give me a link about what TCP options does emscripten support?

JohnTitor: I usually search items on https://github.com/emscripten-core/emscripten.

zonyitoo: https://github.com/emscripten-core/emscripten/blob/main/system/lib/libc/musl/include/netinet/tcp.h

Yes, you are right.

zonyitoo: https://github.com/ifduyue/musl/blob/master/include/netinet/tcp.h

MUSL upstream have already defined these new options.

JohnTitor: > From the path of the file, it seems that it is copied from MUSL. So can we assume that it will eventually include these new options in the future?

I guess so. But it may cause an unwanted failure to include non-declared items.

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

src/unix/linux_like/mod.rs
```diff
@@ -957,6 +957,42 @@ pub const TCP_WINDOW_CLAMP: ::c_int = 10;
...
+        pub const TCP_REPAIR_WINDOW: ::c_int = 29;
+        pub const TCP_FASTOPEN_CONNECT: ::c_int = 30;
+        pub const TCP_ULP: ::c_int = 31;
+        pub const TCP_MD5SIG_EXT: ::c_int = 32;
+        pub const TCP_FASTOPEN_KEY: ::c_int = 33;
+        pub const TCP_FASTOPEN_NO_COOKIE: ::c_int = 34;
+        pub const TCP_ZEROCOPY_RECEIVE: ::c_int = 35;
+        pub const TCP_INQ: ::c_int = 36;
+        pub const TCP_CM_INQ: ::c_int = TCP_INQ;
+        // pub const TCP_TX_DELAY: ::c_int = 37;
```
JohnTitor, line 993: You don't need it? It'd be great if we could add a comment describing the context.

zonyitoo: I need it. But some of the CI images doesn't have this option. (Not the latest kernel).

PR #2143, opened 2021-04-08
Add RLIM_NLIMITS constant as alias of RLIMIT_NLIMITS on linux_like

coolreader18:
At least on my machine (amd64 Arch) both `RLIMIT_NLIMITS` and `RLIM_NLIMITS` are defined as the same.

### DIFF
--- a/src/unix/linux_like/linux/musl/b32/arm/mod.rs
+++ b/src/unix/linux_like/linux/musl/b32/arm/mod.rs
@@ -171,6 +171,7 @@ pub const RLIMIT_AS: ::c_int = 9;
 pub const RLIMIT_NPROC: ::c_int = 6;
 pub const RLIMIT_MEMLOCK: ::c_int = 8;
 pub const RLIMIT_NLIMITS: ::c_int = 15;
+pub const RLIM_NLIMITS: ::c_int = RLIMIT_NLIMITS;
 
 pub const MADV_SOFT_OFFLINE: ::c_int = 101;
 pub const MCL_CURRENT: ::c_int = 0x0001;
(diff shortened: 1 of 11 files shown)


src/unix/linux_like/linux/musl/b32/arm/mod.rs
```diff
@@ -171,6 +171,7 @@ pub const RLIMIT_AS: ::c_int = 9;
 pub const RLIMIT_NPROC: ::c_int = 6;
 pub const RLIMIT_MEMLOCK: ::c_int = 8;
 pub const RLIMIT_NLIMITS: ::c_int = 15;
+pub const RLIM_NLIMITS: ::c_int = RLIMIT_NLIMITS;
```
JohnTitor, line 174: I think this should be the same on all the musl targets. Why did you separate?

src/unix/linux_like/linux/uclibc/mips/mod.rs
```diff
@@ -49,6 +49,7 @@ pub const RLIMIT_RSS: ::c_int = 7;
 pub const RLIMIT_NPROC: ::c_int = 8;
 pub const RLIMIT_MEMLOCK: ::c_int = 9;
 pub const RLIMIT_NLIMITS: ::c_int = 15;
+pub const RLIM_NLIMITS: ::c_int = RLIMIT_NLIMITS;
```
JohnTitor, line 52: Same here, I don't think this is MIPS only.

coolreader18: I didn't separate it, I just put it next to the existing `RLIMIT_NLIMITS` constant to be safe. At first I was going to put it in `linux_like/mod.rs`, but I could probably compromise and put it in the root of each env submodule.

coolreader18: Oh, I see, on uclibc it's only in mips. Should I extract it out to `uclibc/mod.rs`, `musl/mod.rs`, etc?

JohnTitor: Hm, okay, we should unify it (them) but it isn't necessary on this PR.

PR #2150, opened 2021-04-28
Don't unconditionally link libiconv on macOS.

goffrie:
This adds `#[link(name = "iconv")]` to just the iconv symbols so that the
library is only linked if the symbols are used.

#2037 added a build.rs directive to always link libiconv on Apple platforms,
but that shouldn't be necessary. With this change, we can see using `otool -L`
that a binary that uses an `iconv` symbol links libiconv, but other binaries
don't.

Note that this can only be seen with a rustc prior to nightly-2021-03-09, as
nightly-2021-03-10 includes https://github.com/rust-lang/rust/pull/82731 which
includes #2037, therefore unconditionally linking all Rust binaries to
libiconv.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -11,7 +11,6 @@ fn main() {
     let align_cargo_feature = env::var("CARGO_FEATURE_ALIGN").is_ok();
     let const_extern_fn_cargo_feature = env::var("CARGO_FEATURE_CONST_EXTERN_FN").is_ok();
     let libc_ci = env::var("LIBC_CI").is_ok();
-    let target = env::var("TARGET").unwrap();
 
     if env::var("CARGO_FEATURE_USE_STD").is_ok() {
         println!(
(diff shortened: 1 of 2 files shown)


Amanieu:
If CI is happy then it's fine.

@bors r+

winterqt:
@goffrie It seems that `libiconv` is still unconditionally linked to all Rust binaries (at least on macOS):

```
$ rustc --version
rustc 1.56.0-nightly (1f0a591b3 2021-07-30)
$ cat main.rs
fn main() {}
$ rustc main.rs
$ otool -L main
main:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.100.5)
	/usr/lib/libresolv.9.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
```

The `libc` version in this compilers `std` is `0.2.98`, which includes this patch. Why would this be happening?

PR #2152, opened 2021-04-30
Make netfilter constants available for musl Linux targets

pulsastrix:
The netfilter constants added in #911 and #926 are currently only available for Linux targets using glibc because they weren't available in the musl-sanitized kernel headers at the time these PRs were made. 

With current versions of the sanitized headers, these constants are available, and this PR moves them to `unix/linux_like/linux/mod.rs` so that they can be used on targets using musl libc.

The kernel header version currently set in `ci/install_musl.sh` already supports these constants, but has different values for e.g. `NFT_TABLE_MAXNAMELEN` than the ones that were already defined for glibc. It seems like the maximum name length for various netfilter objects has been changed in the kernel (the respective commits are [1](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/include/uapi/linux/netfilter/nf_tables.h?id=e46abbcc05aa8a16b0e7f5c94e86d11af9aa2770) [2](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/include/uapi/linux/netfilter/nf_tables.h?id=b7263e071aba736cea9e71cdf2e76dfa7aebd039) [3](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/include/uapi/linux/netfilter/nf_tables.h?id=387454901bd62022ac1b04e15bd8d4fcc60bbed4) and [4](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/include/uapi/linux/netfilter/nf_tables.h?id=615095752100748e221028fc96163c2b78185ae4)). To match these values with the ones that were already defined, this PR also updates the used kernel header version in `ci/install_musl.sh`.


JohnTitor:
Any reason to mark this PR as WIP?

pulsastrix:
Not really, I'll mark it as ready.

JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

Issue #2009, opened 2020-12-24
Add `memset_s` and/or `explicit_bzero` for platforms that support it


niluxv:
target platforms: linux, BSD

I would be interested in adding `memset_s` and `explicit_bzero` from `string.h`. These functions allow securely erasing memory. If a PR would be accepted, I can make one.
One question I would have for linux is: What is the minimum version of libc (glibc/musl etc.) that should be supported? Is this documented somewhere?

@rustbot modify labels: +C-API-request and +O-linux and +O-bsd

joshtriplett:
These functions typically have extra annotations to ensure that compiler optimizations will never remove them. Rust doesn't currently have a means of providing those annotations. Thus, I think we shouldn't add these (and give a false impression of their functionality) until the compiler has a means of doing explicit non-removable zeroing.

niluxv:
I do not see how a compiler could optimize out such call. It can't optimize away an entire FFI call, right? The FFI call could have side effects, as far as the compiler knows... FFI is a very strong optimisation barrier without cross-language/linker-plugin LTO.

As long as the `volatile_set_memory` intrinsic does not have a stable counterpart this is actually the only secure and reasonably fast way to zeroize memory on stable.

niacat:
NetBSD has [explicit_memset(3)](http://man.netbsd.org/explicit_memset.3)

PR #2156, opened 2021-05-02
Add explicit/volatile memset/memzero functions

niluxv:
Adds the following functions to platforms that support them:
* `explicit_bzero` (de facto standard): Glibc, MUSL, FreeBSD, DragonFly BSD, OpenBSD
* `explicit_memset`: NetBSD
* `memset_s` (C11 standard): FreeBSD, DragonFly BSD

These functions are useful for zeroing secret memory.

Closes issue #2009

### DIFF
--- a/src/unix/bsd/freebsdlike/mod.rs
+++ b/src/unix/bsd/freebsdlike/mod.rs
@@ -1593,6 +1593,14 @@ extern "C" {
     pub fn iconv_close(cd: iconv_t) -> ::c_int;
 }
 
+extern "C" {
+    // Added in `FreeBSD` 11.0
+    // Added in `DragonFly BSD` 5.4
+    pub fn explicit_bzero(s: *mut ::c_void, len: ::size_t);
+    // ISO/IEC 9899:2011 ("ISO C11") K.3.7.4.1
+    pub fn memset_s(s: *mut ::c_void, smax: ::rsize_t, c: ::c_int, n: ::rsize_t) -> ::errno_t;
+}
+
 #[link(name = "rt")]
 extern "C" {
     pub fn mq_close(mqd: ::mqd_t) -> ::c_int;
--- a/libc-test/semver/dragonfly.txt
+++ b/libc-test/semver/dragonfly.txt
@@ -1174,6 +1174,7 @@ endpwent
 endservent
 endutxent
 exit_status
+explicit_bzero
 faccessat
 fchdir
 fchflags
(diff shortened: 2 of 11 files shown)


niluxv:
As for the failing FreeBSD tests: `rsize_t` is a `size_t` from the C11 standard annex K, and similarly `errno_t` is an `int`. Should I just replace them in the `memset_s` signature?

src/unix/bsd/freebsdlike/mod.rs
```diff
@@ -1593,6 +1593,14 @@ extern "C" {
     pub fn iconv_close(cd: iconv_t) -> ::c_int;
 }
 
+extern "C" {
+    // Added in `FreeBSD` 11.0
+    // Added in `DragonFly BSD` 5.4
+    pub fn explicit_bzero(s: *mut ::c_void, len: ::size_t);
+    // ISO/IEC 9899:2011 ("ISO C11") K.3.7.4.1
+    pub fn memset_s(s: *mut ::c_void, smax: ::rsize_t, c: ::c_int, n: ::rsize_t) -> ::errno_t;
```
devnexen, line 1601: macOs has also memset_s if that helps.

niluxv: Added `memset_s` to macOS and fixed the signature to use standard C types instead of the C11 annex K types that were used.

src/unix/bsd/freebsdlike/mod.rs
```diff
@@ -1593,6 +1593,14 @@ extern "C" {
     pub fn iconv_close(cd: iconv_t) -> ::c_int;
 }
 
+extern "C" {
```
JohnTitor, line 1596: Why did you declare a new `extern` block instead of using the existing one?

niluxv: No particular reason I think. I started with glibc and there were already 3 different `extern "C"` blocks there (all without special `link` attributes), and I didn't know which one I should add this function to, so I added a `extern "C"` block there. And for the other platforms I copied my code from glibc.

JohnTitor: > 
> 
> No particular reason I think. I started with glibc and there were already 3 different `extern "C"` blocks there (all without special `link` attributes), and I didn't know which one I should add this function to, so I added a `extern "C"` block there. And for the other platforms I copied my code from glibc.

I should've overlooked them, let's use one block here.

PR #2161, opened 2021-05-06
Add fn sync to android targets

DoumanAsh:
Note that it has different signature from posix's one

### DIFF
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -2547,6 +2547,7 @@ extern "C" {
         new_value: *const itimerspec,
         old_value: *mut itimerspec,
     ) -> ::c_int;
+    pub fn sync() -> ::c_int;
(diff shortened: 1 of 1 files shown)


DoumanAsh:
It turns out signature of this function is changing from `int sync(void)` into `void sync(void)` depending on android version so might be not feasible to add

JohnTitor:
Looking at https://github.com/aosp-mirror/platform_bionic/commit/79a8f4b2c9b5096c90608018b0d3ba0cb6216ea8, it's a quite old commit (existing since android-8.0.0_r1) so we don't have to count the use of `int sync(void)` here, I guess (Android 7 or older is now "not supported version" IIRC)?

DoumanAsh:
In my case we're using Android 7 and it is not like these versions would disappear all of the sudden.
But I understand if you want to support only latest versions

JohnTitor:
Hm, then it'd be better to declare on the user codebase instead as I'm not sure there's a graceful way to detect the Android toolchain.

PR #2164, opened 2021-05-08
openbsd adding mmap MAP_CONCEAL flag.

devnexen:


semarie:
@devnexen do you have particular use case to `MAP_CONCEAL` ? or do you want to include it just for completeness ?

As OpenBSD could easily break compatibility between versions, I would prefer to add things in rust libc only if effectively used somewhere. Else it will break some days as rust libc doesn't want ot break compatibility.

From PR point of vue, no problem.

devnexen:
It is used outside of OpenBSD codebase especially where FreeBSD's MAP_NOCORE is, from security standpoint, it has its use.

semarie:
thanks for the information. glad to see it will be used

PR #2166, opened 2021-05-09
freebsd update adding cpu affinity api subset

devnexen:


src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -121,6 +124,10 @@ s! {
         pub pve_fsid: u32,
         pub pve_path: *mut ::c_char,
     }
+
+    pub struct cpuset_t {
+        __bits: [::c_long; (256 + _BITSET_BITS - 1) / _BITSET_BITS],
```
JohnTitor, line 129: Is it intentional that the field is hidden?

devnexen: code consumers are not really supposed to deal directly with thus the various CPU_* "macros". But it that s too much of controversy it can be made public.

JohnTitor (approved):

PR #2168, opened 2021-05-11
Add IP_BOUND_IF for Apple

pinkisemils:
To allow for binding a socket to an interface or querying a socket to see which interface it's bound to, `IP_BOUND_IF` must be used - I've added it to `libc`. 

I've added it as an apple-wide constant (both for macOS and iOS), but I'm not sure if it's the same value on iOS.


JohnTitor:
Thanks!
r? @JohnTitor @bors r+

JohnTitor (approved):

PR #2170, opened 2021-05-12
Fix type for Android st_nlink

tweksteen:


Amanieu:
@JohnTitor This changes `st_nlink` from `i32` to `u32`. Is this an acceptable breaking change?

Amanieu:
Actually these are already unsigned, so this isn't a breaking change.

@bors r+

PR #1638, opened 2020-01-06
Add missing AArch64 HWCAP_* and HWCAP2_* consts

linkmauve:
Also move `getauxval()` where it should be.

PR #2172, opened 2021-05-13
musl: add missing AArch64 HWCAP_* consts

### LINKED ISSUE #2171: musl: HWCAP_* consts missing on aarch64 targets
It looks like #1638 added these constants, but for glibc only. #2109 added a [`TODO-linux.txt`](https://github.com/rust-lang/libc/blob/ed1399a/libc-test/semver/TODO-linux.txt) which made a note of this.

I just received an in-the-wild report about this (https://github.com/RustCrypto/hashes/issues/273) and noticed there wasn't a specific tracking issue for it, so I thought I'd open one.

At first glance it seems like this could be resolved by a PR similar to #1638, but adding the constants to `linux/musl/b64/aarch64/mod.rs`.

Is that the case? If so, I could try to open a PR to add these constants. Edit: opened #2172.

tarcieri:
Closes #2171.

I effectively just copied and pasted these from #1638 after confirming that they are identical to musl's [aarch64/bits/hwcap.h](https://github.com/bminor/musl/blob/4433115/arch/aarch64/bits/hwcap.h) and that musl's `getauxval()` also takes a `c_ulong`.

It seems like if these are added it would address some of [`TODO-linux.txt`](https://github.com/rust-lang/libc/blob/ed1399a/libc-test/semver/TODO-linux.txt)?


Amanieu:
I think that does address the TODO-linux.txt, you can remove them from this file.

tarcieri:
@Amanieu done

tarcieri:
Build failure looks unrelated

joshtriplett:
Let's give it another try:
@bors r+

PR #2173, opened 2021-05-13
freebsd update adding `elf_aux_info` from 12.x release.

devnexen:


JohnTitor:
@devnexen Could you drop 8ece773?

PR #2176, opened 2021-05-15
openbsd adding few more netdb constants

devnexen:


semarie (changes requested):
please include `libc-test/semver/openbsd.txt` additions, else it is fine.

semarie (approved):
LGTM thanks

PR #2179, opened 2021-05-16
apple adding thread policy api

devnexen:


libc-test/semver/apple.txt
```diff
@@ -585,6 +586,7 @@ KERN_RAGE_THREAD
 KERN_SAFEBOOT
 KERN_SAVED_IDS
 KERN_SECURELVL
+KERN_SUCCESS
```
JohnTitor, line 589: Looks like we don't cover all the items added by this PR.

JohnTitor (approved):

PR #2182, opened 2021-05-16
freebsd: add realhostname api from libutil

devnexen:


src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1612,6 +1618,13 @@ extern "C" {
         string: *const ::c_char,
         attrnamespace: *mut ::c_int,
     ) -> ::c_int;
+    pub fn realhostname(host: *mut ::c_char, hsize: ::size_t, ip: *const ::in_addr) -> ::c_int;
+    pub fn realhostname_sa(
```
JohnTitor, line 1622: We should also add them to the semver list.

JohnTitor (approved):

PR #2188, opened 2021-05-22
openbsd update to latest release for few constants.

devnexen:
Those values might be correct in 6.8 but not in 6.9. Handling release N and release N-1 would be too complicated I suggest we prioritize the lastest.

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/freebsd13/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/freebsd13/mod.rs
@@ -231,6 +231,7 @@ extern "C" {
     pub fn fdatasync(fd: ::c_int) -> ::c_int;
 
     pub fn getrandom(buf: *mut ::c_void, buflen: ::size_t, flags: ::c_uint) -> ::ssize_t;
+    pub fn getentropy(buf: *mut ::c_void, buflen: ::size_t) -> ::c_int;
     pub fn elf_aux_info(aux: ::c_int, buf: *mut ::c_void, buflen: ::c_int) -> ::c_int;
 }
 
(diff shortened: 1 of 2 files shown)


semarie:
@JohnTitor is it possible to just deprecated these constants (and remove them "one day") ?

semarie:
else, values are right and it LGTM.

JohnTitor:
Yup, I'd prefer deprecating as per https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md#breaking-change-policy.

src/unix/bsd/freebsdlike/freebsd/freebsd13/mod.rs
```diff
@@ -231,6 +231,7 @@ extern "C" {
     pub fn fdatasync(fd: ::c_int) -> ::c_int;
 
     pub fn getrandom(buf: *mut ::c_void, buflen: ::size_t, flags: ::c_uint) -> ::ssize_t;
+    pub fn getentropy(buf: *mut ::c_void, buflen: ::size_t) -> ::c_int;
```
semarie, line 234: change in wrong PR ?

src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -936,6 +936,10 @@ pub const NET_RT_STATS: ::c_int = 4;
 pub const NET_RT_TABLE: ::c_int = 5;
 pub const NET_RT_IFNAMES: ::c_int = 6;
 #[doc(hidden)]
+#[deprecated(
+    since = "0.2.95",
+    note = "Possbily increasing over the releases and might not be so used in the field"
```
semarie, line 941: small typo in "Possibly" (and same in other deprecated change)

semarie (changes requested):

PR #2169, opened 2021-05-12
Bump version to `0.2.95`

pinkisemils:
Hiya, I'm eager to use the recently added constant in #2168. Thus, here's a PR to bump the `libc` version to allow a new version to be released.

PR #2192, opened 2021-05-23
netbsd mapping auxiliary vectors type

devnexen:


devnexen:
most likely last PR before 0.2.95, there is #2169 awaiting :-)

JohnTitor (approved):

PR #2199, opened 2021-05-26
openbsd/netbsd siginfo_t adding si_addr field accessor.

devnexen:


Amanieu:
What is the point of this function? The `si_addr` field is already available.

devnexen:
There is software relying on this :-) this is how I spotted this

Amanieu:
There is a precedent in FreeBSD which is already doing this.

@bors r+

PR #2201, opened 2021-05-27
Signal controlling constants for windows

vorner:
This is replacement of #1626, because I have lost the repository with that branch in the meantime. The content is the same (unless I've made some mistake copying it over), just rebased onto current master.


JohnTitor (approved):

JohnTitor:
Thanks!
r? @JohnTitor @bors r+

JohnTitor:
Oh wait, we should add them to the semver list: https://github.com/rust-lang/libc/blob/master/libc-test/semver/windows.txt
@bors r-

PR #2207, opened 2021-05-31
bsd adding setproctitle (setproctitle_fast for FreeBSD too).

### LINKED ISSUE #2197: Add setproctitle
FreeBSD:

```     
#include <sys/types.h>
#include <unistd.h>

void setproctitle(const	char *fmt, ...);

void setproctitle_fast(const char *fmt,	...);
```

Certain BSD's have setproctitle to change the process title, and other unices have different ways of changing the process title (as shown in ps and such).

For FreeBSD see: https://www.freebsd.org/cgi/man.cgi?query=setproctitle&apropos=0&sektion=0&manpath=FreeBSD+13.0-RELEASE+and+Ports&arch=default&format=html

I'm not sure what's needed to get support for setproctitle, at least on platforms where it's supported. I'm not proficient at interacting with C from Rust but hopefully I can still help.

devnexen:
Closes #2197

### DIFF
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -2113,6 +2113,8 @@ extern "C" {
 
     // Added in `NetBSD` 7.0
     pub fn explicit_memset(b: *mut ::c_void, c: ::c_int, len: ::size_t);
+
+    pub fn setproctitle(fmt: *const ::c_char, ...);
 }
 
 #[link(name = "util")]
--- a/src/unix/bsd/netbsdlike/openbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/openbsd/mod.rs
@@ -1566,6 +1566,8 @@ extern "C" {
 
     // Added in `OpenBSD` 5.5
     pub fn explicit_bzero(s: *mut ::c_void, len: ::size_t);
+
+    pub fn setproctitle(fmt: *const ::c_char, ...);
 }
 
 cfg_if! {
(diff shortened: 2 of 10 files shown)


JohnTitor:
cc @semarie

src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -1566,6 +1566,7 @@ extern "C" {
 
     // Added in `OpenBSD` 5.5
     pub fn explicit_bzero(s: *mut ::c_void, len: ::size_t);
+    pub fn setproctitle(fmt: *const ::c_char, ...);
```
semarie, line 1569: mostly cosmetic remark: the position of the function addition is a bit misleading as it seems it is part of a block "Added in `OpenBSD` 5.5" (which isn't right for setproctitle).

starting a new block with a newline might be preferable

src/unix/bsd/netbsdlike/netbsd/mod.rs
```diff
@@ -2113,6 +2113,7 @@ extern "C" {
 
     // Added in `NetBSD` 7.0
     pub fn explicit_memset(b: *mut ::c_void, c: ::c_int, len: ::size_t);
+    pub fn setproctitle(fmt: *const ::c_char, ...);
```
semarie, line 2116: same remark than for OpenBSD. the function isn't "Added in `NetBSD` 7.0"

semarie (commented):
globally fine with me, just two small nits

JohnTitor (approved):

JohnTitor:
Thanks both, @bors r+

PR #2216, opened 2021-06-03
Add gethostuuid and uuid_t on macOS

Thomasdezeeuw:
This adds the `gethostuuid` function and the `uuid_t` for macOS.

I'm not sure the function signature is correct. It's defined as the following in C:

```C
int gethostuuid(uuid_t id, const struct timespec *wait);

typedef unsigned char   __darwin_uuid_t[16];
typedef __darwin_uuid_t uuid_t;
```

Because C doesn't really has arrays, the accept parameter can't be `uuid_t` in Rust, as Rust does have arrays and thus expect an array to be passed by value, where the C function expects a pointer.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -361,6 +361,8 @@ fn test_apple(target: &str) {
     cfg.skip_roundtrip(move |s| match s {
         // FIXME: this type has the wrong ABI
         "max_align_t" if i686 => true,
+        // Can't return an array from a C function.
+        "uuid_t" => true,
         _ => false,
     });
     cfg.generate("../src/lib.rs", "main.rs");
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
Also this is not available on iOS I think:

```C
#if defined(__IPHONE_OS_VERSION_MIN_REQUIRED) && (__IPHONE_OS_VERSION_MIN_REQUIRED < __IPHONE_7_0)
int gethostuuid(uuid_t, const struct timespec *) __OSX_AVAILABLE_BUT_DEPRECATED_MSG(__MAC_NA, __MAC_NA, __IPHONE_2_0, __IPHONE_5_0, "gethostuuid() is no longer supported");
#else
int gethostuuid(uuid_t, const struct timespec *) __OSX_AVAILABLE_STARTING(__MAC_10_5, __IPHONE_NA);
#endif
```

PR #2218, opened 2021-06-05
netbsd/freebsd adding accept_filter_arg as SO_ACCEPTFILTER only makes…

devnexen:
… sense with


Amanieu:
CI seems to be failing. I think our tests can't handle the `-` in the array size computation.

JohnTitor (approved):

PR #2224, opened 2021-06-09
Add AF_VSOCK to uclibc mips

kolapapa:
https://github.com/kraj/uclibc-ng/blob/master/libc/sysdeps/linux/common/bits/socket.h#L131


JohnTitor:
It shouldn't be MIPS-specific, right? Then it makes more sense to place on uclibc/mod.rs.

JohnTitor (approved):

JohnTitor:
Thanks! r? @JohnTitor @bors r+

PR #2225, opened 2021-06-09
Update FIXME for WASI

JohnTitor:
r? @ghost

### DIFF
--- a/.github/workflows/bors.yml
+++ b/.github/workflows/bors.yml
@@ -127,7 +127,10 @@ jobs:
           powerpc64le-unknown-linux-gnu,
           s390x-unknown-linux-gnu,
           riscv64gc-unknown-linux-gnu,
-          # FIXME: Figure out why this is disabled.
+          # FIXME: A recent nightly causes a linker failure:
+          # https://github.com/rust-lang/rust/issues/76679
+          # See this comment for more details:
+          # https://github.com/rust-lang/libc/pull/2225#issuecomment-880696737
           #wasm32-wasi,
           sparc64-unknown-linux-gnu,
           wasm32-unknown-emscripten,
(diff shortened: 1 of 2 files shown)


JohnTitor:
Now we run into this:
```
Error: failed to run main module `/checkout/target/wasm32-wasi/debug/deps/main-28cbf70091fdc75c.wasm`

Caused by:
    0: failed to read input file
    1: No such file or directory (os error 2)
```

This happens on every test, @sunfishcode would you have time to check it? Seems it's related to https://github.com/rust-lang/libc/pull/1437#issuecomment-515681183 and I guess I forgot to set a correct flag or something else therefore it cannot find wasm file?

JohnTitor:
Some notes:
- `RUSTFLAGS='-C link-args=-Wl,-v'` fixes "Bad relocation type" failure but it seems that it makes the build not produce any wasm files.
  - Actually, the command works fine on my end (ArchLinux) with the below env vars:

```dockerfile
ENV CARGO_TARGET_WASM32_WASI_RUNNER=wasmtime \
  CARGO_TARGET_WASM32_WASI_LINKER=/wasi-libc/bin/clang \
  CC_wasm32_wasi=/wasi-libc/bin/clang \
  PATH=$PATH:/rust/bin \
  RUSTFLAGS='-Ctarget-feature=-crt-static'
```

The problem is that without the above link-args flag, the build won't work on Ubuntu, i.e., default llvm, clang, lld packages.
I've tried xxx-12 packages but it causes "`cc` not found" failure.
Another way would be to use https://github.com/WebAssembly/wasi-sdk's master.

JohnTitor:
So, I couldn't find the best solution here, https://github.com/rust-lang/rust/issues/76679 is quite painful... I'm going to keep disabling WASI CI until the issue is fixed.
@bors r+

PR #2228, opened 2021-06-10
Add `mallinfo2` support

jonas-schievink:
This function was added in glibc 2.33 and fixes a shortcoming of the mallinfo API: it was unable to handle memory usage of more than 2 GB due to its use of `int` as the field types. This was fixed by duplicating the API and changing them to `size_t`.


JohnTitor (approved):

JohnTitor:
Looks good! @bors r+

PR #2236, opened 2021-06-12
solaris based systems update

devnexen:
basis to do cpu affinity with some illumos specifics.

### DIFF
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -25,6 +25,8 @@ pub type wchar_t = ::c_int;
 pub type nfds_t = ::c_ulong;
 pub type projid_t = ::c_int;
 pub type zoneid_t = ::c_int;
+pub type psetid_t = ::c_int;
+pub type processorid_t = ::c_int;
 
 pub type suseconds_t = ::c_long;
 pub type off_t = ::c_long;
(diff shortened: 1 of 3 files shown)


src/unix/solarish/mod.rs
```diff
@@ -2227,6 +2241,8 @@ extern "C" {
     pub fn labs(i: ::c_long) -> ::c_long;
     pub fn rand() -> ::c_int;
     pub fn srand(seed: ::c_uint);
+    pub fn getentropy(buf: *mut ::c_void, buflen: ::size_t) -> ::c_int;
+    pub fn getrandom(bbuf: *mut ::c_void, buflen: ::size_t, flags: ::c_uint);
```
JohnTitor, line 2245: Shouldn't it return `ssize_t`?

devnexen: oh right I just forgot the return type

RalfJung: Is using the same type for Illumos and Solaris correct here? `rand`/`getrandom` used to use different return types for these two OSes (see e.g. https://github.com/rust-random/rand/pull/730), and glibc also mentions Solaris having a non-standard return type ([here](https://www.gnu.org/software/gnulib/manual/html_node/getrandom.html)).

PR #2237, opened 2021-06-13
Add more MNT_ flags, especially on FreeBSD

asomers:


src/unix/bsd/apple/mod.rs
```diff
@@ -3358,18 +3358,10 @@ pub const DLT_LOOP: ::c_uint = 108;
 pub const BPF_ALIGNMENT: ::c_int = 4;
 
 // sys/mount.h
-pub const MNT_RDONLY: ::c_int = 0x00000001;
```
devnexen, line 3361: just curious any reason removing them ?

asomers: I'm not removing them altogether, just moving them into bsd/mod.rs .  Those constants have the same value on all of the BSDs.

devnexen: ah right

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

JohnTitor:
Ugh: `qemu-system-sparc64: terminating on signal 15 from pid 1533 (timeout)`
@bors retry

PR #2238, opened 2021-06-13
solaris adding few cpu affinity related functions

devnexen:


src/unix/solarish/mod.rs
```diff
@@ -2642,6 +2642,13 @@ extern "C" {
     pub fn pset_list(pset: *mut psetid_t, numpsets: *mut ::c_uint) -> ::c_int;
     pub fn pset_setattr(pset: psetid_t, attr: ::c_uint) -> ::c_int;
     pub fn pset_getattr(pset: psetid_t, attr: *mut ::c_uint) -> ::c_int;
+    pub fn processor_bind(
+        idtype: ::idtype_t,
+        id: ::id_t,
+        new_binding: ::processorid_t,
+        old_binding: *mut *mut processorid_t,
```
JohnTitor, line 2649:
```suggestion
        old_binding: *mut processorid_t,
```

I think it should be: https://docs.oracle.com/cd/E19109-01/tsolaris8/817-0881/6mglbsoma/index.html and https://github.com/illumos/illumos-gate/blob/9ecd05bdc59e4a1091c51ce68cce2028d5ba6fd1/usr/src/uts/common/sys/processor.h#L133

JohnTitor (approved):

PR #2189, opened 2021-05-22
openbsd update map sigcontext for amd64

devnexen:

PR #2242, opened 2021-06-15
OpenBSD: add signal-stack machine context struct on x86_64.

cfallin:
The `ucontext_t` struct on OpenBSD (which is actually a typedef to
`struct sigcontext`) represents the on-stack saved state that is
available for a signal handler on OpenBSD to read and/or modify. Its
definition comes from <machine/signal.h>, and was copied from OpenBSD
6.9, which is the current version. Note that OpenBSD does not guarantee
ABI compatibility across versions, so this would have to be updated if
the frame format ever does change; however, it is very useful to have
(it is not possible to know the faulting PC in a signal handler
otherwise), so I suspect including this definition is worthwhile.

### DIFF
--- a/src/unix/bsd/netbsdlike/openbsd/x86_64.rs
+++ b/src/unix/bsd/netbsdlike/openbsd/x86_64.rs
@@ -4,6 +4,44 @@ pub type c_long = i64;
 pub type c_ulong = u64;
 pub type c_char = i8;
 
+s! {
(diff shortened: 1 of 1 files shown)


semarie:
there is already on-going work in #2189 to add `sigcontext_t` definition. thanks.

cfallin:
Ah, that's fantastic, thanks! I hadn't seen that PR, sorry. Will close this PR now.

PR #2246, opened 2021-06-16
pthread_spinlock api remaining bsd systems

devnexen:


src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -383,6 +385,13 @@ s! {
         pub dlpi_phdr: *const Elf_Phdr,
         pub dlpi_phnum: Elf_Half,
     }
+
+    // pthread.h
+
+    pub struct __c_anonymous_pthread_spinlock {
```
semarie, line 391: `pthread_spinlock` struct is internal and isn't in public header on purpose. the visible part is only a pointer, nothing else.

it is the same kind of type as `pthread_t` which is defined as `pub type pthread_t = ::uintptr_t` in Rust.

src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -1530,6 +1539,12 @@ extern "C" {
     pub fn pthread_main_np() -> ::c_int;
     pub fn pthread_set_name_np(tid: ::pthread_t, name: *const ::c_char);
     pub fn pthread_stackseg_np(thread: ::pthread_t, sinfo: *mut ::stack_t) -> ::c_int;
+    pub fn pthread_spin_init(lock: *mut pthread_spinlock_t, pshared: ::c_int) -> ::c_int;
```
semarie, line 1542: if I don't mess myself, pthread_spin_ functions could be commun with at least NetBSD.

semarie (changes requested):

src/unix/bsd/freebsdlike/dragonfly/mod.rs
```diff
@@ -193,6 +195,10 @@ s! {
     pub struct cpumask_t {
         ary: [u64; 4],
     }
+
+    pub struct __c_anonymous_pthread_spinlock_s {
```
semarie, line 199: it isn't exactly my area but I think if the OS define a struct as opaque, it should be kept opaque.

`pthread_spinlock_t` is opaque in DragonflyBSD too: [sys/sys/_pthreadtypes.h](https://github.com/DragonFlyBSD/DragonFlyBSD/blob/c163a4d7ee9c6857ee4e04a3a2cbb50c3de29da1/sys/sys/_pthreadtypes.h)

libc-test/build.rs
```diff
@@ -381,6 +381,7 @@ fn test_openbsd(target: &str) {
         "limits.h",
         "link.h",
         "locale.h",
+        "machine/spinlock.h",
```
semarie, line 384: this definition shouldn't be need now that you don't use `_atomic_lock_t` type.

the tests passes on OpenBSD without it.

semarie (changes requested):

semarie:
it mess to me you reintroduced `__c_anonymous_pthread_spinlock` on OpenBSD while removing it from DragonflyBSD...

devnexen:
ah yes wrong rebasing

semarie:
rebasing still a bit wrong :
- `if ty.starts_with("__c_anonymous_") {` isn't need anymore in `test_openbsd()` function
- pthread_spin_ functions could be merged with NetBSD inside `src/unix/bsd/netbsdlike/mod.rs`

devnexen:
true. Merci beaucoup for your review ;-)

semarie (approved):
fine with me. thanks

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #2249, opened 2021-06-18
darwin add few more debug malloc zone api

devnexen:


src/unix/bsd/apple/mod.rs
```diff
@@ -77,6 +77,8 @@ pub type thread_latency_qos_policy_t = *mut thread_latency_qos_policy;
 pub type thread_throughput_qos_policy_data_t = thread_throughput_qos_policy;
 pub type thread_throughput_qos_policy_t = *mut thread_throughput_qos_policy;
 
+pub type malloc_zone_t = [::uintptr_t; 18]; // FIXME: keeping 'private' for now
```
Amanieu, line 80: Could you hide the array behind a struct with a private field? That way changing the representation in the future won't be a breaking change.

PR #2252, opened 2021-06-20
linux adding semid_ds struct and seminfo incidently.

### LINKED ISSUE #2002: Provide definition of struct semid_ds for Linux
Target triplet: x86_64-unknown-linux-gnu

Documentation for `struct semid_ds`: https://man7.org/linux/man-pages/man2/semctl.2.html

```c
       The semid_ds data structure is defined in <sys/sem.h> as follows:

           struct semid_ds {
               struct ipc_perm sem_perm;  /* Ownership and permissions */
               time_t          sem_otime; /* Last semop time */
               time_t          sem_ctime; /* Creation time/time of last
                                             modification via semctl() */
               unsigned long   sem_nsems; /* No. of semaphores in set */
           };
```

Note that libc already has definition of `semctl` (function that needs `struct semid_ds`) for Linux: https://github.com/rust-lang/libc/blob/0.2.81/src/unix/linux_like/linux/mod.rs#L2777-L2782

libc also has definition of `struct semid_ds` for MacOS (added via #1159): https://github.com/rust-lang/libc/blob/0.2.81/src/unix/bsd/apple/mod.rs#L553-L564

devnexen:
close #2002

### DIFF
--- a/libc-test/semver/linux-gnu.txt
+++ b/libc-test/semver/linux-gnu.txt
@@ -566,6 +566,8 @@ pthread_setname_np
 pututxline
 pwritev2
 qsort_r
+semid_ds
+seminfo
 setutxent
 setxattr
 sgetspent_r
--- a/src/unix/linux_like/linux/gnu/b32/mod.rs
+++ b/src/unix/linux_like/linux/gnu/b32/mod.rs
@@ -14,6 +14,7 @@ pub type __u64 = ::c_ulonglong;
 pub type __fsword_t = i32;
 pub type fsblkcnt64_t = u64;
 pub type fsfilcnt64_t = u64;
+pub type __syscall_ulong_t = ::c_ulong;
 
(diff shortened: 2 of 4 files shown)


JohnTitor (approved):

JohnTitor:
Thanks!
r? @JohnTitor @bors r+

libc-test/semver/linux.txt
```diff
@@ -2971,6 +2971,8 @@ sem_unlink
 sembuf
 semctl
 semget
+semid_ds
+seminfo
```
JohnTitor, line 2975: Oh, should be in linux-gnu.txt instead.

JohnTitor:
It fails on style check.

JohnTitor:
@bors delegate+
In case this fails, you can `@bors try` to check the build but not merge.

PR #2263, opened 2021-06-29
netbsd: sched_param api addition

devnexen:


JohnTitor (approved):

JohnTitor:
Looks good to me, @semarie could you take a look?

semarie:
OpenBSD doesn't have this interface.

technically the structure exists (but nothing use it) and the functions are under `#if 0` in the `sched.h` header. so it would be better to just not declare it too.

JohnTitor:
Thanks for checking!
@bors r+

PR #2264, opened 2021-07-01
Define `S_I{R,W,X}{USR,GRP,OTH}` and `S_IS{VTX,UID,GID}` for wasi.

sunfishcode:
WASI doesn't have full permissions setting, but it's gaining some
support, so define the POSIX names that wasi-libc defines, and let
WASI libc and WASI sort out what to do with them.


JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

PR #2271, opened 2021-07-05
linux timer api remove unnecessary double linkage to librt.

devnexen:


JohnTitor:
This won't link librt on musl, is it correct behavior?

12101111:
> This won't link librt on musl, is it correct behavior? (Moreover, any build with the `std` also won't link librt but it seems fine?:
> 
> https://github.com/rust-lang/libc/blob/90aa0e32eca33ccec050c034c438b1b703f16449/src/unix/mod.rs#L302-L304
> )

librt.a of musl is empty file:

```
> cat /usr/lib/librt.a
!<arch>
```

And build of `libc` with `std` feature will pull `std` and a prebuilt `libc` with `rustc-dep-of-std` feature

https://github.com/rust-lang/rust/blob/72b0c7dfe9bcb55096d8b81a67e683730f1e3c15/library/std/Cargo.toml#L19

JohnTitor (approved):
Thanks for clarifying! Then it should be fine. There're still some uses of `name = "rt"`, I think they all also can be removed, right?
Anyway, this is good to go as-is, let's r+ :)

PR #2273, opened 2021-07-06
Use inlined script to upload docs to gh-pages

JohnTitor:
...because the action we're using is outdated. Confirmed that it worked on my fork: https://github.com/JohnTitor/libc/actions/runs/1002739375
r? @ghost


JohnTitor:
Self r+'ing as this is just CI cleanup.
@bors r+

PR #2280, opened 2021-07-09
proposes to disable (at least for now) few flags for arm64 as they ar…

devnexen:
…e missing in raspberry


devnexen:
I just propose here, if not merged as it s surely available in other arm64 boards, this is fine by me.

Amanieu:
I don't think this is necessary? It doesn't hurt to have some extra constants defined, it's just that trying to use them will return a runtime error since the libc/kernel doesn't support them.

The libc tests will fail due to the missing constants in the C headers, but we only guarantee that these tests pass on our CI platforms.

devnexen:
understood :)

PR #2283, opened 2021-07-12
apple add mach_vm_map, mmap does not provide custom alignment so

devnexen:
mach_vm_map is only way.

### DIFF
--- a/libc-test/semver/apple.txt
+++ b/libc-test/semver/apple.txt
@@ -729,6 +729,7 @@ MAXTHREADNAMESIZE
 MCL_CURRENT
 MCL_FUTURE
 MDMBUF
+MEMORY_OBJECT_NULL
 MH_MAGIC
 MH_MAGIC_64
 MINCORE_INCORE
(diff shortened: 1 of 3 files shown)


JohnTitor:
We've deprecated mach-related items since #1364 and a change like this should go to the `mach` crate, I think.

devnexen:
I thought that but the crate had not been updated since 2019

JohnTitor (approved):
Uhm, that's unfortunate, we definitely should consider not suggesting that crate on the deprecation note and open an issue to discuss further.
Anyway, this PR should be fine then, thanks for clarifying!

libc-test/semver/apple.txt
```diff
@@ -1700,6 +1700,7 @@ mach_port_t
 mach_thread_self
 mach_timebase_info
 mach_timebase_info_data_t
+mach_vm_map
```
JohnTitor, line 1703: Ah, the list seems incomplete.

PR #2195, opened 2021-05-25
Add SIGRTMIN/SIGRTMAX on Linux

clarfonthey:
Closes #1883

PR #2304, opened 2021-07-28
Add SIGRTMAX and SIGRTMIN on linux-like systems

### LINKED ISSUE #1883: libc expose SIGRTMIN and SIGRTMAX
Exposing SIGRTMIN and SIGRTMAX might be problematic as they dynamically change and will therefore be hard to encode as constants. Perhaps exposing them in the form of functions is the best solution? I have very little experience porting from C to Rust so I am not sure what the best way of doing that would be. If someone gave me a point in the right direction then I would be willing to try to tackle this.

Target triplet: x86_64-linux-gnu

Man page for signals:
https://www.man7.org/linux/man-pages/man7/signal.7.html

Other crates looking for feature:
https://github.com/nix-rust/nix/issues/1143

mkroening:
Fixes #1883. This is based on https://github.com/rust-lang/libc/pull/2195, which has not seen activity in a while.

The original issue was resolved by making `__libc_current_sigrtmax` and `__libc_current_sigrtmax` `pub`.

I also added a test similar to `errqueue`'s, where the actual value of the macro is compared to `libc`'s implementation.

Closes #2195


src/unix/linux_like/mod.rs
```diff
@@ -1362,6 +1362,16 @@ cfg_if! {
     }
 }
 
+safe_f! {
```
JohnTitor, line 1365: We have another `safe_f!` here, the below should be placed there. And we may want to make them `const fn`s.

src/unix/linux_like/mod.rs
```diff
@@ -1480,6 +1490,9 @@ safe_f! {
 }
 
 extern "C" {
+    pub fn __libc_current_sigrtmax() -> ::c_int;
+    pub fn __libc_current_sigrtmin() -> ::c_int;
```
JohnTitor, line 1494: I think it's fine to expose but I'd like to make it `#[doc(hidden)]`.

JohnTitor (changes requested):

src/unix/linux_like/mod.rs
mkroening: Done. :+1:

src/unix/linux_like/mod.rs
mkroening: I moved it to the other `safe_f!`, did I get the `const`ness right?

JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

mkroening:
[Looks](https://github.com/rust-lang/libc/runs/3196028707) like the `const`ness is not allowed here, since we call `__libc_current_sigrtmax`.

Should I remove the `const`ness again?

JohnTitor:
Ah yes! I forgot that `__libc_*`s weren't `const`s.

mkroening:
Done. :+1:

PR #2306, opened 2021-07-29
Upgrade `semverver` to 0.1.47

JohnTitor:
Finally we released it on crates.io 🎉
r? @ghost


JohnTitor:
Hm, why doesn't `continue-on-error: true` work here? Investigating...

JohnTitor:
Oh well, actually there was another failure:
```
 qemu-system-sparc64: terminating on signal 15 from pid 3193 (timeout)
```

@bors retry

PR #2307, opened 2021-07-29
Remove wasm test workaround

JohnTitor:
r? @ghost


JohnTitor:
Seems it now only runs js tests, it means we can remove the workaround :)

PR #2310, opened 2021-07-31
Support for the ESP-IDF framework (Xtensa and RiscV arch)

ivmarkov:
Dear all,

This PR is implementing support for the [ESP-IDF](https://github.com/espressif/esp-idf) newlib-based framework, which is the open source SDK provided by Espressif for their MCU family (esp32, esp32s2, esp32c3 and all other forthcoming ones).

Note that this is the second PR on that topic. Approx. an year ago, @reitermarkus contributed an [initial set of changes](https://github.com/rust-lang/libc/commit/0dec549f8d649d661c4992cf59a2b6283eef7569#diff-6c07c29bb7b11b27a308055cca03f299266fd8f05af5a65208b64ff715359c89) which are merged already.

Note also that this PR has a [sibling PR against Rust's libStd](https://github.com/rust-lang/rust/pull/87666) which enables **full STD support for the Espressif chipsets** (that is, modulo process support as we are obviously talking about non-kernel bare metal platform here).

A short overview of the changes:
* The original contribution of @reitermarkus is renamed from `xtensa.rs` to `espidf.rs` and all branching through this patch is no longer done based on `target_arch = "xtensa"`, but based on `target_os = "espidf"` (this `target_os` value is to be used by the upcoming Rust targets for the ESP-IDF framework). The primary reason for this change is that branching for ESP-IDF based only on the architecture is no longer valid: the newer Espressif chips (esp32c3 and other upcoming ones) are based on the RISCV32IM(A)C architecture, so this patch now supports both Xtensa and RiscV32. Moreover, I would expect that - given the popularity of the riscv ISA - there will be *other* ports of newlib to riscv which will surely have the layout and sizes of the structures etc. different from the ESP-IDF framework.  
*  The `pthread` structures had sizes which did not match what is used in the ESP-IDF. Ditto for the various `*_INITIALIZER` constants, which do not use 0x00, but 0xff sequences there
* The BSD socket API on ESP-IDF is prefixed with `lwip_`. Rather than doing heavyweight proxying in Rust libStd, it is best to just address this here with a custom `link_name`, just like it had been done for MacOS and other systems.
* Similar to the `l4re` case, the libc crate should NOT issue a link command  to the CRT lib for the ESP-IDF, due to the way the SDK is linked
* Various other small fixes: primarily - declaration of standard APIs that are available in the ESP-IDF SDK

I should also admit that the patch has one little ESP-IDF-specific cheat:
* Since the `pthread` support in ESP-IDF is still lacking RW-locks, I've implemented it as a temporary workaround by simply remapping (with `"link_name"`) the `pthread_rwlock_*` symbols to their equivalent `pthread_mutex_*` symbols. While this implementation is suboptimal, it reduces significantly the PR changeset surface of the other PR which is against Rust libStd.

### DIFF
--- a/src/unix/newlib/mod.rs
+++ b/src/unix/newlib/mod.rs
@@ -33,16 +33,14 @@ s! {
         pub ai_protocol: ::c_int,
         pub ai_addrlen: socklen_t,
 
-        #[cfg(not(all(libc_cfg_target_vendor, target_arch = "powerpc",
-              target_vendor = "nintendo")))]
-        #[cfg(target_arch = "xtensa")]
+        #[cfg(target_os = "espidf")]
         pub ai_addr: *mut sockaddr,
 
         pub ai_canonname: *mut ::c_char,
 
-        #[cfg(not(all(libc_cfg_target_vendor, target_arch = "powerpc",
-              target_vendor = "nintendo")))]
-        #[cfg(not(target_arch = "xtensa"))]
+        #[cfg(not(any(
+            target_os = "espidf",
+            all(libc_cfg_target_vendor, target_arch = "powerpc", target_vendor = "nintendo"))))]
         pub ai_addr: *mut sockaddr,
 
         pub ai_next: *mut addrinfo,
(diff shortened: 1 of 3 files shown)


Amanieu:
> Since the `pthread` support in ESP-IDF is still lacking RW-locks, I've implemented it as a temporary workaround by simply remapping (with `"link_name"`) the `pthread_rwlock_*` symbols to their equivalent `pthread_mutex_*` symbols. While this implementation is suboptimal, it reduces significantly the PR changeset surface of the other PR which is against Rust libStd.

This is not OK. `pthread_rwlock` is very different from `pthread_mutex` and libc should not expose one as the other.

Note that rwlocks can be implemented [using mutexes and condition variables](https://en.wikipedia.org/wiki/Readers%E2%80%93writer_lock#Using_a_condition_variable_and_a_mutex), so perhaps ESP-IDF could add pthread_rwlock support in the future.

In the meantime, `pthread_rwlock` should not be exposed by libc.

The rest of the changes look fine.

ivmarkov:
> Note that rwlocks can be implemented [using mutexes and condition variables](https://en.wikipedia.org/wiki/Readers%E2%80%93writer_lock#Using_a_condition_variable_and_a_mutex), so perhaps ESP-IDF could add pthread_rwlock support in the future.

Indeed. That's what I was doing inside the libStd patch itself, by re-using and modifying a variant of [this implementation](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/wasm/atomics/rwlock.rs). However it resulted in way too may changes all over the place (related to const fn and static rwlocks), so the cleanest approach is to just have native phtread_rwlock support - somehow.

Out of pure curiosity: are there any *observable* differences between rwlock and a mutex (that is, putting performance concerns with many readers aside, as that won't be the case on this tiny MCU which can only spin a handful of threads anyway)?

Condvars in rust seem to be deeply tied to mutexes, so that's one of the reasons I though assembling this cheat is not so un-shameful.

> 
> In the meantime, `pthread_rwlock` should not be exposed by libc.
> 
> The rest of the changes look fine.

Great. I'll push-force the removal of the rwlock cheat shortly.

Amanieu:
> Out of pure curiosity: are there any observable differences between rwlock and a mutex (that is, putting performance concerns with many readers aside, as that won't be the case on this tiny MCU which can only spin a handful of threads anyway)?

Actually there is, but it's quite subtle and specific to POSIX rwlocks. `pthread_rwlock_rdlock` has the requirement that multiple read locks from the same thread must not deadlock (at least with the default mutex type).

But even without this, I can imagine some contrived situations where forward progress can only happen after multiple threads have acquired a read lock and are waiting for some other condition, which would not work if rwlock was implemented as a mutex.

> Indeed. That's what I was doing inside the libStd patch itself, by re-using and modifying a variant of this implementation. However it resulted in way too may changes all over the place (related to const fn and static rwlocks), so the cleanest approach is to just have native phtread_rwlock support - somehow.

Actually I would suggest implementing `pthread_rwlock` support in C in ESP-IDF itself using the algorithm I linked to. This would allow both C and Rust code to use rwlocks.

Alternatively, it might be acceptable to stub out rwlocks in the standard library as unimplemented for now.

ivmarkov:
> > Indeed. That's what I was doing inside the libStd patch itself, by re-using and modifying a variant of this implementation. However it resulted in way too may changes all over the place (related to const fn and static rwlocks), so the cleanest approach is to just have native phtread_rwlock support - somehow.
> 
> Actually I would suggest implementing `pthread_rwlock` support in C in ESP-IDF itself using the algorithm I linked to. This would allow both C and Rust code to use rwlocks.

We are aligned - that's what I'm saying too. Otherwise, the libStd changes become way too unwieldly.

> 
> Alternatively, it might be acceptable to stub out rwlocks in the standard library as unimplemented for now.

They are used *within* the `sys\unix` libStd implementation itself unfortunately (at multiple places, and growing).
So we do need them. As you say - ideally as a simple but `pthread-` compatible implementation.

ivmarkov:
I've force-pushed the removal of the pthread-rwlock hack. 
When / if this MR gets merged, shall we raise the version of libc?
We need this to be done so that when / if the libStd MR gets merged, it gets these definitions (via an updated dependency to libc).

src/unix/newlib/mod.rs
```diff
@@ -232,23 +230,37 @@ s! {
...
-        size: [0; __SIZEOF_PTHREAD_MUTEX_T],
+        size: [__PTHREAD_INITIALIZER_BYTE; __SIZEOF_PTHREAD_MUTEX_T],
     };
     pub const PTHREAD_COND_INITIALIZER: pthread_cond_t = pthread_cond_t {
-        size: [0; __SIZEOF_PTHREAD_COND_T],
+        size: [__PTHREAD_INITIALIZER_BYTE; __SIZEOF_PTHREAD_COND_T],
     };
     pub const PTHREAD_RWLOCK_INITIALIZER: pthread_rwlock_t = pthread_rwlock_t {
-        size: [0; __SIZEOF_PTHREAD_RWLOCK_T],
+        size: [__PTHREAD_INITIALIZER_BYTE; __SIZEOF_PTHREAD_RWLOCK_T],
```
Amanieu, line 239: What is the plan here? If ESPIDF gains a pthread_rwlock implementation in the future then these numbers will likely change. I would `cfg` rwlock out until it is supported by ESPIDF.

ivmarkov: `pthread_mutex_t` on ESP-IDF is implemented as a pointer to a structure *allocated on the heap*. I would expect the same would hold true for `pthread_rwlock_t`, hence I kept it the same size.

These definitions might also allow me to quickly prototype an "out of tree" implementation of these in the next days/weeks and test without altering libc again.

ivmarkov: JFYI, [here](https://github.com/espressif/esp-idf/blob/master/components/pthread/pthread.c#L574).

Amanieu:
OK I'm happy with rwlock as it is, with the understanding that it will soon be implemented in ESP-IDF.

@bors r+

PR #2313, opened 2021-08-02
freebsd add subset of libprocstat

devnexen:


JohnTitor (approved):

JohnTitor:
Seems you messed up a branch by a rebase.

m-ou-se:
This change broke the freebsd build in the `rust-lang/rust` CI. See https://github.com/rust-lang/rust/pull/88478#issuecomment-912419683

lnicola:
Maybe `libprocstat` needs to be installed on the system?

devnexen:
normally freebsd comes with since it s a system library not a third party package one and even looking at the git repo in the 11.0 release era it was already that way https://github.com/freebsd/freebsd-src/tree/releng/11.0/lib

lnicola:
Well, I ran `./src/ci/docker/run.sh dist-x86_64-freebsd` and got the same error:

```
  = note: /usr/local/x86_64-unknown-freebsd11/bin/ld: cannot find -lprocstat
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

So this is a Docker container, not a real FreeBSD install, right? Then it could be missing some libraries.

devnexen:
It seems ; you did all right in your side.

m-ou-se:
cc @rust-lang/infra ^

pietroalbini:
We're cross-compiling FreeBSD from a Linux container, yes. https://github.com/rust-lang/rust/blob/master/src/ci/docker/host-x86_64/dist-x86_64-freebsd/Dockerfile would need to be updated to make sure the library is present, but I'm not that familiar with FreeBSD cross-compilation so can't give more accurate pointers :(

devnexen:
would it be possible to do similarly as here instead ? https://github.com/uutils/coreutils/blob/2bd556e252fd562777cabe052a2eb811d44c3167/.github/workflows/CICD.yml#L534

pietroalbini:
That's not really feasible with our current CI configuration: we're using cross-compilation for most platforms (everything but Windows and macOS) to standardize our CI as much as possible, and running FreeBSD in a VM would add too much complexity.

I took a quick look at the Docker image, and it seems like the relevant libraries are extracted here, maybe try to include the missing library? https://github.com/rust-lang/rust/blob/e2750baf53aaa60db95f10759f6cf9463dc5a6bd/src/ci/docker/scripts/freebsd-toolchain.sh#L52-L58

devnexen:
ah right in that case it makes sense.

PR #2314, opened 2021-08-02
netbsd log api utmpx flavor addition

devnexen:


devnexen:
odd those intermittent ci issues.

Amanieu:
It doesn't seem to be intermittent, the semver test is failing: https://github.com/rust-lang/libc/runs/3220767531

devnexen:
ok but does not seem to occur in all PR tough

PR #2324, opened 2021-08-07
netbsd/openbsd adding more IPC_* flags.

devnexen:
SHM albeit obsoleteare used in couple of crates.


semarie:
fine with me regarding openbsd. thanks.

PR #2327, opened 2021-08-08
Fix termios constants on redox

coolreader18:
The termios constants on redox were just completely wrong, so I got the correct values from https://gitlab.redox-os.org/redox-os/relibc/-/blob/master/src/header/termios/redox.rs


coolreader18:
Would you be able to cut a new release with this? I'm excited about finally getting termios stuff working on redox, and this was a really prevalent issue on the userspace side.

PR #2331, opened 2021-08-11
sched_get_priority_max/min apply to all BSDs and Solaris

rtzoeller:
`sched_get_priority_max()` and `sched_get_priority_min()` apply to all the BSDs as Solaris-like operating systems.

Applying to all Unix-like OSes failed due to emscripten, among possibly others.


rtzoeller:
I'll update this to drop support for the emscripten targets for now, but in parallel to that I've reached out to the emscripten folks on Discord to see if they'd be willing to stub out these functions without implementing the rest of `sched_*` (like OpenBSD does).

PR #2332, opened 2021-08-12
Add more items in apple targets

GuillaumeGomez:


GuillaumeGomez:
@JohnTitor Is it okay to upgrade the minimum supported rust version? Otherwise I'll simply remove `vnode_info_path` and `proc_vnodepathinfo` from the added structs.

src/unix/bsd/apple/mod.rs
```diff
@@ -678,6 +678,47 @@ s! {
...
+    pub struct vnode_info {
+        pub vi_stat: vinfo_stat,
+        pub vi_type: ::c_int,
+        pub vi_pad: ::c_int,
+        pub vi_fsid: ::fsid_t,
+    }
+
+    pub struct vnode_info_path {
+        pub vip_vi: vnode_info,
+        pub vip_path: [::c_char; ::MAXPATHLEN as usize],
```
JohnTitor, line 715: We could nest the array like `[[::c_char; 32]; 32]` to auto-derive traits.

GuillaumeGomez: Wow, quite the dark magic, I like it!

GuillaumeGomez:
@JohnTitor: The checkers aren't happy, so I guess this solution cannot work...

GuillaumeGomez:
@JohnTitor Seems like it worked! \o/

JohnTitor:
Thanks!
@bors r+

PR #1642, opened 2020-01-11
<sys/timex.h>

dfoxfranke:
Adds bindings for `ntp_adjtime` & co. for Linux, FreeBSD/NetBSD/Darwin, and Solaris/IllumOS.

PR #2333, opened 2021-08-12
Add <sys/timex.h> bindings for linux-musl

dfoxfranke:
Adds bindings for `adjtimex()` & co. for linux-musl. Also adds `clock_adjtime()` to linux-glibc.

### DIFF
--- a/src/unix/linux_like/linux/musl/mod.rs
+++ b/src/unix/linux_like/linux/musl/mod.rs
@@ -249,6 +249,36 @@ s! {
         pub ch_size: ::Elf32_Word,
         pub ch_addralign: ::Elf32_Word,
     }
+
+    pub struct timex {
+        pub modes: ::c_uint,
+        pub offset: ::c_long,
+        pub freq: ::c_long,
+        pub maxerror: ::c_long,
+        pub esterror: ::c_long,
+        pub status: ::c_int,
+        pub constant: ::c_long,
+        pub precision: ::c_long,
+        pub tolerance: ::c_long,
+        pub time: ::timeval,
+        pub tick: ::c_long,
+        pub ppsfreq: ::c_long,
+        pub jitter: ::c_long,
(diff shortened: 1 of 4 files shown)


MarkAtwood:
AWS OSPO review, no OSPO concerns.

src/unix/linux_like/linux/musl/mod.rs
```diff
@@ -249,6 +249,36 @@ s! {
...
+        pub ppsfreq: ::c_long,
+        pub jitter: ::c_long,
+        pub shift: ::c_int,
+        pub stabil: ::c_long,
+        pub jitcnt: ::c_long,
+        pub calcnt: ::c_long,
+        pub errcnt: ::c_long,
+        pub stbcnt: ::c_long,
+        pub tai: ::c_int,
+        pub __padding: [::c_int; 11],
```
JohnTitor, line 274: A padding field doesn't need to be public generally, is there any reason?

dfoxfranke: I suppose it doesn't matter much, but if this one should be private then I should've made the [glibc](https://docs.rs/libc/0.2.99/libc/struct.timex.html) ones private as well when I contributed #1642. So I can either keep this one public, be inconsistent, or make a breaking change to the glibc version. Up to you, I have no strong feelings on the matter.

JohnTitor: Ah makes sense! Let's keep it as-is for now.

JohnTitor (approved):

dfoxfranke:
Looks like a CI test is failing because <http://ftp.debian.org/debian/dists/testing/main/installer-s390x/20200314/images/generic/initrd.debian> has gone 404 (maybe releated to the release of Debian 11)? Obviously unrelated to this PR.

JohnTitor:
Fixed on master, @bors retry

PR #2335, opened 2021-08-13
openbsd: add sys/param.h include for MAXPATHLEN in tests

semarie:
unbreak OpenBSD test target after #2332 .

I checked if the value was right, but not tested it. my bad.


JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

PR #2341, opened 2021-08-14
Add IPV6_FREEBIND to Fuchsia

Thomasdezeeuw:
No idea if this is the correct value, I copied it from Linux.


Thomasdezeeuw:
@tamird can you confirm the value is correct?

Amanieu:
Are you sure this flag is actually supported on fuchsia? It doesn't seem to be handled here: https://fuchsia.googlesource.com/fuchsia/+/refs/heads/main/sdk/lib/fdio/socket.cc

tamird:
Does not appear to exist on Fuchsia. https://cs.opensource.google/search?q=IPV6_FREEBIND&sq=&ss=fuchsia

I  think we could upgrade the definitions from musl but haven't had a need for it yet.

cc @ghanan94

Thomasdezeeuw:
Closing this since Fuchsia doesn't seem to have the constant.

PR #2345, opened 2021-08-15
haiku posix_fadvise/fallocate addition.

devnexen:


devnexen:
The same linux s390x seems to fail since yesterday

Amanieu:
Should be fixed by #2347

@bors r+

PR #2346, opened 2021-08-15
msghdr on linux musl 64 bits, padding fields made public

### LINKED ISSUE #2344: msghdr for musl cannot be instantiated because of private padding fields
If I attempt to build

```rust
 let mut header = libc::msghdr {
            msg_name: std::ptr::null_mut(),
            msg_namelen: 0,
            msg_iov: std::ptr::null_mut(),
            msg_iovlen: 0,
            msg_control: std::ptr::null_mut(),
            msg_controllen: 0,
            msg_flags: 0, 
        };
```

on a generic linux target this works, but with `--target x86_64-unknown-linux-musl` I receive:

```
error: cannot construct `msghdr` with struct literal syntax due to inaccessible fields
  --> init/src/main.rs:27:26
   |
27 |         let mut header = libc::msghdr {
   |                          ^^^^^^^^^^^^

error: aborting due to previous error; 1 warning emitted
```

This is because of private padding used in the musl version of these types: https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/linux/musl/b64/mod.rs#L92

It is notable that in the types below the padding is made public, making this particular struct inconsistent.

Can be worked around with `mem::zeroed` but I think these fields should ideally be made public

devnexen:
closes #2344

### DIFF
--- a/src/unix/linux_like/linux/musl/b64/mod.rs
+++ b/src/unix/linux_like/linux/musl/b64/mod.rs
@@ -94,16 +94,16 @@ s! {
         pub msg_namelen: ::socklen_t,
         pub msg_iov: *mut ::iovec,
         #[cfg(target_endian = "big")]
-        __pad1: ::c_int,
+        pub __pad1: ::c_int,
         pub msg_iovlen: ::c_int,
         #[cfg(target_endian = "little")]
-        __pad1: ::c_int,
+        pub __pad1: ::c_int,
         pub msg_control: *mut ::c_void,
         #[cfg(target_endian = "big")]
-        __pad2: ::c_int,
+        pub __pad2: ::c_int,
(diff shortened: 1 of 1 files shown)


JohnTitor:
Hmm, I don't think we should change them. IIUC constructing it with `mem::zeroed` prevents us from being wrong with a padding or unused field. And users have to specify these values, which isn't ideal. So forcing it makes some sense IMHO.

devnexen:
yes you are right I ve seen cmshdr with public padding fields but we can initialize with Maybe uninit or mem zeroed indeed.

JohnTitor:
Yes, some of the padding/unused fields are public as I (or other reviewers) overlooked or there's some reason.

PR #2354, opened 2021-08-19
Haiku/fix tests

nielx:
This modifies the semver tests to reduce the symbols in the unix platform to what Haiku supports. I have added the symbols to the platforms that inherit the unix symbols, but I could do with a double check that I did not miss any.


JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

PR #2355, opened 2021-08-21
openbsd: export `struct sigcontext` for x86_64

semarie:
based on initial work from @devnexen (see #2189)

tested on OpenBSD (7.0-beta) with rustc 1.54.0

### DIFF
--- a/src/unix/bsd/netbsdlike/openbsd/x86_64.rs
+++ b/src/unix/bsd/netbsdlike/openbsd/x86_64.rs
@@ -3,6 +3,112 @@ use PT_FIRSTMACH;
 pub type c_long = i64;
 pub type c_ulong = u64;
 pub type c_char = i8;
+pub type ucontext_t = sigcontext;
+
+s! {
+    pub struct sigcontext {
+        pub sc_rdi: ::c_long,
+        pub sc_rsi: ::c_long,
+        pub sc_rdx: ::c_long,
+        pub sc_rcx: ::c_long,
+        pub sc_r8: ::c_long,
+        pub sc_r9: ::c_long,
+        pub sc_r10: ::c_long,
+        pub sc_r11: ::c_long,
+        pub sc_r12: ::c_long,
+        pub sc_r13: ::c_long,
+        pub sc_r14: ::c_long,
+        pub sc_r15: ::c_long,
+        pub sc_rbp: ::c_long,
+        pub sc_rbx: ::c_long,
+        pub sc_rax: ::c_long,
+        pub sc_gs: ::c_long,
+        pub sc_fs: ::c_long,
+        pub sc_es: ::c_long,
+        pub sc_ds: ::c_long,
+        pub sc_trapno: ::c_long,
+        pub sc_err: ::c_long,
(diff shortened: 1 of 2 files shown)


src/unix/bsd/netbsdlike/openbsd/x86_64.rs
```diff
@@ -3,6 +3,110 @@ use PT_FIRSTMACH;
...
+        pub sc_es: ::c_long,
+        pub sc_ds: ::c_long,
+        pub sc_trapno: ::c_long,
+        pub sc_err: ::c_long,
+        pub sc_rip: ::c_long,
+        pub sc_cs: ::c_long,
+        pub sc_rflags: ::c_long,
+        pub sc_rsp: ::c_long,
+        pub sc_ss: ::c_long,
+        sc_fpstate: *mut fxsave64,
```
JohnTitor, line 36: Should this be private?

src/unix/bsd/netbsdlike/openbsd/x86_64.rs
```diff
@@ -3,6 +3,110 @@ use PT_FIRSTMACH;
...
+        pub fx_xmm: [[u64; 2]; 16],
+        __fx_unused3: [u8; 96],
+    }
+}
+
+cfg_if! {
+    if #[cfg(feature = "extra_traits")] {
+        impl PartialEq for fxsave64 {
+            fn eq(&self, other: &fxsave64) -> bool {
+                return {self.fx_fcw} == {other.fx_fcw} &&
```
JohnTitor, line 65: Other places don't use curly brackets, let's get rid of them for consistency.

src/unix/bsd/netbsdlike/openbsd/x86_64.rs
semarie: hum, no.

src/unix/bsd/netbsdlike/openbsd/x86_64.rs
semarie: curly brackets are required for `fxsave64` fields: it creates a temporary storage, the value is copied from `fxsave64` field, and operation is done on this storage. due to the fact `fxsave64` is packed, direct access of a field implies an unaligned access, and rustc complains with `error: reference to packed field is unaligned` (see https://github.com/rust-lang/rust/issues/82523)

semarie: in libc, `epoll_event` in linux-like is also `packed`. the field access is done using explicit `let` (for copying the value in aligned storage) but there is only two fields in `epoll_event`. it is why I used temporary storage with curly brackets.

JohnTitor: Ah, I see! Then could you add a comment to explain it? r=me once it's added and commits are squashed.

semarie: done. thanks.

JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

PR #2357, opened 2021-08-22
haiku ucontext for x86_64

devnexen:


JohnTitor:
Seems good, but I'd like to double-check as `ucontext_t` needed some tweaks in the past and we don't have CI for Haiku.
@nielx Could you check it if you have some time?

nielx:
I will have a look!

src/unix/haiku/mod.rs
```diff
@@ -352,6 +352,66 @@ s_no_extra_traits! {
...
+    pub struct fpu_state {
+        pub control: ::c_ushort,
+        pub status: ::c_ushort,
+        pub tag: ::c_ushort,
+        pub opcode: ::c_ushort,
+        pub rip: ::c_ulong,
+        pub rdp: ::c_ulong,
+        pub mxcsr: ::c_uint,
+        pub mxcsr_mask: ::c_uint,
+        _fpreg: [::c_uchar; 16],
```
nielx, line 366: Is this correct?
The definition in headers/posix/arch/x86_64/signal.h is:
```
	union {
		struct x86_64_fp_register fp[8];
		struct x86_64_fp_register mmx[8];
	};
```

And then the definition of a `x86_64_fp_register` is:

```
struct x86_64_fp_register {
	unsigned char value[10];
	unsigned char reserved[6];
};
```
So it seems that the size should be 8*16 bytes?

src/unix/haiku/mod.rs
```diff
@@ -352,6 +352,66 @@ s_no_extra_traits! {
...
+        _xmm: [[::c_uchar; 16]; 16],
+        _reserved_416_511: [[::c_uchar; 3]; 32],
```
nielx, line 368: Any reason why you are making me do math of 3*32 instead of just setting it to 96 bytes like in the header?

src/unix/haiku/mod.rs
```diff
@@ -352,6 +352,66 @@ s_no_extra_traits! {
...
+    }
+
+    #[cfg(target_arch = "x86_64")]
+    pub struct xstate_hdr {
+        pub bv: ::c_ulong,
+        pub xcomp_bv: ::c_ulong,
+        _reserved: [[::c_uchar; 2]; 24],
```
nielx, line 375: Again, any reason why you deviate from the header here? it is just defined as 48 c_chars.

src/unix/haiku/mod.rs
```diff
@@ -352,6 +352,66 @@ s_no_extra_traits! {
...
+    #[cfg(target_arch = "x86_64")]
+    pub struct fpu_state {
+        pub control: ::c_ushort,
+        pub status: ::c_ushort,
+        pub tag: ::c_ushort,
+        pub opcode: ::c_ushort,
+        pub rip: ::c_ulong,
+        pub rdp: ::c_ulong,
+        pub mxcsr: ::c_uint,
+        pub mxcsr_mask: ::c_uint,
```
nielx, line 365: The name in the header is `mscsr_mask`.

nielx (changes requested):
I think there may be some issues. One thing that might be helpful is to run the automated tests yourself. In the root of the source, do a `./ci/run.sh --target=x86_64-unknown-haiku` and you will get the feedback on whether it works or not.

devnexen:
> I think there may be some issues. One thing that might be helpful is to run the automated tests yourself. In the root of the source, do a `./ci/run.sh --target=x86_64-unknown-haiku` and you will get the feedback on whether it works or not.

Unfortunately I tried to run CI already but I run beta 3 version and the ctest2 crate compilation gets stuck

nielx:
If you are running in a VM, you might want to give it more memory. My 2GB Haiku Beta3 VM can run the tests.

devnexen:
I set it from start with 4GB the overall VM does not get stuck just the crate compilation but I can use the system doing other things.

nielx:
In that case I am not quite sure what is going on. In any case, the remarks I made are things that also trip up the tests.

Two other points:
* What is the rationale for implementing this in x86_64 only (and not in x86?)
* Are objects of these types always created by the system, or is the user expected to create them themselves? In the latter case, all members should be public, even the reserved data members. Otherwise they cannot be initialized by the user.

devnexen:
> In that case I am not quite sure what is going on. In any case, the remarks I made are things that also trip up the tests.
> 
> Two other points:
> 
> * What is the rationale for implementing this in x86_64 only (and not in x86?)
I thought about it but as a second PR once this one cleared up. also crates using it tend to focus on 64 bits (for now at least).
> * Are objects of these types always created by the system, or is the user expected to create them themselves? In the latter case, all members should be public, even the reserved data members. Otherwise they cannot be initialized by the user.
Mostly created from signal handlers/crash purpose thus users usually tend to look at frame pointer, program counters, generic registers etc ....

PR #2358, opened 2021-08-22
Add `strtof` for all platforms that have `strtod`

hcsch:
All platforms that have `strtod` very likely also have `strtof`. Having `strtof` allows for fuzzing of [`hexf-parse::parse_hexf32`](https://github.com/lifthrasiir/hexf) comparing against `strtof`, which can't (easily) be done with `strtod`, due to float rounding and over-/underflow behavior.

I'm unsure of whether the `strtof` declaration in `src/unix/mod.rs` also needs a link name adjustment for x86 macos. I haven't been able to find anything about this after a rough search in this repo and on Google. I haven't added it for now since only `strtod` (edit: out of the `strtoX` functions) seems to have the changed link name (and I don't have a mac handy to test the available symbols with), but feel free to tell me if it's needed and I'll adjust the PR accordingly.

### DIFF
--- a/libc-test/semver/android.txt
+++ b/libc-test/semver/android.txt
@@ -3215,6 +3215,7 @@ strsignal
 strspn
 strstr
 strtod
+strtof
 strtok
 strtol
 strtoul
--- a/libc-test/semver/unix.txt
+++ b/libc-test/semver/unix.txt
@@ -814,6 +814,7 @@ strrchr
 strspn
 strstr
 strtod
+strtof
 strtok
 strtol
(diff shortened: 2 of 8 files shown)


JohnTitor (approved):

JohnTitor:
Looks good, thanks!
@bors r+

hcsch:
Should I open a version bump PR after this or is one upcoming already? I'd like to use `strtof` for the mentioned use case of fuzzing `hexf-parse` while comparing against it.

PR #2360, opened 2021-08-23
haiku adding cpuid_info data.

devnexen:


Amanieu:
@JohnTitor Do we have any precedent for handling anonymous structs here?

JohnTitor:
We do, you can find some existing structs prefixed with `__c_anonymous`. I usually accept such a change with a check if it matches the declaration (in this case, I'd check `cpuid_info`'s member).

PR #2361, opened 2021-08-24
Bump to 0.2.101

hcsch:
New release to allow for usage of `strtof` introduced in #2358 for
fuzzing of `hexf-parse` while comparing against `strtof` results.


hcsch:
Seems the macOS 1.30.0 build channel didn't even get to building the code, since it failed to retrieve `rustc`: https://github.com/rust-lang/libc/runs/3420467609

PR #2362, opened 2021-08-24
Add L4re support for arm

Steav005:
This PR adds L4re support for arm with uclibc similar to previous PRs for L4re on x86. 

I already tested these changes with a bootstrapped rust compiler using the L4Re Linker `L4Bender` as I described it in:
https://github.com/humenda/rustl4re/pull/3

There is currently no arm L4re target in the official [Rust repo](https://github.com/rust-lang/rust).
And I do not know if it is a good Idea to add one right know since the linker `L4Bender` is not present:
https://github.com/rust-lang/rust/pull/54425


JohnTitor:
Closing in favor of the above comment, feel free to re-open if there's a reason to have it on this crate.

PR #2364, opened 2021-08-25
Fix statfs for DragonFlyBSD

rtzoeller:
The statfs struct was missing several padding areas and used incorrect array lengths, causing all values to be incorrect.

DragonFlyBSD's own documentation seems to be wrong here, which is likely how this ended up being incorrect in the first place.

Source: https://github.com/DragonFlyBSD/DragonFlyBSD/blob/27ea30e396c40d8feb44d76002b79153e96ecf8f/sys/sys/mount.h#L95
Documentation: https://man.dragonflybsd.org/?command=statfs&section=2

### DIFF
--- a/src/unix/bsd/freebsdlike/dragonfly/mod.rs
+++ b/src/unix/bsd/freebsdlike/dragonfly/mod.rs
@@ -247,6 +247,7 @@ s_no_extra_traits! {
     }
 
     pub struct statfs {
+        __spare2: ::c_long,
         pub f_bsize: ::c_long,
         pub f_iosize: ::c_long,
         pub f_blocks: ::c_long,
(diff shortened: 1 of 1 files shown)


src/unix/bsd/freebsdlike/dragonfly/mod.rs
```diff
@@ -256,15 +257,18 @@ s_no_extra_traits! {
...
         pub f_owner: ::uid_t,
-        pub f_type: i32,
-        pub f_flags: i32,
+        pub f_type: ::c_int,
+        pub f_flags: ::c_int,
         pub f_syncwrites: ::c_long,
         pub f_asyncwrites: ::c_long,
         pub f_fstypename: [::c_char; 16],
-        pub f_mntonname: [::c_char; 90],
+        pub f_mntonname: [::c_char; 80],
```
Amanieu, line 265: Do we have constants that we could use here?

rtzoeller: The C source defines `MFSNAMELEN` and `MNAMELEN`, but I don't see any equivalents in rust's libc. I can add these as constants if you'd prefer.

Amanieu: You're right, none of the other targets use a constant. I guess it's fine as it is.

PR #2373, opened 2021-08-27
solarish adding ucontext struct

devnexen:


src/unix/solarish/x86_64.rs
```diff
@@ -0,0 +1,129 @@
...
+        pub status: u32,
+        pub xstatus: u32,
+    }
+}
+
+s_no_extra_traits! {
+    #[cfg(libc_union)]
+    pub union __c_anonymous_fp_reg_set {
+        pub fpchip_state: __c_anonymous_fpchip_state,
+        pub f_fpregs: [[u32; 13]; 10],
```
Amanieu, line 26: This really should be `[u32; 130]` but that may not work with older Rust versions used in CI.

@JohnTitor Thoughts?

devnexen: yes why I did just for precaution

JohnTitor: Yes, we've been using this workaround in some places and it should work fine.

PR #2375, opened 2021-08-28
android prctl adding special flags

devnexen:


src/unix/linux_like/android/mod.rs
```diff
@@ -2438,6 +2438,13 @@ pub const PF_VSOCK: ::c_int = AF_VSOCK;
 pub const PROP_VALUE_MAX: ::c_int = 92;
 pub const PROP_NAME_MAX: ::c_int = 32;
 
+// sys/prctl.h
+pub const PR_SET_VMA: ::c_int = 0x53564d41;
+pub const PR_SET_VMA_ANON_NAME: ::c_int = 0;
+// aliases used but not from header
+pub const ANDROID_PR_SET_VMA: ::c_int = PR_SET_VMA;
+pub const ANDROID_PR_SET_VMA_ANON_NAME: ::c_int = PR_SET_VMA_ANON_NAME;
```
JohnTitor, line 2446: Where do these aliases come from?

devnexen: sorry I missed your comment... seen them here and there on chromium for ex

devnexen: actually, after second thoughts prefer to remove those.

JohnTitor (approved):

PR #2379, opened 2021-08-29
windows bcrypt api, solves also linkage issue with calling convention…

devnexen:
… from system to stdcall

### DIFF
--- a/src/windows/mod.rs
+++ b/src/windows/mod.rs
@@ -497,7 +497,7 @@ extern "C" {
     pub fn aligned_free(ptr: *mut ::c_void);
 }
 
-extern "system" {
+extern "stdcall" {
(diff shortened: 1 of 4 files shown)


Amanieu:
I think this is out of scope for libc, these APIs are provided by the winapi crate: https://docs.rs/winapi/0.3.9/winapi/shared/bcrypt/index.html

src/windows/mod.rs
```diff
@@ -497,7 +497,7 @@ extern "C" {
     pub fn aligned_free(ptr: *mut ::c_void);
 }
 
-extern "system" {
+extern "stdcall" {
```
Amanieu, line 500: `system` is already equivalent to `stdcall` on 32-bit x86. On all other platforms it is equivalent to `C`. All this change does it break non-x86 platforms.

PR #2387, opened 2021-09-03
Revert "fuchsia add getrandom equivalent zx_cprng_draw."

erickt:
This reverts commit 5b8616e478d3f76cfd4bc94f9dd4499103c41d51.

I'm on the Fuchsia Rust team, and we would like this reverted for a few reasons. First, these functions are not actually exposed in our `libc`. Instead, it comes from our `libzircon` library. Second, we haven't solidified these APIs, nor have we committed to exactly which library to expose them from.

Instead, we've created https://crates.io/crates/fuchsia-cprng to expose this functionality. This lets us more easily track our downstream users, and help them update if we ever make a breaking change.

cc @devnexen and @anp

### DIFF
--- a/libc-test/semver/fuchsia.txt
+++ b/libc-test/semver/fuchsia.txt
@@ -1384,6 +1384,3 @@ utimensat
 vhangup
 vmsplice
 waitid
-zx_cprng_add_entropy
-zx_cprng_draw
-zx_status_t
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):
Thanks for the follow-up! It'd be ideal to have some notice period, but this should be urgent thus I'm going just to r+. Sorry for the breakage.

PR #2390, opened 2021-09-05
Move Linux's `POLLRDHUP` into `linux_like` and fix its type.

sunfishcode:
This fixes two errors in #2247.

 - It moves the definitions of `POLLRDHUP` out of `linux_like/linux`
   and into `linux_like`, so that they're available on Android as well.

 - It changes the type from `c_int` to `c_short` to match the other
   `POLL*` flags.

This second change is a breaking change, and I'm fine reverting it and
leaving it as the old type if that's the right thing to do.

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -3037,8 +3037,16 @@ pub const SOL_CAN_BASE: ::c_int = 100;
 pub const CAN_INV_FILTER: canid_t = 0x20000000;
 pub const CAN_RAW_FILTER_MAX: ::c_int = 512;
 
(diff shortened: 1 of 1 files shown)


JohnTitor:
I'm not sure how many users are affected by this change but I'd like to follow our breaking change policy, just in case: https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md#breaking-change-policy

sunfishcode:
Sounds good. I've now changed the PR to just introduce `deprecated` warnings for now.

JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

PR #2396, opened 2021-09-12
Support for the ESP-IDF framework - 3 forgotten mappings

ivmarkov:
When implementing [the initial support for ESP-IDF](https://github.com/rust-lang/libc/pull/2310) a month ago, I managed to forget to properly decorate these 3 libc bindings for the ESP-IDF.

So now we are in the uncomfortable situation that using STD's `TcpListener::bind` results in a linkage error.

This PR is fixing this, as well as properly exposing two additional ESP-IDF APIs, implemented via LwIP.


JohnTitor (approved):
Thanks for the follow-up!

PR #2398, opened 2021-09-12
Add more redox clock constants

coolreader18:


Amanieu:
Are you sure about the value for `CLOCKS_PER_SEC`?

https://github.com/redox-os/newlib/blob/master/newlib/libc/include/machine/time.h

coolreader18:
Hm. Well, I got this from [relibc](https://gitlab.redox-os.org/redox-os/relibc/-/blob/master/src/header/time/constants.rs), which I think is the default libc on redox?

PR #2399, opened 2021-09-13
android fix dev_t type for 32 and 64 bits separately.

### LINKED ISSUE #2394: Size of dev_t/ino_t on 32b android
`libc::dev_t`/`libc::ino_t` are u64 for linux (32/64) and android (64 only), but only u32 for android 32. This causes some weird types for some fields, like in `libc::stat` (the struct, not the function), for which `st_dev` and `st_ino` are `libc::c_ulonglong`, u64. 

This causes some strange results where libc::stat.st_dev and libc::stat.st_dev are u64, but `dev_t`/`ino_t` can be u32.

this code compiles for the linux and android64 target I tested, but not for `i686-linux-android` and `armv7-linux-androideabi` 
```rust
use core::mem::MaybeUninit;
use libc;

fn main() {
    let mut stat = unsafe { MaybeUninit::<libc::stat>::zeroed().assume_init() }; // bad but irrelevant
    let dev: libc::dev_t = stat.st_dev;
    let ino: libc::ino_t = stat.st_ino;
}
```

Real life issue:
When compiling the crate [`fslock`](https://crates.io/crates/fslock), targeting `i686-linux-android` or `armv7-linux-androideabi` with its feature `multilock` enabled, a type error happens because the code assumes stat.st_dev is of type dev_t, and stat.st_ino is of type ino_t. This seems like a reasonable assumption, and it is very strange to me that it does not hold on some platforms.

devnexen:
closes #2394

### DIFF
--- a/src/unix/linux_like/android/b32/mod.rs
+++ b/src/unix/linux_like/android/b32/mod.rs
@@ -8,6 +8,7 @@ pub type off64_t = ::c_longlong;
 pub type sigset_t = ::c_ulong;
 pub type socklen_t = i32;
 pub type time64_t = i64;
+pub type dev_t = u32;
 pub type __u64 = ::c_ulonglong;
 
 s! {
--- a/src/unix/linux_like/android/b64/mod.rs
+++ b/src/unix/linux_like/android/b64/mod.rs
@@ -6,6 +6,7 @@ pub type c_ulong = u64;
 pub type mode_t = u32;
 pub type off64_t = i64;
 pub type socklen_t = u32;
+pub type dev_t = u64;
 
 s! {
     pub struct sigset_t {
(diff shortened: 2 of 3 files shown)


trinity-1686a:
I'm sorry, I think I failed to make myself understood.
This patch changes nothing to my problem. `dev_t` is already 64b on 64b android, and 32b on 32b android.

I'm questioning if it shouldn't be 64b on both platforms, as it is for linux.
And even more, I'm questioning why `stat.st_dev` is of type `dev_t` and `stat.st_ino` of type `ino_t` when building for linux, or for 64b android, but not when building for 32b android.

devnexen:
well according to android headers dev_t is of uint32_t on 32 bits

PR #2401, opened 2021-09-16
Add more apple types

GuillaumeGomez:


GuillaumeGomez:
cc @JohnTitor

GuillaumeGomez:
The failure seems legit since I'm computing since inside a const. What would be the correct way to do so otherwise?

GuillaumeGomez:
Silly me: the answer was simply to use macros. So I created a new macro to handle const definition. :)

GuillaumeGomez:
So macros don't compute the value directly. Sadness...

GuillaumeGomez:
Updated!

PR #2403, opened 2021-09-17
uclibc: fix semver files, add misc symbols

skrap:
This commit gets uclibc support working again, and fixes the semver lists to segregate gnu- and musl- specific symbols to those libc variants.

There is a [separate PR](https://github.com/rust-lang/rust/pull/88952) against rust itself to add armv7-unknown-linux-uclibceabihf as a tier 3 platform, and this commit is in support of that effort.  Until that's done, we should not use the CI for that platform, but I'm including the CI file here in anticipation of turning it on in the not-too-distant future.


libc-test/build.rs
```diff
@@ -2846,6 +2846,35 @@ fn test_linux(target: &str) {
...
+            | "IPV6_HDRINCL"
+            | "IPV6_MULTICAST_ALL"
+            | "IPV6_PMTUDISC_INTERFACE"
+            | "IPV6_PMTUDISC_OMIT"
+            | "IPV6_ROUTER_ALERT_ISOLATE"
+            | "O_TMPFILE"
+            | "PACKET_MR_UNICAST"
+            | "PTRACE_EVENT_STOP"
+            | "PTRACE_O_EXITKILL"
+            | "PTRACE_O_SUSPEND_SECCOM"
```
Amanieu, line 2868: This looks like a typo?

skrap: Yep, thanks! Removed.

Amanieu:
A lot of tests failing with "constant is never used". Are you sure these are exported publicly all the way to the crate root?

skrap:
> A lot of tests failing with "constant is never used". Are you sure these are exported publicly all the way to the crate root?

Ah, that's for the mips build, which is a target I did not test.  I'll fix.  (There's a few uclibc targets which I don't really know how to build for, but I should be able to figure it out.)

skrap:
Hm, the breakage is now in the `x86_64-unknown-l4re-uclibc` target.  I don't know how to build this tier 3 target (I'm not familiar with the L4RE project), so I may just take my best guess as to the fixes and push.

Amanieu:
Let's give it a try!

@bors r+

PR #2406, opened 2021-09-17
Raise libc's FreeBSD ABI to 12

asomers:
FreeBSD 11 will be EoL on 30-Sept-2021.  Update libc's ABI to the
minimum supported version of FreeBSD.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -43,16 +43,16 @@ fn main() {
     let libc_check_cfg = env::var("LIBC_CHECK_CFG").is_ok() || rustc_minor_ver >= 80;
 
     // The ABI of libc used by std is backward compatible with FreeBSD 12.
-    // The ABI of libc from crates.io is backward compatible with FreeBSD 11.
+    // The ABI of libc from crates.io is backward compatible with FreeBSD 12.
     //
     // On CI, we detect the actual FreeBSD version and match its ABI exactly,
     // running tests to ensure that the ABI is correct.
     let which_freebsd = if libc_ci {
-        which_freebsd().unwrap_or(11)
+        which_freebsd().unwrap_or(12)
     } else if rustc_dep_of_std {
         12
     } else {
-        11
+        12
     };
     match which_freebsd {
         x if x < 10 => panic!("FreeBSD older than 10 is not supported"),


Amanieu:
LGTM, but cc @JohnTitor just in case.

JohnTitor:
If rustc drops it, I think it's fine. Rustc's CI currently uses FreeBSD 11 (https://github.com/rust-lang/rust/blob/35c8f2612ffea42cc09350accdba934e85a19f35/src/ci/docker/scripts/freebsd-toolchain.sh#L8), could you ask infra folks if it's okay, and make a PR? When it's approved, I'll r+ this one.

asomers:
rust PR: https://github.com/rust-lang/rust/pull/89083 .

JohnTitor:
Triage: I guess we should wait for https://github.com/rust-lang/rust/issues/89058, and I wonder if we could release this along with the musl change.

asomers:
@JohnTitor there is a possibility that this change could cause breakage downstream, for example if a downstream crate assumes something about the size of ino_t.  Is there any way to use crater to check whether downstream crates will still compile?

workingjubilee:
Tagging this as breakage-candidate on the same "should be included in the upcoming review" rationale.

workingjubilee:
Also I hit up grep.app and searched for `libc::ino_t`, and I didn't find many users of it at all. Most of them were perfectly "by value" uses. The closest I can find is this thingie, "Reverie", which assumes that `libc::ino_t` will fit inside u32... and in that case, it's done [via a direct cast rather than unsafe code](https://github.com/facebookexperimental/reverie/blob/03cbd6044df77fe5d601bd59b49e7d9c91b7846c/reverie-ptrace/src/gdbstub/commands/base/_vFile.rs).

Also, it's Linux-only to begin with.

jrtc27:
Is any progress happening on this? Given Rust's continued inflexibility regarding OS ABIs, this is blocking FreeBSD/riscv from having a working Rust toolchain.

workingjubilee:
- FCP: https://github.com/rust-lang/rust/issues/89058
- Rust toolchain update: https://github.com/rust-lang/rust/pull/97944

@asomers Please resolve the merge conflict.

asomers:
Merge conflict resolved, @workingjubilee .

workingjubilee:
My inclination is also to Just Merge This Already and have it go out libc `0.2.{next}`, with disregard to users (incorrectly) assuming things about the type of `ino_t` (hopefully they used the type alias fully knowing its value can differ from platform to platform?), but given the concern initially raised by @asomers, [I started a conversation on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/351149-t-libs.2Fcrates/topic/FreeBSD.20and.20libc) to feel out if there's consensus or additional concerns.

Also after further reading the Reverie thing is a weird gdb thing that has its own definition and any resemblance to `stat` is in truth incidental, apparently?

I feel like https://github.com/rust-lang/libc/pull/3139 is a better example of something we _might_ want to wait for 0.3.0 before accepting. Might.

JohnTitor:
Given https://github.com/rust-lang/rust/pull/114521 has been merged and https://github.com/rust-lang/rust/issues/89058 is now accepted, we can go ahead with this change. Removing FreeBSD 11 (and older) support should be announced somewhere (we may deprecate the modules first).
@bors r+

JohnTitor:
@bors retry r-
Ah, hm, I misunderstood this change. does this force users to use FreeBSD 12 unconditionally? Then I guess we have to make a period to announce the removal.

JohnTitor:
Maybe making a Rust inside blog post before releasing would be an alternative way to announce, I guess.

BlackHoleFox:
@JohnTitor A +1 from me at least on the blog post idea. It seemed to work fine for the Apple minimum bumps and I have yet to see anyone calling for my head for lack of notice.

tgross35:
We can probably drop this without an announcement at this point since FreeBSD 11 has been EOL for over three years, long enough that I don't think anyone would be reasonably expecting support. FreeBSD 12 is EOL but we won't drop that here until Rust moves off of it and the platform docs merge (https://github.com/rust-lang/rust/pull/129220).

@rustbot author for a rebase

asomers:
@tgross35 since this PR was opened main and 0.2 have become separate branches.  Since breaking changes are allowed in main, the barrier to merging this PR should be even lower, right?

tgross35:
That is correct. I think we can probably drop 11 from both branches, then maybe drop 12 from `main`? If there is a policy that makes sense then we can put it in the readme.

asomers:
> That is correct. I think we can probably drop 11 from both branches, then maybe drop 12 from `main`? If there is a policy that makes sense then we can put it in the readme.

I would love to do that.  The alternative suggestion that I would make would be to drop 11 from libc-0.2, for riscv64 only.  The reason being that FreeBSD 11 didn't even support riscv64, so there are no backwards-compatibility concerns.  Then after some period of time if there are no complaints from riscv users we could drop it for all architectures.

tgross35:
> > That is correct. I think we can probably drop 11 from both branches, then maybe drop 12 from `main`? If there is a policy that makes sense then we can put it in the readme.
> 
> I would love to do that. The alternative suggestion that I would make would be to drop 11 from libc-0.2, for riscv64 only. The reason being that FreeBSD 11 didn't even support riscv64, so there are no backwards-compatibility concerns. Then after some period of time if there are no complaints from riscv users we could drop it for all architectures.

I think we may as well drop 11 entirely if we intend to drop it at all, but if there are concerns that the version is still relevant then maybe the blog post idea is worth it? Feel free to author something short, just submit a PR adding to [this directory](https://github.com/rust-lang/blog.rust-lang.org/tree/999512e190ed857180377c04ed4e1c529749eeee/posts/inside-rust).

In any case you are the target maintainer so I'm fine with whatever you decide here :) just update this PR with as is reasonable.

asomers:
@tgross35 I've rebased the PR.  I think we can merge this to main now.  I tested it by running the test suites of several crates with a patched libc 0.2.164 that used a FreeBSD 12 ABI.  All passed.  Interestingly, one (cap-std) failed with the *unmodified* libc, but passed with the modified one.  That's a direct result of the 64-bit inode change.

[✓] bfffs
[✓] bsd-pidfile-rs
[✓] cap-ftpd
[✓] cap-std   ( fails with vanilla libc; works with patched )
[✓] capsicum-net
[✓] capsicum-rs
[✓] fsx-rs
[✓] fuse-ufs
[✓] gstat-rs
[✓] libnv-rs
[✓] mdconfig
[✓] mio
[✓] mio-aio
[✓] nix
[✓] rctl
[✓] rustix
[✓] sysctl-rs
[✓] tokio
[✓] xattr
[✓] xfuse
[✓] ztop

tgross35 (approved):

PR #2410, opened 2021-09-19
BSD add utrace calls

devnexen:


semarie (approved):
it built fine on openbsd

PR #2414, opened 2021-09-21
Add initial support for m68k-unknown-linux-gnu

glaubitz:
This adds initial support for m68k-unknown-linux-gnu to libc.

CC @ricky26 @jrtc27


src/unix/linux_like/linux/gnu/b32/m68k/mod.rs
```diff
@@ -0,0 +1,859 @@
...
+        pub cgid: ::gid_t,
+        pub mode: ::mode_t,
+        __seq: u32,
+        __pad1: u32,
+        __glibc_reserved1: u64,
+        __glibc_reserved2: u64,
+    }
+
+    pub struct stat64 {
+        pub st_dev: ::dev_t,
```
jrtc27, line 50: Missing __pad1 after this

src/unix/linux_like/linux/gnu/b32/m68k/mod.rs
```diff
@@ -0,0 +1,859 @@
...
+        pub f_fsid: ::c_ulong,
+        __f_unused: ::c_int,
+        pub f_flag: ::c_ulong,
+        pub f_namemax: ::c_ulong,
+        __f_spare: [::c_int; 6],
+    }
+
+    pub struct shmid_ds {
+        pub shm_perm: ::ipc_perm,
+        __glibc_reserved1: ::c_uint,
```
jrtc27, line 104: Technically these are all c_ulong

src/unix/linux_like/linux/gnu/b32/m68k/mod.rs
```diff
@@ -0,0 +1,859 @@
...
+        pub shm_cpid: ::pid_t,
+        pub shm_lpid: ::pid_t,
+        pub shm_nattch: ::shmatt_t,
+        __glibc_reserved5: ::c_ulong,
+        __glibc_reserved6: ::c_ulong,
+    }
+
+    pub struct msqid_ds {
+        pub msg_perm: ::ipc_perm,
+        __glibc_reserved1: ::c_uint,
```
jrtc27, line 121: These all need to be shifted down one field (reserved1 is after stime, etc) (and are technically c_ulong again)

src/unix/linux_like/linux/gnu/b32/m68k/align.rs
```diff
@@ -0,0 +1,7 @@
+s_no_extra_traits! {
+    #[allow(missing_debug_implementations)]
+    #[repr(align(8))]
+    pub struct max_align_t {
+        priv_: [f32; 4]
```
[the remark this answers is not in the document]
glaubitz: Hmm, I think the above was actually your recommendation.

Or did the ABI change in the mean time?

src/unix/linux_like/linux/gnu/b32/m68k/mod.rs
glaubitz: Done. FWIW, I just noticed that glibc upstream has replaced reserved1 through reserved3 in the mean time:

```C
struct msqid_ds
{
  struct ipc_perm msg_perm;     /* structure describing operation permission */
#if __TIMESIZE == 32
  __time_t msg_stime;           /* time of last msgsnd command */
  unsigned long int __msg_stime_high;
  __time_t msg_rtime;           /* time of last msgsnd command */
  unsigned long int __msg_rtime_high;
  __time_t msg_ctime;           /* time of last change */
  unsigned long int __msg_ctime_high;
#else
  __time_t msg_stime;           /* time of last msgsnd command */
  __time_t msg_rtime;           /* time of last msgsnd command */
  __time_t msg_ctime;           /* time of last change */
#endif
  __syscall_ulong_t __msg_cbytes; /* current number of bytes on queue */
  msgqnum_t msg_qnum;           /* number of messages currently on queue */
  msglen_t msg_qbytes;          /* max number of bytes allowed on queue */
  __pid_t msg_lspid;            /* pid of last msgsnd() */
  __pid_t msg_lrpid;            /* pid of last msgrcv() */
  __syscall_ulong_t __glibc_reserved4;
  __syscall_ulong_t __glibc_reserved5;
};
```

jrtc27: Yes, though same layout. I think it makes sense to use the old names here so it matches other 32-bit architectures, then they can all be switched at once if needed in rust.

src/unix/linux_like/linux/gnu/b32/m68k/mod.rs
```diff
@@ -0,0 +1,859 @@
...
+    pub struct shmid_ds {
+        pub shm_perm: ::ipc_perm,
+        __glibc_reserved1: ::c_uint,
+        pub shm_atime: ::time_t,
+        __glibc_reserved2: ::c_uint,
+        pub shm_dtime: ::time_t,
+        __glibc_reserved3: ::c_uint,
+        pub shm_ctime: ::time_t,
+        __glibc_reserved4: ::c_uint,
+        pub shm_segsz: ::size_t,
```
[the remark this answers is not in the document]
jrtc27: That too yes

JohnTitor:
If it works fine on the platform, I could rubber-stamp this.

glaubitz:
I will have to fix the remaining issues first that @jrtc27 has mentioned. I'll ping you back once we're ready.

src/unix/linux_like/linux/gnu/b32/m68k/align.rs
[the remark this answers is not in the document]
glaubitz: So, it should be:

```
#[allow(missing_debug_implementations)]
    #[repr(align(2))]
    pub struct max_align_t {
        priv_: [i8; f12]
    }
}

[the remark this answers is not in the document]
Comment (glaubitz) [reply src/unix/linux_like/linux/gnu/b32/m68k/align.rs:5]: If I understood you correctly, I will change the PR accordingly and fix the issues below as well.

File: src/unix/linux_like/linux/gnu/b32/m68k/mod.rs (hunk 5 above)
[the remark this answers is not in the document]
Comment (glaubitz) [reply src/unix/linux_like/linux/gnu/b32/m68k/mod.rs:111]: Should be fixed now.

File: src/unix/linux_like/linux/gnu/b32/m68k/align.rs (hunk 4 above)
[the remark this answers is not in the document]
Comment (glaubitz) [reply src/unix/linux_like/linux/gnu/b32/m68k/align.rs:5]: Should be fixed now.

**Comment by glaubitz** [comment]
I will have to review all the constants again before we can consider this ready-for-merge.

**Comment by glaubitz** [comment]
@jrtc27 @JohnTitor I have updated the definitions now and they should now all be correct.

**Comment by glaubitz** [comment]
Could someone clear the S-waiting-on-author? Thanks!

File: src/unix/linux_like/linux/gnu/b32/m68k/align.rs [hunk 6]
```diff
@@ -0,0 +1,7 @@
+s_no_extra_traits! {
+    #[allow(missing_debug_implementations)]
+    #[repr(align(2))]
+    pub struct max_align_t {
+        priv_: [i8; f12]
```
Amanieu, line 5: What `f12` here?

glaubitz: I made that change because @jrtc27 suggested it here https://twitter.com/colonia00/status/1490059057159417858?s=20&t=XN5ECDY5QUuYaXohzbcCjQ

Although I'm not sure about the correct syntax. If you can help me with the correct values, it would be appreciated.

glaubitz: I made the change here because @jrtc27 suggested it here https://github.com/rust-lang/libc/pull/2414#pullrequestreview-759790831

I'm not sure whether it's correct though. If you could help me with the proper values that would be appreciated.

Amanieu: I mean, why is it `f12` instead of `12`? It doesn't seem like it would compile as it is.

glaubitz: Oh, that's a typo. It did compile though.

Thanks for catching this, will fix this in a minute.

glaubitz: Fixed and pushed. Thank you for catching this!

glaubitz:
@jrtc27 Could you briefly comment on the `max_align_t` struct? I'm not sure I got it right:

```rust
#[allow(missing_debug_implementations)]
#[repr(align(2))]
pub struct max_align_t {
       priv_: [i8; 12]
}
```

@Amanieu I'm not sure whether the current `max_align_t` matches what the glibc does.

Amanieu:
My recommendation would be to check against a m68k C toolchain:

```c
#include <stddef.h>
int foo = sizeof(max_align_t);
```

Generate assembly code with `-S` and check the value of `foo`.

@bors r-

Let me know once you've checked the value.

glaubitz:
@Amanieu Perfect, thank you! That's what I was looking for!

glaubitz:
OK, it's 20 bytes:

```Assembler
#NO_APP
        .file   "max_align.c"
        .text
        .globl  foo
        .data
        .align  2
        .type   foo, @object
        .size   foo, 4
foo:
        .long   20
        .ident  "GCC: (Debian 11.2.0-14) 11.2.0"
        .section        .note.GNU-stack,"",@progbits
```

Not sure how that translates into the Rust struct.

Would that be:

```Rust
#[allow(missing_debug_implementations)]
#[repr(align(2))]
pub struct max_align_t {
       priv_: [i8; 20]
}
```

?

glaubitz:
@Amanieu I think it should be correct now. The overall size is 20 bytes which can be represented by `[i8; 20]`.

PR #2415, opened 2021-09-21
Add struct sock_txtime and related flags

ghedo:
These are needed to use the SO_TXTIME socket option on Linux, which is
already exposed.


JohnTitor (approved):

ghedo:
Just realized I didn't add the new things to `libc-test/semver/linux.txt`, not sure if that would explain the failure on mips-musl though...

JohnTitor:
The semver check is unrelated, something on MIPS is different.

ghedo:
According to https://github.com/rust-lang/libc/blob/master/ci/docker/mips-unknown-linux-musl/Dockerfile#L9, it looks like the mips+musl CI builds use Linux 4.14, but `sock_txtime` requires 4.19, so I put the new struct and constants behind `#[cfg(not(all(target_env = "musl", target_arch = "mips")))]` which I think should fix the problem for those builds.

Does this seem reasonable @JohnTitor?

JohnTitor:
Thanks for investigating, let's try again :)
@bors r+

PR #2417, opened 2021-09-23
linux/android aarch64 add user_regs_struct and user_fpsimd_struct

devnexen:


JohnTitor (approved):

irgstg:
Hi,

Seems like there is a wide awareness around this issue:
https://github.com/gnzlbg/ctest/pull/89
https://github.com/rust-lang/rust/issues/54341

Do you think that a PR which includes `user_regs_struct` alone, while opening an issue regarding `user_fpsimd_struct` would be a better idea?

Thanks ;)

irgstg:
I think it'll pass w/o any problems now @Amanieu

irgstg:
I don't see how this fail is relate to this PR

devnexen:
test time out it seems. it happens on occasion.

PR #2426, opened 2021-09-27
fix clippy warnings

aman2909verma:
This pull request fixes some of the Clippy warnings in the repo.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -122,14 +122,18 @@ fn rustc_minor_nightly() -> Option<(u32, bool)> {
 
 fn which_freebsd() -> Option<i32> {
     let output = std::process::Command::new("freebsd-version").output().ok();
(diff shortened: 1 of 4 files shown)


JohnTitor:
I don't think it's a good idea to apply Clippy lints to the libc crate, as it's a bit special crate and has to support an old Rust. For instance, we cannot use `?` operators on an older Rust.

aman2909verma:
I have reverted the conditional operator change but all the other Clippy violations can be fixed as it is good for the code to not have Clippy warnings.

JohnTitor:
I didn't mean it was the only concern, other parts make it easier to compare with the C code. I'm seeing you're sending Clippy fix PRs to random repos, please refrain from such a PR.

PR #2430, opened 2021-10-03
linux musl bring ucontext api for related archs

devnexen:


kaniini:
This pull request is wrong.  The `getcontext` et al API are not provided by musl directly, but by my libucontext library: https://gitea.treehouse.systems/ariadne/libucontext

PR #2435, opened 2021-10-06
add `devname_r` to freebsd and dragonfly

i509VCB:
## Rationale
I need this function specifically to re-implement some parts of `libdrm` in Rust for FreeBSD: https://cgit.freedesktop.org/drm/libdrm/tree/xf86drm.c#n905 Just so happened that dragonfly also has `devname_r` so that is included.


JohnTitor (approved):
Thanks!

PR #2440, opened 2021-10-10
add PT_* ELF segment type const definitions for android

name1e5s:


src/unix/linux_like/android/mod.rs
```diff
@@ -1867,6 +1867,26 @@ pub const MFD_CLOEXEC: ::c_uint = 0x0001;
...
+// so we can use that type here to avoid having to cast.
+pub const PT_NULL: u32 = 0;
+pub const PT_LOAD: u32 = 1;
+pub const PT_DYNAMIC: u32 = 2;
+pub const PT_INTERP: u32 = 3;
+pub const PT_NOTE: u32 = 4;
+pub const PT_SHLIB: u32 = 5;
+pub const PT_PHDR: u32 = 6;
+pub const PT_TLS: u32 = 7;
+pub const PT_NUM: u32 = 8;
```
JohnTitor, line 1881: Is this declared actually? I cannot find it on the source.

name1e5s: Seems that `PT_NUM ` is defined as "Number of defined types" on linux, but android didn't define this const, I will remove the definition now.

JohnTitor:
Thanks! @bors r+

PR #2441, opened 2021-10-11
Add Apple RTF_DEAD and RTF_GLOBAL

tklauser:
RTF_DEAD was added in macOS 10.13.
RTF_GLOBAL was added in macOS 11.0.


JohnTitor (approved):
Thanks!

PR #2444, opened 2021-10-12
Define constants for CAN RAW socket options

djarek:
Add setsockopt constants from `linux/can/raw.h`, which allow configuring CAN_RAW sockets.

Signed-off-by: Damian Jarek <damian.jarek93@gmail.com>


JohnTitor:
Thanks! @bors r+

PR #2448, opened 2021-10-14
Android supports EPOLLEXCLUSIVE, clean up EPOLL* flags

rtzoeller:
Add `EPOLLEXCLUSIVE` to Android, with the same value as on Linux.

Consolidate `EPOLL*` flags which are common to all `linux_like/*/mod.rs` files to `linux_like/mod.rs`, including listing them in ascending order which is how Linux defines them.


rtzoeller:
@JohnTitor it looks like the macOS semver check failed, but it's pretty clearly unrelated to these changes. I'm not sure why that would be.

JohnTitor:
It's due to a previous PR and won't block bors, don't worry.

PR #2449, opened 2021-10-15
Add svm_flags field to sockaddr_vm

rtzoeller:
Testing `libc` on Fedora 34 emits the following error:

```
bad field offset svm_zero of sockaddr_vm: rust: 12 (0xc) != c 13 (0xd)
bad field size svm_zero of sockaddr_vm: rust: 4 (0x4) != c 3 (0x3)
bad field type svm_zero of sockaddr_vm: rust: 0x7ffe56b1165c != c 0x7ffe56b1165d
```

Kernel version:

```
$ uname -r
5.14.9-200.fc34.x86_64
```


JohnTitor:
Adding a new public feild to a struct is a breaking change and we should tell users that we're going to do it, otherwise user code would be broken suddenly.

james-callahan:
Has there been any followup to this?
Without the `svm_flags` field you can no longer use e.g. nitro enclaves on AWS.

See also https://github.com/rust-vsock/vsock-rs/issues/29

PR #2455, opened 2021-10-15
Adds pthread_mutexattr_[g|s]etrobust and pthread_mutex_consistent bindings

awidegreen:
Adds pthread_mutexattr_[g|s]etrobust and pthread_mutex_consistent bindings

FreeBSD:
https://cgit.freebsd.org/src/tree/include/pthread.h?id=65436b2e1207a98a1c752c14f8c059238c0eafda#n140

Linux:
https://sourceware.org/git?p=glibc.git;a=blob;f=sysdeps/htl/bits/pthreadtypes.h;h=74127aea488a69af8fb63b8f353307e5d401a62b;hb=HEAD#l83

### DIFF
--- a/libc-test/semver/freebsd.txt
+++ b/libc-test/semver/freebsd.txt
@@ -1650,9 +1650,12 @@ pthread_getcpuclockid
 pthread_getthreadid_np
 pthread_kill
 pthread_main_np
+pthread_mutex_consistent
 pthread_mutex_timedlock
 pthread_mutexattr_getpshared
 pthread_mutexattr_setpshared
+pthread_mutexattr_getrobust
+pthread_mutexattr_setrobust
 pthread_rwlockattr_getpshared
 pthread_rwlockattr_setpshared
 pthread_setaffinity_np
--- a/libc-test/semver/linux.txt
+++ b/libc-test/semver/linux.txt
@@ -2897,9 +2897,12 @@ pthread_getattr_np
 pthread_getcpuclockid
 pthread_getschedparam
 pthread_kill
+pthread_mutex_consistent
 pthread_mutex_timedlock
 pthread_mutexattr_getpshared
 pthread_mutexattr_setpshared
+pthread_mutexattr_getrobust
+pthread_mutexattr_setrobust
 pthread_rwlockattr_setpshared
 pthread_setaffinity_np
 pthread_setschedparam
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -609,6 +609,8 @@ pub const EXTATTR_NAMESPACE_SYSTEM: ::c_int = 2;
 
 pub const PTHREAD_STACK_MIN: ::size_t = MINSIGSTKSZ;
 pub const PTHREAD_MUTEX_ADAPTIVE_NP: ::c_int = 4;
+pub const PTHREAD_MUTEX_STALLED: ::c_int = 0;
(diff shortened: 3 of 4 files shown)


src/unix/bsd/freebsdlike/mod.rs
```diff
@@ -1156,6 +1156,8 @@ pub const PTHREAD_MUTEX_ERRORCHECK: ::c_int = 1;
 pub const PTHREAD_MUTEX_RECURSIVE: ::c_int = 2;
 pub const PTHREAD_MUTEX_NORMAL: ::c_int = 3;
 pub const PTHREAD_MUTEX_DEFAULT: ::c_int = PTHREAD_MUTEX_ERRORCHECK;
+pub const PTHREAD_MUTEX_STALLED: ::c_int = 0;
+pub const PTHREAD_MUTEX_ROBUST: ::c_int = 1;
```
JohnTitor, line 1160: Are those actually declared on DragonFlyBSD? I couldn't find any references.

src/unix/bsd/freebsdlike/mod.rs
```diff
@@ -1549,6 +1551,12 @@ extern "C" {
         attr: *mut pthread_mutexattr_t,
         pshared: ::c_int,
     ) -> ::c_int;
+    pub fn pthread_mutexattr_getrobust(
+        attr: *const pthread_mutexattr_t,
+        robust: *mut ::c_int,
```
JohnTitor, line 1556: `attr` isn't `const` on FreeBSD: https://github.com/freebsd/freebsd-src/blob/9bb8a4091c4f63dacda5108f4f994f7f6b35bbf1/include/pthread.h#L323-L324

JohnTitor (changes requested):
Also, we need to fix the style check failure.

src/unix/bsd/freebsdlike/mod.rs
awidegreen: they are not, I moved it to FreeBSD instead.

src/unix/bsd/freebsdlike/mod.rs
awidegreen: fixed

src/unix/linux_like/mod.rs
```diff
@@ -1600,6 +1600,10 @@ extern "C" {
         attr: *mut pthread_mutexattr_t,
         pshared: ::c_int,
     ) -> ::c_int;
+    pub fn pthread_mutexattr_setrobust(
```
awidegreen, line 1603: @JohnTitor I'm uncertain if that should really be here. I followed the definition of `pthread_mutexattr_*pshared` which are split like that as well:
* `pthread_mutexattr_setpshared` in `src/unix/linux_like/mod.rs`
* `pthread_mutexattr_getpshared` in `src/unix/linux_like/linux/mod.rs`

JohnTitor: It should available on Android, Emscripten, and Linux, right? Then it's fine. It's up to the contributors to decide which module to place it in and we put it where we need.

awidegreen: Right. It seems that Android [doesn't have](https://android.googlesource.com/platform/external/pthreads/+/refs/heads/master/pthread.h) support for those functions, Empscripten [has](https://github.com/emscripten-core/emscripten/blob/f46fbc5a1eb5dc878c3bdd5a505fc17142875961/system/lib/libc/musl/include/pthread.h). I'll will move things around and add the definitions for Emscripten as well.

JohnTitor:
Thanks! @bors r+

PR #2456, opened 2021-10-15
Upgrade macOS on CI to 11 (Big Sur)

JohnTitor:
Let's see if we can use it.

r? @ghost


devnexen:
yes and I use the 11.6 release version.

JohnTitor:
Okay, now CI is green :) @bors r+

PR #2457, opened 2021-10-15
windows deprecate CONTEXT and related types.

devnexen:


src/windows/msvc/x86_64/align.rs
```diff
@@ -40,12 +41,16 @@ cfg_if! {
 }
 
 s! {
+    #[doc(hidden)]
+    #[deprecated(since = "0.2.104", note = "better using winapi crate")]
```
JohnTitor, line 45: It's a bit longer but something like this is clearer to me.

```suggestion
    #[deprecated(since = "0.2.104", note = "use `winapi` crate instead; we're going to remove it in a future release")]
```

Issue #547, opened 2017-03-03
Rust Roadmap: libc v1.0 release


coder543:
- According to [the roadmap](https://blog.rust-lang.org/2017/02/06/roadmap.html), Rust [should have 1.0-level crates for essential tasks](https://github.com/rust-lang/rust-roadmap/issues/11).
- [According to crates.io](https://crates.io/crates?sort=downloads), `libc` is the most downloaded crate of all time.

Therefore, it follows logically that libc must have a v1.0 release according to the Rust roadmap.

I am opening this issue so that a plan can be formulated and then executed upon.

- What would it take to get `libc` to v1.0?
    - I see some issues marked as `1.0-breakage`, so those would need to be done, but it looks like people are holding off on doing them until closer to the currently-nebulous release date for v1.0.
- When should a v1.0 release happen?
    - We could wait until the end of the year, or we could go ahead and make good on this portion of the roadmap, starting with `libc`.

lambda:
Aside from the issues tagged [`1.0-breakage`](https://github.com/rust-lang/libc/issues?q=is%3Aissue+is%3Aopen+label%3A1.0-breakage), which seem simple enough to deal with, I think that the big outstanding question is  #180, that `libc`'s `c_void` is not the same type as `std`'s `c_void` (also filed as rust-lang/rust#31536). There's been [some substantial discussion of this on internals](https://internals.rust-lang.org/t/solve-std-os-raw-c-void/3268), and there is an [active RFC with considerably more discussion](https://github.com/rust-lang/rfcs/pull/1783).

SimonSapin:
I’m very opposed to making a 1.0 release just because. Upgrading from 0.1 to 0.2 was a huge pain. In Servo we had to coordinate upgrades of 52 dependencies: https://github.com/servo/servo/issues/8608. Upgrading only some dependency but not others would cause build errors because they typically use some libc type (like `c_void`) in their public API.

And wasn’t just Servo. crates.io lists 1210 reverse dependencies for libc now.

I think that a much better justification than just getting a nicer-looking version number should be needed to force the ecosystem to go through that again.

coder543:
> I'm very opposed to making a 1.0 release just because.

It's on the official Rust Roadmap. It's very hard to argue that `libc` _isn't_ essential. The reasoning is detailed on [the issue here.](https://github.com/rust-lang/rust-roadmap/issues/11) That alone should be reasoning enough to do the upgrade.

[...]

PR #2458, opened 2021-10-16
Add missing OpenBSD sysctl(2) name definitions

n1000:
Add a bunch of missing sysctl(2) related definitions. These definitions expand the number of sysctl parameters that can be queried without using directly hard coded constants.

Tested with libc-test on OpenBSD 6.9.


n1000:
> do you really intented to use all of them ?

That is sort of a hard question to answer.. :)

The motivation for doing this was that I was tinkering on some [sysctl-rs](https://github.com/johalun/sysctl-rs) project for fun (to continue learning Rust). I was looking into what it would take to add OpenBSD support to this existing project.

As part of that I realized that most of the sysctl definitions were missing, and thought to add them.

I'm not even completely sure I would ever finish the sysctl-rs OpenBSD changes, but thought these libc changes may be useful regardless.

However, if there is a big concern on versioning issues in libc, we can drop this PR as well, as I am not requesting it for any serious  reason...

semarie:
Again, I am not against adding things, but it doesn't fit with the way OpenBSD and Rust are working togther.

What I mean is adding things now could means that later it will not be possible to changes things which would be really required for someone else.

The problem with Rust libc is there is no support for versioning: once declared it is expected that the value will exist forever (whereas OpenBSD remove and replace obsolete things). Currently, we are dealing with that with deprecation notice and ignoring values in testsuite, but it is polluting namespace and it adds (small) maintenance overhead. Problems will be bigger when conflictious things will surfaced : coexistence will not be possible without a big overhead (you could look how FreeBSD is dealing with FreeBSD 11 to 12 transition regarding some type definitions).

JohnTitor:
I'm thinking of this and yes, currently, due to the versioning, it's not a good idea to declare stuff that _potentially_ required by someone at once. We could introduce breaking changes incrementing a minor version just like other `0.x.y` crates but it may be painful to users that use unrelated platforms. Thus, I'm going to close this PR for now, but at some point, we should discuss versioning along with #547.

PR #2461, opened 2021-10-17
Fix most libc-test compile warnings and errors for DragonFly 

rtzoeller:
libc-test is failing on my DragonFly BSD system with a large number of errors, most of which are just a result of headers not being included.

```
[root@ ~/libc/libc-test]# uname -a
DragonFly  6.0-RELEASE DragonFly v6.0.1-RELEASE #0: Wed Oct 13 21:06:17 CDT 2021     root@:/usr/obj/usr/src/sys/X86_64_GENERIC  x86_64

[root@ ~/libc/libc-test]# cargo test 2>&1 >/dev/null | wc -l
    6015
```

Including a few additional headers, most of these warnings and errors are fixed:
```
[root@ ~/libc/libc-test]# cargo test 2>&1 >/dev/null | wc -l
    561
```


JohnTitor (approved):
Thanks! Unfortunately, we don't have CI for DragonFlyBSD and it's hard to find currently :(

PR #2462, opened 2021-10-17
Deprecate XU_NGROUPS on DragonFly

rtzoeller:
This constant was only ever defined on FreeBSD.

https://github.com/freebsd/freebsd-src/blob/a7d137fcbcac7182d4fcdc97a46b10edc5c7041d/sys/sys/ucred.h
https://github.com/DragonFlyBSD/DragonFlyBSD/blob/c2e500ae4c06466901674883b4593bd9c249cdc1/sys/sys/ucred.h


JohnTitor:
I'd generally deprecate items first (and then remove them) regardless of the targets as we cannot see the _complete_ number of users depending on it.

JohnTitor (approved):

PR #2468, opened 2021-10-20
FreeBSD: add eventfd (since 13)

unrelentingtech:
This already exposes constants in the non-versioned code like #2465 so this should wait on #2465…

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -1764,8 +1764,13 @@ fn test_freebsd(target: &str) {
         _ => cfg.define("_WANT_FREEBSD11_STAT", None),
     };
 
-    let freebsdlast = match freebsd_ver {
-        Some(12) | Some(13) => true,
+    let freebsd12 = match freebsd_ver {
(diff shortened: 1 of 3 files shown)


libc-test/build.rs
```diff
@@ -2018,6 +2027,9 @@ fn test_freebsd(target: &str) {
             // FIXME: https://github.com/rust-lang/libc/issues/1272
             "execv" | "execve" | "execvp" | "execvpe" | "fexecve" => true,
 
+            // These functions were added in FreeBSD 13:
+            "eventfd" if Some(13) <= freebsd_ver => true,
```
JohnTitor, line 2031: I think the condition here is wrong.

JohnTitor:
I think this is now good to go, is there any reason to be WIP?

unrelentingtech:
Yeah, should be fine

JohnTitor (approved):

JohnTitor:
Okay, thanks! @bors r+

PR #2477, opened 2021-10-24
linux: Add open_how and related flags

Fanael:
This makes `openat2` usable.


JohnTitor:
> 
> 
> NB: `open_how` should probably be non-exhaustive, because future kernel versions may add new members to the structure, to quote the man page:
> 
> > Any future extensions to `openat2`() will be implemented as new fields appended to the above structure, with a zero value in a new field resulting in the kernel behaving as though that extension field was not present.  Therefore, the caller _must_ zero-fill this structure on initialization.
> 
> but `#[non_exhaustive]` is not supported by the MSRV. What's the best way of dealing with that scenario using what we can rely on?

We have a workaround to support a newer feature, see build.rs:
https://github.com/rust-lang/libc/blob/90ebf43bb6524a1febc8b8e14aa42dad1eea6aa6/build.rs#L47-L50

Fanael:
Updated and worked around the CI build failure. I'm not sure how to make style check happy now though.

JohnTitor:
Uhm, I think we can have `non_exhaustive.rs` just like `align.rs`?

Fanael:
Updated.

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #2483, opened 2021-10-27
Move Linux's `POLLRDHUP` into `linux_like` and fix its type.

sunfishcode:
This was originally posted as #2390, but since it was a breaking change,
that PR instead just added `deprecated` warnings. There haven't been any
concerns for a while, so this is now a PR to do the change actually, per
the [breaking-change-policy].

This fixes two errors in #2247.

 - It moves the definitions of `POLLRDHUP` out of `linux_like/linux`
   and into `linux_like`, so that they're available on Android as well.

 - It changes the type from `c_int` to `c_short` to match the other
   `POLL*` flags.

[breaking-change-policy]: https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md#breaking-change-policy


JohnTitor (approved):
This is the first time to do an actual change since we have that policy. Let's see if the period (1-2 months, i.e. 3 releases) is enough.

PR #2487, opened 2021-10-30
Document stat.st_blksize changes on DragonFly BSD

rtzoeller:
The tests are currently failing on DragonFly BSD, due to [changes to `stat.st_blksize`](https://github.com/DragonFlyBSD/DragonFlyBSD/commit/34c6728ee290889398ff2f1305a91f08aa6b0b35) among other reasons.

`st_blksize` is now an `i64`, and has moved to take up a previously reserved struct member. Deprecate the struct, to indicate it will be modified in a subsequent release.

### DIFF
--- a/src/unix/bsd/freebsdlike/dragonfly/mod.rs
+++ b/src/unix/bsd/freebsdlike/dragonfly/mod.rs
@@ -106,6 +106,11 @@ s! {
         pub f_uid_uuid: ::uuid_t,
     }
 
+    #[deprecated(
+        since = "0.2.107",
+        note = "stat.st_blksize is an i64 and stat.st_qspare1 is replaced with \
+                stat.st_blksize in DragonFly 5.8"
+    )]
     pub struct stat {
         pub st_ino: ::ino_t,
         pub st_nlink: ::nlink_t,
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -699,6 +699,7 @@ extern "C" {
         all(target_os = "freebsd", any(freebsd11, freebsd10)),
         link_name = "fstat@FBSD_1.0"
     )]
+    #[cfg_attr(target_os = "dragonfly", allow(deprecated))]
     pub fn fstat(fildes: ::c_int, buf: *mut stat) -> ::c_int;
 
     pub fn mkdir(path: *const c_char, mode: mode_t) -> ::c_int;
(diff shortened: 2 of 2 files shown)


Amanieu:
What is the long-term solution for this? I don't see how the deprecation warning is helpful since there is no action anyone can do in their code to resolve it.

rtzoeller:
@Amanieu in a future release, we should update the struct definition to match DragonFly's. This is just an announcement of that future breaking change. 

My understanding was that this was the correct course of action, per the breaking change policy and previous conversations with @JohnTitor.

If we should handle this differently, let me know and I can update the PR.

Amanieu:
I will let @JohnTitor review this since you have previously discussed it with him.

r? @JohnTitor

src/unix/bsd/freebsdlike/dragonfly/mod.rs
```diff
@@ -106,6 +106,10 @@ s! {
         pub f_uid_uuid: ::uuid_t,
     }
 
+    #[deprecated(
+        since = "0.2.106",
+        note = "stat.st_blksize is an i64 and replaces stat.st_qspare1 in DragonFly 5.8"
```
JohnTitor, line 111: I think something like this is clearer:

```suggestion
        note = "stat.st_blksize is an i64 and stat.st_qspare1 is replaced with stat.st_blksize in DragonFly 5.8"
```

src/unix/bsd/freebsdlike/dragonfly/mod.rs
```diff
@@ -106,6 +106,10 @@ s! {
         pub f_uid_uuid: ::uuid_t,
     }
 
+    #[deprecated(
+        since = "0.2.106",
```
JohnTitor, line 110:
```suggestion
        since = "0.2.107",
```

JohnTitor (changes requested):
Yeah, let's deprecate first so that users can notice the change and prepare their own struct if they want to use an older one.
Left some comments about version and note.

JohnTitor:
Thanks! @bors r+

JohnTitor:
CI should be fixed now, @bors retry

JohnTitor:
Some functions need `#[allow(deprecated)]`.

rtzoeller:
@JohnTitor fixed! For some reason I couldn't see the actual error in the log on GitHub, and had to download it and view it locally.

JohnTitor:
> For some reason I couldn't see the actual error in the log on GitHub, and had to download it and view it locally.

Hm, it's displayed fine on my env, maybe loading the log failed at that time?

Anyway, the build should be fixed, @bors r+

PR #2488, opened 2021-10-30
Deprecate FIODGNAME on DragonFly, add FIODNAME

rtzoeller:
`FIODGNAME` is FreeBSD specific, but `FIODNAME` exists on DragonFly which appears to provide similar functionality.

[FreeBSD definition](https://github.com/freebsd/freebsd-src/blob/dc6dd769de63c4eceb8899205a5d780d9f278fd2/sys/sys/filio.h#L60)
[DragonFly definition](https://github.com/DragonFlyBSD/DragonFlyBSD/blob/20f6ddd0df90767e1eba2d12dfa8e1769be7cec7/sys/sys/filio.h#L59)

`FIODGNAME` is appearing in the `libc-test` failures on DragonFly. Deprecate it, to be removed in a subsequent release.


JohnTitor (approved):

JohnTitor:
Thanks for correcting it! @bors r+

PR #2489, opened 2021-10-30
DragonFly's devname_r takes a size_t len parameter

rtzoeller:
#2435 introduced this function on DragonFly BSD with FreeBSD's signature, which is incorrect. Deprecate it, to be fixed in a subsequent release.

https://github.com/DragonFlyBSD/DragonFlyBSD/blob/c163a4d7ee9c6857ee4e04a3a2cbb50c3de29da1/include/stdlib.h#L285


JohnTitor (approved):

rtzoeller:
Build failures look unrelated.

JohnTitor:
Ah, wait, forgot that the version was updated, @bors r- retry

rtzoeller:
@JohnTitor I've updated the version! Thanks for catching that.

Issue #2484, opened 2021-10-27
libc 0.2.105 broken as libstd dep on macOS


bjorn3:
# Reproduce

Update libc in rust-lang/rust to version 0.2.105. Then try to compile on macOS.

# Errors

```
error[E0412]: cannot find type `Option` in this scope
   --> /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.105/src/unix/bsd/apple/b64/x86_64/mod.rs:113:19
    |
113 |         pub size: Option<unsafe extern "C" fn(
    |                   ^^^^^^ not found in this scope
    |
help: consider importing this enum
    |
178 |         use Option;
    |

error[E0412]: cannot find type `Option` in this scope
   --> /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.105/src/unix/bsd/apple/b64/x86_64/mod.rs:117:21
    |
117 |         pub malloc: Option<unsafe extern "C" fn(
    |                     ^^^^^^ not found in this scope
    |
help: consider importing this enum
    |
178 |         use Option;
    |

error[E0412]: cannot find type `Option` in this scope
   --> /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.105/src/unix/bsd/apple/b64/x86_64/mod.rs:121:21
    |
164 |         pub pressure_relief: Option<unsafe extern "C" fn(
    |                              ^^^^^^ not found in this scope
    |
help: consider importing this enum
    |
178 |         use Option;
    |

error[E0412]: cannot find type `Option` in this scope
   --> /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.105/src/unix/bsd/apple/b64/x86_64/mod.rs:168:30
    |
168 |         pub claimed_address: Option<unsafe extern "C" fn(
    |                              ^^^^^^ not found in this scope
    |
help: consider importing this enum
    |
178 |         use Option;
    |

warning: an associated function with this name may be added to the standard library in the future
 --> compiler-builtins/src/float/pow.rs:8:19
  |
8 |     let mut pow = i32::abs_diff(b, 0);
  |                   ^^^^^^^^^^^^^
  |
  = note: `#[warn(unstable_name_collisions)]` on by default
  = warning: once this associated item is added to the standard library, the ambiguity may cause an error or change in behavior!
[...]

PR #2494, opened 2021-10-30
Release 0.2.106 to incorporate build fix for #2484

joshtriplett:


joshtriplett:
(already tagged and uploaded, since the previous version broke CI on macOS)

PR #2504, opened 2021-11-02
sparc64: Update debian image to 11 (2021-10-20) to fix segfault on CI

JohnTitor:
Fixes #2503
r? @ghost


JohnTitor:
Well, it _does_ fail on nightly-2021-10-31 (https://github.com/rust-lang/libc/runs/4085913022?check_suite_focus=true)? I guess it's caused by a change of rustc but actually not...?

PR #2505, opened 2021-11-03
Add getresgid and getresuid to DragonFly, FreeBSD and OpenBSD

rtzoeller:
The functions have the same signatures as on Linux.


semarie (approved):
LGTM for openbsd part

PR #2508, opened 2021-11-03
Add TCGETS2 and TCSETS2 (and variants) ioctl constants for Linux.

de-vri-es:
I noticed that the `termios2` struct is already exposed, but the ioctl constants to use it are not. This PR adds the `TCGETS2`, `TCSETS2`, `TCSETSW2` and `TCSETSF2` on Linux so that you can actually do something with the `termios2` struct.

The `powerpc` architecture is notably missing, because it does not seem to support the `TCGETS2`/`TCSETS2` ioctls.

I *think* the constants are correct for all platforms, but I'm also not 100% sure. Do the unit tests verify the values for all supported platforms, by any chance?


de-vri-es:
Hmm, the test failure is because the C test doesn't `#include <linux/termios.h>`. However, that would conflict with `#include <termios.h>`. Is there a way to change the include files for the tests for specific symbols?

src/unix/linux_like/linux/gnu/b32/sparc/mod.rs
```diff
@@ -508,6 +508,10 @@ pub const TCGETS: ::c_ulong = 0x40245408;
 pub const TCSETS: ::c_ulong = 0x80245409;
 pub const TCSETSW: ::c_ulong = 0x8024540a;
 pub const TCSETSF: ::c_ulong = 0x8024540b;
+pub const TCGETS2: ::c_ulong = 0x402c542a;
+pub const TCSETS2: ::c_ulong = 0x802c542b;
+pub const TCSETSW2: ::c_ulong = 0x802c542c;
+pub const TCSETSF2: ::c_ulong = 0x802c542d;
```
Amanieu, line 514: Sparc uses a different definition than other platforms:
```c
// sparc
#define TCGETS2		_IOR('T', 12, struct termios2)
#define TCSETS2		_IOW('T', 13, struct termios2)
#define TCSETSW2	_IOW('T', 14, struct termios2)
#define TCSETSF2	_IOW('T', 15, struct termios2)

// everyone else
#define TCGETS2		_IOR('T', 42, struct termios2)
#define TCSETS2		_IOW('T', 43, struct termios2)
#define TCSETSW2	_IOW('T', 44, struct termios2)
#define TCSETSF2	_IOW('T', 45, struct termios2)
```

src/unix/linux_like/linux/gnu/b64/sparc64/mod.rs
```diff
@@ -587,6 +587,10 @@ pub const TCGETS: ::c_ulong = 0x40245408;
 pub const TCSETS: ::c_ulong = 0x80245409;
 pub const TCSETSW: ::c_ulong = 0x8024540a;
 pub const TCSETSF: ::c_ulong = 0x8024540b;
+pub const TCGETS2: ::c_ulong = 0x402c542a;
+pub const TCSETS2: ::c_ulong = 0x802c542b;
+pub const TCSETSW2: ::c_ulong = 0x802c542c;
+pub const TCSETSF2: ::c_ulong = 0x802c542d;
```
Amanieu, line 593: Same with sparc64.

src/unix/linux_like/linux/gnu/b32/sparc/mod.rs
de-vri-es: Good catch. Updated. I think I had the direction bits for `mips` wrong too :(

Updated those now too, so it's `0x40` for read and `0x80` for write.

de-vri-es:
> @bors r+

Oops, I pushed a new commit to fix  `mips` after the `r+`.

de-vri-es:
Ugh, the termios2 struct is 4 bytes larger on mips. Updated the constants again.

Yay for the test suite though <3

Amanieu:
Yea the tests here are a lifesaver.

@bors r+

de-vri-es:
I had a typo in TCSETSF2 for sparc.

Also, I noticed that the constants are all `ulong`, even though ioctl takes an int for the `request` parameter. I *could* turn these new ones from this PR into `uint`, which would have made the typo impossible (I had one too many digit). But I guess that the existing ones should stay `ulong` for backwards compatibility.

Or I can just leave them as `ulong` for consistency with the other constants.

src/unix/linux_like/android/mod.rs
```diff
@@ -1338,6 +1338,10 @@ pub const TCGETS: ::c_int = 0x5401;
 pub const TCSETS: ::c_int = 0x5402;
 pub const TCSETSW: ::c_int = 0x5403;
 pub const TCSETSF: ::c_int = 0x5404;
+pub const TCGETS2: ::c_ulong = 0x802c542a;
+pub const TCSETS2: ::c_ulong = 0x402c542b;
+pub const TCSETSW2: ::c_ulong = 0x402c542c;
+pub const TCSETSF2: ::c_ulong = 0x402c542d;
```
Amanieu, line 1344: These should probably be `c_int` for consistency with the other values in this file.

de-vri-es: Yes, good point. Updated.

de-vri-es:
\o/

Thanks for the reviews!

PR #2511, opened 2021-11-04
Add `procctl` data constants on freebsd

niluxv:
The constants to be passed in the `data` argument of `procctl` on freebsd.

Source for the constants: <https://github.com/freebsd/freebsd-src/blob/main/sys/sys/procctl.h>.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2003,15 +2003,32 @@ fn test_freebsd(target: &str) {
             // Added in FreeBSD 13.0 (r356667)
             "GRND_INSECURE" if Some(13) > freebsd_ver => true,
 
-            // Added in FreeBSD 12.1 (r343964 and r345228)
-            "PROC_ASLR_CTL" | "PROC_ASLR_STATUS" | "PROC_PROCCTL_MD_MIN"
+            // Added in FreeBSD 12.1 (r343964)
+            "PROC_ASLR_CTL"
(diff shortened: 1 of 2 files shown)


niluxv:
Hm, seems that the ASLR and PROTMAX constants weren't there in FreeBSD 11 (and the PROTMAX ones also not in 12). But then `PROC_ASLR_CTL`, `PROC_ASLR_STATUS`, `PROC_PROTMAX_CTL` and `PROC_PROTMAX_STATUS` also weren't there in 11, so why do they not error?

devnexen:
little hint if that helps https://github.com/rust-lang/libc/blob/3bde54ffef10bc05c0f3d6498d5a42eabfa4bee9/libc-test/build.rs#L2007

niluxv:
So IIUC the policy for constants is to just ignore them in tests for older FreeBSD versions (i.e. **not** defining the constants only for the appropriate versions of FreeBSD (`freebsd12` and `freebsd13` dirs), but keeping them globally in `freebsd/mod.rs`)?
Did that in the last commit.

devnexen:
freebsd1*/mod.rs are more likely, for example, for constants which had been updated b/w releases.

niluxv:
Ok. Thanks you!

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #2521, opened 2021-11-07
Add TIOCM_* constants for Illumos and Solaris.

de-vri-es:
Values taken from https://github.com/illumos/illumos-gate/blob/252adeb303174e992b64771bf9639e63a4d55418/usr/src/uts/common/sys/termios.h#L466-L476

Since Illumos forked from OpenSolaris 10, I'm assuming that Solaris is still using the same values for binary compatibility. But since Solaris is no longer open source, I couldn't find a way to verify it myself.


JohnTitor (approved):
Let's assume so :)

PR #2522, opened 2021-11-07
Fix/remove deprecated DragonFly items

rtzoeller:
These items were recently deprecated on DragonFly, either because the platform does not define them or because they were out of date.


rtzoeller:
I'm not sure how long the maintainers prefer to wait after deprecating items before fixing them, but I thought I'd put a PR up now and let the maintainers merge it when they felt it was appropriate.

My hope is that it can be merged without waiting too many releases, since DragonFly is a tier 3 target for rust.

JohnTitor:
I think about a month is enough, maybe? Looking at the download stats on https://crates.io/crates/libc, some of the active users have updated but right now is a bit early, I feel.

rtzoeller:
@JohnTitor it's been a little under a month (~4 weeks), but based on the download stats it looks like we're likely reaching the point of diminishing returns. I don't see any open PRs calling for a new release (suggesting the next one is at least a few days away); are we at a point where this can be merged?

JohnTitor:
Okay, let's assume most active users should've got the deprecation note, @bors r+

JohnTitor (approved):

PR #2528, opened 2021-11-11
Scope in Send, Sync for wasi

workingjubilee:
Without this, libc fails during the build of rustc. See https://github.com/rust-lang/rust/pull/90681
This dependency was introduced in rust-lang/libc#2499.
Alternative solution: It's not clear this is necessary for core to finish building,
so maybe we should `#[cfg]` it out in that case, instead?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,7 +63,7 @@ cfg_if! {
         use core::clone::Clone;
         #[doc(hidden)]
         #[allow(unused_imports)]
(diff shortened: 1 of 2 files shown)


workingjubilee:
Oh I should have probably asked for... r? JohnTitor

workingjubilee:
Oh I should have probably asked for... r? @JohnTitor

JohnTitor:
Timeout, @bors retry

PR #2530, opened 2021-11-11
Move more ioctl constants and flags to linux::arch::* modules.

de-vri-es:
This PR moves a few things to `linux::arch::*` modules and out of libc specific modules. Specifically, it:
* Moves the `TIOCM*` ioclt constants and associated flags.
* Moves the `BOTHER` flag.
* Adds the `IBSHIFT` flag.

Without this PR, there are some arch+libc combinations that are missing some of these constants. Since these are all ioctl numbers or arguments to ioctls, they are not specific to any libc flavour. `BOTHER` and `IBSHIFT` don't even appear in glibc or musl headers.


JohnTitor (approved):
Sweet, thanks!

JohnTitor:
@de-vri-es According to https://github.com/rust-lang/libc/runs/4218181867?check_suite_focus=true, this changed some consts' type, it's a breaking change. If that change is necessary, I'd like to deprecate them first then do the actual change in favor of https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md#breaking-change-policy, could you revert the breaking change part?

de-vri-es:
Ah I didn't notice that aarch64-musl has different types for the ioctl constants compared to the glibc targets D:

However, reverting that would mean putting everything back in libc+arch specific modules. Or alternatively, I could add a `pub(crate) type Ioctl` definition to the libc+arch specific modules type and use that from the arch specific modules.

What do you think is best?

JohnTitor:
Sounds fine, one problem is that older Rust doesn't support `pub(crate)` unfortunately so we could add a doc comment instead.

de-vri-es:
Alright, sounds good. I think I'll have time for that tomorrow. If this is blocking a release it might be best to simply revert the whole PR until I can put the fix in.

de-vri-es:
Fix proposed in #2550

PR #2535, opened 2021-11-15
solarish dl_iterate_phdr api addition

devnexen:


pfmooney:
When rebased and de-conflicted with the current (and soon-to-be-merged) illumos bits, this causes the tests to fail.

PR #2549, opened 2021-11-17
Add more items for FreeBSD

GuillaumeGomez:


GuillaumeGomez:
This is now ready!

JohnTitor:
It'd be great if you could also rebase this to test with a new FreeBSD 14 CI, as it currently doesn't gate bors (I guess #2545 will be merged in an hour).

GuillaumeGomez:
Sure! Please ping me then so I can update my PR.

JohnTitor:
@GuillaumeGomez It has been merged now!

GuillaumeGomez:
CI is happy! \o/

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -856,13 +866,75 @@ pub const CTL_HW: ::c_int = 6;
...
+
 pub const CTL_SYSCTL_DEBUG: ::c_int = 0;
 pub const CTL_SYSCTL_NAME: ::c_int = 1;
 pub const CTL_SYSCTL_NEXT: ::c_int = 2;
 pub const CTL_SYSCTL_NAME2OID: ::c_int = 3;
 pub const CTL_SYSCTL_OIDFMT: ::c_int = 4;
 pub const CTL_SYSCTL_OIDDESCR: ::c_int = 5;
 pub const CTL_SYSCTL_OIDLABEL: ::c_int = 6;
+cfg_if! {
+    if #[cfg(any(freebsd13, freebsd14))] {
```
JohnTitor, line 933: Let's avoid using `cfg(freebsdN)` here, it makes it hard to maintain FreeBSD N modules when removing/adding a module (that's why I moved `kinfo_proc`).

GuillaumeGomez: The problem is that there isn't a way to say "all freebsd version starting X".

GuillaumeGomez: Also the problem with `kinfo_proc` was different: it was included in all version by default, in here it's only included in freebsd13 and freebsd14. If freebsd15 is released, it won't be impacted.

JohnTitor: I meant, if we add FreeBSD 15, for instance, we will have to tweak each cfgs, it'll bring more tasks. But we declare them on FreeBSD 13 and FreeBSD 14, we'll just have to copy the FreeBSD 14 module, and if something is changed, we need to change them only.
However, I just remembered #2465, TL;DR newly added consts don't have to be gated and we can just declare them for simplicity.

GuillaumeGomez: Oh nice! Updating then.

GuillaumeGomez:
Simplified consts handling as suggested. :)

JohnTitor:
Hm, seems there are still cfg'd consts, is there any reason that they should be?

GuillaumeGomez:
Definitely not, my bad, completely missed them... Updating.

GuillaumeGomez:
Done!

JohnTitor (approved):

JohnTitor:
Great, thanks! @bors r+

PR #2550, opened 2021-11-17
Use libc specific type for architecture specific ioctl defines on Linux.

de-vri-es:
This PR should fix the type change of some ioctl constants on Linux introduced by #2530.

It does this by adding a `#[doc(hidden)]` type called `Ioctl`, which is defined in libc specific modules and used in arch specific modules.

However, when doing this I noticed that when I added `TCGETS2`, `TCSETS2`, ... in #2508, I unconditionally used `c_ulong`. This is inconsistent with the other ioctl constants for `musl` and `uclibc`. This PR also changes those to use the libc specific types. However, PR #2508 has already been released in 0.2.107, so technically that is also a semver incompatible change. The impact is limited to new constants introduced in the last release, and only on `musl` and `uclibc` targets.

So what is more important here? Consistency in the type of ioctl constants, or being very strict with backwards compatibility? If it is the latter, I'll revert `TCGETS2` etc to `c_ulong`  for this PR.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2740,6 +2740,9 @@ fn test_linux(target: &str) {
             | "Elf64_Shdr" | "Elf32_Sym" | "Elf64_Sym" | "Elf32_Ehdr" | "Elf64_Ehdr"
             | "Elf32_Chdr" | "Elf64_Chdr" => ty.to_string(),
 
+            "Ioctl" if gnu => "unsigned long".to_string(),
+            "Ioctl" => "int".to_string(),
+
             t if is_union => format!("union {}", t),
 
             t if t.ends_with("_t") => t.to_string(),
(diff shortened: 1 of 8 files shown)


de-vri-es:
Adjusted the definition now to be `pub(crate)` when `cfg(doc)` is enabled. This ensures that the documentation says `c_ulong` /`c_int` instead of `Ioctl` with a broken link.

If I remember correctly, `cfg(doc)` is newer than `pub(crate)`, so this shouldn't even break doc generation on older versions of rust.

JohnTitor (approved):
The latest version was released 12 days ago, we could assume the impact of breaking changes is quite small (, or even _nothing_ if you don't depend on that). I think it's a reasonable trade-off so am going to r+. Thanks for the nice cleanup!

PR #2557, opened 2021-11-21
openbsd add ktrace fn

devnexen:


semarie:
ktrace(2) interface is very specific to OpenBSD and strongly linked to OS internals. what do you want to do with such function and/or struct definition ? outside rewriting ktrace(8) or kdump(8) system utilities, I don't see real usage here.

semarie:
for reference, ktrace(2) is a syscall to enable tracing at kernel level of process execution. The trace is saved in a file.

devnexen:
> outside rewriting ktrace(8) or kdump(8) system utilities, I don't see real usage here.

that s a bit the idea even tough I am not the one willing to.

semarie:
do you have more information to share on this rewrite and the purpose of it ? I doubt it would be accepted to be merged upstream, too many requirements would be missing for importing Rust code for now.

I don't want to be the bad guy here, but for now I don't see the point of importing such interface in libc crate. but I agree it could belongs to it. No strong opposition for importing it, but due to the "intented for internal use status" of this interface, any change on it has high probability to be a breaking change, and will forbid the update on Rust side.

devnexen:
yes that would be the first batch, but no offence if this is not merged tough.

JohnTitor:
I'm not a maintainer of FreeBSD nor familiar with it so going to think of it with @semarie's comments. So, right now, the crate leans against stability, and here isn't the best place to declare such an internal API, I guess. I think we should have another crate specifically designed for it, like `mach2`, how about?

PR #2565, opened 2021-11-22
Add devstat items

GuillaumeGomez:


src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -380,6 +503,97 @@ s! {
...
+        pub generation: ::c_long,
+        pub numdevs: ::c_int,
+    }
+
+    struct statinfo {
+        pub cp_time: [::c_long; CPUSTATES],
+        pub tk_nin: ::c_long,
+        pub tk_nout: ::c_long,
+        pub dinfo: *mut devinfo,
+        pub snap_time: ::c_double,
```
GuillaumeGomez, line 595: To be noted: this is supposed to be a `long double`, but I have absolutely no idea how to bind such a type... And the C spec is wonderful about it: https://gcc.gnu.org/onlinedocs/gcc/Floating-Types.html

GuillaumeGomez:
I'm having trouble with errors like:

```
error: incompatible pointer types returning 'devstat_type_flags *' from a function with result type 'struct devstat_type_flags *'
```

They are C enums but the comparison considers them as structs? Just in case, the C enums are defined [here](https://github.com/freebsd/freebsd-src/blob/ee7485bef53df9108a20e497dac6341d030248ce/sys/sys/devicestat.h#L65-L132).

GuillaumeGomez:
@JohnTitor Any idea about why I'm getting this error when using enums?

GuillaumeGomez:
So about the enum/struct comparison, I think `ctest` is somehow considering either the C enum as a struct or the rust enum as a struct. I'll try to check if the issue comes from there.

GuillaumeGomez:
CI passed! \o/

Whoever will review this, please check very carefully if the `c_longdouble` implementation is acceptable. The goal here is simply to have the size right (because it's used as a struct field). It's not meant to be used as is.

GuillaumeGomez:
Fixed the conflict.

GuillaumeGomez:
Rebased.

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -41,10 +41,214 @@ pub type caddr_t = *mut ::c_char;
...
 
+// This is mosty a hack to keep the length somewhat similar. However, based on
+// https://gcc.gnu.org/onlinedocs/gcc/Floating-Types.html, it might very well be incorrect.
+cfg_if! {
+    if #[cfg(any(target_pointer_width = "16", target_pointer_width = "32"))] {
+        pub type c_longdouble = [u8; 8];
+    } else {
+        pub type c_longdouble = [u8; 16];
+    }
+}
```
Amanieu, line 52: Have you checked that this actually works? I imagine that from a calling-convention point of view, a `long double` is passed/returned differently from an array of `u8`.

GuillaumeGomez: Do you know how to check it? I tried it on the C API and valgrind seemed fine with it but no clue if it isn't just plain luck at this point...

Amanieu: I did a few tests on godbolt:
- On x86 & x86_64, long doubles are passed in 80-bit x87 float registers. There is no way to represent this with normal Rust code.
- On aarch64, long doubles are passed in 128-bit SIMD registers. You can probably emulate this by using `uint64x2_t` from the (currently unstable) stdarch intrinsics.

In conclusion, I don't think it is practical to bind these API using Rust at the moment. I would recommend using a wrapper written in C to convert these values to a normal `double`.

In the longer term, I plan on adding arch-specific types such as `f80`, `f128`, `f16`, etc to stdarch, but this is unlikely to happen any time soon.

GuillaumeGomez: Would it make sense to make it private and remove the double functions using `long double` and keeping it for the `statinfo` struct?

GuillaumeGomez: We only test FreeBSD on one arch? :o

Amanieu: We only test it on x86_64. But we do test that it builds on x86_64, i686, aarch64 and powerpc64.

GuillaumeGomez: I'm not sure what you mean: are the types actually checked on the arch you mentioned or is it simply running a `cargo build`?

Amanieu: The types are only checked on x86_64 I think.

GuillaumeGomez: Oh, that's pretty bad. :-/

GuillaumeGomez:
@Amanieu In the end, I simply commented everything related to `long double`. For the time being, I think it's the easiest solution unfortunately.

Amanieu:
I would prefer if the code was properly deleted, it's messy to leave commented code around. Also, the build.rs changes related to long double should also be removed.

GuillaumeGomez:
Ok, done as well.

libc-test/build.rs
```diff
@@ -2206,6 +2220,10 @@ fn test_freebsd(target: &str) {
             // https://github.com/gnzlbg/ctest/issues/68
             "lio_listio" => true,
 
+            // It returns a `long double`, but it's a nightmare to bind correctly in rust
+            // for the moment, so it's a best effort thing...
+            "devstat_compute_etime" => true,
```
Amanieu, line 2225: This one too.

GuillaumeGomez: Sorry, good catch!

lnicola:
Hello! As you might know, `rust-analyzer` is included as a submodule in `rust-lang/rust`.

We generally try to keep our dependencies up to date, and one of those is `libc`, so we often use a newer version than `rustc` itself. This isn't a problem, but sometimes `libc` implements new APIs for FreeBSD and requires linking to extra libraries, which are not available in the `rust-lang/rust` CI image. I've seen this happen for `libkvm`, `libprocstat` and `libdevstat`, see for example https://github.com/rust-lang/rust/pull/91909.

Adding new APIs is fine, of course, but would you mind pinging me next time so I can update the FreeBSD toolchain script when bumping RA to a newer version? Thanks!

PR #2566, opened 2021-11-22
memfd_create is a glibc API, which is not present on uclibc

skrap:
I think the title says it all!  The `armv7-unknown-linux-uclibceabihf` tier-3 target does not build without this change, and I suspect all other uclibc builds are broken as well.

It's a good reminder to try to add some CI validation of the uclibc build, which I will try to do in a followup PR.

To test this, get and unpack https://toolchains.bootlin.com/downloads/releases/toolchains/armv7-eabihf/tarballs/armv7-eabihf--uclibc--bleeding-edge-2020.08-1.tar.bz2

Note instances of `PATH_HERE` below which you'll have to customize for your system.

config.toml:
```
[target.armv7-unknown-linux-uclibceabihf]
linker = "/PATH_HERE/buildroot-armv7/output/host/usr/bin/arm-buildroot-linux-uclibcgnueabihf-gcc"
runner = "qemu-arm -L /PATH_HERE/armv7-eabihf--uclibc--bleeding-edge-2020.08-1/arm-buildroot-linux-uclibcgnueabihf/sysroot/"
```

Then run:
```
cd libc-test
CC_armv7_unknown_linux_uclibceabihf=/PATH_HERE/armv7-eabihf--uclibc--bleeding-edge-2020.08-1/bin/arm-buildroot-linux-uclibcgnueabihf-gcc cargo +nightly test -Zbuild-std --target armv7-unknown-linux-uclibceabihf
```

(if there's a way to include that `CC_armv7_etc` cross CC in the config.toml, I'm all ears!)

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -139,6 +139,7 @@ RUST_NIGHTLY_LINUX_TARGETS="\
 aarch64-fuchsia \
 armv5te-unknown-linux-gnueabi \
 armv5te-unknown-linux-musleabi \
+armv7-unknown-linux-uclibceabihf \
 i686-pc-windows-gnu \
 riscv64gc-unknown-linux-gnu \
 wasm32-wasi \
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -139,7 +139,6 @@ RUST_NIGHTLY_LINUX_TARGETS="\
 aarch64-fuchsia \
 armv5te-unknown-linux-gnueabi \
 armv5te-unknown-linux-musleabi \
-armv7-unknown-linux-uclibceabihf \
 i686-pc-windows-gnu \
 riscv64gc-unknown-linux-gnu \
 wasm32-wasi \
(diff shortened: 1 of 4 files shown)


ci/build.sh
```diff
@@ -139,6 +139,7 @@ RUST_NIGHTLY_LINUX_TARGETS="\
 aarch64-fuchsia \
 armv5te-unknown-linux-gnueabi \
 armv5te-unknown-linux-musleabi \
+armv7-unknown-linux-uclibceabihf \
```
skrap, line 142: Is this the right way to enable CI testing for this target?  I couldn't really untangle how this was meant to work.

Amanieu: This actually needs to go into `RUST_LINUX_NO_CORE_TARGETS` instead because the uclibc targets are not distributed via rustup and need to be built with `-Zbuild-std`.

skrap: Fixed.  I added a comment to help the next person.  I tried to test this via `ci/run-docker.sh armv7-unknown-linux-uclibceabihf`

But I get the following error: `ci/run.sh: 112: cargo: not found`

I suspect there's something messed up with my setup because the same issue occurs for `x86_64-unknown-linux-gnu`.

skrap: Oh I found it - i had a toolchain override in place to my stage2 custom build which did not include cargo.

PR #2567, opened 2021-11-23
Remove unwanted #[doc(hidden)] and deprecated attributes on some FreeBSD constants

GuillaumeGomez:
Based on FreeBSD [source code](https://github.com/freebsd/freebsd-src/blob/main/sys/net/if.h#L142-L174), it is definitely not valid.

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -1962,11 +1962,6 @@ pub const IFF_POINTOPOINT: ::c_int = 0x10;
 pub const IFF_KNOWSEPOCH: ::c_int = 0x20;
 /// (d) resources allocated
 pub const IFF_RUNNING: ::c_int = 0x40;
-#[doc(hidden)]
-#[deprecated(
-    since = "0.2.54",
-    note = "IFF_DRV_RUNNING is deprecated. Use the portable IFF_RUNNING instead"
-)]
 /// (d) resources allocate
 pub const IFF_DRV_RUNNING: ::c_int = 0x40;
 /// (n) no address resolution protocol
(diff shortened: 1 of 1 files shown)


JohnTitor:
@asomers You're the one that originally approved these deprecations, is this correct?

asomers-ax:
Yes, and I still think it's appropriate to deprecate them.  An academic distinction perhaps, because the two pairs of #defines are identical.  But the non-deprecated versions are portable, whereas the deprecated ones aren't.  Even within FreeBSD itself, non-kernel code uses `IFF_RUNNING` and `IFF_OACTIVE`

GuillaumeGomez:
The comment very specifically says (for `IFF_RUNNING` and `IFF_OACTIVE`):

> Old names for driver flags so that user space tools can continue to use the old (portable) names.

So at the minimum, it could be the opposite, but certainly not the current situation.

asomers-ax:
The new names were only introduced for in-kernel drivers.  userland never switched to them.  IMHO Rust's libc should never have used the new, non-portable names since for userland code they have no advantage over the portable ones.

GuillaumeGomez:
They exist, and freebsd source code is saying that they should be preferred, so I think it's definitely not up to us to tell libc's users that they're deprecated when they're obviously not.

JohnTitor:
What's the current status of this PR? I haven't used FreeBSD and cannot determine whether it should be used from the libc crate or not... At least the discussion seems to fail to reach an agreement, the point here is if there's a use of these items in a userland, right?

GuillaumeGomez:
Based on freebsd source code, they shouldn't be marked as deprecated. ^^'

JohnTitor:
Re-visiting. Reading https://github.com/freebsd/freebsd-src/commit/292ee7be1c404a624dda68ddff1bd988a1851e5a, I agree with @asomers-ax that they should be deprecated. If I understand correctly, they were renamed only for kernel/device driver development, and any user applications can use the old names without any issues. And exposing `IFF_DRV_*` without any information would cause some misuse, I guess.

> and freebsd source code is saying that they should be preferred

@GuillaumeGomez I'm not sure what part is saying that, do you mean "... if you're in kernel"?

> I think it's definitely not up to us to tell libc's users that they're deprecated when they're obviously not.

Exposing `IFF_DRV_*` without any information would cause some misuse, I guess. In light of that, deprecating seems one of the good options to me.

tgross35:
Since this has been inactive for a while and the target maintainer indicated a preference to leave these deprecated, I'm going to close this. The above linked commit message makes it sound like the `_DRV_` defines are intended for driver use, we probably don't gain much by exposing them in userspace when the current codebase seems to express intent in keeping the portable versions around https://github.com/freebsd/freebsd-src/blob/87e87fecb139d60a00837f004de0d37dbecd205c/sys/net/if.h#L166-L173.

Cc @asomers in case anything changed.

PR #2569, opened 2021-11-25
linux/android add ptracec_peeksiginfo_args struct.

devnexen:


Amanieu:
Musl headers are incompatible with the kernel headers: you can't include both musl's `signal.h` and the kernel's `ptrace.h` in the same C file since they both define the same structs.

devnexen:
finally :-) thanks @Amanieu for your patience.

PR #2585, opened 2021-12-10
Revert "Fix invalid freebsd version selection"

asomers:
This reverts commit 53e79b886bf4e4d3682ae6669e3585b3586e5255.

PR #2581 unintentionally changed libc's bindings from the FreeBSD 11 ABI
to the native ABI of the build host.  This is causing breakage for many
downstream crates.


Amanieu:
@bors r+

I've yanked libc 0.1.110 which has the issue.

PR #2589, opened 2021-12-12
freebsd proposal to move sockcred2 usage in the root mod while ignori…

devnexen:
…ng it for FreeBSD 12.

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -945,6 +945,17 @@ s! {
         pub generation: ::c_long,
         pub numdevs: ::c_int,
     }
+
+    pub struct sockcred2 {
+        pub sc_version: ::c_int,
+        pub sc_pid: ::pid_t,
+        pub sc_uid: ::uid_t,
+        pub sc_euid: ::uid_t,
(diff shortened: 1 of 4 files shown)


Amanieu:
cc @asomers @GuillaumeGomez

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -3701,6 +3712,15 @@ f! {
         let (idx, offset) = (cpu / bitset_bits, cpu % bitset_bits);
         0 != cpuset.__bits[idx] & (1 << offset)
     }
+
+    pub fn SOCKCRED2SIZE(ngrps: usize) -> usize {
```
GuillaumeGomez, line 3716: Won't it be an issue on freebsd 11 to link to a non-existing function?

GuillaumeGomez: Ah no, nevermind me...

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me since the struct has the same layout between 13 and 14 and the function is implemented, not linked.

asomers (approved):
LGTM.  This is the best way to do with new symbols in the OS, rather than changed symbols.

PR #2590, opened 2021-12-12
Make some freebsd private struct fields public and upgrade crate version to 0.2.112

GuillaumeGomez:
Otherwise we can't use the structs at all...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "libc"
-version = "0.2.111"
+version = "0.2.112"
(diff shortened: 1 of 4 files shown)


Amanieu:
Can't you use the struct with `mem::zeroed`, like C code does with `memset`? Or do you actually need to look at the value of the private field?

GuillaumeGomez:
It is a linked list, I need the `next` field to be able to get the next element. The problem here isn't the initialization (I already used `zeroed` :wink: ).

GuillaumeGomez:
Can you publish the new version please? :)

PR #2603, opened 2021-12-23
freebsd: Correctly detect FreeBSD version

amshafer:
This change removes the CI check that applies to freebsd version cfg. This is causing problems where the freebsd version defaults to 11 when CI is not enabled, some types may be incorrect on a > 11 system,
and compilation failures ensue.
A concrete example is libc::dev_t. This was defaulting to u32 on my FreeBSD 13-STABLE desktop, when it should have been u64. I couldn't compile anything that uses libc::devname_r as a result.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -25,13 +25,13 @@ fn main() {
     // On CI, we detect the actual FreeBSD version and match its ABI exactly,
     // running tests to ensure that the ABI is correct.
     match which_freebsd() {
-        Some(10) if libc_ci || rustc_dep_of_std => {
+        Some(10) if rustc_dep_of_std => {
             println!("cargo:rustc-cfg=freebsd10")
         }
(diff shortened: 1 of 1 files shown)


JohnTitor:
Note that there was a previous attempt about that, #2581, and it was reverted as it causes an unwanted regression. See the discussion on the mentioned PR.

unrelentingtech:
> couldn't compile anything that uses libc::devname_r as a result

This means we're missing something like

```rust
    #[cfg_attr(all(target_os = "freebsd", freebsd11), link_name = "devname_r@FBSD_1.0")]
```

I really think we should add versioned symbols to **ALL** functions already.

JohnTitor:
I'm going to close this with the above reason, we should discuss how we treat these cfgs first, I think.

PR #2604, opened 2021-12-24
solarish couple of strings fn.

devnexen:


devnexen:
@Amanieu test time out :-)

PR #2615, opened 2022-01-04
define TIOCGWINSZ as c_ulong under arm-uclibc.

lancethepants:
I'm trying to bring up a new target for rust, `armv7-unknown-linux-uclibceabi (softfloat)`. https://github.com/rust-lang/rust/pull/92383 Looks like a lot of work has already been hashed out from the recent addition of `armv7-unknown-linux-uclibceabihf`. The only issue I'm currently seeing is when I encounter `TIOCGWINSZ` in a couple places. This is the error I see.
```
   Compiling termize v0.1.1 (/mmc/.cargo/registry/src/github.com-1285ae84e5963aae/termize-0.1.1)
error[E0277]: the trait bound `u32: From<i32>` is not satisfied
  --> src/platform/unix.rs:12:43
   |
12 |     let mut result = ioctl(STDOUT_FILENO, TIOCGWINSZ.into(), &mut window);
   |                      -----                ^^^^^^^^^^^^^^^^^ the trait `From<i32>` is not implemented for `u32`
   |                      |
   |                      required by a bound introduced by this call
   |
   = help: the following implementations were found:
             <u32 as From<Ipv4Addr>>
             <u32 as From<NonZeroU32>>
             <u32 as From<bool>>
             <u32 as From<char>>
           and 2 others
   = note: required because of the requirements on the impl of `Into<u32>` for `i32`

error[E0277]: the trait bound `u32: From<i32>` is not satisfied
```
I see the error in the `termize` crate, and also `shell.rs` in the cargo source.

My current fix is to define TIOCGWINSZ as c_ulong under arm-uclibc.

I don't want to break anything with the established `armv7-unknown-linux-uclibceabihf` target so perhaps @skrap could chime in and take a look. Maybe a better solutions exists.


skrap:
> I don't want to break anything with the established `armv7-unknown-linux-uclibceabihf` target so perhaps @skrap could chime in and take a look. Maybe a better solutions exists.

Since `ioctl` is defined in uclibc as `extern int ioctl (int __fd, unsigned long int __request, ...) __THROW;`, it seems that this change is the right one.   If you wanted to save some heartache for some future devs, you could change all of the constants from `man ioctl_tty` which appear here to the same type.

Thanks for contributing!

lancethepants:
I've updated my pull request. See if that looks alright.  Looks like there's a broader discussion on ioctls going on now though.

Amanieu:
ioctl is part of the kernel API, which means that the constants have the same value no matter which libc is used. It doesn't make sense to put these in modules that are specific to a particular libc.

The difference between gnu which uses `c_ulong` while musl and uclibc which use `c_int` only exists in Rust. The only thing we should care about is that the type matches the one used for the `request` parameter of the `ioctl` function.

We already have a typedef (`Ioctl`) which maps to the type used by the `request` parameter of the `ioctl` function for a particular libc. What needs to be done is:
- Change `Ioctl` for uclibc to `c_ulong` so it matches the `ioctl` signature.
- Define any ioctl constants in the common linux code, not libc-specific code, and using the `Ioctl` type instead of hard-coding `c_int` or `c_ulong`.

skrap:
Yeah, that's great!  @lancethepants do you feel like that's something you want to try out?  Or is it more than you want to take on?

lancethepants:
I'll give it a try, consolidate all the ioctl code.  It's true it's not libc-specific, but it is architecture specific, there are different values for different systems. I think it will probably all end up going under the `arch` directory.

Amanieu:
The tests failed on MIPS: https://github.com/rust-lang/libc/runs/4749363042?check_suite_focus=true

It seems that some headers are missing in the testsuite. Have a look at `libc-test/build.rs`.

lancethepants:
I had run into a nearly identical issue for x86 when defining a different ioctl that had not before been previously defined stopped compilation in the same exact manner. The mips issue is an ioctl that had only been defined under musl, and not gnu, which is where the issue cropped up.  I've undefined those ioctl for non-musl-mips, and maybe we'll see how it goes. I may need to undefine it for all non-musl builds, and maybe a few more issues like this one could pop up as well. We thought just defining everything and just being done with it would be nice, and I still think it's the way to go, just a few corner cases I think. I don't fully understand what the compilation issue is, it talks like the ioctl value is undefined, but then it worked when we had not defined the ioctl. Defining it somehow is messing it up I think.

lancethepants:
So the ppc error I did encounter the wrong value when doing this. Didn't know there was a test for it.  Everything I can find shows that the value is currently incorrect in rust.
https://github.com/golang/go/issues/19560
https://github.com/golang/go/commit/197f9ba11d4559cbd19350ca652da7881f4b273f

What's the best way to proceed? Just fix these one at a time until we pass all the test? It'd be nice if I could test all the tiers on my own with my own system in an automated fashion, but that alone would be a project for me.

Amanieu:
The failing test is on 32-bit ppc which has that constant as `0x403c7413`. On ppc64 it is `0x402c7413` (which matches the values in go that you linked to).

lancethepants:
Looking through linux source I can only find a single definition of tcgets (and friends) for all ppc. golang does define 32bit the same as 64 as well.
https://github.com/golang/go/blame/master/src/cmd/vendor/golang.org/x/sys/unix/zerrors_linux_ppc.go#L415
 Is there a ppc maintainer that can chime in because I do think it is wrong in rust. Where would I find them? 

@Amanieu Is it possible to allow all tests to run regardless if one fails? I think things can get hashed out quicker if I can see all results for all architectures and tiers.

Amanieu:
> Looking through linux source I can only find a single definition of tcgets (and friends) for all ppc. golang does define 32bit the same as 64 as well. [golang/go@`master`/src/cmd/vendor/golang.org/x/sys/unix/zerrors_linux_ppc.go#L415 (blame)](https://github.com/golang/go/blame/master/src/cmd/vendor/golang.org/x/sys/unix/zerrors_linux_ppc.go#L415) Is there a ppc maintainer that can chime in because I do think it is wrong in rust. Where would I find them?

Our CI tests libc by comparing the constants against the ones defined in C by creating a C test program which includes the relevant headers and compares the values with the expected one. We need to figure out why the C test program is giving a different value. You can try it yourself with `ci/run_docker.sh powerpc-unknown-linux-gnu`.

> @Amanieu Is it possible to allow all tests to run regardless if one fails? I think things can get hashed out quicker if I can see all results for all architectures and tiers.

You can remove `fail-fast: true` from the CI scripts to force all tests to complete even if one fails.

skrap:
Just to confirm, a simple PPC 32-bit program (compiled with https://toolchains.bootlin.com/downloads/releases/toolchains/powerpc-440fp/tarballs/powerpc-440fp--glibc--stable-2021.11-1.tar.bz2) does confirm that the C #define is 0x403c7413

```#include <sys/ioctl.h>
#include <stdio.h>
#include <termios.h>

void main() {
	printf("TCGETS: 0x%x\n", TCGETS);
}
```

yields:

```
$ qemu-ppc -L powerpc-440fp--uclibc--stable-2021.11-1/powerpc-buildroot-linux-uclibc/sysroot/ ./test
TCGETS: 0x403c7413
```

lancethepants:
So it turns out with TCGETS and a few others on powerpc(64) differ depending on libc. musl differs from the others for some reason, so that what I made libc do.

glibc & uclibc: 0x403c7413
musl: 0x402c7413

Amanieu:
There are a few commented-out variants of `TIOCGISO7816`, `TIOCSISO7816` and `FIOQSIZE`, what up with those?

lancethepants:
TIOCGISO7816 &TIOCSISO7816 were causing compilation issues in some places, notably x86_64 and others too I think. They weren't ever previously defined but I left them commented out so everything is included if not defined.  I couldn't verify FIOQSIZE on a couple arches, but it wasn't previously defined for those arches either.  I'm just trying to define as much as a can, and can verify, without breaking compilation.

I wrote a little program to help me get all the ioctls values by downloading every arch toolchain from bootlin (save hexagon) and compiling and running with qemu, taking from  @skrap 's idea in his previous post.
[test.txt](https://github.com/rust-lang/libc/files/7878155/test.txt)

Amanieu:
LGTM!

@bors r+

lancethepants:
Pushed a commit to fix sparc compilation.
After seeing https://github.com/rust-lang/libc/pull/2630 my last commit consolidates all ioctl functions to 1 per libc.

lancethepants:
Squashed and rebased on Master.

kaniini:
the musl changes look good

lancethepants:
Push a change that should fix the riscv issue.

lancethepants:
Pushed a commit to fix sparc64 issue. Interested to see how ci runs s390x. I'm getting errors on my machine from HEAD on Master.

PR #2625, opened 2022-01-13
Add `ip_mreqn` on more platforms

tamird:
See individual commits; please do not squash on merge. @Thomasdezeeuw

### DIFF
--- a/libc-test/semver/android.txt
+++ b/libc-test/semver/android.txt
@@ -2844,6 +2844,7 @@ intptr_t
 ioctl
 iovec
 ip_mreq
+ip_mreqn
 ip_mreq_source
 ipv6_mreq
 isalnum
--- a/libc-test/semver/apple.txt
+++ b/libc-test/semver/apple.txt
@@ -1724,6 +1724,7 @@ in6_pktinfo
 in_pktinfo
 initgroups
 integer_t
+ip_mreqn
 ipc_perm
 kern_return_t
 kevent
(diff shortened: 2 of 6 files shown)


tamird:
The failure on freebsd 14 seems to be spurious, but I don't think I can cause it retry.

Mek101:
Just push a new commit

Amanieu:
You can do
```
git commit --amend --no-edit
git push --force
```
To trigger a CI rebuild.

tamird:
Is that necessary to get a review?

Amanieu:
The code looks fine, I'd just like CI to pass before merging.

@bors r+

src/unix/linux_like/android/b64/x86_64/mod.rs
```diff
@@ -4,6 +4,12 @@ pub type greg_t = i64;
 pub type __u64 = ::c_ulonglong;
 
 s! {
+    pub struct ip_mreqn {
+        pub imr_multiaddr: ::in_addr,
+        pub imr_address: ::in_addr,
+        pub imr_ifindex: ::c_int,
+    }
```
Amanieu, line 11: This struct needs to be provided on all Android targets, not just x86_64. Otherwise the CI checks on merge will fail (bors tests some extra targets).

tamird: Done, thanks.

PR #2633, opened 2022-01-18
Move constants from linux/magic.h out of gnu and uclibc

jirutka:
These constants are provided by Linux (`linux/magic.h`), not libc.

/cc @kaniini


jirutka:
Why do you test alpine targets on Ubuntu instead of Alpine?

Amanieu:
Alpine doesn't help in this case since we need a cross-toolchain for a different arch.

If you look at ci/docker/mips-unknown-linux-musl/Dockerfile you will see that we are actually using a MIPS musl toolchain from openwrt for the specific CI builder that is failing. The kernel headers in that toolchain seem to predate binderfs support in the kernel.

You could try updating the toolchain used in CI, which should resolve this failure.

Amanieu:
Superseded by #2639.

PR #2634, opened 2022-01-19
uclibc: Don't specialize PTRACE_O_MASK for uclibc

skrap:
This constant comes from the kernel headers, so its value should not depend on the libc type.

To test: see instructions in https://github.com/rust-lang/libc/pull/2566


Amanieu:
That may be true but it may have a different value in the headers we check against in CI. I expect this to fail unless you update CI to use a more recent uclibc version.

@bors try

skrap:
@Amanieu Does bors actually run the unit tests on targets in `RUST_LINUX_NO_CORE_TARGETS`? If so, I don't see where!

skrap:
(Failure seems unrelated to this PR.)

lancethepants:
This commit passed sparc64 test on my system. Looking at the error I kind of wonder if the virtual machine just failed to start. Maybe a re-run is needed.

Amanieu:
Hmm you're right, I think it only checks that libc compiles, not that its definitions are valid.

In that case,

@bors r+

Amanieu:
The SPARC CI is a bit flaky and sometimes fails to start, we just restart it if that happens.

PR #2636, opened 2022-01-19
uclibc: fix O_TMPFILE value, hide some unused fields to fix libc-test

skrap:
This was a bit of a yak-shave.

First, O_TMPFILE was wrong because 020000000 is decimal, not octal.  :facepalm:   This was, of course, causing all sorts of things to be horrible.

That led me to switch to a newer toolchain which actually defined O_TMPFILE, so that the libc tests pass on `armv7-unknown-linux-uclibceabihf`.

That revealed that certain field names had changed since the last time I ran the libc tests on this architecture, which revealed that some struct fields (e.g. `__unused`) were being marked as `pub` where all other architectures left them as private.

So, this patch might seem somewhat sprawling, but without all of these changes it's not possible to run libc-test!

To test:
* get https://toolchains.bootlin.com/downloads/releases/toolchains/armv7-eabihf/tarballs/armv7-eabihf--uclibc--bleeding-edge-2021.11-1.tar.bz2
* unpack somewhere, and call that `$TOOLCHAIN`
```
cd libc-test

PATH=$TOOLCHAIN/bin:$PATH \
  CC_armv7_unknown_linux_uclibc=armv7-buildroot-linux-uclibc-gcc \
  CARGO_TARGET_ARMV7_UNKNOWN_LINUX_UCLIBCEABIHF_LINKER=armv7-buildroot-linux-uclibc-gcc \
  CARGO_TARGET_ARMV7_UNKNOWN_LINUX_UCLIBCEABIHF_RUNNER="qemu-arm -L $TOOLCHAIN/arm-buildroot-linux-uclibcgnueabihf/sysroot/" \
  cargo +nightly test --target armv7-unknown-linux-uclibceabihf -Zbuild-std=core,std    
```


skrap:
(ref https://github.com/nix-rust/nix/pull/1603)

skrap:
@Amanieu I would like to get CI 'libc-test` working for this target, but seeing as it's Tier-3 at the moment, I'm not sure if that's a change which would be accepted.  Do you have any opinions or advice there?

Amanieu:
I'm OK with adding uclibc to CI even though it is tier 3. It seems sorely needed consider the issues found in this PR :)

Could you add it in this PR?

Amanieu:
Actually I'm going to merge this as it is so that it makes the next libc release. You can add uclibc to CI in a separate PR.

@bors r+

PR #2639, opened 2022-01-21
Consolodate the Linux `*_SUPER_MAGIC` constants.

sunfishcode:
Move the Linux `*_SUPER_MAGIC` constants out of
android/uclibc/gnu-specific directories and into the common `linux_like` directory
so that they're available on all Linux-family platforms, and so that that all
Linux-family platforms have all the constants.


Amanieu:
I think this is the same as #2633: the musl version used in CI doesn't have the BINDERFS magic, which is causing the check to fail.

PR #2640, opened 2022-01-21
add CI for armv7-unknown-linux-uclibceabihf

skrap:
As discussed in https://github.com/rust-lang/libc/pull/2636, this is a framework for CI on targets with no prebuilt std.  They will build via `nightly` and `-Zbuild-std`, and run the unit tests via qemu-user (or whatever else the dockerfile specifies).

It seems like I could write some more documentation about how to add more targets of this type, but perhaps I can get a round of feedback on the mechanism here before I go and write that.

### DIFF
--- a/.github/workflows/bors.yml
+++ b/.github/workflows/bors.yml
@@ -151,6 +151,29 @@ jobs:
       - name: Execute run-docker.sh
         run: LIBC_CI=1 sh ./ci/run-docker.sh ${{ matrix.target }}
 
+  # These targets are tier 3 or otherwise need to have CI build std via -Zbuild-std.
+  # Because of this, only the nightly compiler can be used on these targets.
+  docker_linux_build_std:
+    name: Docker Linux Build-Std Targets
+    needs: [docker_linux_tier1, style_check]
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: true
+      max-parallel: 12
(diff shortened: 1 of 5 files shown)


skrap:
I don't know that I agree with shellcheck's warning here.  I _want_ to split `EXTRA_CARGO_ARGS` on spaces.  shellcheck's docs [call out](https://github.com/koalaman/shellcheck/wiki/SC2086#exceptions)  this valid use case, but their sh-compatible workaround is to put the whole call in a function and use `$@`, which feels like a pretty big change (and IMO pretty ugly).  Instead I'll just make the solution less general and not try to build a general extra args feature.

Amanieu:
LGTM!

@bors r+

Amanieu:
CI is also spewing a bunch of errors along the lines "/toolchain/relocate-sdk.sh: 39: file: not found".

skrap:
> CI is also spewing a bunch of errors along the lines "/toolchain/relocate-sdk.sh: 39: file: not found".

Yeah, I just pushed a fix for that, thanks for the heads up.

skrap:
Huh, I don't think bors ran anything for that commit.  Did I break it somehow??

Amanieu:
There seems to have been some issue with the macOS runners in CI. I restarted the jobs and it seems to be working now.

PR #2642, opened 2022-01-22
Define ip_mreqn on all Linux platforms

tamird:
Updates #1558.

PR #2643, opened 2022-01-22
Add ip_mreqn to Linux s390x

Thomasdezeeuw:
I'm not 100% sure this layout is correct, however since it's the same on
all other existing (at the time out writing) architectures I hope it's
the same on s390x.


tamird:
I also sent https://github.com/rust-lang/libc/pull/2642 in case we can just define this everywhere.

Thomasdezeeuw:
Closing in favour of https://github.com/rust-lang/libc/pull/2642.

PR #2647, opened 2022-01-24
solaris based systems priv api flags updates

devnexen:


src/unix/solarish/mod.rs
```diff
@@ -2233,6 +2233,14 @@ pub const __PROC_PROTECT: ::c_uint = 0x0008;
 pub const NET_MAC_AWARE: ::c_uint = 0x0010;
 pub const NET_MAC_AWARE_INHERIT: ::c_uint = 0x0020;
 pub const PRIV_AWARE_RESET: ::c_uint = 0x0040;
+pub const PRIV_XPOLICY: ::c_uint = 0x0080;
+pub const PRIV_PFEXEC: ::c_uint = 0x0080;
```
pfmooney, line 2237: Shouldn't this be `0x0100`?

pfmooney (changes requested):

src/unix/solarish/mod.rs
devnexen: good catch bad c/p :)

jclulow: Did you run the tests?  They generally catch transcription errors like this.

devnexen: it does I do in other platforms but in my illumos VM tests get stuck and mostly I have to reboot the VM.

PR #2650, opened 2022-01-24
Define the `MFD_HUGE_*` macros for Linux and Android.

sunfishcode:
This defines the `MFD_HUGE_*` flags which can accompany the `MEMFD_HUGETLB`
flag in Linux's [`memfd_create`] call.

[`memfd_create`]: https://man7.org/linux/man-pages/man2/memfd_create.2.html

### DIFF
--- a/ci/docker/mips-unknown-linux-musl/Dockerfile
+++ b/ci/docker/mips-unknown-linux-musl/Dockerfile
@@ -6,11 +6,11 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 RUN mkdir /toolchain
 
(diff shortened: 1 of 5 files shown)


sunfishcode:
The failure here:

> error: component download failed for rust-std-armv7-linux-androideabi: could not download file from 'https://static.rust-lang.org/dist/2018-10-25/rust-std-1.30.0-armv7-linux-androideabi.tar.xz' to '/home/runner/.rustup/downloads/441f7ae731b9a5b952957adc8338ec644a222592695ff9c096178330bcd5cad2.partial'

appears unrelated to this PR.

PR #2657, opened 2022-01-27
freebsd 14 add new PT_GETREGSET/PT_SETREGSET flag

devnexen:


devnexen:
I may have put this a bit early :) most likely the fbsd 14 image will have to catch up first.

Amanieu:
Ping me when you think the image will have been updated so we can merge this.

PR #2658, opened 2022-01-27
windows/msvc: export stdio file descriptor numbers

KapJI:
These are useful for using in `dup2` arguments, for example.

They are probably valid for every compiler but here it is in Microsoft example: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fileno?view=msvc-170#example

### DIFF
--- a/src/windows/msvc/mod.rs
+++ b/src/windows/msvc/mod.rs
@@ -1,6 +1,11 @@
 pub const L_tmpnam: ::c_uint = 260;
 pub const TMP_MAX: ::c_uint = 0x7fff_ffff;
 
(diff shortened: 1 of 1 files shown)


KapJI:
Well, MSVC apparently doesn't declare such constant. MSDN suggests to use `fileno(stdin)` instead.

Amanieu:
We don't expose anything in the libc crate that isn't exposed by the underlying libc. This is checked by our CI, which is failing now because MSVC doesn't expose these constants.

In this case you should probably just hardcode the file descriptor numbers 0, 1 and 2 in your code directly.

PR #2560, opened 2021-11-22
several illumos fixes, and some additional termios constants

jclulow:
I wanted to add some of the new baud rate constants that [we added recently](https://www.illumos.org/issues/13975).  On the way there, I had to make a variety of fixes to get the tests to pass on illumos systems.

PR #2661, opened 2022-01-30
solarish lgrp api subset

devnexen:


src/unix/solarish/mod.rs
```diff
@@ -2242,6 +2305,15 @@ pub const PRIV_USER: ::c_uint = PRIV_DEBUG
...
     | PRIV_PFEXEC;
 
+// sys/lgrp_user.h
+pub const LGRP_COOKIE_NONE: ::lgrp_cookie_t = 0;
+pub const LGRP_AFF_NONE: ::lgrp_affinity_t = 0x0;
+pub const LGRP_AFF_WEAK: ::lgrp_affinity_t = 0x10;
+pub const LGRP_AFF_STRONG: ::lgrp_affinity_t = 0x100;
+pub const LGRP_RSC_COUNT: ::lgrp_rsrc_t = 2;
+pub const LGRP_RSC_CPU: ::lgrp_rsrc_t = 0;
+pub const LGRP_RSC_MEM: ::lgrp_rsrc_t = 1;
```
pfmooney, line 2315: These names are misspelled. 
```diff
-pub const LGRP_RSC_COUNT: ::lgrp_rsrc_t = 2;
-pub const LGRP_RSC_CPU: ::lgrp_rsrc_t = 0;
-pub const LGRP_RSC_MEM: ::lgrp_rsrc_t = 1;
+pub const LGRP_RSRC_COUNT: ::lgrp_rsrc_t = 2;
+pub const LGRP_RSRC_CPU: ::lgrp_rsrc_t = 0;
+pub const LGRP_RSRC_MEM: ::lgrp_rsrc_t = 1;
```

src/unix/solarish/mod.rs
```diff
@@ -2852,6 +2924,35 @@ extern "C" {
...
+        content: ::lgrp_content_t,
+    ) -> ::c_int;
+    pub fn lgrp_mem_size(
+        cookie: ::lgrp_cookie_t,
+        lgrp: ::lgrp_id_t,
+        tpe: ::lgrp_mem_size_flag_t,
+        content: ::lgrp_content_t,
+    ) -> ::lgrp_mem_size_t;
+    pub fn lgrp_nlgrps(cookie: ::lgrp_cookie_t) -> ::c_int;
+    pub fn lgrp_view(cookie: ::lgrp_cookie_t) -> ::lgrp_view_t;
```
pfmooney, line 2955: These require `-llgrp` to link, and should probably be structured in a section similar to what @jclulow did in #2560 [for sendfile](https://github.com/rust-lang/libc/pull/2560/files#diff-907e77bc1adacd9de6453a42791090511ad8d2bb6b15445780a3394f9f3661b8R2843)

pfmooney (changes requested):

src/unix/solarish/mod.rs
```diff
@@ -61,6 +67,53 @@ impl ::Clone for ucred_t {
     }
 }
 
+#[cfg_attr(feature = "extra_traits", derive(Debug))]
+pub enum lgrp_content_t {
```
pfmooney, line 71: The added enums should probably be annotated `repr()`.  I'm not sure if `repr(C)` would be most appropriate, or an explicitly sized type.

src/unix/solarish/mod.rs
```diff
@@ -61,6 +67,53 @@ impl ::Clone for ucred_t {
...
+    LGRP_CONTENT_ALL = 0,
```
pfmooney, line 72: Why is this the only (new) enum type that is assigned a discriminant? I imagine they should be consistent in this manner?

pfmooney (commented):
With your recent changes, the tests now pass (when cherry-picked on top of the updated bits for #2560).  I had a few lingering questions around the enum definitions that perhaps the libc maintainers could chime in on.

src/unix/solarish/mod.rs
devnexen: I forgot to remove as I had another plan originally.

src/unix/solarish/mod.rs
devnexen: `repr(u32)` ?

Amanieu: As a general rule, try to avoid using Rust enums to represent C enums. The problem is that Rust is very strict about assuming that an enum will only ever contain the given values and optimizes based on that.

For example in this particular case, Rust wil represent `None` in `Option<lgrp_content_t>` with the value 2 since it knows that this never a valid value for `lgrp_content_t` (which can only be 0 or 1 according to this definition).

Prefer using a typedef for `c_int` and constants instead so that Rust code doesn't break when an enum contains an unexpected value.

jclulow:
Hey, a quick note for everybody: Patrick and I are fixing up the test suite on illumos in #2560 right now.  As far as I know that's ready to merge.

Once that merges, @devnexen will be able to run the tests for _this_ change to demonstrate that they're correct and there are no regressions.

Can we hold off on merging this until that's sorted and we have test results here?  Thanks.

Amanieu:
Sure, ping me once the tests pass.

devnexen:
@jclulow do you build rust from source on illumos ? packaged version is too old.

jclulow:
> @jclulow do you build rust from source on illumos ? packaged version is too old.

I generally use `rustup`, like I imagine folks do on other platforms.  The Rust project produces binaries for illumos.

devnexen:
I tried but does not seem to work (libgcc_s.so no such file or directory... is it because it s 32 bits in 64?)

jclulow:
> I tried but does not seem to work (libgcc_s.so no such file or directory... is it because it s 32 bits in 64 mode?)

What distribution are you using?

devnexen:
openindiana but ... should it matter ?

jclulow:
> openindiana but ... should it matter ?

You may need to install a package to get that file; e.g., http://pkg.openindiana.org/hipster/en/search.shtml?token=%2Fusr%2Flib%2Famd64%2Flibgcc_s.so.1&action=Search

devnexen:
in case it helps I also needed this 

`basename   dir    usr/gcc/4.9/lib/gcc/i386-pc-solaris2.11/4.9.4/include/ssp pkg:/developer/gcc-49@4.9.4-2021.0.0.8`

now it works thanks ;-) , indeed I use rustup in other platforms usually haiku is the only remaining os not working with

src/unix/solarish/mod.rs
Amanieu: @devnexen This comment still needs to be addressed/

devnexen: Yes I m fixing along now I have working env.

devnexen:
@jclulow @pfmooney I think I ve fixed all tests pass now.

jclulow:
> @jclulow @pfmooney I think I ve fixed all tests pass now.

Awesome!  Can you paste `uname -a` and test run output, like https://github.com/rust-lang/libc/pull/2560#issuecomment-1026213120 ?

pfmooney:
> @jclulow @pfmooney I think I ve fixed all tests pass now.

I'm seeing tests failures when I run with your change locally:
```
RUNNING ALL TESTS
bad lgrp_content_t signed: rust: 1 (0x1) != c 0 (0x0)
bad lgrp_lat_between_t signed: rust: 1 (0x1) != c 0 (0x0)
bad lgrp_mem_size_flag_t signed: rust: 1 (0x1) != c 0 (0x0)
bad lgrp_view_t signed: rust: 1 (0x1) != c 0 (0x0)
```

devnexen:
ah that might come from the fact they are enums.

pfmooney:
> ah that might come from the fact they are enums in reality.

How did you run the tests?  I'm confused about why our results differ.

devnexen:
no they did not I just did not notice in the bottom the rest. should be better now.

jclulow:
> no they did not I just did not notice in the bottom the rest. should be better now.

Awesome!  Can you paste `uname -a` and test run output, like https://github.com/rust-lang/libc/pull/2560#issuecomment-1026213120 ?

devnexen:
`SunOS openindianavbox 5.11 illumos-c8f4a9f75f i86pc i386 i86pc`

`dcarlier@openindianavbox:~/Contribs/libc/libc-test$ cargo test
   Compiling libc-test v0.2.117 (/export/home/dcarlier/Contribs/libc/libc-test)
    Finished test [unoptimized + debuginfo] target(s) in 15.38s
     Running test/cmsg.rs (/export/home/dcarlier/Contribs/libc/target/debug/deps/cmsg-18c4afdd13027561)

running 5 tests
test t::test_cmsg_data ... ok
test t::test_cmsg_len ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.49s

     Running test/errqueue.rs (/export/home/dcarlier/Contribs/libc/target/debug/deps/errqueue-4722944b212c6786)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running test/linux_elf.rs (/export/home/dcarlier/Contribs/libc/target/debug/deps/linux_elf-d1da74545cb2c958)
PASSED 0 tests
     Running test/linux_fcntl.rs (/export/home/dcarlier/Contribs/libc/target/debug/deps/linux_fcntl-2e24f7d0a34ff9e8)
PASSED 0 tests
     Running test/linux_if_arp.rs (/export/home/dcarlier/Contribs/libc/target/debug/deps/linux_if_arp-b2fb1c1458da6387)
PASSED 0 tests
     Running test/linux_ipv6.rs (/export/home/dcarlier/Contribs/libc/target/debug/deps/linux_ipv6-cb013354d9cb392d)
PASSED 0 tests
     Running test/linux_strerror_r.rs (/export/home/dcarlier/Contribs/libc/target/debug/deps/linux_strerror_r-6aec8256525fb5f1)
PASSED 0 tests
     Running test/linux_termios.rs (/export/home/dcarlier/Contribs/libc/target/debug/deps/linux_termios-26cf8337fd1a5def)
PASSED 0 tests
     Running test/main.rs (/export/home/dcarlier/Contribs/libc/target/debug/deps/main-4e5114f819c81262)
RUNNING ALL TESTS
PASSED 7316 tests
     Running test/semver.rs (/export/home/dcarlier/Contribs/libc/target/debug/deps/semver-c6da4dd438cfd52f)
PASSED 1 tests
     Running test/sigrt.rs (/export/home/dcarlier/Contribs/libc/target/debug/deps/sigrt-ed5b7344692da899)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s`

libc-test/build.rs
```diff
@@ -954,6 +955,15 @@ fn test_solarish(target: &str) {
         }
     });
 
+    cfg.skip_type(move |ty| match ty {
+        "sighandler_t"
+        | "lgrp_content_t"
+        | "lgrp_lat_between_t"
+        | "lgrp_mem_size_flag_t"
+        | "lgrp_view_t" => true,
```
pfmooney, line 963: It would be worth getting to the bottom of why these fail the tests.  Are all other enums defined in rust-libc ignored this why?  If not, why do we require them to be skipped here?

libc-test/build.rs
```diff
@@ -954,6 +955,15 @@ fn test_solarish(target: &str) {
         }
     });
 
+    cfg.skip_type(move |ty| match ty {
+        "sighandler_t"
+        | "lgrp_content_t"
```
pfmooney, line 960: If these exclusions do end up being required, they should be added up at line 815

pfmooney (changes requested):

libc-test/build.rs
devnexen: ah yes there is an existing one.

libc-test/build.rs
devnexen: not sure we have the choice since libc see the size as 0.

pfmooney: Which libc?  In C, their `sizeof` evaluates to 4 like normal `int` types.

devnexen: I meant the crate..

`RUNNING ALL TESTS
bad lgrp_content_t signed: rust: 1 (0x1) != c 0 (0x0)
bad lgrp_lat_between_t signed: rust: 1 (0x1) != c 0 (0x0)
bad lgrp_mem_size_flag_t signed: rust: 1 (0x1) != c 0 (0x0)
bad lgrp_view_t signed: rust: 1 (0x1) != c 0 (0x0)`

Amanieu: I think what the error is saying here is that there is a signed/unsigned mismatch between the Rust type and the C type. Are C enums unsigned on solaris?

devnexen: sorry my mistake seems they are when I change the signedness.

devnexen: can you retry pls @pfmooney ?

PR #2668, opened 2022-02-04
feat(riscv64,gnu): add some const PTRACE_*

CoelacanthusHex:
* PTRACE_GETFPREGS
* PTRACE_SETFPREGS
* PTRACE_GETFPXREGS
* PTRACE_SETFPXREGS
* PTRACE_GETREGS
* PTRACE_SETREGS

get from sys/ptrace.h of riscv64 glibc

Signed-off-by: Coelacanthus <coelacanthus@outlook.com>


libc-test/semver/linux-gnu-reicv64gc.txt
```diff
@@ -0,0 +1,6 @@
+PTRACE_GETFPREGS
```
Amanieu, line 1: Shouldn't the name of this file be `linux-gnu-riscv64gc.txt`?

CoelacanthusHex: Oh this was my mistake, I should have checked for these low-level errors before pushing. fixed.

PR #2673, opened 2022-02-06
linux glibc/android memory policies flags

devnexen:


src/unix/linux_like/linux/gnu/mod.rs
```diff
@@ -1121,6 +1121,13 @@ pub const REG_EEND: ::c_int = 14;
 pub const REG_ESIZE: ::c_int = 15;
 pub const REG_ERPAREN: ::c_int = 16;
 
+// linux/mempolicy.h
+pub const MPOL_DEFAULT: ::c_int = 0;
+pub const MPOL_PREFERRED: ::c_int = 1;
+pub const MPOL_BIND: ::c_int = 2;
+pub const MPOL_INTERLEAVE: ::c_int = 3;
+pub const MPOL_LOCAL: ::c_int = 4;
```
Amanieu, line 1129: Since this is taken directly from the kernel headers it should probably go in the libc-independent parts of the code so it works on musl/uclibc as well.

Amanieu:
Emscripten pretends to be Linux but isn't really linux... It should be excluded here.

PR #2676, opened 2022-02-11
Add mcontext_t and ucontext_t for ARM Linux

Amanieu:


src/unix/linux_like/linux/gnu/b32/arm/align.rs
```diff
@@ -4,4 +4,15 @@ s_no_extra_traits! {
     pub struct max_align_t {
         priv_: [i64; 2]
     }
+
+    #[allow(missing_debug_implementations)]
```
devnexen, line 8: do you plan ultimately to go all through the extra_traits work ?

Amanieu: I'm honestly quite skeptical about the whole concept of implementing these traits in the first place. `Debug` is probably fine but `Eq` and `Hash` make absolutely no sense for `ucontext_t`. I'll add them for completeness though.

devnexen: Yes most likely less useful in this case indeed and it s tedious to implement :-)

PR #2679, opened 2022-02-11
flopen from libutil on freebsd

devnexen:


Amanieu:
Failing on FreeBSD 11

PR #2695, opened 2022-02-21
illumos add few fcntl consts

devnexen:


pfmooney:
> note, not sure they re all illumos exclusive.

Looking at the git history can be a good indicator for that.
These all date back to the (Open)Solaris days and thus should be present on both platforms.

PR #2696, opened 2022-02-22
Fix `siginfo_t` for Illumos

dylni:
The `siginfo_t` struct appears to be incorrect for Illumos. It currently has `si_code` before `si_errno` and appears to be too large. I'm not familiar with this platform, so please let me know if this fix is inaccurate.

`siginfo_t`:  
https://github.com/illumos/illumos-gate/blob/74e98f497ae878ae481ccba1f91fd8dc26fa6627/usr/src/boot/sys/sys/signal.h#L196

siginfo.h:  
https://github.com/illumos/illumos-gate/blob/74e98f497ae878ae481ccba1f91fd8dc26fa6627/usr/src/uts/common/sys/siginfo.h#L164

I am not sure if the `packed()` attribute should remain, but I copied it from the Solaris definition.

### DIFF
--- a/src/unix/solarish/illumos.rs
+++ b/src/unix/solarish/illumos.rs
@@ -21,6 +21,45 @@ s! {
     }
 }
 
+s_no_extra_traits! {
+    #[cfg_attr(all(
+            any(target_arch = "x86", target_arch = "x86_64"),
+            libc_packedN
+        ), repr(packed(4)))]
+    #[cfg_attr(all(
+            any(target_arch = "x86", target_arch = "x86_64"),
+            not(libc_packedN)
+        ), repr(packed))]
+    pub struct siginfo_t {
+        pub si_signo: ::c_int,
+        pub si_errno: ::c_int,
+        pub si_code: ::c_int,
+        si_pid: ::pid_t,
+        si_uid: ::uid_t,
+        si_status: ::c_int,
+        pub si_addr: *mut ::c_void,
+        __pad: [u8; 36],
+    }
+
+    impl siginfo_t {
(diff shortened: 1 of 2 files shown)


jclulow:
> `siginfo_t`: https://github.com/illumos/illumos-gate/blob/74e98f497ae878ae481ccba1f91fd8dc26fa6627/usr/src/boot/sys/sys/signal.h#L196

That's not the right header.  Not that it's under `usr/src/boot` which was imported with our copy of the FreeBSD boot loader.

jclulow:
Can I ask what lead you to make this change?  Did something fail to compile or work correctly on an illumos system?

dylni:
@jclulow Yes, the motivation is this issue:
https://github.com/dylni/process_control/issues/12

libc currently does not define `siginfo_t::si_status` on Illumos, which is required by process_control. The signal codes are also required so were included with this change.

If you let me know the correct definition, I can certainly update this PR.

pfmooney:
> @jclulow Yes, the motivation is this issue: [dylni/process_control#12](https://github.com/dylni/process_control/issues/12)
> 
> libc currently does not define `siginfo_t::si_status` on Illumos, which is required by process_control. The signal codes are also required so were included with this change.
> 
> If you let me know the correct definition, I can certainly update this PR.

I'm working on an update to expose the `siginfo_t` fields you need using function accessors like the other unix-y platforms have done.  I should have it posted in the next day or so.

dylni:
Thank you @pfmooney! Please also include `CLD_EXITED` and the related constants from this PR if possible.

PR #2697, opened 2022-02-22
solarish: Expose siginfo_t data as functions

pfmooney:
This will bring illumos (and Solaris) functionality in line with the other UNIX (and UNIX-like) platforms, where unions often hamper access to those data fields.

### DIFF
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -28,6 +28,7 @@ pub type zoneid_t = ::c_int;
 pub type psetid_t = ::c_int;
 pub type processorid_t = ::c_int;
 pub type chipid_t = ::c_int;
+pub type ctid_t = ::id_t;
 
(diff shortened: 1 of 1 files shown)


pfmooney:
Hopefully this will address the needs that @dylni brought up in #2696.

pfmooney:
The normal libc-test suite runs clean (both with and without the extra_traits feature).  I also sketched up a little test program which imported `siginfo_t` instances allocated and populated from C to check the rust-libc parsing.  Those results were OK as well.

pfmooney:
> broken_heart Test failed - [checks-actions](https://github.com/rust-lang/libc/runs/5304820352?check_suite_focus=true)

I removed the compilation-conditional around `si_pad` for now to appease the compiler.

pfmooney:
Thanks for the prompt review.

PR #2706, opened 2022-03-03
Add SocketCan J1939 constants and structs

abalmos:
Add SocketCan J1939 constants and structs. 

Blocking a PR to `nix` to wrap SocketCan's j1939 module.


abalmos:
Looks like 

```
enum {
	J1939_NLA_PAD,
	J1939_NLA_BYTES_ACKED,
	J1939_NLA_TOTAL_SIZE,
	J1939_NLA_PGN,
	J1939_NLA_SRC_NAME,
	J1939_NLA_DEST_NAME,
	J1939_NLA_SRC_ADDR,
	J1939_NLA_DEST_ADDR,
};
```

wasn't added to the kernel until 5.15. 

I actually don't need these for my work. Should I just remove them from the PR or is there another way to deal with the CI being on an older kernel?

Amanieu:
You can add an exception here: https://github.com/abalmos/libc/blob/a5470ce96f994347e14c5659442a7c086d2665c4/libc-test/build.rs#L3168

abalmos:
@Amanieu Thanks! I *think* I pushed the fix. I am on 5.15, so the tests ran okay for me before this change

abalmos:
@Amanieu I might need help with this error. It seems unrelated to this PR.

Amanieu:
```
bad J1939_NO_ADDR value at byte 0: rust: 0 (0x0) != c 255 (0xff)
```
This means that the Rust value of the constant doesn't match the C value.

src/unix/linux_like/linux/mod.rs
```diff
@@ -3185,6 +3200,47 @@ pub const CAN_RAW_RECV_OWN_MSGS: ::c_int = 4;
 pub const CAN_RAW_FD_FRAMES: ::c_int = 5;
 pub const CAN_RAW_JOIN_FILTERS: ::c_int = 6;
 
+// linux/can/j1939.h
+pub const SOL_CAN_J1939: ::c_int = SOL_CAN_BASE + CAN_J1939;
+
+pub const J1939_MAX_UNICAST_ADDR: ::c_uchar = 0xfd;
+pub const J1939_IDLE_ADDR: ::c_uchar = 0xfe;
+pub const J1939_NO_ADDR: ::c_uchar = 0;
```
Amanieu, line 3208:
```suggestion
pub const J1939_NO_ADDR: ::c_uchar = 0xff;
```

abalmos:
Thanks @Amanieu! I'm not sure why `libc-test` worked on my machine from the get-go, but glad to see the CI passing now.

libc-test/build.rs
```diff
@@ -3228,10 +3229,16 @@ fn test_linux(target: &str) {
             | "IFLA_PERM_ADDRESS"
             | "IFLA_PROTO_DOWN_REASON" => true,
 
-            // FIXME: They require recent kernel header:
-            | "CAN_J1939"
-            | "CAN_RAW_FILTER_MAX"
-            | "CAN_NPROTO" => true,
```
Amanieu, line 3234: You need to preserve the old exceptions, they are needed on some targets.

abalmos:
@Amanieu Re-added.

Sorry about that, I didn't realize there was more test environments then what was in GitHub actions.

abalmos:
@Amanieu how do I check the error for this CI?

Amanieu:
Click on the [checks-actions](https://github.com/rust-lang/libc/runs/5426751365?check_suite_focus=true) link from bors. Ignore the semver tests, we allow failure for those.

abalmos:
I had tried that but I don't have any meaningful output. I just see that it failed:

SmartSelect_20220304-182632_Chrome

Amanieu:
You need to use the bar on the left to select the actual build step that failed. Here is the failure:
```
  running: "musl-gcc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-march=armv6" "-marm" "-mfpu=vfp" "-Wall" "-Wextra" "-Wall" "-Wextra" "-Werror" "-Wno-unused-parameter" "-Wno-type-limits" "-Wno-address-of-packed-member" "-Wno-unknown-warning-option" "-Wno-deprecated-declarations" "-D_GNU_SOURCE" "-D__GLIBC_USE_DEPRECATED_SCANF" "-o" "/checkout/target/arm-unknown-linux-musleabihf/debug/build/libc-test-87c193be97bba046/out/main.o" "-c" "/checkout/target/arm-unknown-linux-musleabihf/debug/build/libc-test-87c193be97bba046/out/main.c"
  cargo:warning=/checkout/target/arm-unknown-linux-musleabihf/debug/build/libc-test-87c193be97bba046/out/main.c:97:10: fatal error: linux/can/j1939.h: No such file or directory
  cargo:warning=   97 | #include <linux/can/j1939.h>
  cargo:warning=      |          ^~~~~~~~~~~~~~~~~~~
  cargo:warning=compilation terminated.
`

abalmos:
I am stumped by this one ... it looks like the `linux/can/j1939.h` header is not present in the CI, yet it is running 20.04 and *should* have it based on Ubuntu repos. We develop on an arm board running Debian Buster without problem.

Amanieu:
I think we should just go ahead and upgrade the version of the kernel headers used when testing musl to the latest available version.

abalmos:
@Amanieu I would be happy to try to work on that, but the CI is using the latest set of headers available 
 here: https://github.com/sabotage-linux/kernel-headers

I don't know too much about musl and how all hooks with rust. Would you have any pointers on getting started with said upgrade?

Amanieu:
Fair enough. In that case I think this is fine to land.

@bors r+

abalmos:
@Amanieu Is there any good way to run these tests locally?

abalmos:
I accidentally put `typedefs` into `skip_struct` rather then `skip_types`. 

I also found a comment that explains that testing kernel definitions in glibc targets only is sufficient, so I cleaned up the additions a bit by moving my skipped constant into that block.

*Hopefully* this version will pass :)

PR #2708, opened 2022-03-04
Correct the size of certain types on espidf platform

MabezDev:
This was initially discovered in https://github.com/esp-rs/rust/issues/92, the reason stat fails on the esp-idf platform is because the `stat` struct has a different layout on the Rust side compared to the C side.

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -23,8 +23,15 @@ pub type uintptr_t = usize;
 pub type ssize_t = isize;
 
 pub type pid_t = i32;
-pub type uid_t = u32;
-pub type gid_t = u32;
+cfg_if! {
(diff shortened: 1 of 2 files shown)


Amanieu:
LGTM! Is there a reason why this PR is still a draft?

MabezDev:
Apologies! I was double checking the c typedef's are the same in both our xtensa & RISC based toolchains, and forgot to mark this as ready. Should be good to go now :)

MabezDev:
CI failures seems unrelated? Looks like they're also failing on master.

gyscos:
Note that this broke building std on these `espidf` platforms.

See https://github.com/rust-lang/rust/issues/95924

PR #2709, opened 2022-03-05
openbsd hash search api

devnexen:


semarie:
it seems to me that `ACTION` is an enum (and not just `uint`).

devnexen:
I know but I ve been told in libc we prefer to map this as int/unsigned int.

Amanieu:
Rust enums don't behave like C enums and shouldn't be used in C bindings. Rust enums include the assumption that the value must be one of the defined variants and the compiler relies on this assumption for optimizations.

@bors r+

PR #2711, opened 2022-03-06
haiku build fixes

devnexen:


Amanieu:
CI style check is failing:
```
src/unix/haiku/native.rs:939 - use cfg_if! and submodules instead of #[cfg]
src/unix/haiku/mod.rs:1531 - line longer than 100 chars
```

PR #2712, opened 2022-03-06
macOS getcontext/setcontext

anqurvanillapy:
Signed-off-by: anqurvanillapy <anqurvanillapy@gmail.com>


anqurvanillapy:
Well, now I believe I could write my own `extern "C"` block in my project, and making things happen here is not suitable due to the deprecation. Closing this.

PR #2713, opened 2022-03-06
Make statx available for all Linux targets

alyssais:
I _think_ all this should be available on all Linux targets now, but I'm not very familiar with uClibc, so I'm not sure about it.

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -58,6 +58,38 @@ impl ::Clone for fpos64_t {
 }
 
 s! {
+    pub struct statx {
+        pub stx_mask: u32,
+        pub stx_blksize: u32,
+        pub stx_attributes: u64,
+        pub stx_nlink: u32,
+        pub stx_uid: u32,
+        pub stx_gid: u32,
+        pub stx_mode: u16,
+        __statx_pad1: [u16; 1],
+        pub stx_ino: u64,
+        pub stx_size: u64,
+        pub stx_blocks: u64,
+        pub stx_attributes_mask: u64,
+        pub stx_atime: ::statx_timestamp,
+        pub stx_btime: ::statx_timestamp,
+        pub stx_ctime: ::statx_timestamp,
+        pub stx_mtime: ::statx_timestamp,
+        pub stx_rdev_major: u32,
+        pub stx_rdev_minor: u32,
+        pub stx_dev_major: u32,
+        pub stx_dev_minor: u32,
+        pub stx_mnt_id: u64,
+        __statx_pad2: u64,
+        __statx_pad3: [u64; 12],
+    }
+
+    pub struct statx_timestamp {
+        pub tv_sec: i64,
+        pub tv_nsec: u32,
+        pub __statx_timestamp_pad1: [i32; 1],
+    }
+
     pub struct rlimit64 {
         pub rlim_cur: rlim64_t,
         pub rlim_max: rlim64_t,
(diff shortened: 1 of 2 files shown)


alyssais:
Oh, is there no musl/uClibc CI?

Amanieu:
We do have it but it is only run by bors, not the normal CI for PRs.

@bors r+

Amanieu:
You need to change libc-test/build.rs to include the statx header file in the C tests.

JohnTitor:
Ping @alyssais, could you address the above comment?

src/unix/linux_like/linux/mod.rs
```diff
@@ -3431,6 +3490,14 @@ extern "C" {
...
     pub fn shm_open(name: *const c_char, oflag: ::c_int, mode: mode_t) -> ::c_int;
 
+    pub fn statx(
+        dirfd: ::c_int,
+        pathname: *const c_char,
+        flags: ::c_int,
+        mask: ::c_uint,
+        statxbuf: *mut statx,
+    ) -> ::c_int;
+
```
germag, line 3500: AFIAK the current version of musl (1.2.3) doesn't implement a wrapper for`SYS_statx`.
But, I think it's worth leaving the rest (struct statx, constant, etc) available for all linux target

germag:
I'm quite interested in this PR

src/unix/linux_like/linux/mod.rs
alyssais: My mistake, you're correct.

alyssais:
I've rebased the PR and updated it to drop the statx function, which as @germag pointed out, I was wrong about being present on Musl.  Hopefully this also avoids the need to add a header to build.rs, although I'm not sure.

alyssais:
Thinking about it some more, is it right to expose even the structs if Musl doesn't? I thought it did, but it turns out the `struct statx` in Musl is internal-only, so to use it in a C program one would have to import <linux/stat.h>.  What if Musl came up with its own statx API that used different structs?

germag:
sorry for stepping in, are there any additional problems with this PR

JohnTitor:
> You need to change libc-test/build.rs to include the statx header file in the C tests.

The diff doesn't include the change mentioned above, have you tested that libc-test passes?

alyssais:
> > Thinking about it some more, is it right to expose even the structs if Musl doesn't? I thought it did, but it turns out the `struct statx` in Musl is internal-only, so to use it in a C program one would have to import <linux/stat.h>. What if Musl came up with its own statx API that used different structs?
> 
> if that were to happen, which I don't think it would, they would be incompatible with both the glibc and linux, since the `statx` struct is defined in the kernel headers.

I followed up on this in the #musl IRC channel, and the consensus there was that there's no guarantee that kernel headers and libc headers will be compatible for the same functionality, and that the libc crate probably shouldn't expose interfaces that come straight from the kernel but are not exposed by libc.

If libc does this in other places, then maybe it's okay, but I'd really like to hear the thoughts of a libc crate maintainer on this specific issue before I go further with this, since I'm not confident it's the way to go.

workingjubilee:
Mmm. I don't think we should be exposing this for all Linux targets. Mister Musl is correct, here. For instance, there's `AT_{ElfAuxval}` constants that are exposed by musl and glibc on a target-neutral basis, but that are _not_ defined in a target-neutral way by the kernel. Those libcs have made their own commitments, effectively, on how to make that work going forward, in the (admittedly unlikely) event the kernel introduces a new definition which clashes. I believe it's okay for us to "round up" a bit for certain definitions, but the lack of support from a notable and widely-used libc makes me hesitate. Fixing a struct definition can be easier than fixing a function definition, but it is definitely still annoying to do and liable to be highly breaking.

JohnTitor:
Triage: The arguments above make some sense to me. Let's close this PR for now and continue the discussion if someone wants to support it strongly. Thanks @alyssais for the PR anyway!

PR #2714, opened 2022-03-08
Horizon (Nintendo 3DS) getrandom function and fixes

AzureMarker:
This PR adds `getrandom`, conforming to the [Linux spec](https://man7.org/linux/man-pages/man2/getrandom.2.html), to the `horizon` OS (Nintendo 3DS).

The 3DS doesn't have a full libc implementation, and its randomness API is pretty complicated: https://github.com/rust-random/getrandom/pull/248. For this reason (see the linked PR for more details), the randomness implementation is abstracted by using the Linux `getrandom` interface.

This PR also fixes some types on the horizon platform. See the commits and diff.

cc: @ian-h-chamberlain @Meziu


src/unix/newlib/mod.rs
```diff
@@ -242,6 +219,57 @@ s! {
     }
 }
 
+cfg_if! {
+    if #[cfg(target_os = "horizon")] {
+        pub type sigset_t = ::c_ulong;
+
+        s! {
+            pub struct stat {
```
AzureMarker, line 227: This is throwing some errors in the style check. What's the recommended way to have a different `stat` struct and `sigset_t` type for this target? The `s` macro doesn't allow `cfg_if` in its body, which is why we went with this approach.

AzureMarker: I think flattening it is a no-go because it doesn't match the toolchain. There might be padding/ABI issues (I assume).

Amanieu: You need to move the Horizon-specific definitions to the `horizon/mod.rs` module. Generic definitions can go into a `generic` module used by non-horizon platforms.

AzureMarker: Updated to use a `generic` module.

PR #2717, opened 2022-03-08
Add support for Apple WatchOS

vladimir-ea:
Hi, I am hoping to add Apple WatchOS compile targets and these are prerequisite changes. 

Many thanks!
Vlad.


vladimir-ea:
This is the main PR: https://github.com/rust-lang/rust/pull/94736

src/unix/bsd/apple/mod.rs
```diff
@@ -1172,9 +1172,9 @@ s_no_extra_traits! {
         pub ifi_noproto: u64,
         pub ifi_recvtiming: u32,
         pub ifi_xmittiming: u32,
-        #[cfg(any(target_arch = "arm", target_arch = "x86"))]
+        #[cfg(any(target_arch = "arm", target_arch = "x86", target_pointer_width = "32"))]
```
Amanieu, line 1175: I think it might be better to change this to just `#[cfg(target_pointer_width = "32")]` and `#[cfg(target_pointer_width = "64")]`.

src/unix/bsd/apple/mod.rs
```diff
@@ -5522,7 +5522,7 @@ extern "C" {
 }
 
 cfg_if! {
-    if #[cfg(any(target_arch = "arm", target_arch = "x86"))] {
+    if #[cfg(any(target_arch = "arm", target_arch = "x86", target_pointer_width = "32"))] {
```
Amanieu, line 5525: Same here.

src/unix/bsd/apple/mod.rs
vladimir-ea: Thanks I will change this.

vladimir-ea:
@[Amanieu](https://github.com/Amanieu) updated as you suggested, sorry for the delay.

PR #2718, opened 2022-03-08
solarish pr(x)map data addition

devnexen:


devnexen:
cargo test --manifest-path libc-test/Cargo.toml 
   Compiling libc-test v0.2.119 (/export/home/dcarlier/Contribs/libc/libc-test)
    Finished test [unoptimized + debuginfo] target(s) in 11.14s
     Running test/cmsg.rs (target/debug/deps/cmsg-5f9f95dde8ae0a16)

running 5 tests
test t::test_cmsg_data ... ok
test t::test_cmsg_len ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.49s

     Running test/errqueue.rs (target/debug/deps/errqueue-25af991f283a08ca)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running test/linux_elf.rs (target/debug/deps/linux_elf-6f30f58b9b37962a)
PASSED 0 tests
     Running test/linux_fcntl.rs (target/debug/deps/linux_fcntl-878291074fed22f4)
PASSED 0 tests
     Running test/linux_if_arp.rs (target/debug/deps/linux_if_arp-bd69c11e5e71e04f)
PASSED 0 tests
     Running test/linux_ipv6.rs (target/debug/deps/linux_ipv6-ad453147d6bef766)
PASSED 0 tests
     Running test/linux_strerror_r.rs (target/debug/deps/linux_strerror_r-5f39b274aca2de03)
PASSED 0 tests
     Running test/linux_termios.rs (target/debug/deps/linux_termios-f2eaefc93d6b68ba)
PASSED 0 tests
     Running test/main.rs (target/debug/deps/main-4108ccb0a54f93cc)
RUNNING ALL TESTS
PASSED 7985 tests
     Running test/semver.rs (target/debug/deps/semver-90705d34f9f80c06)
PASSED 1 tests
     Running test/sigrt.rs (target/debug/deps/sigrt-84ea52b695c379bd)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

libc-test/build.rs
```diff
@@ -733,6 +733,7 @@ fn test_solarish(target: &str) {
     cfg.define("_XOPEN_SOURCE", Some("700"));
     cfg.define("__EXTENSIONS__", None);
     cfg.define("_LCONV_C99", None);
+    cfg.define("_STRUCTURED_PROC", Some("1"));
```
jclulow, line 736: For what did you need this define?

libc-test/build.rs
```diff
@@ -783,6 +784,7 @@ fn test_solarish(target: &str) {
         "sys/mman.h",
         "sys/mount.h",
         "sys/priv.h",
+        "sys/procfs.h",
```
jclulow, line 787: I believe this should be `procfs.h`, not `sys/procfs.h`, as per [proc(4)](https://illumos.org/man/4/proc).

jclulow (changes requested):

libc-test/build.rs
devnexen: to enable new struct instead (see sys/procfs.h and sys/old_procfs.h)

libc-test/build.rs
devnexen: sure for further mapping that might be better.

libc-test/build.rs
pfmooney: If you use the plain `procfs.h` include as noted below, that will be defined for you.

jclulow: Right, `procfs.h` is the public header we should be using for this.

src/unix/solarish/mod.rs
```diff
@@ -460,6 +462,38 @@ s! {
         pub pi_fputypes: [::c_char; PI_FPUTYPE as usize],
         pub pi_clock: ::c_int,
     }
+
+    pub struct prmap_t {
+        pub pr_vaddr: ::uintptr_t,
+        pub pr_size: ::size_t,
+        pub pr_mapname: [[::c_char; 32]; 2],
```
jclulow, line 469: Why did this change from `[::c_char; PRMAPSZ as usize]`?

devnexen: the failure might have to do with the fact it s an array sized > 32 (happened before).

pfmooney: If definitions implied by the `s!` macro cannot handle the >32 length, we should move this struct outside that macro, rather than altering the type.

devnexen: @jclulow see for ex https://github.com/rust-lang/libc/blob/8b05ed7ebb9cf24a3473764a3dca3f9f3397477f/src/unix/bsd/netbsdlike/netbsd/mod.rs#L514 there are couple more examples here and there.

Amanieu: I agree with @pfmooney here, I think it would be better to disable types that can't be expressed precisely on older rustc versions.

pfmooney: > I agree with @pfmooney here, I think it would be better to disable types that can't be expressed precisely on older rustc versions.

We could also just provide our own impls of `Eq` and `Hash` which handled the long array members here.

devnexen: That s what I started to do in fact.

Amanieu:
This is still failing in CI.

devnexen:
still an issue with old rust.

PR #2719, opened 2022-03-09
Create optionally-available __int128 typedefs and use them for ARM64 definitions.

### LINKED ISSUE #2524: `user_fpsimd_struct` on aarch64
On aarch64, `user_fpsimd_struct` includes a `__uint128_t` type:
https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/aarch64/sys/user.h;h=9927b3ac3c43096d85f65421c89c8bef5e0bbce9;hb=a85c93c42433aa52d4c5bc18e1dff038a82415bf#l32

There is a known issue regarding `__uint128_t` / `__int128_t` alignment to be different from `u128` / `i128` in rust.
I know that some of you guys faced this issue in some occasions (e.g. #2008).
In #2417 he tried few different ways to commit this struct, which fails tests.

Should we consider any other workaround, if there is any to consider?

Gankra:
Potentially fixes #2524, see the comments in the patch for details.

### DIFF
--- a/src/fixed_width_ints.rs
+++ b/src/fixed_width_ints.rs
@@ -18,3 +18,78 @@ pub type uint16_t = u16;
 pub type uint32_t = u32;
 #[deprecated(since = "0.2.55", note = "Use u64 instead.")]
 pub type uint64_t = u64;
+
+cfg_if! {
+    if #[cfg(all(libc_int128, target_arch = "aarch64", not(target_os = "windows")))] {
+        // This introduces partial support for FFI with __int128 and
+        // equivalent types on platforms where Rust's definition is validated
+        // to match the standard C ABI of that platform.
+        //
+        // Rust does not guarantee u128/i128 are sound for FFI, and its
+        // definitions are in fact known to be incompatible. [0]
+        //
+        // However these problems aren't fundamental, and are just platform
+        // inconsistencies. Specifically at the time of this writing:
+        //
+        // * For x64 SysV ABIs (everything but Windows), the types are underaligned.
+        // * For all Windows ABIs, Microsoft doesn't actually officially define __int128,
+        //   and as a result different implementations don't actually agree on its ABI.
+        //
+        // But on the other major aarch64 platforms (android, linux, ios, macos) we have
+        // validated that rustc has the right ABI for these types. This is important because
+        // aarch64 uses these types in some fundamental OS types like user_fpsimd_struct,
+        // which represents saved simd registers.
+        //
+        // Any API which uses these types will need to `#[ignore(improper_ctypes)]`
+        // until the upstream rust issue is resolved, but this at least lets us make
+        // progress on platforms where this type is important.
+        //
+        // The supported architectures and OSes is intentionally very restricted,
+        // as careful work needs to be done to verify that a particular platform
+        // has a conformant ABI.
+        //
+        // [0]: https://github.com/rust-lang/rust/issues/54341
+
+        /// C `__int128` (a GCC extension that's part of many ABIs)
+        pub type __int128 = i128;
+        /// C `unsigned __int128` (a GCC extension that's part of many ABIs)
+        pub type __uint128 = u128;
+        /// C __int128_t (alternate name for [__int128][])
+        pub type __int128_t = i128;
+        /// C __uint128_t (alternate name for [__uint128][])
+        pub type __uint128_t = u128;
+
+        cfg_if! {
+            if #[cfg(libc_const_assert)] {
+                // NOTE: if you add more platforms to here, you may need to cfg
+                // these consts. They should always match the platform's values
+                // for `sizeof(__int128)` and `_Alignof(__int128)`.
+                const _SIZE_128: usize = 16;
+                const _ALIGN_128: usize = 16;
+
+                /// Since Rust doesn't officially guarantee that these types
+                /// have compatible ABIs, we const assert that these values have the
+                /// known size/align of the target platform's libc. If rustc ever
+                /// tries to regress things, it will cause a compilation error.
+                ///
+                /// This isn't a bullet-proof solution because e.g. it doesn't
+                /// catch the fact that llvm and gcc disagree on how x64 __int128
+                /// is actually *passed* on the stack (clang underaligns it for
+                /// the same reason that rustc *never* properly aligns it).
+                const _ASSERT_128_COMPAT: () = {
+                    assert!(core::mem::size_of::<__int128>() == _SIZE_128);
+                    assert!(core::mem::align_of::<__int128>() == _ALIGN_128);
+
+                    assert!(core::mem::size_of::<__uint128>() == _SIZE_128);
+                    assert!(core::mem::align_of::<__uint128>() == _ALIGN_128);
+
+                    assert!(core::mem::size_of::<__int128_t>() == _SIZE_128);
+                    assert!(core::mem::align_of::<__int128_t>() == _ALIGN_128);
+
+                    assert!(core::mem::size_of::<__uint128_t>() == _SIZE_128);
+                    assert!(core::mem::align_of::<__uint128_t>() == _ALIGN_128);
+                };
+            }
+        }
+    }
+}
--- a/src/unix/linux_like/android/b64/aarch64/mod.rs
+++ b/src/unix/linux_like/android/b64/aarch64/mod.rs
@@ -382,3 +382,10 @@ cfg_if! {
         pub use self::align::*;
     }
 }
+
+cfg_if! {
+    if #[cfg(libc_int128)] {
+        mod int128;
+        pub use self::int128::*;
+    }
+}
--- a/build.rs
+++ b/build.rs
@@ -60,6 +60,11 @@ fn main() {
         println!("cargo:rustc-cfg=libc_align");
     }
 
+    // Rust >= 1.26 supports i128 and u128:
+    if rustc_minor_ver >= 26 || rustc_dep_of_std {
+        println!("cargo:rustc-cfg=libc_int128");
+    }
+
     // Rust >= 1.30 supports `core::ffi::c_void`, so libc can just re-export it.
     // Otherwise, it defines an incompatible type to retaining
     // backwards-compatibility.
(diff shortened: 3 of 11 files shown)


Gankra:
See https://github.com/rust-lang/rust/issues/54341#issuecomment-1062498329 for details on my investigation into this.

I will happily fill in more OS/arches if anyone can run https://github.com/Gankra/abi-checker on that target and get full passes. I only have access to aarch64 Android (phone) and macOS (coworker). I'm kinda blindly assuming iOS is the same as macOS here.

Gankra:
For context, properly defining these types is important for saving a crashing process' registers (i.e. when generating a breakpad-style minidump, which I am currently working on porting to Rust for Android).

src/fixed_width_ints.rs
```diff
@@ -18,3 +18,75 @@ pub type uint16_t = u16;
...
+                assert!(std::mem::align_of::<__int128_t>() == ALIGN_128);
+
+                assert!(std::mem::size_of::<__uint128_t>() == SIZE_128);
+                assert!(std::mem::align_of::<__uint128_t>() == ALIGN_128);
+            };
+        }
+    }
+}
+
+pub use self::supported_128::*;
```
bjorn3, line 92: nit: missing trailing newline

Gankra:
In the upstream issue nagisa notes that rust uniformly has this repr on all aarch64 targets -- but I'd still like to verify that the primary C compilers for those targets agree.

src/unix/linux_like/android/b64/aarch64/mod.rs
```diff
@@ -55,6 +55,12 @@ s! {
         pub pc: u64,
         pub pstate: u64,
     }
+
+    pub struct user_fpsimd_struct {
```
Amanieu, line 59: Can you add this for Linux as well? It's the same struct, it's defined in the kernel headers.

src/fixed_width_ints.rs
```diff
@@ -18,3 +18,70 @@ pub type uint16_t = u16;
 pub type uint32_t = u32;
 #[deprecated(since = "0.2.55", note = "Use u64 instead.")]
 pub type uint64_t = u64;
+
+cfg_if! {
+    if #[cfg(all(target_arch = "aarch64",
+        any(target_os = "android", target_os="macos", target_os = "ios")))] {
```
Amanieu, line 24: I would feel comfortable just doing `not(windows)` here. That way all non-Windows platforms automatically get support and this is confirmed by the static assert.

Amanieu (commented):
Great work! I think this was sorely needed.

Gankra:
So since putting up this PR I spent a bit more time digging into what can be done to polyfill this type on x64 (since SysV actually passes structs the same as ints, in general). Unfortunately on x64 (linux, at least) clang and gcc actually disagree on the pass-on-stack convention for `__int128`:

https://twitter.com/Gankra_/status/1502320041315946499

x64 SysV fails to specify what "pushing" a value actually means and so clang packs __int128 while gcc properly aligns it. This can introduce padding and differences and completely desynchronize the offsets to args.

In discussing this on Twitter some Apple engineers also pointed out that clang actually has macos-specific ABI differences for backcompat with itself (details vague, no documentation for this in general). Similarly a coworker noted that stuff like this is a problem for clang plugins when the plugin and compiler weren't both consistently built by clang or gcc.

This now has me increasingly paranoid about claiming u128/i128 are ABI-compatible on ARM64 without any way for me to properly compare gcc and clang both targetting these platforms. And also makes me less inclined to willy-nilly declare that Rust agrees with C compilers for "all non-windows OSes". Even asserting they agree on size and align isn't sufficient because the compilers can still disagree on how the value is pushed to the stack. :(

I'm not sure what libc's policy is on divergent impls for one platform, and how to declare one "canonical" over the others.

Unfortunately I can't really test this stuff out better -- the "gcc" on my phone's termux is just an alias for clang, and similarly I expect that will be the case for m1 macbooks.

------------

If anyone has better access to arm64 platforms and can setup clang/gcc properly, they can test this out by checking out:

https://github.com/Gankra/abi-checker

and running

```
cargo run -- --tests ui128
```

This will compare CC to rustc and CC to itself. That is a good start for rust matching the "canonical" implementation. 

To exhaustively try out all the combos of clang+gcc+rust to check if the ABI is actually consistent between all impls, you can do:

```
cargo run -- --tests ui128 --pairs clang_calls_gcc gcc_calls_clang rustc_calls_gcc gcc_calls_rustc rustc_calls_clang clang_calls_rustc clang_calls_clang gcc_calls_gcc rustc_calls_rustc
```

------------

Otherwise you can manually do something like [this godbolt](https://godbolt.org/z/Prx8MTjrP).

Amanieu:
I ran your ABI checker on AArch64 Linux, all the tests pass:
```
Final Results:
ui128::c::clang_calls_gcc        all 226/226 passed
ui128::c::gcc_calls_clang        all 226/226 passed
ui128::c::rustc_calls_gcc        all 226/226 passed
ui128::c::gcc_calls_rustc        all 226/226 passed
ui128::c::rustc_calls_clang      all 226/226 passed
ui128::c::clang_calls_rustc      all 226/226 passed
ui128::c::clang_calls_clang      all 226/226 passed
ui128::c::gcc_calls_gcc          all 226/226 passed
ui128::c::rustc_calls_rustc      all 226/226 passed
ui128::cdecl::rustc_calls_rustc  all 226/226 passed
ui128::stdcall::rustc_calls_rustc all 226/226 passed
ui128::fastcall::rustc_calls_rustc all 226/226 passed

2712 passed, 0 failed, 0 completely failed, 33 skipped
```

Gankra:
Oh nice!! that helps my confidence with going forward with this a lot. I'll clean this up on monday.

Gankra:
Also lmao @ rustc actually managing to do *something* on linux when you ask it to fastcall/stdcall with itself, I hadn't actually every tested that.

Gankra:
Applied review requests and squashed.

src/fixed_width_ints.rs
```diff
@@ -18,3 +18,74 @@ pub type uint16_t = u16;
...
+        /// have compatible ABIs, we const assert that these values have the
+        /// known size/align of the target platform's libc. If rustc ever
+        /// tries to regress things, it will cause a compilation error.
+        ///
+        /// This isn't a bullet-proof solution because e.g. it doesn't
+        /// catch the fact that llvm and gcc disagree on how x64 __int128
+        /// is actually *passed* on the stack (clang underaligns it for
+        /// the same reason that rustc *never* properly aligns it).
+        const _ASSERT_128_COMPAT: () = {
+            assert!(std::mem::size_of::<__int128>() == SIZE_128);
```
Amanieu, line 78: This should use `core` instead of `std`.

src/unix/linux_like/android/b64/aarch64/mod.rs
```diff
@@ -55,6 +55,12 @@ s! {
...
+        pub vregs: [__uint128_t; 32],
```
Amanieu, line 60: This type needs to be imported:
```suggestion
        pub vregs: [::__uint128_t; 32],
```

Gankra:
```
error: constant is never used: `SIZE_128`
  --> src/fixed_width_ints.rs:65:9
   |
65 |         const SIZE_128: usize = 16;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: the lint level is defined here
  --> src/lib.rs:18:38
   |
18 | #![cfg_attr(libc_deny_warnings, deny(warnings))]
   |                                      ^^^^^^^^
   = note: `#[deny(dead_code)]` implied by `#[deny(warnings)]`

error: constant is never used: `ALIGN_128`
  --> src/fixed_width_ints.rs:66:9
   |
66 |         const ALIGN_128: usize = 16;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `libc` due to 2 previous errors
```

??????? how ???????

bjorn3:
Maybe `const _ASSERT_128_COMPAT: () = { ... }` doesn't mark them as used because it isn't used itself? Would `const _: () = { ... }` work?

Gankra:
It looks like prefixing them with _'s is the best approach.

pushed fix

Gankra:
uhhh looks like the last remaining issue is... rust 1.13 support(????)

not sure what to do about that...

bjorn3:
`proc_macro::is_available()` was stabilized on rustc 1.57 (https://github.com/rust-lang/rust/commit/09dd213cd2253015b80605e374d624fcf3c9aea0). CI uses a nightly from a couple of weeks before the moment it was stabilized, but because it is still 1.57, proc_macro2 thinks it was stabilized and tries to use it. 2021-10-11 is the first nightly with `is_available()` stabilized I think. You can either upgrade to it or pin proc_macro2 to 1.0.35 which doesn't use `is_available()` yet.

bjorn3:
For the 1.13 issue you could conditionally compile it. The build script already detects a couple of rustc versions.

Gankra:
I *think* I added the cfgs?

Since these defs are conditional I added a fallback for the old macos definition. That seems like the right thing to do?

JohnTitor:
> i128 was added in Rust 1.26, you might want to add a check for that in build.rs.
> 
> For the unreachable in garando_syntax I have no idea. cc @JohnTitor
> 
> ```
> thread 'main' panicked at 'internal error: entered unreachable code', /cargo/registry/src/github.com-1ecc6299db9ec823/garando_syntax-0.1.0/src/ext/expand.rs:337:21
> ```

I don't have enough time to investigate, but the code is quite old and I suspect that a newer thing here triggers that panic.
FWIW here's the repo: https://github.com/JohnTitor/garando

Gankra:
sorry i am unfortunately stuck in the situation of compile-by-ci here because I don't have direct access to these platforms on my dev machine.

* macos freaked out from the way i cfg'd, went with cfg_if with the s!'s being internal
* musl aarch64 complained it didn't have fpsimd; added it since it's a kernel API so it should be portable?

Gankra:
arghh nope macros bad style now 😭

Gankra:
Is there a "right" way to express what I'm doing with macos here? Should I just drop updating the macos def? Or not make it backcompat?

Amanieu:
I would just drop the type on old rustc which don't support i128. The type definition is incorrect in any case.

Amanieu:
You can ignore the failure on sparc, it's spurious, just retry the build.

Gankra:
* const asserts only exist as of 1.57 so I added another cfg for that

* had to make a bunch more types optional behind int128 support in macos since the type is transitively used by several other ones.

Gankra:
trying to instead cfg the fields and keep the struct around, since the fallout from removing it seems kinda nasty

Gankra:
ok!!! this one actually seems to work and satisfy the nightmare cfg macro rube-goldberg machine (remaining failures are the pre-existing ones in the repo itself).

Gankra:
ok coming back to this with a fresh mind and the design of libc i think now has congealed into a coherent form for me. Hopefully this works now...

PR #2731, opened 2022-03-22
Add pthread_mutexattr_setprotocol and pthread_mutexattr_getprotocol on Linux

rtzoeller:
Include the PTHREAD_PRIO_* constants.


rtzoeller:
AFAICT the failures were not related to this change.

Amanieu:
We currently have emscripten as a linux-like platform but it doesn't support this function.

Amanieu:
To clarify, the actual failure is in the emscripten tests (click on the link from bors, then look for the asmjs builder on the left), which doesn't have these functions.

Amanieu:
It seems these functions don't exist on Android? Or our Android version in CI is too old. Feel free to upgrade it if that is the case.

rtzoeller:
I don't have a good way to validate the functionality on Android, so I've updated the PR to be Linux only.

PR #2732, opened 2022-03-24
Two improvements of the latest i128 numbers change

est31:
Two improvements of #2719:

* First, a typo/grammar fix
* Second, a fix to avoid usage of the anti pattern of `==` inside an `assert` macro. Since a few releases, `assert` is usable inside `const` expressions, but for the purpose of `==`, superior alternatives exist that also work at compile time but print the two values at a mismatch. `assert_eq` is not yet usable in const contexts due to formatting.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -87,9 +87,9 @@ fn main() {
         println!("cargo:rustc-cfg=libc_ptr_addr_of");
     }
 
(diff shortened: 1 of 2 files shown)


Amanieu:
cc @Gankra

Gankra:
TIL! neat!

Gankra:
looks like a CI error, retry seems appropriate

PR #2741, opened 2022-04-06
Add network types for Windows

GuillaumeGomez:


Amanieu:
I'm not sure libc is the right crate for these Windows-specific APIs. They are already available on Windows via the winapi crate.

Amanieu:
On windows these types are not part of libc (msvcrt) but are part of winsock which is part of the win32 API. I don't think these belong in the libc crate.

PR #2743, opened 2022-04-08
linux: add GRND_INSECURE constant

zx2c4:
This is available on kernels ≥5.6.


Amanieu:
This constant is not present on the images we use in CI (probably a too-old kernel). You can add an ignore rule in libc-test/build.rs.

zx2c4:
Oh thanks. I was wondering how that worked. Will fix and force push.

zx2c4:
Hm that's not a very informative failure...

Amanieu:
You need to look on the left and select the actual builder that failed: https://github.com/rust-lang/libc/runs/5963890427?check_suite_focus=true

zx2c4:
> You need to look on the left and select the actual builder that failed: https://github.com/rust-lang/libc/runs/5963890427?check_suite_focus=true

Ah, thanks. Should be all set now.

zx2c4:
@bors r+

(@Amanieu - I'm trying to get it to re-do the thing you did, but I don't know if me writing it will have any effect.)

zx2c4:
> 💔 Test failed - [checks-cirrus-freebsd-11](https://cirrus-ci.com/task/5392635555741696)

Surely that's a flake, right?

PR #2744, opened 2022-04-09
Add pthread_set/get schedparam functions to FreeBSD.

vityafx:
FreeBSD has these functions and this provides an interface to use them.

In particular, the [`thread-priority`](https://github.com/vityafx/thread-priority/runs/5881232298) crate can't be built without those.


vityafx:
I am not sure I understand what this error is about. Can you help me with it, please?

Amanieu:
I think it's just a spurious failure, let's try again.

@bors retry

PR #2747, opened 2022-04-09
Disable broken CI jobs

Amanieu:
These are temporarily removed until someone comes to fix them.

The semver jobs have been broken for over a year now, but are ignored by bors. Now we just disable them completely since they cause confusion when reading the build logs.

The uclibc targets were recently broken by https://github.com/rust-lang/rust/pull/95688. They are missing a `si_addr()` function on `struct siginfo`.

### DIFF
--- a/.github/workflows/bors.yml
+++ b/.github/workflows/bors.yml
@@ -274,6 +274,7 @@ jobs:
         shell: bash
 
   semver_linux:
+    if: ${{ false }} # This is currently broken
     name: Semver Linux
     runs-on: ubuntu-20.04
     continue-on-error: true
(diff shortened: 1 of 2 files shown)


Amanieu:
cc @skrap since you added one of the uclibc jobs.

skrap:
> cc @skrap since you added one of the uclibc jobs.

I'll submit a followup restoring the uclibc job.  Thanks for the CC!  (I wonder if there should be some standard way that target maintainers be notified if their targets break?)

Amanieu:
Maintainers for tier2/tier3 targets are normally listed [here](https://doc.rust-lang.org/nightly/rustc/platform-support.html), but this is a relatively recent policy so most targets are still missing dedicated maintainers.

PR #2750, opened 2022-04-11
Add siginfo accessors for uclibc

pfmooney:
With rust-lang/rust#95688 switching to the libc-provided `si_addr()` accessor on `siginfo_t`, it became apparent that the uclibc target was lacking that implementation (see rust-lang/rust#95866).

It would be nice to provide the same accessor in uclibc as the other UNIX-y platforms.

CC: @asomers @name1e5s @skrap - If you have a testing env for this, could you try it out?

### DIFF
--- a/src/unix/linux_like/linux/uclibc/mod.rs
+++ b/src/unix/linux_like/linux/uclibc/mod.rs
@@ -81,6 +81,32 @@ s! {
     }
 }
 
+impl siginfo_t {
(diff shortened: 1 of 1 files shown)


skrap:
> CC: @asomers @name1e5s @skrap - If you have a testing env for this, could you try it out?

Trying it out now!

skrap:
Builds and tests pass with `armv7-eabihf--uclibc--bleeding-edge-2021.11` from https://toolchains.bootlin.com/ !

PR #2751, opened 2022-04-11
Add unix adjtime support.

guyru:


guyru:
> broken_heart Test failed - [checks-actions](https://github.com/rust-lang/libc/runs/5996515299?check_suite_focus=true)

What does the checks-action test mean?

Amanieu:
That's just the action that reports the build status at the end. You need to find the actual failure on the list on the left. In this case: https://github.com/rust-lang/libc/runs/5996464765?check_suite_focus=true

Amanieu:
adjtime seems to be missing on Android.

PR #1344, opened 2019-05-16
Fix Android build jobs

gnzlbg:

PR #2758, opened 2022-04-13
Add misc constants and functions for android

fkm3:
These are the main diffs present in the downstream android opensource
project repo.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -1803,6 +1803,9 @@ fn test_android(target: &str) {
             // Added in glibc 2.25.
             "getentropy" => true,
 
+            // Added in API level 28, but some tests use level 24.
+            "getrandom" => true,
+
             _ => false,
         }
     });
--- a/src/unix/linux_like/android/b32/arm.rs
+++ b/src/unix/linux_like/android/b32/arm.rs
@@ -501,6 +501,9 @@ pub const SYS_pwritev2: ::c_long = 393;
 pub const SYS_pkey_mprotect: ::c_long = 394;
 pub const SYS_pkey_alloc: ::c_long = 395;
 pub const SYS_pkey_free: ::c_long = 396;
+pub const SYS_io_uring_setup: ::c_long = 425;
+pub const SYS_io_uring_enter: ::c_long = 426;
+pub const SYS_io_uring_register: ::c_long = 427;
 
 // offsets in mcontext_t.gregs from sys/ucontext.h
 pub const REG_R0: ::c_int = 0;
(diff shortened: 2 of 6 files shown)


Amanieu:
CI failure seems to be because we are using an old version of the Android emulator. You should try updating the CI scripts to use a newer version.

fkm3:
Thanks. It looks like it is using android SDK version 24 for arm and version 28 for x86, but `getrandom` was only added in version 28. I'll try bumping the version. https://github.com/rust-lang/libc/commit/eb0794a5d8af5b7b370b01c5ae4f7f5f0c14bc0d seems to have chosen different versions on purpose though, so I'm a little wary.

Amanieu:
This seems to have been due to an unavailability of emulator images for certain SDK versions: https://github.com/rust-lang/libc/pull/1344#issuecomment-494346531

fkm3:
There are no system images for armeabi-v7a after API version 24. It may be that support was dropped. I'm unsure.

I'm going to try moving the `getrandom` declaration into `b64/mod.rs` and then only bump the API version for aarch64.

Amanieu:
Maybe try using an arm64 image? It should support both 32-bit and 64-bit ARM.

Amanieu:
Seems like x86_64 is still running a v24 image.

fkm3:
I've updated x86_64 as well now. The newer `system.img` doesn't work easily with `e2fs`, so I had to add some extra code to workaround it. PTAL, thanks for sticking with me!

fkm3:
Switched from `dd` to `fallocate` (12.892 secs vs 0.438 on my machine). Hopefully that will fix the time out.

Amanieu:
I think there might be something wrong with the arm image being used?

fkm3:
Are there logs for the timeout?

Amanieu:
https://github.com/rust-lang/libc/runs/6141722587?check_suite_focus=true

fkm3:
Thanks. I can't find a solution. It seems like an over constrained problem.

Some notes, possibly incorrect:

* `getrandom` requires API version >= 28
* Android emulator support for arm64 on a x64 host was dropped somewhere in the API range (24, 28].
* There is some support for running arm on the x64 emulator ([link](https://android-developers.googleblog.com/2020/03/run-arm-apps-on-android-emulator.html)), but I'm not sure if it is relevant to the CI tests. Additionally, running the x64 emulator on x64 requires KVM (which I assume the CI doesn't support, if it runs on github actions)

I wonder if we can avoid the android emulator altogether with some combination of qemu + the arm64 system image (similar to what the android x64 tests are doing). I might give that a try, but will probably be slow to follow up.

Amanieu:
An alternative solution might be to simply exclude the new symbols from tests (see `libc-test/build.rs`) and keep using the old Android emulator image.

fkm3:
Sorry for the slow follow up. 

> An alternative solution might be to simply exclude the new symbols from tests (see libc-test/build.rs) and keep using the old Android emulator image.

I've uploaded new changes that try this approach.

I've also added another io_uring constant.

Thanks!

JohnTitor (approved):

PR #2760, opened 2022-04-19
Add netbsd's syscall.h.

m-ou-se:
This adds https://ftp.netbsd.org/pub/NetBSD/NetBSD-current/src/sys/sys/syscall.h

(I only need `SYS___futex`, but I added the rest as well.)


m-ou-se:
I don't know if these are stable, or how to find out if we can rely on these constants.

thomcc:
I think asking on one of the [NetBSD mailing lists](https://www.netbsd.org/mailinglists/) would be appropriate. Perhaps `tech-userlevel`?

semarie:
from [cvs syscall.h history](http://cvsweb.netbsd.org/bsdweb.cgi/src/sys/sys/syscall.h), the values seems relatively stable (in sense they doesn't change). but beware that it isn't the same stability than Linux: values could be removed if the syscall gone.

if you need some of them (because no libc wrapper), just adding the minimum required constants should be safe. else, you could lead to problems regarding the stability provided by rust libc crate (no symbols removed) versus the stability provided by the target (where symbols could be removed between major OS version).

thomcc:
Hm, I wonder if it's better to just put it inside the stdlib, where it's not exposed publicly and would just be used for the internal implementation... we'd have to fix it if it were removed, but it wouldn't break anybody downstream.

(Well... sort of. I guess it would mean if it were removed, code using older libstd would stop working, hm...)

m-ou-se:
> Hm, I wonder if it's better to just put it inside the stdlib

Yeah maybe. In that case we should only merge https://github.com/rust-lang/libc/pull/2762 and keep `SYS___futex` inside `std`.

semarie:
@Amanieu but does Boost why refuse to remove them if NetBSD do a breaking change ?

Amanieu:
The syscalls are part of the public headers so I guess you would just be expected to rebuild boost with the updated headers.

semarie:
@Amanieu for me, the problem is if **all** of these defines are commited in rust libc, it means that problems will arrived when NetBSD will remove one of them: rust libc stability will not permit to remove them.

Please note that importing only `SYS_futex` (or what it is effectivily required for rust-lang/rust/issues/93740) would be fine. as it is limiting the possible problems.

Amanieu:
In that case I'll just close this for now, we can just hard-code the sycall number in the standard library.

PR #2761, opened 2022-04-19
Add OpenBSD's futex.h.

m-ou-se:
This adds OpenBSD's [src/sys/sys/futex.h](https://cvsweb.openbsd.org/src/sys/sys/futex.h?rev=1.2&content-type=text/x-cvsweb-markup).

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -506,7 +506,7 @@ fn test_openbsd(target: &str) {
             // https://marc.info/?l=openbsd-cvs&m=154723400730318
             "mincore" => true,
 
-            // futex() has volative arguments, but that doesn't exist in Rust.
+            // futex() has volatile arguments, but that doesn't exist in Rust.
             "futex" => true,
 
             _ => false,
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -434,6 +434,7 @@ fn test_openbsd(target: &str) {
         "signal.h",
         "string.h",
         "sys/file.h",
+        "sys/futex.h",
         "sys/ioctl.h",
         "sys/ipc.h",
         "sys/mman.h",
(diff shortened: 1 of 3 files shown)


semarie:
Some points:
- `libc-test/build.rs` needs to include `futex.h` (for OpenBSD), else the CI will fail (github actions doesn't run it, but I am)
- `libc-test/semver/openbsd.txt` needs to include new public symbols
- `timespec` struct needs to be declared used inside `src/unix/bsd/netbsdlike/openbsd/mod.rs`

m-ou-se:
@semarie Thanks! How about now?

semarie:
@m-ou-se my local CI still doesn't pass due to `volatile` arguments in the function (rust doesn't understand volatile).
so could you add a "futex" entry inside `cfg.skip_fn()` with a comment, in order to avoid checking the function ?

something like:
```diff
diff --git a/libc-test/build.rs b/libc-test/build.rs
index 27497fa7189..11c9c1d2e5c 100644
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -505,6 +506,9 @@ fn test_openbsd(target: &str) {
             // https://marc.info/?l=openbsd-cvs&m=154723400730318
             "mincore" => true,
 
+            // futex has volative arguments, Rust doesn't understand them
+            "futex" => true,
+
             _ => false,
         }
     });
```

else, with it, all tests passed. I am fine with your PR.

semarie (approved):

libc-test/build.rs
```diff
@@ -505,6 +506,9 @@ fn test_openbsd(target: &str) {
             // https://marc.info/?l=openbsd-cvs&m=154723400730318
             "mincore" => true,
 
+            // futex() has volative arguments, but that doesn't exist in Rust.
```
thomcc, line 509:
```suggestion
            // futex() has volatile arguments, but that doesn't exist in Rust.
```

PR #2766, opened 2022-04-21
freebsd kinfo_kstat/sigtramp additions

devnexen:


src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -449,6 +449,19 @@ s! {
         pub kve_path: [[::c_char; 32]; 32],
     }
 
+    pub struct kinfo_kstack {
+        pub kkst_tid: ::lwpid_t,
+        pub kkst_state: ::c_int,
+        pub kkst_trace: [[::c_char; 32]; 32],
```
Amanieu, line 455: I'd rather keep the original size and just use a `cfg()` to only make this type available on newer rustc versions.

Amanieu:
Wait, this isn't the right cfg. It seems that we don't have one for this specific feature. Basically in [1.47](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1470-2020-10-08) most traits were updated to be implemented for arrays of any length. Add a new cfg in build.rs for this.

devnexen:
not sure FreeBSD 11 still exist https://pkg.freebsd.org/

Amanieu:
Yes, see #2779.

JohnTitor:
To revive the logs, @bors try

JohnTitor (approved):

JohnTitor:
@devnexen Are you still working on this?

devnexen:
> @devnexen Are you still working on this?

No I do not think so.

PR #2768, opened 2022-04-26
Promote pthread_getname_np and pthread_setname_np from glibc to linux

colincross:
musl libc added pthread_setname_np in 1.1.16 and pthread_getname_np
in 1.2.3, and uClibc has had them since v1.0.20.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -3401,6 +3401,9 @@ fn test_linux(target: &str) {
             // Not defined in uclibc as of 1.0.34
             "gettid" if uclibc => true,
 
+            // Needs musl 1.2.3 or later.
+            "pthread_getname_np" if musl => true,
+
             _ => false,
         }
     });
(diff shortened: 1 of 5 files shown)


Amanieu:
It seems pthread_getname_np is missing on MIPS with musl. Maybe we are using an older version in CI? If so then feel free to upgrade it to the latest version.

colincross:
Is the musl version coming from https://github.com/rust-lang/rust/blob/master/src/ci/docker/scripts/musl-toolchain.sh#L57?

Amanieu:
Oh right, we're stuck on 1.1.x for now due to #2088. For now you can just add an exception for these symbols in libc-test/build.rs.

Xuanwo:
PR https://github.com/rust-lang/libc/pull/2768 seems break `pthread_getname_np` on `musl` target, related issue: https://github.com/tikv/pprof-rs/issues/142

PR #2770, opened 2022-04-28
Add umtx_op to FreeBSD.

m-ou-se:
See https://www.freebsd.org/cgi/man.cgi?query=_umtx_op&sektion=2&manpath=FreeBSD+11.0-RELEASE+and+Ports and https://github.com/FreeBSD/freebsd-src/blob/stable/11/sys/sys/umtx.h.


m-ou-se:
@Amanieu Can you publish a new version of this crate to crates.io, so we can start using the additions in std?

Amanieu:
I'm waiting for bors to finish going through the PR queue, then I'll publish a release.

PR #2775, opened 2022-05-02
Enable aio_{read,write}v on FreeBSD

asomers:
When I originally added these symbols, I put them in the freebsd13
module.  But I needn't have, since they didn't change any existing
symbols.  Instead, as totally new symbols, they should've gone into the
freebsd module.  This way they can actually be used.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2352,6 +2352,8 @@ fn test_freebsd(target: &str) {
 
             // Added in FreeBSD 13.
             "getlocalbase" if Some(13) > freebsd_ver => true,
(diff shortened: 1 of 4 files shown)


asomers:
Ahh, that test failure is because FreeBSD's package servers have finally stopped serving packages for 11.  It's been EoL for a while.  So libc has two choices:
* Stop using pkg in CI, or
* Stop running CI on 11, which makes it hard to guarantee that we won't accidentally change the ABI.

Amanieu:
I've removed the FreeBSD 11 CI in #2779.

PR #2779, opened 2022-05-05
Remove FreeBSD 11 from CI

Amanieu:
FreeBSD 11 is EOL and packages for it are no longer provided, which causes the CI job to fail.

cc @asomers


Amanieu:
Blocked on https://github.com/rust-lang/homu/pull/169

PR #2792, opened 2022-05-17
Add NetBSD ki_sigset_t, kinfo_proc2, and kinfo_lwp structs

Freaky:


libc-test/semver/netbsd.txt
```diff
@@ -1275,8 +1275,10 @@ jrand48
 kevent
 key_t
 killpg
-kinfo_vmentry
```
Amanieu, line 1278: Why was this removed?

Freaky: It wasn't removed, I just sorted this hunk - it's under kinfo_proc2.

PR #2801, opened 2022-05-26
apple adding `IPV6_BOUND_IF` socket constant.

### LINKED ISSUE #2793: Consider adding socketopt const "IPV6_BOUND_IF" as we need to set/get bound interface on macOS
Consider add this to `bsd/apple/mod.rs`:
```rust
pub const IPV6_BOUND_IF: ::c_int = 125;
```
We really need this constant to perform modification on socket options(to bind socket to an interface, more specifically). Though we can temporarily define this on our own, but since `IP_BOUND_IF` is already defined in libc crate, we'd like to keep consistency.
I can make a simple PR if you agree to add this.
Thanks for your hard work.

devnexen:
closes #2793


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #2803, opened 2022-05-27
unix: setreuid/setregid

edigaryev:
This adds [`setreuid()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/setreuid.html) and [`setregid()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/setregid.html) on `unix`.

### DIFF
--- a/libc-test/semver/unix.txt
+++ b/libc-test/semver/unix.txt
@@ -761,6 +761,8 @@ setpgid
 setsid
 setsockopt
 setuid
+setreuid
+setregid
 setvbuf
 shm_open
 shm_unlink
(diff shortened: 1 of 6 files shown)


JohnTitor (approved):
Makes sense, thanks!

libc-test/semver/unix.txt
```diff
@@ -761,6 +761,8 @@ setpgid
 setsid
 setsockopt
 setuid
+setreuid
```
devnexen, line 764: nit: ideally in alphabetical order

JohnTitor: Good catch! I'm going to fix once this is merged.

PR #2807, opened 2022-05-28
solarish systems SOMAXCONN constant

devnexen:


pfmooney:
This matches the value of the constant on illumos, although I'm not sure if it's actually referenced by the kernel or libc anymore for applying resource limits. (As opposed to a more dynamic tunable in the network stack.)

devnexen:
In a multiplatform context when other oses define it.

JohnTitor (approved):

JohnTitor:
Sounds sensible, @bors r+

PR #2808, opened 2022-05-28
linux SOMAXCONN rework proposal.

devnexen:
glibc had increased to 4096 since couple of years already.


JohnTitor (approved):
I'm not sure if it's enough period to have a breakage, but reading https://github.com/bminor/glibc/commit/96958e2700f5b4f4d1183a0606b2b9848a53ea44, the higher value seems okay to me.

Amanieu:
This is probably fine, but the constant seems to be informative anyways: the actual limit is a sysctl option that can vary at runtime.

@bors r+

PR #2813, opened 2022-06-01
Fix the definition of sigevent

asomers:
It was originally defined back before rust could represent C unions.  So
instead of defining the union field correctly, it simply defined that
union's most useful field.  Define it correctly now.

Include a backwards-compatibility mechanism: Rename sigevent's old
definition and hide it.  Implement Deref and DerefMut from sigevent to
the old definition, so consumers will still be able to use the old field
name.

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -1355,6 +1355,10 @@ s! {
     // structure exists for backwards-compatibility with consumers that still
     // try to access that one member.
     #[doc(hidden)]
+    #[deprecated(
+        since = "0.2.127",
+        note = "Use sigevent instead"
+    )]
     pub struct sigevent_0_2_126 {
         pub sigev_notify: ::c_int,
         pub sigev_signo: ::c_int,
@@ -1664,6 +1668,7 @@ s_no_extra_traits! {
     }
 }
 
+#[allow(deprecated)]
 impl ::core::ops::Deref for sigevent {
     type Target = sigevent_0_2_126;
 
@@ -1672,6 +1677,7 @@ impl ::core::ops::Deref for sigevent {
     }
 }
 
+#[allow(deprecated)]
 impl ::core::ops::DerefMut for sigevent {
     fn deref_mut(&mut self) -> &mut Self::Target {
         unsafe { &mut *(self as *mut Self as *mut sigevent_0_2_126) }
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -1434,6 +1434,12 @@ s_no_extra_traits! {
         pub sigev_signo: ::c_int,
         pub sigev_value: ::sigval,
         pub _sigev_un: __c_anonymous_sigev_un,
+        /// Exists just to prevent the struct from being safely constructed,
+        /// because the Debug, Hash, PartialImpl, and
+        /// Deref<Target=sigevent_0_2_0126> trait impls might read uninitialized
+        /// fields of _sigev_un.  This field may be removed once those trait
+        /// impls are.
+        _private: ()
     }
 
     pub struct ptsstat {
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -888,6 +888,12 @@ s! {
         pub profhz: ::c_int,
     }
 
+    #[cfg(libc_union)]
+    pub struct __c_anonymous_sigev_thread {
+        pub _function: *mut ::c_void,   // Actually a function pointer
+        pub _attribute: *mut ::pthread_attr_t,
+    }
+
     pub struct __c_anonymous_stailq_entry_devstat {
         pub stqe_next: *mut devstat,
     }
@@ -1343,6 +1349,21 @@ s! {
         pub strchange_instrms: u16,
         pub strchange_outstrms: u16,
     }
+
+    // When sigevent was first added to libc, Rust still didn't support unions.
+    // So the definition only included one of the union's member.  This
+    // structure exists for backwards-compatibility with consumers that still
+    // try to access that one member.
+    #[doc(hidden)]
+    pub struct sigevent_0_2_126 {
+        pub sigev_notify: ::c_int,
+        pub sigev_signo: ::c_int,
+        pub sigev_value: ::sigval,
+        pub sigev_notify_thread_id: ::lwpid_t,
+        #[cfg(target_pointer_width = "64")]
+        __unused1: ::c_int,
+        __unused2: [::c_long; 7]
+    }
 }
 
 s_no_extra_traits! {
@@ -1393,16 +1414,22 @@ s_no_extra_traits! {
         __reserved: [::c_long; 4]
     }
 
+    // Can't correctly impl Debug for unions
+    #[allow(missing_debug_implementations)]
+    #[cfg(libc_union)]
+    pub union __c_anonymous_sigev_un {
+        pub _threadid: ::__lwpid_t,
+        pub _sigev_thread: __c_anonymous_sigev_thread,
+        pub _kevent_flags: ::c_ushort,
+        __spare__: [::c_long; 8],
+    }
+
+    #[cfg(libc_union)]
     pub struct sigevent {
         pub sigev_notify: ::c_int,
         pub sigev_signo: ::c_int,
         pub sigev_value: ::sigval,
-        //The rest of the structure is actually a union.  We expose only
-        //sigev_notify_thread_id because it's the most useful union member.
-        pub sigev_notify_thread_id: ::lwpid_t,
-        #[cfg(target_pointer_width = "64")]
-        __unused1: ::c_int,
-        __unused2: [::c_long; 7]
+        pub _sigev_un: __c_anonymous_sigev_un,
     }
 
     pub struct ptsstat {
@@ -1637,6 +1664,20 @@ s_no_extra_traits! {
     }
 }
 
+impl ::core::ops::Deref for sigevent {
+    type Target = sigevent_0_2_126;
+
+    fn deref(&self) -> &Self::Target {
+        unsafe { &*(self as *const Self as *const sigevent_0_2_126) }
+    }
+}
+
+impl ::core::ops::DerefMut for sigevent {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        unsafe { &mut *(self as *mut Self as *mut sigevent_0_2_126) }
+    }
+}
+
 cfg_if! {
     if #[cfg(feature = "extra_traits")] {
         impl PartialEq for utmpx {
@@ -1831,20 +1872,46 @@ cfg_if! {
                 self.sigev_notify == other.sigev_notify
                     && self.sigev_signo == other.sigev_signo
                     && self.sigev_value == other.sigev_value
-                    && self.sigev_notify_thread_id
-                        == other.sigev_notify_thread_id
+                    // sigev_notify indicates which union fields are valid
+                    && match self.sigev_notify {
+                        ::SIGEV_NONE => true,
+                        ::SIGEV_SIGNAL => true,
+                        ::SIGEV_THREAD => unsafe {
+                            self._sigev_un._sigev_thread
+                                == other._sigev_un._sigev_thread
+                        },
+                        ::SIGEV_KEVENT => unsafe {
+                            self._sigev_un._kevent_flags
+                                == other._sigev_un._kevent_flags
+                        },
+                        ::SIGEV_THREAD_ID => unsafe {
+                            self._sigev_un._threadid
+                                == other._sigev_un._threadid
+                        },
+                        _ => false
+                    }
             }
         }
         impl Eq for sigevent {}
         impl ::fmt::Debug for sigevent {
             fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
-                f.debug_struct("sigevent")
-                    .field("sigev_notify", &self.sigev_notify)
-                    .field("sigev_signo", &self.sigev_signo)
-                    .field("sigev_value", &self.sigev_value)
-                    .field("sigev_notify_thread_id",
-                           &self.sigev_notify_thread_id)
-                    .finish()
+                let mut ds = f.debug_struct("sigevent");
+                ds.field("sigev_notify", &self.sigev_notify);
+                ds.field("sigev_signo", &self.sigev_signo);
+                ds.field("sigev_value", &self.sigev_value);
+                // The sigev_notify field indicates which union fields are valid
+                unsafe {
+                    match self.sigev_notify {
+                        ::SIGEV_THREAD => ds.field("_sigev_thread",
+                            &self._sigev_un._sigev_thread),
+                        ::SIGEV_KEVENT => ds.field("_kevent_flags",
+                            &self._sigev_un._kevent_flags),
+                        ::SIGEV_THREAD_ID => ds.field("_threadid",
+                            &self._sigev_un._threadid),
+                        _ => &mut ds
+                    }
+                };
+                ds.finish()
             }
         }
         impl ::hash::Hash for sigevent {
@@ -1852,7 +1919,15 @@ cfg_if! {
                 self.sigev_notify.hash(state);
                 self.sigev_signo.hash(state);
                 self.sigev_value.hash(state);
-                self.sigev_notify_thread_id.hash(state);
+                // The sigev_notify field indicates which union fields are valid
+                unsafe {
+                    match self.sigev_notify {
+                        ::SIGEV_THREAD => self._sigev_un._sigev_thread.hash(state),
+                        ::SIGEV_KEVENT => self._sigev_un._kevent_flags.hash(state),
+                        ::SIGEV_THREAD_ID => self._sigev_un._threadid.hash(state),
+                        _ => ()
+                    };
+                }
             }
         }
 
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -888,12 +888,6 @@ s! {
         pub profhz: ::c_int,
     }
 
-    #[cfg(libc_union)]
-    pub struct __c_anonymous_sigev_thread {
-        pub _function: *mut ::c_void,   // Actually a function pointer
-        pub _attribute: *mut ::pthread_attr_t,
-    }
-
     pub struct __c_anonymous_stailq_entry_devstat {
         pub stqe_next: *mut devstat,
     }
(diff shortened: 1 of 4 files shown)


asomers:
@Amanieu what can we do about backwards-compat?  It should only be a problem for crates that uses the fairly obscure `sigev_notify_thread_id` field.  That includes Nix, which I can fix.  Can we use crater in "grep mode" just to find affected crates?

JohnTitor:
>  Can we use crater in "grep mode" just to find affected crates?

AFAIK no, the cheapest way is `cargo check`.

JohnTitor:
@Amanieu Semi-related, is there any discussion/progress about libc crate versioning? I'd like to resolve the root cause of such an issue someday...

Amanieu:
No, there's not been any discussion on this. We could bring it up at the next libs team meeting, but I'm not really sure what needs to be done here.

asomers:
Then here's what I suggest:
* Run Crater in check mode on all of libc's direct dependencies.
* If Nix is the only affected crate, then we can make the change, and publish a near synchronous patch release of Nix uses the newer libc crate.
* If there are a lots of crates that would be affected, then we change the PR to create a `sigevent2` structure and deprecate the original.  At some future date we rename `sigevent2` to `sigevent` and make `sigevent2` a deprecated type alias.

Does that sound reasonable?  Is craterbot already activated for this repo?

Amanieu:
Crater is really designed to test builds of the Rust compiler & standard library, not external crates like libc. I'm sure it would be possible to adapt it, but I'm not familiar with crater and wouldn't know where to start.

As an alternative, it's possible to download the source code for all crates published on crates.io: https://twitter.com/m_ou_se/status/1433085053056262144

This would allow grepping for any uses of `sigev_notify_thread_id` to see if any crates other than nix would break from this change.

asomers:
I used a command inspired by that Twitter post to download and search every crate version.  What I found was:
* Nix
* A few Nix forks, all since abandoned
* The autd-timer and autd3-timer crates, both abandoned and deleted from their upstream repo with the comments "start over" and "start over v2"
* The async-timer crate, which used this symbol up until v0.2.7, but then stopped.  And FWIW it never used that symbol correctly.
* A bunch of false positives

asomers:
@Amanieu do you approve of the plan to fix libc and deal with the breakage by immediately publishing patch releases of Nix?

Amanieu:
I'm concerned about the amount of breakage: there's a lot of crates that aren't using the latest major version of Nix. Do you think we could implement a backward-compatibility hack using `Deref`/`DerefMut` so that `.sigev_notify_thread_id` continues to work?

asomers:
We would definitely patch more than just the current major version.  I think I should be able to do the top five or six.  But what about this DerefMut proposal.  How would that work?

Amanieu:
If we implement `DerefMut` for `sigevent` such that it dereferences to an anonymous union with a `sigev_notify_thread_id`, existing users of `sigev_notify_thread_id` should still continue working. We also don't need to worry about accessing uninitialized data since reading out of a union is still unsafe.

asomers:
That's a good idea, and it looks like it works!  I'll amend the PR.

asomers:
@Amanieu the remaining CI failure is due to a limitation in the style checker.  It doesn't understand that the type declaration is part of an impl block.  I'm not sure how to fix that.  Do you have any ideas?

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -990,6 +996,21 @@ s! {
...
+    // When sigevent was first added to libc, Rust still didn't support unions.
+    // So the definition only included one of the union's member.  This
+    // structure exists for backwards-compatibility with consumers that still
+    // try to access that one member.
+    #[doc(hidden)]
+    pub struct sigevent_0_2_126 {
+        pub sigev_notify: ::c_int,
+        pub sigev_signo: ::c_int,
+        pub sigev_value: ::sigval,
+        pub sigev_notify_thread_id: ::lwpid_t,
```
Amanieu, line 1009: I would prefer if this was a `union` that just contains a single field. This forces the access to the `sigev_notify_thread_id` field to be in an `unsafe` block, which is necessary since this may be an uninitialized field in the union.

Amanieu:
Not sure about how to deal with the style checker, @JohnTitor any ideas?

JohnTitor:
> Not sure about how to deal with the style checker, @JohnTitor any ideas?

Fixed in #2824

src/unix/bsd/freebsdlike/freebsd/mod.rs
asomers: But the whole point of this structure is for backwards-compatibility with older libc versions, and they didn't require `unsafe`.  Wouldn't it break backwards-compatibility then, if we put this field into a union?

Amanieu: It just needs to be backwards-compatible enough for existing users, which is basically just nix. Does nix access this field outside of an unsafe block?

asomers: Yes, existing Nix does not use `unsafe`.

asomers: Also, I don't think that it should be considered unsafe to access the union member when using the backwards-compatible definition.  Since it's defined as a struct with a bunch of padding fields, the whole structure should always be initialized, and it shouldn't be a problem to read from the "union" member.

asomers:
Rebased and squashed.

src/unix/bsd/freebsdlike/freebsd/mod.rs
Amanieu: I think this is most likely fine, but I would like a second opinion from @JohnTitor.

JohnTitor: Hmm, I'm not sure if I've got 100% of the context here so don't take my thought as a strong opinion, but I think it doesn't have to be _perfectly correct_ while it's just for back-compat.

asomers:
Fourth time's the charm :crossed_fingers:

asomers:
Ok, I got the Android CI stuff running locally.  It turns out that Android requires a `skip_field_type` for `_sigev_un`.  I don't know why the other OSes don't.  I'm going to squash and merge.

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1202,6 +1229,20 @@ s_no_extra_traits! {
     }
 }
 
+impl ::core::ops::Deref for sigevent {
+    type Target = sigevent_0_2_126;
+
+    fn deref(&self) -> &Self::Target {
+        unsafe { &*(self as *const Self as *const sigevent_0_2_126) }
```
dtolnay, line 1236: This Deref impl exposes UB to safe code on freebsd.

```rust
fn main() {
    let sigevent = libc::sigevent {
        sigev_notify: 0,
        sigev_signo: 0,
        sigev_value: libc::sigval {
            sival_ptr: 0 as *mut _,
        },
        _sigev_un: libc::__c_anonymous_sigev_un {
            _kevent_flags: 0,
        },
    };
    let _val = sigevent.sigev_notify_thread_id;
}
```

Tested in miri using https://github.com/rust-lang/miri/pull/2221:

```console
$ cargo miri run --target x86_64-unknown-freebsd

error: Undefined Behavior: type validation failed: encountered uninitialized bytes, but expected initialized bytes
  --> src/main.rs:12:16
   |
12 |     let _val = sigevent.sigev_notify_thread_id;
   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type validation failed: encountered uninitialized bytes, but expected initialized bytes
   |
   = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
   = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
           
   = note: inside `main` at src/main.rs:12:16
```

@Amanieu I thought this exact unsoundness issue came up in a recent T-libs-api meeting. Did we end up deciding that this is "okay enough" as a tradeoff? I didn't see it discussed explicitly here on the PR.

Screenshot from 2022-06-20 21-58-56

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1396,28 +1437,62 @@ cfg_if! {
...
+                                == other._sigev_un._sigev_thread
+                        },
+                        ::SIGEV_KEVENT => unsafe {
+                            self._sigev_un._kevent_flags
+                                == other._sigev_un._kevent_flags
+                        },
+                        ::SIGEV_THREAD_ID => unsafe {
+                            self._sigev_un._threadid
+                                == other._sigev_un._threadid
+                        },
```
dtolnay, line 1455: All these unsafe blocks are reading and comparing potentially uninitialized data from the union.

```rust
fn main() {
    let sigevent = libc::sigevent {
        sigev_notify: libc::SIGEV_THREAD,
        sigev_signo: 0,
        sigev_value: libc::sigval {
            sival_ptr: 0 as *mut _,
        },
        _sigev_un: libc::__c_anonymous_sigev_un {
            _kevent_flags: 0,
        },
    };
    let _eq = sigevent == sigevent;
}
```

```console
error: Undefined Behavior: type validation failed: encountered uninitialized raw pointer
    --> /git/libc/src/unix/bsd/freebsdlike/freebsd/mod.rs:233:1
     |
233  | / s! {
234  | |     pub struct aiocb {
235  | |         pub aio_fildes: ::c_int,
236  | |         pub aio_offset: ::off_t,
...    |
1013 | |     }
1014 | | }
     | |_^ type validation failed: encountered uninitialized raw pointer
     |
     = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
     = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
             
     = note: inside `<libc::__c_anonymous_sigev_thread as std::cmp::PartialEq>::eq` at /git/libc/src/unix/bsd/freebsdlike/freebsd/mod.rs:876:9
     = note: inside `<libc::sigevent as std::cmp::PartialEq>::eq` at /git/libc/src/unix/bsd/freebsdlike/freebsd/mod.rs:1445:29
note: inside `main` at src/main.rs:12:15
    --> src/main.rs:12:15
     |
12   |     let _eq = sigevent == sigevent;
     |               ^^^^^^^^^^^^^^^^^^^^
     = note: this error originates in the derive macro `PartialEq` (in Nightly builds, run with -Z macro-backtrace for more info)
```

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1396,28 +1437,62 @@ cfg_if! {
...
+                ds.field("sigev_value", &self.sigev_value);
+                // The sigev_notify field indicates which union fields are valid
+                unsafe {
+                    match self.sigev_notify {
+                        ::SIGEV_THREAD => ds.field("_sigev_thread",
+                            &self._sigev_un._sigev_thread),
+                        ::SIGEV_KEVENT => ds.field("_kevent_flags",
+                            &self._sigev_un._kevent_flags),
+                        ::SIGEV_THREAD_ID => ds.field("_threadid",
+                            &self._sigev_un._threadid),
```
dtolnay, line 1475: Same here.

```rust
fn main() {
    let sigevent = libc::sigevent {
        sigev_notify: libc::SIGEV_THREAD,
        sigev_signo: 0,
        sigev_value: libc::sigval {
            sival_ptr: 0 as *mut _,
        },
        _sigev_un: libc::__c_anonymous_sigev_un {
            _kevent_flags: 0,
        },
    };
    println!("{:?}", sigevent);
}
```

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1396,28 +1437,62 @@ cfg_if! {
...
                 self.sigev_notify.hash(state);
                 self.sigev_signo.hash(state);
                 self.sigev_value.hash(state);
-                self.sigev_notify_thread_id.hash(state);
+                // The sigev_notify field indicates which union fields are valid
+                unsafe {
+                    match self.sigev_notify {
+                        ::SIGEV_THREAD => self._sigev_un._sigev_thread.hash(state),
+                        ::SIGEV_KEVENT => self._sigev_un._kevent_flags.hash(state),
+                        ::SIGEV_THREAD_ID => self._sigev_un._threadid.hash(state),
```
dtolnay, line 1492: Same here.

I think either these impls need to be removed because `sigev_notify` is not trustworthy, or constructing `sigevent` with a user-controlled `sigev_notify` needs to be unsafe in some way, or constructing `sigevent` needs to be disallowed altogether.

dtolnay (changes requested):

src/unix/bsd/freebsdlike/freebsd/mod.rs
asomers: But that field will never be uninitialized, because no program should ever use both `_sigev_un` and `sigev_notify_thread_id`.  The latter only exists to provide backwards-compatibility with the pre-union code.  So anybody accessing that field should already have initialized the entire structure.

src/unix/bsd/freebsdlike/freebsd/mod.rs
asomers: I agree.  I think we should remove most trait impls for any struct that contains a union (see #2816).  But how to do it without breaking existing code?

src/unix/bsd/freebsdlike/freebsd/mod.rs
dtolnay: I get that what you said is what programs _should_ do. If the libc crate ensures that programs do as you described, then that's great. If the **programmer** needs to ensure they do as you described or else they get memory unsafety (which is what is currently implemented in this PR) then we need to make them write `unsafe` somewhere.

asomers: The problem is that the old, non-union version of `sigevent` is already in the wild.  And since it isn't `unsafe`, there simply isn't any way for libc to fix the definition of `sigevent`, be backwards-compatible, *and* be `unsafe`.  The deref compromise suggested by @Amanieu is IMHO the best compromise.
FWIW there is exactly one currently-maintained crate on crates.io that uses the old `sigev_notify_thread_id` field, and I'm a maintainer of that crate, so I can fix any resulting bugs.  It's my intention to release a new version of Nix as soon as libc releases a new version with this PR.  However, the backwards-compatibility is still useful for users of older Nix versions.

dtolnay: > there simply isn't any way for libc to fix the definition of `sigevent`, be backwards-compatible, *and* be `unsafe`

Before this PR, libc::sigevent cannot be constructed in safe code because it has private padding fields. After this PR it can be constructed as shown above, which is a soundness bug. Preserving the property that it cannot be constructed in safe code is not a breaking change. The safety requirement on constructing a sigevent can be that you never expose it in a way that would mismatch the sigev_notify with the active variant of _sigev_un.

You can use #\[non_exhaustive\] if libc uses that these days, or just a zero sized private field with appropriate comment.

dtolnay (changes requested):
`_private` looks good.

The 1.13.0 build is still going to fail like in https://github.com/rust-lang/libc/runs/6977334685 though. Please either get `cargo +1.13.0 build` working, or open a separate PR to remove the 1.13.0 job from CI because that is a ridiculous version to support.

asomers:
@Amanieu @dtolnay I don't know how to satisfy both the style checker and Rust 1.13.0.  Older versions of Rust (tested on 1.14.0) seem to expand the macro incorrectly.  If I place the `#[cfg(libc_union)]` attribute directly over the struct, then recent rustc versions work but 1.14.0 seems to ignore it when expanding the s! macro.  Hard to be sure exactly, because cargo-expand does not work with Rust 1.14.0.  Placing `#[cfg(libc_union)]` before s! works, but that fails the style checker.

What do you want to do ?  Raise the MSRV to 1.19.0? Relax the style checker?  Or is there a version of cargo-expand that works with such ancient Rust versions, which might help me debug the macro?

JohnTitor:
> open a separate PR to remove the 1.13.0 job from CI because that is a ridiculous version to support.

Why such aggressive language? Are you mocking the maintenance efforts that have been made?
Whatever the policy is, I'm hesitant to break it with a patch version while it's alive. And I don't think following it is "ridiculous".

> What do you want to do ? Raise the MSRV to 1.19.0? Relax the style checker? Or is there a version of cargo-expand that works with such ancient Rust versions, which might help me debug the macro?

I prefer to tweak the style checker, maybe adding support for "ignore this block" or "ignore this file" would be sufficient.

asomers:
BTW, the current changes in this PR include the required style checker tweaks.  However, if https://github.com/rust-lang/libc/pull/2845 concludes , then I'll be able to revert the style checker changes and go back to preferred style.

asomers:
@dtolnay since #2845 has stalled, would you consider merging this PR, despite its style changes?

workingjubilee:
This PR has seen a lot of action, and it starts with a note of "...will fail to build with these changes". I skimmed it and am not sure where it's at right now, so I'm marking it as a breakage candidate since we're likely to start evaluating things we want to pull in for a 0.3 release real soon now, and I don't want this to get missed in the review for that, **even if it is not breaking anything**. However if you could make the PR's opening message match what the current status of this PR is, that would be greatly appreciated.

asomers:
> This PR has seen a lot of action, and it starts with a note of "...will fail to build with these changes". I skimmed it and am not sure where it's at right now, so I'm marking it as a breakage candidate since we're likely to start evaluating things we want to pull in for a 0.3 release real soon now, and I don't want this to get missed in the review for that, **even if it is not breaking anything**. However if you could make the PR's opening message match what the current status of this PR is, that would be greatly appreciated.

Updated, @workingjubilee !  The current status of this PR is that, as far as I am concerned, it is complete.  However, there is a trilemma: we cannot have both backwards-compatibility AND rust 1.13.0 support AND an unmodified style checker.  The current state of the PR relaxes the style checker, but I would be ok with relaxes either or both of the other conditions, too.

asomers:
@workingjubilee as explained in the current PR summary, this change should not cause any breakage.  Can we move forward with it?  And as far as the trilemma goes, which is your preference to drop?

* 100% Backwards-compatibility (though only one crate will be affected)
* Rust 1.13.0 support
* An unmodified style checker

workingjubilee:
:thinking:

workingjubilee:
Let me see if I can get a reduced version of #2813 through.

asomers:
@workingjubilee do you need help with said reduced version?

asomers:
@workingjubilee I rebased on main.  This version retains Rust 1.14.0 compatibility, but relaxes the style checker.

asomers:
BTW, I don't think I ever explained the motivation for this change.  Without this change, it is impossible to use a multi-threaded Tokio runtime with POSIX AIO.  Only a single-threaded runtime can be used.  Switching to a multi-threaded runtime by using this libc branch improves my application's performance by 180% .  So the PR isn't just for correctness's sake; it's got real value too.

SteveLauC:
If #3630 is a rebase of this PR, we can then close this PR? 🤔

tgross35:
Considering the challenges this PR has had and how long it has been open, I think we are better off not worrying about 0.2 and just going forward with a fix for 1.0 via https://github.com/rust-lang/libc/pull/3630.

PR #2817, opened 2022-06-04
ucontext usage update on macos 64 bits.

### LINKED ISSUE #2812: Aarch64 macOS: ucontext_t is not FFI-Safe
When using `ucontext_t` in `getcontext` (manually added, the API seems not to be exported on aarch64 mac in libc), like so:
```rust
extern "C" {
    fn getcontext(ucp: *mut ucontext_t) -> c_int;
}
```
the compiler complains that ucontext_t is not ffi-safe, as 128 bit integers are not supported in FFI.
I did not find any use of 128 bit ints, hence, I assume the compiler is overly smart for the neon state(?) which is:
```rust
    #[repr(align(16))]
    pub struct __darwin_arm_neon_state64 {
        pub __v: [[u64; 2]; 32],
        pub __fpsr: u32,
        pub __fpcr: u32,
    }
````

When I hand-roll the ucontext_t struct, including a neon-state like this, the compiler no longer complains.
```rust
#[repr(C, align(16))]
pub struct arm_neon_state64 {
    pub opaque: [u8; (32 * 16) + (2 * mem::size_of::<u32>())],
}
```
Coincidentally, my testcase would segfault after the original getcontext, which may be related.

This is on macos, aarch64, with libc 0.2.121 and rust 1.59.0

devnexen:
closes #2812

### DIFF
--- a/src/unix/bsd/apple/b64/aarch64/align.rs
+++ b/src/unix/bsd/apple/b64/aarch64/align.rs
@@ -15,6 +15,7 @@ s! {
         pub uc_link: *mut ::ucontext_t,
         pub uc_mcsize: usize,
         pub uc_mcontext: mcontext_t,
+        __mcontext_data: __darwin_mcontext64,
     }
 
     pub struct __darwin_mcontext64 {


src/unix/bsd/apple/b64/mod.rs
```diff
@@ -109,6 +109,9 @@ extern "C" {
         path2: *const ::c_char,
         options: ::c_uint,
     ) -> ::c_int;
+
+    #[allow(deprecated)]
```
JohnTitor, line 113: Why is this attribute needed?

libc-test/build.rs
```diff
@@ -324,6 +324,9 @@ fn test_apple(target: &str) {
             // macOs 12 minimum
             "backtrace_async" => true,
 
+            // deprecated api
+            "getcontext" => true,
```
JohnTitor, line 328: I'm not sure if adding a deprecated API is a good idea. If there are no good alternatives, users should report it upstream, adding it here just delays the problem.

devnexen: I hesitated to be honest.

domenukk: There isn't any good alternative afaik, and I'm not sure if Apple cares about low-level APIs enough to provide one

JohnTitor: My concern is that deprecated items are likely to be removed, i.e. they would introduce a breaking change. Such a code should belong to userland IMHO.

JohnTitor: @devnexen By the way, we don't have to ignore it on the test anymore, right?

JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

JohnTitor:
crates.io failure @bors retry

PR #2822, opened 2022-06-10
Add doc aliases on __error() function

GuillaumeGomez:
We recommend using `__errno_location` everywhere except on apple and freebsd targets, which is a bit confusing.

I can instead reexport it as `__errno_location` if you prefer?

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -5020,6 +5020,8 @@ extern "C" {
         thread_info_out: thread_info_t,
         thread_info_outCnt: *mut mach_msg_type_number_t,
     ) -> kern_return_t;
+    #[cfg_attr(doc, doc(alias = "__errno_location"))]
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):

JohnTitor:
Makes sense!

> I can instead reexport it as `__errno_location` if you prefer?

I'm hesitant to add non-existing names, it may cause some confusion if someone wants to use it only on those platforms.
Let's add doc aliases for now.

JohnTitor:
Ah, [doc alias is available on Rust >= 1.48.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md#rustdoc-5), we should add only on these Rust versions using `cfg_attr`. Otherwise CI will complain about it, I think.

GuillaumeGomez:
Good point! Updating it then.

GuillaumeGomez:
Updated!

JohnTitor:
Thanks! @bors r+

PR #2829, opened 2022-06-19
memfd_create additional flags for FreeBSD.

devnexen:


Amanieu:
Tests are failing. It seems some constants have the wrong value.

PR #2835, opened 2022-06-25
Add MAP_GUARD for FreeBSD 12+

jjl:
As per https://github.com/freebsd/freebsd-src/blob/main/sys/sys/mman.h#L103


jjl:
I got confused. Seems it's already there

PR #2841, opened 2022-07-10
Fix the type of file flags on OpenBSD

asomers:
In C they're defined as macros, but since they're used with chflags(2)
they should have the same type as that function's flags argument.


semarie (approved):
the chflags(2) prototype is `int chflags(const char *path, unsigned int flags);`, so yes, it makes sense.

LGTM

PR #2842, opened 2022-07-11
Fix FreeBSD CPU_ macros

rtzoeller:
The current definitions conflate bits and bytes, leading to panics.

Found while trying to add support for `nix::sched::sched_setaffinity` to FreeBSD.


rtzoeller:
FYI @asomers

asomers:
Good catch.  Maybe there should be unit tests for those functions?

rtzoeller:
@asomers I'll improve the test coverage in `nix` when I add FreeBSD support for the wrapper.

PR #2848, opened 2022-07-18
add crt-static for android

Bryanskiy:


src/lib.rs
```diff
@@ -26,10 +26,6 @@
 #![deny(missing_copy_implementations, safe_packed_borrows)]
 #![cfg_attr(not(feature = "rustc-dep-of-std"), no_std)]
 #![cfg_attr(feature = "rustc-dep-of-std", no_core)]
-#![cfg_attr(
-    feature = "rustc-dep-of-std",
-    feature(native_link_modifiers, native_link_modifiers_bundle)
-)]
```
JohnTitor, line 32: Could you explain why this change is necessary?

JohnTitor (commented):
Also, it'd be helpful to provide the context/reason for this change.

src/lib.rs
petrochenkov: `native_link_modifiers` and `native_link_modifiers_bundle` are already stable, so these features produce `#[warn(stable_features)]` in `rustc-dep-of-std` mode (which turns into an error when building standard library in rust-lang/rust).

petrochenkov:
>Also, it'd be helpful to provide the context/reason for this change.

This is the libc part of supporting statically linked executables on Android, the rustc part is https://github.com/rust-lang/rust/pull/99421.

Amanieu:
```
  = note: /android/ndk-arm/bin/../lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: cannot find -lrt
          /android/ndk-arm/bin/../lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: cannot find -lpthread
```

src/unix/mod.rs
```diff
@@ -346,10 +346,17 @@ cfg_if! {
...
+            cfg(target_feature = "crt-static"))]
+        #[link(name = "m", kind = "static", modifiers = "-bundle",
+            cfg(target_feature = "crt-static"))]
+        #[link(name = "m", cfg(not(target_feature = "crt-static")))]
+        #[link(name = "c", cfg(not(target_feature = "crt-static")))]
+        extern {}
     } else if #[cfg(any(target_os = "macos",
                         target_os = "ios",
                         target_os = "watchos",
-                        target_os = "android",
```
petrochenkov, line 352: With these changes as is the case of Android *without* `rustc-dep-of-std` is never covered, so this line needs to be kept.
Sorry, didn't notice this previously.

petrochenkov:
Does anybody know why libc on musl and redox doesn't link anything at all in regular no-std mode (without `rustc-dep-of-std`)?
It looks like a bug but people apparently lived with this for years?

petrochenkov:
Implementation for glibc is also incorrect, but in a different way - in regular no-std mode the libraries are linked dynamically even if `crt-static` is enabled.

petrochenkov:
Ah, I see, all of this is indeed incorrect but cannot be properly fixed on stable version of libc because link time `cfg` is unstable.
(But unstable features can be used in `feature = "rustc-dep-of-std"` mode.)

Anyway, this PR is about Android and the least bad thing to do for this target would be to follow what glibc does (i.e. https://github.com/rust-lang/libc/pull/2848#discussion_r933900736).

PR #2851, opened 2022-07-23
Android's prop_info type fix.

### LINKED ISSUE #2844: Android `prop_info` has changed size
libc defines `prop_info` for Android, as an opaque type but one whose size presumably matters (since libc fills in fields to match its size). However, Android appears to have changed the size of the private fields: `name` is now a 0-length array at the end.

This structure doesn't seem to ever be handled by value, only by pointer, so I don't think we need to specify its private fields at all.

devnexen:
closes #2844.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -1664,9 +1664,6 @@ fn test_android(target: &str) {
             "termios2" => true,
             // uc_sigmask and uc_sigmask64 of ucontext_t are an anonymous union
             "ucontext_t" => true,
-            // 'private' type
-            "prop_info" => true,
-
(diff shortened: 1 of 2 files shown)


Amanieu:
Changing the type to `c_void` seems incorrect. However you could define it as an empty enum.

Amanieu:
After thinking about it, I don't think this is the proper fix. `prop_info` itself isn't a private type, it's just that its value field is variable-length. Rather than changing the whole struct into an opaque type, we should just change its last field to be zero-length to match upstream.

Amanieu:
@joshtriplett What are your thoughts on this since you raised the issue?

Issue #2850, opened 2022-07-22
MSRV policy for libc crate is work in progress on the libs-team repo


workingjubilee:
It's difficult to observe potentially important maintenance policy decisions if they happen in another repo. That said, I don't think it's inherently more or less proper to have the discussion in
- https://github.com/rust-lang/libs-team/issues/72

to be clear. As has already been mentioned by one participant in that discussion, while this is clearly an important discussion for the libc crate, it may have ecosystem-wide impacts. Thus, treating it as a discussion for the libs team in general seems appropriate in other ways. Nonetheless, I raise it as an issue here anyways, if only to accomplish the feat of providing a redirect. It's a potential bug in the human interface of this crate, rather than the machine interface.

JohnTitor:
I've pinned this issue and retitled it for awareness, hope repo visitors find this.

JohnTitor:
Let's track this issue on #3248!

PR #2852, opened 2022-07-26
README.md: Clarify (lack of) MSRV policy

joshtriplett:
MSRV policy is still being discussed on an ongoing basis in
https://github.com/rust-lang/libs-team/issues/72 . In the interim,
clarify in README that libc doesn't currently have an MSRV policy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,9 @@ libc = "0.2"
 
 ## Rust version support
 
-The minimum supported Rust toolchain version is **Rust 1.13.0** . APIs requiring
+The minimum supported Rust toolchain version is currently **Rust 1.13.0**.
+(libc does not currently have any policy regarding changes to the minimum
+supported Rust version; such policy is a work in progress.) APIs requiring
 newer Rust features are only available on newer Rust toolchains:
 
 | Feature              | Version |


README.md
```diff
@@ -42,7 +42,9 @@ libc = "0.2"
 
 ## Rust version support
 
-The minimum supported Rust toolchain version is **Rust 1.13.0** . APIs requiring
+The minimum supported Rust toolchain version is currently **Rust 1.13.0**.
+(libc does not currently have any policy regarding changes to the minimum
+supported Rust version; such policy is a work in progress.) APIs requiring
```
cole-h, line 47: Would it be valuable to link to the specific issue where this policy is being discussed? Or maybe even a short description of "in the libc issues" or something, so people know where to go (if they don't come to this PR and find the backlink here, that is).

joshtriplett: @cole-h I don't think a link from the README to a github issue would be appropriate.

I would be happy to add some indication of where discussion is taking place, except that even that seems to be a matter of some consternation in places, and I'm concerned about such documentation being interpreted as normative (this is where the discussion *should* take place) rather than informative (this is where the discussion is in practice taking place).

JohnTitor (approved):
LGTM! Thank you for your hard work to set up MSRV policy <3

README.md
JohnTitor: For now, I've pinned https://github.com/rust-lang/libc/issues/2850 so that repo visitors can notice there's a discussion on the libs-team issue tracker.

cole-h: OK, makes sense! Thank you both!

joshtriplett: @JohnTitor Good idea!

PR #2859, opened 2022-08-03
libc 0.2.127

joshtriplett:


JohnTitor:
@Amanieu Note that this didn't bump up the libc-test version.

joshtriplett:
@JohnTitor Good catch.

Do we actually use that version number anywhere? Is there value in bumping it every time, or might it make sense to just set it to 0.0.1 and leave it that way?

I can send a PR if there's value in it.

JohnTitor:
@joshtriplett Some contributors use the libc-test crate on their CI, see https://github.com/rust-lang/libc/issues/2065 for the context.

joshtriplett:
@JohnTitor Ah, I see. Alright, will fix. (And when we can move MSRV forward far enough in the future, we can use workspace inheritance and have that version only updated in one place.)

Issue #2860, opened 2022-08-03
error LNK2019: unresolved external symbol printf


Tastaturtaste:
### Description
Building a program using printf does not work with the linker error _LNK2019: unresolved external symbol printf referenced in function ..._

### A Minimum Working Example without any dependencies (except libc) that shows the issue
```
use libc::{c_char, printf};
fn main() {
    unsafe {
        printf(b"Hello World!\n\0" as *const u8 as *const c_char);
    }
}
```
### The target triple
x86_64-pc-windows-msvc
### Instructions to reproduce
  uploads to github gists, etc.)

1. Create a new project with `cargo new reproduce`
2. `cd` ./reproduce
3. `cargo add libc`
4. copy the minimum example above in the main.rs file
5. `cargo run`
### Further information
[This issue](https://github.com/rust-lang/rust/issues/42830) apparently runs into the same problem.

From [this microsoft documentation](https://docs.microsoft.com/en-us/cpp/porting/visual-cpp-change-history-2003-2015?view=msvc-170#stdio_and_conio) it seems that the bug depends on the Windows SDK version used:

> If you are linking with Windows SDK 8.1 or earlier, you might encounter these unresolved external symbol errors. In that case, you should resolve the error by adding legacy_stdio_definitions.lib to the linker input as described previously.

I successfully tried the fix outlined in [this comment](https://github.com/rust-lang/rust/issues/42830#issuecomment-310484898) by separating out the affected `print`- and `scan`- family functions (currently there are only the two functions `printf` and `fprintf`) into their own `extern "C"` block and annotating that with `#[cfg_attr(all(windows, target_env = "msvc"), link(name = "legacy_stdio_definitions", kind = "dylib"))]`.

According to [this documentation from microsoft](https://docs.microsoft.com/en-us/cpp/porting/overview-of-potential-upgrade-issues-visual-cpp?view=msvc-170#libraries) the "legacy_stdio_definitions.lib" is only supplied with Visual Studio 15 and later. I don't know if there is a way to detect at compile time the Visual Studio version used in the toolchain and only conditionally apply the fix I outlined above.

PR #2861, opened 2022-08-03
Fix for https://github.com/rust-lang/libc/issues/2860

Tastaturtaste:
This PR is intended to fix https://github.com/rust-lang/libc/issues/2860. 
As indicated in the issue, this fix requires linking against "legacy_stdio_definitions.lib", [which is only provided with Visual Studio >=15](https://docs.microsoft.com/en-us/cpp/porting/overview-of-potential-upgrade-issues-visual-cpp?view=msvc-170#libraries). I don't know how this could be checked conditionally at compile time though.

### DIFF
--- a/src/windows/mod.rs
+++ b/src/windows/mod.rs
@@ -277,6 +277,16 @@ impl ::Clone for fpos_t {
     }
 }
 
+// Special handling for all print and scan type functions because of https://github.com/rust-lang/libc/issues/2860
+#[cfg_attr(
+    all(windows, target_env = "msvc"),
+    link(name = "legacy_stdio_definitions")
+)]
+extern "C" {
+    pub fn printf(format: *const c_char, ...) -> ::c_int;
+    pub fn fprintf(stream: *mut FILE, format: *const c_char, ...) -> ::c_int;
+}
+
 extern "C" {
     pub fn isalnum(c: c_int) -> c_int;
     pub fn isalpha(c: c_int) -> c_int;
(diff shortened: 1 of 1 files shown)


Amanieu:
It seems that we are linking against symbols that have been changed in the latest CRT? Could we change libc to link against the new symbols instead? The UCRT must expose some version of printf, right?

@rustbot ping windows

Amanieu:
cc @ChrisDenton

ChrisDenton:
Yeah. these aren't exported by the ucrt and haven't been for a long time. IIRC printf &co are implemented in the C headers as wrappers around [internal functions](https://docs.microsoft.com/en-us/cpp/c-runtime-library/internal-crt-globals-and-functions?view=msvc-170). So I guess the options are:

1. Leave it up to users to workaround this issue (aka the status quo).
2. Ignore the warnings and try to recreate the C headers in Rust using the internal symbols. This could break at any time.
3. Reimplement `prinf` in libc
4. Try to link `legacy_stdio_definitions.lib` which is a library that provide these (and other) symbols

Amanieu:
2 & 3 are straight out: we don't want this kind of maintenance burden on libc.

I think the best thing to do is to just link against `legacy_stdio_definitions`.

@bors r+

Tastaturtaste:
Thanks for the review!
It seems the task runner for Docker Linux Tier2 (sparc64-unknown-linux-gnu) failed with some kind of timeout:
> terminating on signal 15 from pid 2252 (timeout)

PR #2862, opened 2022-08-04
Add `gnu_get_[version|release]` for glibc

LegNeato:
See <https://sourceware.org/git/?p=glibc.git;a=blob;f=csu/version.c;h=8c0ed79c01223e1f12b54d19f90b5e5b7dd78d27;hb=HEAD#l51> and <http://refspecs.linux-foundation.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/libc-gnu-get-libc-version-1.html>

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2675,7 +2675,7 @@ fn test_linux(target: &str) {
                "elf.h",
                "fcntl.h",
                "glob.h",
-               "gnu/libc-version.h",
+               [gnu]: "gnu/libc-version.h",
                "grp.h",
                "iconv.h",
                "ifaddrs.h",
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2675,6 +2675,7 @@ fn test_linux(target: &str) {
                "elf.h",
                "fcntl.h",
                "glob.h",
+               "gnu/libc-version.h",
                "grp.h",
                "iconv.h",
                "ifaddrs.h",
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -1326,6 +1326,11 @@ extern "C" {
     pub fn malloc_trim(__pad: ::size_t) -> ::c_int;
 }
 
+extern "C" {
+    pub fn gnu_get_libc_release() -> *const ::c_char;
+    pub fn gnu_get_libc_version() -> *const ::c_char;
+}
+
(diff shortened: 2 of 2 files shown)


LegNeato:
I don't think this failure was caused by my code, any action on me?

libc-test/build.rs
```diff
@@ -2675,6 +2675,7 @@ fn test_linux(target: &str) {
                "elf.h",
                "fcntl.h",
                "glob.h",
+               "gnu/libc-version.h",
```
Amanieu, line 2678: It seems that the glibc versions we use in CI don't have this header file, which is causing the failure. Was this header added recently? If so then perhaps it would be better to add these symbols to the whitelist in libc-test/build.rs, with a note about which version of glibc they were introduced in.

LegNeato:
When I had the include path wrong it failed in CI. Once I had the correct include path it succeeded. So something in CI was able to pick it up.

The functions were introduced in version 2.1, which was [released in 1999](https://sourceware.org/glibc/wiki/Glibc%20Timeline):

<https://sourceware.org/git/?p=glibc.git;a=blob;f=csu/Versions;h=8e1b21948e8d7568c3e61caeaeb719cdaf110272;hb=HEAD#l6>

I don't see anything to suggest it is optional / can be turned off, though [the GNU manual notes](https://www.gnu.org/software/gnulib/manual/html_node/Glibc-gnu_002flibc_002dversion_002eh.html):

> This function is missing on some platforms: macOS 11.1, FreeBSD 13.0, NetBSD 9.0, OpenBSD 6.7, Minix 3.1.8, AIX 5.1, HP-UX 11, IRIX 6.5, Solaris 11.4, Cygwin 2.9, mingw, MSVC 14, Android 9.0.

It looks like alpine linux uses `ulibc`, which _does not_ include the header. Does `homu` use alpine linux as it's docker image base? If so that could account for the discrepancy...CI uses an image with glibc so the header was found and passed, `homu` uses alpine so it can't find the header and the merge fails.

libc-test/build.rs
Amanieu: Ah I see the problem, it's failing on a musl target where this header doesn't exist. You need to prefix it with `[gnu]:`:
```
[gnu]: "gnu/libc-version.h",
```

LegNeato: Ah cool, I'll get to this in a day or so, thanks for finding the issue!

PR #2863, opened 2022-08-04
Bump libc-test version to match libc

joshtriplett:


JohnTitor:
Could you also update the version of the libc dependency? It rarely happens but just in case I'd avoid a potential version mismatching.

joshtriplett:
On Thu, Aug 04, 2022 at 02:05:40AM -0700, Yuki Okushi wrote:
> Could you also update the version of the libc dependency? It rarely happens but just in case I'd avoid a potential version mismatching.

Done.

JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

JohnTitor:
@Amanieu Could you take care of the new release?

PR #2864, opened 2022-08-04
Socket constants necessary for ESP-IDF support in socket2

ivmarkov:
I would like to upstream the support for [ESP-IDF](https://github.com/espressif/esp-idf) in the [socket2 crate](https://github.com/esp-rs-compat/socket2). (Which in turn is necessary for Rust async networking support on top of ESP-IDF.)

To tidy up the patch and remove [this](https://github.com/rust-lang/socket2/compare/master...esp-rs-compat:socket2:master#diff-bbdb7821e30b05ce3afbaa9fbc1526c18d6670d65479baaca3beaf4ea6174891R80) and [this](https://github.com/rust-lang/socket2/compare/master...esp-rs-compat:socket2:master#diff-bbdb7821e30b05ce3afbaa9fbc1526c18d6670d65479baaca3beaf4ea6174891R413) TODOs I need these constants defined in the `libc` crate

### DIFF
--- a/src/unix/newlib/espidf/mod.rs
+++ b/src/unix/newlib/espidf/mod.rs
@@ -83,6 +83,9 @@ pub const MSG_DONTROUTE: ::c_int = 0x4;
 pub const MSG_WAITALL: ::c_int = 0x02;
 pub const MSG_MORE: ::c_int = 0x10;
 pub const MSG_NOSIGNAL: ::c_int = 0x20;
+pub const MSG_TRUNC: ::c_int = 0x04;
+pub const MSG_CTRUNC: ::c_int = 0x08;
+pub const MSG_EOR: c_int = 0x08;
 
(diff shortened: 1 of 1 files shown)


ivmarkov:
Note that I've also added the signature of `eventfd` to the PR (yes - ESP-IDF supports an [API](https://github.com/espressif/esp-idf/blob/master/components/vfs/include/esp_vfs_eventfd.h#L67) similar to Linux's `eventfd`). 

This is necessary for [upstreaming my changes to the `polling` crate](https://github.com/smol-rs/polling/compare/master...esp-rs-compat:polling:master#diff-9bc7469460a924fbd3fb3bb8e3adb149f4f4f9f1f641d12e11cea2974542336dR107) which is also part of my general effort to upstream the Rust async support for ESP-IDF.

PR #2867, opened 2022-08-09
Bump up to v0.2.129

JohnTitor:
... because v0.2.128 has a Cargo bug: https://github.com/rust-lang/libc/issues/2866

Signed-off-by: Yuki Okushi <jtitor@2k36.org>


aganders3:
Thanks for the quick fix!

PR #2868, opened 2022-08-10
Bump to 0.2.130

teamplayer3:
Quick fix for target `riscv32imc-esp-espidf`.

PR #2869, opened 2022-08-10
ESP-IDF - add missing double-colon

ivmarkov:
The previous PR was having one of the new `const`s wrongly defined which broke the ESP-IDF build.


Amanieu:
@bors r-

I'm going to merge #2868 instead.

PR #2874, opened 2022-08-14
Add FUTEX_LOCK_PI2 on Linux

rtzoeller:
Tested using a modified version of the [linux-futex crate](https://github.com/m-ou-se/linux-futex/).


rtzoeller:
Hmm, `FUTEX_LOCK_PI2` was [added to the Linux kernel in 5.14](https://man7.org/linux/man-pages/man2/futex.2.html), but the Docker images are using Ubuntu Focal which uses an older kernel/headers.

What's the best way to handle this?

JohnTitor (approved):

PR #2876, opened 2022-08-16
linux: add missing netfilter definitions

phi-gamma:
Bring ``libc`` up to parity with Linux v5.18 headers wrt. to Netfilter.

``libc-test`` fails due to unrelated issues on my Fedora box but the
changes should be inocuous enough IMO.

Note there’s a warning in the code indicating potential trouble due to [this
issue](https://github.com/rust-lang/libc/pull/1628). I’ll sort this out if
the CI run fails.

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -2086,6 +2086,7 @@ pub const NFNLGRP_CONNTRACK_EXP_UPDATE: ::c_int = 5;
 pub const NFNLGRP_CONNTRACK_EXP_DESTROY: ::c_int = 6;
 pub const NFNLGRP_NFTABLES: ::c_int = 7;
 pub const NFNLGRP_ACCT_QUOTA: ::c_int = 8;
+pub const NFNLGRP_NFTRACE: ::c_int = 9;
 
 pub const NFNETLINK_V0: ::c_int = 0;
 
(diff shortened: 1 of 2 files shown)


phi-gamma:
The “Docker Linux Tier1” CI run fails like so:

    cargo:warning=/checkout/target/i686-unknown-linux-gnu/debug/build/libc-test-4eae72712e4eb4b5/out/main.c:30472:66: error: 'NFNL_SUBSYS_HOOK' undeclared here (not in a function); did you mean 'NFNL_SUBSYS_ULOG'?

So the caveat about musl relying on pre-historic kernel headers
still seems to apply. Could this change be made a compile-time
conditional? To it sounds rather suboptimal to let one specific
target hold back the crate like this.

Let’s see about those other changes.

Amanieu:
You can add exceptions to the tests in `libc-test/build.rs`. You will see many other cases where exceptions are added due to old kernel headers in CI. Make sure to mark the kernel version the constants were introduced in so we can remove the exceptions when upgrading the CI images.

phi-gamma:
2022-08-16 (Tuesday), ***@***.*** (Amanieu):

> You can add exceptions to the tests in `libc-test/build.rs`.
> You will see many other cases where exceptions are added due to
> old kernel headers in CI. Make sure to mark the kernel version
> the constants were introduced in so we can remove the exceptions
> when upgrading the CI images.

Thanks for the pointer, I’ll give it a shot and resubmit!

src/unix/linux_like/linux/mod.rs
```diff
@@ -2101,15 +2102,31 @@ pub const NFNL_SUBSYS_CTNETLINK_TIMEOUT: ::c_int = 8;
...
+cfg_if! {
+    if #[cfg(target_env = "musl")] {
+        // FIXME: CI kernel headers lag bit behind;
+        // NFNL_SUBSYS_HOOK was added with v5.14.
+        pub const NFNL_SUBSYS_COUNT: ::c_int = 12;
+    } else {
+        pub const NFNL_SUBSYS_HOOK: ::c_int = 12;
+        pub const NFNL_SUBSYS_COUNT: ::c_int = 13;
+    }
+}
```
Amanieu, line 2114: The exception in build.rs should apply to both glibc and musl, so you shouldn't need to add this cfg_if.

phi-gamma: True. I could swear I submitted that in an earlier revision and it resulted in a CI failure,
prompting me to add that ``cfg_if``. Anyways, CI seems to be happy now without it.

phi-gamma:
2022-08-20 (Saturday), ***@***.*** (bors):

> :broken_heart: Test failed - [checks-actions](https://github.com/rust-lang/libc/runs/7931592633?check_suite_focus=true)
> 

Sigh. Those were introduced with v5.4. Confusingly only
``mips-unknown-linux-gnu`` seems to be affected by this but not
other platforms with older kernels.

Exceptions for those ``NFULA_*`` constants added.

phi-gamma:
> S-waiting-on-author

@JohnTitor could you clarify what you’re waiting for? I updated the pr to address
the issues that became apparent during the bors run.

JohnTitor:
Oh, missed you pushed a commit after the failure, @bors r+

phi-gamma:
> Oh, missed you pushed a commit after the failure, @bors r+

No worries, thank you!

phi-gamma:
> > @@ -2101,15 +2102,31 @@ pub const NFNL_SUBSYS_CTNETLINK_TIMEOUT: ::c_int = 8;
>  pub const NFNL_SUBSYS_CTHELPER: ::c_int = 9;
>  pub const NFNL_SUBSYS_NFTABLES: ::c_int = 10;
>  pub const NFNL_SUBSYS_NFT_COMPAT: ::c_int = 11;
> -pub const NFNL_SUBSYS_COUNT: ::c_int = 12;
> +cfg_if! {
> +    if #[cfg(target_env = "musl")] {
> +        // FIXME: CI kernel headers lag bit behind;
> +        // NFNL_SUBSYS_HOOK was added with v5.14.
> +        pub const NFNL_SUBSYS_COUNT: ::c_int = 12;
> +    } else {
> +        pub const NFNL_SUBSYS_HOOK: ::c_int = 12;
> +        pub const NFNL_SUBSYS_COUNT: ::c_int = 13;
> +    }
> +}
> 
> The exception in build.rs should apply to both glibc and musl, so you shouldn't need to add this cfg_if.

I tried that initially but without the ``cfg_if`` the
*x86_64-unknown-linux-gnu* run fails because of the value
mismatch for ``NFNL_SUBSYS_COUNT``.

PR #2883, opened 2022-08-19
Add `confstr` and guaranteed `_CS_*` constants on apple

thomcc:
I actually just want `_CS_DARWIN_USER_TEMP_DIR`, and the `confstr` function, but I also have added the other "guaranteed" `_CS_*` values (e.g. present in the [output of `man confstr`](https://gist.github.com/thomcc/12a27c7998019b6da9fa4b539bdbca44)).

This is apparently a [POSIX API](https://pubs.opengroup.org/onlinepubs/9699919799/functions/confstr.html) but I have no idea which other platforms support it, nor what values should be provided. It's somewhat important on Darwin (the only way to access the user temp dir), but I'm not sure what other use it has, so I've only provided it on Darwin.

There are a few other constants in Darwin `unistd.h`, but they're omitted since they're only conditionally present, and they provide values that... don't actually work[^1]. As a result, I've only provided the ones documented by `confstr` documentation.

[^1]: For example, `_CS_POSIX_V6_LPBIG_OFFBIG_CFLAGS` nominally provides values that can be passed to a c compiler, but it gives `-W 64` for me, which does not seem to work with `clang`.

### DIFF
--- a/libc-test/semver/apple.txt
+++ b/libc-test/semver/apple.txt
@@ -1507,6 +1507,10 @@ XATTR_SHOWCOMPRESSION
 XUCRED_VERSION
 YESEXPR
 YESSTR
+_CS_PATH
+_CS_DARWIN_USER_DIR
+_CS_DARWIN_USER_TEMP_DIR
+_CS_DARWIN_USER_CACHE_DIR
 _IOFBF
 _IOLBF
 _IONBF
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #2889, opened 2022-08-24
linux: add rtnetlink mcast group definitions

phi-gamma:
These come in two flavors:

- RTNLGRP_* are bit indexes (arguments to setsockopt(2)) as used
  in the kernel with test_bit() but also userspace (see libnl
  examples).

- RTMGRP_* are bitmasks not used in the kernel; their use
  seems to be deprecated, at least according to comments in
  libnl, but documentation still references them.

The rationale for adding these definitions is that they're needed
to subscribe to kernel events via Netlink multicast groups.

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -2669,6 +2669,62 @@ pub const ARPD_LOOKUP: ::c_ushort = 0x02;
 pub const ARPD_FLUSH: ::c_ushort = 0x03;
 pub const ATF_MAGIC: ::c_int = 0x80;
 
+// userspace compat definitions for RTNLGRP_*
+pub const RTMGRP_LINK: ::c_int = 0x00001;
+pub const RTMGRP_NOTIFY: ::c_int = 0x00002;
+pub const RTMGRP_NEIGH: ::c_int = 0x00004;
+pub const RTMGRP_TC: ::c_int = 0x00008;
+pub const RTMGRP_IPV4_IFADDR: ::c_int = 0x00010;
+pub const RTMGRP_IPV4_MROUTE: ::c_int = 0x00020;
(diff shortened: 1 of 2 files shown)


src/unix/linux_like/linux/mod.rs
```diff
@@ -2669,6 +2669,64 @@ pub const ARPD_LOOKUP: ::c_ushort = 0x02;
...
+pub const RTNLGRP_MPLS_NETCONF: ::c_uint = 0x1d;
+pub const RTNLGRP_IPV4_MROUTE_R: ::c_uint = 0x1e;
+pub const RTNLGRP_IPV6_MROUTE_R: ::c_uint = 0x1f;
+pub const RTNLGRP_NEXTHOP: ::c_uint = 0x20;
+pub const RTNLGRP_BRVLAN: ::c_uint = 0x21;
+pub const RTNLGRP_MCTP_IFADDR: ::c_uint = 0x22;
+pub const RTNLGRP_TUNNEL: ::c_uint = 0x23;
+pub const RTNLGRP_STATS: ::c_uint = 0x24;
+pub const __RTNLGRP_MAX: ::c_uint = 0x25;
+pub const RTNLGRP_MAX: ::c_uint = 0x24;
```
JohnTitor, line 2728: If these values are fairly easy to be updated and would cause a breaking change, I guess it doesn't fit with the libc crate (we generally don't update such a value as we don't have any supporting policy for Linux kernel version, etc. So it might be inconvenient to use after a while).

phi-gamma: Makes sense, thanks for clarifying. These constants are gone from the latest iteration.

JohnTitor:
Thanks! @bors r+

phi-gamma:
Thanks @JohnTitor @Amanieu !

PR #2891, opened 2022-08-26
Add makedev for the BSDs

asomers:
Also, make Linux's makedev function safe and const.

### DIFF
--- a/libc-test/Cargo.toml
+++ b/libc-test/Cargo.toml
@@ -67,6 +67,11 @@ path = "test/cmsg.rs"
 harness = true
 
 [[test]]
+name = "makedev"
+path = "test/makedev.rs"
+harness = true
+
+[[test]]
 name = "errqueue"
 path = "test/errqueue.rs"
 harness = true
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -22,6 +22,15 @@ fn do_cc() {
             }
             cmsg.compile("cmsg");
         }
+
+        if target.contains("linux")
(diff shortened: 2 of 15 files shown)


asomers:
cc @musikid

semarie:
LGTM for openbsd.

just a question about the type used for the function definition: why using `c_uint` (and casting it to `dev_t`) instead of `dev_t` directly ?

asomers:
> LGTM for openbsd.
> 
> just a question about the type used for the function definition: why using `c_uint` (and casting it to `dev_t`) instead of `dev_t` directly ?

Just because that's what the existing Linux definition does.

JohnTitor (approved):

JohnTitor:
👍, @bors r+

asomers:
@JohnTitor I think this should fix the bors error.  But I don't know how to run the tests on emscripten except by asking bors to merge.

JohnTitor:
We could do `@bors try` but I think it's okay to r+ in this case.
@bors r+

asomers:
@JohnTitor I think if you do " bors delegate=asomers" then that will give me permission to use "bors try".  It'll also give me permission to merge the PR, but I promise not to do that unless you say so ;) .

JohnTitor:
👍, @bors delegate+

asomers:
bors try

asomers:
@JohnTitor it finally passes CI, I've squashed my commits.

JohnTitor:
Thanks! @bors r+

PR #2900, opened 2022-09-05
fcntl F_DUPFD_CLOEXEC has different values on Solaris and Illumos

gco:


src/unix/solarish/illumos.rs
```diff
@@ -42,6 +42,9 @@ pub const F_OFD_SETLKW: ::c_int = 49;
 pub const F_FLOCK: ::c_int = 53;
 pub const F_FLOCKW: ::c_int = 54;
 
+// fcntl F_DUPFD_CLOEXEC has different values on Solaris and Illumos
```
jclulow, line 45:
```suggestion
```

I would remove the comments.  The separate file structure already denotes different constants for illumos and Solaris, and the file would be full of comments like this if we noted every difference.

src/unix/solarish/solaris.rs
```diff
@@ -36,6 +36,9 @@ pub const TCP_KEEPIDLE: ::c_int = 0x1d;
 pub const TCP_KEEPCNT: ::c_int = 0x1e;
 pub const TCP_KEEPINTVL: ::c_int = 0x1f;
 
+// fcntl F_DUPFD_CLOEXEC has different values on Solaris and Illumos
```
jclulow, line 39:
```suggestion
```

jclulow (changes requested):

jclulow (approved):
Looks good to me -- thanks!

JohnTitor (approved):

PR #2901, opened 2022-09-05
Solaris TCP_KEEPINTVL and TCP_KEEPCNT have wrong values

gco:


pfmooney:
I would have expected `libc-test` to catch this (I don't see that we're excluding those constants).  Do the tests run clean with the change?

pfmooney:
> I would have expected `libc-test` to catch this (I don't see that we're excluding those constants). Do the tests run clean with the change?

Sorry, I misread this as being the top-level (solarish) ones.  Carry on!

gco:
> I would have expected libc-test to catch this (I don't see that we're excluding those constants). Do the tests run clean with the change?

Yeah-- nobody has run libc-test on Solaris recently, it doesn't work at all.  Maybe a PR for that later.

pfmooney:
> > I would have expected libc-test to catch this (I don't see that we're excluding those constants). Do the tests run clean with the change?
> 
> Yeah-- nobody has run libc-test on Solaris recently, it doesn't work at all. Maybe a PR for that later.

Understandable.  It was not a small list to get it running cleanly on illumos.  If it will require a significant delta from the existing combined "solarish" test setup, it might be worth splitting the Solaris test stuff out too (just as was done for the actual library sources)

Regardless, thanks for chasing down these issues.

JohnTitor:
Are there any public resources to check these values on Solaris?

gco:
> Are there any public resources to check these values on Solaris?

It's unfortunate, but installing Solaris in a VM is probably the most straightforward way.

[Oracle Solaris 11.4 CBE](https://blogs.oracle.com/solaris/post/announcing-the-first-oracle-solaris-114-cbe) (common build environment) is free to use for personal use and free/open-source developers.  

FWIW, libc-test does report the following (amongst many other complaints) after some effort getting it to run:

```
bad TCP_KEEPCNT value at byte 0: rust: 30 (0x1e) != c 31 (0x1f)
bad TCP_KEEPINTVL value at byte 0: rust: 31 (0x1f) != c 30 (0x1e)
```

JohnTitor:
Thanks, hmm, then adding it to libc-test would be the best way here. For now, I'm just going to r+ this, thanks again for fixing them!

JohnTitor (approved):

PR #2902, opened 2022-09-06
apple: Add the attrlist set of functions

beetrees:
Add the attrlist set of functions for apple platforms.


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

src/unix/bsd/apple/mod.rs
```diff
@@ -5562,6 +5737,58 @@ extern "C" {
     ) -> ::sysdir_search_path_enumeration_state;
 
     pub static vm_page_size: vm_size_t;
+
+    pub fn getattrlist(
+        path: *const ::c_char,
+        attrList: *mut ::c_void,
```
Dr-Emann, line 5743: Per the [man page](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/getattrlist.2.html), I think the `attrList` parameters for these functions should be `*mut attrlist`?

Dr-Emann: Ah, yes, does look like the header has it declared as `void *`. Womp. Oh well, just means I need an extra `ptr.cast()`.

PR #2904, opened 2022-09-09
Add new Linux system call constants

cd-work:
This adds the latest Linux system call constants based on the Kernel's system call table:

https://github.com/torvalds/linux/blob/master/arch/x86/entry/syscalls/syscall_64.tbl


cd-work:
I have no idea how the process for adding new system call constants generally works, but I saw that https://github.com/rust-lang/libc/pull/2114 seemingly just added all the stuff manually and decided to do the same.

I personally need the `landlock` system calls, so that's my motivation for the PR. So I thought I might as well upstream these changes.

I've only checked the system call table for x86_64 and extrapolated all the other architectures from there. It seems like for existing syscalls they're mostly the same with some constant offset, so I've tried to just keep that pattern going.

JohnTitor (approved):
If you're sure they are valid on each arch and aren't supposed to be changed in the long term, I think it's totally fine.
Thanks for contributing!

cd-work:
The build failed on a component download on Windows, so I don't think it's related to this PR.

PR #2914, opened 2022-09-18
fix wrong definitions of getpwent_r and getgrent_r on solarish os

### LINKED ISSUE #2908: Wrong `getpwent_r` and `getgrent_r` definitions on illumos and solaris
Definitions according to their man pages:
```c
struct passwd *getpwent_r(struct passwd *pwd, char  *buffer, int buflen);
struct group *getgrent_r(struct group  *grp,  char  *buffer, int bufsize);
```
* [solaris getpwent_r man page](https://sites.google.com/a/diviware.com/unix/unix-page/solaris-page/solaris-man-pages/3c/getpwent_r)
* [solaris getgrent_r man page](https://sites.google.com/a/diviware.com/unix/unix-page/solaris-page/solaris-man-pages/3c/getgrent_r)
* [illumos getpwent_r man page](https://illumos.org/man/3C/getpwent)
* [illumos getgrent_r man page](https://illumos.org/man/3C/getgrent)

`libc` exposes them as:

```rust
pub unsafe extern "C" fn getgrent_r(
    grp: *mut group,
    buf: *mut c_char,
    buflen: size_t,
    result: *mut *mut group
) -> c_int

pub unsafe extern "C" fn getpwent_r( 
    pwd: *mut passed, 
    buf: *mut c_char, 
    buflen: size_t, 
    result: *mut *mut passed
) -> c_int
```

SteveLauC:
Closes #2908 

* [man page for `getpwent_r`](https://illumos.org/man/3C/getpwnam)
* [man page for `getgrent_r`](https://illumos.org/man/3C/getgrnam)

You may find the definitions for `getpwnam_r/getpwuid_r/getgrnam_r/getgruid_r` exposed by `libc` are also wrong:
```c
struct passwd *getpwnam_r(const char *name, struct passwd *pwd,
            char *buffer, int buflen);
```
```rust
pub fn getpwnam_r(
    name: *const ::c_char,
    pwd: *mut passwd,
    buf: *mut ::c_cha
    buflen: ::size_t,
    result: *mut *mut passwd,
) -> ::c_int;
```
But actually they are **correct** as there are the POSIX-conforming definitions (see `Standard conforming` section of above man pages):
```
Standard conforming

       cc [ flag...] file... -D_POSIX_PTHREAD_SEMANTICS [ library... ]

       int getpwnam_r(const char *name, struct passwd *pwd, char *buffer,
            size_t bufsize, struct passwd **result);


       int getpwuid_r(uid_t uid, struct passwd *pwd, char *buffer,
            size_t bufsize, struct passwd **result);
```

`getpwent_r/getgrent_r` don't get lucky, they do not have the POSIX-conforming alternatives.

To double check this, I searched its [source code](https://github.com/illumos/illumos-gate/blob/master/usr/src/lib/libc/port/gen/getpwnam_r.c):
```shell
$ rg "__posix_getpwnam_r"
port/mapfile-vers
1582:   __posix_getpwnam_r;

port/gen/getpwnam_r.c
152:__posix_getpwnam_r(const char *name, struct passwd *pwd, char *buffer,

$ rg "__posix_getpwent_r"

$
```

### DIFF
--- a/src/unix/solarish/compat.rs
+++ b/src/unix/solarish/compat.rs
@@ -1,6 +1,7 @@
 // Common functions that are unfortunately missing on illumos and
 // Solaris, but often needed by other crates.
 
+use core::cmp::min;
 use unix::solarish::*;
 
 const PTEM: &[u8] = b"ptem\0";
@@ -168,4 +169,52 @@ pub unsafe fn forkpty(
     }
 
     0
+}
+
+pub unsafe fn getpwent_r(
+    pwd: *mut passwd,
+    buf: *mut ::c_char,
+    buflen: ::size_t,
+    result: *mut *mut passwd,
+) -> ::c_int {
+    let old_errno = *::___errno();
+    *::___errno() = 0;
+    *result = native_getpwent_r(
+        pwd,
+        buf,
+        min(buflen, ::c_int::max_value() as ::size_t) as ::c_int,
+    );
+
+    let ret = if (*result).is_null() {
+        *::___errno()
+    } else {
+        0
+    };
+    *::___errno() = old_errno;
+
+    ret
+}
+
+pub unsafe fn getgrent_r(
+    grp: *mut ::group,
+    buf: *mut ::c_char,
+    buflen: ::size_t,
+    result: *mut *mut ::group,
+) -> ::c_int {
+    let old_errno = *::___errno();
+    *::___errno() = 0;
+    *result = native_getgrent_r(
+        grp,
+        buf,
+        min(buflen, ::c_int::max_value() as ::size_t) as ::c_int,
+    );
+
+    let ret = if (*result).is_null() {
+        *::___errno()
+    } else {
+        0
+    };
+    *::___errno() = old_errno;
+
+    ret
 }
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -3016,24 +3016,14 @@ extern "C" {
     ) -> ::c_int;
     #[cfg_attr(
         any(target_os = "solaris", target_os = "illumos"),
-        link_name = "__posix_getpwent_r"
+        link_name = "getpwent_r"
     )]
-    pub fn getpwent_r(
-        pwd: *mut passwd,
-        buf: *mut ::c_char,
-        buflen: ::size_t,
-        result: *mut *mut passwd,
-    ) -> ::c_int;
+    fn native_getpwent_r(pwd: *mut passwd, buf: *mut ::c_char, buflen: ::c_int) -> *mut passwd;
     #[cfg_attr(
         any(target_os = "solaris", target_os = "illumos"),
-        link_name = "__posix_getgrent_r"
+        link_name = "getgrent_r"
     )]
-    pub fn getgrent_r(
-        grp: *mut ::group,
-        buf: *mut ::c_char,
-        buflen: ::size_t,
-        result: *mut *mut ::group,
-    ) -> ::c_int;
+    fn native_getgrent_r(grp: *mut ::group, buf: *mut ::c_char, buflen: ::c_int) -> *mut ::group;
     #[cfg_attr(
         any(target_os = "solaris", target_os = "illumos"),
         link_name = "__posix_sigwait"


JohnTitor:
This is a breaking change, did you make sure that the change doesn't break any code? 

cc @ctrlcctrlv who originally added this

ctrlcctrlv:
I mean, I have not tested personally on Solaris, but it looks right to me.

SteveLauC:
> This is a breaking change, did you make sure that the change doesn't break any code?

Not sure about this. So what are we going to do about this PR?

JohnTitor:
> Not sure about this. So what are we going to do about this PR?

The libc crate follows SemVer and changing declarations will break existing code (unless it's totally wrong and doesn't work at all), we have to confirm that the change here doesn't confuse anyone.

ctrlcctrlv:
I'm to blame for this via #934. I have two questions.

1. Why isn't the right solution to add POSIX conforming versions of these functions to Illumos?
2. Why did CI pass even though `__posix_get(pw|gr)ent_r` doesn't exist?

ctrlcctrlv:
Well, failing an answer to №2, I've gotten OmniOS (an Illumos distribution) working on my own. I suggest I just author a patch to Illumos to add the missing functions, making this issue moot. Does anyone have an opinion on that course of action?

SteveLauC:
> Does anyone have an opinion on that course of action?

Sounds good to me:)

jclulow:
I've had a look at this, and I suspect that the right call is probably to implement, in Rust, the behaviour you're expecting based on the other platforms as a wrapper around the native function.  I don't believe there's any substantial difference otherwise, so you could presumably do something a bit like this:

```rust
extern "C" {
    #[cfg_attr(link_name  = "getpwent_r")]
    pub fn native_getpwent_r(pwd: *mut passwd, buf: *mut ::c_char, buflen: ::c_int) -> *mut passwd;
}

pub unsafe fn getpwent_r(
    pwd: *mut passwd,
    buf: *mut ::c_char,
    buflen: ::size_t,
    result: *mut *mut passwd,
) -> ::c_int {
    let res = native_getpwent_r(passwd, buf, buflen.try_into().unwrap());
    if res.is_null() {
        -1
    } else {
        *result = res;
        0
    }
}
```

This would go in the [`src/unix/solarish/compat.rs`](https://github.com/rust-lang/libc/blob/master/src/unix/solarish/compat.rs) file, alongside the handful of similar routines that are present there.

We can and likely eventually will figure out a path for reworking these routines in illumos, but it will take time and coordination.  Even when that lands it will be months or years before users have updated all their machines past LTS releases of various distributions to get the new symbol available.

SteveLauC:
@jclulow's walkaround looks good and I just added it. Any opinion on this?

(friendly ping @JohnTitor)

src/unix/solarish/compat.rs
```diff
@@ -169,3 +169,42 @@ pub unsafe fn forkpty(
 
     0
 }
+/// Standard conforming wrapper around `native_getpwent_r()`
+///
+/// TODO: remove this entry and rename `native_getpwent_r` to `getpwent_r` in
+/// `mod.rs` when the standard conforming definitoin is added to the upstream
+/// kernel
```
jclulow, line 176: I would probably remove this whole doc comment and the matching one on the next function, for a few reasons:

- I don't think we usually put doc comments on any public symbols in the libc crate?
- It won't actually be strictly necessary to do this, as the existing `getpwent_r()` will never go away or change behaviour (illumos does not break backwards compatibility for public symbols) and in fact it may be counter productive to do it too soon (for the same reason we're adding it in the first place)
- It is not actually a function that conforms to any standard of which I'm aware; rather it's a GNU extension with regrettably different arguments

src/unix/solarish/mod.rs
```diff
@@ -3016,24 +3016,18 @@ extern "C" {
...
-        link_name = "__posix_getpwent_r"
+        link_name = "getpwent_r"
     )]
-    pub fn getpwent_r(
-        pwd: *mut passwd,
-        buf: *mut ::c_char,
-        buflen: ::size_t,
-        result: *mut *mut passwd,
-    ) -> ::c_int;
+    pub fn native_getpwent_r(pwd: *mut passwd, buf: *mut ::c_char, buflen: ::c_int) -> *mut passwd;
```
jclulow, line 3021: It's not clear to me if `native_getpwent_r()` and `native_getgrent_r()` should be `pub` or perhaps rather `pub(crate)` -- or maybe no access qualifier is necessary because they're only used from a child module?  If they're `pub` it would seem that we'd have to keep exposing them forever, because removing them would then break libc crate backwards compatibility.

jclulow (commented):
Thanks for taking a swing at this!  I have put a few thoughts in comments.

src/unix/solarish/mod.rs
SteveLauC: I removed the `pub` qualifier since these two functions are only used inside this child module

src/unix/solarish/compat.rs
```diff
@@ -169,3 +170,33 @@ pub unsafe fn forkpty(
...
     0
 }
+
+pub unsafe fn getpwent_r(
+    pwd: *mut passwd,
+    buf: *mut ::c_char,
+    buflen: ::size_t,
+    result: *mut *mut passwd,
+) -> ::c_int {
+    let res = native_getpwent_r(pwd, buf, buflen.try_into().unwrap());
```
JohnTitor, line 180: I haven't checked deeply but this `TryInto` may break our MSRV?

JohnTitor (commented):
If this doesn't introduce any breaking change (i.e. ABI change, MSRV breaking, liking change), I'm fine to accept, has anyone confirmed that?

src/unix/solarish/compat.rs
jclulow: It might be better just to clamp the value at `::c_int::MAX`.  It's not incorrect to do so, because we would just not be using the entire (apparently massive) buffer passed by the user.

SteveLauC: Yes, `TryInto` needs rust 1.34 or above. The current MSRV is 1.13.

@jclulow Would you like to test this on illumos? (I am sorry I haven't got a VM to work...) Normally this bufsize is obtained through `sysconf(_SC_GETPW/GR_R_SIZE_MAX)` at runtime. If the value is in the range of `int` on illumos, then we can use `as` to cast it here

src/unix/solarish/compat.rs
```diff
@@ -169,3 +170,33 @@ pub unsafe fn forkpty(
...
+    if res.is_null() {
+        -1
+    } else {
+        *result = res;
+        0
+    }
+}
```
jclulow, line 187: Hey, sorry about this, but I was reading the manual page again today and I think I missed an edge case in my original suggestion.  Critically, when there are no more entries, for glibc:

```
RETURN VALUE
       On success, these functions return 0 and *pwbufp is a  pointer  to  the
       struct  passwd.   On  error,  these functions return an error value and
       *pwbufp is NULL.

ERRORS
       ENOENT No more entries.
```

Contrast this with the manual for the illumos version:

```
RETURN VALUES
...
       Applications wishing to check for error situations should set errno to
       0 before calling getpwnam(), getpwnam_r(), getpwuid(), getpwuid_r(),
       getpwent(), getpwent_r(), fgetpwent(), and fgetpwent_r(). If these non-
       reentrant functions return a null pointer and errno is non-zero, an
       error occurred.
```

So I think we need to clear `errno` ourselves here, and return `ENOENT` explicitly; i.e., something like:

```rust
*___errno() = 0;
let res = native_getpwent_r(pwd, buf, buflen.try_into().unwrap());
if res.is_null() {
    if *___errno() == 0 {
        *___errno() = ENOENT;
    }
    -1
} else {
    *result = res;
    0
}

File: src/unix/solarish/compat.rs (hunk 3 above)
Comment (jclulow) [reply src/unix/solarish/compat.rs:180]: I think we only care about the type, not what values users might or might not pass, so probably something like this would work:

```rust
buflen.min(::c_int::MAX as ::size_t) as ::c_int
```

That way I think you would be able to pass any legal value up to and including the maximum `int` value.  The argument in the illumos libc version is definitely an `int`:

```
struct passwd *
getpwent_r(struct passwd *result, char *buffer, int buflen)
{
        nss_XbyY_args_t arg;
        char            *nam;
...
```

jclulow: And yes, either @pfmooney or myself can definitely test the final patch on an illumos system!

src/unix/solarish/compat.rs
```diff
@@ -169,3 +170,33 @@ pub unsafe fn forkpty(
...
+    buflen: ::size_t,
+    result: *mut *mut ::group,
+) -> ::c_int {
+    let res = native_getgrent_r(grp, buf, buflen.try_into().unwrap());
+    if res.is_null() {
+        -1
+    } else {
+        *result = res;
+        0
+    }
```
jclulow, line 201: Same sort of change required for this, as the convention with `getgrent_r()` is the same:

```
       Applications wishing to check for error situations should set errno to
       0 before calling getgrnam(), getgrnam_r(), getgrent(),
       getgrent_r()getgrgid(), getgrgid_r(), fgetgrent(), and fgetgrent_r().
       If these functions return a null pointer and errno is non-zero, an
       error occurred.
```

src/unix/solarish/compat.rs
SteveLauC: Yes, glibc returns 0 and sets `errno` to `ENOENT` when there are no more entries. But different UNIX (or UNIX-like) OSes actually have different behaviors in this case. For example:

* freebsd: [getpwent_r man page](https://www.freebsd.org/cgi/man.cgi?query=getpwent&sektion=3)

   On success: return 0, `result` will be non-NULL
   On error: return the error number, `result` will be NULL
   End: return 0, `result` will be NULL.

* dragonfly: [getpwent_r man page](https://leaf.dragonflybsd.org/cgi/web-man?command=getpwent_r&section=ANY)

  On success: return 0 and `result` will be non-NULL
  On error: return the error number, `result` will be NULL
  End: return 0, `result` will be NULL
  > The behavior on freebsd and dragonfly are the same.

* netbsd: [getpwent_r man page](https://man.netbsd.org/getpwent_r.3)

  On success: return 0, `result` will be non-NULL,
  On error: return a non-zero value (`errno` is set), `result` will be NULL
  End: return 0, `result` will be NULL

* Linux: [getpwent_r man page](https://man7.org/linux/man-pages/man3/getpwent_r.3.html)

  On success: return 0, `result` will be non-NULL
  On error: return the error number (**not ENOENT**), `result` will be NULL
  End: return `ENOENT` and `result` will be NULL.

Are we going to adopt the glibc one?

SteveLauC: If we follow the glibc one, then the impl will be something like:

```rust
pub unsafe fn getpwent_r(
    pwd: *mut passwd,
    buf: *mut ::c_char,
    buflen: ::size_t,
    result: *mut *mut passwd,
) -> ::c_int {
    * ::___errno() = 0;
    let res = native_getpwent_r(pwd, buf, buflen.min(::c_int::MAX as ::size_t) as ::c_int);
    if res.is_null() {
        *result = core::ptr::null_mut();
        if * ::___errno() == 0 {
            return ENOENT;
        } else {
            return * ::___errno();
        }
    } else {
        *result = res;
        0
    }
}
```

jclulow: Oh bother.  Thank you for comparing the BSDs, which I had not done.  That is somewhat unfortunate!  The FreeBSD behaviour seems to be the same as what the standardised `getpwnam_r()` and `getgrnam_r()` offer, which probably makes sense here.  Clearly this is always going to be an ifdef soup for people.

I think probably we should do what FreeBSD is doing here because it's so much like the existing standard, so, something like:

```rust
let olderror = *___errno();
*___errno() = 0;
*result = native_getpwent_r(pwd, buf, buflen);
let reterr = if (*result).is_null() {
    *___errno()
} else {
    0
};
*___errno() = olderror;
reterr
```

That way, we can use `errno` ourselves but not stomp on it, and presumably we can just be included in whatever conditional compilation in `nix`, et al, is already in-place for FreeBSD.  This is pretty much what our POSIX-compliant `getgrnam_r()` wrapper does in illumos today; i.e.,

```c
/*
 * POSIX.1c standard version of the function getgrnam_r.
 * User gets it via static getgrnam_r from the header file.
 */
int
__posix_getgrnam_r(const char *name, struct group *grp, char *buffer,
    size_t bufsize, struct group **result)
{
        int nerrno = 0;
        int oerrno = errno;

        if ((*result = getgrnam_r(name, grp, buffer, (uintptr_t)bufsize))
            == NULL) {
                        nerrno = errno;
        }
        errno = oerrno;
        return (nerrno);
}
```

SteveLauC: Done.

SteveLauC:
I just squashed and rebased, could I trouble you guys for another review:)

SteveLauC:
Current APIs are the same as the original ones, so no API break here:
* [doc for `getpwent_r`](https://rust-lang.github.io/libc/x86_64-unknown-illumos/doc/libc/fn.getpwent_r.html)
* [doc for `getgrent_r`](https://rust-lang.github.io/libc/x86_64-unknown-illumos/doc/libc/fn.getgrent_r.html)

ctrlcctrlv:
Thanks a lot @jclulow for helping out. As I said in IRC, I think your solution offers the best of my proposed solution and the solution @SteveLauC originally proposed.

JohnTitor (approved):

JohnTitor:
LGTM, thanks for clarifying!
@jclulow I'm not sure if you approved this or not, could you approve explicitly?

Meanwhile, let's see if CI agrees, @bors try

SteveLauC:
CI [`Build Channels macOS (1.13.0)`](https://github.com/rust-lang/libc/actions/runs/3177914048/jobs/5178927777#logs) and [`Docker Linux Tier2 (sparc64-unknown-linux-gnu)`](https://github.com/rust-lang/libc/actions/runs/3177914048/jobs/5178943511#logs) failed, seems unrelated to this pr

JohnTitor:
CI issue has been fixed, @bors try

JohnTitor:
Since this PR gets no response for a month, I'd like to go ahead assuming @jclulow overall agreed with the change (if you have any objections/concerns, let us know!).
@bors r+

PR #2920, opened 2022-09-23
Exposed malloc_size on MacOSX

naricc:
`malloc_size` is available in the OSX libc, but not exposed by rust libc. This change exposed it.

This is my first PR to libc, or to any rust library. So please be liberal with feedback.


naricc:
Abandoning now that I found the contribution instructions.

PR #2921, opened 2022-09-23
Exposed malloc_size on Mac

naricc:
This change exposes malloc_size, which is available in libc on MacOSX.

### DIFF
--- a/libc-test/semver/apple.txt
+++ b/libc-test/semver/apple.txt
@@ -1951,7 +1951,9 @@ mach_vm_offset_t
 mach_vm_size_t
 madvise
 malloc_default_zone
+malloc_good_size
 malloc_printf
+malloc_size
 malloc_statistics_t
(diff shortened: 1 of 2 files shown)


JohnTitor:
Could you also add it on https://github.com/rust-lang/libc/blob/master/libc-test/semver/apple.txt?

devnexen:
I may suggest to add `malloc_good_size` as well since they re somewhat related, but no worries it can be added in a separate PR otherwise :).

naricc:
@JohnTitor I've added it to `apple.txt`. I've got to admit I am a little confused though; what is `apple.txt` actually used for?

@devnexen That is a good idea; I may as well add it while I am doing this, since they tend to be used together.

JohnTitor (approved):

JohnTitor:
Let me check if it's spurious or not, @bors retry

PR #2925, opened 2022-09-25
CPU_SETSIZE constant to dragonflybsd.

devnexen:


JohnTitor:
Seems it doesn't belong to cpuset.h but cpumask.h on DragonflyBSD: https://github.com/DragonFlyBSD/DragonFlyBSD/blob/c163a4d7ee9c6857ee4e04a3a2cbb50c3de29da1/sys/sys/cpumask.h#L36

Looking at its declaration, it can be different from FreeBSD, I guess?

devnexen:
> Seems it doesn't belong to cpuset.h but cpumask.h on DragonflyBSD: https://github.com/DragonFlyBSD/DragonFlyBSD/blob/c163a4d7ee9c6857ee4e04a3a2cbb50c3de29da1/sys/sys/cpumask.h#L36
> 
> Looking at its declaration, it can be different from FreeBSD, I guess?

I did a quick c test and there is only one arch.

src/unix/bsd/freebsdlike/dragonfly/mod.rs
```diff
@@ -1066,6 +1066,8 @@ pub const CPUCTL_MSRSBIT: ::c_int = 0xc0106305;
 pub const CPUCTL_MSRCBIT: ::c_int = 0xc0106306;
 pub const CPUCTL_CPUID_COUNT: ::c_int = 0xc0106307;
 
+pub const CPU_SETSIZE: ::size_t = ::mem::size_of::<::cpumask_t>();
```
JohnTitor, line 1069: It seems you forgot to multiply 8.

devnexen: true  I ve been too fast here

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #2930, opened 2022-09-28
solarish amd64 regset constants

devnexen:


JohnTitor (approved):

JohnTitor:
LGTM, @jclulow @pfmooney could you review when you have some time?

pfmooney:
I think it looks OK, but I have one question (directed to @devnexen):

Are there existing or potential consumers which are driving these additions?  There have been bits added in the past -- [MR_HDR_AOUT](https://github.com/rust-lang/libc/blob/198beb0d45ef9ae2be1f4a6d35446a2822c13d76/src/unix/solarish/illumos.rs#L59) is the most recent example -- an effectively Uncommitted interfaces which, when [cleaned up](https://github.com/illumos/illumos-gate/commit/fec047081731fd77caf46ec0471c501b2cb33894#diff-33b9ac79706b16f65c95e909fde32771b388f896bee74a006d13fefdf877fe91) caused the libc-test suite to start failing on new machines.

I'm happy to see additions if they will (or are likely to) be used, and are not exposing bits of the OS which are less than [Committed](https://illumos.org/man/7/attributes).

Thanks

devnexen:
This particular set is indeed used within a getcontext part, it would be more readable using these constants.

JohnTitor:
> I'm happy to see additions if they will (or are likely to) be used

> This particular set is indeed used within a getcontext part, it would be more readable using these constants.

I'm going to r+ considering this is approved with the above.

@bors r+

PR #2931, opened 2022-09-29
Use `::Option` and not `Option` for `pthread_jit_write_callback_t`

thomcc:
The `pthread_jit_write_callback_t` type alias (introduced by https://github.com/rust-lang/libc/pull/2896) uses `Option` rather than `::Option`, which causes problems when trying to update the `libc` dependency that `std` has.

(Please cut a release after merging this <3)


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

JohnTitor:
@thomcc 0.2.134 is now available including this: https://crates.io/crates/libc/0.2.134

thomcc:
Thank you!

Issue #2934, opened 2022-09-30
Rust binding to LFS64 ABI-compat symbols rather than public names on musl libc


richfelker:
It's come to our attention as part of [removing the legacy LFS64 glibc-ABI-compat symbols](https://www.openwall.com/lists/musl/2022/09/30/3) that Rust's libc crate is apparently binding to the LFS64 ("large file support") symbol names like `fopen64`, etc. rather than the standard symbol names. In the musl API/ABI (defined by our public headers) the LFS64 interfaces exist only as macros remapping to the standard function names, and the nonstandard LFS64 names are purely a glibc-ABI-compat mechanism that will not be linkable in the future (i.e. only available to already-dynamic-linked binaries referencing these symbols). This is part of an effort to get rid of the macros (which cause a lot of havok on various C++ projects), which is blocked on having the symbols present, because the prevalence of autoconf link-only tests makes it unsafe to have the symbols linkable without declarations in the headers.

What needs to happen here in order for Rust to reference only the standard libc interface names on musl targets?

bossmc:
Is the position that the `xxx64` calls are simply aliases to `xxx` guaranteed to be true going forwards (i.e.  are the `xxx` functions already effectively the `xxx64` versions)?  The `libc` crate can perform the same aliasing when targeting `-musl` fairly easily.  An alternative would be to remove these symbols from the `libc` crate on the `musl` targets, but that feels like a bigger breaking change in the rust-world and would be a problem if the ABI ever changed such that `xxx64` actually did something other than be aliased to `xxx`.

Similar question around `off64_t` and friends, musl currently defines these to their non-64 counterparts, is that also a position we can rely on going forwards?

I'm guessing the answers to the above are all "yes" since I'm assuming that musl's offset/size types have always been 64-bit and the methods have always used the 64-bit underlying syscalls.  Hence there being no need to have separate names for compatibility reasons.

Lastly is there a list of affected symbols or is it simply "all the ones that end with `64`"?

richfelker:
> Is the position that the `xxx64` calls are simply aliases to `xxx` guaranteed to be true going forwards

[...]

PR #2935, opened 2022-09-30
Alias all LFS64 symbols to their non-LFS64 counterparts on musl

bossmc:
As per https://github.com/rust-lang/libc/issues/2934 the LFS64 symbols on musl-libc are simply aliases to the non-LFS64 symbols.  Currently this is done both in the header files (as `#define` entries) and in the library (as aliasing symbols).  There is a desire in musl to drop the ABI compatibility shims (the symbol aliases) - currently the `libc` crate exports the LFS64 symbols by `extern`-ing the compatibility shims which will fail if musl removes them.

This changes the musl build of libc to replicate the aliasing that's in the C header files (with `pub use xxx as xxx64`) so the API from the `libc` crate is unchanged, but the crate is now compatible with the upcoming musl release.

I've also checked and all the LFS64 types (e.g. `off64_t`) are already the same as their non-LFS64 equivalents.

This is an annoying change to test, `libc-test` seems expect to build against `musl 1.1.24` (in fact, does Rust even support `musl 1.2`?  It's not obvious that it does... e.g. see https://github.com/rust-lang/libc/blob/d99c37d97c7a075ffc7f4260f2dd134d1ee3daaa/src/unix/linux_like/linux/musl/mod.rs#L288-L292)


JohnTitor:
This is a breaking change, we have to deprecate them on musl first otherwise user code depending on them would be broken.

bossmc:
Unless I've missed something, this isn't a breaking change (at a crate API level at least) - all the types and functions that used to exist still do and still have the same name/arguments/structure/alignment etc.  The only difference is that when rust code calls `xxx64(*const xxx64_t)` it will actually call the `xxx` function in the musl library (which it was already doing since the `xxx64` symbol in the library was a link-time alias to the `xxx` symbol).

I've done a bit more work on this (ensuring that the complex types are rust-level aliases to each other on `musl` so they can be passed to either function without failing type checks), though I'm not sure it's complete (I've not done this for the structs in musl's `b64` or `b32` modules yet - and I'm unsure what needs to happen in the `b32` case).

src/unix/linux_like/linux/musl/mod.rs
```diff
@@ -756,6 +745,73 @@ extern "C" {
...
+    let _dirent: ::dirent64 = dirent;
+    let _ino: ::ino64_t = ino;
+    let _flock: ::flock64 = flock;
+    let _off: ::off64_t = off;
+    let _pos: ::fpos64_t = pos;
+    let _rlimit: ::rlimit64 = rlimit;
+    let _stat: ::stat64 = stat;
+    let _statfs: ::statfs64 = statfs;
+    let _statvfs: ::statvfs64 = statvfs;
+}
```
JohnTitor, line 774: This could be moved to test, I guess?

src/unix/linux_like/mod.rs
```diff
@@ -1776,8 +1728,65 @@ extern "C" {
     pub fn uname(buf: *mut ::utsname) -> ::c_int;
 }
 
+// LFS64 extensions
+//
+// * musl has 64-bit versions only so aliases the LFS64 symbols to the standard ones
+// * ulibc doesn't have preadv64/pwritev64 (TODO: Or does it?!)
```
JohnTitor, line 1734: We prefer FIXME over TODO, and this is unclear what we should do, I think we could just drop that.

bossmc: Ah, whoops, that TODO was for me 😄 the existing code didn't declare `preadv64/pwritev64` on `ulibc` and I wanted to write a comment explaining why they were omitted here.  I've just had a look at indeed `ulibc` doesn't have these entrypoints (no idea why not) so the TODO can be removed.

src/unix/linux_like/linux/musl/mod.rs
bossmc: Could be, though this is merely a compile-time check that the interchangeable types are aliases (rather than identically defined separate types) it's not a test in the classical sense.

An alternative might be to `#[cfg(test)]` this function?

bossmc:
> No, we don't support musl 1.2.0 or higher as it contains the time64 change. We've tried to update, and the discussion is ongoing: #2088 I'm not sure how it's related to this change though, does it matter?

There's a proposed patch available today to remove the symbols from the musl library on the `1.2.3` branch, to test this change has sufficiently removed the references to the LFS64 symbols, I plan to use that patched library.  Unfortunately, this also entails making a few changes to the `libc` crate (e.g. fixing up that linked struct definition) to get the tests passing which are changes that don't want to go into the crate while it's targeting `1.1`.

Maybe this PR should be deferred until @wesleywiser's 1.2 support lands - since these symbols won't be disappearing in the `1.1.x` releases.

colincross:
I tested this against ToT musl that includes the LFS64 symbol removal (https://git.musl-libc.org/cgit/musl/commit/?id=246f1c811448f37a44b41cd8df8d0ef9736d95f4), and it required the attached addition to the second patch:
[dirent64.patch.txt](https://github.com/rust-lang/libc/files/11114685/dirent64.patch.txt).

bossmc:
Thanks @colincross I'd not noticed the `extra-traits` feature.  Apparently there's a catchup merge/rebase needed too now 💯

Amanieu:
Style check is failing, but otherwise this looks good to merge. It is not a breaking change.

bossmc:
From the discussion in #3248 this probably needs some markups.  In that discussion @joshtriplett would prefer the libc crate to expose LFS64 symbols even on Musl 1.2.4 so cross-libc-targeting crates don't need to use `#[cfg]` to find the 64-bit compatible file operations.  This seems reasonable (if a little unusual, not following the model used elsewhere in the libc crate) but it exposes another few questions about _how_ the libc crate exposes the aliased symbols.

Currently the implementation exposes aliased types with `type <foo>64 = <foo>` and aliased functions with `use <foo> as <foo>64` this has a few impacts:

* The two types are not distinct, so (e.g.) trait `impl`s on them overlap where they wouldn't for e.g. glibc
* The function type signatures are "wrong", e.g. the `<foo>64` function takes `*mut <bar>` not `*mut <bar>64` arguments
* The aliased functions are "the same" function (e.g. casting them to raw pointers will give the same value)

The use of `use` is also fairly messy, `use` pulls the named symbol into the current module in every namespace (value, function, type) unless there's already a symbol of that name defined directly in the current module.  This, combined with the name clash in `stat()/stat64() vs struct stat/struct stat64` means that it's not immediately obvious where symbols are coming from and if they're correct.

An alternative approach would be for the musl backend to define each LFS64-affected type twice and to define shim functions of the form:

```
fn stat64(path: *const c_char, buf: *mut stat64) -> c_int {
  return stat(path, buf.cast());
}
```

Which assumes:

* The inliner will do the right thing in release builds
* The two structures are `.cast()`-able to each other which would want a regression test (or deterministic codegen?)

Thoughts?

Amanieu:
I believe there was some confusion in the discussion in #3248. @joshtriplett was talking about time64, not LFS64.

You're right that changing the `*64` types would be a breaking change due to possibe `impl`s, and the alternative solution would be better.

bossmc:
Ok, I've added the set of shim functions (in `.../musl/lfs64.rs`) and restored the duplicate definitions of the structs.  I think this is now the correct shape for review!

bossmc:
I've tested this against musl 1.1.24 (x86_64 and i686) and against musl 1.2.4 (same architectures) - the former passes cleanly and the latter fails only on the padding changes that are already fixed in #3068.

src/unix/linux_like/linux/musl/lfs64.rs
```diff
@@ -0,0 +1,251 @@
+#[no_mangle]
```
Amanieu, line 1: We shouldn't use `no_mangle` here: we don't actually want to define a `creat64` symbol, we just want what is effectively a `#define` alias for it.

Amanieu: In fact these should be `#[inline]`.

bossmc: Hmmm, correct, I was under the impression that `extern "C"` functions are implicitly `#[no_mangle]` but it seems that's not the case.

`#[inline]` or `#[inline(always)]`?  This might be a valid use case for the rarely used second form...

Amanieu: Please use plain `#[inline]`. The general policy is to only use `#[inline(always)]` for performance reasons if justified by benchmarks.

bossmc: No problem, done!

Mark-Simulacrum:
Closing & reopening to hopefully fix a bors bug which is preventing other repos from moving forward...

mqudsi:
Does this need to be done for other `*64` functions? I'm seeing similar build errors regarding `fcntl64` trying to statically compile sqlite for x86_64 musl under Linux.

PR #2942, opened 2022-10-06
introduce X86_FEATURES for linux glibc/musl

### LINKED ISSUE #2929: Missing HWCAP definitions for x86_64
Target Triple: stable-x86_64-unknown-linux-gnu
API: https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/cpufeature.h

The `aarch64` target enjoys is HWCAP fields (bitfields in the returned value from `getauxval(AT_HWCAP)`, but these constants are absent in glibc and musl x86_64. Can we add them?

devnexen:
Closes #2929


JohnTitor:
It's unfortunate that such a big change is introduced without a test, I'm not sure if it really fits the libc crate. Could you clarify why tests are missing and those should be here?

devnexen:
I m not sure it will be accepted since it s not really libc but more kernel space, it took me a while to add those :-) but I understand if you close.

JohnTitor:
@rust-lang/libc Any thoughts on this? I'm slightly in favor of closing this but if anyone has an objection, drop a comment here.

thomcc:
@devnexen is there a reason we can't cover these with tests? What header are they in?

devnexen:
They are in header only found in kernel source

joshtriplett:
I'd like to see these added. But is there some way we could avoid duplicating them between glibc and musl, and for that matter between x86 and x86-64? They're identical in all four cases.

In general, things from the Linux kernel headers should be identical across all C libraries.

joshtriplett:
:+1:, this new version looks much better to me, and seems reasonable to merge.

@JohnTitor Any objections, or would you be fine with merging this?

JohnTitor:
I'm still concerned these don't have a test.
And as far as I've noticed, these're also _unstable_, i.e. could be changed easily. Example commits are https://github.com/torvalds/linux/commit/d45476d9832409371537013ebdd8dc1a7781f97a, https://github.com/torvalds/linux/commit/356edeca2e811642b5848605c2f5f9a7c6c32112, https://github.com/torvalds/linux/commit/f098addbdb44c8a565367f5162f3ab170ed9404a. Since we don't have any policy for the supported kernel version, we cannot change it soonish when something is changed/removed (currently we haven't changed/removed items directly but have deprecated them). So I'm not sure if these really fit with this crate.

joshtriplett:
I don't think we can easily test these unless the kernel headers are available on the test system.

And I do think these are important for users making use of auxv.

JohnTitor:
Then these could be provided via another crate like mach on macOS.

joshtriplett:
My apologies, I only just now realized that this is adding *all* the CPU feature constants, not just those visible in auxv. This should only add those visible in binary form to userspace; the visible ones are stable.

joshtriplett:
Also, arguably, the visible ones should be moved to a uapi header in the kernel.

JohnTitor:
If you only add items that you want to use and these are stable and small enough to review, I'm fine to accept ignoring tests.

JohnTitor:
@devnexen Could you address https://github.com/rust-lang/libc/pull/2942#issuecomment-1280270087? I think we could merge once it's done.

JohnTitor (approved):

JohnTitor:
@joshtriplett Could you review the new diff?

src/unix/linux_like/linux/arch/generic/mod.rs
```diff
@@ -111,6 +111,72 @@ cfg_if! {
...
+        pub const X86_FEATURE_AVX512F: ::c_ulong = 288 + 16;
+        pub const X86_FEATURE_AVX512DQ: ::c_ulong = 288 + 17;
+        pub const X86_FEATURE_AVX512IFMA: ::c_ulong = 288 + 21;
+        pub const X86_FEATURE_AVX512PF: ::c_ulong = 288 + 26;
+        pub const X86_FEATURE_AVX512ER: ::c_ulong = 288 + 27;
+        pub const X86_FEATURE_AVX512CD: ::c_ulong = 288 + 28;
+        pub const X86_FEATURE_AVX512BW: ::c_ulong = 288 + 30;
+        pub const X86_FEATURE_AVX512VL: ::c_ulong = 288 + 31;
+        pub const X86_FEATURE_AVX512_BF16: ::c_ulong = 384 + 5;
+        pub const X86_FEATURE_AVX512VBMIK: ::c_ulong = 512 + 1;
```
workingjubilee, line 153: I can't find this feature name on torvalds/linux master, is this a typo?

devnexen: good point.

workingjubilee:
I don't know if we should be exposing these if they only are mentioned in the kernel, even if they are in the uapi headers (and they currently aren't, afaict) because as mentioned in https://github.com/rust-lang/libc/pull/2713, pedantically but correctly, the libc is allowed to have a varying definition here, and I can't find them exposed by musl or glibc either.

This is traditionally something where people have just bypassed the libc and kernel by using the CPUID interface, because on x86, that doesn't require privileges.

Issue #2870, opened 2022-08-10
`libiconv` is still unconditionally linked to all binaries on macOS


winterqt:
```
$ rustc -Vv
rustc 1.64.0-nightly (93ffde6f0 2022-07-23)
binary: rustc
commit-hash: 93ffde6f04d3d24327a4e17a2a2bf4f63c246235
commit-date: 2022-07-23
host: aarch64-apple-darwin
release: 1.64.0-nightly
LLVM version: 14.0.6
$ cat main.rs
fn main() {}
$ rustc main.rs
$ otool -L main
main:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.100.5)
	/usr/lib/libresolv.9.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
```

The `libc` version in this compiler's `std` is 0.2.126, which includes #2150, which *should* have fixed this.

JohnTitor:
The problem is that I don't have a macOS machine to debug, cc @goffrie (again).

bjorn3:
I don't think rustc cares about if functions in an extern block are used, but unconditionally considers the library referenced by the extern block used.

goffrie:
Interesting, I'm pretty sure you're correct. I thought I tested #2150 manually but I might have fooled myself.

In that case I'm not sure that we can actually do anything. libiconv is part of the macOS base system at least so it shouldn't pose a problem at runtime.

thomcc:
It poses a problem in several cases any build system that tries to accurately track dependencies or linkage has to deal with this -- you see this with the Nix discussion that seems to have referenced the issue. It causes various conflicts with C code occasionally, slows down linking/loading/startup/dlsym/etc, and so on. It would also cause massive issues for all Rust programs if `libiconv` ever bumped to version 8.0.0 (it's on 7.0.0), or if it bumped the soname to `libiconv.3`. I also think the fact that it isn't re-exported from `libSystem.B.{tbd,dylib}` is likely deliberate (probably due to the size -- all those conversions are not free).

[...]

PR #2944, opened 2022-10-07
Don't link against iconv on apple targets when used by `std`

thomcc:
See https://github.com/rust-lang/libc/issues/2870 (and specifically my, erm, rant in https://github.com/rust-lang/libc/issues/2870#issuecomment-1271104762). I think we should probably remove it outright, but we definitely shouldn't have it in every Rust program by giving it to `std`.

FIxes https://github.com/rust-lang/libc/issues/2870

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -5873,7 +5873,10 @@ cfg_if! {
     }
 }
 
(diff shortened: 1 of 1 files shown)


JohnTitor:
Makes sensible to me, thanks!
@bors r+

Issue #2948, opened 2022-10-08
Missing XFS_SUPER_MAGIC for android target


MrCroxx:
Target Triple: *-linux-android
API: https://android.googlesource.com/platform/bionic/+/master/libc/include/sys/vfs.h#105

Bionic libc for Android has exported `XFS_SUPER_MAGIC`, but rust libc hasn't.

I'm not that familiar with Android, but when I was fixing the same problem for rust lib `nix`, I found that rust libc missed the constant for android target and I checked the bionic libc source code. I will send a PR if there is no problem.

PR #2949, opened 2022-10-08
add xfs super magic to android targets

MrCroxx:
`XFS_SUPER_MAGIC` is defined in libc for android, but is missing in rust libc.

https://android.googlesource.com/platform/bionic/+/master/libc/include/sys/vfs.h#105

fix: #2948


JohnTitor (approved):

JohnTitor:
You don't have to open an issue when you immediately submit a PR.
@bors r+

PR #2955, opened 2022-10-10
[DNM] testing

JohnTitor:
r? @ghost

Signed-off-by: Yuki Okushi <jtitor@2k36.org>

### DIFF
--- a/.github/workflows/bors.yml
+++ b/.github/workflows/bors.yml
@@ -8,166 +8,19 @@ on:
 
 permissions: {}
 jobs:
-  docker_linux_tier1:
-    permissions:
-      actions: write # to cancel workflows (rust-lang/simpleinfra/github-actions/cancel-outdated-builds)
-      contents: read # to fetch code (actions/checkout)
-
-    name: Docker Linux Tier1
-    runs-on: ubuntu-22.04
-    strategy:
-      fail-fast: true
-      matrix:
-        target: [
-          i686-unknown-linux-gnu,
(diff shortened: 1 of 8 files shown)


chriswailes:
1. The use of make_standalone_toolchain.py is no longer recommended and using it with the compiler wrappers is explicitly not supported. See https://developer.android.com/ndk/guides/other_build_systems
2. The Android Platform Support document says that rustc should target the lowest level API supported by a target for a given NDK unless there is a specific reason a higher API level is needed. If libc needs to use API level 28 we should update the rustc CI and document it.
3. Updates to aarch-linux-android/Dockerfile, arm-linux-androideabi/Dockerfile, and x86_64-linux-android/Dockerfile are missing.

JohnTitor:
>     2. The Android Platform Support document says that `rustc` should target the lowest level API supported by a target for a given NDK unless there is a specific reason a higher API level is needed. If `libc` needs to use API level 28 we should update the `rustc` CI and document it.

It doesn't need to match them, the libc crate CI is independent.

>     3. Updates to aarch-linux-android/Dockerfile, arm-linux-androideabi/Dockerfile, and x86_64-linux-android/Dockerfile are missing.

Have you read the PR title? This is currently the testing status, it's intentional.

chriswailes:
Would it be acceptable to re-open https://github.com/rust-lang/libc/pull/2956?

JohnTitor:
I don't think we should change anything here until https://github.com/rust-lang/rust/issues/103673 is resolved.

PR #2956, opened 2022-10-10
Update the CI to use Android NDK r25b

chriswailes:


JohnTitor:
Thanks for the PR! I've, however, updated the toolchain on https://github.com/rust-lang/libc/pull/2955, closing in favor of that.

chriswailes:
A couple of comments on #2955:

1. The use of `make_standalone_toolchain.py` is no longer recommended and using it with the compiler wrappers is explicitly not supported.  See https://developer.android.com/ndk/guides/other_build_systems
2. The Android Platform Support document says that `rustc` should target the lowest level API supported by a target for a given NDK unless there is a specific reason a higher API level is needed.  If `libc` needs to use API level 28 we should update the `rustc` CI and document it.
3. Updates to `aarch-linux-android/Dockerfile`, `arm-linux-androideabi/Dockerfile`, and `x86_64-linux-android/Dockerfile` are missing.

PR #2957, opened 2022-10-11
add some time functions on glibc and musl

SteveLauC:
#### man pages

* [asctime/ctime man page](https://man7.org/linux/man-pages/man3/ctime.3.html)
* [strftime](https://man7.org/linux/man-pages/man3/strftime.3.html)
* [strptime](https://man7.org/linux/man-pages/man3/strptime.3.html)

I didn't add `ctime()/ctime_r()` on musl because they involve the `time_t` type, which elicits [a deprecation wraning](https://github.com/rust-lang/libc/pull/1956).

Is it fine to add these two functions on `musl`, they will have the same definitions  as the `glibc` ones:
```rust
pub fn ctime(timep: *const time_t) -> *mut ::c_char;
pub fn ctime_r(timep: *const time_t, buf: *mut ::c_char) -> *mut ::c_char;
```
If it's ok, I will add them:)

### DIFF
--- a/libc-test/semver/linux-gnu.txt
+++ b/libc-test/semver/linux-gnu.txt
@@ -654,3 +654,7 @@ utmpx
 utmpxname
 euidaccess
 eaccess
+asctime_r
+ctime_r
+strftime
+strptime
--- a/libc-test/semver/linux-musl.txt
+++ b/libc-test/semver/linux-musl.txt
@@ -50,3 +50,6 @@ reallocarray
 timex
 euidaccess
 eaccess
+asctime_r
+strftime
+strptime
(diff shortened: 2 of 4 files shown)


joshtriplett:
While in general we provide bindings to most functions even if they're not advisable to call, is there a specific reason you need `asctime` (and `ctime`)? It returns a pointer to an internal libc buffer that gets overwritten, so it's very easy to misuse, and needs copying out for safety.

No objections if you really do need `asctime`, just want to confirm.

SteveLauC:
Yes, their static return value can be overwritten, which is not safe.

I don't have any specific reason to add them, since the reentrant functions provide the same functionality. I added them for the reason that not all platforms have such reentrant successors, so exposing them may keep some kind of consistency between those systems.

But they should be considered obsolete, will remove them:)

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #2556, opened 2021-11-21
Add missing net/if.h and net/if_mib.h items

GuillaumeGomez:

PR #2960, opened 2022-10-13
implement ifreq for linux

folkertdev:
defines the `ifreq` struct, for use with `ioctl`. I used the definition from the [manpages](https://man7.org/linux/man-pages/man7/netdevice.7.html)

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -603,6 +603,21 @@ s! {
         pub src_length: ::__u64,
         pub dest_offset: ::__u64,
     }
+
+    pub struct __c_anonymous_ifru_map {
+        pub mem_start: ::c_ulong,
+        pub mem_end: ::c_ulong,
+        pub base_addr: ::c_ushort,
+        pub irq: ::c_uchar,
+        pub dma: ::c_uchar,
+        pub port: ::c_uchar,
+    }
+
+   pub struct in6_ifreq {
+       pub ifr6_addr: ::in6_addr,
+       pub ifr6_prefixlen: u32,
+       pub ifr6_ifindex: ::c_int,
+   }
 }
 
 s_no_extra_traits! {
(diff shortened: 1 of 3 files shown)


folkertdev:
I'm not sure what the CI error means. It seems to complain about type I did not touch?!

src/unix/linux_like/linux/mod.rs
```diff
@@ -603,6 +603,21 @@ s! {
...
         pub dest_offset: ::__u64,
     }
+
+    pub struct ifru_map {
+        mem_start: ::c_ulong,
+        mem_end: ::c_ulong,
+        base_addr: ::c_ushort,
+        irq: ::c_uchar,
+        dma: ::c_uchar,
+        port: ::c_uchar,
```
JohnTitor, line 613: I'd recommend marking fields as public for convenience.

src/unix/linux_like/linux/mod.rs
```diff
@@ -690,6 +705,27 @@ s_no_extra_traits! {
         #[cfg(not(all(target_arch = "x86_64", target_pointer_width = "32")))]
         pad: [::c_long; 4],
     }
+
+    pub union ifr_ifru {
```
JohnTitor, line 709: Should be under on libc_union.

JohnTitor:
Okay I checked the source and realized that it's anonymous: https://github.com/torvalds/linux/blob/5bfc75d92efd494db37f5c4c173d3639d4772966/include/uapi/linux/if.h#L234-L275

Luckily, we have a prior art on FreeBSD: https://github.com/rust-lang/libc/pull/2556
I think following a similar way would make CI green, but I'd suggest using them on your local once, to confirm the behavior.

folkertdev:
I can't seem to run the tests locally, at least not in the way that the README describes

    > cargo test
       Compiling libc-test v0.2.135 (/home/folkertdev/rust/libc/libc-test)
    The following warnings were emitted during compilation:

    warning: /home/folkertdev/rust/libc/target/debug/build/libc-test-dbaae0e5290e28ae/out/main.c:129:10: fatal error: linux/openat2.h: No such file or directory
    warning:   129 | #include <linux/openat2.h>
    warning:       |          ^~~~~~~~~~~~~~~~~
    warning: compilation terminated.

should this work? I ran these installs based on the dockerfile: `gcc libc6-dev ca-certificates linux-headers-generic`.

In any case, the implementation now more closely mimics the freebsd one. I added a `Debug` impl, but not `Eq` or `Hash` because the linux version of the `ifru` union contains a pointer, so equality/hashing would not have the behavior we typically expect in rust.

JohnTitor:
I guess your glibc/OS version is old. For example, the libc-test currently requires Ubuntu 22.04 and its glibc version.

folkertdev:
ok, well not much I can do about that in the short run. I may not have fully understood your suggestion: is there something in the freebsd code that I did not copy (correctly) that would fix the current CI issue?

JohnTitor:
> is there something in the freebsd code that I did not copy (correctly) that would fix the current CI issue?

Looking at the source, ifru_map is also anonymous, it seems. And about in6_ifreq, it's declared on include/uapi/linux/ipv6.h but the test doesn't include it, you have to add it to `test_linux` func of https://github.com/rust-lang/libc/blob/master/libc-test/build.rs (I'm not sure if it requires a special handling like in6.h though). I don't think these get rid of all the errors, but let's do them and check what errors remain.

folkertdev:
allright, got CI to pass. Thanks @JohnTitor for your help!

JohnTitor:
LGTM, could you make sure these are usable for your use case? (If you already did it, great!) Also, it'd be great if you could squash commits into one as the current commits would be noisy when git-blaming.

folkertdev:
commits are squashed, and I verified that this works for our use case

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #2961, opened 2022-10-14
fix wrong xattr syscalls on NetBSD

SteveLauC:
I am not 100% sure that these bindings are wrong because I can not find them in the [NetBSD manual](https://man.netbsd.org/lsetxattr).

But according to [sys/sys/xattr.h](https://github.com/NetBSD/src/blob/ac36f3f9fee8aff1c69d653503149faaa531f752/sys/sys/xattr.h#L62-L76), they seems to be wrong:

```c
int	setxattr(const char *, const char *, const void *, size_t, int);
int	lsetxattr(const char *, const char *, const void *, size_t, int);

int	fremovexattr(int, const char *);
```

```rust
// Current bindings in `libc`:

pub fn setxattr(
    path: *const ::c_char,
    name: *const ::c_char,
    value: *const ::c_void,
    size: ::size_t,
) -> ::c_int;
pub fn lsetxattr(
    path: *const ::c_char,
    name: *const ::c_char,
    value: *const ::c_void,
    size: ::size_t,
) -> ::c_int;
pub fn fremovexattr(
    fd: ::c_int, 
    path: *const ::c_char, 
    name: *const ::c_char
) -> ::c_int;
```

Friendly ping @devnexen who originally added them in #2514 .

### DIFF
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -2977,12 +2977,14 @@ extern "C" {
         name: *const ::c_char,
         value: *const ::c_void,
         size: ::size_t,
+        flags: ::c_int,
     ) -> ::c_int;
     pub fn lsetxattr(
         path: *const ::c_char,
         name: *const ::c_char,
         value: *const ::c_void,
         size: ::size_t,
+        flags: ::c_int,
     ) -> ::c_int;
     pub fn fsetxattr(
         filedes: ::c_int,
@@ -2996,7 +2998,7 @@ extern "C" {
     pub fn flistxattr(filedes: ::c_int, list: *mut ::c_char, size: ::size_t) -> ::ssize_t;
     pub fn removexattr(path: *const ::c_char, name: *const ::c_char) -> ::c_int;
     pub fn lremovexattr(path: *const ::c_char, name: *const ::c_char) -> ::c_int;
-    pub fn fremovexattr(fd: ::c_int, path: *const ::c_char, name: *const ::c_char) -> ::c_int;
+    pub fn fremovexattr(fd: ::c_int, name: *const ::c_char) -> ::c_int;
 
     pub fn string_to_flags(
         string_p: *mut *mut ::c_char,


JohnTitor:
I think this is a breaking change as it changes the interface. I'm, however, fine to accept it as long as they aren't usable at all across all NetBSD versions (that makes it possible to assume there's no real user). Could you make sure of that?

SteveLauC:
Yes, this is a breaking change. I will try to fire up a NetBSD VM to test the current APIs.

SteveLauC:
Hi, guys, I just have NetBSD installed (including Rust toolchain and the GCC compiler), so I am ready to test the existing bindings. 
Screenshot from 2022-11-04 17-39-43

But I may need some help on how to test them. 

What values should be passed for those function parameters that do not exist? (the `path` argument in `removexattr`) For parameters that are not exposed, should we just ignore them (`flag` argument of `setxattr` and `lsetxattr`)? 

Friendly ping @JohnTitor

SteveLauC:
Though EA is not enabled, I still tested the following code snippet:

```rust
use libc::{__errno, c_char, c_void, lsetxattr, perror, fremovexattr, setxattr};
use std::{        
    ptr::null_mut,
    os::unix::io::AsRawFd,     
    fs::File,                                  
};       
            
fn test_setxattr() {        
    println!("Testing setxattr:");
    let res = unsafe {
        setxattr(
            "/home/steve/workspace/rust/Cargo.toml\0".as_ptr() as *mut c_char,
            "user.first\0".as_ptr() as *mut c_char,
            "first\0".as_ptr() as *mut c_void,
            5,                                                              
        )                     
    };               
                                                                                          
    if res == -1 {
        unsafe {
            perror(null_mut());
            println!("errno = {}", *__errno());
        }                      
    } else {                                   
        println!("success");
    }       
    println!();             
}    
               
 
fn test_lsetxattr() {
    println!("Testing lsetxattr:");
    let res = unsafe {
        lsetxattr(   
            "/home/steve/workspace/rust/Cargo.toml\0".as_ptr() as *mut c_char,
            "user.first\0".as_ptr() as *mut c_char,
            "user.first\0".as_ptr() as *mut c_char,
            "first\0".as_ptr() as *mut c_void,
            5,
        )
    };

    if res == -1 {
        unsafe {
            perror(null_mut());
            println!("errno = {}", *__errno());
        }
    } else {
        println!("success");
    }
    println!();
}

fn test_fremovexattr() {
    println!("Testing fremovexattr:");
    let file = File::open("/home/steve/workspace/rust/Cargo.toml").unwrap();
    let fd = file.as_raw_fd();
    let res = unsafe{
        fremovexattr(fd, "user.fist\0".as_ptr() as *mut c_char, null_mut() as *mut c_char)
    };
    if res == -1 {
        unsafe {
            perror(null_mut());
            println!("errno = {}", *__errno());
        }
    } else {
        println!("success");
    }
    println!();
}

fn main() {
    test_setxattr();
    test_lsetxattr();
    test_fremovexattr();
}

```shell
$ cargo r -q
Testing setxattr:
Not supported
errno = 86

Testing lsetxattr:
Not supported
errno = 86

Testing fremovexattr:
Not supported
errno = 86
```

The existing bindings seemed to work, at least they showed me that EA is not supported.

SteveLauC:
> [NetBSD 10 will be the first official release with full extended attribute support in FFS](https://mail-index.netbsd.org/tech-kern/2022/05/24/msg028105.html)

If we would like to test these bindings further, we have to wait for the NetBSD 10 release

SteveLauC:
> Has the situation changed here? I'm having trouble finding headers for these.

Hi, looks like the header file is still there: https://github.com/NetBSD/src/blob/a623cf0b61440d709d496b4325d530e07f25fb94/sys/sys/xattr.h#L61-L77

tgross35 (approved):
...I read that file about five times and missed the extra `int` each time. Thanks for the link, changes LGTM but please rebase.

SteveLauC:
Emm, thanks for rebasing for me 🫡

tgross35:
The button was there so figured why not :)

PR #2963, opened 2022-10-14
linux fixes recvmmsg flags type for musl/emscripten.

### LINKED ISSUE #2945: Mismatch between flag type and function argument for `recvmmsg`
The function `recvmmsg` for `musl` and `emscripten` has argument type `c_uint` for flags, but the flag types as defined in this package are `c_int`.

Problematic function definitions here:
https://github.com/rust-lang/libc/blob/7ecea9ac6d4e2093a6619d2fc0fd3b843ee21aa4/src/unix/linux_like/linux/musl/mod.rs#L706
https://github.com/rust-lang/libc/blob/6eee580b51455131b10b780087025379a591ceb6/src/unix/linux_like/emscripten/mod.rs#L1873

I encounted this when compiling a dependency that used `libc`:

```
#18 580.4    Compiling solana-streamer v1.10.40
#18 580.8 error[E0308]: mismatched types
#18 580.8   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/solana-streamer-1.10.40/src/recvmmsg.rs:99:70
#18 580.8    |
#18 580.8 99 |         unsafe { libc::recvmmsg(sock_fd, &mut hdrs[0], count as u32, MSG_WAITFORONE, &mut ts) };
#18 580.8    |                                                                      ^^^^^^^^^^^^^^ expected `u32`, found `i32`
#18 580.8    |
#18 580.8 help: you can convert an `i32` to a `u32` and panic if the converted value doesn't fit
#18 580.8    |
#18 580.8 99 |         unsafe { libc::recvmmsg(sock_fd, &mut hdrs[0], count as u32, MSG_WAITFORONE.try_into().unwrap(), &mut ts) };
#18 580.8    |                                                                                    ++++++++++++++++++++
#18 580.8 
#18 580.9 For more information about this error, try `rustc --explain E0308`.
#18 580.9 error: could not compile `solana-streamer` due to previous error
#18 580.9 warning: build failed, waiting for other jobs to finish...
#18 601.8 error: build failed
```

MSG_WAITFORONE is defined as a `c_int`:
https://github.com/rust-lang/libc/blob/183820f6330d3d7a107632dfc87ef38212046b69/src/unix/linux_like/mod.rs#L777

devnexen:
closes #2945

### DIFF
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -759,23 +759,95 @@ pub const PF_IEEE802154: ::c_int = AF_IEEE802154;
 pub const PF_CAIF: ::c_int = AF_CAIF;
 pub const PF_ALG: ::c_int = AF_ALG;
 
+#[deprecated(
+    since = "0.2.136",
+    note = "recvmmsg call expects an unsigned type on musl/emscripten"
+)]
 pub const MSG_OOB: ::c_int = 1;
+#[deprecated(
+    since = "0.2.136",
+    note = "recvmmsg call expects an unsigned type on musl/emscripten"
+)]
 pub const MSG_PEEK: ::c_int = 2;
+#[deprecated(
+    since = "0.2.136",
+    note = "recvmmsg call expects an unsigned type on musl/emscripten"
+)]
 pub const MSG_DONTROUTE: ::c_int = 4;
+#[deprecated(
+    since = "0.2.136",
+    note = "recvmmsg call expects an unsigned type on musl/emscripten"
+)]
 pub const MSG_CTRUNC: ::c_int = 8;
+#[deprecated(
+    since = "0.2.136",
+    note = "recvmmsg call expects an unsigned type on musl/emscripten"
+)]
 pub const MSG_TRUNC: ::c_int = 0x20;
+#[deprecated(
+    since = "0.2.136",
+    note = "recvmmsg call expects an unsigned type on musl/emscripten"
+)]
 pub const MSG_DONTWAIT: ::c_int = 0x40;
+#[deprecated(
+    since = "0.2.136",
+    note = "recvmmsg call expects an unsigned type on musl/emscripten"
+)]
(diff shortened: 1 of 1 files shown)


JohnTitor:
Changing their types would break existing code, I'd prefer to follow https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md#breaking-change-policy first.

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

asomers:
This change is causing deprecation warnings downstream, and there's nothing that downstream crates can do about it except to suppress the warning.  I don't think the policy laid out in https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md#breaking-change-policy is appropriate in this case.  That policy would make sense if upstream really were changing an API, and especially if upstream were removing it.  But in this case musl/emscripten didn't change anything.  Instead, libc was simply wrong.  I think libc should just go ahead and fix the definition for musl/emscripten.
A deprecation notice, IMHO, should only be used if there's some way for a downstream crate to be compatible with both the deprecated libc definitions and the fixed ones.

JohnTitor:
> A deprecation notice, IMHO, should only be used if there's some way for a downstream crate to be compatible with both the deprecated libc definitions and the fixed ones.

Users can prepare the upcoming type changes if they depend on it.

niluxv:
This change is breaking CI for `rustix` (and probably other crates), which has `deny(warnings)` in the CI. And as @asomers already said, there is nothing (except `allow(deprecated)`) the user can do about this. Deprecating items is a mechanism for soft removal, not for announcing modifications to existing APIs.

thomcc:
Yes, as I mentioned [here](https://github.com/rust-lang/libc/pull/2978#issuecomment-1289937463), these deprecations don't feel like the right way to handle API changes -- it makes sense to handle removals that way, but it's unclear what downstream users are supposed to do for changes other than `#[allow(...)]` them.

PR #2964, opened 2022-10-16
adds putenv call to win32.

### LINKED ISSUE #2926: putenv and friends are missing on Windows
[_putenv, _wputenv](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv-wputenv?view=msvc-170), [_putenv_s, _wputenv_s](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv-s-wputenv-s?view=msvc-170) are currently missing.

They're important because, on Windows, `std::env::set_var` is implemented via `SetEnvironmentVariableW` WinAPI, which has a separate table from CRT (libc) implementation. As a result, if you're linking Rust with C code and want C code to see some dynamically set environment variables, `putenv` and friends are your only option.

devnexen:
closes #2926.

### DIFF
--- a/libc-test/semver/windows.txt
+++ b/libc-test/semver/windows.txt
@@ -261,6 +261,8 @@ popen
 printf
 ptrdiff_t
 putchar
+putenv
+putenv_s
 puts
 raise
 rand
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #2969, opened 2022-10-18
illumos: Remove obsolete (and private) a.out define

pfmooney:
This cleans up an obsolete definition which [has been removed](https://github.com/illumos/illumos-gate/commit/fec047081731fd77caf46ec0471c501b2cb33894#diff-33b9ac79706b16f65c95e909fde32771b388f896bee74a006d13fefdf877fe91L163) in upstream illumos, and thus is causing libc-test to fail on up-to-date machines.

### DIFF
--- a/src/unix/solarish/illumos.rs
+++ b/src/unix/solarish/illumos.rs
@@ -56,8 +56,6 @@ pub const SOL_FILTER: ::c_int = 0xfffc;
 
 pub const MADV_PURGE: ::c_int = 9;
 
-pub const MR_HDR_AOUT: ::c_uint = 0x3;
-
 pub const B1000000: ::speed_t = 24;
 pub const B1152000: ::speed_t = 25;
 pub const B1500000: ::speed_t = 26;


JohnTitor:
Could you deprecate it first instead of deleting directly? See https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md#breaking-change-policy for the reason.

pfmooney:
> Could you deprecate it first instead of deleting directly? See https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md#breaking-change-policy for the reason.

I can, if you insist.

This was part of a Private (not documented and subject to change/removal at any time) OS interface which probably should not have been added to rust-libc in the first place.  As noted in [the ticket for removing `a.out` exec functionality](https://www.illumos.org/issues/14780),  support was limited to SPARC (which itself is deprecated on illumos).  AFAIK, no work has been done to support Rust on illumos on SPARC.

All that said, I'll defer to what you think best.

Thanks.

JohnTitor (approved):

JohnTitor:
Thanks for clarifying! So, hmm, I guess we could assume there's no real user (or misunderstanding use) then. Given that, it seems okay to accept as-is. @bors r+

PR #2970, opened 2022-10-18
solarish: Fix libsendfile symbol dependency

pfmooney:
A number of function symbols were erroneously grouped under the `libsendfile` link section.  This should correct the situation.


JohnTitor (approved):

JohnTitor:
Good catch, thanks! @bors r+

PR #2978, opened 2022-10-24
follow-up on #2963, changing MSG* constant types for musl/emscripten.

devnexen:


JohnTitor:
Could you just remove the deprecation attrs? We have to figure out how we deliver breaking changes first.

thomcc:
Yeah, deprecations here cause [failures](https://github.com/rust-lang/rust/actions/runs/3317024558/jobs/5479447742) when used in the stdlib...

thomcc:
I started a discussion about this in https://rust-lang.zulipchat.com/#narrow/stream/351149-t-libs.2Fcrates/topic/Deprecations.2Ftype.20changes.20in.20.60libc.60.20.28for.20musl.2C.20emscripten.29.

It's tricky because these aren't really being removed, just changed -- We also still need them both before and after in the stdlib (and I'm sure this is true for others too). So, I'm not sure the policy in the CONTRIBUTING.md guide is actually good for cases like this (it seems really geared for removals...)

JohnTitor (approved):
Let's remove deprecations on 0.2 first. We can discuss how we treat breaking changes later.

PR #2979, opened 2022-10-25
Add MADV constants for Redox

jackpot51:
Required for `memmap2` crate


src/unix/redox/mod.rs
```diff
@@ -634,6 +634,12 @@ pub const PROT_READ: ::c_int = 0x0004;
 pub const PROT_WRITE: ::c_int = 0x0002;
 pub const PROT_EXEC: ::c_int = 0x0001;
 
+pub const MADV_NORMAL: ::c_int = 0;
+pub const MADV_RANDOM: ::c_int = 1;
+pub const MADV_SEQUENTIAL: ::c_int = 2;
```
JohnTitor, line 639: These are swapped, I guess?

jackpot51: They are not swapped, afaict.

JohnTitor: Oh, I looked at https://github.com/redox-os/newlib/blob/0eeb4c1d32aab0dbfd2a8f20aa326ae2d98ca1c6/winsup/cygwin/include/sys/mman.h#L52-L53 and thought so, maybe I checked the wrong place. Could you tell me the right reference for future review?

jackpot51: Newlib is dead code that has not been used for years, and you are looking at the Windows support code in newlib.

Relibc is the C library used by Redox OS. It is written in Rust, of course, and you can see the MADV constants here: https://gitlab.redox-os.org/redox-os/relibc/-/blob/master/src/header/sys_mman/mod.rs#L23

JohnTitor (approved):

PR #2985, opened 2022-10-28
Enforce order of any `s_*!` macro calls

flba-eb:
Before this change, only the order of `s!` was checked. After, it also checks `s_no_extra_traits!` and `s_paren!`. Only the order is checked, not the number of calls. This is required because multiple calls have to be allowed.


JohnTitor:
Thanks!
@bors r+

PR #2986, opened 2022-10-31
Add FICLONE ioctl for linux aarch64

dtolnay:
The ioctl number is the same on aarch64 as on x86_64:

- https://github.com/sunfishcode/linux-raw-sys/blob/v0.1.2/src/aarch64/ioctl.rs#L215-L216
- https://github.com/sunfishcode/linux-raw-sys/blob/v0.1.2/src/x86_64/ioctl.rs#L215-L216


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #2987, opened 2022-10-31
Add Trusty platform support

randomPoison:
This PR adds support for the [Trusty secure operating system](https://source.android.com/docs/security/features/trusty). This upstreams [the patch that we have been using](https://cs.android.com/android/platform/superproject/+/master:external/rust/crates/libc/patches/trusty.patch;l=1;drc=122e586e93a534160230dc10ae3474cf31dd8f7f) internally, cleaned up to meet the project's style conventions.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -139,6 +139,12 @@ cfg_if! {
 
         mod hermit;
         pub use hermit::*;
+    } else if #[cfg(target_os = "trusty")] {
+        mod fixed_width_ints;
+        pub use fixed_width_ints::*;
+
+        mod trusty;
+        pub use trusty::*;
     } else if #[cfg(all(target_env = "sgx", target_vendor = "fortanix"))] {
         mod fixed_width_ints;
         pub use fixed_width_ints::*;
--- a/src/trusty.rs
+++ b/src/trusty.rs
@@ -1,0 +1,66 @@
+pub use core::ffi::c_void;
+
+pub type size_t = usize;
+pub type ssize_t = isize;
+
+cfg_if! {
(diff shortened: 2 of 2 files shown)


JohnTitor:
For maintenance reasons, I'd like to add target support when it's official (the official targets have target maintainers and we could ensure whether a target is still active or not). At a glance, there's no Trusty-related target on rust-lang/rust repo, could you propose adding it there first?

randomPoison:
I've put up a PR to add the trusty compilation targets: https://github.com/rust-lang/rust/pull/103895

Is that enough to move this PR forward? Or did you want to wait until the new targets are approved in the main Rust project before merging here? Because that PR depends on the new APIs added in this PR, which compilates things 😅

src/lib.rs
```diff
@@ -139,6 +139,12 @@ cfg_if! {
 
         mod hermit;
         pub use hermit::*;
+    } else if #[cfg(target_os = "trusty")] {
+        mod fixed_width_ints;
+        pub use fixed_width_ints::*;
+
+        mod trusty;
+        pub use trusty::*;
```
JohnTitor, line 147: I wonder if we could move the module under linux_like or somewhere else similar to that. I'm not familiar with Trusty but is it completely different from Linux-like OS (i.e. exposing linux-like items on Trusty is harmful)?

JohnTitor:
I see https://github.com/rust-lang/compiler-team/issues/582 is approved and it's a good time to go ahead.
Could you check my above comment and rebase to re-run CI?

JohnTitor:
Closing as inactive, feel free to resolve the above comment and re-submit a PR if you're still interested in this change. Thanks for your contribution anyway!

PR #2995, opened 2022-11-06
Add rand48 functions

carbotaniuman:


src/unix/linux_like/linux/mod.rs
```diff
@@ -4343,6 +4353,7 @@ extern "C" {
 
     pub fn pthread_getname_np(thread: ::pthread_t, name: *mut ::c_char, len: ::size_t) -> ::c_int;
     pub fn pthread_setname_np(thread: ::pthread_t, name: *const ::c_char) -> ::c_int;
+
```
JohnTitor, line 4356: This change isn't necessary, could you drop?

PR #3001, opened 2022-11-16
Add sys/ucontext.h signatures for linux aarch64 glibc

dtolnay:
### `getcontext`, `setcontext`, `makecontext`, `swapcontext`

From \<sys/ucontext.h\>. The specification for these was removed from POSIX.1-2008 in favor of POSIX threads, but glibc continues to ship an implementation for aarch64 just as it does for x86_64.

Libc crate's existing x86_64 binding with the same signatures added in this PR:

https://github.com/rust-lang/libc/blob/bbf929d2c8355fa19384b3551c5874c866be465f/src/unix/linux_like/linux/gnu/b64/x86_64/mod.rs#L810-L813

Glibc implementation:

- https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/aarch64/getcontext.S
- https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/aarch64/setcontext.S
- https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/aarch64/makecontext.c
- https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/aarch64/swapcontext.S

<br>

### ~~`iopl`, `ioperm`~~

~~From \<sys/io.h\>. These are functions for accessing x86 I/O ports. ARM has no such I/O ports in the architecture. Linux's `man 2` for both functions contains: _"This call is mostly for the i386 architecture.  On many other architectures it does not exist or will always return an error."_ Glibc ships one of these "always return an error" implementation of both functions for aarch64.~~

~~Matching signatures from x86_64:~~

https://github.com/rust-lang/libc/blob/bbf929d2c8355fa19384b3551c5874c866be465f/src/unix/linux_like/linux/gnu/b64/x86_64/mod.rs#L814-L815

~~Glibc implementation:~~

- ~~https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/arm/ioperm.c~~

~~The implementation has `unsigned int` for the argument of `iopl` but I've used int in the PR to match the Linux docs, which seems more authoritative. Unclear why glibc diverges from this but it doesn't make a difference in the ABI.~~

### DIFF
--- a/src/unix/linux_like/linux/gnu/b64/aarch64/align.rs
+++ b/src/unix/linux_like/linux/gnu/b64/aarch64/align.rs
@@ -49,3 +49,10 @@ s! {
         pub cgroup: ::c_ulonglong,
     }
 }
+
+extern "C" {
+    pub fn getcontext(ucp: *mut ucontext_t) -> ::c_int;
+    pub fn setcontext(ucp: *const ucontext_t) -> ::c_int;
+    pub fn makecontext(ucp: *mut ucontext_t, func: extern "C" fn(), argc: ::c_int, ...);
+    pub fn swapcontext(uocp: *mut ucontext_t, ucp: *const ucontext_t) -> ::c_int;
+}


thomcc (approved):
Seems fine. Matches declaration in headers, assuming our `struct ucontext` is right (which this PR doesn't add).

The iopl/ioperm stuff is a bit odd to have but I suppose this allows for following a consistent error handling path rather than `cfg`ing it out.

thomcc:
> The implementation has unsigned int for the argument of iopl but I've used int in the PR to match the Linux docs, which seems more authoritative

This also matches the way we [expose `iopl`](https://github.com/rust-lang/libc/blob/bbf929d2c8355fa19384b3551c5874c866be465f/src/unix/linux_like/linux/gnu/b64/x86_64/mod.rs#L814) on x86, which is somewhat important.

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

dtolnay:
```console
/checkout/target/aarch64-unknown-linux-gnu/debug/build/libc-test-5a973a03a651b470/out/main.c: In function '__test_fn_iopl':
/checkout/target/aarch64-unknown-linux-gnu/debug/build/libc-test-5a973a03a651b470/out/main.c:53524:24: error: 'iopl' undeclared (first use in this function)
53524 |                 return iopl;
      |                        ^~~~
/checkout/target/aarch64-unknown-linux-gnu/debug/build/libc-test-5a973a03a651b470/out/main.c:53524:24: note: each undeclared identifier is reported only once for each function it appears in
/checkout/target/aarch64-unknown-linux-gnu/debug/build/libc-test-5a973a03a651b470/out/main.c: In function '__test_fn_ioperm':
/checkout/target/aarch64-unknown-linux-gnu/debug/build/libc-test-5a973a03a651b470/out/main.c:53529:24: error: 'ioperm' undeclared (first use in this function); did you mean 'ipc_perm'?
53529 |                 return ioperm;
      |                        ^~~~~~
      |                        ipc_perm
```

I did a bit of digging in glibc, it turns out glibc ships *definitions* of `iopl` and `ioperm` on aarch64 but does not ship a *declaration* of them in any header. :exploding_head: I've removed those two from the PR.

dtolnay:
Can't tell what failed. :crying_cat_face:

thomcc:
https://github.com/rust-lang/libc/actions/runs/3502179666/jobs/5866446110#step:5:2455 is supposed to print `^PASSED .* tests` or `^test result: ok` to stdout. https://github.com/rust-lang/libc/blob/master/ci/runtest-android.rs#L41-L47. I'm not sure why this is happening though.

It doesn't look like it's caused by this PR, but it also doesn't seem to be failing in other PRs.

thomcc:
Should be fixed by https://github.com/rust-lang/libc/pull/3005

JohnTitor:
Sorry for the inconvenience! @bors retry

JohnTitor:
CI for 1.13 fails because it (`ucontext_t` precisely) needs `libc_align`. Could you declare them on align.rs? (We could remove this cfg after setting MSRV to a more recent version but for now let's follow the current policy.)

dtolnay:
:+1: Moved to align.rs so that they're properly conditional on `libc_align`.

JohnTitor:
Thanks! @bors r+

PR #3003, opened 2022-11-17
linux musl adding `PIDFD_NONBLOCK` constant.

### LINKED ISSUE #3002: `PIDFD_NONBLOCK` missing from Linux musl toolchains
`PIDFD_NONBLOCK` is an alias for `O_NONBLOCK` used with [`pidfd_open`](https://man7.org/linux/man-pages/man2/pidfd_open.2.html).

It is missing from all the `musl` toolchains.

These toolchains do however have `SYS_pidfd_open` which would suggest they should have this value.

devnexen:
closes #3002


JohnTitor:
Could you add them to https://github.com/rust-lang/libc/blob/master/libc-test/semver/linux-musl.txt?

JohnTitor (approved):

src/unix/linux_like/linux/musl/mod.rs
```diff
@@ -590,6 +590,9 @@ pub const EFD_NONBLOCK: ::c_int = ::O_NONBLOCK;
 
 pub const SFD_NONBLOCK: ::c_int = ::O_NONBLOCK;
 
+pub const SOCK_NONBLOCK: ::c_int = O_NONBLOCK;
```
JohnTitor, line 593: Seems we already have `SOCK_NONBLOCK`

PR #3007, opened 2022-11-20
Fix the loongarch64 kernel ABI

xen0n:
The initial loongarch64 support code went in too early, even before the upstream kernel ABI has finalized, and was not adjusted since then. No one with enough knowledge of LoongArch was involved in the initial review, so we have been shipping broken LoongArch support, but luckily the rustc port is not merged yet so no real damage has been done.

Fix the following discrepancies:

- There is no longer {g,s}etrlimit, only prlimit64.
- There is no longer fstat and newfstatat, only statx.
- MINSIGSTKSZ and SIGSTKSZ now have different values.
- The binary sysctl syscall was removed from Linux long before the existence of upstream Linux/LoongArch port (5.5 vs 5.19) so even a wrapper does not make sense.

There might be more but these are the most obvious.

cc @xry111 @zhaixiaojuan

### DIFF
--- a/src/unix/linux_like/linux/gnu/b64/loongarch64/mod.rs
+++ b/src/unix/linux_like/linux/gnu/b64/loongarch64/mod.rs
@@ -325,8 +325,6 @@ pub const SYS_vmsplice: ::c_long = 75;
 pub const SYS_splice: ::c_long = 76;
 pub const SYS_tee: ::c_long = 77;
 pub const SYS_readlinkat: ::c_long = 78;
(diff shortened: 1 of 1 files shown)


xry111:
LGTM.

JohnTitor (approved):

JohnTitor:
Thank you for the detailed explanation!

> but luckily the rustc port is not merged yet so no real damage has been done.

I'm going to r+ in favor of the above, @bors r+

Issue #3004, opened 2022-11-17
`sockaddr_storage` shouldn't have internal padding bytes


stevenengler:
A common use-case of `sockaddr_storage` is to store a socket address in a `sockaddr_storage`, inspect the `ss_family` field and the socket address length, and then cast it to the correct socket address type (for example `sockaddr_in`). For example, this is what the nix crate currently does using a union: https://github.com/nix-rust/nix/blob/fbebb21dd8df447a1408795b4b5706d9ca6c55df/src/sys/socket/addr.rs#L1537-L1540

The rust libc library defines `sockaddr_storage` [as the following](https://docs.rs/libc/0.2.137/src/libc/unix/linux_like/mod.rs.html#232) on Linux:

```rust
pub struct sockaddr_storage {
    pub ss_family: sa_family_t,
    __ss_align: ::size_t,
    #[cfg(target_pointer_width = "32")]
    __ss_pad2: [u8; 128 - 2 * 4],
    #[cfg(target_pointer_width = "64")]
    __ss_pad2: [u8; 128 - 2 * 8],
}
```

On Linux x86-64, this results in the following representation:

```
type: `unix::linux_like::sockaddr_storage`: 128 bytes, alignment: 8 bytes
    field `.ss_family`: 2 bytes
    padding: 6 bytes
    field `.__ss_align`: 8 bytes, alignment: 8 bytes
    field `.__ss_pad2`: 112 bytes
```

From what I understand, the 6 padding bytes in `sockaddr_storage` should cause a cast from memory initialized as a `sockaddr_storage` to another type like `sockaddr_in` to invoke UB since these padding bytes are considered to be uninitialized memory. For example these padding bytes of `sockaddr_storage` overlap with the `sockaddr_in.sin_port` field, so reading the `sin_port` field after casting from a `sockaddr_storage` would be UB. And in the nix case, the `ss: libc::sockaddr_storage = mem::zeroed()` may also not zero these 6 padding bytes, possibly leading to UB when it's later read as a different socket address type if those padding bytes weren't overwritten during the `ptr::copy`.

On recent glibc versions (such as 2.31 on Ubuntu 20.04), they define `sockaddr_storage` as:

```c
#define __ss_aligntype  unsigned long int
#define _SS_PADSIZE \
  (_SS_SIZE - __SOCKADDR_COMMON_SIZE - sizeof (__ss_aligntype))

struct sockaddr_storage
  {
[...]

PR #3010, opened 2022-11-21
Rearrange `sockaddr_storage` padding/alignment fields on Linux and Fuchsia

stevenengler:
Part of #3004.

Previously on Linux, the `sockaddr_storage` structure had padding bytes between the `ss_family` and `__ss_align` fields. The `__ss_align` field has now been moved to the end of the structure to eliminate these padding bytes, matching recent glibc versions: https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/bits/socket.h;h=4f1f810ea1d9bf00ff428e4e7c49a52c71620775;hb=c804cd1c00adde061ca51711f63068c103e94eef#l190

After the PR on Linux x86-64:

```
print-type-size type: `unix::linux_like::sockaddr_storage`: 128 bytes, alignment: 8 bytes
print-type-size     field `.ss_family`: 2 bytes
print-type-size     field `.__ss_pad2`: 118 bytes
print-type-size     field `.__ss_align`: 8 bytes
```

These moved fields are private but they are used in the `sockaddr_storage`s `PartialEq`, `Debug`, and `Hash` implementations, so the exact behaviour may change slightly, but I don't think anyone should be depending on this.

~(It looks like [Fuchsia](https://github.com/rust-lang/libc/blob/73c25f4e9d66054d1496c693b72d60caea00c4a9/src/fuchsia/mod.rs#L910) has the same issue, but I didn't modify its structure because I don't know much about it, or how to test it.)~

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -909,8 +909,8 @@ s_no_extra_traits! {
 
     pub struct sockaddr_storage {
         pub ss_family: sa_family_t,
+        __ss_pad2: [u8; 128 - 2 - 8],
         __ss_align: ::size_t,
-        __ss_pad2: [u8; 128 - 2 * 8],
     }
 
     pub struct utsname {
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -231,11 +231,11 @@ s_no_extra_traits! {
 
     pub struct sockaddr_storage {
         pub ss_family: sa_family_t,
-        __ss_align: ::size_t,
         #[cfg(target_pointer_width = "32")]
-        __ss_pad2: [u8; 128 - 2 * 4],
+        __ss_pad2: [u8; 128 - 2 - 4],
         #[cfg(target_pointer_width = "64")]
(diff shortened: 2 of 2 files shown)


asomers:
I think Fuchsia's definition is here, and yes I think you should fix it in libc.
https://fuchsia.googlesource.com/fuchsia/+/refs/heads/main/zircon/third_party/ulib/musl/include/sys/socket.h#376

stevenengler:
> I think Fuchsia's definition is here, and yes I think you should fix it in libc. https://fuchsia.googlesource.com/fuchsia/+/refs/heads/main/zircon/third_party/ulib/musl/include/sys/socket.h#376

Added a fixup commit to change Fuchsia's definition.

asomers (approved):

JohnTitor (approved):
The upstream change is old enough and looking at https://github.com/rust-lang/libc/issues/1412 and https://github.com/bminor/glibc/releases/tag/glibc-2.26, we could assume the effect is quite limited.

PR #3013, opened 2022-11-23
handle c circular dependence (linux gnu)

BelovDV:
Repeating libc allows solve [this fixme](https://github.com/rust-lang/rust/blob/master/compiler/rustc_codegen_ssa/src/back/link.rs#:~:text=//%20HACK/FIXME%3A%20Fixup,the%20libc%20crate.) in rust compiler.


petrochenkov:
This is not exactly what the hack in rustc does, but `-lc -lgcc_eh -lgcc -lc` implemented here should work fine, I checked what interdependencies these libraries have.

JohnTitor (approved):

JohnTitor:
Thank you for the PR and review! @bors r=JohnTitor,petrochenkov

PR #3015, opened 2022-11-26
adding a handful of get/set*ent calls for android.

### LINKED ISSUE #3014: `get`/`setpwent` is missing on aarch64 android
The `get`/`setpwent` functions are not available on the aarch64-linux-android target, however libc on this platform (at least on my phone) does have this symbol in the libc library.
I'm happy to add this api, but it would be my first PR, so I wanted to first check there wasn't some reason not to include it.

Also I don't know if I'd need to add some more functions/types?
A dependency of a package I'm using doesn't compile because of this issue, I don't actually know the api at all, so any help would be appreciated.

This api is available in this package for x86_64 linux, so would it just be a case of porting that code over?

devnexen:
close #3014

### DIFF
--- a/libc-test/semver/android.txt
+++ b/libc-test/semver/android.txt
@@ -2915,6 +2915,7 @@ dlsym
 dup
 dup2
 duplocale
+endgrent
 endservent
 epoll_create
 epoll_create1
@@ -3014,6 +3015,7 @@ getegid
 getenv
 geteuid
 getgid
+getgrent
 getgrgid
 getgrgid_r
 getgrnam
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):

JohnTitor:
The failure is:

```
 CANNOT LINK EXECUTABLE "/data/local/tmp/main-7b710ea257d908af": cannot locate symbol "setpwent" referenced by "/data/local/tmp/main-7b710ea257d908af"...
```

Is it missing on Arm or is our emulator old? Could you check?

PR #3017, opened 2022-11-26
Add `get`/`set`/`endpwent` to android target

Z-snails:
See #3014.


Z-snails:
Oh this has been done already in #3015.

PR #3018, opened 2022-11-26
Migrate from highfive to triagebot

ehuss:
This migrates this repository from using the highfive bot to using triagebot (aka rustbot).

This should not be merged without coordinating the removal of the highfive webhook and/or merging https://github.com/rust-lang/highfive/pull/433.

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -3,4 +3,23 @@ allow-unauthenticated = [
     "C-*", "O-*", "S-*"
 ]
 
+[autolabel."S-waiting-on-review"]
+new_pr = true
+
 [assign]
+contributing_url = "https://github.com/rust-lang/libc/blob/master/CONTRIBUTING.md"
+
+[assign.owners]
+"*" = ["@JohnTitor"]
+
+[mentions."src/unix/bsd/netbsdlike/openbsd"]
+message = "Some changes occurred in OpenBSD module"
+cc = ["@semarie"]
+
+[mentions."src/unix/bsd/netbsdlike/mod.rs"]
+message = "Some changes occurred in OpenBSD module"
+cc = ["@semarie"]
+
+[mentions."src/unix/solarish"]
+message = "Some changes occurred in solarish module"
+cc = ["@jclulow", "@pfmooney"]


JohnTitor (approved):

JohnTitor:
Thank you! @bors r+

ehuss:
Oh, please don't merge yet. This needs infra work first.

> **This should not be merged**

ehuss:
@JohnTitor This should not have been merged. I'm concerned that the bots will conflict with one another and possibly get into a loop. Can you revert it please?

JohnTitor:
Ah sorry, I didn't read the description carefully!
However, is there any blocker of https://github.com/rust-lang/highfive/pull/433? I'm not sure why it's still open.

JohnTitor:
> I'm concerned that the bots will conflict with one another and possibly get into a loop.

Also, I'm not sure how the loop is caused, is it an implementation issue? If I opened a revert PR, the issue could occur, right? Then I think waiting for the above highfive PR to be merged would be fine as the rate of PRs here isn't high.

ehuss:
> However, is there any blocker of [rust-lang/highfive#433](https://github.com/rust-lang/highfive/pull/433)? I'm not sure why it's still open.

We have to wait for infra to do it.  It can take time for people to get around to things, and this wasn't intended to be something dealt with quickly.

> Also, I'm not sure how the loop is caused, is it an implementation issue?

I haven't tested, but the concern is that both bots will be handling `r?` commands. In the welcome message, both bots print an `r?` text. One bot will respond to the other's command and vice-versa. I'm not sure if they will get in a war with each other trying to set the assignment.  I suppose since they only print one message, it should only result in one duplicate, but I'm not sure. There are additional error message comments that get posted that may contain additional `r?` commands, I don't remember.

> If I opened a revert PR, the issue could occur, right?

Hm, I'm not sure, possibly. Is the master branch protected? I don't suppose you can push directly to it?

JohnTitor:
Thanks for clarifying! I see, I'm going to revert then. At least, highfive will print the below for my PRs and `r?` loops won't happen, I guess.

`@JohnTitor: no appropriate reviewer found, use r? to override`

PR #3023, opened 2022-11-30
Enable copy_file_range on FreeBSD

asomers:
PR #2479 did this, but only in the freebsd13 and freebsd14 modules, which was incorrect.  Those modules should only be used for functions that change across FreeBSD versions, and therefore need different ELF symbol versions.  Functions that were newly added since FreeBSD 11 can still go in the base freebsd module.  It will cause no problems for them to be there, and users will see an error at link time if they try to use such a function in an environment that is too old to support it.


asomers:
cc @devnexen

asomers:
This should fix CI on FreeBSD 12

JohnTitor (approved):

JohnTitor:
Makes sense! @bors r+

PR #3024, opened 2022-12-02
redox: long is 32-bits on 32-bit systems

jackpot51:
This ensures that 32-bit systems use i32 for c_long and u32 for c_ulong.

I have also adjusted off_t and time_t to be `long long`, which is what they are in `relibc` now.


JohnTitor (approved):

JohnTitor:
Thank you! @bors r+

PR #3027, opened 2022-12-05
linux: Add AT_SYSINFO_EHDR constant

Phantomical:
When passed to [`getauxval(3)`][2] this is used to retrieve the base address of the vdso mapped into the current process. It is defined the architecture specific `auxvec.h` headers but [it is defined to the same value in all of them][1] so I've put it in `src/unix/linux_like/linux/mod.rs` instead of in the arch submodule.

[1]: https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/torvalds/linux%24+%23define+AT_SYSINFO_EHDR&patternType=standard&sm=1
[2]: https://man7.org/linux/man-pages/man3/getauxval.3.html


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #3030, opened 2022-12-07
Try readding all inotify flags

carbotaniuman:
Adds back some inotify flags that were once too new, but it appears the inexorable process of time has made the tests pass now.


JohnTitor (approved):

JohnTitor:
Looks good, thank you! @bors r+

PR #3035, opened 2022-12-12
Adding RTEXT_FILTER* constants from linux/rtnetlink.h

jreppnow:
Adding some constants needed for some specific netlink-route interactions (fetching information on a network interface). They are defined at ```linux/rtnetlink.h```.


jreppnow:
The ```RTEXT_FILTER_MST``` constant is available locally for me (Linux 6.0.12), but was not available on your CI and presumably is not available on the majority of active Linux systems yet (added to the kernel header ~9 months ago), so I removed it for this PR. I personally don't need it for my use case.

JohnTitor:
Thanks! @bors r+

jreppnow:
The CI seems to fail since MUSL is still using modified headers based on Linux 4.19.* (https://github.com/sabotage-linux/kernel-headers), which are also missing ```RTEXT_FILTER_MRP``` (added 3 years ago), ```RTEXT_FILTER_CFM_CONFIG``` and ```RTEXT_FILTER_CFM_STATUS``` (added 2 years ago). Linux 4.19 is a longterm release with support until Dec, 2024. 

I only need ```RTEXT_FILTER_VF``` and ```RTEXT_FILTER_SKIP_STATS``` at the moment, so I would be fine with only adding the first four constants (all added more than 7 years ago), but I guess it would be nice if there was a way to have them included for systems where they are available? 

@JohnTitor What is your preference? Remove the "newer" constants and only add what's available for all LTR kernels? Or is there possibly a way to conditionally include this only for specific Linux versions?

JohnTitor:
Our musl image is too old because we haven't made a decision around the [time64](https://musl.libc.org/time64.html) breaking change.
Anyway, adding newer consts doesn't break the build for older musl users but the libc crate users have to care about it when using, so it should be totally fine to add them. Let's ignore them on the tests, you could tweak the below to do that:

https://github.com/rust-lang/libc/blob/72fba0ecc76bf385d8f7b687fa8e46a1f6c0ae43/libc-test/build.rs#L3385

jreppnow:
@JohnTitor Sorry for taking so long. I added the exception as requested, here is hoping it goes through this time.

JohnTitor:
Sorry for the delay!
@bors r+

jreppnow:
@JohnTitor I left in both exceptions/```if```s now, hope that's fine. 

I wonder if I can also re-try? @bors try

JohnTitor:
LGTM, thank you.
@bors r+

PR #3036, opened 2022-12-13
Add missing string-to-number functions from ISO C

LegionMammal978:
These should be the observable API changes:

- ISO C90 function `atol()` is added to all `fuchsia`, `unix`, `vxworks`, `wasi`, and `windows` targets.
- ISO C90 function `atof()` is added to `android`, `redox`, and `vxworks`.
- ISO C99 functions `atoll()`, `strtoll()`, and `strtoull()` are added to all `fuchsia`, `unix`, `vxworks`, `wasi`, and `windows` targets.

Also, I wasn't exactly sure where to insert the new functions, so I just tried to place them near related functions.

### DIFF
--- a/libc-test/semver/android.txt
+++ b/libc-test/semver/android.txt
@@ -2854,7 +2854,10 @@ arphdr
 arpreq
 arpreq_old
 atexit
+atof
 atoi
+atol
+atoll
 bind
 blkcnt_t
 blksize_t
@@ -3504,7 +3507,9 @@ strtod
 strtof
 strtok
 strtol
+strtoll
 strtoul
+strtoull
 strxfrm
 suseconds_t
 swapoff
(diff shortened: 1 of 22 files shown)


JohnTitor:
Do you really want to use all of the items? The huge addition makes review harder and may cause unneeded breaking change (if upstream changes ABI, it would introduce a breaking change here. It'd be troublesome when there's no real user but we have to think about the effect of that breaking change).
Also, changing the `clock_t` type is a breaking change and this PR cannot be merged until we (maintainers) figure out how to set the release policy and ship breaking changes, so I'd recommend dropping the changes that you don't need and separating diff into some PRs.

LegionMammal978:
Alright, I've scaled this PR back to just the string-to-number functions, since those were what I was originally missing.

JohnTitor (approved):

JohnTitor:
Great, thanks! @bors r+

PR #3037, opened 2022-12-14
Add support for the unstable `check-cfg` feature behind an environment variable

pietroalbini:
`check-cfg` ([Rust](https://doc.rust-lang.org/stable/unstable-book/compiler-flags/check-cfg.html), [Cargo](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#check-cfg)) is an unstable features that warns when you write an unknown `#[cfg]` (likely due to a typo). The feature works out of the box for default cfgs and features provided by Cargo, but requires providing the list of extra cfgs when custom ones are used.

This PR adds the `LIBC_CHECK_CFG` environment variable. When enabled, the build script will use the `cargo:rustc-check-cfg` println to instruct the compiler of all the possible cfgs set by libc. The build script was also refactored to ensure all cfgs are accounted for, and a CI job using `-Z check-cfg` was added.

This PR is best reviewed commit-by-commit.

## Why is this needed?

The main motivation for this PR is that `rust-lang/rust` enforces `check-cfg` across the whole codebase. Normally this is not a problem for dependencies like `libc`, as Cargo caps the lints and thus doesn't show the generated warnings.

When developing support for new targets though, it's helpful to use a custom libc fork to develop the libc port and the std port together. Unfortunately doing that today results in a bunch of compilation errors, since lints are not capped with `path` dependencies. My goal with this PR is to address that shortcoming, as we'd then be able to set the `LIBC_CHECK_CFG=1` environment variable in the Rust build system and remove the compilation errors.

This PR might also be helpful for libc maintainers, as the CI check might spot typos in `#[cfg]`s.

### DIFF
--- a/.github/workflows/bors.yml
+++ b/.github/workflows/bors.yml
@@ -317,6 +317,23 @@ jobs:
         run: LIBC_CI=1 TOOLCHAIN=${{ matrix.toolchain }} WIN_TARGET=${{ matrix.target }} sh ./ci/build.sh
         shell: bash
 
+  check_cfg:
+    permissions:
+      actions: write # to cancel workflows (rust-lang/simpleinfra/github-actions/cancel-outdated-builds)
+      contents: read # to fetch code (actions/checkout)
+
+    name: "Check #[cfg]s"
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: rust-lang/simpleinfra/github-actions/cancel-outdated-builds@master
+        with:
+          github_token: "${{ secrets.GITHUB_TOKEN }}"
+      - uses: actions/checkout@v3
+      - name: Setup Rust toolchain
+        run: TOOLCHAIN=nightly sh ./ci/install-rust.sh
+      - name: Build with check-cfg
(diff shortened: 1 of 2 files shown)


JohnTitor:
Thanks for the detailed explanation, that sounds sensible! Could you fix Cirrus CI's failures?

pietroalbini:
I'll get to that in the new year (this is work-related and I'm on vacation).

pietroalbini:
Rebased on top of master. I have no clue why FreeBSD is failing though, my PR only changes how `cfg`s are set in the build script :(

pietroalbini:
Turns out the problem was enabling `rerun-if-env-changed`, which did hide some of the default behavior of Cargo... Removed it, this is ready for review.

JohnTitor:
Sorry for the delay, I'll review in a few days!

JohnTitor (approved):
Thanks!

JohnTitor:
Could you resolve the merge conflict (I made it to remove the semver checks, sorry)?

pietroalbini:
Rebased, and added `ohos` to both `target_os` and `target_env`.ò

JohnTitor:
Great, thank you! @bors r+

JohnTitor:
@bors delegate+
r=me once the old rustc becomes happy.

PR #3040, opened 2022-12-18
NetBSD: add two more errno values

0-wiz-0:
available in NetBSD-current since 2020

### DIFF
--- a/libc-test/semver/netbsd.txt
+++ b/libc-test/semver/netbsd.txt
@@ -1542,3 +1542,5 @@ EXTATTR_NAMESPACE_EMPTY
 extattr_list_fd
 extattr_list_file
 extattr_list_link
+EOWNERDEAD
+ENOTRECOVERABLE
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -1423,6 +1423,12 @@ pub const ENOLINK: ::c_int = 95;
 pub const EPROTO: ::c_int = 96;
 pub const EOWNERDEAD: ::c_int = 97;
 pub const ENOTRECOVERABLE: ::c_int = 98;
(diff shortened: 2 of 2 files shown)


workingjubilee:
Why would this be a "breakage candidate"? Because it changes `ELAST`? `ELAST` on OS that use this convention is defined as being the largest errno. It's only intended to be used to allow checks to see if an error is within the known values for errno, and things of that nature, it's not really directly used by any API. It is "versioning in userspace", as it were.

JohnTitor:
> Why would this be a "breakage candidate"? Because it changes `ELAST`? `ELAST` on OS that use this convention is defined as being the largest errno. It's only intended to be used to allow checks to see if an error is within the known values for errno, and things of that nature, it's not really directly used by any API. It is "versioning in userspace", as it were.

Yes, and I think it _could_ cause a bug (or any weird behavior due to lack of errno) if the last value points to a newer number while the user env is older.
As you mentioned on https://github.com/rust-lang/libc/issues/3131, we just should not provide such a const, I think.

workingjubilee:
Noting that https://github.com/rust-lang/libc/pull/1902 was handled by leaving the value in place and deprecating it, though that time we said we'd change it.

JohnTitor:
@0-wiz-0 Could you apply this?

> * Merge this change with the updated `ELAST` value, but also with a deprecation notice. People who are using `ELAST` "correctly" don't have to change anything immediately, so it doesn't introduce bugs in code that was written to be future-proof, but they can use the deprecation notice to prepare. People who were using it incorrectly have a heads up.

@rustbot author

0-wiz-0:
@JohnTitor I've added a deprecation warning as requested. Did I do it correctly?

JohnTitor (approved):

JohnTitor:
Yeah, thanks!
@bors r+

PR #3041, opened 2022-12-20
adding KERNEL_VERSION macro for linux.

devnexen:


JohnTitor:
Could we have a test like sigrt?

libc-test/test/linux_kernel_version.rs
```diff
@@ -0,0 +1,17 @@
+//! Compare libc's SIGRTMAX and SIGRTMIN functions against the actual C macros
```
JohnTitor, line 1: This comment should also be updated.

JohnTitor:
Thanks :)
@bors r+

PR #3043, opened 2022-12-22
linux-like: add AT_RECURSIVE constant

lucab:
This adds the `AT_RECURSIVE` constant on Linux and Android.
It is defined as part of the kernel uapi in `fcntl.h`:
 * https://github.com/torvalds/linux/blob/v6.1/include/uapi/linux/fcntl.h#L112


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

Issue #3045, opened 2022-12-23
Missing `struct ctl_info` of MacOS


siegfried:
This structure is used with the `CTLIOCGINFO` ioctl to translate from a kernel control name to a control id.
`CTLIOCGINFO` is missing as well.

PR #3046, opened 2022-12-23
Add struct ctl_info

siegfried:
Add the struct mentioned in #3045 

Not sure if I should add `CTLIOCGINFO ` because it is wrapped in a macro:

```c
#define CTLIOCGINFO     _IOWR('N', 3, struct ctl_info)  /* get id from name */
```

The C definition of `ctl_info` is:
```c
/*!
 *       @defined MAX_KCTL_NAME
 *   @discussion Kernel control names must be no longer than
 *       MAX_KCTL_NAME.
 */
#define MAX_KCTL_NAME   96

/*
 * Controls destined to the Controller Manager.
 */

/*!
 *       @struct ctl_info
 *       @discussion This structure is used with the CTLIOCGINFO ioctl to
 *               translate from a kernel control name to a control id.
 *       @field ctl_id The kernel control id, filled out upon return.
 *       @field ctl_name The kernel control name to find.
 */
struct ctl_info {
	u_int32_t   ctl_id;                             /* Kernel Controller ID  */
	char        ctl_name[MAX_KCTL_NAME];            /* Kernel Controller Name (a C string) */
};
```

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -83,6 +83,11 @@ fn main() {
         println!("cargo:rustc-cfg=libc_non_exhaustive");
     }
 
+    // Rust >= 1.47 supports long array:
+    if rustc_minor_ver >= 47 || rustc_dep_of_std {
+        println!("cargo:rustc-cfg=libc_long_array");
+    }
+
     if rustc_minor_ver >= 51 || rustc_dep_of_std {
         println!("cargo:rustc-cfg=libc_ptr_addr_of");
     }
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -3313,6 +3313,8 @@ pub const MINCORE_MODIFIED: ::c_int = 0x4;
 pub const MINCORE_REFERENCED_OTHER: ::c_int = 0x8;
 pub const MINCORE_MODIFIED_OTHER: ::c_int = 0x10;
 
+pub const CTLIOCGINFO: c_ulong = 0xc0644e03;
+
 //
 // sys/netinet/in.h
 // Protocols (RFC 1700)
@@ -5929,5 +5931,12 @@ cfg_if! {
         pub use self::b64::*;
     } else {
         // Unknown target_arch
+    }
+}
+
+cfg_if! {
+    if #[cfg(libc_long_array)] {
+        mod long_array;
+        pub use self::long_array::*;
     }
 }
--- a/libc-test/semver/apple.txt
+++ b/libc-test/semver/apple.txt
@@ -266,6 +266,7 @@ CTLFLAG_RD
 CTLFLAG_RW
 CTLFLAG_SECURE
 CTLFLAG_WR
+CTLIOCGINFO
 CTLTYPE
 CTLTYPE_INT
 CTLTYPE_NODE
@@ -1844,6 +1845,7 @@ copyfile
 copyfile_flags_t
 cpu_subtype_t
 cpu_type_t
+ctl_info
 difftime
 dirfd
 disconnectx
--- a/src/unix/bsd/apple/long_array.rs
+++ b/src/unix/bsd/apple/long_array.rs
@@ -1,0 +1,8 @@
+s! {
+    pub struct ctl_info {
+        pub ctl_id: u32,
+        pub ctl_name: [::c_char; MAX_KCTL_NAME],
+    }
+}
+
+pub const MAX_KCTL_NAME: usize = 96;


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

src/unix/bsd/apple/mod.rs
```diff
@@ -231,6 +231,11 @@ impl ::Clone for sysdir_search_path_domain_mask_t {
 }
 
 s! {
+    pub struct ctl_info {
+        pub ctl_id: u32,
+        pub ctl_name: [::c_char; 96],
```
JohnTitor, line 236: The size here is too big for older rustc, we could nest the array to pass CI, see the below for example:
https://github.com/rust-lang/libc/blob/72fba0ecc76bf385d8f7b687fa8e46a1f6c0ae43/src/unix/bsd/apple/mod.rs#L833-L835

siegfried: Is there an option to compile differently for old and new `rustc`? I can image this workaround can cause some confusions when type checking.

JohnTitor: Maybe we could add cfg under https://github.com/rust-lang/libc/blob/master/build.rs.

siegfried: I add a cfg to compile `ctl_info` only for new `rustc`, please review. Thanks.

build.rs
```diff
@@ -40,6 +40,11 @@ fn main() {
         println!("cargo:rustc-cfg=libc_deny_warnings");
     }
 
+    // Rust > 1.13 supports long array:
+    if rustc_minor_ver > 13 || rustc_dep_of_std {
```
JohnTitor, line 44: Could you use `=>` to keep the consistency with other cfgs?

JohnTitor (commented):
It'd also be great if you could squash commits into one.

build.rs
siegfried: Done.

JohnTitor:
Could you squash commits into one to clean up the history?

siegfried:
> Could you squash commits into one to clean up the history?

Sure, if you have trouble squash merging on your side.

JohnTitor:
This repo uses bors and the bot only can merge PRs via merge commits. So you have to do that before merging.

siegfried:
Squashed.

JohnTitor:
👍, @bors r+

siegfried:
Hmm, what is wrong again?

JohnTitor:
Ah, that's because we tried to impl Copy/Clone for non-existing item on the `s!` macro. Sadly, we cannot apply the attribute for impl, so we have to use cfg_if to completely hide the struct on 1.13.

siegfried:
Thanks. Good to know. Now I've got it updated.

siegfried:
Ah damn, cannot pass style check then. Let try another fix.

siegfried:
Looks like `cfg_if!` is the only option here. Is there anyway to get around by the style check? Or is there any way I can define the struct out of `s!`?

JohnTitor:
One way to pass the style check would be to cut it out into a separate file, like https://github.com/rust-lang/libc/blob/master/src/unix/align.rs. Could you try it?

siegfried:
I'm confused by the new style error. Any idea? @JohnTitor

JohnTitor:
Ugh, it says `mod` should be placed at the end of the file.

JohnTitor (approved):
Awesome, thanks!

siegfried:
Still not working?

JohnTitor:
```
+ cargo +1.13.0 build -vv --no-default-features --target aarch64-apple-ios
warning: unused manifest key: package.categories
warning: unused manifest key: package.categories
warning: unused manifest key: package.categories
    Updating registry `[https://github.com/rust-lang/crates.io-index`](https://github.com/rust-lang/crates.io-index%60)
error: failed to parse registry's information for: serde

Caused by:
  Feature `playground` depends on `serde_derive` which is not an optional dependency.
Consider adding `optional = true` to the dependency
```

I believe it's unrelated to this PR, let me investigate after my day job.

JohnTitor:
https://github.com/rust-lang/crates.io/pull/5929 should fix the issue, let's try out, @bors retry

siegfried:
I rebased the commit to the tip.

JohnTitor:
The failure is valid:
```
error[E0277]: the trait bound `[i8; 96]: core::fmt::Debug` is not satisfied
   --> src/unix/bsd/apple/long_array.rs:4:9
    |
4   |           pub ctl_name: [::c_char; MAX_KCTL_NAME],
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `[i8; 96]` cannot be formatted using `:?`; if it is defined in your crate, add `#[derive(Debug)]` or manually implement it
    | 
   ::: src/lib.rs
    |
93  | / cfg_if! {
94  | |     if #[cfg(windows)] {
95  | |         mod fixed_width_ints;
96  | |         pub use fixed_width_ints::*;
...   |
156 | |     }
157 | | }
    |
    = help: the trait `core::fmt::Debug` is not implemented for `[i8; 96]`
    = note: required because of the requirements on the impl of `core::fmt::Debug` for `&[i8; 96]`
    = note: required for the cast to the object type `core::fmt::Debug`

error[E0277]: the trait bound `[i8; 96]: core::cmp::Eq` is not satisfied
   --> src/unix/bsd/apple/long_array.rs:4:9
    |
4   |           pub ctl_name: [::c_char; MAX_KCTL_NAME],
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `core::cmp::Eq` is not implemented for `[i8; 96]`
    | 
   ::: src/lib.rs
    |
93  | / cfg_if! {
94  | |     if #[cfg(windows)] {
95  | |         mod fixed_width_ints;
96  | |         pub use fixed_width_ints::*;
...   |
156 | |     }
157 | | }
    |
    = help: the following implementations were found:
              <[T; 19] as core::cmp::Eq>
              <[T; 22] as core::cmp::Eq>
              <[T; 10] as core::cmp::Eq>
              <[T; 32] as core::cmp::Eq>
            and 30 others
    = note: required by `core::cmp::AssertParamIsEq`

error[E0277]: the trait bound `[i8; 96]: core::hash::Hash` is not satisfied
   --> src/unix/bsd/apple/long_array.rs:4:9
    |
4   |           pub ctl_name: [::c_char; MAX_KCTL_NAME],
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `core::hash::Hash` is not implemented for `[i8; 96]`
    | 
   ::: src/lib.rs
    |
93  | / cfg_if! {
94  | |     if #[cfg(windows)] {
95  | |         mod fixed_width_ints;
96  | |         pub use fixed_width_ints::*;
...   |
156 | |     }
157 | | }
    |
    = help: the following implementations were found:
              <[T; 20] as core::hash::Hash>
              <[T] as core::hash::Hash>
              <[T; 23] as core::hash::Hash>
              <[T; 12] as core::hash::Hash>
            and 30 others
    = note: required by `core::hash::Hash::hash`

error[E0369]: binary operation `==` cannot be applied to type `[i8; 96]`
   --> src/unix/bsd/apple/long_array.rs:4:9
    |
4   |           pub ctl_name: [::c_char; MAX_KCTL_NAME],
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | 
   ::: src/lib.rs
    |
93  | / cfg_if! {
94  | |     if #[cfg(windows)] {
95  | |         mod fixed_width_ints;
96  | |         pub use fixed_width_ints::*;
...   |
156 | |     }
157 | | }
    |
    = note: an implementation of `std::cmp::PartialEq` might be missing for `[i8; 96]`

error[E0369]: binary operation `!=` cannot be applied to type `[i8; 96]`
   --> src/unix/bsd/apple/long_array.rs:4:9
    |
4   |           pub ctl_name: [::c_char; MAX_KCTL_NAME],
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | 
   ::: src/lib.rs
    |
93  | / cfg_if! {
94  | |     if #[cfg(windows)] {
95  | |         mod fixed_width_ints;
96  | |         pub use fixed_width_ints::*;
...   |
156 | |     }
157 | | }
    |
    = note: an implementation of `std::cmp::PartialEq` might be missing for `[i8; 96]`
```

1.19 cannot impl it and we have to use `s_no_extra_traits!` and manually impl them, I think.

siegfried:
How about support >= 1.20 only? Updated.

JohnTitor:
It will require 1.47, but it sounds good to me as I think it's old enough 👍

siegfried:
Sorry, I just saw this. Now I change the version to `>= 1.47`. Please review.

JohnTitor:
No worries, thanks! @bors r+

Issue #3044, opened 2022-12-23
`AF_SYSTEM` and `AF_SYS_CONTROL` types do not fit in `sockaddr_ctl`


siegfried:
I'm trying to assign [AF_SYSTEM](https://github.com/rust-lang/libc/blob/f4bc85111b2403a9c9cb49890305cb6b32c5af41/src/unix/bsd/apple/mod.rs#L3567) to [sockaddr_ctl.sc_family](https://github.com/rust-lang/libc/blob/f4bc85111b2403a9c9cb49890305cb6b32c5af41/src/unix/bsd/apple/mod.rs#L641) which is a `u_char`, but AF_SYSTEM is `c_int`.

`AS_SYS_CONTROL` should be `u16` in `sockaddr_ctl.ss_sysaddr` as well yet its definition is `c_int`.
```rust
pub const AF_SYS_CONTROL: ::c_int = 2;
```

tgross35:
Are these flags ever passed to functions?

`main` is now for 1.0, so your PR can probably be adjusted.

PR #3047, opened 2022-12-23
Fix AF_SYSTEM and AF_SYS_CONTROL

siegfried:
Trying to fix the issue #3044 

Here is the definition I found:

```c
/*!
 *       @struct sockaddr_ctl
 *       @discussion The controller address structure is used to establish
 *               contact between a user client and a kernel controller. The
 *               sc_id/sc_unit uniquely identify each controller. sc_id is a
 *               unique identifier assigned to the controller. The identifier can
 *               be assigned by the system at registration time or be a 32-bit
 *               creator code obtained from Apple Computer. sc_unit is a unit
 *               number for this sc_id, and is privately used by the kernel
 *               controller to identify several instances of the controller.
 *       @field sc_len The length of the structure.
 *       @field sc_family AF_SYSTEM.
 *       @field ss_sysaddr AF_SYS_KERNCONTROL.
 *       @field sc_id Controller unique identifier.
 *       @field sc_unit Kernel controller private unit number.
 *       @field sc_reserved Reserved, must be set to zero.
 */
struct sockaddr_ctl {
	u_char      sc_len;     /* depends on size of bundle ID string */
	u_char      sc_family;  /* AF_SYSTEM */
	u_int16_t   ss_sysaddr; /* AF_SYS_KERNCONTROL */
	u_int32_t   sc_id;      /* Controller unique identifier  */
	u_int32_t   sc_unit;    /* Developer private unit number */
	u_int32_t   sc_reserved[5];
};
```

Shall I alias AF_SYS_CONTROL to AF_SYS_KERNCONTROL?


src/unix/bsd/apple/mod.rs
```diff
@@ -3606,7 +3606,7 @@ pub const PF_ISDN: ::c_int = AF_ISDN;
 pub const PF_KEY: ::c_int = pseudo_AF_KEY;
 pub const PF_INET6: ::c_int = AF_INET6;
 pub const PF_NATM: ::c_int = AF_NATM;
-pub const PF_SYSTEM: ::c_int = AF_SYSTEM;
+pub const PF_SYSTEM: ::c_int = 32;
```
siegfried, line 3609: I changed the type of `PF_SYSTEM` to `c_int` regarding the man page of SOCKET(2):

```
NAME
     socket – create an endpoint for communication

SYNOPSIS
     #include <sys/socket.h>

     int
     socket(int domain, int type, int protocol);

DESCRIPTION
     socket() creates an endpoint for communication and returns a descriptor.

     The domain parameter specifies a communications domain within which communication will take place; this selects the protocol family which should be used.  These
     families are defined in the include file ⟨sys/socket.h⟩.  The currently understood formats are

           PF_LOCAL        Host-internal protocols, formerly called PF_UNIX,
           PF_UNIX         Host-internal protocols, deprecated, use PF_LOCAL,
           PF_INET         Internet version 4 protocols,
           PF_ROUTE        Internal Routing protocol,
           PF_KEY          Internal key-management function,
           PF_INET6        Internet version 6 protocols,
           PF_SYSTEM       System domain,
           PF_NDRV         Raw access to network device,
           PF_VSOCK        VM Sockets protocols

     The socket has the indicated type, which specifies the semantics of communication.  Currently defined types are:

           SOCK_STREAM
           SOCK_DGRAM
           SOCK_RAW

     A SOCK_STREAM type provides sequenced, reliable, two-way connection based byte streams.  An out-of-band data transmission mechanism may be supported.  A SOCK_DGRAM
     socket supports datagrams (connectionless, unreliable messages of a fixed (typically small) maximum length).  SOCK_RAW sockets provide access to internal network
     protocols and interfaces.  The type SOCK_RAW, which is available only to the super-user.
```

src/unix/bsd/apple/mod.rs
```diff
@@ -3564,11 +3564,11 @@ pub const AF_E164: ::c_int = AF_ISDN;
 pub const pseudo_AF_KEY: ::c_int = 29;
 pub const AF_INET6: ::c_int = 30;
 pub const AF_NATM: ::c_int = 31;
-pub const AF_SYSTEM: ::c_int = 32;
+pub const AF_SYSTEM: ::c_uchar = 32;
 pub const AF_NETBIOS: ::c_int = 33;
 pub const AF_PPP: ::c_int = 34;
 pub const pseudo_AF_HDRCMPLT: ::c_int = 35;
-pub const AF_SYS_CONTROL: ::c_int = 2;
+pub const AF_SYS_CONTROL: ::c_ushort = 2;
```
siegfried, line 3571: The old examples uses the value of `AF_SYS_CONTROL` for `AF_SYS_KERNCONTROL`. I can also add `AF_SYS_KERNCONTROL` and leave `AF_SYS_CONTROL` untouched.

siegfried:
Is there anything else needed from me?

JohnTitor:
Changing types is a breaking change and we don't ship it on minor releases. I'll revisit once we decide to do a major release.

tgross35:
We have an upcoming major release where we can do breaking changes.

It seems unusual to only change one `AF_*` type - should they all be changed? same for `PF_`

src/unix/bsd/apple/mod.rs
tgross35: Just change to `AF_SYSTEM as c_int` rather than duplicating the literal

siegfried:
Updated. Sorry for the delay. Please review.

tgross35:
Just to clarify; is `sc_family` only ever set to `AF_SYSTEM` and `ss_sysaddr` is only ever `AF_SYS_CONTROL`?

src/unix/bsd/apple/mod.rs
```diff
@@ -4066,14 +4066,14 @@ pub const AF_E164: c_int = AF_ISDN;
...
-pub const AF_SYSTEM: c_int = 32;
+pub const AF_SYSTEM: c_uchar = 32;
 pub const AF_NETBIOS: c_int = 33;
 pub const AF_PPP: c_int = 34;
 pub const pseudo_AF_HDRCMPLT: c_int = 35;
 pub const AF_IEEE80211: c_int = 37;
 pub const AF_UTUN: c_int = 38;
 pub const AF_VSOCK: c_int = 40;
-pub const AF_SYS_CONTROL: c_int = 2;
+pub const AF_SYS_CONTROL: c_ushort = 2;
```
tgross35, line 4076:
```suggestion
pub const AF_SYS_CONTROL: c_ushort = 2;
```

It looks like the field is `u_int16_t`, so this can just be `u16` https://developer.apple.com/documentation/kernel/sockaddr_ctl/1809252-ss_sysaddr.

siegfried:
> Just to clarify; is `sc_family` only ever set to `AF_SYSTEM` and `ss_sysaddr` is only ever `AF_SYS_CONTROL`? Meaning no other constants are ever used for these fields.

I think so regarding to the document.

src/unix/bsd/apple/mod.rs
siegfried: Updated. Waiting on CI.

tgross35 (approved):
LGTM, please squash. I think the CI failure is unrelated.

src/unix/bsd/apple/mod.rs
siegfried: The CI test failed after I change `AF_SYS_CONTROL` to `u16`. Any idea why? @tgross35

tgross35: It doesn't look related, the android CI has been flaky.

siegfried:
Squashed.

PR #3048, opened 2022-12-23
linux-like: add OPEN_TREE_CLONE and OPEN_TREE_CLOEXEC

lucab:
This adds the `OPEN_TREE_CLONE` and `OPEN_TREE_CLOEXEC` constants on Linux and Android.
Those are used by the `open_tree()` syscall and defined in `mount.h`:
 * https://github.com/torvalds/linux/blob/v6.1/include/uapi/linux/mount.h#L61-L65


src/unix/linux_like/mod.rs
```diff
@@ -616,6 +616,8 @@ pub const MS_LAZYTIME: ::c_ulong = 0x2000000;
 pub const MS_ACTIVE: ::c_ulong = 0x40000000;
 pub const MS_MGC_VAL: ::c_ulong = 0xc0ed0000;
 pub const MS_MGC_MSK: ::c_ulong = 0xffff0000;
+pub const OPEN_TREE_CLONE: ::c_uint = 0x01;
+pub const OPEN_TREE_CLOEXEC: ::c_uint = O_CLOEXEC as ::c_uint;
```
lucab, line 620: Note for reviewers: these constants are typed as `uint` because they are used in the corresponding `open_tree()` syscall, which is [defined](https://github.com/torvalds/linux/blob/v6.1/fs/namespace.c#L2504) with an `unsigned flags` argument:
```
SYSCALL_DEFINE3(open_tree, int, dfd, const char __user *, filename, unsigned, flags)
```

JohnTitor, line 620: These have been added in linux_like mod but I guess emscripten doesn't have them?

lucab: I'm honestly not very familiar with emscripten, but yes it doesn't seem to have these (also, no `open_tree()` at all).
At the same time, I think those are not defined in any C lib either, only in the Linux kernel uapi headers.

Is this a concern? Should I move these constant to `android/` and `linux/` instead?

JohnTitor: > I'm honestly not very familiar with emscripten, but yes it doesn't seem to have these (also, no `open_tree()` at all). At the same time, I think those are not defined in any C lib either, only in the Linux kernel uapi headers.

If I checked correctly, it doesn't have yet:
https://github.com/emscripten-core/emscripten/blob/6a93f7db16cd8e5214b4c3c08a361ccab9ffc67e/system/lib/libc/musl/include/sys/mount.h

> Is this a concern? Should I move these constant to `android/` and `linux/` instead?

It's not a big deal as the build passes fine anyway. But I'd prefer to add items only if a target has it, because it would produce a compile error instead of a weird/unclear behavior due to a wrong const/flag usage, I feel the former is more helpful for end users.

lucab: Ack. I've moved them to Android plus Linux only then.

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

lucab:
Amended to add mising `linux/mount.h` includes in `libc-test` logic.

lucab:
CI failure:
```
/checkout/target/i686-unknown-linux-musl/debug/build/libc-test-edac14eb712d5cd2/out/main.c:120:10: fatal error: linux/mount.h: No such file or directory
120 | #include <linux/mount.h>
    |          ^~~~~~~~~~~~~~~
```

I'm a bit out of clues about why it cannot find `linux/mount.h` specifically, while all the other `linux/*.h` kernel headers seem fine. Is there some weird setup for kernel headers in CI? How can I trigger bors tests myself to dig into this?

JohnTitor:
Hmm, our kernel version on musl is indeed quite old, IIUC but `mount.h` is more senior than it and I'm not sure why that happens.
You can find our CI structure here: https://github.com/rust-lang/libc/tree/master/ci
You're also now able to use `@bors try`, feel free to test on bors!
@bors delegate+

lucab:
Ok, so the problem with musl tests is that the headers snapshot at https://github.com/sabotage-linux/kernel-headers is from version `4.19.88`. There has been a rework/split of kernel headers after that, but anyway the `OPEN_TREE_*` constants are more recent than that version.

I've reworked the tests to take that into account, and bors CI is green now.
This PR should be ready for stamping and merging now.

PR #3051, opened 2022-12-25
linux: add rtnetlink ifinfomsg attribute enumerals

tones111:


src/unix/linux_like/linux/mod.rs
```diff
@@ -3019,6 +3019,19 @@ pub const ALG_SET_AEAD_AUTHSIZE: ::c_int = 5;
...
+pub const IF_OPER_NOTPRESENT: u8 = 1;
+pub const IF_OPER_DOWN: u8 = 2;
+pub const IF_OPER_LOWERLAYERDOWN: u8 = 3;
+pub const IF_OPER_TESTING: u8 = 4;
+pub const IF_OPER_DORMANT: u8 = 5;
+pub const IF_OPER_UP: u8 = 6;
+
+pub const IF_LINK_MODE_DEFAULT: u8 = 0;
+pub const IF_LINK_MODE_DORMANT: u8 = 1;
+pub const IF_LINK_MODE_TESTING: u8 = 2;
```
JohnTitor, line 3033: Should these really be `u8`? For example, `IFF_*` consts are declared as `c_int` and I'm concerned that type inconsistency confuses users.

tones111:
Updated to use c_int.

The kernel declares these symbols within enums ([if.h](https://github.com/torvalds/linux/blob/master/include/uapi/linux/if.h#L167-L182)) without specifying specific values, so it seems unlikely they'll exceed u8::MAX.  I think the compiler has some liberty to determine the underlying size of integer to use, but c_int is more consistent with the other symbols in this file.

Thanks for the review

src/unix/linux_like/linux/mod.rs
tones111: Updated to use c_int

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

JohnTitor:
Some items need to be excluded on sparc64 test.

tones111:
Thank goodness for CI.  I Think we've finally got this pull request through the checks.  Thanks for your help.

JohnTitor:
👍 @bors r+

tones111:
It looks like the sparc64 target is still failing, but I can't tell from the logs what's wrong.  Do you have any suggestions?

JohnTitor:
It's a timeout failure, so @bors retry

PR #3053, opened 2022-12-26
linux: add additional netlink interface attribute tags

tones111:


tones111:
It looks like this change depends on a newer version of linux api headers than what the CI is using.  What kernel version is the libc crate attempting to target?

JohnTitor:
> What kernel version is the libc crate attempting to target?

You could check it at https://github.com/rust-lang/libc/tree/master/ci/docker and related files under the ci dir.
Anyway, it's totally fine to skip tests for them until we bump the image. Could you add the consts here with a comment?
https://github.com/rust-lang/libc/blob/72fba0ecc76bf385d8f7b687fa8e46a1f6c0ae43/libc-test/build.rs#L3385

JohnTitor (approved):

JohnTitor:
Thank you! @bors r+

PR #3064, opened 2023-01-11
Fix compile error for riscv64-linux-android

colincross:
Fixes:
error[E0369]: cannot subtract `char` from `char`
  --> prebuilts/rust/linux-x86/1.64.0/src/stdlibs/vendor/libc/src/unix/linux_like/android/b64/riscv64/mod.rs:63:53
   |
63 | pub const COMPAT_HWCAP_ISA_I: ::c_ulong = 1 << ('I' - 'A');
   |                                                 --- ^ --- char
   |                                                 |
   |                                                 char


JohnTitor (approved):

JohnTitor:
Thank you for fixing it! @bors r+

PR #3072, opened 2023-01-18
add `user_regs` for linux on arm

Freax13:
This struct is used for `PTRACE_GETREGS` & `PTRACE_SETREGS`. We mirror the name used in `glibc`. This struct is called `pt_regs` in the kernel. Instead of a `uregs` array, we use separate named fields.

> - \[ ] Edit corresponding file(s) under `libc-test/semver` when you add/remove item(s)

AFAICT there's no file for linux on arm (not aarch64). Am I missing something here?


JohnTitor (approved):

JohnTitor:
If there's not, you can just ignore it :) Since we have a lot of supported env, covering everything is hard and ongoing.
@bors r+

PR #3074, opened 2023-01-19
linux: add PTRACE_SYSCALL_INFO_*

mbyzhang:
Add `PTRACE_SYSCALL_INFO_*` for `ptrace_syscall_info.op`.


src/unix/linux_like/linux/gnu/mod.rs
```diff
@@ -874,6 +874,10 @@ pub const PTRACE_INTERRUPT: ::c_uint = 0x4207;
 pub const PTRACE_LISTEN: ::c_uint = 0x4208;
 pub const PTRACE_PEEKSIGINFO: ::c_uint = 0x4209;
 pub const PTRACE_GET_SYSCALL_INFO: ::c_uint = 0x420e;
+pub const PTRACE_SYSCALL_INFO_NONE: ::__u8 = 0;
+pub const PTRACE_SYSCALL_INFO_ENTRY: ::__u8 = 1;
+pub const PTRACE_SYSCALL_INFO_EXIT: ::__u8 = 2;
+pub const PTRACE_SYSCALL_INFO_SECCOMP: ::__u8 = 3;
```
JohnTitor, line 880: Have you made sure the `__u8` type works well with other declarations? For example, if a fn needing these consts requires another type, it would result in causing type conversion.

mbyzhang: `PTRACE_SYSCALL_INFO_*` are basically variants of `ptrace_syscall_info.op`, which has type `__u8`. See https://github.com/devnexen/libc/blob/master/src/unix/linux_like/linux/gnu/mod.rs#L344. Currently there are no other declarations that need these consts.

JohnTitor (approved):

PR #3084, opened 2023-01-22
xous: add initial C definitions

xobs:
This adds initial C definitions to Xous. There is no C library, so this mostly serves to add C-compatible exports to the `libc` crate.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -151,6 +151,12 @@ cfg_if! {
 
         mod wasi;
         pub use wasi::*;
+    } else if #[cfg(target_os = "xous")] {
+        mod fixed_width_ints;
+        pub use fixed_width_ints::*;
+
+        mod xous;
+        pub use xous::*;
     } else {
         // non-supported targets: empty...
     }
--- a/src/xous.rs
+++ b/src/xous.rs
@@ -1,0 +1,49 @@
+//! Xous C type definitions
+
+pub type c_schar = i8;
+pub type c_uchar = u8;
+pub type c_short = i16;
+pub type c_ushort = u16;
+pub type c_int = i32;
+pub type c_uint = u32;
+pub type c_float = f32;
+pub type c_double = f64;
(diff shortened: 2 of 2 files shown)


xobs:
Note that I have NOT added a file to `libc/semver/` because Xous has no C functions, and so this file would be empty.

JohnTitor:
We're pausing accepting _external_ targets currently, see https://github.com/rust-lang/libc/pull/3032 for details.

xobs:
What is classified as an *external* target? Is it only tier 2 and lower?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "libc"
-version = "0.2.139"
+version = "0.2.140"
```
JohnTitor, line 3: Could you drop this? As mentioned in https://github.com/rust-lang/libc/blob/master/.github/PULL_REQUEST_TEMPLATE.md, A version bump PR shouldn't contain any other changes to avoid blocking a release when a review takes time.

JohnTitor:
Ah I thought there was no official target for xous but it's not true (that's what we discussed in the above linked PR, read it for the context). r=me once the above is resolved.

Cargo.toml
xobs: Sure, I misunderstood that statement and thought that a PR should contian a version bump as a commit without any other changes.

xobs:
Alright, I removed that commit. Note that the freebsd build is broken due to what apparently is a missing symbol in curl, preventing it from downloading things.

 r=johntitor

JohnTitor (approved):

JohnTitor:
Thanks!

> Note that the freebsd build is broken due to what apparently is a missing symbol in curl, preventing it from downloading things.

It has been fixed on master, it happens because the branch here is a bit old, so no problem at all :)

JohnTitor:
Oh I forgot: @bors r+

PR #3089, opened 2023-01-23
Add ioctl FS_IOC_{G,S}ETVERSION and FS_IOC_{G,S}ETFLAGS

ETKNeil:
This PR focus on adding support for the constants behind the call of ioctl(FS_IOC_G/SETFLAGS) and ioctl(FS_IOC_G/SETVERSION)

Ressources:
Linux: https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/fs.h#L220
Android: https://cs.android.com/android/platform/superproject/+/master:prebuilts/vndk/v31/arm/include/bionic/libc/kernel/uapi/linux/fs.h;l=155?q=FS_APPEND_FL&ss=android%2Fplatform%2Fsuperproject
Android (bis): https://cs.android.com/android/platform/superproject/+/master:bionic/libc/kernel/uapi/linux/fs.h;l=161;drc=4794e479f4b485be2680e83993e3cf93f0f42d03?q=FS_APPEND_FL&sq=&ss=android%2Fplatform%2Fsuperproject

Note: Thoses calls are a bit of a mess historically, however the constants did not change since


JohnTitor:
Seems the values are different on i686-unknown-linux-gnu.

ETKNeil:
> Seems the values are different on i686-unknown-linux-gnu.

Indeed, I fail to expand the macros, I did so for all the supported arch now

src/unix/linux_like/linux/arch/generic/mod.rs
```diff
@@ -204,14 +204,34 @@ pub const BLKIOOPT: ::Ioctl = 0x1279;
...
-pub const FS_IOC32_GETVERSION: ::Ioctl = 1;
-pub const FS_IOC32_SETVERSION: ::Ioctl = 2;
+cfg_if! {
+    // Those type are constructed using the _IOC macro
+    // DD-SS_SSSS_SSSS_SSSS-TTTT_TTTT-NNNN_NNNN
+    // where D stands for direction (either None (00), Read (01) or Write (11))
+    // where S stands for size (int, long, struct...)
+    // where T stands for type ('f','v','X'...)
+    // where N stands for NR (NumbeR)
+    if #[cfg(any(target_arch = "x86"))] {
```
JohnTitor, line 214: This `any` doesn't do anything and should go to an arch-specific file.

JohnTitor: https://doc.rust-lang.org/reference/conditional-compilation.html

ETKNeil: Uhm I don't see an arch specific for x86 and x86_64 thus why I added them to generic, note sparc, powerpc and mips which are all the arch specific supported by libc for linux-like are covered so by doing so I think all case are covered.

The superfluous any were removed

JohnTitor: You could make a new module like `mips`. This structure is for arch-dependent declarations and using cfgs would lose the meaning here.

JohnTitor:
Could you also fix the merge conflict and squash commits into one?

ETKNeil:
> Could you also fix the merge conflict and squash commits into one?

Done

JohnTitor (approved):

JohnTitor:
Sorry for the delay! @bors r+

fsavy-tehtris:
Could we retry this please? The CI logs have expired so we can't know what went wrong.

fsavy-tehtris:
@rustbot review

(not sure if this is the way to go, sorry for the noise)

fsavy-tehtris:
Looks like this is stuck in Bors queue @JohnTitor

JohnTitor:
Superseded by #3396, thanks anyway!

PR #3090, opened 2023-01-26
Solaris/Illumos: use correct types for getrandom(2) flags

josephlr:
On Solaris (and any other platform that supports it), the `getrandom(2)` syscall has signature:

```rust
fn getrandom(buf: *mut c_void, buflen: size_t, flags: c_uint) -> ssize_t;
```
so the flag constants (`GRND_NONBLOCK`, `GRND_RANDOM`, etc...) should be of type `c_uint`.

I'm not sure if this sort of "bug fix" counts as a breaking change, there weren't any Solaris/Illumos files under `libc-test/semver`.

Signed-off-by: Joe Richey <joerichey@google.com>

### DIFF
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -1280,8 +1280,8 @@ pub const FILENAME_MAX: ::c_uint = 1024;
 pub const L_tmpnam: ::c_uint = 25;
 pub const TMP_MAX: ::c_uint = 17576;
 
-pub const GRND_NONBLOCK: ::c_int = 0x0001;
-pub const GRND_RANDOM: ::c_int = 0x0002;
+pub const GRND_NONBLOCK: ::c_uint = 0x0001;
+pub const GRND_RANDOM: ::c_uint = 0x0002;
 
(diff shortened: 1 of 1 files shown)


jclulow (approved):
This looks right to me, and critically brings this platform into line with all the other ones, which was our intent at the OS level when we added this interface with a matching signature.  I think it regrettably suggests nobody actually compiled and used the code in the original PR that added these.

josephlr:
> I think it regrettably suggests nobody actually compiled and used the code in the original PR that added these.

I've actually been using these flags in the `getrandom` crate: https://github.com/rust-random/getrandom/blob/6536b9e60c0101ba7c555af90bdfd2e47c9a5384/src/solaris_illumos.rs#L41 but we need an `as` cast to make the flag the right type. I'd imagine the flags are very rarely used, because most of the time, `flags == 0`, which works regardless of the particular integer type.

JohnTitor (approved):

JohnTitor:
Thanks for the explanation! It makes sense to call it a bug fix if it currently needs an assertion.
@bors r+

PR #3093, opened 2023-01-27
Add wasi-emulated-mman cargo feature

### LINKED ISSUE #3091: Add emulated features from `wasi-libc` (like `mmap`)
When compiling to `wasm32-wasi`, `libc` does not include some emulated `wasi-libc` functions like `mmap`. When using `wasi-libc` directly, you can set some extra parameters to enable these emulated functions like:
```
-D_WASI_EMULATED_MMAN   # for clang
-lwasi-emulated-mman    # for lld
```

What would be the appropriate fix for this? I was hoping a cargo feature or something similar could solve the problem. I am a bit new here, so if I could get some guidance I would be happy to PR to supply these additions. (I see that changes would be needed for https://github.com/rust-lang/libc/blob/master/src/wasi.rs, but I am not sure what would need to change in build.rs or elsewhere to accommodate this.)

GregoryConrad:
Fixes #3091 by adding a `wasi-emulated-mman` cargo feature that adds bindings for `wasi-libc`'s emulated mman.

I am not sure if I implemented this completely correctly; feedback would be highly appreciated!

Some general comments/questions:
1. I did *not* implement `mmap64` here, as I was unsure how to handle C's `#define mmap64 mmap`. I only started learning Rust a couple months ago so I am not sure what would be analogous here.
2. I am not 100% sure what is up with the `_GNU_SOURCE` and `_BSD_SOURCE` conditional includes in the C header, so I included the relevant bindings in the hopes that I could get feedback on that from someone here.
3. ~The unix/linux-like `MAP_FAILED` is implemented as `pub const MAP_FAILED: *mut ::c_void = !0 as *mut ::c_void;`, which confused me. The `mman.h` in `wasi-libc` is `#define MAP_FAILED ((void *) -1)`. What is the correct rust binding for this constant?~ I failed to realize the `!0` is just the bitwise operator; all set here.
4. There is no `libc-test/semver/wasi.txt`?
5. And finally, a big newbie question: originally, I thought this library provided the target's libc (like wasi-libc), but that appears to be wrong as far as I can tell. How exactly does the target's libc get built & linked? Is this overridable using environment variables with cargo at build time (e.g. `CC="clang ... --sysroot=..." cargo build --target="..."`)? Or through some different mechanism?

### DIFF
--- a/src/wasi.rs
+++ b/src/wasi.rs
@@ -434,174 +434,106 @@ pub const NOEXPR: ::nl_item = 0x50001;
 pub const YESSTR: ::nl_item = 0x50002;
 pub const NOSTR: ::nl_item = 0x50003;
 
-cfg_if! {
-    if #[cfg(feature = "wasi-emulated-mman")] {
-        // wasi-libc @ 5d8a140, musl 1.2.3: libc-top-half/musl/include/sys/mman.h
+pub const MAP_FAILED: *mut ::c_void = !0 as *mut ::c_void;
 
-        pub const MAP_FAILED: *mut ::c_void = !0 as *mut ::c_void;
+pub const MAP_SHARED: ::c_int = 0x01;
+pub const MAP_PRIVATE: ::c_int = 0x02;
+pub const MAP_SHARED_VALIDATE: ::c_int = 0x03;
+pub const MAP_TYPE: ::c_int = 0x0f;
+pub const MAP_FIXED: ::c_int = 0x10;
+pub const MAP_ANON: ::c_int = 0x20;
+pub const MAP_ANONYMOUS: ::c_int = MAP_ANON;
+pub const MAP_NORESERVE: ::c_int = 0x4000;
+pub const MAP_GROWSDOWN: ::c_int = 0x0100;
+pub const MAP_DENYWRITE: ::c_int = 0x0800;
+pub const MAP_EXECUTABLE: ::c_int = 0x1000;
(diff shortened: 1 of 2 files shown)


src/wasi.rs
```diff
@@ -434,6 +434,175 @@ pub const NOEXPR: ::nl_item = 0x50001;
...
+                name = "c",
+                kind = "static",
+                modifiers = "-bundle",
+                cfg(target_feature = "crt-static")
+            )
+        )]
+        #[cfg_attr(
+            feature = "rustc-dep-of-std",
+            link(name = "c", cfg(not(target_feature = "crt-static")))
+        )]
```
GregoryConrad, line 554: I grabbed this from the `extern "C"` block later in the file. As far as I can tell, this just specifies how to handle static linking (adding the "bundle" modifier) when compiling to `crt-static` when building as a dep of std, so I copied it to this new extern block.
If this is wrong, please let me know!

JohnTitor:
As mentioned in https://github.com/rust-lang/libc/pull/3097#issuecomment-1416284583, we don't accept any platform-specific feature.

GregoryConrad:
@JohnTitor Ok, then what is the best course of action here? Remove the cargo feature (but keep my additions without the feature check), and if someone uses libc without the proper wasi-libc compilation flags enabled, they will just get linking/build failures?

JohnTitor:
It should be separated into another crate as it may require additional setup as you said. The libc crate should be minimal to be usable on many hosts/targets.

PR #3097, opened 2023-01-31
Hide `iconv` functions on Apple devices behind a feature flag

winterqt:
See #2870 and #2944 for context. I believe the comment in the code explains the changes well enough, but if you'd like me to add a description to the commit as well (or adjust any wording etc.), let me know.

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -5908,19 +5908,19 @@ cfg_if! {
 }
 
 // These require a dependency on `libiconv`, and including this when built as
-// part of `std` means every Rust program gets it. Ideally we would have a link
-// modifier to only include these if they are used, but we do not.
-#[cfg_attr(not(feature = "rustc-dep-of-std"), link(name = "iconv"))]
-extern "C" {
-    pub fn iconv_open(tocode: *const ::c_char, fromcode: *const ::c_char) -> iconv_t;
-    pub fn iconv(
-        cd: iconv_t,
-        inbuf: *mut *mut ::c_char,
-        inbytesleft: *mut ::size_t,
-        outbuf: *mut *mut ::c_char,
-        outbytesleft: *mut ::size_t,
-    ) -> ::size_t;
-    pub fn iconv_close(cd: iconv_t) -> ::c_int;
+// part of `std` means every Rust program gets it. Additionally, when not built
+// as part of `std`, merely using `libc` on an Apple target will pull in `libiconv`.
+//
+// Therefore, due to these functions very low usage numbers on the platform, we hide it
+// behind a feature flag.
+//
+// Ideally we would have a link modifier to only include these if they are used, but we do not.
+cfg_if! {
+    if #[cfg(feature = "iconv-apple")] {
+        #[link(name = "iconv")]
+        mod iconv;
+        pub use self::iconv::*;
+    }
 }
 
 cfg_if! {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ align = []
 rustc-dep-of-std = ['align', 'rustc-std-workspace-core']
 extra_traits = []
 const-extern-fn = []
+iconv-apple = []
 # use_std is deprecated, use `std` instead
 use_std = [ 'std' ]
 
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ libc = "0.2"
   If you use Rust >= 1.62, this feature is implicitly enabled.
   Otherwise it requires a nightly rustc.
 
+* `iconv-apple`: Enables usage of `iconv_open`, `iconv`, and `iconv_close` on Apple devices.
+
 * **deprecated**: `use_std` is deprecated, and is equivalent to `std`.
 
 ## Rust version support
--- a/src/unix/bsd/apple/iconv.rs
+++ b/src/unix/bsd/apple/iconv.rs
@@ -1,0 +1,11 @@
+extern "C" {
+    pub fn iconv_open(tocode: *const ::c_char, fromcode: *const ::c_char) -> iconv_t;
+    pub fn iconv(
+        cd: iconv_t,
+        inbuf: *mut *mut ::c_char,
+        inbytesleft: *mut ::size_t,
+        outbuf: *mut *mut ::c_char,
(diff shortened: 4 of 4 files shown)


src/unix/bsd/apple/mod.rs
```diff
@@ -5908,9 +5908,14 @@ cfg_if! {
...
-// modifier to only include these if they are used, but we do not.
-#[cfg_attr(not(feature = "rustc-dep-of-std"), link(name = "iconv"))]
+// part of `std` means every Rust program gets it. Additionally, when not built
+// as part of `std`, merely using `libc` on an Apple target will pull in `libiconv`.
+//
+// Therefore, due to these functions very low usage numbers on the platform, we hide it
+// behind a feature flag.
+//
+// Ideally we would have a link modifier to only include these if they are used, but we do not.
+#[cfg_attr(eature = "libiconv-apple", link(name = "iconv"))]
```
thomcc, line 5918:
```suggestion
#[cfg_attr(feature = "libiconv-apple", link(name = "iconv"))]
```

winterqt: Oops, thanks.

winterqt:
What would be the best path forward for `libc-test`? Should I just add the `apple-iconv` to its dependency on `libc`, or do something more complex?

JohnTitor:
Uhm, I hesitate to add a platform feature flag as we support a lot of platforms and the flag management can be complex easily 😱(we've done a similar thing to Rust versions and it introduces a ton of complexity, huh).
Perhaps we should have asked that those functions be separated into a platform-specific crate rather than added to the libc crate.
@rust-lang/crate-maintainers Any thoughts on this?

(By the way, this is a breaking change and I'd like to ship it in a major release.)

Amanieu:
I still don't understand what the problem is with linking against `iconv` on apple targets. This library is guaranteed to always be present, so it shouldn't cause any issues?

winterqt:
@Amanieu See @thomcc's excellent comment [here](https://github.com/rust-lang/libc/issues/2870#issuecomment-1271104762) for some info.

joshtriplett:
:+1: for just dropping these functions entirely in a subsequent major release, if we ever make a major release.

winterqt:
@JohnTitor @joshtriplett Do you think there's any chance of a major release happening within, say, a few months? A year? Just curious, I appreciate your insights, thanks! 😊

JohnTitor:
We're going to make a major change to musl in https://github.com/rust-lang/libc/pull/3068 and I guess we could publish this along with that PR. It will be merged once [FCP](https://github.com/rust-lang/libc/pull/3068#issuecomment-1381151618) passes.
I cannot tell the exact date and time for that, as it requires agreement by team members, but it could happen within a few months (hopefully).

thomcc:
> See @thomcc's excellent comment

That was about `libstd`'s usage. I'd don't love that we link to it in `libc`, but there are reasons I haven't pursued this.

> I still don't understand what the problem is with linking against `iconv` on apple targets. This library is guaranteed to always be present, so it shouldn't cause any issues?

It's mostly compelling since it's practically unused[^2], and linking with it means it must be available on the host[^host] breaks some minimalist (or perhaps naïve) cross-compilation setups, where a full SDK is not available, but a linker and a libSystem stub are (for example, this is the case with `zig`'s cross compilation support, unless you bring along the `.tbd` files you need). IOW, the library is guaranteed to be present on the target, but not on the host, and it is needed in both places (as linking a mach-o requires the linker know which dylib provides each symbol).

Concretely, this probably is one of the first things hit by folks who try to cross compile rust to macOS by using `zig {cc,ld}` directly. I'm unsure how reasonable it is to expect cross compilation to work in such cases, and users will probably hit something after this anyway, and are likely better off with something like `cargo-zigbuild` anyway, which shaves off these kinds of sharp edges (it [bundles](https://github.com/rust-cross/cargo-zigbuild/blob/0b7a8ee3acd9a11adc55d654eb44fc18f13901be/src/macos/libiconv.tbd) `libiconv.tbd` for this reason).

---

In any case, I don't think we should removing the items (even if they're nearly unused), and I *really* don't think we should add a feature flag for this. I do think it's annoying that Rust users may have to cart around semi-vestigial `libiconv.tbd` in their cross compilation setup, even if they don't need it (but perhaps if it weren't that it might be something else[^3]). So, I have two thoughts here:

1. If we care about that kind of use case, then we should just change the PR to remove the `#[link]` attribute and deprecate the items (with a message explaining a fix). I don't know if this is technically breaking, but it seems like it unlikely to break anybody.

2. If we don't care, then we probably should wait until `raw-dylib` is supported on macOS (which sadly seems [stalled](https://github.com/rust-lang/rust/pull/100703)). At that point, we could change the attribute to `#[link(name = "iconv", kind = "raw-dylib")]`. Then (IIUC) we'll generate the tbd automatically when compiling, avoiding such problems. That said, it may be quite a while before that is available on stable.

I do think the hassle this causes in practice is not nearly worth it given how few people (if any) depend on it, but I also think libc should not make the breaking API changes it does (altho I'm unsure how much the `link` removal would count...). So I don't know.

[^2]: [virtually nobody on github](https://cs.github.com/?scopeName=All+repos&scope=&q=libc%3A%3Aiconv+language%3Arust) use libc's declarations, and the code that does want to call `iconv` seems to [explicitly link anyway](https://cs.github.com/?scopeName=All+repos&scope=&q=%2Fname%5Cs*%3D%5Cs*%22iconv%22%2F+language%3Arust). Of course, neither of those searches are absolute proof of anything, but it's a ballpark idea of usage.

[^3]: IOW, perhaps it's better that cross compilation does not work out of the box. I don't think we want to make any guarantees that it will (even in `std`), nor do we want the bug reports that could come if it worked and then we began linking to another system library.

JohnTitor:
That's insightful, thank you!

> 1. If we care about that kind of use case, then we should just change the PR to remove the `#[link]` attribute and deprecate the items (with a message explaining a fix).

Do you mean we should deprecate items first and then remove them? If so, how about deprecating on 0.2.x and removing on 0.3.x (along with time64 change)?

workingjubilee:
Given that it seems much more likely that we will be pushing out a 0.3 release sometime "soon" (highly relatively speaking): Yeah, we should probably instead do a deprecation and `doc(hidden)`.

PR #3098, opened 2023-02-02
add SO_TS_* constants for FreeBSD

folkertdev:
These constants are used in timestamping of packets on a socket. The definitions come from https://github.com/bminor/newlib/blob/master/newlib/libc/sys/rtems/include/sys/socket.h#L165. 

For context, here is an example of some of these being used on FreeBSD for packet timestamping in NTP: https://github.com/ntpsec/ntpsec/blob/master/ntpd/ntp_packetstamp.c.


JohnTitor (approved):

JohnTitor:
Thanks 👍 @bors r+

PR #3106, opened 2023-02-06
Add BLKPG and BLKRRPART ioctls

IsaccBarker:
This adds the following IOCTLs to `src/unix/linux_like/linux/arch/generic/mod.rs`: BLKPG and BLKRRPART.

### BLKPG
Modifies the kernel's in-memory representation of the partition table.

### BLKRRPART
Force a re-read of the partition tables.

#### Note
Concerning bumping the crate version, should I have done that? The pre-pr checklist says to do so (or at least that's how I read it), but the contribution guidelines only mention it in the context of requiring a new release.


JohnTitor:
> Concerning bumping the crate version, should I have done that? The pre-pr checklist says to do so (or at least that's how I read it), but the contribution guidelines only mention it in the context of requiring a new release.

A PR shouldn't contain any other changes if it bumps up the crate version.
The exact wording is below:

> Your PR that bumps up the crate version doesn't contain any other changes

Is this confusing? I'm not a native English speaker and any suggestion is welcome!

IsaccBarker:
Ah, sorry! I think I got confused because I was thinking that "any other changes" referred to tacking on other changes unrelated to the PR. I didn't know why that needed to be mentioned, but that makes much more sense, thank you!

As for the `x86_64-unknown-linux-gnu` CI check failing, I think that's because it's not pulling in the header where these things are defined. I'm not entirely sure how to fix that.

JohnTitor:
> I didn't know why that needed to be mentioned

The reason is that a PR is often stuck with a failure like yours and it can delay a release 🚢 

> As for the `x86_64-unknown-linux-gnu` CI check failing, I think that's because it's not pulling in the header where these things are defined. I'm not entirely sure how to fix that.

Our CI config is on https://github.com/rust-lang/libc/tree/master/ci/docker and the root `ci` dir.
In this case, I think the kernel version is old and we have to ignore test for them. You could do so by tweaking the test like:
https://github.com/rust-lang/libc/blob/78d774fbc1c256d6cfae8ea71082b4dc68724634/libc-test/build.rs#L3683

The Linux version comment is helpful when we upgrade CI images and remove FIXME.

IsaccBarker:
Sorry for my absence, life is hectic.

JohnTitor:
Sorry for the delay, too. Could you do a rebase instead of merge and squash commits into one?

JohnTitor:

PR #3109, opened 2023-02-07
linux: add more constants and FUTEX_OP for futex

voskh0d:
Add FUTEX_BITSET_MATCH_ANY, FUTEX_OP_* constant and a const fonction, FUTEX_OP() to replace the macro in C.


JohnTitor (approved):

voskh0d:
Hi, I've removed the 'const' which was causing some issue with the previous commit. Also, I think making this function was anyway incorrect :)

PR #3124, opened 2023-02-19
FreeBSD: add AT_RESOLVE_BENEATH, Linux-compatible clock aliases

valpackett:
- sorry for the spam xD *now* I should be done with that area of the API…
- +mirror one other thing that [was done](https://reviews.freebsd.org/D30988) in the headers to reduce portability annoyance :)


JohnTitor (approved):

JohnTitor:
Thanks, @bors r+

PR #3126, opened 2023-02-22
add p_name field in kinfo_proc struct on OpenBSD

semarie:
pthread_get_name_np() and pthread_set_name_np() are now using a kernel storage  and could be viewed from outside the process.

Reference: https://github.com/openbsd/src/commit/cef5a146e600a27064f0ea2aa25fc5f8663cb9b7


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #3130, opened 2023-02-28
Remove duplicated atof function declaration

flba-eb:
This adapts to the changes done with PR #3036, removing a function declaration which would exist twice otherwise.

cc: @gh-tr, @samkearney


JohnTitor (approved):

JohnTitor:
Good catch, thanks!
@bors r+

PR #3133, opened 2023-03-03
linux/musl: add copy_file_range syscall wrapper

slp:
musl supports copy_file_range since v1.1.24, so expose the corresponding wrapper here too.


JohnTitor:
Thanks!
@bors r+

JohnTitor:
```
terminating on signal 15 from pid 2416 (timeout)
```

@bors retry

PR #1635, opened 2019-12-31
Add TCP FastOpen support for macOS

zonyitoo:
fix #1632

Issue #1632, opened 2019-12-30
Add TCP FastOpen support for macOS


zonyitoo:
## The `connectx` system call

This function is commonly used for inititiating a TCP Fast Open connection.

[Available since iOS 9 and OS X 10.11](https://en.wikipedia.org/wiki/TCP_Fast_Open)

```c
// Original definition in /usr/include/sys

__API_AVAILABLE(macosx(10.11), ios(9.0), tvos(9.0), watchos(2.0))
int connectx(int, const sa_endpoints_t *, sae_associd_t, unsigned int,
    const struct iovec *, unsigned int, size_t *, sae_connid_t *);

__API_AVAILABLE(macosx(10.11), ios(9.0), tvos(9.0), watchos(2.0))
int disconnectx(int, sae_associd_t, sae_connid_t);
```

Definition:

```c
int connectx(int socket,
             const sa_endpoints_t *endpoints,
             sae_associd_t associd /* Reserved, always be SAE_ASSOCID_ANY */,
             unsigned int flags,
             const struct iovec *iov,
             unsigned int iovcnt,
             size_t *len,
             sae_connid_t *connid /* Reserved, always be NULL */);
```

Example usage:

```c
struct sockaddr_in sa;
memcpy(&sa, /* TARGET ADDR */, sizeof(struct sockaddr_in));
sa.sin_len = sizeof(struct sockaddr_in);
sa_endpoints_t endpoints;
memset((char *)&endpoints, 0, sizeof(endpoints));
endpoints.sae_dstaddr    = (struct sockaddr *)&sa;
endpoints.sae_dstaddrlen = /* TARGET ADDR LEN */;

s = connectx(sockfd,
             &endpoints,
             SAE_ASSOCID_ANY,
             CONNECT_DATA_IDEMPOTENT | CONNECT_RESUME_ON_READ_WRITE,
             NULL,
             0,
             NULL,
             NULL);
```

### Variable definitions

- `SAE_ASSOCID_ANY`
    ```c
    // Defined in /usr/include/sys/socket.h (MacOSX.sdk)
    #define SAE_ASSOCID_ANY 0
    ```

- `CONNECT_DATA_IDEMPOTENT` and `CONNECT_RESUME_ON_READ_WRITE`
    ```c
    // Defined in /usr/include/sys/socket.h (MacOSX.sdk)
    #define CONNECT_RESUME_ON_READ_WRITE    0x1 /* resume connect() on read/write */
[...]

PR #3135, opened 2023-03-03
macos: add TCP_FASTOPEN_FORCE_ENABLE

database64128:
This is used to disable the backoff mechanism, otherwise TFO is basically unusable.

Updates #1632 and #1635

## Definition

https://github.com/apple/darwin-xnu/blob/2ff845c2e033bd0ff64b5b6aa6063a1f8f65aa32/bsd/netinet/tcp.h#L310

## Related Commits/PRs

- https://github.com/zonyitoo/tokio-tfo/pull/5
- https://github.com/database64128/tfo-go/commit/c980f6b18b76910d79908eba489e5f16a98278a9


database64128:
Did I miss anything? Not sure why the test failed.

JohnTitor:
The source you linked seems old, there's no update since 2021. For example, https://github.com/apple-oss-distributions/xnu/blob/5c2921b07a2480ab43ec66f5b9e41cb872bc554f/bsd/netinet/tcp_private.h#L141 shows it's now a private type. But it may be inaccurate as I cannot make sure it's official. Could you check the source directly? I don't have macOS env.

database64128:
> Could you check the source directly? I don't have macOS env.

I also don't have macOS env. This only came up because I was helping test an iOS app that uses my tfo-go module and it does not reliably utilize TFO. I was able to make TFO consistently work on my iOS 15 & 16 devices by setting `TCP_FASTOPEN_FORCE_ENABLE` to 1. Tests in my tfo-go repo suggest that this also works on macOS.

@zonyitoo Can you help with this?

@JohnTitor If it's indeed a private type, does that mean it's no longer eligible for inclusion in this project?

zonyitoo:
I could verify that `setsockopt` with `0x218` returns `0`.

```rust
setsockopt(0x6, 0x6, 0x218)		 = 0 0
```

Client will always try to send TFO cookie even if the remote endpoint doesn't support TFO (`python -m SimpleHTTPServer 80`):
[client_tfo.pcapng.zip](https://github.com/rust-lang/libc/files/10887910/client_tfo.pcapng.zip)

database64128:
@zonyitoo Thanks for the pcap file. Can you check the header file on your system to see if it's indeed part of the private API? I think that's also what they meant in their comment:

> Could you check the source directly?

zonyitoo:
Source Path:

```
/Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/System/Library/Frameworks/Kernel.framework/Versions/A/Headers/netinet/tcp_private.h
```

```c
#define MPTCP_ALTERNATE_PORT            0x216
#define MPTCP_FORCE_ENABLE              0x217
#define TCP_FASTOPEN_FORCE_ENABLE       0x218
#define MPTCP_EXPECTED_PROGRESS_TARGET  0x219
#define MPTCP_FORCE_VERSION             0x21a
```

Well, yes, it is part of the private API. Because it doesn't included in:

```
$ ls /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/netinet/
bootp.h      icmp_var.h   igmp.h       in.h         in_systm.h   ip.h         ip_icmp.h    tcp.h        tcp_seq.h    tcp_var.h    udp.h
icmp6.h      if_ether.h   igmp_var.h   in_pcb.h     in_var.h     ip6.h        ip_var.h     tcp_fsm.h    tcp_timer.h  tcpip.h      udp_var.h
```

So it is impossible to use it directly from C.

JohnTitor:
There's no solid policy about the private items, but if it's subject to change in the future or needs additional setup, this crate's stability policy doesn't work well, I think.

karelbilek:
Can you please share some details about this? Only references to "brutal backoff mechanisms" and TCP_FASTOPEN_FORCE_ENABLE is basically this thread, and the other linked threads.

(My question is not related to rust at all; I just saw this thread when doing this PR for dnsdist https://github.com/PowerDNS/pdns/pull/16071 )

PR #3136, opened 2023-03-03
linux: move DCCP_ constants from linux/gnu to linux

### LINKED ISSUE #3132: dccp exports are tied to glibc on linux
the various DCCP_ exports such as `DCCP_SOCKOPT_SERVICE` are tied to `linux/gnu` (or android). that means they are not present on linux/musl.

however, these defines don't come from the libc. for instance:

```c
#include <sys/socket.h>
#include <stdio.h>

int main(void) {
  printf("%d\n", DCCP_SOCKOPT_SERVICE);
}
```

fails to compile on either glibc or musl. this is because they actually come from `<linux/dccp.h>`.

since they come from the kernel headers, they should also be present in linux/musl/mod.rs (or wherever they are meant to be defined here.)

(mostly an upstream version of a downstream bug report: https://github.com/rust-lang/socket2/issues/405)

nekopsykose:
closes #3132

these constants come from linux headers, so they should be exposed for "linux", not just glibc.

this change exposes them for linux/musl and linux/uclibc.

of note, android contains these same constants, but moving it to linux-like would also expose them on emscripten, which does not have it.

(not sure if things are placed in the correct place, but this looks like an okayish start)


nekopsykose:
not sure if the failing freebsd-14 ci is related to any of this

zonyitoo:
@JohnTitor Could you rerun the CI? I don't think there is anything related to freebsd in this PR.

JohnTitor (approved):

JohnTitor:
Re-running doesn't help it here as the failure is _valid_, it has been fixed on master already so we just need
@bors r+

PR #3137, opened 2023-03-04
linux/musl/s390x: change f_* constants to uint from ulong

nekopsykose:
musl defines these as `unsigned`, not `unsigned long`
https://github.com/bminor/musl/blob/7d756e1c04de6eb3f2b3d3e1141a218bb329fcfb/arch/s390x/bits/statfs.h#L2

mostly relevant to also fixing https://github.com/nix-rust/nix/pull/1835  

that said, i don't know if this is a huge breaking change or not, only that the current one isn't correct afaict

### DIFF
--- a/src/unix/linux_like/linux/musl/b64/s390x.rs
+++ b/src/unix/linux_like/linux/musl/b64/s390x.rs
@@ -60,33 +60,33 @@ s! {
     }
 
     pub struct statfs {
-        pub f_type: ::c_ulong,
-        pub f_bsize: ::c_ulong,
+        pub f_type: ::c_uint,
+        pub f_bsize: ::c_uint,
         pub f_blocks: ::fsblkcnt_t,
         pub f_bfree: ::fsblkcnt_t,
         pub f_bavail: ::fsblkcnt_t,
         pub f_files: ::fsfilcnt_t,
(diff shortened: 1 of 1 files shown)


Amanieu:
This should also fix the definition of the `*_SUPER_MAGIC` constants that are used with statfs.

nekopsykose:
good point, i'll have a look

nekopsykose:
what would the correct definition be? the supermagic constants are normally kernel header origined, and so musl doesn't (re)export anything for them that i can see.

but perhaps you mean just changing every c_uint to c_int in `src/unix/linux_like/mod.rs` under the s390x block? i can do that if you confirm (i'm not 100% sure)

nekopsykose:
ah, or maybe you meant it's already fixed, since now they are both uint?

Amanieu:
Those constants should have the same type as the `f_type` field on `statfs`.

nekopsykose:
ok, that makes sense. that means these existing changes makes them align correctly, since they are c_uint presently.

thank you!

Mark-Simulacrum:
Bors appears to be confused about the commits for this PR, going to try closing and reopening (please re-approve @Amanieu if needed).

PR #3138, opened 2023-03-04
Add support for OpenHarmony

Amanieu:
This PR adds support for [OpenHarmony](https://gitee.com/openharmony/docs/) targets:
- `aarch64-linux-ohos`
- `arm-linux-ohos`

Compiler team MCP: https://github.com/rust-lang/compiler-team/issues/568

OpenHarmony uses a fork of musl with minor modifications, so most of the code is shared with other musl targets. Additionally, although OpenHarmony uses musl 1.2, it is still ABI-compatible with musl 1.1 so the existing bindings should continue to work until https://github.com/rust-lang/libc/pull/3068 is merged.

A PR to add the targets to rustc will follow after this is merged.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -3081,7 +3081,7 @@ fn test_linux(target: &str) {
 
     // target_env
     let gnu = target.contains("gnu");
-    let musl = target.contains("musl");
+    let musl = target.contains("musl") || target.contains("ohos");
     let uclibc = target.contains("uclibc");
 
     match (gnu, musl, uclibc) {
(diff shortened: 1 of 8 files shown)


notgull:
Out of curiosity, does OpenHarmony provide a new polling system, or does it just use Epoll like Linux does?

Amanieu:
These OpenHarmony targets run on top of the Linux kernel and therefore support the same standard mechanisms that Linux provides. However OpenHarmony also supports running on embedded devices & microcontrollers with a custom kernel (LiteOS), but these targets are not currently supported in this PR.

JohnTitor (approved):

JohnTitor:
Thanks!
@bors r+

PR #3139, opened 2023-03-05
removing freebsd 11 EOL proposal.

devnexen:


JohnTitor:
Note that there're some overrides with https://github.com/rust-lang/libc/pull/2406.

workingjubilee:
- FCP: https://github.com/rust-lang/rust/issues/89058
- Rust toolchain update: https://github.com/rust-lang/rust/pull/97944

@devnexen Please resolve the merge conflict.

build.rs
```diff
@@ -56,18 +56,13 @@ fn main() {
...
     // On CI, we detect the actual FreeBSD version and match its ABI exactly,
     // running tests to ensure that the ABI is correct.
     match which_freebsd() {
-        Some(10) if libc_ci || rustc_dep_of_std => set_cfg("freebsd10"),
-        Some(11) if libc_ci => set_cfg("freebsd11"),
-        Some(12) if libc_ci => set_cfg("freebsd12"),
-        Some(13) if libc_ci => set_cfg("freebsd13"),
-        Some(14) if libc_ci => set_cfg("freebsd14"),
-        Some(_) | None => set_cfg("freebsd11"),
+        Some(12) if libc_ci => println!("cargo:rustc-cfg=freebsd12"),
```
asomers, line 62: This looks like a merge conflict. You should use `set_cfg` instead of `println!`

src/unix/bsd/freebsdlike/freebsd/freebsd11/mod.rs
```diff
@@ -1,488 +0,0 @@
-// APIs that were changed after FreeBSD 11
-
-// The type of `nlink_t` changed from `u16` to `u64` in FreeBSD 12:
-pub type nlink_t = u16;
```
asomers, line 4: All of the stuff that was in this file should now be moved into freebsdlike/freebsd/mod.rs  For example, move all three `nlink_t` definitions from freebsd{12,13,14}.rs into that file.

asomers (changes requested):

workingjubilee:
I [started a conversation on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/351149-t-libs.2Fcrates/topic/FreeBSD.20and.20libc) to see how people feel about this.

build.rs
```diff
@@ -74,8 +72,6 @@ fn main() {
 
     match which_freebsd {
         x if x < 10 => panic!("FreeBSD older than 10 is not supported"),
```
tgross35, line 72: Could you update this comment?

src/unix/bsd/freebsdlike/mod.rs
```diff
@@ -1572,26 +1572,10 @@ extern "C" {
     pub fn getutxid(ut: *const utmpx) -> *mut utmpx;
     pub fn getutxline(ut: *const utmpx) -> *mut utmpx;
     pub fn initgroups(name: *const c_char, basegid: crate::gid_t) -> c_int;
-    #[cfg_attr(
-        all(target_os = "freebsd", any(freebsd11, freebsd10)),
-        link_name = "kevent@FBSD_1.0"
-    )]
-    pub fn kevent(
```
asomers, line 1571: Did you really mean to remove this function?  I think you should just remove the `link_name` instead.  The function is definitely needed.

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -53,6 +53,11 @@ pub type sctp_assoc_t = u32;
 
 pub type eventfd_t = u64;
 
+pub type nlink_t = u64;
+pub type dev_t = u64;
+pub type ino_t = c_ulong;
```
asomers: This should be u64.  Otherwise `stat` is broken on x86.
```suggestion
pub type ino_t = u64;
```

asomers (changes requested):
These are the only two problems I've found in my testing so far.  But that begs the question: how did you test these changes?

asomers:
But what did you do within that VM?  What projects did you test?  How did you patch them to use libc main branch?

devnexen:
Usually, I just do ```cargo test --manifest-path libc-test/Cargo.toml```.

devnexen:
> > Usually, I just do `cargo test --manifest-path libc-test/Cargo.toml`. I did not try to use in any project.
> 
> For a change like this, that really isn't sufficient. You need to run the test suites of a selection of other crates, patched to use this branch of libc.

Ok I ll try

src/unix/bsd/mod.rs
```diff
@@ -644,15 +644,7 @@ extern "C" {
     pub fn strerror_r(errnum: c_int, buf: *mut c_char, buflen: size_t) -> c_int;
     pub fn abs(i: c_int) -> c_int;
     pub fn labs(i: c_long) -> c_long;
-    #[cfg_attr(
-        all(target_os = "freebsd", any(freebsd12, freebsd11, freebsd10)),
-        link_name = "rand@FBSD_1.0"
-    )]
     pub fn rand() -> c_int;
-    #[cfg_attr(
-        all(target_os = "freebsd", any(freebsd12, freebsd11, freebsd10)),
```
asomers, line 653: Why are you removing this link name?  FreeBSD 12 still needs it.

src/unix/bsd/mod.rs
```diff
@@ -644,15 +644,7 @@ extern "C" {
     pub fn strerror_r(errnum: c_int, buf: *mut c_char, buflen: size_t) -> c_int;
     pub fn abs(i: c_int) -> c_int;
     pub fn labs(i: c_long) -> c_long;
-    #[cfg_attr(
-        all(target_os = "freebsd", any(freebsd12, freebsd11, freebsd10)),
```
asomers, line 648: Here, too.  Why remove the link name?

src/unix/bsd/mod.rs
```diff
@@ -864,10 +848,6 @@ extern "C" {
         all(target_os = "macos", target_arch = "x86"),
         link_name = "wait4$UNIX2003"
     )]
-    #[cfg_attr(
-        all(target_os = "freebsd", any(freebsd12, freebsd11, freebsd10)),
-        link_name = "wait4@FBSD_1.0"
```
asomers, line 869: Like the others, I think that removing this link_name was unintended.  However, it is coincidentally the right thing to do.  It never should've been there in the first place.  `wait4` hasn't ever been modified since FreeBSD first adopted ELF symbol versions.

asomers (changes requested):
@tgross35 the breakage shown in the Nix crate came from switching libc-0.2 to libc-1.0.  But it's important for the submitter to check that nix doesn't break any _worse_ on libc-1.0 with this PR than without it.
And I agree that there _shouldn't_ be any breakage as a result of removing support for the FreeBSD 10 and 11 ABIs from libc.  However, there may be as a result of mistakes in the PR.  The fact that the submitter didn't catch the accidental removal of kevent or changing the type of ino_t suggests that he didn't test it very closely.

PR #3142, opened 2023-03-07
add S_IRWX* constants to wasi

connor4312:
It seems like all other platforms have `/S_IRWX[OGU]/`, and the `libc` crate looks for them, but they were missing in WASI.


JohnTitor (approved):

JohnTitor:
Thanks a lot! @bors r+

PR #3144, opened 2023-03-09
Add more declarations for AIX

ecnelises:
Add more constants and function declarations to make it more compatible to code written for other existing unix-like systems.

All added constants are verified on 64-bit AIX system. All added function are verified as linkable symbols.


src/unix/aix/mod.rs
```diff
@@ -6,23 +6,27 @@ pub type blkcnt_t = ::c_long;
 pub type clock_t = ::c_int;
 pub type daddr_t = ::c_long;
 pub type dev_t = ::c_ulong;
+pub type fpos64_t = ::c_longlong;
 pub type fsblkcnt_t = ::c_ulong;
 pub type fsfilcnt_t = ::c_ulong;
+pub type idtype_t = ::c_int;
 pub type ino_t = ::c_ulong;
-pub type key_t = ::c_long;
+pub type key_t = ::c_int;
```
JohnTitor, line 14: Type changing would be a breaking change, is this wrong at all or was it changed at some point? If latter, dropping it for now would be a safe way.

ecnelises: Sorry, previous definition was wrong and should be 4 bytes on both 32 and 64 bits.

JohnTitor (approved):

JohnTitor:
Thank you! @bors r+

PR #3152, opened 2023-03-17
mac time.h functions

shua:
I used these in https://github.com/shua/graf because I'm used to them, and don't really want to pull in time or chrono just for some stuff already present in libc.h . Compilation worked fine on my linux target, but failed on mac osx because these weren't defined in rust-lang/libc (but these are all present in 13.1 SDK).

### DIFF
--- a/libc-test/semver/apple.txt
+++ b/libc-test/semver/apple.txt
@@ -1819,6 +1819,8 @@ arc4random
 arc4random_buf
 arc4random_uniform
 arphdr
+asctime
+asctime_r
 attrgroup_t
 attribute_set_t
 attrlist
@@ -1845,6 +1847,8 @@ copyfile
 copyfile_flags_t
 cpu_subtype_t
 cpu_type_t
+ctime
+ctime_r
 ctl_info
 difftime
 dirfd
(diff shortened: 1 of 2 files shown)


shua:
It's currently failing because the new functions are unused, but I don't know how to fix this. I'm not sure why these functions are unused but the others are considered to be used. I tried `git grep` for some of the other functions defined in `src/unix/bsd/apple/mod.rs` but never found where they're being used, only where they're defined.

JohnTitor:
Because they're already defined and available on macOS via reexports: https://github.com/rust-lang/libc/blob/bac14427831aaff9b79dcdd125c03109414215c5/src/unix/mod.rs#L1269

Have you tried importing them actually?

shua:
ah, thanks, I misunderstood the errors. I updated the PR to remove the time.h functions and struct that were already defined in `src/unix/mod.rs` and `src/unix/linux_like/mod.rs`..

shua:
@JohnTitor are you the right person to ask for a review on this?

JohnTitor:
Yeah, could you squash commits into one?

shua:
The test failure seems unrelated to my changes. The test that failed was targeting freebsd 14, but my changes only touched two files, both specific to macos.

JohnTitor:
It's fixed on master :)
@bors r+

JohnTitor (approved):

PR #3154, opened 2023-03-17
_lwp api solarish addition.

devnexen:


pfmooney:
I'd have to do more looking, but I'm pretty sure that those are private symbols that external consumers should not be building dependencies upon.

jclulow:
I am also inclined not to put these in here.  Where were you hoping to consume these symbols?

There are better alternatives for some of them; e.g., `thr_kill()` is better than `_lwp_kill()`.

devnexen:
> I'd have to do more looking, but I'm pretty sure that those are private symbols that external consumers should not be building dependencies upon.

Fair enough, it was not obvious from their respective man pages.

PR #3155, opened 2023-03-19
definitions for linux hardware timestamping

folkertdev:
Definitions can be found here https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/net_tstamp.h#L76

These definitions are relevant for (PTP and NTP) hardware timestamping


folkertdev:
I'm not sure what this CI error means

```
  cargo:warning=/checkout/target/i686-unknown-linux-gnu/debug/build/libc-test-1d150dc772ee3115/out/main.c:35676:80: error: 'HWTSTAMP_FLAG_BONDED_PHC_INDEX' undeclared here (not in a function)
  cargo:warning=35676 |             static const int __test_const_HWTSTAMP_FLAG_BONDED_PHC_INDEX_val = HWTSTAMP_FLAG_BONDED_PHC_INDEX;
  cargo:warning=      |                                                                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  cargo:warning=/checkout/target/i686-unknown-linux-gnu/debug/build/libc-test-1d150dc772ee3115/out/main.c:35682:68: error: 'HWTSTAMP_FLAG_LAST' undeclared here (not in a function)
  cargo:warning=35682 |             static const int __test_const_HWTSTAMP_FLAG_LAST_val = HWTSTAMP_FLAG_LAST;
  cargo:warning=      |                                                                    ^~~~~~~~~~~~~~~~~~
  cargo:warning=/checkout/target/i686-unknown-linux-gnu/debug/build/libc-test-1d150dc772ee3115/out/main.c:35688:68: error: 'HWTSTAMP_FLAG_MASK' undeclared here (not in a function)
  cargo:warning=35688 |             static const int __test_const_HWTSTAMP_FLAG_MASK_val = HWTSTAMP_FLAG_MASK;
  cargo:warning=      |                                                                    ^~~~~~~~~~~~~~~~~~
  cargo:warning=cc1: note: unrecognized command-line option '-Wno-unknown-warning-option' may have been intended to silence earlier diagnostics
```

maybe the system libc on CI does not (yet?) have these symbols? They are not important to me so I'd be OK with removing them (just thought I might as well add them for completeness)

JohnTitor:
> maybe the system libc on CI does not (yet?) have these symbols?

Correct.

JohnTitor (approved):

JohnTitor:
Thank you, @bors r+

folkertdev:
I'm not sure what went wrong here. I assume the action needs to just be re-run?

folkertdev:
I added two extra constants that are needed (`SIOCSHWTSTAMP` and `SIOCGHWTSTAMP`) and rebased. Maybe that makes bors happier?

though CI seems to be broken at the moment.

JohnTitor:
The FreeBSD failure is unrelated and it's already fixed on master.
But https://github.com/rust-lang/libc/pull/3155#issuecomment-1481006102 is legit. That's because (probably) musl toolchain on CI is old, could you skip tests on that env?

folkertdev:
ok, rebased again to fix that freebsd problem

> could you skip tests on that env

how would I do this? do you mean a `cfg` for the undefined constants?

JohnTitor:
https://github.com/rust-lang/libc/blob/master/libc-test/build.rs is a place to tweak, like:
https://github.com/rust-lang/libc/blob/68e06adafa8f793b094a23b210d39b27c813600c/libc-test/build.rs#L1739-L1754

folkertdev:
allright, I added logic to ignore the constant for `i686` and `mips`. that should make the tests that failed before pass, but there might be other (cancelled) ci runs that also should skip this constant. Is there a better way to figure that out than trial and error?

JohnTitor:
There's a trade-off, we could disable the fail-fast but we may want to find failures fast.
@bors r+

JohnTitor:
(Another way would be running the same command as CI on your local.)

libc-test/build.rs
```diff
@@ -3724,6 +3724,8 @@ fn test_linux(target: &str) {
                 => true,
             "SCTP_FUTURE_ASSOC" | "SCTP_CURRENT_ASSOC" | "SCTP_ALL_ASSOC" | "SCTP_PEER_ADDR_THLDS_V2" => true, // linux 5.5+
 
+            "HWTSTAMP_TX_ONESTEP_P2P" if i686 || mips || aarch64_musl => true,
```
JohnTitor, line 3727: I think the below should work like others:

```suggestion
            // FIXME: Requires more recent kernel headers
            "HWTSTAMP_TX_ONESTEP_P2P" if mips || musl => true,
```

folkertdev: yes, fixed. A couple more linux flavors to get through

JohnTitor: The condition doesn't seem to be changed. It should `mips || musl` as we use older kernels for these targets.

folkertdev: though maybe musl is more relevant than aarch here?

JohnTitor: We use the same kernel version for _all_ the musl targets across CI: https://github.com/rust-lang/libc/blob/c64bfc5fb36414e02ad7977b32df765876632383/ci/install-musl.sh#L82
So, CPU arch doesn't involve here, we just need to ignore `musl` on the if cond (the same applies to sparc64).

folkertdev: so `i686 || mips || musl`, or only `musl` even?

JohnTitor: Ah, I mean the condition should be `sparc64 || musl`.

JohnTitor: For sparc64, see:
https://github.com/rust-lang/libc/blob/c64bfc5fb36414e02ad7977b32df765876632383/ci/docker/sparc64-unknown-linux-gnu/Dockerfile#L1-L2

JohnTitor: i686 itself isn't related to the failure, it fails due to the musl toolchain.

folkertdev: makes sense, I've updated it

JohnTitor:
Let's try again, @bors r+

folkertdev:
we're getting closer. I'm not sure what this is trying to do though:

```
RUNNING ALL TESTS
rust[8] = 0 != 8 (C): Rust "hwtstamp_config" -> C
rust[9] = 0 != 9 (C): Rust "hwtstamp_config" -> C
rust[10] = 0 != 10 (C): Rust "hwtstamp_config" -> C
rust[11] = 0 != 11 (C): Rust "hwtstamp_config" -> C
thread 'main' panicked at 'some tests failed', /checkout/target/sparc64-unknown-linux-gnu/debug/build/libc-test-298fd54f1fef40cd/out/main.rs:12:21

Caused by:
  process didn't exit successfully: `/test-runner-linux sparc64 /checkout/target/sparc64-unknown-linux-gnu/debug/deps/main-973f2bcc439c5f8d` (exit status: 1)
```

should we skip the `hwtstamp_config` struct on this target?

folkertdev:
rather looks like this is an ABI roundtrip test. That this fails is extremely surprising to me, given that the struct is just three `int`s, so nothing tricky is going on. Should we just ignore the roundtrip for `sparc64`?

JohnTitor:
I guess https://github.com/rust-lang/libc/issues/1558 is related. Yeah, skipping would be fine 👍

folkertdev:
allright let's hope that's it

PR #3165, opened 2023-03-24
add memmem GNU extension for haiku

devnexen:


JohnTitor (approved):

nielx:
@devnexen What is this libunix.so you try to link against? It does not seem to be part of Haiku.

devnexen:
ah yes it s part of libgnu I can see in the last haiku snapshot, seems it s already taking care of.

JohnTitor:
@devnexen I'm always happy to see your PRs and your contributions are great, but please make sure everything isn't broken on your end before submitting a PR (IIRC a similar thing happened by your PR in the past). CI doesn't check for all the targets, unfortunately (and I'm not the expert for all the supported targets, sorry).

PR #3172, opened 2023-03-27
FreeBSD: Ignore test for remove const

JohnTitor:
This brings us some problems, let's pin it.


JohnTitor:
`pkg install` doesn't work nicely on a pinned version, weird. Let's just fix the test and investigate later to unblock CI.
@bors r+

PR #3175, opened 2023-03-29
Switch to 64 bit file and time APIs on GNU libc for 32bit systems

snogge:
Use the 64 bit types and APIs included in GNU libc also for 32-bit systems.  These are the types and APIs used when you compile your C code against GNU libc headers with the preprocessor options -D_FILE_OFFSET_BITS=64 -D_TIME_BITS=64.

This is required for 2038 compatible software built for 32-bit systems. 

This PR is not done, I need some guidance with finishing it.  
1.  Does it need some sort of configuration options to turn this on and off?
2. There is a conflict for some of the stat functions where two signatures link against the same symbol.  For C this is not a problem, but we probably need to make `struct stat` and `struct stat64` be different names for the same struct.  I don't know how to do that.
3. It's possible that some types should be moved/split/joined in the file hierarchy.  I'd appreciate some feedback on that, but I'm waiting on the answer to question 1 before moving forward. 

I've done one commit for each modified type so it is easier to review the changes in isolation.


snogge:
Oops, left some debug/test commits in there.  Removed now.

snogge:
Hi @JohnTitor , do you have any feedback on this change? 
Is there anything I can do to facilitate the review?
I'd appreciate any hints on solving question 2:

> 2. There is a conflict for some of the stat functions where two signatures link against the same symbol. For C this is not a problem, but we probably need to make struct stat and struct stat64 be different names for the same struct. I don't know how to do that.

I'm not a Rust developer and not sure how I can create that type alias.

JohnTitor:
Thanks for the PR! But I think this is a huge breaking change and we have to discuss it before reviewing the changes. Could you open an issue instead?

snogge:
Sure, I'll do that.

snogge:
I'm working on updating this PR to handle problem 2 mentioned above and to pass tests on more platforms.  Expect an update later this week.

snogge:
Sorry for the many pushes, that was meant only for my fork.  Anyway, now the PR should handle the duplicate functions mentioned in problem 2, and also pass tests on more platforms.

snogge:
Just rebased on main. Passes all tests in main.yml and bors.yml workflows. 
Still not very clean though, feedback on code style would be welcome.

I considered adding a cargo feature to control this, but as that would be braking the ABI I guess I shouldn't?

src/unix/mod.rs
```diff
@@ -62,17 +62,26 @@ s! {
...
+        pub tv_usec: i64
     }
 
     // linux x32 compatibility
     // See https://sourceware.org/bugzilla/show_bug.cgi?id=16437
     pub struct timespec {
         pub tv_sec: time_t,
+        #[cfg(all(gnu_time64_abi, any(target_arch = "powerpc", target_arch = "mips")))]
+        __pad: i32,
         #[cfg(all(target_arch = "x86_64", target_pointer_width = "32"))]
```
plugwash, line 79: Would it make more sense to use target_endian here, rather than hardcoding specific architectures?

snogge:
I just removed the CI-changes I pushed by mistake....again.

plugwash:
Hi, I applied a patch series based on this PR to the rust-libc package in Debian. Following Debian's descisions on which architectures to apply the time64 transition to I added a condition to disable the "gnu_time64_abi" setting on i386.

However, it seems that this patch series causes API breaks, even if the "gnu_time64_abi" config is not set. For example https://ci.debian.net/packages/r/rust-fs-at/testing/i386/45768200/#S6

plugwash:
Yeah, it looks like the missing f_flags issue affects at least arm too.

snogge:
I've reconstructed the PR to support an environment variable `RUST_LIBC_TIME_BITS` which can be set to `32` to force 32-bit time on 32-bit platforms.  The default, based on the cargo environment variables is 64-bit time for those platforms. 
I've also modified the CI workflows to test these variants.

src/unix/mod.rs
snogge: Fixed now.

src/unix/linux_like/linux/mod.rs
```diff
@@ -322,7 +322,12 @@ s! {
     }
 
     pub struct input_event {
+        #[cfg(not(gnu_time64_abi))]
         pub time: ::timeval,
+        #[cfg(gnu_time64_abi)]
+        __sec: ::c_ulong,
+        #[cfg(gnu_time64_abi)]
+        __usec: ::c_ulong,
         pub type_: ::__u16,
```
plugwash, line 331: This breaks the evdev crate with no obvious way of fixing it (rust-evdev has other errors too, but they are errors I can deal with)

692s error[E0560]: struct `input_event` has no field named `time`
692s    --> src/lib.rs:265:13
692s     |
692s 265 |             time: self.time.as_raw(),
692s     |             ^^^^ `input_event` does not have this field
692s     |
692s     = note: available fields are: `__sec`, `__usec`

The original C header deals with this using macros, which are defined differently for the non-time64 and time64 cases.
(non-time64)

#define input_event_sec time.tv_sec
#define input_event_usec time.tv_usec

(time64)

#define input_event_sec  __sec
#define input_event_usec __usec

Not sure how this can be best dealt with in rust, the best I can think of would be to add get/set/get_mut methods.

snogge: Thank you for testing.
Does this definition of `input_event` work - it will require changes in using code
```rust
    pub struct input_event {
        #[cfg(any(target_pointer_width = "64", not(gnu_time64_abi)))]
        pub input_event_sec: ::time_t,
        #[cfg(all(target_pointer_width = "32", gnu_time64_abi))]
        pub input_event_sec: ::c_ulong,

        #[cfg(not(gnu_time64_abi))]
        pub input_event_usec: ::suseconds_t,
        #[cfg(gnu_time64_abi)]
        pub input_event_usec: ::c_ulong,

        #[cfg(target_arch = "sparc64")]
        _pad1: u32,

        pub type_: ::__u16,
        pub code: ::__u16,
        pub value: ::__s32,
    }
```

snogge: Thank you for testing.
Does this definition of `struct input_event` work for evdev-rs - it will require code changes.
```rust
    pub struct input_event {
        #[cfg(any(target_pointer_width = "64", not(gnu_time64_abi)))]
        pub input_event_sec: ::time_t,
        #[cfg(all(target_pointer_width = "32", gnu_time64_abi))]
        pub input_event_sec: ::c_ulong,

        #[cfg(any(target_pointer_width = "64", not(gnu_time64_abi)))]
        pub input_event_usec: ::suseconds_t,
        #[cfg(all(target_pointer_width = "32", gnu_time64_abi))]
        pub input_event_usec: ::c_ulong,

        #[cfg(target_arch = "sparc64")]
        _pad1: ::c_int,

        pub type_: ::__u16,
        pub code: ::__u16,
        pub value: ::__s32,
    }
```

plugwash: Structurally this looks like a sensible compromise and one to which downstream code could be adapted fairly easily.

I still think the definition is probablly wrong on 32-bit linux systems with 64-bit time and C libraries other than glibc.

snogge: You're probably right, I'll have to think about this some more. 
But I'll update with this struct definition for now.

plugwash:
After patching the filetime crate to build against the patched libc crate, i'm seeing a test failure. I guess there is some sort of problem with the file time APIs.

git clone https://github.com/plugwash/filetime
cd filetime
cargo test

    ---- tests::set_file_times_test stdout ----
    thread 'tests::set_file_times_test' panicked at src/lib.rs:449:9:
    assertion `left == right` failed: modification time should be updated
      left: FileTime { seconds: 1073741822, nanos: 0 }
     right: FileTime { seconds: 20000, nanos: 0 }

snogge:
You need to use the `utimensat_time64` syscall in https://github.com/plugwash/filetime/blob/main/src/unix/linux.rs#L44
I haven't (yet) figured out why the  call to `set_file_times` on https://github.com/plugwash/filetime/blob/main/src/lib.rs#L438 works, but something about the use of only the `mtime` argument causes the `mtime` to be set to the value of `UTIME_OMIT`.

plugwash:
Thanks a lot for the advice on the filetime issue.

I'm putting together a list of known fallout from the time64 update for rust in Debian along with the patches we are carrying to deal with it, it's possible I've missed some stuff. Some of these patches also make debian-specific assumptions, they should be regarded as pointers to stuff that needs fixing and potential, not nessacerally patches that should be adopted as-is.

bindgen needs to pass time64 defines libclang:
* bindgen - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/bindgen/debian/patches/pass-time64-flags-to-libclang.patch

timespec now contains padding:
* alsa - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/coreutils/debian/patches/use-mem-zeroed.diff
* coreutils - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/coreutils/debian/patches/use-mem-zeroed.diff
* laurel - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/laurel/debian/patches/use-timespec-new.patch
* pipewire - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/pipewire/debian/patches/use-mem-zeroed.patch
* pleaser - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/pleaser/debian/patches/use-mem-zeroed.patch
* serial-unix - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/serial-unix/debian/patches/use-mem-zeroed.patch
*  vmm-sys-util - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/vmm-sys-util/debian/patches/use-mem-zeroed.patch (note: upstream don't support 32-bit and we may well drop 32-bit support for this package in Debian too soon, but for now we are shipping this patch).

glibc has inconsitent definitions of suseconds_t and struct timeval

input_event defition change:
* evdev-rs - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/evdev-rs/debian/patches/time64.patch

to be continued...........

snogge:
> I'm putting together a list of known fallout from the time64 update for rust in Debian along with the patches we are carrying to deal with it, it's possible I've missed some stuff. Some of these patches also make debian-specific assumptions, they should be regarded as pointers to stuff that needs fixing and potential fixes, not nessacerally patches that should be adopted as-is.

This is great!

> 
> timespec now contains padding:

I don't see what we can do about this. 

<snip>

> glibc has inconsitent definitions of suseconds_t and struct timeval:

I have no idea why glibc did not choose to redefine `suseconds_t` to a 64 bit type instead of using `__suseconds64_t` => `int64_t`.

>     * filetime - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/filetime/debian/patches/suseconds_t-workaround.patch
> 
>     * laurel - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/laurel/debian/patches/use-timespec-new.patch

Not going via `libc::timespec` just looks better to me.

>     * libpulse-binding - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/libpulse-binding/debian/patches/avoid-suseconds_t.patch

https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/libpulse-binding/debian/patches/avoid-suseconds_t.patch#L34
Using `time_t` here is easy but maybe a bit misleading.  `time_t` is never involved in defining this member type in glibc.

>     * net2 - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/net2/debian/patches/suseconds_t-workaround.patch
> 
>     * nix - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/nix/debian/patches/time_t_workarounds.diff

https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/nix/debian/patches/time_t_workarounds.diff#L66
Shouldn't that be `i64`?  `suseconds_t` is a signed integer type. 

>     * socket2 - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/socket2/debian/patches/suseconds_t-workaround.patch
> 
>     * unix-socket - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/unix-socket/debian/patches/suseconds_t-workaround.patch
> 
>     * vsock - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/vsock/debian/patches/suseconds_t-workaround.patch
> 
> 
> input_event defition change:
> 
>     * evdev-rs - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/evdev-rs/debian/patches/time64.patch

I've pushed the new input_event struct that looks like (except for conditionals and stuff)
```rust
    pub struct input_event {
        pub input_event_sec: ::time_t,
        pub input_event_sec: ::c_ulong,
    };
```
I think it is correct for all archs, I cannot figure out any where it shouldn't be.

> 
> explicit syscalls:
> 
>     * filetime - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/filetime/debian/patches/use-correct-syscall-on-time64.patch
>       to be continued...........

Just use the `libc::utimensat` function instead?  It will call the correct glibc wrapper around the correct syscall.
I think filetime calls a some other libc functions anyway.

> 
> bindgen needs to pass time64 defines libclang:
> 
>     * bindgen - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/bindgen/debian/patches/pass-time64-flags-to-libclang.patch
> 
> 
> issues related to bindgen:
> 
>     * selinux-sys - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/selinux-sys/debian/patches/fix-bindgen.diff

snogge:
> > input_event defition change:
> > ```
> > * evdev-rs - https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/evdev-rs/debian/patches/time64.patch
> > ```
> 
> I've pushed the new input_event struct that looks like (except for conditionals and stuff)
> 
> ```rust
>     pub struct input_event {
>         pub input_event_sec: ::time_t,
>         pub input_event_sec: ::c_ulong,
>     };
> ```
> 
> I think it is correct for all archs, I cannot figure out any where it shouldn't be.

@plugwash, I reread your comment about this, and I now I can figure out where it wouldn't work. :)

snogge:
@JohnTitor @tgross35 @joshtriplett 
I apologize for taking this shotgun-like approach for attention.
This PR is now ~1.5 years old, and I'm still not sure whether there is any plan to include it in the 1.0 release and whether my approach is acceptable. 
I realize it is a large and controversial change, but I'd _really_ like some feedback on the PRs viability and the choices made.

snogge:
@tgross35  Any updates?  It's been almost two months now.

tgross35:
Timing, I was actually just looking at this :) I think we just need to start merging small parts of this to unblock things. Would it be possible to split the following changes to a new PR?

- The changes to `src/unix/linux_like/linux/mod.rs`
- Adding `gnu_time64_abi` to `ALLOWED_CFGS` in `build.rs`

If we can get that merged as a basis then the rest (testing, platform support, defaults) can fall into place as we figure it out.

tgross35:
Actually, a more general request than that. Could uapi-related changes be split to a PR that we can merge first?

I mentioned this in the musl PR at https://github.com/rust-lang/libc/pull/3791#discussion_r1849480058 but we should have a config `linux_time_bits64` that corresponds to `__USE_TIME_BITS64` used by uapi (example in [`struct input_event`](https://github.com/torvalds/linux/blob/bf9aa14fc523d2763fc9a10672a709224e8fcaf4/include/uapi/linux/input.h#L28-L47), my [suggested rust field implementation](https://github.com/rust-lang/libc/pull/3791#discussion_r1849493446)). This should be used for anything that isn't glibc-specific.

Doing this first makes the rest of the changes here a lot smaller and unblocks the musl PR as well.

tgross35:
Per the above, I think getting the Linux `cfg` in place is a reasonable first separate first since it affects all libraries and is a smaller portion of this change. After that is in place it will be easier to merge the rest.

Let me know if you need any help here.

@rustbot author

snogge:
I'm working on a new PR, but it's been a while so I need to double check some things.

snogge:
@tgross35 new PR for `linux_time_bits64`: #4148

tgross35:
After https://github.com/rust-lang/libc/pull/4148 merges this can be updated to finish the rest. I think the goal should be similar to #4148: introduce two environment variables `RUST_LIBC_UNSTABLE_GLIBC_FILE_OFFSET_BITS` and `RUST_LIBC_UNSTABLE_GLIBC_TIME_BITS` that correspond to `_FILE_OFFSET_BITS` and `_TIME_BITS` described at https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html. To keep with smaller PRs, one of these can be split off (it probably makes sense to do file first since it stands on its own but time depends on it).

plugwash:
IMO anything that is to be merged upstream needs to be tested and cross/checked against the C definitions both with and without time64 enabled. 32-bit non-time64 systems are not going to go away immediately both because people will not upgrade immediately, and because debian/ubuntu decided to exclude i386 from the time64 transition.

When applying time64 patches derived from this PR in Debian, and adding a patcht to disable time64 on i386, I ran into some issues with the stat structures, some of which were noticed immediately and some of which took more time to notice. I fixed these for i386 in https://salsa.debian.org/rust-team/debcargo-conf/-/blob/8838a57bc76436e0a7f42ff7baadaac0faf3585b/src/libc/debian/patches/time64-108-fix-stat-i386.patch but I suspect there are similar issues for other architectures.

snogge:
> @snogge are there parts of this PR that are still outstanding since #4345 and #4433?

* The default should be switched to 64bit time on master
* I think the `input_event` struct should be changed to use the `input_event_X` macros. We have discussed this before.
* I have a change to improve how the `RUST_LIBC_UNSTABLE_GNU_*_BITS` variables are read in `build.rs` which makes switching the default easier.

I hop to start pushing PRs this week.

tgross35:
@snogge I am going to close this, since I believe everything here has been superseded by a few PRs that have merged (with the exception of https://github.com/rust-lang/libc/pull/4654). Thank you very much for getting the ball rolling here!

PR #3177, opened 2023-03-29
Bump version to 0.2.141

arjentz:
Bump version as requested in https://github.com/rust-lang/socket2/issues/405

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "libc"
-version = "0.2.140"
+version = "0.2.141"
 authors = ["The Rust Project Developers"]
 license = "MIT OR Apache-2.0"
 readme = "README.md"
--- a/libc-test/Cargo.toml
+++ b/libc-test/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "libc-test"
-version = "0.2.140"
+version = "0.2.141"
 authors = ["The Rust Project Developers"]
 license = "MIT OR Apache-2.0"
 build = "build.rs"
@@ -12,7 +12,7 @@ A test crate for the libc crate.
 
 [dependencies.libc]
 path = ".."
-version = "0.2.140"
+version = "0.2.141"
 default-features = false
 
 [build-dependencies]


JohnTitor:
Let me merge some more PRs in the review queue.

jbg:
FYI, upwards of 48,500 crates currently fail to build on all musl targets because libc 0.2.140 is broken on musl. IMHO, it may be worth considering yanking 0.2.140 and/or publishing a quick 0.2.141 even without the extra merged PRs.

arjentz:
Maybe I failed to express the urgency of this PR. I figured it'd be merged and released pretty quickly (and I didn't have the numbers handy). Thanks @jbg!

@JohnTitor It would be great if we could get the release out ASAP

JohnTitor:
It's interesting that you don't have even a single word of appreciation for my maintaining this crate and making releases, whereas you immediately downvote me for asking you to wait a bit to take care of the PR waiting for review. _Sigh_.

jbg:
> > FYI, upwards of 48,500 crates currently fail to build on all musl targets because libc 0.2.140 is broken on musl. IMHO, it may be worth considering yanking 0.2.140 and/or publishing a quick 0.2.141 even without the extra merged PRs.
> 
> @jbg Seems you misunderstand how this crate works, it just exposes consts/fns/structs etc. We've also tested the build for musl on CI and it's green. If you refer to the socket2 issue, its build is broken because it uses an _undeclared_ item on this crate, note that it's not this crate's fault.
> 
> But yeah, let's make a new release now then, @bors r+

Thanks very much for the release, and apologies for the misunderstanding, I mistakently thought a `libc` patch release had made those constants conditional on glibc targets and broken `socket2`, rather than a `socket2` patch release starting to depend on `libc` constants that were unexpectedly only exported on glibc targets.

Is there a way the docs could show when constants are target-specific? The 0.2.140 docs don't indicate in any way that `DCCP_*` is only exported for glibc targets in that version. Maybe rustdoc doesn't provide a convenient way to do that?

JohnTitor:
Yeah, given how we (re)export items, IIRC it's difficult to show items on docs.rs per target.
But we have platform-specific docs on gh-pages: https://rust-lang.github.io/libc/#platform-specific-documentation
Does it work for you?

jbg:
Yes, these are excellent.

I think despite the link from README they might be easily missed, though, given how docs.rs is in many people's muscle memory.

I wonder if it's possible somehow to get rustdoc to produce the little yellow tags (like the `Unix` and `Windows` ones on https://docs.rs/tokio/latest/tokio/signal/index.html) for target-specific re-exports. I think somehow adding a [#[doc(cfg)] attribute](https://doc.rust-lang.org/rustdoc/unstable-features.html#doccfg-recording-what-platforms-or-features-are-required-for-code-to-be-present) to the re-exports would do it, but that would require the docs to be built with nightly rustdoc (which docs.rs does use).

JohnTitor:
`#[doc(cfg(...))]` _might_ be a good solution here, although it should be quite hard as this crate supports a lot of targets...

jbg:
I'll have a play and if it's not too ugly I'll submit a PR for consideration :)

Thomasdezeeuw:
Thanks @JohnTitor!

JohnTitor:
Published on crates.io: https://crates.io/crates/libc/0.2.141

arjentz:
Thanks a lot @JohnTitor

PR #3178, opened 2023-03-30
[DRAFT] MSVC: Add `msvc_crt_link` options

ChrisDenton:
> **Note**
> This only applies when libc is used as a dependency of std.
>
> PR is for testing/discussion purposes only. It should not be merged yet.
>
> See https://github.com/rust-lang/rust/issues/39016 for more background

Currently libc hard codes the MSVC CRT that's used by the compiler. Users can select between two CRTs (dynamic and static) by toggling the `target-feature=crt-static` option. However, this is far from ideal given the number of possible configurations, especially when attempting to co-exist with existing C/C++ builds.

At a minimum, it's also useful to be able to use the debug or release versions of these libraries. The following table shows these possible combinations:

||Release|Debug|
|--|--|--|
|**Dynamic**|`msvcrt`|`msvcrtd`|
|**Static**|`libcmt`|`libcmtd`|

Furthermore, users may want to disable linking the crt entirely so that they can fully control what is linked and how.

To support these different configurations, this PR adds a new cfg, `msvc_crt_link=option`, where `option` is the link configuration option. This can be set in global rustflags (using `--cfg`) so that it's applied consistently. The two possible options are:

* `debug`: Use the debug versions of the CRT libraries
* `nocrt`: disable all CRT linking. This obviously overrides all other options because nothing is linked

### DIFF
--- a/src/windows/mod.rs
+++ b/src/windows/mod.rs
@@ -254,10 +254,97 @@ pub const SIG_GET: ::sighandler_t = 2;
 pub const SIG_SGE: ::sighandler_t = 3;
 pub const SIG_ACK: ::sighandler_t = 4;
 
-// inline comment below appeases style checker
-#[cfg(all(target_env = "msvc", feature = "rustc-dep-of-std"))] // " if "
-#[link(name = "msvcrt", cfg(not(target_feature = "crt-static")))]
-#[link(name = "libcmt", cfg(target_feature = "crt-static"))]
+// MSVC C RunTime (CRT) link options.
+//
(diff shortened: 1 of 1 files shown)


JohnTitor:
Triage: This has been open for more than six months. Is there anything I can help here?

ChrisDenton:
Hmm... I don't think so, I'd hoped this would be part of a wider discussion on solving the issue but that discussion doesn't look like it's going to happen any time soon. I'll remove this for now and maybe revisit if/when I can gather more interest. It is something that needs broader input so I wouldn't want to move forward without that.

PR #3062, opened 2023-01-09
android: add missing syscall constants

lucab:
This adds all syscall constants that are available in NDK `r21d` but were so far missing in this library.
See `semver/android.txt` changes for the complete list.

Refs:
 * https://android.googlesource.com/platform/bionic/+/refs/tags/ndk-r21d/libc/kernel/uapi/asm-generic/unistd.h
 * https://android.googlesource.com/platform/bionic/+/refs/tags/ndk-r21d/libc/include/bits/glibc-syscalls.h

PR #3179, opened 2023-03-31
android: add new syscall constants

sunfishcode:
Similar to #3062, this adds all syscall constants that are available in NDK `r25c` but were so far missing in this library. See `semver/android.txt` changes for the complete list.

Refs:
 * https://android.googlesource.com/platform/bionic/+/refs/tags/ndk-r25c/libc/kernel/uapi/asm-generic/unistd.h
 * https://android.googlesource.com/platform/bionic/+/refs/tags/ndk-r25c/libc/include/bits/glibc-syscalls.h

### DIFF
--- a/libc-test/semver/android.txt
+++ b/libc-test/semver/android.txt
@@ -2290,7 +2290,9 @@ SYS_clock_gettime
 SYS_clock_nanosleep
 SYS_clock_settime
 SYS_clone
+SYS_clone3
 SYS_close
+SYS_close_range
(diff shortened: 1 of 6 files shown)


JohnTitor:
We use r21d on CI and I guess the test would fail, have you run it on your end?

sunfishcode:
Ah, no, I have not run the CI tests with the Android SDK. So, this should wait until the CI is updated.

PR #3000, opened 2022-11-11
Report the actual error when failing to get the rustc version

bjorn3:
This should help with pinpointing the issue in rust-lang/crater#663.

PR #3180, opened 2023-04-01
build.rs: print rustc stderr if exit status != 0

jsha:
I was trying to run benchmarks locally with rustc-perf and found that many of them failed to build with a message from libc's build.rs "Failed to get rustc version." I made this change locally to help debug, and I think it would be generally useful. In my case it quickly revealed that rustc was failing to find libLLVM and so `rustc --version` was emitting nothing on stdout.

I think this may have been part of what was intended with #3000 and might help debug https://github.com/rust-lang/crater/issues/663.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,7 @@
 use std::env;
 use std::process::Command;
 use std::str;
+use std::string::String;
 
 // List of cfgs this build script is allowed to set. The list is needed to support check-cfg, as we
 // need to know all the possible cfgs that this script will set. If you need to set another cfg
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):

build.rs
```diff
@@ -181,6 +181,11 @@ fn rustc_minor_nightly() -> (u32, bool) {
         .output()
         .ok()
         .expect("Failed to get rustc version");
+    if !output.status.success() {
+        let stderr = std::string::String::from_utf8_lossy(output.stderr.as_slice());
+        eprintln!("failed to run rustc: {stderr}");
```
JohnTitor, line 186: This syntax doesn't work on old rustc.

JohnTitor:
Note that you saw the build _on PRs_, not bors. The bors' build is here: https://cirrus-ci.com/build/5602401026048000
The FreeBSD failure happens on your PR because your branch is old, it's fixed on master so rebasing should resolve it.
The actual error is here (you can find the link from bors comment): https://github.com/rust-lang/libc/actions/runs/4601182601/jobs/8128728585#step:4:237

It's a network failure which means we should: @bors retry

Issue #1239, opened 2019-02-04
CMSG_NXTHDR hits a Bus Error on sparc64-unknown-linux-gnu


asomers:
PR #1235 adds a new test program to ensure that the cmsg(3) family of functions are correctly implemented on all targets.  However, when run in QEMU `CMSG_NXTHDR` hits a Bus Error.  This could be an error in libc's `CMSG_NXTHDR` implementation for that platform, or it could be an error in Linux's definition of the macro (doesn't work on unaligned buffers or something like that).  Less likely would be an error in QEMU.  Lacking access to sparc hardware, I didn't investigate further.

https://travis-ci.com/rust-lang/libc/jobs/175199673

gnzlbg:
This reproduces on master, that is, updating the Debian version did not fix this (it did fix the s390x issue). 

Chances are, there is a bug in our implementation.

gnzlbg:
I'm also seeing sometimes a:

```
running 5 tests
test t::test_cmsg_data ... ok
error: process didn't exit successfully: `qemu-arm -L /musl-arm /checkout/target/arm-unknown-linux-musleabihf/debug/deps/cmsg-7238071f2e3df703` (signal: 11, SIGSEGV: invalid memory reference)
The command "if [[ $TRAVIS_OS_NAME = "linux" ]] && [[ $BUILD_ONLY != "1" ]]; then sh ci/run-docker.sh $TARGET; else sh ci/run.sh $TARGET; fi" exited with 101.
```

seg fault on the arm build jobs in the cmsg tests. Restarting the build job fixes the issue, but it might mean that something fishy is going on there.

PR #3183, opened 2023-04-02
Use aligned `cmsghdr` structs `test_cmsg_nxthdr`

### LINKED ISSUE #3181: `test_cmsg_nxthdr` fails due to alignment checks for pointer dereferences
Example run: https://cirrus-ci.com/task/5904059698249728?logs=test#L85:
```
running 5 tests
test t::test_cmsg_data ... ok
test t::test_cmsg_len ... ok
test t::test_cmsg_nxthdr ... FAILED
failures:
---- t::test_cmsg_nxthdr stdout ----
thread 't::test_cmsg_nxthdr' panicked at 'misaligned pointer dereference: address must be a multiple of 0x4 but is 0x1b6966ed4d1', libc-test/test/cmsg.rs:74:25
```

It's caused by https://github.com/rust-lang/rust/pull/98112

ChrisDenton:
Fixes #3181.

I could find no reason for using unaligned structs in this test.

### DIFF
--- a/libc-test/test/cmsg.rs
+++ b/libc-test/test/cmsg.rs
@@ -55,34 +55,32 @@ mod t {
     // https://github.com/rust-lang/libc/issues/1239
     #[cfg(not(target_arch = "sparc64"))]
     #[test]
-    // FIXME: This triggers alignment checks for pointer dereferences:
-    // https://github.com/rust-lang/libc/issues/3181
-    #[ignore]
     fn test_cmsg_nxthdr() {
         use std::ptr;
+        // Helps to align the buffer on the stack.
+        #[repr(align(8))]
+        struct Align8<T>(T);
 
-        let mut buffer = [0u8; 256];
+        const CAPACITY: usize = 512;
+        let mut buffer = Align8([0_u8; CAPACITY]);
         let mut mhdr: msghdr = unsafe { mem::zeroed() };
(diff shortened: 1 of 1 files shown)


ChrisDenton:
I'm trying to get feedback in case there's something I've missed. I don't quite understand why it's structured the way it is.

ChrisDenton:
It looks like #1239 might have been caused by this too?

JohnTitor:
> It looks like #1239 might have been caused by this too?

Not sure, could you remove the cfg and related comments?

ChrisDenton:
Ok, that seems to have worked?

JohnTitor:
CI on PRs doesn't run on Sparc64, we need:
@bors try

ChrisDenton:
Oops, I missed a `#[cfg]`.

 @bors try

ChrisDenton:
Hm, that still failed. I'll try bumping up the alignment...

ChrisDenton:
Ok, I don't think I can fix sparc today. It's possible there may still be a problem in the test, or perhaps in `CMSG_NXTHDR`. But I can't immediately see it. Honestly it would probably be worth rewriting some of the `CMSG_` "macros" using modern pointer methods in case that shows up anything. But that's a job for another time.

For now I've aligned the buffer and also cleaned up a few of the pointer bits in case that triggers any debug asserts in the future. Nothing major though.

PR #3189, opened 2023-04-06
Remove AIX specific definition of sigval

ecnelises:
AIX definition of sigval is actually the same as other unix. Remove the union definition as other platforms treat it like a pointer after #3185.


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #3195, opened 2023-04-11
openbsd 73 add get/setthrname calls

devnexen:


semarie:
any reason to want these somehow internals syscalls ? Rust libc already provides `pthread_get_name_np` and `pthread_set_name_np` for the get/set the thread name.

to cite the man page:
```
     The getthrname() and setthrname() system calls are specific to OpenBSD;
     pthread_get_name_np(3) and pthread_set_name_np(3) operate on the same
     thread name in a more portable way.
```

semarie:
the tid is showed by `ps`, but I don't think that Rust libc has interface to get it. it isn't the standard way to reference a thread.

please understand that just adding functions because they exists could hurt later. OpenBSD doesn't care about backward compatibility specially for such internals functions.

devnexen:
you re making a good point, I did not realise it was an internal function.

PR #3201, opened 2023-04-14
Netlink support for FreeBSD 13.2

### LINKED ISSUE #3194: Netlink support on FreeBSD 13.2+
FreeBSD 13.2 (here x86_64-unknown-freebsd) gained the Netlink layer, which seems to be
supported today only for Linux-like and Fuschia.

Bindings at https://github.com/rust-netlink/netlink-sys need a couple of bits from libc do build, most notably `PF_NETLINK`, `sockaddr_nl`, `SOL_NETLINK` and friends.

ydirson:
Closes #3194, to the point that [Xen Guest Agent](https://gitlab.com/xen-project/xen-guest-agent) can works on FreeBSD.

Requires breaking-change #3367.

Downstream support in `rust-netlink`:
- [x] `netlink-packet-route`: FreeBSD support available since v0.18
- [x] `rtnetlink`: FreeBSD support available since v0.14.1 (from https://github.com/rust-netlink/rtnetlink/pull/49)

Also see https://github.com/rust-lang/libc/pull/3555 for a backport to `libc-0.2` branch using a feature flag to avoid the breaking change by default.

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -1388,6 +1388,14 @@ s_no_extra_traits! {
         pub sdl_data: [::c_char; 46],
     }
 
+    pub struct sockaddr_nl {
+        pub nl_len: ::c_uchar,
+        pub nl_family: ::sa_family_t,
+        nl_pad: ::c_ushort,
+        pub nl_pid: u32,
+        pub nl_groups: u32
+    }
+
     pub struct mq_attr {
         pub mq_flags: ::c_long,
         pub mq_maxmsg: ::c_long,
@@ -1785,6 +1793,34 @@ cfg_if! {
                 self.sdl_alen.hash(state);
                 self.sdl_slen.hash(state);
                 self.sdl_data.hash(state);
+            }
+        }
+
+        impl PartialEq for sockaddr_nl {
+            fn eq(&self, other: &sockaddr_nl) -> bool {
+                self.nl_len == other.nl_len &&
+                    self.nl_family == other.nl_family &&
+                    self.nl_pid == other.nl_pid &&
+                    self.nl_groups == other.nl_groups
+            }
+        }
+        impl Eq for sockaddr_nl {}
+        impl ::fmt::Debug for sockaddr_nl {
+            fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
+                f.debug_struct("sockaddr_nl")
+                    .field("nl_len", &self.nl_len)
+                    .field("nl_family", &self.nl_family)
+                    .field("nl_pid", &self.nl_pid)
+                    .field("nl_groups", &self.nl_groups)
+                    .finish()
+            }
+        }
+        impl ::hash::Hash for sockaddr_nl {
+            fn hash<H: ::hash::Hasher>(&self, state: &mut H) {
+                self.nl_len.hash(state);
+                self.nl_family.hash(state);
+                self.nl_pid.hash(state);
+                self.nl_groups.hash(state);
             }
         }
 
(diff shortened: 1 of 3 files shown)


ydirson:
The freebsd14 warning about `NETLINK_GENERIC` not being used is likely about its shadowing. Should we add a `cfg` restriction to the one in `freensd/mod.rs`? I'm not even sure it has any usefulness, and is not listed in the semver list for freebsd, so maybe we should just nuke it instead?

JohnTitor:
> The freebsd14 warning about `NETLINK_GENERIC` not being used is likely about its shadowing. Should we add a `cfg` restriction to the one in `freensd/mod.rs`? I'm not even sure it has any usefulness, and is not listed in the semver list for freebsd, so maybe we should just nuke it instead?

Because it's already defined:
https://github.com/rust-lang/libc/blob/0b758afcdc34c9a72508a481ba151857d0b469de/src/unix/bsd/freebsdlike/freebsd/mod.rs#L3342

ydirson:
I see it's already defined.  The problem is, FreeBSD has 2 defines with same name and different values, each for a different API and defined in their own header file.  The separate header files allow this to work for C by creating a poor man's namespacing, but how are we to handle that in the Rust case?

Some options I can see include:
- change the definition of the other FreeBSD API providing the clashing define (breaking API for them, no clue how widespread its use is in Rust or even if it is used at all - can we know that?)
- use a different name for the FreeBSD Netlink API's const: that make the FreeBSD Rust Netlink API incompatible with that of Linux-like OS, which does not look clean and has to be handled specially in crates building on this to provide the rustified API (or really on any crate using the libc Netlink API - that raises some documentation issues to make sure this shape-changing API would be properly used)
- change the definition of the Netlink API in libc crate, possibly with a compatibility layer on linux-like for transition (which would make the Netlink API more special than the rest of libc)

What do you think?

asomers:
Regarding NETLINK_GENERIC, ask @GuillaumeGomez .  He's the one who added the definition from if_mib.h .  Also, grep.app shows that his sysinfo crate is the only one that's using it as he intended.  There are many more that try to use your version.  Including https://github.com/bytecodealliance/rustix/blob/db143034d80dbe2dc62368b732378c89f16e7569/src/net/types.rs#L800 , which seems to be confused by the definition in libc.  One version or the other will have to change.  I suspect that the least overall pain to the community would come by removing or renaming the definition from if_mib.h

Also, you should move your definitions out of the freebsd13 and freebsd14 modules and into the freebsd/mod.rs file.  The former two should only be used for stuff that _changes_ between versions.  Symbols that newly appear in a newer version of FreeBSD can safely be defined by libc in for all OS versions.  For example, `ino_t` changed between versions, so it must be defined in freebsd11, freebsd12, etc.

GuillaumeGomez:
FreeBSD and its handling of APIs is a nightmare. They don't care about compatibility, they change types/constants instead of creating new ones. So if `NETLINK_GENERIC` exists in two different places with different values, to prevent breaking libc's API, here's what I recommend: adding docs to the current `NETLINK_GENERIC` telling where to look for which values and then create both constants in two submodules (which maybe mention each others?).

asomers:
What do you mean @GuillaumeGomez ?  This problem is caused by if anything too much backwards-compatibility, not too little.  The NETLINK_GENERIC mib that you're using was added in 1996, and it still works.  The NETLINK_GENERIC that @ydirson wants, by contrast, is based on an RFC written in 2003.  By then it was too late to change the name of the old constant, but changing the name of the new constant would break programs' cross-platform compatibility.  Hence isolating them in separate headers.

Is there any precedent for adding submodules to libc?

GuillaumeGomez:
I mean that if they wanted to change the value of `NETLINK_GENERIC`, they should have created a new constant (called `NETLINK_GENERIC2` for example) instead of just changing the value. They often break API, the worst being in structs where they change the fields order or even remove/add some.

> Is there any precedent for adding submodules to libc?

No clue but if there is a good enough reason, I think it could be ok (up to them though!).

asomers:
> I mean that if they wanted to change the value of `NETLINK_GENERIC`, they should have created a new constant (called `NETLINK_GENERIC2` for example) instead of just changing the value. They often break API, the worst being in structs where they change the fields order or even remove/add some.

Are you talking about stuff like `struct stat`?  That, I think, is really cool.  FreeBSD uses ELF symbol versioning to change structs like that.  That means zero problems for C code; old code will still compile, and old binaries will still run.  And yet new code can take advantage of new features _without_ introducing new syscalls like Linux's `stat64`.  The only problems come when using FFI, like Rust does.  Rust basically chose to be easily cross-compilable at the expense of not being able to take advantage of ELF symbol versions.

> 
> > Is there any precedent for adding submodules to libc?
> 
> No clue but if there is a good enough reason, I think it could be ok (up to them though!).

I don't see why it couldn't be done.  I think it's preferable to move the NETLINK_GENERIC mib into a submodule, because unlink the NETLINK_GENERIC protocol family it is platform-specific.  If we do that, what do you think the submodule should be named, and what other symbols should go with it?

GuillaumeGomez:
> Are you talking about stuff like `struct stat`? That, I think, is really cool. FreeBSD uses ELF symbol versioning to change structs like that. That means zero problems for C code; old code will still compile, and old binaries will still run. And yet new code can take advantage of new features _without_ introducing new syscalls like Linux's `stat64`. The only problems come when using FFI, like Rust does. Rust basically chose to be easily cross-compilable at the expense of not being able to take advantage of ELF symbol versions.

I don't know enough about this to comment. From my understanding, having multiple types depending on which *minimum* version you support is **much** better and simpler. Anyway, there is no point debating over whether or not FreeBSD took the right decision here.

ydirson:
> Also, you should move your definitions out of the freebsd13 and freebsd14 modules and into the freebsd/mod.rs file.

Pushed this, but:
- it won't compile anywhere before the name clash gets solved, so I stacked a temporary commit to comment the if_mib definition out to keep the branch usable
- libc-test now reports problems on `freebsd12`, which was the reason why I had moved the code in version-specific modules

asomers:
>     * libc-test now reports problems on `freebsd12`, which was the reason why I had moved the code in version-specific modules
You will have to add an exception in libc-test.  See copy_file_range for an example.

ydirson:
Starting to introduce namespaces in libc will have an impact on those generated tests, we'll likely need someone mastering this part?

ydirson:
Now rebased on top of #3367 to get rid of the previous hack around `NETLINK_GENERIC`

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -3359,51 +3512,6 @@ pub const IFDR_MSG_SIZE: ::c_int = 64;
 pub const IFDR_REASON_MSG: ::c_int = 1;
 pub const IFDR_REASON_VENDOR: ::c_int = 2;
 
-// sys/net/if_mib.h
-
-/// non-interface-specific
-pub const IFMIB_SYSTEM: ::c_int = 1;
```
GuillaumeGomez, line 3365: This is a breaking change btw (moving a constant to a different module).

ydirson: Yes, this is handled separately as #3367, on which this PR is based

ydirson:
Rebased onto 0.2.151

ydirson:
Since 0.2 was branched out I cannot really rebase this branch any more to keep it usable (rustix just pushed a new version demanding libc 0.2.152), so I pushed the new rebased version to https://github.com/xcp-ng/rust-libc/tree/freebsd-netlink-0.2

ydirson:
This last push removes the pcap/nflog part, which is indeed not directly related, and not necessary for our use case.

That removed part is at https://github.com/xcp-ng/rust-libc/tree/freebsd-pcap-nflog in case it is useful for a new PR.

ydirson:
... and rebased to avoid failure of obsolete freebsd12 build job

ydirson:
@tgross35 What does this `S-blocked` tag imply?  Anything that could be added to `CONTRIBUTING.md`?

tgross35:
It just means that something needs to happen before this can move forward. In this case I believe that is https://github.com/rust-lang/libc/pull/3367.

ydirson:
Makes sense, and now I realize that #3367 is itself blocked, and indeed needs a rebase (and maybe it should be added to the 1.0 milestone too) - but I only saw it by chance because of your post in #3248, received no notification of the tag change, and @bors did not report the need to rebase either :shrug:

tgross35:
No need to rebase just yet, I need to read into things better and figure out what series of events needs to happen. I'll post an update when I figure that out.

tgross35:
Sorry this has sat like so many libc PRs, but I've been thinking about this a bit recently. Here is what I would like to do:

PR 1:

1. Move `if_mib` constants to new files within `src/new`, since the new structure splits things up by header file anyway.
2. Reexport everything from `if_mib`, like we do for other existing modules in `src/new`
3. Add a `FIXME(1.0)` to drop this re-export

That should mean no changes to the public API

PR 2:

1. Add netlink within `src/new` 
2. Make both the netlink and the if_mib modules public

This means that:

* Existing users of `if_mib` can continue to use that
* Anyone aware of this change will be able to pick importing from `if_mib` or from `netlink`
* Once 1.0 releases, anybody who was relying on `if_mib` from root will break. This is intentional to avoid problems from quietly switching to netlink.
* Some months after the 1.0 release, we can consider re-exporting from `netlink` if we think that's worth it

@asomers @JohnTitor any thoughts?

@rustbot author

asomers:
I think that @tgross35 's proposal is probably the only thing that will work while preserving backwards-compatibility for all current users.

dybucc:
not sure if the pr author is still on this. if not, i've picked up where they left off and finished things up as per [^1] in [^2].

note for these changes to be complete; namely, for the if_mib module to also be made public, [^3] needs merging first. i've also commented there with updated changes. ci neither passes at the moment.

[^2]: https://github.com/dybucc/libc/commit/bf155a5abfd0e890ca9849f49a255dd259cdbe07
[^3]: https://github.com/rust-lang/libc/pull/3367

dybucc:
An updated PR has been submitted at #5326.

tgross35:
Closing since https://github.com/rust-lang/libc/pull/5326 supersedes it, please take a look at that PR if you get the chance.

PR #3206, opened 2023-04-19
Add PF_ROUTE-related structures on Apple systems

roblabla:
This commit adds the following structures from [net/route.h](https://github.com/roblabla/MacOSX-SDKs/blob/master/MacOSX13.3.sdk/usr/include/net/route.h):

- rt_msghdr
- rt_msghdr2
- rt_metrics

and the following structures from [net/if.h](https://github.com/roblabla/MacOSX-SDKs/blob/master/MacOSX13.3.sdk/usr/include/net/if.h):

- ifa_msghdr
- ifma_msghdr
- ifma_msghdr2

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -435,6 +435,80 @@ s! {
         pub ifm_data: if_data,
     }
 
+    pub struct ifa_msghdr {
+        pub ifam_msglen: ::c_ushort,
+        pub ifam_version: ::c_uchar,
+        pub ifam_type: ::c_uchar,
+        pub ifam_addrs: ::c_int,
+        pub ifam_flags: ::c_int,
+        pub ifam_index: ::c_ushort,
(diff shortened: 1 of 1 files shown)


JohnTitor:
The test still fails, could you check?
@rustbot author

roblabla:
And this is why I shouldn't trust apple's manpages. They are filled with lies, lies and more lies.

This should be all fixed now.

JohnTitor (approved):

JohnTitor:
Thank you! @bors r+

nyurik:
@roblabla I'm not certain if this is related, but CI no longer passes on MacOS, and all errors relate to `rmx_state` and other MacOS specific stuff.  Any idea what might be causing it? Thx!

roblabla:
rmx_state was removed in macOS 14.4. https://github.com/rust-lang/libc/pull/3776 fixes it.

PR #2480, opened 2021-10-24
ARMv6K Horizon OS support

Meziu:
Added support for the Horizon OS Nintendo 3DS system `Newlib` declarations.

Edit: after some changes, networking is now stable with standardized API.

PR #3209, opened 2023-04-20
armv7 PSVita OS support via newlib

nikarh:
Added support for ps vita via newlib.
Similar to #2480, but for a different device.

I'm working on adding support of std for vita, and this is essentially a prerequisite.


JohnTitor:
We received a report about a similar target in the past (sorry, I cannot reveal it as it's sensitive), so I'd like to confirm with the core team if everything is OK before accepting this.

JohnTitor (approved):

JohnTitor:
Ah, we can go ahead then. Thank you for clarifying! @bors r+

PR #3211, opened 2023-04-21
uclibc/mips: fixed SA_* mismatched types

cppcoffee:
use ulong type uniformly.

### DIFF
--- a/src/unix/linux_like/linux/uclibc/mips/mod.rs
+++ b/src/unix/linux_like/linux/uclibc/mips/mod.rs
@@ -23,10 +23,10 @@ pub const ECOMM: ::c_int = 70;
 pub const EPROTO: ::c_int = 71;
 pub const EDOTDOT: ::c_int = 73;
 
-pub const SA_NODEFER: ::c_int = 0x40000000;
-pub const SA_RESETHAND: ::c_int = 0x80000000;
-pub const SA_RESTART: ::c_int = 0x10000000;
-pub const SA_NOCLDSTOP: ::c_int = 0x00000001;
+pub const SA_NODEFER: ::c_uint = 0x40000000;
+pub const SA_RESETHAND: ::c_uint = 0x80000000;
+pub const SA_RESTART: ::c_uint = 0x10000000;
+pub const SA_NOCLDSTOP: ::c_uint = 0x00000001;
 
 pub const EPOLL_CLOEXEC: ::c_int = 0x80000;
 


cppcoffee:
The following error occurs when compiling `nix` crate on the mipsel platform:

```
   --> /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/nix-0.24.3/src/macros.rs:70:35
    |
70  |                       const $Flag = libc::$Flag $(as $cast)*;
    |                                     ^^^^^^^^^^^ expected `u32`, found `i32`
    |
   ::: /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/nix-0.24.3/src/sys/signal.rs:436:1
    |
436 | / libc_bitflags!{
437 | |     /// Controls the behavior of a [`SigAction`]
438 | |     #[cfg_attr(docsrs, doc(cfg(feature = "signal")))]
439 | |     pub struct SaFlags: SaFlags_t {
...   |
461 | |     }
462 | | }
    |
    = note: this error originates in the macro `libc_bitflags` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I found the uclibc/arm definition of SA_* using the ulong type: https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/linux/uclibc/arm/mod.rs#L472- L478 

Modified to use ulong type uniformly, can compile `nix` now.

Please review it.

JohnTitor:
This is a breaking change and I'd recommend deprecating them first to inform users.

cppcoffee:
Hi @JohnTitor , deprecated attribute has been added, please review it.

JohnTitor:
Please do not change their types, also it'd be great if could squash commits into one once it's done!

cppcoffee:
Hi @JohnTitor , Does that mean adding only the `deprecated` attribute?

JohnTitor:
Yes, we should inform users that we're going to introduce a breaking change before actually doing so. By this users can prepare that change and we can prevent their code from being broken.

src/unix/linux_like/linux/uclibc/mips/mod.rs
```diff
@@ -23,10 +23,14 @@ pub const ECOMM: ::c_int = 70;
 pub const EPROTO: ::c_int = 71;
 pub const EDOTDOT: ::c_int = 73;
 
-pub const SA_NODEFER: ::c_int = 0x40000000;
-pub const SA_RESETHAND: ::c_int = 0x80000000;
-pub const SA_RESTART: ::c_int = 0x10000000;
-pub const SA_NOCLDSTOP: ::c_int = 0x00000001;
+#[deprecated(since = "0.2.143", note = "Use c_ulong instead.")]
```
JohnTitor, line 26:
```suggestion
#[deprecated(since = "0.2.143", note = "we 're going to change its type to `c_ulong` in a future release")]
```

cppcoffee:
Hi @JohnTitor , Add `deprecated` attribute only, please review.

cppcoffee:
OK, I'll take care of it when I can rebase it.

tgross35:
@cppcoffee could you squash this so it drops the merge commits? Looks like this

```
# Assuming this repo (rust-lang, not your fork) is called `upstream`
git fetch upstream

git rebase -i upstream/main
# This command opens an editor. For everything EXCEPT the first line,
# replace `pick` with `f` or `fixup`.
#
# There might not actually be any lines besides the first here, if not
# just close it.
#
# When you close the editor, it will have combined the commits.

# Resolve conflicts if needed, follow the instructions it gives

# Push to this branch with the new history
git push --force-with-lease
```

Also a couple different ways to do it here https://stackoverflow.com/questions/5189560/how-do-i-squash-my-last-n-commits-together

---

I will backport only this PR (to target stable, 0.2) but as soon as it merges, you can create a new PR doing the actual fix (which will be in 1.0 when that happens). I opened https://github.com/rust-lang/libc/issues/3887 to make sure we don't forget this.

tgross35:
Hey @cppcoffee, thank you for updating but this needs a rebase and squash - not merge (no merge commits policy). Could you try to follow the instructions from my comment? If that doesn't work then let me know and I can help you figure it out (or worst case I can do it for you, but I think you should be able to get it working).

cppcoffee:
Rebase done.

tgross35:
@cppcoffee thank you for getting this up to date. Unfortunately after some digging, it seems like this wouldn't be correct - see https://github.com/rust-lang/libc/issues/3887#issuecomment-2319939888.

What would be correct is to change all the `SA_*` constants to be `::c_uint` on mips ulibc. And then open an issue/PR to [`nix`](https://github.com/nix-rust/nix) to do the same. If this sounds right, could you update this PR to do that?

You can just do the change without deprecation notices, it just won't get released until 1.0.

cppcoffee:
Oh, it looks like `nix` is using the wrong type.

cppcoffee:
Hi @tgross35, Use the `c_uint` type for `SA_*`.

tgross35:
LGTM, but please rebase and squash

tgross35 (approved):

tgross35:
@cppcoffee could you rebase and squash as mentioned above?

cppcoffee:
rebase and squash done.

tgross35:
Thanks! There are some CI issues but it's not related to your changes.

PR #3173, opened 2023-03-28
Upstream a few more constants from rustix.

sunfishcode:
Add a definition of `CIBAUD` for powerpc and powerpc64 on glibc.

And a definition for `RLIM64_INFINITY` on linux_like platforms.

PR #3212, opened 2023-04-21
Define `FICLONE` on mips and power.

sunfishcode:
PR #3173 added definitions of `FICLONE` for mips and power to a generic file, however mips and power have their own versions of this file, so move their `FICLONE` definitions into their own files.


JohnTitor (approved):

JohnTitor:
Good call, thanks! @bors r+

PR #3214, opened 2023-04-21
redox adding few calls.

devnexen:


JohnTitor (approved):

JohnTitor:
Seems an import is needed: https://github.com/rust-lang/libc/actions/runs/4770931323/jobs/8482757206#step:5:6368

PR #3222, opened 2023-04-25
haiku: fix incorrect linked library.

jessicah:
Introduced in fb2a763; using a non-existent "libunix". This is in libgnu.

### DIFF
--- a/src/unix/haiku/mod.rs
+++ b/src/unix/haiku/mod.rs
@@ -2055,7 +2055,7 @@ extern "C" {
     ) -> ::c_int;
 }
 
-#[link(name = "unix")]
+#[link(name = "gnu")]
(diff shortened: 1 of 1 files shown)


JohnTitor:
I'm not familiar with Haiku so just to confirm, should every Haiku users have the libgnu, i.e. does it need an additional setup? If it should it's fine.

jessicah:
Haiku has non-posix extensions in libbsd and libgnu; they're part of the system, however.

nielx:
Can confirm that this fixes the build, the incorrect linking was introduced in #3165.

JohnTitor (approved):

JohnTitor:
It should be fine then, thanks for clarifying! @bors r+

PR #3226, opened 2023-04-26
Don't link `legacy_stdio_definitions` from std

ChrisDenton:
std on windows-msvc does not use `printf` or `fprintf` so never needs the `legacy_stdio_definitions.lib` import library and will always work fine without it.

### DIFF
--- a/src/windows/mod.rs
+++ b/src/windows/mod.rs
@@ -278,13 +278,17 @@ impl ::Clone for fpos_t {
 }
 
 // Special handling for all print and scan type functions because of https://github.com/rust-lang/libc/issues/2860
-#[cfg_attr(
-    all(windows, target_env = "msvc"),
-    link(name = "legacy_stdio_definitions")
-)]
-extern "C" {
-    pub fn printf(format: *const c_char, ...) -> ::c_int;
-    pub fn fprintf(stream: *mut FILE, format: *const c_char, ...) -> ::c_int;
+cfg_if! {
+    if #[cfg(not(feature = "rustc-dep-of-std"))] {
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):

src/windows/mod.rs
```diff
@@ -278,13 +278,17 @@ impl ::Clone for fpos_t {
...
+    if #[cfg(not(feature = "rustc-dep-of-std"))] {
+        #[cfg_attr(
+            all(windows, target_env = "msvc"),
+            link(name = "legacy_stdio_definitions")
+        )]
+        extern "C" {
+            pub fn printf(format: *const c_char, ...) -> ::c_int;
+            pub fn fprintf(stream: *mut FILE, format: *const c_char, ...) -> ::c_int;
+        }
+    }
```
Urgau, line 291:
```suggestion
#[cfg(not(feature = "rustc-dep-of-std"))]
#[cfg_attr(
    all(windows, target_env = "msvc"),
    link(name = "legacy_stdio_definitions")
)]
extern "C" {
    pub fn printf(format: *const c_char, ...) -> ::c_int;
    pub fn fprintf(stream: *mut FILE, format: *const c_char, ...) -> ::c_int;
}
```
nit: no need to use `cfg_if!` here

ChrisDenton: That was what I originally did. but the style checker complained.

OwenLu88:
Which version will update?  Now, I using "rustc 1.69.0 (84c898d65 2023-04-16)". Is it the latest version?

PR #3230, opened 2023-04-28
Added getentropy to vita target

nikarh:
This is a late addition to https://github.com/rust-lang/libc/pull/3209. This definition is required in order to implement random in std correctly.

As the previous PR, getentropy is a standard C function, should be implemented by newlib provider, and is not Sony's intellectual property.


JohnTitor (approved):
Thank you!

PR #3235, opened 2023-05-05
Disable tests using build-std temporarily

JohnTitor:


JohnTitor:
The fix will be available soonish: https://github.com/rust-lang/cargo/pull/12088

PR #3238, opened 2023-05-07
more bsd things: kenv, setlogin, reboot, SCM_ constants

valpackett:


semarie:
Tests are fine on OpenBSD (it only tests the `setlogin()` addition on OpenBSD).
I can't really comment for the support in all BSDs, but it is an old function (it comes from 4.3BSD-Reno).

JohnTitor (approved):

JohnTitor:
Thank you both, LGTM! @bors r+

PR #3246, opened 2023-05-11
Fix loongarch64 bindings

heiher:
This PR aims to fix the LoongArch64 bindings and incorporate tests for LoongArch.

- libc-test: PASSED


JohnTitor:
The struct mcontext_t has breaking changes, but the loongarch64 target is relatively new and I guess we could assume there's (almost) no user-facing issue. Could we confirm that assumption somehow?

heiher:
Indeed, this is a bug fix. Currently, we are in the early bring-up stage of LoongArch, and it's a good opportunity to fix this issue.

src/unix/linux_like/linux/gnu/b64/loongarch64/align.rs
```diff
@@ -17,10 +17,10 @@ s! {
...
     #[repr(align(16))]
     pub struct mcontext_t {
-        pub sc_pc: ::c_ulonglong,
-        pub sc_regs: [::c_ulonglong; 32],
-        pub sc_flags: ::c_ulong,
-        pub sc_extcontext: [u64; 0],
+        pub __pc: ::c_ulonglong,
+        pub __gregs: [::c_ulonglong; 32],
+        pub __flags: ::c_uint,
+        pub __extcontext: [::c_ulonglong; 0],
```
heiher, line 23: https://github.com/bminor/glibc/blob/d6c72f976c61d3c1465699f2bcad77e62bafe61d/sysdeps/unix/sysv/linux/loongarch/sys/ucontext.h#L46-L52

heiher:
`struct mcontext_t` is an architecture-specific structure, and I conducted a thorough review of all libc users, but I did not find any merged code that utilizes it. It is now known that this part is unusable in the released versions of libc, but it is indeed a viable way to fix the issue. The good news is that it was found in the early phase.

heiher:
V2: Add `struct user_regs_struct` and `struct user_fp_struct`.

src/unix/linux_like/linux/gnu/b64/loongarch64/mod.rs
```diff
@@ -189,6 +189,21 @@ s! {
...
+        pub csr_badv: u64,
+        pub reserved: [u64; 10],
+
+    }
+
+    pub struct user_fp_struct {
+        pub fpr: [u64; 32],
+        pub fcc: u64,
+        pub fcsr: u32,
+    }
```
heiher, line 206: https://github.com/bminor/glibc/blob/d6c72f976c61d3c1465699f2bcad77e62bafe61d/sysdeps/unix/sysv/linux/loongarch/sys/user.h#L24-L40

```c

struct user_regs_struct
{
  /* Saved main processor registers. */
  uint64_t regs[32];

  /* Saved special registers. */
  uint64_t orig_a0;
  uint64_t csr_era;
  uint64_t csr_badv;
  uint64_t reserved[10];
};

struct user_fp_struct {
  uint64_t    fpr[32];
  uint64_t    fcc;
  uint32_t    fcsr;
};
```

heiher:
@JohnTitor Is it okay in your opinion?

JohnTitor (approved):

JohnTitor:
Yeah, thank you for clarifying! @bors r+

PR #3247, opened 2023-05-12
Add linux canxl constants and canxl frame struct

marcelbuesing:
Add constants and structs related to CANXL support more info related to CANXL [here](https://lore.kernel.org/linux-can/20220724074402.117394-1-socketcan@hartkopp.net/).


marcelbuesing:
Not sure yet but my assumption is that it's currently failing because it was just added in the kernel in `v6.1` and docker base image (Ubuntu 22.04.2)  is based on 5.19. I guess this would require ubuntu 23.04 to make it work.

JohnTitor:
You can skip tests for now by tweaking below:
https://github.com/rust-lang/libm/pull/281

A FIXME comment describing what version introduces these items would be helpful so that we can remove them once upgrading CI images.

marcelbuesing:
> You can skip tests for now by tweaking below:
> 
> https://github.com/rust-lang/libc/blob/e1c4a383d1f049518f0595614632ff773b4c4c5f/libc-test/build.rs#L3476
> 
> A FIXME comment describing what version introduces these items would be helpful so that we can remove them once upgrading CI images.

Thank you I just added them to the build.rs

JohnTitor:
@joshtriplett Please do not merge directly, we only run full CI via bors.

PR #3252, opened 2023-05-17
android: add memmem

tibordp:
Adding `memmem` to Android bindings.

Android NDK has had `memmem` probably since the beginning but definitely as early as Android 2.2.

https://android.googlesource.com/platform/ndk.git/+/refs/tags/android-2.2_r1/build/platforms/android-3/arch-arm/usr/include/string.h#44


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #3255, opened 2023-05-26
Fixed vita libc definitions

nikarh:
This fixes definitions that were incorrect in my initial PR (#3209). As with the previous one, this relies on open-source newlib implementation and doesn't use Sony internal API's or expose Sony internal data structures.

This fixes:
- fs API that uses `dirent` and stat
- sockets
- pthreads

A couple of explanations
- Unfortunately `dirent` in vita newlib is not very POSIX'y, before `d_name` it has a field with an internal data structure, which is of no use for std (no inodes, no file type inside), so I've just stubbed it as an `__offset`. Also, Vita doesn't expose inodes - `dirent` doesn't have it at all, and in `stat` it is always `0`. I am not sure what would be the best approach here. Maybe I should move the POSIX `dirent` to `generic.rs` and reexpose it in all targets, and redefine it in `vita/mod.rs`?
- All pthread types on Vita are pointers, and the backing data structure is allocated by corresponding init functions, so their sizeof's are all 4. I also changed `pthread_attr_t` and `pthread_rwlockattr_t` to reflect their sizes and not be constant.

The fixes in std will be in rust-lang/rust#111819


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

src/unix/newlib/mod.rs
```diff
@@ -160,8 +171,12 @@ s! {
     }
 
     pub struct dirent {
+        #[cfg(not(target_os = "vita"))]
         pub d_ino: ino_t,
+        #[cfg(not(target_os = "vita"))]
         pub d_type: ::c_uchar,
+        #[cfg(target_os = "vita")]
+        __offset: [u8; 88],
```
riking, line 179: This is a repr(Rust) struct, so this offset doesn't actually offset d_name. In fact, the compiler will preferentially place d_name at offset 0 because it has a higher natural alignment

nikarh:
Hey, thanks for reviewing this!

In vita newlib dirent is weird and contains a 88 byte sized internal data
structure at offset 0 (which we neither really want to expose nor care
about here). And all other fields defined before this offset are filtered
by a macro for vita.
Plus this structure should be defined in s! macro which adds repr(C).

riking:
Yes, I saw the s! macro after writing that and deleted my comment.

PR #3257, opened 2023-05-29
linux-gnu: add putpwent/putgrent

superwhiskers:
adds the [`putpwent(3)`](https://www.man7.org/linux/man-pages/man3/putpwent.3.html) and [`putgrent(3)`](https://www.man7.org/linux/man-pages/man3/putgrent.3.html) glibc extensions. these don't appear to be present anywhere else from what i could tell, correct me if i'm wrong.


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #3261, opened 2023-06-01
Unify definitions of `siginfo_t`, `statvfs` and `statfs` in `musl` targets

bossmc:
During #2935 I noticed there were multiple identical definitions of some of the `bits/***.h` types in the  musl target, as well as a few places where a type was defined twice in the module tree (leading to the "upper-most" definition being the one exported by the library, in contradiction to the expectation that the "most-specific" definition would be used.

This change moves the definitions of `struct statvfs(64)` and `struct siginfo_t` to be global for all `musl` targets (see https://git.musl-libc.org/cgit/musl/tree/include/sys/statvfs.h and https://git.musl-libc.org/cgit/musl/tree/include/signal.h which are architecture-agnostic headers) and `struct statfs(64)` to be global for all 64-bit `musl` targets (see https://git.musl-libc.org/cgit/musl/tree/arch/generic/bits/statfs.h).  This also required moving `fsblkcnt64_t` and `fsfilcnt64_t` to be musl-wide too (for use in `struct statvfs64`).

It also removes a bunch of redundant (and unreachable) definitions in the `riscv64` and `riscv32` targets (there seems to be a `riscv32` target in the crate, but not in `musl` itself or at least there's no `arch/riscv32` folder in tree).

Upshot of the above is that this change has no externally visible effect, if the more specific types were intended to be used they weren't being so removing them is a no-op.  To actually use more specific type definitions one would need to `cfg` out the general definition as well as providing the specific one.

<details>

<summary>To find most of these issues I used this process</summary>

```
$ for target in $(rustc --print target-list | grep musl)
do
  echo $target
  RUSTDOCFLAGS="--output-format json --document-private-items" cargo +nightly doc -Z build-std=core --target $target
done
$ for json in target/**/doc/libc.json
do
  echo $json
  jq '.index[] | select(.inner | keys[0] == "struct") | .name' $json | sort | uniq -d
done
```

The first command uses rustdoc to create a JSON representation of the API of the crate for each (`musl`) target and the second searches that output for two exported structs of the same name within a single target.  Where there's a duplicate, only one of the two symbols is actually usable (and due to import rules, symbols defined locally take precedence over symbols imported from submodules so the less specific symbol is the one that wins).

You can do similar tests for `enum`, `typedef`, `union`, `constant` by changing the second command in the obvious way, you can also do the same for `function` though you need to additionally filter on `extern "C"` (since e.g. there's many many `clone` functions defined in the crate):

```
$ jq '.index[] | select(.inner | keys[0] == "function") | select(.inner.function.header.abi | (type == "object" and keys[0] == "C"))
| .name' $json | sort | uniq -d
```

</details>

It feels like adding the checks in that methodology to CI for each target would be a good way to catch issues where a more specific definition is masked by a less-specific one.

### DIFF
--- a/src/unix/linux_like/linux/musl/mod.rs
+++ b/src/unix/linux_like/linux/musl/mod.rs
@@ -144,6 +144,26 @@ s! {
         pub sa_restorer: ::Option<extern "C" fn()>,
     }
 
+    // `mips*` targets swap the `s_errno` and `s_code` fields otherwise this struct is
+    // target-agnostic (see https://www.openwall.com/lists/musl/2016/01/27/1/2)
+    pub struct siginfo_t {
+        pub si_signo: ::c_int,
+        #[cfg(not(target_arch = "mips"))]
+        pub si_errno: ::c_int,
+        pub si_code: ::c_int,
+        #[cfg(target_arch = "mips")]
+        pub si_errno: ::c_int,
+        #[doc(hidden)]
+        #[deprecated(
+            since = "0.2.54",
+            note = "Please leave a comment on \
+                  https://github.com/rust-lang/libc/pull/1316 if you're using \
+                  this field"
+        )]
+        pub _pad: [::c_int; 29],
+        _align: [usize; 0],
+    }
+
     pub struct statvfs {
         pub f_bsize: ::c_ulong,
         pub f_frsize: ::c_ulong,
--- a/src/unix/linux_like/linux/musl/mod.rs
+++ b/src/unix/linux_like/linux/musl/mod.rs
@@ -158,9 +158,8 @@ s! {
         #[doc(hidden)]
         #[deprecated(
             since = "0.2.54",
-            note = "Please leave a comment on \
-                  https://github.com/rust-lang/libc/pull/1316 if you're using \
-                  this field"
+            note = "Please leave a comment on https://github.com/rust-lang/libc/pull/1316 \
+                  if you're using this field"
         )]
         pub _pad: [::c_int; 29],
         _align: [usize; 0],
@@ -457,6 +456,40 @@ s! {
         pub tcpi_rcv_ooopack: u32,
         #[cfg(target_arch = "loongarch64")]
         pub tcpi_snd_wnd: u32,
+    }
+
+    // MIPS implementation is special (see mips arch folders)
+    #[cfg(not(target_arch = "mips"))]
+    pub struct statfs {
+        pub f_type: ::c_ulong,
+        pub f_bsize: ::c_ulong,
+        pub f_blocks: ::fsblkcnt_t,
+        pub f_bfree: ::fsblkcnt_t,
+        pub f_bavail: ::fsblkcnt_t,
+        pub f_files: ::fsfilcnt_t,
+        pub f_ffree: ::fsfilcnt_t,
+        pub f_fsid: ::fsid_t,
+        pub f_namelen: ::c_ulong,
+        pub f_frsize: ::c_ulong,
+        pub f_flags: ::c_ulong,
+        pub f_spare: [::c_ulong; 4],
+    }
+
+    // MIPS implementation is special (see mips arch folders)
+    #[cfg(not(target_arch = "mips"))]
+    pub struct statfs64 {
+        pub f_type: ::c_ulong,
+        pub f_bsize: ::c_ulong,
+        pub f_blocks: ::fsblkcnt64_t,
+        pub f_bfree: ::fsblkcnt64_t,
+        pub f_bavail: ::fsblkcnt64_t,
+        pub f_files: ::fsfilcnt64_t,
+        pub f_ffree: ::fsfilcnt64_t,
+        pub f_fsid: ::fsid_t,
+        pub f_namelen: ::c_ulong,
+        pub f_frsize: ::c_ulong,
+        pub f_flags: ::c_ulong,
+        pub f_spare: [::c_ulong; 4],
     }
 }
 
--- a/src/unix/linux_like/linux/musl/mod.rs
+++ b/src/unix/linux_like/linux/musl/mod.rs
@@ -148,6 +148,8 @@ s! {
 
     // `mips*` targets swap the `s_errno` and `s_code` fields otherwise this struct is
     // target-agnostic (see https://www.openwall.com/lists/musl/2016/01/27/1/2)
+    //
+    // FIXME(union): C implementation uses unions
     pub struct siginfo_t {
         pub si_signo: ::c_int,
         #[cfg(not(target_arch = "mips"))]
@@ -177,12 +179,12 @@ s! {
         #[cfg(target_endian = "little")]
         pub f_fsid: ::c_ulong,
         #[cfg(target_pointer_width = "32")]
-        __f_unused: ::c_int,
+        __pad: ::c_int,
         #[cfg(target_endian = "big")]
         pub f_fsid: ::c_ulong,
         pub f_flag: ::c_ulong,
         pub f_namemax: ::c_ulong,
-        __f_spare: [::c_int; 6],
+        __f_reserved: [::c_int; 6],
     }
 
     pub struct statvfs64 {
@@ -197,12 +199,12 @@ s! {
         #[cfg(target_endian = "little")]
         pub f_fsid: ::c_ulong,
         #[cfg(target_pointer_width = "32")]
-        __f_unused: ::c_int,
+        __pad: ::c_int,
         #[cfg(target_endian = "big")]
         pub f_fsid: ::c_ulong,
         pub f_flag: ::c_ulong,
         pub f_namemax: ::c_ulong,
-        __f_spare: [::c_int; 6],
+        __f_reserved: [::c_int; 6],
     }
 
     pub struct termios {
--- a/src/unix/linux_like/linux/musl/mod.rs
+++ b/src/unix/linux_like/linux/musl/mod.rs
@@ -19,7 +19,9 @@ pub type shmatt_t = ::c_ulong;
 pub type msgqnum_t = ::c_ulong;
 pub type msglen_t = ::c_ulong;
 pub type fsblkcnt_t = ::c_ulonglong;
+pub type fsblkcnt64_t = ::c_ulonglong;
 pub type fsfilcnt_t = ::c_ulonglong;
+pub type fsfilcnt64_t = ::c_ulonglong;
 pub type rlim_t = ::c_ulonglong;
 
 cfg_if! {
@@ -173,6 +175,26 @@ s! {
         pub f_files: ::fsfilcnt_t,
         pub f_ffree: ::fsfilcnt_t,
         pub f_favail: ::fsfilcnt_t,
+        #[cfg(target_endian = "little")]
+        pub f_fsid: ::c_ulong,
+        #[cfg(target_pointer_width = "32")]
+        __f_unused: ::c_int,
+        #[cfg(target_endian = "big")]
+        pub f_fsid: ::c_ulong,
+        pub f_flag: ::c_ulong,
+        pub f_namemax: ::c_ulong,
+        __f_spare: [::c_int; 6],
+    }
+
+    pub struct statvfs64 {
+        pub f_bsize: ::c_ulong,
+        pub f_frsize: ::c_ulong,
+        pub f_blocks: ::fsblkcnt64_t,
+        pub f_bfree: ::fsblkcnt64_t,
+        pub f_bavail: ::fsblkcnt64_t,
+        pub f_files: ::fsfilcnt64_t,
+        pub f_ffree: ::fsfilcnt64_t,
+        pub f_favail: ::fsfilcnt64_t,
         #[cfg(target_endian = "little")]
         pub f_fsid: ::c_ulong,
         #[cfg(target_pointer_width = "32")]
--- a/src/unix/linux_like/linux/gnu/b32/riscv32/mod.rs
+++ b/src/unix/linux_like/linux/gnu/b32/riscv32/mod.rs
@@ -70,21 +70,6 @@ s! {
         pub f_spare: [::c_long; 4],
     }
 
-    pub struct statvfs {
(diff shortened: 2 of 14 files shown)


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

bossmc:
Turns out I was both too aggressive and not aggressive enough - every target uses the same `statfs` definition apart from the `mips` targets (that re-arranges the fields) so I've lifted `statfs` to by musl-wide.  OTOH `mips` also re-orders the fields in `siginfo_t` so I've corrected the (global) definition to get that right, I considered adding a `mips`-specific definition but that seemed overkill.

bossmc:
This MR is in danger of going stale so I'd like to resolve the style issue.  However I can't find a way to write the code that meets all the conflicting requirements/intentions:

1. the style check only allows one `s!` per file (so I can't have a `#[cfg(not(mips))] s! { struct statvfs ... }` alongside the common `s!`)
2. The `s!` macro can't have a `cfg_if` nested inside it (it possibly could but it's quite tricky to make work)
3. If you use a `#[cfg(not(mips))]` inside the `s!` macro it removes the struct definition but keeps the `impl Copy`/`impl Clone` which leads to dual-definitions when the real version is added.
4. Moving the `impl Copy` and `impl Clone` blocks to a `#[derive(Copy, Clone)]` combined with (3) works, but the auto-derived `Clone` implementation doesn't use the fact that it's `Copy` to optimize itself to a `memcpy` in debug mode

I'm trying to avoid having 10 identical copies of this type defined simply because `mips` needs a special copy (I'm also baffled why `mips` has so many mips-only layout differences compared to other targets when it's a fairly recent addition to the linux architectures but that ship has long sailed)

tgross35:
Hey @bossmc, sorry this has been stale for so long. If you are interested in rebasing so we can merge this than that would be excellent; if not, no worries.

I do need to note that I don't think this is backportable, meaning we will only get the changes once 1.0 is released (which may still be a few months off). I have to assume this isn't a problem since most of this is refactoring, but just an FYI.

> This MR is in danger of going stale so I'd like to resolve the style issue. However I can't find a way to write the code that meets all the conflicting requirements/intentions:
> 
>     1. the style check only allows one `s!` per file (so I can't have a `#[cfg(not(mips))] s! { struct statvfs ... }` alongside the common `s!`)
> 
>     2. The `s!` macro can't have a `cfg_if` nested inside it (it possibly could but it's quite tricky to make work)
> 
>     3. If you use a `#[cfg(not(mips))]` inside the `s!` macro it removes the struct definition but keeps the `impl Copy`/`impl Clone` which leads to dual-definitions when the real version is added.
> 
>     4. Moving the `impl Copy` and `impl Clone` blocks to a `#[derive(Copy, Clone)]` combined with (3) works, but the auto-derived `Clone` implementation doesn't use the fact that it's `Copy` to optimize itself to a `memcpy` in debug mode
> 
> 
> I'm trying to avoid having 10 identical copies of this type defined simply because `mips` needs a special copy (I'm also baffled why `mips` has so many mips-only layout differences compared to other targets when it's a fairly recent addition to the linux architectures but that ship has long sailed)

Your 3+4 is totally fine here - how well things optimize in debug mode is essentially a nonissue. For cases where it is, it more common for those crates to just [compile dependencies in release mode](https://stackoverflow.com/a/60752451/5380651), so nothing for us to worry about here.

tgross35:
Let me know if you need anything else. Just `@rustbot review` when you are ready.

@rustbot author

bossmc:
Thanks!  I've made that change and rebased to (what was) `main`.

@rustbot review

src/macros.rs
```diff
@@ -82,16 +82,11 @@ macro_rules! s {
...
             #[allow(deprecated)]
             $(#[$attr])*
             pub struct $i { $($field)* }
         }
-        #[allow(deprecated)]
-        impl ::Copy for $i {}
-        #[allow(deprecated)]
-        impl ::Clone for $i {
-            fn clone(&self) -> $i { *self }
-        }
```
tgross35, line 94: These changes are a nice cleanup, but would you mind moving them to a separate PR? To keep different types of changes atomic.

tgross35 (changes requested):
A couple small nits and this lgtm, please rebase and squash.

Regarding your top post - I am entirely in favor of using `rustdoc` JSON output for testing. If you are up to it, please do consider putting something like this in the `libc-test` crate.

bossmc:
Not a problem, I've spun out https://github.com/rust-lang/libc/pull/4038 to separate the two changes.

Once that's in, I'll rebase this change for final review.

tgross35:
Hey, feel free to hold off on rebasing for another day or two. I'm trying to get some larger refactoring in and sync up the branches a bit more before merging anything else.

tgross35:
Alright the big changes are done, rebase at your convenience!

bossmc:
The docker-based tests seem to be failing for networking/DNS reasons?  Otherwise I've rebased, which wasn't too complicated.

@rustbot review

tgross35:
Yeah, it seems like Ubuntu must cut off the repos when they go EOL, which is a bit surprising. I'm trying to fix it in https://github.com/rust-lang/libc/pull/4119.

It looks like the new macro body formatting wants an extra comma, you should be able to just run `./ci/style.sh` locally to apply that all.

bossmc:
Thanks, I don't know if you want to stay on the old ubuntu releases but the packages are saved off to https://old-releases.ubuntu.com/releases/ once a release goes EOL.

I've fixed up the style issue too.

tgross35:
Thanks for the links, we'll need to update at some point so might as well do it now. Though if it takes me more than a couple hours due to the time change I'll probably reconsider :)

src/unix/linux_like/linux/musl/mod.rs
```diff
@@ -164,6 +185,26 @@ s! {
         __f_spare: [::c_int; 6],
     }
 
+    pub struct statvfs64 {
```
tgross35, line 188: Could you add a comment to `statvfs64` and `statfs64` that we always assume largefile64? Since these types seem to be gated by `_LARGEFILE64_SOURCE`.

src/unix/linux_like/linux/musl/mod.rs
```diff
@@ -164,6 +185,26 @@ s! {
...
+        pub f_favail: ::fsfilcnt64_t,
+        #[cfg(target_endian = "little")]
+        pub f_fsid: ::c_ulong,
+        #[cfg(target_pointer_width = "32")]
+        __f_unused: ::c_int,
+        #[cfg(target_endian = "big")]
+        pub f_fsid: ::c_ulong,
+        pub f_flag: ::c_ulong,
+        pub f_namemax: ::c_ulong,
+        __f_spare: [::c_int; 6],
```
tgross35, line 205: Could this be renamed to `__reserved`? https://github.com/kraj/musl/blob/ffb23aef7b5339b8c3234f4c6a93c488dc873919/include/sys/statvfs.h#L27

src/unix/linux_like/linux/musl/mod.rs
```diff
@@ -164,6 +185,26 @@ s! {
...
+        pub f_blocks: ::fsblkcnt64_t,
+        pub f_bfree: ::fsblkcnt64_t,
+        pub f_bavail: ::fsblkcnt64_t,
+        pub f_files: ::fsfilcnt64_t,
+        pub f_ffree: ::fsfilcnt64_t,
+        pub f_favail: ::fsfilcnt64_t,
+        #[cfg(target_endian = "little")]
+        pub f_fsid: ::c_ulong,
+        #[cfg(target_pointer_width = "32")]
+        __f_unused: ::c_int,
```
tgross35, line 200: I don't see this field in the definition at https://github.com/kraj/musl/blob/ffb23aef7b5339b8c3234f4c6a93c488dc873919/include/sys/statvfs.h#L14-L28, any idea where it comes from?

src/unix/linux_like/linux/musl/mod.rs
```diff
@@ -144,6 +146,25 @@ s! {
...
+        #[cfg(target_arch = "mips")]
+        pub si_errno: ::c_int,
+        #[doc(hidden)]
+        #[deprecated(
+            since = "0.2.54",
+            note = "Please leave a comment on https://github.com/rust-lang/libc/pull/1316 \
+                  if you're using this field"
+        )]
+        pub _pad: [::c_int; 29],
+        _align: [usize; 0],
```
tgross35, line 165: You could probably make the struct `#[repr(align(...))]` to get ride of the old hack

src/unix/linux_like/linux/musl/mod.rs
```diff
@@ -144,6 +146,25 @@ s! {
...
+        pub si_code: ::c_int,
+        #[cfg(target_arch = "mips")]
+        pub si_errno: ::c_int,
+        #[doc(hidden)]
+        #[deprecated(
+            since = "0.2.54",
+            note = "Please leave a comment on https://github.com/rust-lang/libc/pull/1316 \
+                  if you're using this field"
+        )]
+        pub _pad: [::c_int; 29],
```
tgross35, line 164: Mind adding a comment `// FIXME(union): C implementation uses unions` so we remember to update this before 1.0?

tgross35 (commented):
I double checked the sources again and noticed a couple small things. Most of this is just for consistency, but I'm not sure about s390x and the 32-bit padding fields.

src/unix/linux_like/linux/musl/mod.rs
bossmc: Good question, I think I pulled it from the `b32/arm/mod.rs` definition.  Looking at pahole on a 32-bit musl libc I can see:

```
struct statvfs {
        long unsigned int          f_bsize;              /*     0     4 */
        long unsigned int          f_frsize;             /*     4     4 */
        fsblkcnt_t                 f_blocks;             /*     8     8 */
        fsblkcnt_t                 f_bfree;              /*    16     8 */
        fsblkcnt_t                 f_bavail;             /*    24     8 */
        fsfilcnt_t                 f_files;              /*    32     8 */
        fsfilcnt_t                 f_ffree;              /*    40     8 */
        fsfilcnt_t                 f_favail;             /*    48     8 */
        long unsigned int          f_fsid;               /*    56     4 */

        /* XXX 4 bytes hole, try to pack */

        /* --- cacheline 1 boundary (64 bytes) --- */
        long unsigned int          f_flag;               /*    64     4 */
        long unsigned int          f_namemax;            /*    68     4 */
        int                        __reserved[6];        /*    72    24 */

        /* size: 96, cachelines: 2, members: 12 */
        /* sum members: 92, holes: 1, sum holes: 4 */
        /* last cacheline: 32 bytes */
};
```

So this `__f_unused` is simply explicit padding on 32-bit systems.  We don't seem to add explicit padding elsewhere so I'll remove this.

src/unix/linux_like/linux/musl/mod.rs
```diff
@@ -415,6 +456,40 @@ s! {
         #[cfg(target_arch = "loongarch64")]
         pub tcpi_snd_wnd: u32,
     }
+
+    // MIPS implementation is special (see mips arch folders)
+    #[cfg(not(target_arch = "mips"))]
```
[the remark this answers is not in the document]
bossmc: The `x32` target is not the intel/AMD 32-bit target (that's in `i386` in musl), it's this https://sites.google.com/site/x32abi/.  Rust supports `gnux32` but doesn't support `muslx32` (yet?) so we don't need to worry about this weird architecture.  The padding fields are necessary on the x32 target since we're passing this structure to a 64-bit kernel so we need to shuffle the 32-bit user-space fields into the appropriate places in the kernel structure.

src/unix/linux_like/linux/musl/mod.rs
bossmc: Ah, it _is_ explicit padding... https://git.musl-libc.org/cgit/musl/tree/include/sys/statvfs.h?h=v1.1.24 has:

```
#if __BYTE_ORDER == __LITTLE_ENDIAN
	unsigned long f_fsid;
	unsigned :8*(2*sizeof(int)-sizeof(long));
#else
	unsigned :8*(2*sizeof(int)-sizeof(long));
	unsigned long f_fsid;
#endif
```

Where those non-standard lines are "anonymous bitfields" that adds padding so the `f_fsid` field is as big as two `int`s.

src/unix/linux_like/linux/musl/mod.rs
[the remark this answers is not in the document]
tgross35: Makes sense, thanks for the clarification!

src/unix/linux_like/linux/musl/mod.rs
tgross35: Fair enough. `align = usize` would be nice, that seems worth proposing at some point.

tgross35 (approved):
Changes lgtm, please squash

bossmc:
Anything in particular you'd like me to squash away?  I don't see the benefit of squashing an entire change into one commit, it just  removes possibly useful context for the change.

If you want to squash the whole change, you can also just do that with the "Squash and Merge" button in Github.

tgross35:
Oh no I certainly didn't mean squash into one (that option isn't available with merge queues anyway), should have just said clean up history. I highly appreciate atomic commits with helpful messages; however, for the sake of my sanity when backporting, there can't be any back and forth changes or fixup commits (e.g. I think there was something that added a `cfg_if!` then changed it back to a `#[cfg]` and something where statfs changes were split around a commit of siginfo changes - that would certainly make my cherry picks confusing :) ).

No worries though, I did some cleanup (added a reordering commit, split each type's refactoring into its own commit, squashed some related changes, and left breadcrumbs for what I changed).

Thank you for all the follow up here!

bossmc:
re: squash - understood, and I like the re-sculpted commits you've done.  I've had other maintainers insist on "squash and merge" for all changes, and I've never liked that policy, so sorry I jumped to a conclusion here!

Nice to get these changes in and thanks for the reviews and support.

PR #3262, opened 2023-06-01
s390x-musl: define O_LARGEFILE constant

nekopsykose:
from https://git.musl-libc.org/cgit/musl/tree/arch/s390x/bits/fcntl.h?id=4b125dd408d54487dc8843b9553502aa0c4167f8#n16

it's a nop as noted in https://github.com/rust-lang/libc/commit/75ac488063b968699c48881b59d525cd46ca95eb, but afaik we should still expose the symbol(?)

### DIFF
--- a/src/unix/linux_like/linux/musl/b64/s390x.rs
+++ b/src/unix/linux_like/linux/musl/b64/s390x.rs
@@ -145,6 +145,7 @@ pub const ETIMEDOUT: ::c_int = 110;
 pub const O_APPEND: ::c_int = 1024;
 pub const O_CREAT: ::c_int = 64;
 pub const O_EXCL: ::c_int = 128;
+pub const O_LARGEFILE: ::c_int = 0x8000;
 pub const O_NONBLOCK: ::c_int = 2048;
 pub const SA_NOCLDWAIT: ::c_int = 2;
 pub const SA_ONSTACK: ::c_int = 0x08000000;


bossmc:
I agree that libc should probably expose the value (so code that checks if it is set in a bitmask works?) but I _think_ it's set to `0x8000` on all linux (64-bit?) architectures so probably can be put in `musl/b64/mod.rs` (or even in `musl/mod.rs`, or even even 'linux/mod.rs`?) and have the rest of the definitions removed (see #3261 for other examples of this).

nekopsykose:
> but I think it's set to 0x8000 on all linux (64-bit?) architectures  

doesn't seem to be (note aarch64, for instance), from musl source:  
```
arch/x32/bits/fcntl.h
16:#define O_LARGEFILE 0100000

arch/powerpc/bits/fcntl.h
16:#define O_LARGEFILE 0200000

arch/s390x/bits/fcntl.h
16:#define O_LARGEFILE 0100000

arch/powerpc64/bits/fcntl.h
16:#define O_LARGEFILE 0200000

arch/mipsn32/bits/fcntl.h
16:#define O_LARGEFILE  020000

arch/m68k/bits/fcntl.h
16:#define O_LARGEFILE 0400000

arch/generic/bits/fcntl.h
16:#define O_LARGEFILE 0100000

arch/aarch64/bits/fcntl.h
16:#define O_LARGEFILE 0400000

arch/mips64/bits/fcntl.h
16:#define O_LARGEFILE  020000

arch/mips/bits/fcntl.h
16:#define O_LARGEFILE  020000

arch/arm/bits/fcntl.h
16:#define O_LARGEFILE 0400000
```

> (see https://github.com/rust-lang/libc/pull/3261 for other examples of this).  

ok, that seems much more thorough as a unification project. perhaps you'd like to also add these constants into that too? although in this case, these don't actually seem consistent for all 32 or all 64 bit widths..

nekopsykose:
> (so code that checks if it is set in a bitmask works?)  

my initial motivation is that rust 1.70 built on s390x musl now fails without this exposed. i'm not exactly sure why, as the failure is from `nix` which hasn't changed in forever (this constant is checked [here](https://github.com/nix-rust/nix/blob/c6f9e2332efcf62c751d7a0174bb791e732b90a8/src/fcntl.rs#L117) )

perhaps because of some doc feature.. but it indeed is harmless to expose, so hotpatching it for vendor/libc worked fine.

bossmc:
Huh, there's more divergence than I expected, don't worry about the unification then!

nekopsykose:
no worries, thanks for the review! always nice to see people working on these interfaces too :)

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #3266, opened 2023-06-06
Update crate version to 0.2.146

nikarh:
Contains a fix #3265 for previous yanked release


nikarh:
Looks like the pipeline failed to download nodejs due to a network error

Amanieu:
Published

PR #3268, opened 2023-06-07
add mips32r6 and mips64r6 target_arch detection

chenx97:
This PR introduces `"mips32r6"` and `"mips64r6"` as valid `target_arch` values, and would be the arch values used by Tier-3 targets `mipsisa32r6-unknown-linux-gnu`, `mipsisa32r6el-unknown-linux-gnu`, `mipsisa64r6-unknown-linux-gnuabi64` and `mipsisa64r6el-unknown-linux-gnuabi64`.

This PR is made for https://github.com/rust-lang/rust/pull/112374.

This PR has a corresponding [MCP](https://github.com/rust-lang/compiler-team/issues/632) waiting for review.

mips64r6 has the same ABIs, syscall conventions, and various magic numbers as mips64 does. So this PR just adds mips64r6 to all mips64 detection macros. Similarly, mips32r6 share their ABIs, syscall conventions and various magic with mips, so the PR just adds mips32r6 to where mips is to be detected.


semarie:
I think the modification of `src/unix/bsd/netbsdlike/openbsd/mod.rs` isn't expected. It is modifying an OpenBSD part and not a Linux part. It shouldn't break anything, but it would be confusing.

chenx97:
> I think the modification of `src/unix/bsd/netbsdlike/openbsd/mod.rs` isn't expected. It is modifying an OpenBSD part and not a Linux part. It shouldn't break anything, but it would be confusing.

Changes reverted

semarie:
thanks

chenx97:
Commits of this PR have been changed to only affect GNU/Linux.

JohnTitor:
I'm going to r+ once MCP passes.

chenx97:
> I'm going to r+ once MCP passes.

It just passed.

chenx97:
what's happening to the freebsd CI here?

JohnTitor (approved):

PR #3274, opened 2023-06-15
add MAP_UNINITIALIZED flag (Linux-only) for mmap

raymontag:
MAP_UNINITIALIZED is a Linux-only flag for mmap that is generically defined in mmap-common.h: https://elixir.bootlin.com/linux/latest/source/include/uapi/asm-generic/mman-common.h#L33

### DIFF
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -1290,6 +1290,7 @@ pub const MAP_NORESERVE: ::c_int = 0x04000;
 pub const MAP_POPULATE: ::c_int = 0x08000;
 pub const MAP_NONBLOCK: ::c_int = 0x010000;
 pub const MAP_STACK: ::c_int = 0x020000;
+pub const MAP_UNINITIALIZED: ::c_int = 0x4000000;
 
 pub const EDEADLK: ::c_int = 35;
 pub const ENAMETOOLONG: ::c_int = 36;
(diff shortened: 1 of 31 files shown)


JohnTitor (approved):

JohnTitor:
Seems the musl header here is old and doesn't know these items, we have to ignore them on the test. See this for example: https://github.com/rust-lang/libc/pull/3278#issuecomment-1606100074

JohnTitor:
Ping from triage: @raymontag Could you check the above comment if you have time? Thanks!

JohnTitor:
Closing as inactive, feel free to re-submit if you're still interested in this change. Thank you for the PR anyway!

SteveLauC:
From the source code of glibc:

> sysdeps/unix/sysv/linux/tst-mman-consts.py

```
        # A series of MAP_HUGE_<size> macros are defined by the kernel
        # but not by glibc.  MAP_UNINITIALIZED is kernel-only.
        # MAP_FAILED is not a MAP_* flag and is glibc-only, as is the
        # MAP_ANON alias for MAP_ANONYMOUS.  MAP_RENAME, MAP_AUTOGROW,
        # MAP_LOCAL and MAP_AUTORSRV are in the kernel header for
        # MIPS, marked as "not used by linux"; SPARC has MAP_INHERIT
        # in the kernel header, but does not use it.  The kernel
        # header for HPPA removed a define of MAP_VARIABLE to 0 in
        # Linux 6.2.
```

> MAP_UNINITIALIZED is kernel-only.

which seems to mean that this constant is defined in the kernel, not libcs. And there is a [patch](https://patchwork.kernel.org/project/linux-fbdev/patch/1441832902-28993-5-git-send-email-palmer@dabbelt.com/) which also seems to prove it.

If so, why we are defining this constants across `{libc}/{arch}/mod.rs` but not `linux/mod.rs`

Issue #3160, opened 2023-03-21
test failure: incomplete type for struct af_alg_iv


tones111:
When attempting to run the test suite I get an error that struct af_alg_iv is not defined.

```
$ cd libc-test && cargo test
<snip>
  cargo:warning=/home/paul/src/libc/target/debug/build/libc-test-3ad7e5dad55ad485/out/main.c: In function ‘__test_fsize_af_alg_iv_iv’:
  cargo:warning=/home/paul/src/libc/target/debug/build/libc-test-3ad7e5dad55ad485/out/main.c:25567:34: error: invalid application of ‘sizeof’ to incomplete type ‘__u8[]’ {aka ‘unsigned char[]’}
  cargo:warning=25567 |                     return sizeof(foo->iv);
  cargo:warning=      |                                  ^
```

The offending lines in the test are
```
                  uint64_t __test_fsize_af_alg_iv_iv(void) {
                     struct af_alg_iv* foo = NULL;
                     return sizeof(foo->iv);
                 }
```
which I don't believe can ever work since dereferencing a NULL pointer is undefined behavior.

My /usr/include/linux/if_alg.h defines it as
`struct af_alg_iv {
    __u32 ivlen;
    __u8  iv[];
};`

Compiler explorer shows the [same error](https://godbolt.org/z/P5vno7n77) when trying to implement an equivalent call.  Is this a case of main.c not getting generated correctly?

system information:
```
$ cargo --version -v
cargo 1.68.0 (115f34552 2023-02-26)
release: 1.68.0
commit-hash: 115f34552518a2f9b96d740192addbac1271e7e6
commit-date: 2023-02-26
host: x86_64-unknown-linux-gnu
libgit2: 1.5.0 (sys:0.16.0 vendored)
libcurl: 7.86.0-DEV (sys:0.4.59+curl-7.86.0 vendored ssl:OpenSSL/1.1.1q)
os: Arch Linux [64-bit]
```

kernel: v6.2.7'
gcc: 12.2.1

PR #3275, opened 2023-06-15
Define additional socket ioctls

leftmostcat:
Define several socket-related `ioctl`s which aren't currently captured. Updated to include `ioctl`s from BSD systems.

`libc-test` suite has not been run due to https://github.com/rust-lang/libc/issues/3160.


JohnTitor (approved):

JohnTitor:
Seems the musl header here is old and doesn't know these items, we have to ignore them on the test. See this for example: https://github.com/rust-lang/libc/pull/3278#issuecomment-1606100074

JohnTitor:
Friendly-ping @leftmostcat, could you check the above comment? Thanks!

JohnTitor:

PR #3277, opened 2023-06-16
android: Update linux/if_link.h constants

dragan-cecavac-nordsec:
Update Android `libc/kernel/uapi/linux/if_link.h` IFLA entries.
Reference: https://android.googlesource.com/platform/bionic/+/refs/heads/master/libc/kernel/uapi/linux/if_link.h#159

PR #3278, opened 2023-06-16
android: Add kernel/uapi/linux/neighbour.h constants

dragan-cecavac-nordsec:
Introduce Android constants defined in https://android.googlesource.com/platform/bionic/+/refs/heads/master/libc/kernel/uapi/linux/neighbour.h


JohnTitor (approved):
Thanks!

dragan-cecavac-nordsec:
Hi,
It seems that the failure is the same as with https://github.com/rust-lang/libc/pull/3279#issuecomment-1600505781 so I'll also just wait and hope for the best 🤞

JohnTitor:
Yeah, let's see if it's now fixed, @bors retry

JohnTitor:
s390x CI needs a tweak, let me fix.

dragan-cecavac-nordsec:
It looks to me like some of the definitions are too new for the test setup 😅
Referring to master vs Android 11:
- https://android.googlesource.com/platform/bionic/+/refs/heads/master/libc/kernel/uapi/linux/neighbour.h
- https://android.googlesource.com/platform/bionic/+/refs/tags/android-11.0.0_r1/libc/kernel/uapi/linux/neighbour.h

For starters I could reduce the scope of the changes to cover the testable ones and further on we could see what could be done to widen the test scope, if you agree?

JohnTitor:
You can ignore tests for newer items, see https://github.com/rust-lang/libc/blob/1e8943d432c3c3012eaacb091e07b88d2ca068d6/libc-test/build.rs#L1743

A fixme comment describing available version(s) would be helpful when upgrading CI toolchains :)

dragan-cecavac-nordsec:
Thank you, hopefully this should do.

When it comes to NDK versions listed under `FIXME` I simply checked the most recent sub-version. E.g. `r22b` in case of `r22` and `r25c` in case of `r25`.
After the upgrade, this PR will conflict with https://github.com/rust-lang/libc/pull/3277. Whichever gets merged first, I can adjust the other.

JohnTitor:
Thanks! @bors r+

PR #3282, opened 2023-06-25
Fix ABI compatibility with Emscripten >= 3.1.42

kleisauke:
See:
https://github.com/emscripten-core/emscripten/pull/19569#issuecomment-1605440414

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -70,14 +70,8 @@ fn main() {
         Some(_) | None => set_cfg("freebsd11"),
     }
 
-    match emcc_version() {
-        Some((major, minor, patch))
-            if (major > 3)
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

JohnTitor:
Seems we have to early return if rustc version is too old (I think it's fine as the usage rate should be ultimately low).

kleisauke:
I think commit 92db93cf9e756d9ac51115acaa369ebc0cbff83f would fix the build on older Rust versions.

JohnTitor:
Seems good 👍, @bors r+

PR #3284, opened 2023-06-25
Add missing PS Vita definitions, fix some unused ones

pheki:
This PR improves vita's newlib support for std by adding some missing definitions and fixing some duplicated ones (e.g. `EAI_NONAME` from `src/unix/newlib/vita/mod.rs` was not being used as it was already defined on `src/unix/newlib/mod.rs`)

Previous work: #3209 #3255 

cc @nikarh

### DIFF
--- a/src/unix/newlib/mod.rs
+++ b/src/unix/newlib/mod.rs
@@ -787,13 +787,6 @@ cfg_if! {
 }
 
 cfg_if! {
-    if #[cfg(not(target_os = "vita"))] {
-        mod dirent;
-        pub use self::dirent::*;
-    }
-}
-
-cfg_if! {
     if #[cfg(libc_align)] {
         #[macro_use]
         mod align;
--- a/src/unix/newlib/mod.rs
+++ b/src/unix/newlib/mod.rs
@@ -1,13 +1,7 @@
 pub type blkcnt_t = i32;
 pub type blksize_t = i32;
 
-cfg_if! {
-    if #[cfg(target_os = "vita")] {
-        pub type clockid_t = ::c_uint;
-    } else {
-        pub type clockid_t = ::c_ulong;
-    }
-}
+pub type clockid_t = ::c_ulong;
 
 cfg_if! {
     if #[cfg(any(target_os = "espidf"))] {
@@ -168,16 +162,6 @@ s! {
         pub sa_handler: extern fn(arg1: ::c_int),
         pub sa_mask: sigset_t,
         pub sa_flags: ::c_int,
-    }
-
-    pub struct dirent {
-        #[cfg(not(target_os = "vita"))]
-        pub d_ino: ino_t,
-        #[cfg(not(target_os = "vita"))]
-        pub d_type: ::c_uchar,
-        #[cfg(target_os = "vita")]
-        __offset: [u8; 88],
-        pub d_name: [::c_char; 256usize],
     }
 
     pub struct stack_t {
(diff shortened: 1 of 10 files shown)


src/unix/newlib/mod.rs
```diff
@@ -546,8 +546,16 @@ pub const IFF_LINK2: ::c_int = 0x4000; // per link layer defined bit
 pub const IFF_ALTPHYS: ::c_int = IFF_LINK2; // use alternate physical connection
 pub const IFF_MULTICAST: ::c_int = 0x8000; // supports multicast
 
-pub const TCP_NODELAY: ::c_int = 8193;
-pub const TCP_MAXSEG: ::c_int = 8194;
+cfg_if! {
+    if #[cfg(target_os = "horizon")] {
```
nikarh, line 550: This should be vita, horizon is for 3ds/switch

nikarh (changes requested):
Checked every const/struct/function, apart from horizon macro everything looks correct

src/unix/newlib/mod.rs
pheki: Nice catch!

pheki:
TYSM @nikarh, just fixed that cfg

nikarh (approved):

JohnTitor (approved):

pheki:
Looks like https://github.com/rust-lang/homu/issues/75, should we just retry?

JohnTitor:
Sure, @bors retry

RalfJung:
Is this the PR that is completely halting bors (across RA, miri, libc, ...) right now? Could someone try closing and re-opening it?

RalfJung:
Bummer, this wasn't it seems. :(
We'll have to wait for infra then I guess.

Cc https://github.com/rust-lang/homu/issues/191

src/unix/newlib/mod.rs
```diff
@@ -598,10 +618,19 @@ pub const NI_NAMEREQD: ::c_int = 4;
...
 
-pub const EAI_FAMILY: ::c_int = -303;
-pub const EAI_MEMORY: ::c_int = -304;
-pub const EAI_NONAME: ::c_int = -305;
-pub const EAI_SOCKTYPE: ::c_int = -307;
+cfg_if! {
+    if #[cfg(target_os = "vita")] {
+        pub const EAI_FAMILY: ::c_int = -6;
+        pub const EAI_MEMORY: ::c_int = -10;
+        pub const EAI_NONAME: ::c_int = -2;
```
riking, line 625: The ordering here doesn't really make sense after this patch... maybe this could be `pub use vita::{EAI_NONAME, ...};` ?

pheki: I like that idea, I'll change it

pheki: Just did it

We can't add `pub use` as it goes against the library styling ([pub use must go after all constants](https://github.com/rust-lang/libc/blob/main/ci/style.rs)), but there's already a `pub use self::vita::*` at the end.

pheki:
@JohnTitor can I get a new review on this one? Rustbot had bugged and then I did a small change (@riking's suggestion)

nikarh:
This PR was updated a number of times after it was last reviewed (remaining structures and signatures fixed) and at this point, we aren't planning on adding more changes to this PR.

Also, it is a pre-requisite for rust-lang/socket2/pull/465.

JohnTitor:

pheki:

Done!

JohnTitor:
Thank you! @bors r+

pheki:
Maybe we need to retry? Looks like it failed to start one BSD test on cirrus

JohnTitor:
Seems so, @bors retry

nikarh:
Problem was with `dirent` struct defines in `dirent` module.
Moved the shared `dirent` to `newlib/generic.rs` and added reexports to all targets.

Should I squash the fix to the main commit?

JohnTitor:
If commits are too many, it'd be great to be squashed, but 2 commits are fine by me, thank you!
@bors r+

PR #3286, opened 2023-06-28
android 32 bits fix stat struct proposal.

### LINKED ISSUE #3285: android: Can we use proper `libc` typedefs in `stat` definition?
If we look at the `stat` type's definition for different platforms, we see:
- [x86_64-unknown-linux-gnu](https://rust-lang.github.io/libc/x86_64-unknown-linux-gnu/doc/libc/struct.stat.html):
```rust
st_dev: dev_t
st_ino: ino_t
[...]
```
- [aarch64-linux-android](https://rust-lang.github.io/libc/aarch64-linux-android/doc/libc/struct.stat.html):
  - same

But for [arm-linux-androideabi](https://rust-lang.github.io/libc/arm-linux-androideabi/doc/libc/struct.stat.html) we stop using `libc` typedefs and instead have:
```rust
st_dev: c_ulonglong
st_mode: c_uint
[...]
st_ino: c_ulonglong
```

Is there a (good) reason for this divergence? If I look at [`ino_t`](https://rust-lang.github.io/libc/arm-linux-androideabi/doc/libc/type.ino_t.html) it's actually defined to be a `c_ulong`, not `c_ulonglong`, so that will lead the mismatches all over the place when the typedefs are used. Is this mandated somewhere, a bug in the platform definition that we need to support? This is a nightmare to support in user code.

devnexen:
close #3285


JohnTitor (approved):

tgross35:
Looks like this doesn't agree with the `libc` we are checking against. I think @danielocfb was saying that we should use e.g. `dev_t` rather than `c_ulong`

@rustbot author

tgross35:
@maurer would you mind looking at this? This would be a breaking change to `stat` for libc 1.0.

maurer:
This change is incorrect - the bionic [stat definition](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/sys/stat.h;l=83?q=%22struct%20stat%22%20bionic&ss=android%2Fplatform%2Fsuperproject%2Fmain) looks like the left hand side for 32-bit platforms.

I can comment over on the issue, but the reason this one is different is because the actual value returned here is *not* an `ino_t`. Even outside `bionic` (which you should use as the source of truth for what Android `libc` structures look like), 32-bit ARM `musl` has a [similar header](https://cs.android.com/android/platform/superproject/main/+/main:external/trusty/musl/arch/arm/bits/stat.h?q=%22struct%20stat%22%20musl&ss=android%2Fplatform%2Fsuperproject%2Fmain) which makes the fact that this field is truncated more explicit.

It looks like there may be a couple places where using the "right type" is plausible (e.g. musl thinks that using `dev_t` for the first one could work), and I could talk to the bionic maintainer about making this switch if it would be helpful, but the proposed change would actually just flat out work incorrectly, the new proposed structure has the wrong offsets.

tgross35:
It's been a while, @devnexen would you be able to make this closer match the Android definitions as the maintainer requested? Looks like `stat64` could also be made an alias since they have identical definitions (this may have already happened).

tgross35:
Seems fine to me now, @maurer could you take a look?

See also https://github.com/rust-lang/libc/pull/5069 for changing to the `timespec` types eventually

tgross35:
What's the status of 64-bit time on 32-bit Android actually, is that a supported configuration? Based on https://cs.android.com/android/platform/superproject/+/android-latest-release:bionic/libc/include/sys/types.h;l=97-99;drc=0d34bb1a4ce70114d96ca50aab8012b571d4e7c0 and https://cs.android.com/android/platform/superproject/+/android-latest-release:external/kernel-headers/original/uapi/asm-generic/posix_types.h;l=92;drc=f2abbe6d3587e83f95b5fa7f9da6e4a1f981ad38, I don't see any indication that it is.

Cc @dybucc since you may know or be interested in the answer.

maurer:
Looks fine now.

maurer (approved):

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/3286/023c0a3c184c2d7fcd528b64e867075660a5884d..5987c6262f24c8b9536965e6a6af738e5d88795b)*

dybucc:
>What's the status of 64-bit time on 32-bit Android actually, is that a supported configuration? Based on https://cs.android.com/android/platform/superproject/+/android-latest-release:bionic/libc/include/sys/types.h;l=97-99;drc=0d34bb1a4ce70114d96ca50aab8012b571d4e7c0 and https://cs.android.com/android/platform/superproject/+/android-latest-release:external/kernel-headers/original/uapi/asm-generic/posix_types.h;l=92;drc=f2abbe6d3587e83f95b5fa7f9da6e4a1f981ad38, I don't see any indication that it is.
>
>Cc @dybucc since you may know or be interested in the answer.

What little I know comes from this document [^1]. There's also info there on
32-bit file offset types.

[^1]: <https://android.googlesource.com/platform/bionic/+/main/docs/32-bit-abi.md#is-32_bit-on-lp32-y2038>

tgross35:
> Linux 5.x kernels do offer extra interfaces so that 32-bit processes can pass 64-bit times to/from the kernel, but we do not plan on adding support for these to the C library

Sounds like there's nothing for us to support then - that makes it easy.

PR #3287, opened 2023-06-30
linux: add a few kTLS defintions

gennyble:
related: https://github.com/nix-rust/nix/pull/2065

i think android has support, too, but if I put it in `linux-like` does that imply emscripten support?


fasterthanlime:
Thanks @gennyble for opening this!

I need this PR + a point release to (ultimately) land https://github.com/nix-rust/nix/pull/2065 which in turn unblocks doing in-kernel TLS properly from rust with tokio, cf. https://github.com/hapsoc/ktls/pull/26

(I know, exciting!)

gennyble:
`TLS_GET_RECORD_TYPE` is from a kernel header which might not be available where libc is built. What would you like me to do here?

I see [a skip](https://github.com/rust-lang/libc/blob/main/libc-test/build.rs#L3513) might be an option here?

fasterthanlime:
@JohnTitor would it be possible to get a look at this? Maybe from another maintainer, if you're not able to?

JohnTitor:
> `TLS_GET_RECORD_TYPE` is from a kernel header which might not be available where libc is built. What would you like me to do here?
> 
> I see [a skip](https://github.com/rust-lang/libc/blob/main/libc-test/build.rs?rgh-link-date=2023-06-30T19%3A54%3A42Z#L3513) might be an option?

Correct.

gennyble:
Instead of skipping I added the header to the linux kernel headers for libc-test.

@rustbot review

JohnTitor (approved):

JohnTitor:
Cool, thanks! @bors r+

fasterthanlime:
Thanks for getting this merged! When can we expect this to be included in a crates.io release of libc ? (So we can bump the version nix relies on)

fasterthanlime:
@JohnTitor the latest libc release was cut on June 25, 2023 - I think this means this PR still isn't on crates.io? The problem is we're trying to use those new additions from https://lib.rs/crates/nix, so a git dependency isn't really an option here.

PR #3291, opened 2023-07-05
For NetBSD: add entry for NetBSD/riscv64.

he32:


he32:
I know the code to support NetBSD/riscv64 isn't yet in rust itself; a request with the required changes will be submitted shortly.  However, since this is its own repository, I need to request this separately, and one of them has to go in first.

he32:
THe NetBSD/riscv64 target was merged into rust recently, ref. https://github.com/rust-lang/rust/pull/113356

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
```diff
@@ -0,0 +1,24 @@
...
+    if #[cfg(libc_const_size_of)] {
+        #[doc(hidden)]
+        pub const _ALIGNBYTES: usize = ::mem::size_of::<::c_long>() - 1;
+    } else {
+        #[doc(hidden)]
+        pub const _ALIGNBYTES: usize = 8 - 1;
+    }
+}
+
+// pub const _MAX_PAGE_SHIFT: u32 = 12;
```
JohnTitor, line 19: Why is this commented out? If it's a left-over, please remove it to avoid confusion (note that many contributors touch codebase).

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
```diff
@@ -0,0 +1,24 @@
+use PT_FIRSTMACH;
+
+pub type c_long = i64;
+pub type c_ulong = u64;
+pub type c_char = u8;
+pub type __cpu_simple_lock_nv_t = ::c_int;
+
+// should be pub(crate), but that requires Rust 1.18.0
```
JohnTitor, line 8: You can use cfgs if it should be gated, see https://github.com/rust-lang/libc/blob/main/build.rs

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
he32: Hm, yes, this is a left-over.  I've committed an entry removing this possibly-confusing line.

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
he32: Hm, I'll admit this is copy-pasted from the origin.  That said, this particular comment and construct seems to be prevalent in both freebsd and openbsd target descriptions, and the other netbsd descriptions as well.
And ... I'll admit that it's not self-evident to me how *exactly* to do what the comment implies.

JohnTitor: Ah, I see! Could you remove them then?

he32: I removed the comment from the riscv64.rs description.
Do you want me to remove it from all the other NetBSD descriptions as part of this pull request?

he32: @JohnTitor Maybe I don't quite understand what you are asking.  Can the tidy-up of the conditional in the various netbsd *.rs files please be postponed to a separate pull request?  And can this particular pull request please be applied?  libc is vendored "like crazy", and this is hurting portability for this particular platform.

JohnTitor:
Could you squash commits into one then? We can go ahead once it's done.

he32:
Ping?  I don't think the failed freebsd builds has anything to do with the suggested change.

JohnTitor (approved):

JohnTitor:
Yeah, should be, thanks! @bors r+

PR #3304, opened 2023-07-17
Add secure_getenv function for Linux variants

CuriouslyCurious:
This adds the [`secure_getenv`](https://www.gnu.org/software/libc/manual/html_node/Environment-Access.html#index-secure_005fgetenv) function for Linux variants as it is supported in both GNU libc (2.17+) and musl ([1.1.24+](https://git.musl-libc.org/cgit/musl/commit/?h=v1.1.24&id=ea9525c8bcf6170df59364c4bcd616de1acf8703)).

It is recommended to be used over the regular `getenv`, and there is a [standing issue](https://github.com/rust-lang/rust/issues/92558) for the Rust library to use it instead.

I can't properly test it on my machine as there is some funky stuff going on with musl. I'll open up a [separate issue](https://github.com/rust-lang/libc/issues/3305) with the output as it doesn't concern this fairly simple change.

### DIFF
--- a/libc-test/semver/linux.txt
+++ b/libc-test/semver/linux.txt
@@ -3413,6 +3413,7 @@ sched_setparam
 sched_setscheduler
 sctp_assoc_t
 seccomp_data
+secure_getenv
 seed48
 seekdir
 sem_close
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):

JohnTitor:
Failed on asmjs:
```
   = note: error: undefined symbol: secure_getenv (referenced by top-level compiled C/C++ code)
          warning: Link with `-sLLD_REPORT_UNDEFINED` to get more information on undefined symbols
          warning: To disable errors for undefined symbols use `-sERROR_ON_UNDEFINED_SYMBOLS=0`
          warning: _secure_getenv may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
          Error: Aborting compilation due to previous errors
          emcc: error: '/emsdk-portable/node/16.20.0_64bit/bin/node /emsdk-portable/upstream/emscripten/src/compiler.js /tmp/tmp_0irvcao.json' failed (returned 1)
```

@rustbot author

PR #3307, opened 2023-07-21
upgrade to emscripten 3.1.41

### LINKED ISSUE #3306: emscripten time_t has wrong size
`time_t` is 64bit on emscripten since https://github.com/emscripten-core/emscripten/pull/17401
Currently the undersized struct definition here truncates the last 4 bytes of the `timespec`, I believe, so that `Instant::now()` etc. end up having no sub-second resolution on `wasm32-unknown-emscripten`.

Filed these issues with more details before I landed here:
https://github.com/emscripten-core/emscripten/issues/19872
https://github.com/rust-lang/rust/issues/113852

It is curious to note that `time_t` was only 32-bit for a short while, starting with https://github.com/emscripten-core/emscripten/pull/16966 (edit: according the comment in that PR, the size on wasm32 stayed the same at that point, so I presume it was indeed 64-bit for the first time from 17401)
~~although it has been 32-bit in this repo for 6 years. I am not seeing the same issue before 16966 that I am seeing after 17401, however. So there is probably another glitch or two that either masked the issue before 16966 or (in between 16966 and 17401) enabled the issue to surface as soon as the size became wrong again.~~

Cheers!

benjamin-sieffert:
- Increase size of `time_t` to 64-bit
- Update values of a number of constants
- Remove tests for a lot of headers that have been removed from emscripten
- Bump emscripten version in test environment from `3.1.20` to `3.1.41` and use the recommended nodeJS version

Fixes #3306


JohnTitor:
We have to confirm that this change doesn't negatively affect std, marked as blocked until it's resolved.

ci/emscripten-entry.sh
```diff
@@ -7,6 +7,6 @@ source /emsdk-portable/emsdk_env.sh &> /dev/null
 
 # emsdk-portable provides a node binary, but we need version 8 to run wasm
 # NOTE: Do not forget to sync Node.js version with `emscripten.sh`!
-export PATH="/node-v14.17.0-linux-x64/bin:$PATH"
+export PATH="/node-v16.20.0-linux-x64/bin:$PATH"
```
kleisauke, line 10: This change can be omitted after PR #3321 lands.

ci/emscripten.sh
```diff
@@ -24,5 +22,5 @@ chmod a+rxw -R /emsdk-portable
 # node 8 is required to run wasm
 # NOTE: Do not forget to sync Node.js version with `emscripten-entry.sh`!
 cd /
-curl --retry 5 -L https://nodejs.org/dist/v14.17.0/node-v14.17.0-linux-x64.tar.xz | \
+curl --retry 5 -L https://nodejs.org/dist/v16.20.0/node-v16.20.0-linux-x64.tar.xz | \
```
kleisauke, line 25: ditto.

ci/emscripten-entry.sh
benjamin-sieffert: Thanks! Adjusted

libc-test/build.rs
```diff
@@ -2744,21 +2737,45 @@ fn test_emscripten(target: &str) {
...
-            | "F_GETLK" // to 5
-            | "F_SETLK" // to 6
-            | "F_SETLKW" // to 7
-            | "O_TMPFILE" // to 65
-            | "SIG_IGN" // -1
-                => true,
+            // No personality.h etc.: https://github.com/emscripten-core/emscripten/pull/17704
+            // No sysctl.h: https://github.com/emscripten-core/emscripten/pull/18863
+            n if n.starts_with("PTRACE_") => true,
+            n if n.starts_with("QIF_") => true,
```
JohnTitor, line 2766: You should just remove the declaration instead of skipping.

JohnTitor:
It'd also be nice if you could rebase instead of merging and squashing commits into one.

benjamin-sieffert:
Sorry, I guess this is a bit over my head. I am too unfamiliar with all the details of what exactly is going on in the tests.
I tried moving this PR even further up to 3.1.44 and ran into issues I do not understand at all.
My motivation for this PR was fixing https://github.com/rust-lang/libc/issues/3306 which causes `Instant::now` to have no sub-second precision on emscripten. (There is also an issue with certain `std::fs` functions crashing because of 32/64 bit mismatch on types, but I am not sure if this PR would have fixed that.)
For my use-case I have solved the timing issue by using `emscripten_get_now` instead.
I will abandon this PR and leave it to more involved devs to eventually figure out this emscripten/libc update.

PR #3309, opened 2023-07-27
Add missing targets for documentation and remove duplicates for list

GuillaumeGomez:
I just noticed that the apple targets were not documented on docs.rs so I added them.

I also saw some duplicates in the list. Do you want me to add a check to ensure that there are no duplicates?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ Raw FFI bindings to platform libraries like libc.
 features = ["const-extern-fn", "extra_traits"]
 default-target = "x86_64-unknown-linux-gnu"
 targets = [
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):

JohnTitor:
> Do you want me to add a check to ensure that there are no duplicates?

I think the priority is low as the list wouldn't be updated so frequently. Let's add that if the same error happens in the future!

@bors r+

GuillaumeGomez:
> I think the priority is low as the list wouldn't be updated so frequently. Let's add that if the same error happens in the future!

Noted! Don't hesitate to tell me if it needs to be added, I'll gladly do it.

PR #3310, opened 2023-07-30
Don't include flock support on Solaris

inferiorhumanorgans:
This should disable the flock wrapper and related consts on Solaris, but leave them in place for OpenSolaris derivatives like Illumos which do have a BSD-esque flock.

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -1067,7 +1067,11 @@ extern "C" {
     )]
     pub fn realpath(pathname: *const ::c_char, resolved: *mut ::c_char) -> *mut ::c_char;
 
-    pub fn flock(fd: ::c_int, operation: ::c_int) -> ::c_int;
+    cfg_if! {
+        if #[cfg(not(target_os = "solaris"))] {
+            pub fn flock(fd: ::c_int, operation: ::c_int) -> ::c_int;
+        }
+    }
 
     #[cfg_attr(target_os = "netbsd", link_name = "__times13")]
     pub fn times(buf: *mut ::tms) -> ::clock_t;
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -1915,10 +1915,16 @@ pub const SHUT_RD: ::c_int = 0;
 pub const SHUT_WR: ::c_int = 1;
 pub const SHUT_RDWR: ::c_int = 2;
 
-pub const LOCK_SH: ::c_int = 1;
-pub const LOCK_EX: ::c_int = 2;
-pub const LOCK_NB: ::c_int = 4;
-pub const LOCK_UN: ::c_int = 8;
+// These don't exist on Solaris, but do on OpenSolaris derivatives
+// e.g. https://illumos.org/man/3C/flock
+cfg_if! {
+    if #[cfg(not(target_os = "solaris"))] {
+        pub const LOCK_SH: ::c_int = 1;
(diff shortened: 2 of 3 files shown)


src/unix/solarish/mod.rs
```diff
@@ -1915,10 +1915,16 @@ pub const SHUT_RD: ::c_int = 0;
 pub const SHUT_WR: ::c_int = 1;
 pub const SHUT_RDWR: ::c_int = 2;
 
-pub const LOCK_SH: ::c_int = 1;
-pub const LOCK_EX: ::c_int = 2;
-pub const LOCK_NB: ::c_int = 4;
-pub const LOCK_UN: ::c_int = 8;
+// These don't exist on Solaris, but do on OpenSolaris derivatives
+// e.g. https://illumos.org/man/3C/flock
+cfg_if! {
```
jclulow, line 1920: If something is illumos specific, I suspect we ought to move it to `illumos.rs` rather than introduce more conditional compilation directives?

src/unix/mod.rs
```diff
@@ -1067,7 +1067,11 @@ extern "C" {
     )]
     pub fn realpath(pathname: *const ::c_char, resolved: *mut ::c_char) -> *mut ::c_char;
 
-    pub fn flock(fd: ::c_int, operation: ::c_int) -> ::c_int;
+    cfg_if! {
```
jclulow, line 1070: I think we usually use `#[cfg_attr]` rather than `cfg_if` where possible.  It looks like that in this file at least.

jclulow (changes requested):

src/unix/mod.rs
inferiorhumanorgans: Yeah, clippy flagged it on the first run so I changed to `cfg_if!`.  I've no strong feelings either way.

src/unix/solarish/mod.rs
inferiorhumanorgans: I don't know how much Illumos diverges from Solaris.  If it's a matter of a few things being added to Illumos then I'd think some directives wouldn't hurt.  If it's a matter of things being changed wholesale or a lot of new features being added then, yeah, I'm inclined to agree that a new module is the way to go.

I've no strong feelings either way but I don't have an Illumos install to check against.

jclulow: The divergence only increases over time.  Oracle made a proprietary fork of OpenSolaris more than a decade ago, and we made an open one at the same time.  We've been diverging ever since.

Regardless, I believe the module file already exists, alongside this file.

inferiorhumanorgans: With those constants moved into `illumos.rs` it looks like I've 1.70.0 building successfully for Solaris 11.4 (x86+SPARC) except for clippy. 🎉

JohnTitor (approved):

inferiorhumanorgans:
```
  error: macros cannot expand to foreign items
      --> ../src/unix/mod.rs:1070:5
       |
  1070 | /     cfg_if! {
  1071 | |         if #[cfg(not(target_os = "solaris"))] {
  1072 | |             pub fn flock(fd: ::c_int, operation: ::c_int) -> ::c_int;
  1073 | |         }
  1074 | |     }
       | |_____^
```

So is the answer to pull this out of a `cfg_if` block?

JohnTitor:
Yeah, that should resolve the error.
@bors r-
@rustbot author

JohnTitor:

PR #3324, opened 2023-08-24
Add stat64at function declaration to AIX

ecnelises:
In AIX, when large file API enabled, fstatat will be redirected to stat64at.

Also, fix clockid_t to correct definition, as time-rs builds well with this change.


ecnelises:
Hmm looks it was successfully merged, should I do something to fix it?

JohnTitor:
No, you can ignore it :)

PR #3328, opened 2023-08-27
Add a few declarations for Apple systems

vincentisambart:
I added the following definitions for Apple systems:

- `_POSIX_SPAWN_DISABLE_ASLR`
  - It is a private flag for `posix_spawnattr_setflags` that disables [ASLR](https://en.wikipedia.org/wiki/Address_space_layout_randomization) for the child process.
  - It is mainly used by debuggers. Its declaration in LLDB's source code:
https://github.com/llvm/llvm-project/blob/2051a4121957a4bfc015b5102d06ffd6d0852d33/lldb/tools/debugserver/source/MacOSX/MachProcess.mm#L507-L509
- `pthread_chdir_np` and `pthread_fchdir_np`
  - Private functions that change the working directory only for the current thread.
  - They are declared and documented in Apple's libpthread:
https://github.com/apple-oss-distributions/libpthread/blob/67e155c94093be9a204b69637d198eceff2c7c46/private/pthread/private.h#L44-L89
  - They are for example used by Chromium in [this file](https://chromium.googlesource.com/chromium/src/base/+/master/process/launch_mac.cc).
  - LLDB also uses the [equivalent](https://github.com/apple-oss-distributions/libpthread/blob/main/src/pthread_cwd.c) `__pthread_chdir` and `__pthread_fchdir` in [this file](https://github.com/llvm/llvm-project/blob/main/lldb/source/Host/macosx/objcxx/Host.mm).
- `posix_spawnattr_getbinpref_np` and `posix_spawnattr_setbinpref_np`
  - These methods are public and declared in `/usr/include/spawn.h`.
  - They are pretty close to the already present `posix_spawnattr_getarchpref_np` and `posix_spawnattr_setarchpref_np`, but you can see that in [this file](https://github.com/llvm/llvm-project/blob/main/lldb/source/Host/macosx/objcxx/Host.mm) LLDB does use both `binpref` and `archpref`.


src/unix/bsd/apple/mod.rs
```diff
@@ -4687,12 +4687,13 @@ pub const MNT_SNAPSHOT: ::c_int = 0x40000000;
 pub const MNT_NOBLOCK: ::c_int = 0x00020000;
 
 // sys/spawn.h:
-pub const POSIX_SPAWN_RESETIDS: ::c_int = 0x01;
-pub const POSIX_SPAWN_SETPGROUP: ::c_int = 0x02;
-pub const POSIX_SPAWN_SETSIGDEF: ::c_int = 0x04;
-pub const POSIX_SPAWN_SETSIGMASK: ::c_int = 0x08;
-pub const POSIX_SPAWN_SETEXEC: ::c_int = 0x40;
-pub const POSIX_SPAWN_START_SUSPENDED: ::c_int = 0x80;
+pub const POSIX_SPAWN_RESETIDS: ::c_int = 0x0001;
```
vincentisambart, line 4690: They are declared with 4 digits in `sys/spawn.h`, and having them all 4 digits aligns a bit better so I made them all 4 digits. I can revert that if you prefer.
`posix_spawnattr_setflags` and `posix_spawnattr_getflags` use `short` so I think they should be declared `::c_short` (in C they are defined with `#define` so not really typed), but that would probably break compatibility so I left the type as-is, even for the `_POSIX_SPAWN_DISABLE_ASLR` I added.

src/unix/bsd/apple/mod.rs
```diff
@@ -4687,12 +4687,13 @@ pub const MNT_SNAPSHOT: ::c_int = 0x40000000;
...
+pub const POSIX_SPAWN_SETPGROUP: ::c_int = 0x0002;
+pub const POSIX_SPAWN_SETSIGDEF: ::c_int = 0x0004;
+pub const POSIX_SPAWN_SETSIGMASK: ::c_int = 0x0008;
+pub const POSIX_SPAWN_SETEXEC: ::c_int = 0x0040;
+pub const POSIX_SPAWN_START_SUSPENDED: ::c_int = 0x0080;
+pub const _POSIX_SPAWN_DISABLE_ASLR: ::c_int = 0x0100;
```
vincentisambart, line 4696: Technically it is not defined in `sys/spawn.h` like the others, but I thought putting it with the others, and in increasing value order, would make it easier to find.

libc-test/build.rs
```diff
@@ -328,6 +328,9 @@ fn test_apple(target: &str) {
 
             // FIXME: XCode 13.1 doesn't have it.
             "TIOCREMOTE" => true,
+
+            // Private value used by debuggers.
+            "_POSIX_SPAWN_DISABLE_ASLR" => true,
```
JohnTitor, line 333: I don't think we should expose private consts on libc, it's typically unstable and libc versioning doesn't fit it.

libc-test/build.rs
```diff
@@ -353,6 +356,9 @@ fn test_apple(target: &str) {
             // FIXME: Once the SDK get updated to Ventura's level
             "freadlink" | "mknodat" | "mkfifoat" => true,
 
+            // Private functions
+            "pthread_chdir_np" | "pthread_fchdir_np" => true,
```
JohnTitor, line 360: Same here

vincentisambart:
Removed all the private constants and methods.

JohnTitor (approved):

JohnTitor:
Thank you so much! @bors r+

PR #3334, opened 2023-09-01
Add ALG_SET_DRBG_ENTROPY and ALG_SET_KEY_BY_KEY_SERIAL constants

rzimmerman:
`ALG_SET_KEY_BY_KEY_SERIAL` was added in Linux 6.2:
https://elixir.bootlin.com/linux/v6.2.16/source/include/uapi/linux/if_alg.h#L54

Also adding `ALG_SET_DRBG_ENTROPY` for completeness.


rzimmerman:
Regarding the test failures, I think the jammy package for linux-headers is still tied to kernel version 5.15, so this constant is missing.

JohnTitor:
We recently updated Ubuntu images to 23.10, please rebase onto main.
Note that our musl header is old and you would have to tweak https://github.com/rust-lang/libc/blob/main/libc-test/build.rs.

tgross35:
@rzimmerman are you able to rebase to finish this?

tgross35:
@rzimmerman this should just need a rebase if you are able to get it over the line

tgross35:
Actually this should be all set, https://github.com/rust-lang/libc/pull/3404 added these constants. Closing since I don't think there is anything remaining here.

PR #3335, opened 2023-09-02
Remove deprecation from SA_ONSTACK; fix hexagon-unknown-linux-musl

androm3da:
In ad3f860b3cc7314331bbed1a925bce9d01b247df the RLIM_* values were removed, but leaving behind a deprecation about RLIM_NLIMITS caused it to incorrectly apply instead to SA_ONSTACK.

### DIFF
--- a/src/unix/linux_like/linux/arch/generic/mod.rs
+++ b/src/unix/linux_like/linux/arch/generic/mod.rs
@@ -257,6 +257,8 @@ cfg_if! {
         pub const RLIMIT_NICE: ::__rlimit_resource_t = 13;
         pub const RLIMIT_RTPRIO: ::__rlimit_resource_t = 14;
         pub const RLIMIT_RTTIME: ::__rlimit_resource_t = 15;
+        #[allow(deprecated)]
+        #[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
         pub const RLIMIT_NLIMITS: ::__rlimit_resource_t = RLIM_NLIMITS;
 
     } else if #[cfg(any(target_env = "musl", target_env = "ohos"))] {
@@ -277,16 +279,21 @@ cfg_if! {
         pub const RLIMIT_NICE: ::c_int = 13;
         pub const RLIMIT_RTPRIO: ::c_int = 14;
         pub const RLIMIT_RTTIME: ::c_int = 15;
+        #[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
         pub const RLIM_NLIMITS: ::c_int = 15;
+        #[allow(deprecated)]
+        #[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
         pub const RLIMIT_NLIMITS: ::c_int = RLIM_NLIMITS;
     }
 }
 
 cfg_if! {
     if #[cfg(target_env = "gnu")] {
+        #[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
         pub const RLIM_NLIMITS: ::__rlimit_resource_t = 16;
     }
     else if #[cfg(target_env = "uclibc")] {
+        #[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
         pub const RLIM_NLIMITS: ::__rlimit_resource_t = 15;
     }
 }
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -2811,7 +2811,10 @@ pub const POSIX_MADV_DONTNEED: ::c_int = 4;
 
 pub const RLIM_INFINITY: ::rlim_t = !0;
 pub const RLIMIT_RTTIME: ::c_int = 15;
+#[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
 pub const RLIMIT_NLIMITS: ::c_int = 16;
+#[allow(deprecated)]
+#[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
 pub const RLIM_NLIMITS: ::c_int = RLIMIT_NLIMITS;
 
 pub const MAP_ANONYMOUS: ::c_int = MAP_ANON;
--- a/src/unix/aix/mod.rs
+++ b/src/unix/aix/mod.rs
@@ -1762,6 +1762,7 @@ pub const PRIO_USER: ::c_int = 2;
 pub const RUSAGE_THREAD: ::c_int = 1;
 pub const RLIM_SAVED_MAX: ::c_ulong = RLIM_INFINITY - 1;
 pub const RLIM_SAVED_CUR: ::c_ulong = RLIM_INFINITY - 2;
+#[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
 pub const RLIM_NLIMITS: ::c_int = 10;
 
 // sys/sched.h
--- a/src/unix/haiku/mod.rs
+++ b/src/unix/haiku/mod.rs
@@ -741,6 +741,7 @@ pub const RLIMIT_AS: ::c_int = 6;
 pub const RLIM_INFINITY: ::rlim_t = 0xffffffff;
 // Haiku specific
 pub const RLIMIT_NOVMON: ::c_int = 7;
+#[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
 pub const RLIM_NLIMITS: ::c_int = 8;
 
 pub const RUSAGE_SELF: ::c_int = 0;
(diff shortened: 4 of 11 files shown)


androm3da:
cc @lancethepants 
cc @asomers

asomers (changes requested):
It looks like https://github.com/rust-lang/libc/commit/ad3f860b3cc7314331bbed1a925bce9d01b247df _also_ accidentally removed the deprecation notice from RLIM_NLIMITS.  Can you please restore it?

androm3da:
> It looks like [ad3f860](https://github.com/rust-lang/libc/commit/ad3f860b3cc7314331bbed1a925bce9d01b247df) _also_ accidentally removed the deprecation notice from RLIM_NLIMITS. Can you please restore it?

I didn't look closely enough there, thanks.  Will do.

androm3da:
> It looks like [ad3f860](https://github.com/rust-lang/libc/commit/ad3f860b3cc7314331bbed1a925bce9d01b247df) _also_ accidentally removed the deprecation notice from RLIM_NLIMITS. Can you please restore it?

Do I apply it to the RLIM_NLIMITS in `src/unix/linux_like/linux/arch/generic/mod.rs` or other instances as well?

src/unix/linux_like/linux/arch/generic/mod.rs
```diff
@@ -277,6 +277,7 @@ cfg_if! {
         pub const RLIMIT_NICE: ::c_int = 13;
         pub const RLIMIT_RTPRIO: ::c_int = 14;
         pub const RLIMIT_RTTIME: ::c_int = 15;
+#[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
```
asomers, line 280: Looks like some kind of whitespace problem here.

asomers (changes requested):
Yes, you should apply the deprecation to all instances of RLIM_NLIMITS.

src/unix/linux_like/linux/arch/generic/mod.rs
androm3da: Oh?  But I ran `rustfmt` and the `ci/style.rs` says `good style!`.

Can you offer more details about the whitespace problem?

asomers: The `#[deprecated...]` should have enough leading space to line up with the line it deprecates.  That's how the rest of the crate does it.

androm3da:
> Yes, you should apply the deprecation to all instances of RLIM_NLIMITS.

But I suppose applying it causes the build to fail.  I could locally bypass this warning but is that really what we want?

```
error: use of deprecated constant `unix::linux_like::linux::arch::generic::RLIM_NLIMITS`: Not stable across OS versions
   --> src/unix/linux_like/linux/arch/generic/mod.rs:260:59
    |
260 |         pub const RLIMIT_NLIMITS: ::__rlimit_resource_t = RLIM_NLIMITS;
    |                                                           ^^^^^^^^^^^^
    |
note: the lint level is defined here
   --> src/lib.rs:20:38
    |
20  | #![cfg_attr(libc_deny_warnings, deny(warnings))]
    |                                      ^^^^^^^^
    = note: `#[deny(deprecated)]` implied by `#[deny(warnings)]`

```

asomers:
Yes, you'll have to put a `#[allow(deprecated)]` at each one of those sites.

src/unix/linux_like/linux/arch/generic/mod.rs
```diff
@@ -257,6 +257,7 @@ cfg_if! {
         pub const RLIMIT_NICE: ::__rlimit_resource_t = 13;
         pub const RLIMIT_RTPRIO: ::__rlimit_resource_t = 14;
         pub const RLIMIT_RTTIME: ::__rlimit_resource_t = 15;
+        #[allow(deprecated)]
         pub const RLIMIT_NLIMITS: ::__rlimit_resource_t = RLIM_NLIMITS;
```
asomers, line 261: Hmm.  Come to think of it, we'll probably have to mark `RLIMIT_NLIMITS` as deprecated, too.

asomers (changes requested):

lancethepants:
There are also references to RLIM_NLIMITS in sparc, mips and powerpc if those needs to be labeled deprecated as well.
Untitled

asomers:
Yeah, all of them do.  There's no way to reliably use that constant, given that the OS can change it at any time.

src/unix/linux_like/linux/arch/generic/mod.rs
androm3da: Okay -- got it, thanks.

androm3da:
> Yeah, all of them do. There's no way to reliably use that constant, given that the OS can change it at any time.

Okay - so I'll apply the deprecation to `RLIM{IT,}_NLIMITS` for to all of the targets, then.  And then `#[allow(deprecated)]` for any uses of either within the `libc` crate.

androm3da:
> > Yeah, all of them do. There's no way to reliably use that constant, given that the OS can change it at any time.
> 
> Okay - so I'll apply the deprecation to `RLIM{IT,}_NLIMITS` for to all of the targets, then. And then `#[allow(deprecated)]` for any uses of either within the `libc` crate.

Also: `RLIMIT_NLIMITS` / `RLIM_NLIMITS` appears a few times within `libc-test/semver/` - but no changes are required here?

asomers:
> 
> Also: `RLIMIT_NLIMITS` / `RLIM_NLIMITS` appears a few times within `libc-test/semver/` - but no changes are required here?

I'm not sure.  I don't know how that part works.  But if CI passes, then I wouldn't worry about it.

androm3da:
Not changed: references to `RLIM*_NLIMITS` in fuchsia, aix, emscripten - but I suppose these are bound by the same limitation and I should mark them deprecated there too?

asomers:
Yeah, it ought to be deprecated everywhere.

androm3da:
cc @cramertj 
cc @ecnelises

asomers (changes requested):
What about src/unix/bsd ?

ecnelises:
Looks good to me about AIX. AIX has backward compatible `RLIM_NLIMITS_*` for elder versions, but not actually changed for very long time.

androm3da:
> What about src/unix/bsd ?

Looks like that's already as-expected.

```
$ rg -N -B3 '(RLIM|RLIMIT)_NLIM' src/unix/bsd/
src/unix/bsd/netbsdlike/openbsd/mod.rs
pub const AT_REMOVEDIR: ::c_int = 0x08;

pub const RLIM_NLIMITS: ::c_int = 9;

src/unix/bsd/netbsdlike/netbsd/mod.rs
pub const RLIMIT_NTHR: ::c_int = 11;

pub const RLIM_NLIMITS: ::c_int = 12;

src/unix/bsd/apple/mod.rs
pub const RLIMIT_NPROC: ::c_int = 7;
pub const RLIMIT_NOFILE: ::c_int = 8;
pub const RLIM_NLIMITS: ::c_int = 9;

src/unix/bsd/freebsdlike/freebsd/mod.rs
pub const RLIMIT_KQUEUES: ::c_int = 13;
pub const RLIMIT_UMTXP: ::c_int = 14;
pub const RLIM_NLIMITS: ::rlim_t = 15;

src/unix/bsd/freebsdlike/dragonfly/mod.rs
pub const ELAST: ::c_int = 99;
pub const RLIMIT_POSIXLOCKS: ::c_int = 11;
pub const RLIM_NLIMITS: ::rlim_t = 12;

```

asomers:
> > What about src/unix/bsd ?
> 
> Looks like that's already as-expected.

Ahh, good.  But what about Haiku and nto?

asomers (approved):

androm3da:
Thanks, @asomers 

Are you able to remove the `S-waiting-on-review` label and replace it with whatever's appropriate?  I tried but I don't think I have permission to change this.

androm3da:
> Thanks, @asomers
> 
> Are you able to remove the `S-waiting-on-review` label and replace it with whatever's appropriate? I tried but I don't think I have permission to change this.

gentle ping @asomers

asomers:
No, @androm3da I don't have any permissions in this repo.

ecnelises:
Maybe @JohnTitor ?

androm3da:
I have updated the commit message for ba6e03bcbbfb869056282b91559c9ff059dc17bd to clarfiy that this is a fix for the `libstd` build for the `hexagon-unknown-linux-musl` target.

JohnTitor:

androm3da:

Done.  Sorry, I'd composed the commits independently in order to make it so that if the `RLIM_LIMITS` deprecation caused problems, that part could be reverted without reverting the un-deprecation of `SA_ONSTACK`.  But I probably should've looked closer at the contribution guidelines.

androm3da:
Thanks @JohnTitor, let me know if there's anything else I need to fix. 😁

JohnTitor:
Thanks! @bors r+

JohnTitor (approved):

PR #3337, opened 2023-09-04
Add F_PUNCHHOLE and fpunchhole_t

anacrolix:


src/unix/bsd/apple/mod.rs
```diff
@@ -368,6 +368,13 @@ s! {
         pub fst_bytesalloc: ::off_t,
     }
 
+    pub struct fpunchhole_t {
+        pub fp_flags: ::c_uint, /* unused */
+        pub reserved: ::c_uint, /* (to maintain 8-byte alignment) */
```
devnexen, line 373: nit: there is no need to expose this field, its sole role is mentioned in the comment.

anacrolix: Oh you're right. If I make these private, can people still instantiate the struct?

anacrolix: It might be best to leave them available. I don't believe there's any chance of misuse, and their purposes might change in the future.

devnexen: fair point it s easier to instantiate that way indeed. another note, you can amend your commit message with adding
`close #<issue id>`.

anacrolix: I've added that to the one of the commit messages.

anacrolix:
I've used this change from a real system and it works great. LMK what else is needed.

src/unix/bsd/apple/mod.rs
devnexen: cool, you just need to rebase into 1 commit and you re good :)

anacrolix:
@JohnTitor could you take a look at this?

JohnTitor (approved):

JohnTitor:
Thank you! @bors r+

anacrolix:
Thank you kindly, @JohnTitor and @devnexen

PR #3341, opened 2023-09-06
timerfd api for freebsd 14.

### LINKED ISSUE #3339: timerfd on FreeBSD >= 14.0
Same API as Linux. See [timerfd(2)](https://man.freebsd.org/cgi/man.cgi?query=timerfd&sektion=2&manpath=FreeBSD+15.0-CURRENT) and [`<sys/timerfd.h>`](https://github.com/freebsd/freebsd-src/blob/main/sys/sys/timerfd.h) for details.

Similar to `eventfd` the ABI is new, so no need to gate by `cfg(freebsd14)`.

devnexen:
close #3339

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -4693,6 +4693,11 @@ pub const RB_POWERCYCLE: ::c_int = 0x400000;
 pub const RB_PROBE: ::c_int = 0x10000000;
 pub const RB_MULTIPLE: ::c_int = 0x20000000;
 
+// sys/timerfd.h
+
+pub const TFD_NONBLOCK: ::c_int = ::O_NONBLOCK;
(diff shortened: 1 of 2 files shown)


src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -5406,6 +5411,15 @@ extern "C" {
         infotype: *mut ::c_uint,
         flags: *mut ::c_int,
     ) -> ::ssize_t;
+
+    pub fn timerfd_create(clockid: ::clockid_t, flags: ::c_int) -> ::c_int;
```
asomers, line 5415: Nit: the header and man page both define `clockid` as `int`, not `clockid_t`, though those are the same type.  Is the difference intentional?

devnexen: no I just misread.

asomers (approved):

JohnTitor (approved):

PR #3343, opened 2023-09-07
Move all seccomp consts and structs into top-level mod

### LINKED ISSUE #3342: Fix seccomp constants on musl/all linux platforms
on x86_64-unknown-linux-gnu, some of the seccomp-related constants are provided but they are not on all platforms.

These constants live inside <linux/seccomp.h> and it appears that some of them exist in rust's libc linux/mod.rs
Linux: https://github.com/torvalds/linux/blob/master/include/uapi/linux/seccomp.h#L10
Rust libc linux/mod.rs: https://github.com/rust-lang/libc/blob/835661543db1ec42a6d9a809d69c3c5b5b978b81/src/unix/linux_like/linux/mod.rs#L2190
Rust libc linux/gnu/<platform>/mod.rs https://github.com/rust-lang/libc/blob/835661543db1ec42a6d9a809d69c3c5b5b978b81/src/unix/linux_like/linux/gnu/b64/x86_64/mod.rs#L805

The constants
```
/* Valid operations for seccomp syscall. */
#define SECCOMP_SET_MODE_STRICT		0
#define SECCOMP_SET_MODE_FILTER		1
#define SECCOMP_GET_ACTION_AVAIL	2
#define SECCOMP_GET_NOTIF_SIZES		3
```
from seccomp.h appear to be missing from `unix/linux_like/linux/mod.rs` . Some other newer constants also appear to be missing e.g. `SECCOMP_FILTER_FLAG_NEW_LISTENER`.

If it seems okay to move everything into linux/mod.rs, I'll make a PR for that.

boustrophedon:
Move all seccomp consts and structs into top-level mod    

Seccomp constants and structs were partially defined in the top-level mod.rs for linux and partially outside. This commit moves everything into the top-level mod and adds missing entries as of linux 6.4.12.

Resolves #3342 and [rust-vmm/seccompiler#57](https://github.com/rust-vmm/seccompiler/issues/57)

I couldn't get main to run libc-test cargo test (even without my changes) with either gcc 13.2.1 or clang 16.0.6. They failed with different errors/warnings.


boustrophedon:
style check is failing but I get

```
$ rustc ci/style.rs && ./style src
good style!
```

It looks like it was some indentation issue. Either way, I fixed it.

boustrophedon:
The linux failures are coming from SECCOMP_FILTER_FLAG_WAIT_KILLABLE_RECV which was added last year https://github.com/torvalds/linux/commit/c2aa2dfef243efe213a480a1ee8566507a5152f4

Looks like it was available in about 5.19 or so, and the docker image this is running in is `ubuntu-22.04`  which is running 5.15 [according to this page](https://github.com/actions/runner-images/blob/ubuntu22/20230821.1/images/linux/Ubuntu2204-Readme.md).

I'm not sure what to do here.

boustrophedon:
Hi, anyone have any suggestions for what to do here?

The issue is that the version of the Linux kernel in CI does not support some newer constants for the seccomp syscall, and so CI fails when checking whether the constants exist in the corresponding C headers.

However, the version in CI is also newer than the [minimum supported version of the linux kernel](https://doc.rust-lang.org/stable/rustc/platform-support.html) so in theory this crate may already have constants that are not supported by the minimum version but do happen to be supported by the version in CI.

JohnTitor:
Could you try to do a rebase? We recently updated the Ubuntu image on CI to 23.10, which uses Linux v6.5.

boustrophedon:
@JohnTitor Thanks, it looks like everything is passing now!

JohnTitor (approved):

JohnTitor:
Cool! One concern is that our musl toolchain is quite old and might fail, but let's see if that's the case.
If fails, you can ignore tests like:
https://github.com/rust-lang/libc/blob/240de1f185aef26938b1eb20be1ceb3ec99c35bb/libc-test/build.rs#L3503-L3506

@bors r+

boustrophedon:
Yeah, it looks like it's failing on musl with stuff like
```
cargo:warning=/checkout/target/aarch64-unknown-linux-musl/debug/build/libc-test-dd9ce03b4e008979/out/main.c: In function '__test_field_type_seccomp_notif_sizes_seccomp_data':

  cargo:warning=/checkout/target/aarch64-unknown-linux-musl/debug/build/libc-test-dd9ce03b4e008979/out/main.c:21349:30: error: invalid use of undefined type 'struct seccomp_notif_sizes'
```

```
  cargo:warning=/checkout/target/aarch64-unknown-linux-musl/debug/build/libc-test-dd9ce03b4e008979/out/main.c:21700:38: error: invalid use of undefined type 'struct seccomp_notif_addfd'
```

Could I help out with upgrading musl? If you think it's too much work I'm fine with just adding the exceptions for the failing types and constants.

JohnTitor:
> Could I help out with upgrading musl? If you think it's too much work I'm fine with just adding the exceptions for the failing types and constants.

Sadly it's quite hard. We have to upgrade musl from v1.1.24 to v1.2 or higher, but it would break the compatibility. See https://github.com/rust-lang/libc/pull/3068 for ongoing work. I'd recommend just skipping tests for now.

boustrophedon:
I'm not sure if I'm able to trigger a bors run to test the tier2 targets.

JohnTitor:
tier2 targets are only run on bors (otherwise it consumes a lot of GHA budget on the org), @bors r+

boustrophedon:
So I'm trying to reproduce the failure locally and the first issue I ran into was
```
+ cargo test --no-default-features --manifest-path libc-test/Cargo.toml --target mips-unknown-linux-gnu
    Updating crates.io index
error: failed to write /checkout/Cargo.lock

Caused by:
  failed to open: /checkout/Cargo.lock  

Caused by:
  Read-only file system (os error 30)
```

but the docker run command is

boustrophedon:
Just for reference, the failure in CI looks like
```
PASSED 20 tests
     Running test/linux_strerror_r.rs (target/mips-unknown-linux-gnu/debug/deps/linux_strerror_r-cbe4f10440c1c52a)
RUNNING ALL TESTS
PASSED 1 tests
     Running test/linux_termios.rs (target/mips-unknown-linux-gnu/debug/deps/linux_termios-16515f471baa64e5)
RUNNING ALL TESTS
PASSED 50 tests
     Running test/main.rs (target/mips-unknown-linux-gnu/debug/deps/main-bf0b31ae08956b48)
RUNNING ALL TESTS
rust[4] = 0 != 4 (C): Rust "seccomp_notif_sizes" -> C
rust[5] = 0 != 5 (C): Rust "seccomp_notif_sizes" -> C
thread 'main' panicked at 'some tests failed', /checkout/target/mips-unknown-linux-gnu/debug/build/libc-test-b013fc4656104f6c/out/main.rs:12:21
```

boustrophedon:
Aha, I took another look at the docker command:

```
    docker run \
      --rm \
      --user "$(id -u)":"$(id -g)" \

**Comment by JohnTitor** [comment]
So, we have to build the target using `-Zbuild-std` now, as https://github.com/rust-lang/compiler-team/issues/648 dropped the support. As it's now a tier 3 target, we could downgrade our CI support for MIPS instead. Let me do that, one sec.

**Comment by boustrophedon** [comment]
@JohnTitor Does docker CI run locally for you? It might be an issue with my environment just not matching what the scripts want.

**Comment by Arnavion** [comment]
>Does docker CI run locally for you?

Yes it works fine.

```sh
printf '[toolchain]\nchannel = "nightly-2023-07-04"\n' >rust-toolchain.toml
rustup target add mips-unknown-linux-gnu
cargo generate-lockfile
./ci/run-docker.sh mips-unknown-linux-gnu
```

(Replace `mips-unknown-linux-gnu` with whatever target you want to test.)

> ```
> + cargo test --no-default-features --manifest-path libc-test/Cargo.toml --target mips-unknown-linux-gnu
>     Updating crates.io index
> error: failed to write /checkout/Cargo.lock
> 
> Caused by:
>   failed to open: /checkout/Cargo.lock  
> 
> Caused by:
>   Read-only file system (os error 30)
> ```

This is because you didn't run [`cargo generate-lockfile`](https://github.com/rust-lang/libc/blob/f5172d29959326a45bd6cef88c38896bf3aeb479/ci/install-rust.sh#L73)

Arnavion:
Anyway I rebased your branch on main and test all the targets except mips and android for you and they all pass. So if you rebase and ask JohnTitor to rerun CI it should be fine.

boustrophedon:
I rebased, thanks for helping out with this! @JohnTitor could you re-try CI please?

It might be nice to add instructions in the ci/ directory on how to run everything. I also am not sure if the solution to the uid and bindmount ro/rw stuff is to just run the scripts as root locally, which seems not ideal, or if there's a better way besides modifying the scripts.

Arnavion:
>I also am not sure if the solution to the uid and bindmount ro/rw stuff is to just run the scripts as root locally, which seems not ideal, or if there's a better way besides modifying the scripts.

As I said, add yourself to the `docker` group.

boustrophedon:
Sorry, I forgot to mention my user is in the docker group!

boustrophedon:
Hi again, sorry to bother, @JohnTitor could you re-run the CI tests when you have a chance? Thank you!

JohnTitor:
Sorry for the long delay! I missed this. @bors r+

boustrophedon:
Thank you!

PR #3352, opened 2023-09-13
Add time namespace constant

cd-work:
This constant is a possible bitflag argument for `unshare`: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/sched.h?h=v6.6-rc1#n44

I am unsure about its support in Fuchsia, which is why I have not added the constant there.


JohnTitor (approved):

cd-work:
Seems like some tier2 architectures aren't supported, not sure how this is generally handled.

Would the best place to add this be `/src/unix/linux_like/linux/gnu/mod.rs`?

JohnTitor:
Feel free to add to the targets where you want to use the item.

JohnTitor:

PR #3356, opened 2023-09-18
Change `FD_SETSIZE` to `c_int`

magicant:
`FD_SETSIZE` defines an upper bound for file descriptors that can be added to `fd_set`. It is common to check if a file descriptor is less than `FD_SETSIZE` before adding it to an `fd_set`. See the example below:

```rust
fn main() {
    let path = std::ffi::CStr::from_bytes_with_nul(b"/dev/null\0").unwrap();
    let fd = unsafe { libc::open(path.as_ptr(), libc::O_RDONLY) };
    if fd < 0 {
        eprint!("open failed\n");
        return;
    }

    let mut set = std::mem::MaybeUninit::<libc::fd_set>::uninit();
    unsafe { libc::FD_ZERO(set.as_mut_ptr()) }
    if fd < libc::FD_SETSIZE { // <- type mismatch!
        unsafe { libc::FD_SET(fd, set.as_mut_ptr()) }
    } else {
        eprint!("too large fd\n");
    }
}
```

Unfortunately, this example does not compile because of a type mismatch. `fd` is `c_int` while `FD_SETSIZE` is `usize` (or `size_t`).
Since `FD_SETSIZE` represents the max file descriptor + 1, I think it should have the same type as file descriptors. This pull request modifies the type to `c_int` and adds casts where needed.


sthibaul:
Please also update `src/unix/hurd/mod.rs`, introduced in the meanwhile, thanks!

magicant:
Rebased and updated my commit.

I also changed `_POSIX_OPEN_MAX` as well as `_POSIX_FD_SETSIZE` to `c_int` since the latter is defined as an alias of the former.

sthibaul:
Ah, sorry, my commit indeed modified a lot, up to conflicting yours.

magicant:
No problem, my diff is so small that the conflict can be easily resolved 🙂

JohnTitor (approved):

PR #3357, opened 2023-09-19
Expose `len8_dlc` field of `can_frame` struct.

de-vri-es:
The `can_frame` struct on Linux has a `len8_dlc` field which can be used to stuff a few extra bits in the `dlc` field of the actual CAN frame on the wire (only allowed when the data length itself is 8, and if the hardware CAN controller allows it).

However, that field is not currently exposed. This PR exposes it as `pub len8_dlc: u8` so we can get those extra 2.8 bits from our CAN frames :sweat_smile: 

It would have been nice to name the current `can_dlc` field as `len`, instead of the old deprecated name, but I guess that ship has sailed.

Upstream documentation: https://www.kernel.org/doc/html/latest/networking/can.html#how-to-use-socketcan 

> ```c
> struct can_frame {
>         canid_t can_id;  /* 32 bit CAN_ID + EFF/RTR/ERR flags */
>         union {
>                 /* CAN frame payload length in byte (0 .. CAN_MAX_DLEN)
>                  * was previously named can_dlc so we need to carry that
>                  * name for legacy support
>                  */
>                 __u8 len;
>                 __u8 can_dlc; /* deprecated */
>         };
>         __u8    __pad;   /* padding */
>         __u8    __res0;  /* reserved / padding */
>         __u8    len8_dlc; /* optional DLC for 8 byte payload length (9 .. 15) */
>         __u8    data[8] __attribute__((aligned(8)));
> };
> ```
> 
> Remark: The len element contains the payload length in bytes and should be used instead of can_dlc. The deprecated can_dlc was misleadingly named as it always contained the plain payload length in bytes and not the so called 'data length code' (DLC).
>
> To pass the raw DLC from/to a Classical CAN network device the len8_dlc element can contain values 9 .. 15 when the len element is 8 (the real payload length for all DLC values greater or equal to 8).

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -1580,10 +1580,11 @@ s_no_extra_traits! {
     #[allow(missing_debug_implementations)]
     pub struct can_frame {
         pub can_id: canid_t,
+        // FIXME(1.0): this field was renamed to `len` in Linux 5.11
         pub can_dlc: u8,
         __pad: u8,
         __res0: u8,
-        __res1: u8,
+        pub len8_dlc: u8,
         pub data: [u8; CAN_MAX_DLEN],
     }
 


JohnTitor (approved):

de-vri-es:
I don't see any reason for the failure. Is it correct to assume that this was an unrelated CI failure? :o

JohnTitor:
The failure is _related_. I think mips doesn't have it or our musl header is too old.

de-vri-es:
Oh, sorry, the link just pointed to a job running "exit 1" and I didn't notice the other red check.

de-vri-es:
Ah, yes. Note that the `can_frame` struct comes from a Linux header, not from a glibc or musl header. I'm also pretty sure that `mips` has it too, but the headers are pulled in from the openwrt sdk which includes headers from linux 4.14, and the `len8_dlc` field was added in Linux  commit ea7800565a128 which was released in 5.11.

What would be a good way forward for me to unblock this? Find a different linux mips distro with a newer kernel?

de-vri-es:
Hmm, updating to https://downloads.openwrt.org/releases/23.05.0-rc4/targets/bcm47xx/generic/openwrt-sdk-23.05.0-rc4-bcm47xx-generic_gcc-12.3.0_musl.Linux-x86_64.tar.xz breaks other tests, atleast if I locally run `ci/run-docker.sh mipsel-unknown-linux-musl`.

I guess I need to figure out why first :(

tgross35:
@de-vri-es any chance you could update this? MIPS dropped to tier3 so if that was the cause of failure here, it shouldn't be a problem anymore.

de-vri-es:
Re-applied the changes to current `main`.

de-vri-es:
CI now fails because the kernel headers for the musl tests are taken from https://github.com/sabotage-linux/kernel-headers which is based on linux 4.19, but this field was introduced in linux 5.11 (released in february 2021).

tgross35:
Could you just update skip_field in libc-test/build.rs and leave a `FIXME(musl)`? We should be updating these in the near future

de-vri-es:
Done, and rebased again!

@rustbot ready

de-vri-es:
Huh.. I don't see my changes to `build.rs` anymore in the git history but CI succeeded.. Maybe I messed up the rebase and someone updated the `musl` headers in the mean time? :laughing:

tgross35 (approved):
Thanks!

src/unix/linux_like/linux/mod.rs
```diff
@@ -1583,7 +1583,7 @@ s_no_extra_traits! {
         pub can_dlc: u8,
```
tgross35, line 1583: While you're here, mind adding a comment ``// FIXME(1.0): this field was renamed to `can_len` `` based on the docs?

de-vri-es: Sure!

tgross35:
> Huh.. I don't see my changes to `build.rs` anymore in the git history but CI succeeded.. Maybe I messed up the rebase and someone updated the `musl` headers in the mean time? 😆

Indeed :) since https://github.com/rust-lang/libc/pull/3921 musl is on 6.6

I left one quick unrelated request since you're updating the struct, should be ready in either case.

de-vri-es:
> Indeed :) since #3921 musl is on 6.6

Nice :)

> I left one quick unrelated request since you're updating the struct, should be ready in either case.

Added the comment to `can_dlc` (new name is "`len`") :)

tgross35:
Oops thanks for the name correction. LGTM!

src/unix/linux_like/linux/mod.rs
de-vri-es: Added!

de-vri-es:
Had to force-push again, I had a typo in `FIXME`.

de-vri-es:
@tgross35: Thanks for the reviews!

Issue #3248, opened 2023-05-15
Tracking issue for libc 0.3


joshtriplett:
This issue tracks changes we want to make in libc 0.3.

- [ ] musl 1.2

bossmc:
On the LFS64 symbol removal point (#2935) the current PR is attempting to achieve rust crate API compatibility (by aliasing LFS64 things to their non-LFS64 equivalents), hiding the effective ABI break in musl libc (it's not a real break since those symbols never "really" existed, but they existed enough for the libc crate to wind up depending on them 😞).  If 0.3 is going to be a breaking change we could do a simpler change where we simply stop all the LFS64 functions and types from the libc crate (which is arguably closer to the musl libc API).

My expectation was that we'd merge the API-compatible change in a 0.2.x (and maybe then deprecate the aliases?) and remove them "later" in a 0.3 release, but if 0.3 is imminent, maybe we just drop the items completely?

Note that the rust libc crate has to do "something" here to work with musl 1.2.4+ (@wesleywiser moved rustc's self-contained version to 1.2.3 since that's ABI-compatible with 1.1.24 but 1.2.4 breaks the crate's use of the ABI).

joshtriplett:
@bossmc Figuring out how to deal with LFS64 and time64 has been one of the two biggest blocking issues for this. (The other is having unnamed fields of struct and union types.)

We *could* just try to match the API of musl as closely as possible, but that just pushes the problem onto whoever *uses* libc.

Personally, I would like to minimize the degree to which people have to write a pile of `cfg`s in order to use the 64-bit functions. `libc::foo64` seems preferable to `#[cfg(...)] use libc::foo as foo64; #[cfg(not(...))] use libc::foo64;` (with corresponding complexity for every other function as well as off_t).

Is there some reasonable way we could *generally* make the same function names and types available on every platform that has 64-bit support, so that people only need `cfg` if they want to support platforms that *don't* have 64-bit support?

lvella:
So, the changes for 64 bits version of time and file offset function is just for `musl`? Can't 32 bit `glibc` be included in the package?

JohnTitor:
> Fork a libc-0.2 branch from master on which development for libc 0.2 will continue.

Created the [`libc-0.2` branch](https://github.com/rust-lang/libc/tree/libc-0.2). Once the musl 1.2 PR (or any other breaking changes) gets merged, I'll update the bors config and docs.

ydirson:
[...]

PR #3367, opened 2023-09-27
freebsd: move net/if_mib.h contents to submodule

ydirson:
There is a conflict of NETLINK_GENERIC definitions between net/if_mib.h and netlink/netlink.h.  netlink.h is already exported in the crate root for Linux (and those definitions are already used by at least crates neli and netlink-packet-route), and if_mib is not much used yet, so this moves if_mib contents into its own namespace to leave place for netlink support on FreeBSD (rust-lang/libc#3201).

Looks like it is the first time namespacing is used in the libc crate, so I had to guess a few things, being quite new to this project.

The moved symbols did not appear in semver, so I just added the new ones.  `sysinfo` from @GuillaumeGomez seems to be the only impacted public crate.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2453,6 +2453,33 @@ fn test_freebsd(target: &str) {
             // FIXME: The values has been changed in FreeBSD 15:
             "CLOCK_BOOTTIME" if Some(15) <= freebsd_ver => true,
 
+            // submodule ifmib, cannot be checked today
+            "DOT3COMPLIANCE_STATS"
+            | "DOT3COMPLIANCE_COLLS"
+            | "IFDATA_GENERAL"
+            | "IFDATA_LINKSPECIFIC"
+            | "IFDATA_DRIVERNAME"
+            | "IFMIB_IFCOUNT"
+            | "IFMIB_IFDATA"
+            | "IFMIB_SYSTEM"
+            | "NETLINK_GENERIC"
+            | "dot3ChipSetAMD7990"
+            | "dot3ChipSetAMD79900"
+            | "dot3ChipSetAMD79C940"
+            | "dot3ChipSetIntel82586"
+            | "dot3ChipSetIntel82596"
+            | "dot3ChipSetIntel82557"
+            | "dot3ChipSetNational8390"
+            | "dot3ChipSetNationalSonic"
+            | "dot3ChipSetFujitsu86950"
+            | "dot3ChipSetDigitalDC21040"
+            | "dot3ChipSetDigitalDC21140"
+            | "dot3ChipSetDigitalDC21041"
+            | "dot3ChipSetDigitalDC21140A"
+            | "dot3ChipSetDigitalDC21142"
+            | "dot3ChipSetWesternDigital83C690"
+            | "dot3ChipSetWesternDigital83C790" => true,
+
             _ => false,
         }
     });
--- a/src/unix/bsd/freebsdlike/freebsd/ifmib.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/ifmib.rs
@@ -1,0 +1,44 @@
+// sys/net/if_mib.h
+
+/// non-interface-specific
+pub const IFMIB_SYSTEM: ::c_int = 1;
+/// per-interface data table
+pub const IFMIB_IFDATA: ::c_int = 2;
+
+/// generic stats for all kinds of ifaces
+pub const IFDATA_GENERAL: ::c_int = 1;
+/// specific to the type of interface
+pub const IFDATA_LINKSPECIFIC: ::c_int = 2;
+/// driver name and unit
+pub const IFDATA_DRIVERNAME: ::c_int = 3;
+
+/// number of interfaces configured
+pub const IFMIB_IFCOUNT: ::c_int = 1;
+
+/// functions not specific to a type of iface
+pub const NETLINK_GENERIC: ::c_int = 0;
+
+pub const DOT3COMPLIANCE_STATS: ::c_int = 1;
+pub const DOT3COMPLIANCE_COLLS: ::c_int = 2;
+
+pub const dot3ChipSetAMD7990: ::c_int = 1;
+pub const dot3ChipSetAMD79900: ::c_int = 2;
+pub const dot3ChipSetAMD79C940: ::c_int = 3;
+
+pub const dot3ChipSetIntel82586: ::c_int = 1;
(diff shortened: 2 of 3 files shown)


ydirson:
New revision completely drops the semver part -- will gladly use any hint to get it working and add it back.

GuillaumeGomez:
To avoid the semver issue, the only way I see is to reexport items from the submodule by default.

ydirson:
> To avoid the semver issue, the only way I see is to reexport items from the submodule by default.

That would avoid the breaking change, but that would not work for `NETLINK_GENERIC`, which this PR is all about.

GuillaumeGomez:
Indeed. Let's see if libc's maintainers have another way to do it. But otherwise, I'm very fine with the breaking change. I'll just update `libc` in `sysinfo` and everything should be fine.

ydirson:
I'm baffled by the tests still trying to look at `ifmib::*` after I removed them from semver...

ydirson:
> I'm baffled by the tests still trying to look at `ifmib::*` after I removed them from semver...

OK I realize they come from the generated tests. This part is essentially dedicated to `ctest`, and I don't see a way there to tell a header's contents live in a submodule.
Looks like for now we'll have to flag those definitions not to be checked.

JohnTitor:
Note also that private code might use these items. The libs-maintainers team has been considering making a new major release (0.3? 1.0? I'm not sure though) and I think it'd be better to include this change there. Let's wait for that release.

ydirson:
> Note also that private code might use these items. The libs-maintainers team has been considering making a new major release (0.3? 1.0? I'm not sure though) and I think it'd be better to include this change there. Let's wait for that release.

How can we track that here, when the symbols moved were not included in semver to start with?  Is there a "list of breaking changes" somewhere to add this PR to for consideration?

JohnTitor:
See https://github.com/rust-lang/libc/issues/3248, I've added this to the task list.

ydirson:
Rebased onto 0.2.151

asomers:
What about adding a feature flag to libc?  An off-by-default "ifmib" feature would let crates like @ydirson's use the constants in their correct namespace, not conflicting with the new constant introduced by #3201 , while other crates won't see a change.  And libc could remove the feature flag when it does next make a major release.

ydirson:
As I understand it, such a flag would enable either `ifmib` of `netlink`, so for 0.2 it could make sense to let `ifmib` be the default, and rather add a `netlink` feature flag to disable it and enable netlink support?

asomers:
Whatever it's named, the default option should be the status quo, and the alternative would be to put ifmib into a submodule.  BTW, I've seen other crates that call this "unstable".

ydirson:
> I've seen other crates that call this "unstable".

That would seem a bit too generic here, we could just name it `netlink`.
However I see some moves on the 0.3 front, so should I proceed with this suggestion for 0.2?

ydirson:
rebased

tgross35:
Could we just add a single `NETLINK_GENERIC_IF_MIB`?

ydirson:
> Could we just add a single `NETLINK_GENERIC_IF_MIB`? Or are more conflicts expected?

All software using the `if_mib.h` version will miscompile and need to make a change to get the `NETLINK_GENERIC` they need.  My gut feeling is that anyone using that API is likely to use other consts (but I may be wrong), and the namespacing would prevent miscompilation if someone just casually tries to bump `libc`.  The change would be more fullproof if we forced attention by moving the `ifmibdata` type to the namespace as well.

But I would not object to a simpler change without namespacing if other parties more in contact with that API feel it's OK. @GuillaumeGomez do you think about any FreeBSD devs who would have informed opinion? 

Note I just stumbled by chance (digging into `sysinfo` to check how that API is used there) against recently-merged #4022, adding the same `if_mib,h` API for `apple`, whose code would now need to be changed at the same time.

GuillaumeGomez:
I don't know of any FreeBSD devs so I have no idea.

tgross35:
If the intent is that both netlink and if_mib are available and the user has to select one, then it makes sense to move either everything from if_mib to a separate module, possibly everything from netlink to its own module as well. If instead it is only a couple of variables that conflict then I don't think we need to move everything.

@asomers what is planned here? Based on your above comment

> What about adding a feature flag to libc? An off-by-default "ifmib" feature would let crates like @ydirson's use the constants in their correct namespace, not conflicting with the new constant introduced by #3201 , while other crates won't see a change. And libc could remove the feature flag when it does next make a major release.

this makes it sound a bit like `if_mib` is intended to replace `netlink` at some point.

asomers:
> If the intent is that both netlink and if_mib are available and the user has to select one, then it makes sense to move either everything from if_mib to a separate module, possibly everything from netlink to its own module as well. If instead it is only a couple of variables that conflict then I don't think we need to move everything.
> 
> @asomers what is planned here? Based on your above comment
> 
> > What about adding a feature flag to libc? An off-by-default "ifmib" feature would let crates like @ydirson's use the constants in their correct namespace, not conflicting with the new constant introduced by #3201 , while other crates won't see a change. And libc could remove the feature flag when it does next make a major release.
> 
> this makes it sound a bit like `if_mib` is intended to replace `netlink` at some point.

No.  If anything, more stuff is moving toward netlink.  But I definitely expect both to be used for a long time.  I don't see any solution that doesn't involve moving one thing or the other into a submodule.  Perhaps both.

tgross35:
Wrote down my thoughts for how I would like to proceed: https://github.com/rust-lang/libc/pull/3201#issuecomment-4736374182

@rustbot author

dybucc:
not sure whether the pr author is still on this. if not, i picked up where they left off and did as mentioned in [^1]. see [^2] for changes listed as PR 1 in [^1].

[^2]: https://github.com/dybucc/libc/commit/6b2874579696d2bb90aa1df324049e6f50b2c019

tgross35:
Ping @ydirson, are you still interested in working on this?

Give it about a week and if we don't hear back, @dybucc feel free to start putting up PRs.

ydirson:
> Ping @ydirson, are you still interested in working on this?
> 
> Give it about a week and if we don't hear back, @dybucc feel free to start putting up PRs.

Still interested, but I'm a bit on other things at the moment, wouldn't mind if someone has the motivation to take the lead on this.

dybucc:
Done. See #5325.

tgross35:
I think this should be resolved by the above-linked PR which is the first step in https://github.com/rust-lang/libc/pull/3201#issuecomment-4736374182.

PR #3373, opened 2023-10-06
feat: add close_range for glibc

SteveLauC:
This PR adds [`close_range(2)`](https://man7.org/linux/man-pages/man2/close_range.2.html) for Linux/glibc:

```c
#include <linux/close_range.h>

       int close_range(unsigned int first, unsigned int last,
                       unsigned int flags);
```

musl and uClibc don't have wrappers for this syscall

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -1401,11 +1401,7 @@ extern "C" {
     ) -> ::c_int;
 
     // Added in `glibc` 2.34
-    pub fn close_range(
-        first: ::c_uint,
-        last: ::c_uint,
-        flags: ::c_int,
-    ) -> ::c_int;
+    pub fn close_range(first: ::c_uint, last: ::c_uint, flags: ::c_int) -> ::c_int;
 }
 
 cfg_if! {
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -1399,6 +1399,13 @@ extern "C" {
         envp: *const *mut c_char,
         flags: ::c_int,
     ) -> ::c_int;
+
+    // Added in `glibc` 2.34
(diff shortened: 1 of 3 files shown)


src/unix/linux_like/linux/gnu/mod.rs
```diff
@@ -1399,6 +1399,9 @@ extern "C" {
         envp: *const *mut c_char,
         flags: ::c_int,
     ) -> ::c_int;
+
+    // Added in `glibc` 2.34
+    pub fn close_range(first: ::c_uint, last: ::c_uint, flags: ::c_int) -> ::c_int;
```
SteveLauC, line 1404: This `flags` argument, is of type `unsigned int`, according to the signature given in the man page, but

```sh
$ cd libc-test && cargo t
```
would complain it should have type `int` to me

JohnTitor: glibc version difference? I'd prefer to declare a newer one.

SteveLauC: > glibc version difference?

Not sure about this, let me investigate it a little bit, my glibc is 2.37

```sh
$ ldd --version
ldd (GNU libc) 2.37
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
```

SteveLauC: Just give it a check on glibc

* 2.34
* 2.35
* 2.36

The `flag` argument is of type int, so I guess this is a mistake of the manual

JohnTitor: Seems so, thank you for investigating!

JohnTitor (approved):

PR #3157, opened 2023-03-20
Add pthread_once

DBLouis:

PR #3376, opened 2023-10-07
Add pthread_once

0xcaff:
Close #3157

### DIFF
--- a/libc-test/semver/linux.txt
+++ b/libc-test/semver/linux.txt
@@ -1926,6 +1926,7 @@ PTHREAD_PRIO_PROTECT
 PTHREAD_PROCESS_PRIVATE
 PTHREAD_PROCESS_SHARED
 PTHREAD_STACK_MIN
+PTHREAD_ONCE_INIT
 PTRACE_ATTACH
 PTRACE_CONT
 PTRACE_DETACH
@@ -3373,6 +3374,8 @@ pthread_barrier_wait
 pthread_barrier_destroy
 pthread_barrierattr_t
 pthread_barrier_t
+pthread_once
+pthread_once_t
 ptrace
 ptsname_r
 pwrite64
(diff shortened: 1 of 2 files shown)


0xcaff:
PR in draft while waiting for CI to complete. Maintainer, please feel free to take over and undraft it if the CI is green and I forgot.

0xcaff:
This PR adds pthread_once to linux_like platforms (linux, android, l4re and emscripten). pthread_once seems to be more widely supported than on these linux_like platforms:

* apple's platforms (macOS, iOS, etc.) https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/pthread_once.3.html

* freebsd https://man.freebsd.org/cgi/man.cgi?query=pthread_once&sektion=3&apropos=0&manpath=FreeBSD+7.1-RELEASE

* openbsd https://man.openbsd.org/pthread_once.3

* musl https://git.musl-libc.org/cgit/musl/tree/src/thread/pthread_once.c

* opensolaris https://docs.oracle.com/cd/E19120-01/open.solaris/816-5137/tlib-43636/index.html

Given that this is pretty widely available and not a platform specific feature (pthread_once is part of POSIX), should pthread_once go into src/unix/mod.rs instead? I see pthread_cancel is similar (part of POSIX) but instead of it being defined in src/unix/mod.rs, it is defined in what seems like every platform's module.

I guess it comes down to how bad getting this wrong is. The tests only seem to cover a few platforms (macOS, Windows, Linux) so perhaps we should err on the side of safety by not adding symbols to targets we're not absolutely sure about?

JohnTitor:
If you want to use the item on these platforms, feel free! Note that you don't have to declare an item on all the available platforms exhaustively. You could just move it whenever you need it.

JohnTitor (approved):

JohnTitor:
For now, I'm going to r+, but feel free to move the items if you want :)
@bors r+

PR #3384, opened 2023-10-12
adding ifreq struct for openbsd

devnexen:


src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -533,6 +533,12 @@ s! {
         pub key: *mut ::c_char,
         pub data: *mut ::c_void,
     }
+
+    pub struct ifreq {
+        pub ifr_name: [::c_char; ::IFNAMSIZ],
+        #[cfg(libc_union)]
+        pub ifr_ifru: __c_anonymous_ifr_ifru,
+    }
```
semarie, line 541: in not `libc_union` case, the struct size is wrong and could lead to have `ioctl(2)` write outside the struct.

semarie (changes requested):
I don't have lot of time for reviewing now. I really hope you NEED it, and it isn't just because you see it is missing.

JohnTitor (approved):
Hope you need it, too...

libc-test/build.rs
```diff
@@ -522,7 +522,19 @@ fn test_openbsd(target: &str) {
         "sys/param.h",
     }
 
+    cfg.skip_type(move |ty| {
```
semarie, line 525: this code seems not required (it built fine without it)

semarie (commented):
even if it is already merged, could you adress the two remarks please ?

semarie:
@devnexen ping

devnexen:
sorry I missed that will look into in few hours. cheers.

PR #3393, opened 2023-10-18
feat: Added ifconf struct

Brijeshkrishna:
Adding ifconf struct 

As per defined in C

```c
struct ifconf
  {
    int	ifc_len;			/* Size of buffer.  */
    union
      {
	__caddr_t ifcu_buf;
	struct ifreq *ifcu_req;
      } ifc_ifcu;
  };
```

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -792,6 +792,23 @@ s_no_extra_traits! {
         pub ifr_ifru: ::sockaddr,
     }
 
+    #[cfg(libc_union)]
+    pub union __c_anonymous_ifc_ifcu {
+        pub ifcu_buf: *mut ::c_char,
+        pub ifcu_req: *mut ::ifreq,
+    }
+
+    /*  Structure used in SIOCGIFCONF request.  Used to retrieve interface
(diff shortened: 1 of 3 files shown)


src/unix/linux_like/linux/mod.rs
```diff
@@ -792,6 +792,19 @@ s_no_extra_traits! {
         pub ifr_ifru: ::sockaddr,
     }
 
+    pub union __c_anonymous_ifc_ifcu {
```
devnexen, line 795: since it's a union, it needs to go the `s_extra_no_traits` part, then implementing Eq,PartialEq, ::hash traits. Also for unions, the `#[cfg(libc_union)]` flag needs to be taken in account. You can have a look at the [FreeBSD's implementation](https://github.com/rust-lang/libc/blob/df3f7c1fa899839d4cbd4e969ce9b4d508e6d4d0/src/unix/bsd/freebsdlike/freebsd/mod.rs#L1506) for inspiration.

JohnTitor:
I think you messed up the branch, would be great if you could clean it up and squash commits into one.

JohnTitor:

PR #3395, opened 2023-10-19
linux_like: remove non-sched_priority fields from sched_param for musl

nekopsykose:
for musl, these names were removed in 827aa8fbcac89a63c6efb986871663861500cd13 (back to reserved space).
https://github.com/bminor/musl/commit/827aa8fbcac89a63c6efb986871663861500cd13

since musl doesn't actually touch/implement these options, they shouldn't be exposed.

---

discovered in https://github.com/haileys/bark/issues/4 . afaict, ohos is also musl-1.2 base, so we can drop that too. i can't speak for emscripten, maybe it's similar?

however, merely dropping the cfg like this changes the representation of the struct. afaict we want to match the representation to musl, so we have to keep the padding like musl does. in that case, the definition is:  

```
struct sched_param {
	int sched_priority;
	int __reserved1;
#if _REDIR_TIME64
	long __reserved2[4];
#else
	struct {
		time_t __reserved1;
		long __reserved2;
	} __reserved2[2];
#endif
	int __reserved3;
};
```

i don't know how to map this back to rust

### DIFF
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -108,13 +108,13 @@ s! {
 
     pub struct sched_param {
         pub sched_priority: ::c_int,
-        #[cfg(any(target_env = "musl", target_os = "emscripten", target_env = "ohos"))]
+        #[cfg(any(target_os = "emscripten", target_env = "ohos"))]
         pub sched_ss_low_priority: ::c_int,
-        #[cfg(any(target_env = "musl", target_os = "emscripten", target_env = "ohos"))]
+        #[cfg(any(target_os = "emscripten", target_env = "ohos"))]
         pub sched_ss_repl_period: ::timespec,
-        #[cfg(any(target_env = "musl", target_os = "emscripten", target_env = "ohos"))]
+        #[cfg(any(target_os = "emscripten", target_env = "ohos"))]
         pub sched_ss_init_budget: ::timespec,
-        #[cfg(any(target_env = "musl", target_os = "emscripten", target_env = "ohos"))]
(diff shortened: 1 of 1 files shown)


nekopsykose:
ah, i see https://github.com/rust-lang/libc/pull/1588 touched on this and left a fixme in libc-test..

nekopsykose:
it /should/ be correct to add the correct amount of type padding. for the `#if`, the documentation is in https://github.com/bminor/musl/commit/1febd21d3fb7b78f6fc13e48b3ff2a396a607a15 . what that means for us is we just assume it is true (new binaries).

that makes the struct

```
int sched_priority
int
long [4]
int
```

i can't find a reference on how to add 'invisible' padding (without adding a named type that has to be used). also seemingly these arent `#[repr(C)]`, so i don't know how much this matters or not..

someone that knows more than me would have to elaborate on this, but would be nice to fix the definition, and if the fields are removed for all 3 we can drop the fixme from libc-test too i think

devnexen:
> however, merely dropping the cfg like this changes the representation of the struct. afaict we want to match the representation to musl, so we have to keep the padding like musl does. in that case, the definition is:
> 
> ```
> struct sched_param {
> 	int sched_priority;
> 	int __reserved1;
> #if _REDIR_TIME64
> 	long __reserved2[4];
> #else
> 	struct {
> 		time_t __reserved1;
> 		long __reserved2;
> 	} __reserved2[2];
> #endif
> 	int __reserved3;
> };
> ```
> 
> i don't know how to map this back to rust

I would say in rust's POV, __reserved2 ought to have the same size in both cases ?

PR #3396, opened 2023-10-20
Add ioctl FS_IOC_* version and flag constants

fsavy-tehtris:
This is a tentative to fix https://github.com/rust-lang/libc/pull/3089 because the original author is unlikely to revisit it.

I've added the missing architectures for the constants on Linux, I think they should all be covered now.


JohnTitor (approved):

JohnTitor:
@bors r+
Thanks!

fsavy-tehtris:
I've added the missing `arm` `target_arch` but I was not able to run the Android tests on my machine.

@rustbot review

fsavy-tehtris:
More fixes.
@rustbot review

JohnTitor:

fsavy-tehtris:
Done!

JohnTitor (approved):

fsavy-tehtris:
I _think_ I've fixed the outstanding issues with Android, although the linker does not work in the CI container.
The macOS check broke but I think this is because the related image was updated?

@rustbot review

JohnTitor:
Yeah, the macOS failure is unrelated and has been fixed on main. Let's give another try!
@bors r+

PR #3400, opened 2023-10-21
netbsd/openbsd adding more accessors to siginfo_t.

### LINKED ISSUE #3397: siginfo_t: si_pid, si_status on OpenBSD/NetBSD
POSIX defines the following fields on `siginfo_t` [link](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html):

```
int           si_signo  Signal number. 
int           si_code   Signal code. 
int           si_errno  If non-zero, an errno value associated with 
                        this signal, as described in <errno.h>. 
pid_t         si_pid    Sending process ID. 
uid_t         si_uid    Real user ID of sending process. 
void         *si_addr   Address of faulting instruction. 
int           si_status Exit value or signal. 
long          si_band   Band event for SIGPOLL. 
union sigval  si_value  Signal value.
```

Of particular interest to me are `si_signo`, `si_code` (which are already exposed everywhere), as well as `si_pid` and `si_status`. These allow asynchronously receiving the exit status of a child process via SIGCHLD.

However, on `x86_64-unknown-netbsd` ([documented here](https://man.netbsd.org/siginfo.2)) and `x86_64-unknown-openbsd` ([undocumented but exists here](https://github.com/openbsd/src/blob/master/sys/sys/siginfo.h#L175)) , these fields are not exposed in any way. on `x86_64-unknown-linux-gnu` they are exposed as unsafe methods, and on `x86_64-unknown-freebsd`, as regular struct fields *and* unsafe methods.

I understand that this is due to some implementations choosing to use unions, or having different struct field orders. It would be very helpful if there was a consistent, portable way to access them, or at the very least, expose them on those two targets.

devnexen:
close #3397

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -638,7 +638,7 @@ fn test_windows(target: &str) {
 
             // Windows uppercase structs don't have `struct` in front:
             t if is_struct => {
-                if ty.clone().chars().next().unwrap().is_uppercase() {
+                if ty.chars().next().unwrap().is_uppercase() {
                     t.to_string()
                 } else if t == "stat" {
                     "struct __stat64".to_string()
--- a/src/unix/bsd/netbsdlike/openbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/openbsd/mod.rs
@@ -540,12 +540,46 @@ impl siginfo_t {
         self.si_addr
     }
 
+    pub unsafe fn si_code(&self) -> ::c_int {
+        self.si_code
+    }
+
+    pub unsafe fn si_errno(&self) -> ::c_int {
+        self.si_errno
+    }
+
+    pub unsafe fn si_pid(&self) -> ::pid_t {
+        #[repr(C)]
+        struct siginfo_timer {
+            _si_signo: ::c_int,
+            _si_code: ::c_int,
+            _si_errno: ::c_int,
+            _pad: [::c_int; SI_PAD],
+            _pid: ::pid_t,
+        }
+        (*(self as *const siginfo_t as *const siginfo_timer))._pid
+    }
+
+    pub unsafe fn si_uid(&self) -> ::uid_t {
+        #[repr(C)]
+        struct siginfo_timer {
+            _si_signo: ::c_int,
+            _si_code: ::c_int,
+            _si_errno: ::c_int,
+            _pad: [::c_int; SI_PAD],
+            _pid: ::pid_t,
+            _uid: ::uid_t,
+        }
+        (*(self as *const siginfo_t as *const siginfo_timer))._uid
+    }
+
     pub unsafe fn si_value(&self) -> ::sigval {
         #[repr(C)]
         struct siginfo_timer {
             _si_signo: ::c_int,
-            _si_errno: ::c_int,
             _si_code: ::c_int,
+            _si_errno: ::c_int,
+            _pad: [::c_int; SI_PAD],
             _pid: ::pid_t,
             _uid: ::uid_t,
             value: ::sigval,
@@ -1575,6 +1609,9 @@ pub const NTFS_MFLAG_CASEINS: ::c_int = 0x1;
 pub const NTFS_MFLAG_ALLNAMES: ::c_int = 0x2;
 
 pub const TMPFS_ARGS_VERSION: ::c_int = 1;
+
+const SI_MAXSZ: ::size_t = 128;
+const SI_PAD: ::size_t = (SI_MAXSZ / ::mem::size_of::<::c_int>()) - 3;
 
 pub const MAP_STACK: ::c_int = 0x4000;
 pub const MAP_CONCEAL: ::c_int = 0x8000;
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -60,6 +60,39 @@ impl siginfo_t {
         self.si_addr
     }
 
+    pub unsafe fn si_code(&self) -> ::c_int {
+        self.si_code
+    }
+
+    pub unsafe fn si_errno(&self) -> ::c_int {
+        self.si_errno
+    }
+
+    pub unsafe fn si_pid(&self) -> ::pid_t {
+        #[repr(C)]
+        struct siginfo_timer {
+            _si_signo: ::c_int,
+            _si_errno: ::c_int,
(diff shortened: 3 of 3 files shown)


src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -540,6 +540,39 @@ impl siginfo_t {
...
+        self.si_errno
+    }
+
+    pub unsafe fn si_pid(&self) -> ::pid_t {
+        #[repr(C)]
+        struct siginfo_timer {
+            _si_signo: ::c_int,
+            _si_errno: ::c_int,
+            _si_code: ::c_int,
+            __pad1: ::c_int,
```
semarie, line 557: the `_pad` type size is wrong.

it is defined as `int _pad[SI_PAD]` (an array of `::c_int`), and the number of elements is:
- `#define SI_PAD ((SI_MAXSZ / sizeof (int)) - 3)`
- `#define SI_MAXSZ 128`

src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -540,6 +540,39 @@ impl siginfo_t {
...
+        (*(self as *const siginfo_t as *const siginfo_timer))._pid
+    }
+
+    pub unsafe fn si_uid(&self) -> ::uid_t {
+        #[repr(C)]
+        struct siginfo_timer {
+            _si_signo: ::c_int,
+            _si_errno: ::c_int,
+            _si_code: ::c_int,
+            __pad1: ::c_int,
```
semarie, line 569: same here (wrong type)

semarie (changes requested):
please note that I only checked OpenBSD definitions.

libc-test/build.rs
```diff
@@ -638,7 +638,7 @@ fn test_windows(target: &str) {
 
             // Windows uppercase structs don't have `struct` in front:
             t if is_struct => {
-                if ty.clone().chars().next().unwrap().is_uppercase() {
```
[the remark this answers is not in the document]
devnexen: yes but the build does not pass for me if I don t change it.

src/unix/bsd/netbsdlike/openbsd/mod.rs
devnexen: in fact it s just.a wrong copy paste from netbsd..

src/unix/bsd/netbsdlike/openbsd/mod.rs
devnexen: so even the existing si_value accessor is wrong too (e.g. code/errno inverted).

src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -540,12 +540,46 @@ impl siginfo_t {
...
+        self.si_errno
+    }
+
+    pub unsafe fn si_pid(&self) -> ::pid_t {
         #[repr(C)]
         struct siginfo_timer {
             _si_signo: ::c_int,
+            _si_code: ::c_int,
             _si_errno: ::c_int,
+            _pad: [::c_int; 29],
```
semarie, line 557: sizeof(int) isn't a constant. it isn't the same value on i386 and amd64 for example.

I dunno if rust is able to get the sizeof(int) at build-time.

alternatively, you could define : 
```
pub SI_MAXSZ: ::c_int = 128;
pub SI_PAD: ::c_int = ((SI_MAXSZ / _sizeof_int) - 3);
```

and define a private `_sizeof_int` constant with the right value for each architecture supported.

semarie (changes requested):

src/unix/bsd/netbsdlike/openbsd/mod.rs
devnexen: seems to work with the following commit

semarie: please stick to the C definition.

something like the following should be possible:
```
const SI_MAXSZ: usize = 128;
const SI_PAD: usize = ((SI_MAXSZ / ::mem::size_of::<::c_int>()) - 3);
```

devnexen: oh no hold on my last commit did not go ...

src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -1576,6 +1610,9 @@ pub const NTFS_MFLAG_ALLNAMES: ::c_int = 0x2;
 
 pub const TMPFS_ARGS_VERSION: ::c_int = 1;
 
+pub const SI_MAXSZ: ::size_t = 128;
+pub const SI_PAD: ::size_t = (SI_MAXSZ / ::mem::size_of::<::c_int>()) - 3;
+
```
semarie, line 1615: I would not make them public: there are not used outside the struct definition anywhere inside OpenBSD tree.

semarie (changes requested):
outside the visibility of the constants, I am fine with the OpenBSD version.

src/unix/bsd/netbsdlike/openbsd/mod.rs
devnexen: Hmmm I would object at least for SI_PAD at least don't you think ?

semarie: it is simpler to make it public if need than to hide it later.

but even in C, it wouldn't be needed: it is the size of the padding.
you might need the size of the struct, but not just of a part of the struct (and you could still use `offsetof()` macro). 

here, the "struct" is hidden inside a function. so for me there is really no point to have it public (even if it isn't wrong per se). it is just an internal, or an implementation detail.

devnexen: good point about the struct being within these accessors, ok I ll commit new change.

LunarLambda:
What's stopping this from getting merged?

devnexen:
it will happen at some point no worries :)

JohnTitor (approved):

JohnTitor:
Sorry for the delay! @bors r+

LunarLambda:
thank you very much! would it be possible to publish a release for this?

JohnTitor:
Yeah, I'll make a new release this weekend :)

PR #3401, opened 2023-10-22
feat: Added linux/wireless.h

Brijeshkrishna:
Added linux/wireless.h to unix/linux_like/linux


src/unix/linux_like/linux/mod.rs
```diff
@@ -805,6 +941,45 @@ s_no_extra_traits! {
...
+            pub ap_addr: ::sockaddr,
+            pub addr: ::sockaddr,
+            pub param: iw_param,
+            pub data: iw_point,
+    }
+
+    pub struct iw_event {
+        pub len: __u16,
+        pub cmd: __u16,
+        pub u: iwreq_data,
```
devnexen, line 970: tip: as you set above, iwreq_data falls under the config 'libc_union', if not set then iwreq_data is an unknown type.

src/unix/linux_like/linux/mod.rs
```diff
@@ -805,6 +941,45 @@ s_no_extra_traits! {
...
+    }
+
+    #[cfg(libc_union)]
+    pub union __c_anonymous_iwreq {
+        pub ifrn_name: [c_char; ::IFNAMSIZ],
+    }
+
+    pub struct iwreq {
+        #[cfg(libc_union)]
+        pub ifr_ifrn: __c_anonymous_iwreq,
```
devnexen, line 980: "similarly" if libc_union is not set then iwreq will have the wrong size due to the missing field.

JohnTitor:
Are you sure that you will use _all_ the items here? Adding many items at once makes a review and a potential breaking change hard. Please reduce if you don't need all.

Brijeshkrishna:
Ok, then I will split it into 3 PR.
One for the constants,basic structs and  remaining constants and structs.

PR #3320, opened 2023-08-18
Adding missing macros from linux/if_tun.h

BrandonMFong:
Close https://github.com/rust-lang/libc/issues/3151

I found the missing lines from my debian environment.
```
brandonmfong@kama:~/brando/sources/repo/rust-libc/libc-test$ lsb_release  -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye
```

PR #3405, opened 2023-10-25
feat: add GSO flags for linux and android

elecm:
Add Generic Segmentation Offload (GSO) flags based on [if_tun.h](https://github.com/torvalds/linux/blob/master/include/uapi/linux/if_tun.h#L87) header in Linux  and [if_tun.h](https://android.googlesource.com/platform/bionic/+/HEAD/libc/kernel/uapi/linux/if_tun.h#71) header in Android source code.
Include the `IFF_NO_CARRIER` flag as well (found in the same files mentioned in the previous paragraph).

There is a [PR](https://github.com/rust-lang/libc/pull/3320#issue-1857370692) similar to mine, but that PR doesn't include GSO flags for UDP (`TUN_F_USO4` and `TUN_F_USO6`), and it is only for Linux. And also, I used `c_uint`type for GSO flags because related ioctl function (`#define TUNSETOFFLOAD  _IOW('T', 208, unsigned int)`) get `unsigned int` but that PR uses `c_ushort`.


elecm:
Tests were run on Ubuntu 22.04, but its kernel does not support `IFF_NO_CARRIER`, `TUN_F_USO4`, and `TUN_F_USO6`.
Ubuntu 22.04.3 upgrades its kernel to 6.2 and supports them.

devnexen:
> Tests were run on Ubuntu 22.04, but its kernel does not support `IFF_NO_CARRIER`, `TUN_F_USO4`, and `TUN_F_USO6`. Ubuntu 22.04.3 upgrades its kernel to 6.2 and supports them.

If that is the case, it s necessary to put the aforementioned constants as exceptions [in the skip_const() helper](https://github.com/rust-lang/libc/blob/0644a872dce01c7e744957444efa6aa6a0149848/libc-test/build.rs#L3595).

JohnTitor:

Issue #3317, opened 2023-08-09
Provide zerocopy trait implementations behind feature flag


joshlf:
A common use of `unsafe` code is to access or modify the bytes of libc types (e.g. by transmuting them; [here's an example](https://github.com/bytecodealliance/rustix/blob/main/src/backend/libc/net/ext.rs)). Many libc types could implement the zerocopy traits ([FromBytes](https://docs.rs/zerocopy/latest/zerocopy/trait.FromBytes.html), [AsBytes](https://docs.rs/zerocopy/latest/zerocopy/trait.AsBytes.html), and [Unaligned](https://docs.rs/zerocopy/latest/zerocopy/trait.AsBytes.html)) by using a custom derive, so this would just be a matter of adding `#[cfg_attr(feature = "zerocopy", derive(FromBytes, AsBytes, Unaligned))]` to the types which support it. This would allow all of those uses to be safe.

If this is something that the maintainers would be open to, I'd be happy to put up a PR.

tgross35:
This needs some level of team discussion since it would involve an ecosystem crate. I don't really have any objection for an addition in 1.0 (or later) as long as it is pretty unintrusive.

joshlf:
> This needs some level of team discussion since it would involve an ecosystem crate, I suggested a Zulip discussion on [#3924 (comment)](https://github.com/rust-lang/libc/pull/3924#issuecomment-2356572073). I don't really have any objection for an addition in 1.0 (or later) as long as it is pretty unintrusive.

I've started a [Zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/351149-t-libs.2Fcrates/topic/libc.20optional.20dependency.20on.20zerocopy).

joshlf:
@tgross35 Is there an official process to put this proposal forward somewhere? Not sure whether to take the radio silence on the Zulip thread as disinterest or lack of opposition.

tgross35:
I think the t-libs/crates channel just doesn't get very much traffic. I'll poke things here tomorrow.

joshlf:
Sounds good, thanks!

joshlf:
Per the [Zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/libc.20optional.20dependency.20on.20zerocopy/near/474417305), I'm going to close this for now.

PR #3407, opened 2023-10-25
Add optional zerocopy trait derives

joshlf:


joshlf:
I'm uploading this as a draft early so I can run stuff through CI. Definitely not ready for review yet!

joshlf (commented):
TODO: Update tests to exercise this feature.

joshlf:
@JohnTitor I do a have a question if you happen to have any insight: This PR adds a new Cargo feature which doesn't work on the MSRV of 1.13. I'm looking through the CI infrastructure but don't see an obvious place to put logic to test features only on more recent MSRVs. The README advertises various MSRVs for various features, so I assume such machinery must exist somewhere?

JohnTitor:
You can tweak https://github.com/rust-lang/libc/blob/main/build.rs :)

joshlf:
> You can tweak https://github.com/rust-lang/libc/blob/main/build.rs :)

It looks like the current logic in that file automatically enables features unconditionally if the compiler version is high enough (ie, it doesn't require the user to enable the corresponding Cargo feature). Is that the behavior we want here? I assumed y'all would want this to remain optional, but if that's not the case, I'd be happy to make this unconditional following the logic used with the other features.

JohnTitor:
Ah, hmm, fair enough.
So, let's just declare a normal Cargo feature then (with the MSRV note on README).

joshlf:
Sounds good. Do you have a sense of how I could enable this feature in CI, but only when the toolchain version is 1.61 (zerocopy's MSRV) or greater?

JohnTitor:
It'd be hacky but something like this should be possible:
https://github.com/rust-lang/libc/blob/5e32df0ee7861cfed0ea9c11cfcc59031750b454/ci/build.sh#L69-L79

joshlf:
Sounds good! Does [this](https://github.com/rust-lang/libc/pull/3407/files#diff-0f42fd993ce699d1eaf4007157dd696e37a4ad3f424642cf0e504cfba2faa97e) look good to you?

joshlf:
Also, do these CI tests test on the MSRV (1.13)? I searched for code that enabled it, and found none:

```text
$ rg '1\.13'
README.md
45:The minimum supported Rust toolchain version is currently **Rust 1.13.0**.

ci/build.sh
191:        if [ "${RUST}" != "1.13.0" ]; then

libc-test/build.rs
1198:            "MS_RMT_MASK" => true, // updated in glibc 2.22 and musl 1.1.13
1424:            "MS_RMT_MASK" => true, // updated in glibc 2.22 and musl 1.1.13
```

Cargo.toml
```diff
@@ -130,6 +130,7 @@ cargo-args = ["-Zbuild-std=core"]
 
 [dependencies]
 rustc-std-workspace-core = { version = "1.0.0", optional = true }
+zerocopy = { version = "0.7.16", optional = true, features = ["derive"] }
```
joshlf, line 133: Leaving a TODO for myself:
- [ ] Rename to `zerocopy-0-7`
- [ ] Add `zerocopy-0-6` (to be removed before merge) to test that multiple version trains work

Since libc doesn't specify an edition, we have to use the legacy approach of importing macros via `#[macro_use] extern crate zerocopy;`, which may not play nicely with having macros from multiple versions of the same crate. Instead, we may need to update `build.rs` to detect when the Rust version is early enough that editions are supported, and instruct rustc to use a recent-enough version that we can explicitly import different versions under different names. This would be necessary in case a build requires multiple zerocopy versions simultaneously.

joshlf: See for more discussion: https://github.com/google/zerocopy/issues/557

JohnTitor:
> Also, do these CI tests test on the MSRV (1.13)? I searched for code that enabled it, and found none:

No, I remember that I removed it at some point as something on the test pipeline had been broken.

Cargo.toml
JohnTitor: Once https://github.com/rust-lang/libs-team/issues/72 gets a conclusion, we can specify the edition.

joshlf:
Closing per https://github.com/rust-lang/libc/issues/3317#issuecomment-2389935686

PR #3409, opened 2023-10-27
Add various constants from OpenBSD's sys/exec_elf.h.

ltratt:
This mirrors more of the constants that, for example, are available in linux/mod.rs.

These values are taken from https://github.com/openbsd/src/blob/master/sys/sys/exec_elf.h, though, unsurprisingly, they match those on Linux's `elf.h` (occasionally with different formatting, which I have preserved in this PR).


semarie (approved):
LGTM.
thanks

JohnTitor (approved):

PR #3410, opened 2023-10-27
Use new check-cfg syntax in newer nightly

Urgau:
[MCP636 - Simplify and improve explicitness of the check-cfg syntax](https://github.com/rust-lang/compiler-team/issues/636) introduced a new syntax for check-cfg.

This PR adjust the `build.rs` code to use the new syntax on `rustc >= 75`.
The old syntax is still used on older version since on rust-lang/rust we need to be compatible with both for now.

### DIFF
--- a/.github/workflows/bors.yml
+++ b/.github/workflows/bors.yml
@@ -333,7 +333,7 @@ jobs:
       - name: Setup Rust toolchain
         run: TOOLCHAIN=nightly sh ./ci/install-rust.sh
       - name: Build with check-cfg
-        run: LIBC_CI=1 LIBC_CHECK_CFG=1 cargo build -Z unstable-options -Z check-cfg=features,names,values,output
+        run: LIBC_CI=1 LIBC_CHECK_CFG=1 cargo build -Z unstable-options -Z check-cfg
 
   # These jobs doesn't actually test anything, but they're only used to tell
   # bors the build completed, as there is no practical way to detect when a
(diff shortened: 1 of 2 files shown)


JohnTitor:
I think this would throw the error on _old_ 1.75 nightly, like https://github.com/rust-lang/libc/issues/3066. It'd be great if we could avoid that.

Urgau:
> I think this would throw the error on _old_ 1.75 nightly, like #3066. It'd be great if we could avoid that.

Yes, it would, but I don't know how to avoid that. The version returned by rustc isn't granular enough.

But I also don't think it matter that much since you need the `LIBC_CHECK_CFG=1` env to activate the check cfgs, and as far as I know only rust-lang/rust and rust-lang/libc are setting it.

Urgau:
I've just push a commit to fix the `check_cfg` CI workflow, which [was broken](https://github.com/rust-lang/libc/actions/runs/6642969536/job/18048969844) due to using the old Cargo `-Zcheck-cfg` options.

I also notice that it as been broken for a while, shouldn't bors/CI be gated on it?

JohnTitor:
> I also notice that it as been broken for a while, shouldn't bors/CI be gated on it? Should I gate CI on it?

They actually gate CI, but the first time they noticed it was _after_ merging...: https://github.com/rust-lang/libc/actions/runs/6644020892/job/18052280651
And other PRs are currently blocked.

> > I think this would throw the error on _old_ 1.75 nightly, like #3066. It'd be great if we could avoid that.
> 
> Yes, it would, but I don't know how to avoid that. The version returned by rustc isn't granular enough.
> 
> But I also don't think it matter that much since you need the `LIBC_CHECK_CFG=1` env to activate the check cfgs, and as far as I know only rust-lang/rust and rust-lang/libc are setting it.

Ok then we can go ahead I think, thanks for clarifying!

JohnTitor (approved):

PR #3415, opened 2023-10-29
feat: Added wireless.h constants

Brijeshkrishna:
Added `linux/wireless.h` constants to `Linux`


JohnTitor (approved):

Brijeshkrishna:
Does not rust `nightly` support `const fn` ?

PR #3422, opened 2023-11-03
Add more definitions from linux/tls.h

Arnavion:


src/unix/linux_like/linux/mod.rs
```diff
@@ -3234,6 +3265,41 @@ pub const HWTSTAMP_FILTER_PTP_V2_DELAY_REQ: ::c_uint = 14;
 pub const HWTSTAMP_FILTER_NTP_ALL: ::c_uint = 15;
 
 // linux/tls.h
+pub const TLS_TX: ::c_int = 1;
+pub const TLS_RX: ::c_int = 2;
+
+pub const TLS_1_2_VERSION_MAJOR: ::__u8 = 0x3;
+pub const TLS_1_2_VERSION_MINOR: ::__u8 = 0x3;
+pub const TLS_1_2_VERSION: ::__u16 =
+    ((TLS_1_2_VERSION_MAJOR as ::__u16) << 8) | (TLS_1_2_VERSION_MINOR as ::__u16);
```
Arnavion, line 3274: The actual header uses a macro for this and `TLS_1_3_VERSION`:

```c
#define TLS_VERSION_NUMBER(id)  ((((id##_VERSION_MAJOR) & 0xFF) << 8) | \
                                 ((id##_VERSION_MINOR) & 0xFF))
```

I've inlined the calculation here rather than bind it to a `fn` because I figured there's no reason a user would want to use it.

JohnTitor (approved):

Arnavion:
What does the CI failure mean? That the mips builder's version of the header doesn't define the CHACHA20-POLY1305-related items? The definitions in the header aren't arch-specific so I guess the builder just has old headers [(added in v5.11).](https://github.com/torvalds/linux/commit/923c40c4651ed8b30cbd9fbac0f0ab612216cccc) What should be done?

devnexen:
e.g. `tls12_crypto_info_chacha20_poly1305` is available only from 5.11, you might need to add a bunch of exceptions in `libc-test/build.rs` appropriately.

Arnavion:
Okay, I'll look into that. I tried updating the OpenWRT SDK like https://github.com/rust-lang/libc/commit/7015ee9a42834ea31194ea6b36e3e47f8b39bbdb had done to solve a similar problem in the past, which does pull in new kernel headers (5.15) and fixes the problem, but it also ends up pulling in musl 1.2 which is broken without #1848

Arnavion:
Can I get a bors rerun please?

libc-test/build.rs
```diff
@@ -3593,6 +3593,10 @@ fn test_linux(target: &str) {
             // FIXME: The size of `iv` has been changed since Linux v6.0
             // https://github.com/torvalds/linux/commit/94dfc73e7cf4a31da66b8843f0b9283ddd6b8381
             "af_alg_iv" => true,
+
+            // FIXME: Requires >= 5.11 kernel headers
+            "tls12_crypto_info_chacha20_poly1305" if (mips || mips64) && musl => true,
```
devnexen, line 3598: Did you check there is really no more to add than this ?

Arnavion: Yes, I ran `ci/run-docker.sh` for the mips target locally. The issue is only with Chacha20-Poly1305 definitions, so this one struct and the six consts.

JohnTitor:

Arnavion:
Sure. I was hoping to have the CI run first just in case there are other architectures that are also broken, since the previous CI run cancelled them when mips failed.

Arnavion:
@JohnTitor Can you re-run bors please?

JohnTitor:
👍 @bors r+

Arnavion:
Sigh, i686 failed this time. And again because it cancelled all the other jobs I don't know what else is broken. I'm going to have to run them all locally...

Arnavion:
Every target that uses install-musl.sh also pulls in ancient headers without TLS 1.3 or AES-256-GCM support, so I've updated build.rs to catch those too.

I ran all the targets that CI tests locally and they all pass now. So bors should work too. @JohnTitor Can you rerun it please?

PR #3423, opened 2023-11-03
Add new constants to be used in (linux) sysctl

GuillaumeGomez:
I started using `sysctl` and realized that most constants for it are actually not defined in linux so I added them.


JohnTitor:
CI fails, could you see if the header is imported and if so, skip tests on libc-test?

src/unix/linux_like/linux/mod.rs
```diff
@@ -4029,6 +4029,144 @@ pub const DCCP_SOCKOPT_CCID_TX_INFO: ::c_int = 192;
...
+pub const KERN_BOOTLOADER_TYPE: ::c_int = 67;
+pub const KERN_RANDOMIZE: ::c_int = 68;
+pub const KERN_SETUID_DUMPABLE: ::c_int = 69;
+pub const KERN_SPIN_RETRY: ::c_int = 70;
+pub const KERN_ACPI_VIDEO_FLAGS: ::c_int = 71;
+pub const KERN_IA64_UNALIGNED: ::c_int = 72;
+pub const KERN_COMPAT_LOG: ::c_int = 73;
+pub const KERN_MAX_LOCK_DEPTH: ::c_int = 74;
+
+pub const VM_UNUSED1: ::c_int = 1;
```
devnexen, line 4114: https://lwn.net/Articles/1652/ not sure it is useful to expose those.

devnexen:
overall, you might need to add `linux/sysctl.h` in libc-test/build.rs

src/unix/linux_like/linux/mod.rs
GuillaumeGomez: I suppose not, I'll remove them.

GuillaumeGomez:
Interestingly enough, there are minor differences between linux and android in these constants (`KERN_CAP_BSET` only exist in android, `NET_IRDA`, `KERN_NMI_WATCHDOG` and `KERN_PANIC_ON_NMI` only exist in linux) so I also added them in android.

JohnTitor (approved):

GuillaumeGomez:
Seems like `androidabi` didn't find the `KERN_CAP_BSET` constant. Well, maybe it's too recent so I removed it. We can put it back later on.

JohnTitor:
👍 @bors r+

GuillaumeGomez:
Thanks!

PR #3425, opened 2023-11-06
Deprecate FreeBSD 11 module

JohnTitor:
It reached EoL about 2 years ago and std removed the support, it's time to deprecate and remove it in the future.


JohnTitor:
Let's go ahead for now, we can discuss if we should just remove the module or not later, I think.
@bors r+

JohnTitor:
Hmm, painful, another issue I noticed is that this will emit warnings to std users, which is not desirable. Closing for now to come up with a smarter solution.

PR #3427, opened 2023-11-08
Use a unique name for the Docker image used by CI.

Arnavion:
This allows testing multiple targets in parallel on the same host machine.

### DIFF
--- a/ci/run-docker.sh
+++ b/ci/run-docker.sh
@@ -28,7 +28,7 @@ run() {
     echo "Building docker container for target ${1}"
 
     # use -f so we can use ci/ as build context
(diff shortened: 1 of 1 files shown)


Arnavion:
I needed this because I had to run CI locally for https://github.com/rust-lang/libc/pull/3422 instead of having to wait for someone to approve it, and my machine has enough CPU and RAM to save time by running every target in parallel. It doesn't hurt to always do it even when running only one target as CI does, so I'm upstreaming it.

JohnTitor (approved):

JohnTitor:
Makes sense, thanks! @bors r+

PR #3432, opened 2023-11-11
linux gnu adding pthread_get_minstack helper.

devnexen:


JohnTitor:
I'm concerned that this has a detailed implementation. The libc crate should only be a thin wrapper, and such an implementation should go to other crates (I remember a related topic was discussed _somewhere_, we have a lot of discussing spaces...).

devnexen:
I wish it was a direct call but it s a private symbol. but I understand your concern.

PR #3433, opened 2023-11-11
Add the NI_IDN getnameinfo() extension

adder32:
This PR adds the `NI_IDN` getnameinfo extension.

I've also tried to add the other two extensions (see: https://man7.org/linux/man-pages/man3/getnameinfo.3.html), but the tests would fail (`-Werror`) because those two are deprecated.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -3671,6 +3671,10 @@ fn test_linux(target: &str) {
             if name.starts_with("RLIM64") {
                 return true;
             }
+            // CI fails because musl targets use Linux v4 kernel
+            if name.starts_with("NI_IDN") {
+                return true;
+            }
         }
         match name {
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):

JohnTitor:
Thank you! @bors r+

JohnTitor:
CI fails because musl targets use Linux v4 kernel, could you skip the test for musl? A brief instruction is here: https://github.com/rust-lang/libc/blob/main/CONTRIBUTING.md#test-before-you-commit

adder32:
Let's see if I got it right.

JohnTitor:
Yeah, it should work. Could you squash commits into one?

adder32:
Sure!

JohnTitor:
Thank you! @bors r+

PR #3436, opened 2023-11-12
feat: eventfd_read/write for Android/FreeBSD/Linux

SteveLauC:
Add

```c
int eventfd_read(int fd, eventfd_t *value);
int eventfd_write(int fd, eventfd_t value);
```

for

* Android
* [FreeBSD](https://man.freebsd.org/cgi/man.cgi?query=eventfd&apropos=0&sektion=2&manpath=FreeBSD+13.2-RELEASE&arch=default&format=html)
* [Linux](https://man7.org/linux/man-pages/man2/eventfd.2.html)


SteveLauC:
This FreeBSD 12 CI failure is for the reason that these things are added in FreeBSD 13

From the lesson I learned from my last [PR](https://github.com/rust-lang/libc/pull/3374#pullrequestreview-1662093818), even though a symbol is not available for 12, we should also put it in the `freebsd/mod.rs`

Friendly ask, what should we do about this case :<

devnexen:
> This FreeBSD 12 CI failure is for the reason that these things are added in FreeBSD 13
> 
> From the lesson I learned from my last [PR](https://github.com/rust-lang/libc/pull/3374#pullrequestreview-1662093818), even though a symbol is not available for 12, we should also put it in the `freebsd/mod.rs`
> 
> Friendly ask, what should we do about this case :<

The way I see it, something along those lines in libc-test/build.rs in appropriate places :

```
"eventfd_t" if Some(13) > freebsd_ver => true,
```

same for the calls.

JohnTitor:
Looks good, could you squash commits into one?

SteveLauC:
Accidentally broke my branch, move this PR to #3437 

For the OpenBSD and Solarish maintainers, I am sorry about the ping, please ignore it

PR #3437, opened 2023-11-12
feat: eventfd_read/write for Android/FreeBSD/Linux

SteveLauC:
Add

```c
int eventfd_read(int fd, eventfd_t *value);
int eventfd_write(int fd, eventfd_t value);
```

for

* Android
* [FreeBSD](https://man.freebsd.org/cgi/man.cgi?query=eventfd&apropos=0&sektion=2&manpath=FreeBSD+13.2-RELEASE&arch=default&format=html)
* [Linux](https://man7.org/linux/man-pages/man2/eventfd.2.html)


SteveLauC:
And thanks @devnexen for helping me with the CI failure!

JohnTitor (approved):

PR #3446, opened 2023-11-14
haiku adding a subset of syscalls to be used by fuzzers and calls

devnexen:
interceptors.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -4578,6 +4578,10 @@ fn test_haiku(target: &str) {
     });
 
     cfg.skip_fn(move |name| {
+        // private system calls
+        if name.starts_with("_kern_") {
+            return true;
+        }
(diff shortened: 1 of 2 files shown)


libc-test/build.rs
```diff
@@ -4578,6 +4578,10 @@ fn test_haiku(target: &str) {
     });
 
     cfg.skip_fn(move |name| {
+        // private system calls
```
JohnTitor, line 4581: I remember a similar discussion happened somewhere but private items shouldn't be exposed here, I think.

devnexen: do you mean in the crate or in build.rs ?

JohnTitor: From the crate, I mean, as it's basically unstable.

devnexen: oh ok

PR #3448, opened 2023-11-17
add vm_allocate for apple

hjmallon:
We have vm_deallocate, why not vm_allocate too?

I am not totally sure whether these should live here or in mach2?


JohnTitor (approved):

JohnTitor:
If it's related to mach items it should live in mach2, but indeed it's weird that we don't have `allocate` while we do `deallocate`, so @bors r+

PR #3408, opened 2023-10-26
feat: add new constants from fanotify linux api

ad0:
Since it's introduction in Linux and in this repo (#1699), several new features have been introduced in the fanotify API.
In the original fanotify API, only a limited set of events was supported.  In particular, there was no support for create, delete, and move events.  The support for those events was added in Linux 5.1.

This PR adds missing constants from `fanotify.h` to support the newest features.

PR #3449, opened 2023-11-20
feat: add new struct definitions from fanotify linux api

ad0:
#3408 introduced new constants for the fanotify linux API. Among those constants, some are used to request additional data, which are formatted according to several C structs that are defined in `fanotify.h` but missing from this crate.

As continuation of the first PR, this PR adds missing structures from `fanotify.h` to fully support the newest features.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -3495,7 +3495,8 @@ fn test_linux(target: &str) {
             "type_"
                 if struct_ == "input_event"
                     || struct_ == "input_mask"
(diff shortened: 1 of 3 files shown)


JohnTitor:
Triage: Could you check the test failure?
@rustbot author

tgross35:
https://github.com/rust-lang/libc/pull/3695 added a few of these definitions. Since this has been inactive for a while, I am going to close it.

@ad0 if you get back to this, feel free to open a new PR with whatever is still missing.

PR #3456, opened 2023-11-28
Do not dereference uninhabited types refs in Clone implementations

samueltardieu:
A reference to an uninhabited type should never be dereferenced: this is UB. `Copy` should not be implemented on such a type, and an upcoming Clippy lint (`uninhabited_reference`) may flag such
dereferences as suspicious.

Since those types are not structs, they do not need to get `Copy` and `Clone` implementations. A `missing!` macro limits code duplication.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -120,6 +120,12 @@ macro_rules! s_no_extra_traits {
     );
 }
 
+macro_rules! missing {
+    ($($(#[$attr:meta])* pub enum $i:ident {})*) => ($(
+        $(#[$attr])* #[allow(missing_copy_implementations)] pub enum $i { }
+    )*);
+}
+
 macro_rules! e {
     ($($(#[$attr:meta])* pub enum $i:ident { $($field:tt)* })*) => ($(
         __item! {
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -53,13 +53,9 @@ pub type iconv_t = *mut ::c_void;
 pub type sctp_assoc_t = ::__s32;
 
 pub type eventfd_t = u64;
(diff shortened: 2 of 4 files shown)


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

Nadrieril:
> A reference to an uninhabited type should never be dereferenced: this is UB

This is not quite correct: if the `&!` is safe, then the code can't be reached so dereferencing it is ok. In other words, the `Clone` impl is sound because it can assume that `&self` obeys its safety invariant (which means that `clone` will never be called). A different way to see why this is not UB is that this does not require an `unsafe {}` block.

Now, this `Clone` impl _could_ trigger UB if it was passed a `&self` that breaks the safety invariant. But that's the responsibility of the caller, not the impl.

samueltardieu:
> > A reference to an uninhabited type should never be dereferenced: this is UB
> 
> This is not quite correct: if the `&!` is safe, then the code can't be reached so dereferencing it is ok. In other words, the `Clone` impl is sound because it can assume that `&self` obeys its safety invariant (which means that `clone` will never be called). A different way to see why this is not UB is that this does not require an `unsafe {}` block.
> 
> Now, this `Clone` impl _could_ trigger UB if it was passed a `&self` that breaks the safety invariant. But that's the responsibility of the caller, not the impl.

I agree. However, in the context of libc which necessarily deals with a lot of `unsafe`, obtaining such a reference (for example a `&mut FILE` from a `*mut FILE`) is quite easy and a user could shoot themselves in the foot by calling `.clone()` on them. This is why I think that removing those `Clone` implementations was the best think to do anyway.

Nadrieril:
Oh I see! Then I totally agree, having these impls is a big footgun. Well the whole thing is quite footgunny but that too

PR #3465, opened 2023-12-07
solarish secflags api

devnexen:


devnexen:
image

pfmooney:
These interfaces are not (yet) Committed and probably should not be added to the libc crate like this.

pfmooney:
Because the secflags stuff is not yet Committed when it comes to interface stability, it could be subject to change at any time.  That would be unfortunate for anyone who had begun using it from the libc crate, thinking it could be relied upon.

devnexen:
oh ok I see.

pfmooney:
> oh ok I see.

If you want to play with the secflags stuff, by all means feel free to do so in your own crate/project.  It's fine to build rust things on top of Uncommitted interfaces *if you're willing to accept the risk of them changing in the future*.  Putting `libc` consumers unknowingly at risk of such breakage is not acceptable, though.

Cheers

Issue #3462, opened 2023-12-04
epoll event constants should be declared as u32


tremon015:
The current implementation declares all EPOLL* event constants as `c_int`, but this isn't very useful given that the accompanying field in struct epoll_event is declared as u32.

```
pub const EPOLLIN: c_int = 0x1;

pub struct epoll_event {
    pub events: u32,
    pub u64: u64,
}
```

This makes using these constants very unergonomic:

```
error[E0308]: mismatched types
   --> src/bin/hld.rs:256:46
    |
256 |          if event.events & EPOLLIN == EPOLLIN {
    |                            ^^^^^^^ expected `u32`, found `i32`

error[E0277]: no implementation for `u32 & i32`
   --> src/bin/hld.rs:256:44
    |
256 |          if event.events & EPOLLIN == EPOLLIN {
    |                          ^ no implementation for `u32 & i32`
    |
```

The first thing every user of these constants will do is redefine them as u32. Can this be done in the libc crate itself? This applies to all EPOLL* event descriptors and input flags, but not the EPOLL_CTL_* operation constants.

tremon015:
Taking a look at the kernel source itself, all values are of type `__poll_t`: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/eventpoll.h

And the __poll_t type is
```
typedef unsigned __bitwise __poll_t;
```
from https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/types.h#n59

tremon015:
Well, I can create a pull request with the straightforward s/// easily. However, I'm completely unsure of the further ramifications of this change. So I would really appreciate it if other people approach this request with due care :p

PR #3466, opened 2023-12-08
update epoll constants to match epoll_event struct

tremon015:
Change the type of the EPOLL* event constants to match the type of the events field in struct epoll_event (i.e. u32).

closes: #3462

### DIFF
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -1067,21 +1067,21 @@ pub const UIO_MAXIOV: ::c_int = 1024;
 
 pub const FD_SETSIZE: usize = 1024;
 
-pub const EPOLLIN: ::c_int = 0x1;
-pub const EPOLLPRI: ::c_int = 0x2;
-pub const EPOLLOUT: ::c_int = 0x4;
-pub const EPOLLERR: ::c_int = 0x8;
-pub const EPOLLHUP: ::c_int = 0x10;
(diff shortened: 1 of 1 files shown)


tremon015:
This should probably be considered a breaking change, but I'm unsure how to address that properly. I don't think deprecating the current constants and introducing new ones to reflect the type change is in any way a proper solution. Anyone using these constants in the context of epoll_event will already have these values wrapped as `(constant as u32)`, so for these cases this change will not cause any breakage. However, I'm not convinced there are no other uses for these constants outside of the epoll_event context.

tremon015:
I guess that depends on whether we want kernel types to be exposed directly through the libc create. For as far as I can see, the `__poll_t` type is part of the Linux syscall interface, but not the glibc wrapper. The type is exposed via <linux/types.h> so it is available to userspace, but the glibc header file <sys/epoll.h> does not use the `__poll_t` type at all, it just redefines the values in an enum:

```
enum EPOLL_EVENTS
  {
    EPOLLIN = 0x001,
[..]
    EPOLLET = 1u << 31
  };
```

What would be the rust equivalent of the bitwise attribute?

devnexen:
you re right then u32 seems good enough as is.

samueltardieu (approved):
Seems safe enough, and it is unlikely existing code will break, it may get warnings though which will be easy to fix.

JohnTitor (approved):

PR #3468, opened 2023-12-09
netbsd adding mcontext related data for riscv64

devnexen:
ref: https://github.com/NetBSD/src/blob/0465e5c825effb130eca95499f5ac6454fe0d5ab/sys/arch/riscv/include/mcontext.h#L44

### DIFF
--- a/src/unix/bsd/netbsdlike/netbsd/riscv64.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/riscv64.rs
@@ -3,7 +3,26 @@ use PT_FIRSTMACH;
 pub type c_long = i64;
 pub type c_ulong = u64;
 pub type c_char = u8;
+pub type __greg_t = u64;
 pub type __cpu_simple_lock_nv_t = ::c_int;
+pub type __gregset = [__greg_t; _NGREG];
+pub type __fregset = [__freg; _NFREG];
+
+s! {
+    pub struct mcontext_t {
+        pub __gregs: __gregset,
+        pub __fregs: __fpregset,
+        __spare: [::__greg_t; 7],
+    }
+}
+
+s_no_extra_traits! {
+    #[cfg_attr(feature = "extra_traits", allow(missing_debug_implementations))]
+    pub union __fpreg {
+        pub u_u64: u64,
+        pub u_d: ::c_double,
+    }
+}
 
 pub(crate) const _ALIGNBYTES: usize = ::mem::size_of::<::c_long>() - 1;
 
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):

JohnTitor:
Could you resolve conflicts and remove libc_union cfg?

src/unix/bsd/netbsdlike/netbsd/aarch64.rs
```diff
@@ -9,6 +9,8 @@ pub type __cpu_simple_lock_nv_t = ::c_uchar;
 s! {
     pub struct __fregset {
         pub __qregs: [__c_anonymous__freg; 32],
+        #[cfg(not(libc_union))]
+        pub __qregs: [u128; 1],
```
JohnTitor, line 13: The main branch doesn't need cfg anymore.

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
```diff
@@ -3,11 +3,108 @@ use PT_FIRSTMACH;
...
+            }
+        }
+    }
+    if #[cfg(libc_const_size_of)] {
+        #[doc(hidden)]
+        pub const _ALIGNBYTES: usize = ::mem::size_of::<::c_long>() - 1;
+    } else {
+        #[doc(hidden)]
+        pub const _ALIGNBYTES: usize = 8 - 1;
+    }
```
JohnTitor, line 60: `pub(crate)` can be used I think.

tgross35:
This probably needs a rebase at this point, but otherwise seems like it was waiting for review.

@rustbot review

tgross35:
Can this just be moved to netbsd/mod.rs? I'm not sure what is platform-specific here, if you have a header link that would help.

This should also get the semver test

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
```diff
@@ -3,11 +3,94 @@ use PT_FIRSTMACH;
...
+    if #[cfg(feature = "extra_traits")] {
+        impl PartialEq for __c_anonymous__fpreg {
+            fn eq(&self, other: &__c_anonymous__fpreg) -> bool {
+                unsafe {
+                self.u_u64 == other.u_u64
+                    || self.u_ud == other.u_ud
+                }
+            }
+        }
+        impl Eq for __c_anonymous__fpreg {}
```
tgross35, line 34: These should probably just be dropped, it would be weird to use float equality to compare integers.

tgross35 (changes requested):
I think the trait should be dropped as mentioned. Otherwise lgtm, please add sources to the PR description.

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
```diff
@@ -3,11 +3,72 @@ use PT_FIRSTMACH;
...
+        __spare: [::__greg_t; 7],
+    }
+}
+
+s_no_extra_traits! {
+    #[cfg_attr(feature = "extra_traits", allow(missing_debug_implementations))]
+    pub union __c_anonymous__fpreg {
+        pub u_u64: u64,
+        pub u_d: ::c_double,
+    }
```
tgross35, line 22: Sorry, one other thing - should this just be called `__fpreg`? From https://github.com/NetBSD/src/blob/0465e5c825effb130eca95499f5ac6454fe0d5ab/sys/arch/riscv/include/mcontext.h#L44C7-L47.

It would probably also be good to define 

```rust
pub const _NGREG: usize = 32;
pub const _NFREG: usize = 33;
pub type __gregset_t = [__greg_t; _NGREG];
pub type __fregset_t = [__fpreg; _NFREG];
```

and then use those in mcontext

tgross35 (approved):
Thanks!

PR #3473, opened 2023-12-10
proposal to separate netbsd 10 release from its predecessors which

devnexen:
will be still used for a while.

One of the visible example is the kevent using finally a void pointer like other BSD.


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "libc"
-version = "0.2.150"
+version = "0.2.151"
```
JohnTitor, line 3: Could you drop this change?

JohnTitor:
Maybe we could drop the support on libc 0.3 instead of adding cfg complexity.

devnexen:
> Maybe we could drop the support older NetBSDs on libc 0.3 instead of adding cfg complexity.

a bit torn, but when 0.3 would be released ? netbsd 10 is not really released yet (rc2 only for now).

devnexen:
> Maybe we could drop the support older NetBSDs on libc 0.3 instead of adding cfg complexity.

true but in the field netbsd 8/9 are not going away anytime soon tough.

PR #3474, opened 2023-12-11
update CI freebsd 14 to proper release.

devnexen:

PR #3475, opened 2023-12-11
Stop testing on FreeBSD 12

asomers:
FreeBSD 12.4 will be EoL at the end of the month.  Stop testing on it.  Also, update the 14.0 image to the official release.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,18 +1,4 @@
 task:
-  name: nightly x86_64-unknown-freebsd-12
-  freebsd_instance:
-    image_family: freebsd-12-4
-  setup_script:
-    - pkg install -y libnghttp2 curl
-    - curl https://sh.rustup.rs -sSf --output rustup.sh
-    - sh rustup.sh --default-toolchain nightly -y --profile=minimal
-    - . $HOME/.cargo/env
-  test_script:
-    - . $HOME/.cargo/env
-    - LIBC_CI=1 sh ci/run.sh x86_64-unknown-freebsd
-    - sh ci/run.sh x86_64-unknown-freebsd
-
-task:
   name: nightly x86_64-unknown-freebsd-13
   freebsd_instance:
     image_family: freebsd-13-2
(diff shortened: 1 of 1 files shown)


devnexen:
There is a [PR for it](https://github.com/rust-lang/libc/pull/3474).

.cirrus.yml
```diff
@@ -29,7 +15,7 @@ task:
 task:
   name: nightly x86_64-unknown-freebsd-14
   freebsd_instance:
-    image: freebsd-14-0-rc1-amd64
+    image: freebsd-14-0-release-amd64-ufs
```
devnexen, line 18: now there is a proper 14 release, `image_family: freebsd-14-0` is sufficient.

.cirrus.yml
```diff
@@ -1,17 +1,3 @@
-task:
```
devnexen, line 1: it s tricky ... logically you re right but in the wild there are many users still taking time to migrate (in another project I had a freebsd ... 9.2 user, a bit extreme :) but you get me ..). I may suggest to give a bit of time before removing it but not up to me..

asomers: 9.2?  Wow.  My feeling is that while some users may take some time to upgrade, they also probably aren't using cutting edge Rust software.  So those users can probably make do with older crates.  No need to maintain the latest Rust releases for old FreeBSD release.  At least, that's my philosophy.

devnexen: Fair enough. It seems it takes more than this tough see [what it took for freebsd 11](https://github.com/rust-lang/homu/pull/169).

JohnTitor:
r+ once https://github.com/rust-lang/homu/pull/199 gets merged.

.cirrus.yml
JohnTitor: Note that removing CI doesn't mean that the existing usage on FreeBSD 12 will be broken. The impact of this should be low, I think.

JohnTitor (approved):

PR #3487, opened 2023-12-19
Add `ifreq`, `ifconf` and related constants to Android

arctic-alpaca:
Hi,

this adds `ifreq`, `ifconf` and all related constants defined in Android to Android.
I also added missing symbols to `semver/linux.txt`.


JohnTitor (approved):

arctic-alpaca:
I missed adding some fields to the skip section. While investigating that, I found `test_android()` calls `cfg.skip_field()` twice. The second call overwrites the function set in the first call. I removed the first call and all its contents as it was not used since https://github.com/rust-lang/libc/commit/ab1472eb086cea9d23aa267206dc040feefa673c which introduced the second call.

PR #3488, opened 2023-12-19
Add `newer-api` Cargo feature

JohnTitor:
This is an attempt to accept the `breakage-candidate`s without a major release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -139,6 +139,7 @@ align = []
 rustc-dep-of-std = ['align', 'rustc-std-workspace-core']
 extra_traits = []
 const-extern-fn = []
+newer-api = []
 # use_std is deprecated, use `std` instead
 use_std = ['std']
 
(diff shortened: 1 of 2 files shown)


JohnTitor:
I expect this will gate breaking changes (adding members to struct, changing const's values, etc...) but it would add complexity, any opinions or better options are definitely welcome!
cc @rust-lang/crate-maintainers I'd like to hear anyone's opinion.

Amanieu:
I don't think this will work due to the way Cargo feature unification works. Enabling this feature in one crate could end up breaking a completely unrelated crate. Considering how widely used libc is, I think this will be too brittle for any practical use.

JohnTitor:
Hmm, I see! Do you have any other ways to accept breakage candidates until we can release libc 0.3? 🤔 (Or, would releasing libc 0.3 be the best way here?)

Amanieu:
I think the best way forward would be to follow the implementation plan in #3248.

JohnTitor:
Sure, let's follow it then!

PR #3490, opened 2023-12-19
Add waitid function for OpenBSD

lcheylus:
To compile on OpenBSD a Rust project using [shared_child](https://docs.rs/shared_child/) crate, I need `waitid` libc function.

- Same prototype for `waitid` function on NetBSD and OpenBSD.
- To support it on OpenBSD, move definition from `src/unix/bsd/netbsdlike/netbsd/mod.rs` to `src/unix/bsd/netbsdlike/mod.rs` => available on both BSD OS

### DIFF
--- a/libc-test/semver/openbsd.txt
+++ b/libc-test/semver/openbsd.txt
@@ -1274,4 +1274,5 @@ utimensat
 utmp
 utrace
 wait4
+waitid
 xucred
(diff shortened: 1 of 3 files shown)


JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #3493, opened 2023-12-20
removing freebsd10/11 support proposal.

devnexen:


JohnTitor:
This has some duplicates with https://github.com/rust-lang/libc/pull/2406, let's track this topic there.

PR #3494, opened 2023-12-21
Add linux support for `sched_attr` and related syscalls

matthri:
Hi, 
I this PR should add support for the `sched_attr` config struct and the related `sched_getattr` and `sched_setattr` syscalls.
Since this is my first PR to this repo and I'm not that experienced  with the different standard libraries I encountered some problems and wanted to ask for some assistance/ hints on how this  best implemented in the interests of this project. 

1. **Header file conflict**: The `sched_attr` struct is defined in `linux/sched/types.h` but when I add this file to the includes in the `build.rs` file I get some build errors since the file additionally defines the `sched_param` struct which is also defined by the `bits/types/struct_sched_param.h` header which is included by `pthread.h`.
Can you point me in the right direction how this is best resolved?
2. **Multiple struct versions**: There exist multiple versions of the `sched_attr` struct which are distinguished by the `size` field of the struct. 
Whats in your opinion the best way to represent this in this crate?

Many thanks in advance and sorry for the wall of text above :D

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -850,6 +850,19 @@ s_no_extra_traits! {
         pub d_type: ::c_uchar,
         pub d_name: [::c_char; 256],
     }
+
+    pub struct sched_attr {
+        pub size: ::c_uint,
+        pub sched_policy: ::c_uint,
+        pub sched_flags: ::c_ulonglong,
+        pub sched_nice: ::c_int,
+        pub sched_priority: ::c_uint,
+        pub sched_runtime: ::c_ulonglong,
+        pub sched_deadline: ::c_ulonglong,
+        pub sched_period: ::c_ulonglong,
+        pub sched_util_min: ::c_uint,
+        pub sched_util_max: ::c_uint,
+    }
 }
 
 s_no_extra_traits! {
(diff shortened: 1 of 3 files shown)


src/unix/linux_like/linux/mod.rs
```diff
@@ -4520,6 +4536,20 @@ pub const NET_DCCP: ::c_int = 20;
...
+        pid: ::pid_t,
+        attr: *mut ::sched_attr,
+        size: ::c_uint,
+        flags: ::c_uint
+    ) -> ::c_int {
+        ::syscall(::SYS_sched_getattr, pid, attr, size, flags) as ::c_int
+    }
+    pub fn sched_setattr(pid: ::pid_t, attr: *mut ::sched_attr, flags: ::c_uint) -> ::c_int {
+        ::syscall(::SYS_sched_setattr, pid, attr, flags) as ::c_int
+    }
```
JohnTitor, line 4551: The libc crate is a thin wrapper, and if glibc doesn't expose it, it shouldn't be added.

JohnTitor (changes requested):

src/unix/linux_like/linux/mod.rs
matthri: Thanks for your response. 
Before opening this PR I had seen that there are already some methods in the android target which are not exposed and use the syscall directly, so I hoped this might also be applicable here.

E.g.
https://github.com/rust-lang/libc/blob/e077dbfe7faec63e7ea315722a596f9c062f0701/src/unix/linux_like/android/b32/x86/mod.rs#L591-L615

matthri: I understand, thanks again for your time.

PR #3495, opened 2023-12-22
Simplify build.rs by using the question mark operator

samueltardieu:
The `question_mark` clippy lint currently overlooks some patterns that could be replaced by the question mark operator. This will be fixed shortly by https://github.com/rust-lang/rust-clippy/pull/11994.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -231,20 +231,14 @@ fn rustc_minor_nightly() -> (u32, bool) {
 }
 
 fn which_freebsd() -> Option<i32> {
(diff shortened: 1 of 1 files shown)


JohnTitor:
Thanks! But IIRC the question mark operator on Option was stabilized on 1.22 and this would break 0.2's MSRV, I think.

Once I create a new release (it should be the final release from the main branch, I think), I'm going to r+ this.

JohnTitor (approved):

JohnTitor:
It's time to go! @bors r+

PR #3496, opened 2023-12-24
fix typos in libc

Takashiidobe:
Fixing some typos:

For the change in `src/unix/bsd/apple/mod.rs`, I don't think changing the name of a param changes ABI so I think it's safe to change, but could be wrong. I believe the correct spelling is here: https://opensource.apple.com/source/Libc/Libc-498/darwin/libproc.c.auto.html which indicates it should be `minor`.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -3281,7 +3281,7 @@ fn test_linux(target: &str) {
 
     let mut cfg = ctest_cfg();
     cfg.define("_GNU_SOURCE", None);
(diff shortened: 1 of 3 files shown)


JohnTitor (approved):

JohnTitor:
> For the change in `src/unix/bsd/apple/mod.rs`, I don't think changing the name of a param changes ABI so I think it's safe to change

Correct, thanks!

@bors r+

PR #3498, opened 2023-12-24
Improve the version parser of Emscripten

kleisauke:
Some Emscripten versions come with `-git` attached, so split the version string also on the `-` char.

See: https://github.com/rust-lang/rust/issues/119250.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -274,7 +274,10 @@ fn emcc_version_code() -> Option<u64> {
         return None;
     }
     let version = stdout.unwrap();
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):

JohnTitor:
Cool, thanks! @bors r+

kleisauke:
It looks like [`trim_end_matches`](https://doc.rust-lang.org/std/primitive.str.html#method.trim_end_matches) is only available since Rust 1.30.0. I just simplified this PR to avoid that function.

PR #3499, opened 2023-12-25
Add group calls to android

Takashiidobe:
Following https://github.com/rust-lang/libc/issues/3014, this PR just implements the group side of the calls, since that doesn't require CI to be updated.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2015,6 +2015,15 @@ fn test_android(target: &str) {
             // Added in API level 28, but some tests use level 24.
             "aligned_alloc" => true,
 
+            // Added in API level 26, but some tests use level 24.
+            "getgrent" => true,
+
+            // Added in API level 26, but some tests use level 24.
+            "setgrent" => true,
+
+            // Added in API level 26, but some tests use level 24.
+            "endgrent" => true,
+
             // FIXME: bad function pointers:
             "isalnum" | "isalpha" | "iscntrl" | "isdigit" | "isgraph" | "islower" | "isprint"
             | "ispunct" | "isspace" | "isupper" | "isxdigit" | "isblank" | "tolower"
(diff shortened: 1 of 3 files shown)


JohnTitor (approved):

Takashiidobe:
perfect, looks like I get this error message on arm-linux-androideabi, time to fix it up

```
stderr ---
WARNING: linker: /data/local/tmp/main-418f900fe48d7afe: unsupported flags DT_FLAGS_1=0x8000001
CANNOT LINK EXECUTABLE "/data/local/tmp/main-418f900fe48d7afe": cannot locate symbol "setgrent" referenced by "/data/local/tmp/main-418f900fe48d7afe"...
```

tgross35:
@Takashiidobe would you be able to rebase? I can help figure out the problems if they still exist.

Takashiidobe:
Still get the same error message, and I think I remember why.

[CI is on API 24](https://github.com/rust-lang/libc/blob/ffb7f0c05718643d06bc20b971cedf20d6928d93/ci/android-install-sdk.sh#L18-L25) which is before these calls were introduced ([API 26](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/grp.h;l=56-59?q=grp.h&ss=android%2Fplatform%2Fsuperproject%2Fmain)) so this PR won't pass.

I'm not sure if there's a requirement on API level being capped at 24 (if so, then this PR shouldn't pass until that's changed) or if that's a CI limitation?

tgross35:
You should just be able to exclude tests for the relevant constants in `libc-test/build.rs` the `test_android` function.

Cc @maurer (android change)

tgross35 (approved):
LGTM but I'll wait for maurer to take a look. Please squash & rebase.

maurer:
This is fine/appropriate to put into `libc` in case it's needed for compatibility.

To @Takashiidobe - these functions are performance footguns - Android has too many AIDs, so iterating through them in this fashion can take a nontrivial amount of time, ID based functions are preferred. Iterating via `getgrent` may take up to an eighth of a second to get through everything. They were primarily added for compatibility and tests.

PR #3500, opened 2023-12-25
update android sdk version

Takashiidobe:


JohnTitor:
Note that CI on PRs doesn't trigger all the jobs, you could verify if it works by running run-docker.sh like CI.
You also have to add more tweaks, see https://github.com/rust-lang/libc/pull/3439 (N.B. it's incomplete!).

Takashiidobe:
Will reopen when the android CI is fixed, I couldn't figure out how to fix it.

PR #3502, opened 2023-12-27
Add support for posix_spawn on OpenBSD

nuudlman:


src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -2140,6 +2144,85 @@ extern "C" {
 
     pub fn mimmutable(addr: *mut ::c_void, len: ::size_t) -> ::c_int;
 
+    // Added in `OpenBSD 5.2`
```
semarie, line 2147: I am unsure that the comment with the history add lot of value

src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -2140,6 +2144,85 @@ extern "C" {
...
+    // spawn.h
+    pub fn posix_spawn(
```
semarie, line 2149: I think that lot of functions here could be pushed at higher level and being defined only at one place. It would mean removing the definitions in `netbsdlike/netbsd/mod.rs`, and put them in `netbsdlike/mod.rs` to match netbsd and openbsd (and eventually move it higher, but it will be more work).

for the types, having them here is fine (the implementation are different across systems).

semarie (approved):
just some nits, but the definitions are right.
LGTM

src/unix/bsd/netbsdlike/openbsd/mod.rs
nuudlman: I was just copying what I saw earlier, but I agree

src/unix/bsd/netbsdlike/openbsd/mod.rs
nuudlman: I pulled out the common constants as well (netbsd seems to define 0x40 as well as the posix ones)

semarie (approved):
LGTM thanks.

JohnTitor:

nuudlman:
Done

JohnTitor (approved):

JohnTitor:
Thanks! @bors r+

PR #3505, opened 2023-12-28
Add constants from <include/linux/sched.h>

GuillaumeGomez:
Needed for `sysinfo` as usual. :)

Constants come from https://github.com/torvalds/linux/blob/master/include/linux/sched.h.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -4056,6 +4056,33 @@ fn test_linux(target: &str) {
                 true
             }
 
+            // FIXME: seems to not be available all the time (from <linux/include/sched.h>:
+            "PF_VCPU"
+            | "PF_IDLE"
+            | "PF_EXITING"
+            | "PF_POSTCOREDUMP"
+            | "PF_IO_WORKER"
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
I had to ignore the constants because they are not defined in `linux/sched.h`, no idea why. It's strange because it's the same file but only up to a certain point...

libc-test/build.rs
```diff
@@ -4056,6 +4056,33 @@ fn test_linux(target: &str) {
                 true
             }
 
+            // FIXME: seems to not be available all the time (from <linux/include/sched.h>:
```
JohnTitor, line 4059:
```suggestion
            // FIXME: seems to not be available all the time (from <include/linux/sched.h>:
```

Seems swapped, but yeah, it's weird that 23.10 couldn't find them which has the v6.5 kernel header 🤔

JohnTitor (approved):

GuillaumeGomez:
Thanks for the comment fix and the quick review!

PR #3513, opened 2024-01-05
Add more items from `include/linux/sched.h` header

GuillaumeGomez:
More items coming from the header file `include/linux/sched.h`.

### DIFF
--- a/.github/workflows/full_ci.yml
+++ b/.github/workflows/full_ci.yml
@@ -1,0 +1,348 @@
+name: full CI
+
+on:
+  merge_group:
+  push:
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Finally was able to fix CI. I'm not super happy that the new struct is ignored in tests though, but nothing we can do about it currently...

JohnTitor (approved):

JohnTitor:
The comment is quite helpful when removing, thanks!
@bors r+

GuillaumeGomez:
Seems like not all constants are available on all tested linux. I'll add some cfg conditions to fix it.

PR #3523, opened 2024-01-07
Remove PSP-related code

JohnTitor:
Removes the support to avoid some concerns.

For users: If you depend on the source here, please create a separate crate and use it instead. Sorry for the inconvenience.


overdrivenpotato:
@JohnTitor Out of curiosity, what were the concerns with keeping these function stubs in libc? Is there anything we can do as the `rust-psp` project, to help keep these definitions in?

tgross35:
> @JohnTitor Out of curiosity, what were the concerns with keeping these function stubs in libc? Is there anything we can do as the `rust-psp` project, to help keep these definitions in?

I am pretty late to the picture here but are there any possible licensing issues? It seems like that has come up a bit recently.

JohnTitor:
@tgross35 Yeah, correct.

@overdrivenpotato If you want to know what the issue is, you can contact the Foundation members who manage the licensing issues.

PR #3524, opened 2024-01-07
Bump MSRV to 1.71

JohnTitor:
Bump MSRV to 1.71 whose std has the musl 1.2 support.
This doesn't remove cfg hacks not to make this a _super_ PR.


notgull (commented):
Should this PR also update `rust-version` in `Cargo.toml`?

PR #3525, opened 2024-01-07
Remove cfg hacks for old MSRV

JohnTitor:


semarie:
FYI, it is fine on openbsd

Cargo.toml
```diff
@@ -14,6 +14,7 @@ exclude = ["/ci/*", "/.github/*", "/.cirrus.yml", "/triagebot.toml"]
 description = """
 Raw FFI bindings to platform libraries like libc.
 """
+rust-version = "1.71.0"
```
bjorn3, line 17: Maybe update to the 2021 edition too? The 2015 edition has quite different name resolution rules that are foreign for people having learned rust in 2018 or later as well as anyone who has forgotten the 2015 rules.

JohnTitor: Good point, we should migrate to it, yeah.

JohnTitor: But the problem is that ctest2 uses a quite old librustc_syntax fork and doesn't support newer syntax at all. We have to rewrite ctest2 first :(

djc:
It's very surprising to me that there's no discussion or reference here to what the new MSRV should be. Was this discussed in an issue? If so, please link the issue in the PR description.

In my opinion it would be nice to stick to 1.63 for the time being, matching Tokio and a bunch of other downstreams that would like to support stable Debian.

Amanieu:
This change only applies to the v1.0 branch (main) which is not released yet. The v0.2 branch still has the old MSRV.

ChrisDenton:
Maybe the version should be changed on the main branch to avoid further confusion?

PR #3527, opened 2024-01-08
proposal to separate netbsd 10 release from its predecessors which

devnexen:
will be still used for a while.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -18,6 +18,7 @@ const ALLOWED_CFGS: &'static [&'static str] = &[
     "libc_const_extern_fn_unstable",
     "libc_deny_warnings",
     "libc_thread_local",
+    "netbsd10",
 ];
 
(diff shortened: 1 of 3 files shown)


JohnTitor:
We should discuss and define the support policy for NetBSD before this.

tgross35:
(That being said, it would still be nice to document in the netbsd platform docs that we intend to support the stable releases)

tgross35:
You’re right, there is no ABI change There are just some nonbreaking changes in C that map to breaking changes in Rust, and things that rely on implicit conversion are one of them - as is the case here :)

Bringing thing in line with the other BSDs is worth it though, and the amount of people that will actually notice is probably quite small (users are likely `as’ casting the field to a pointer on NetBSD which means they wouldn’t notice a change) so it’s not too big of a deal.

PR #3531, opened 2024-01-08
Add additional Linux AF_PACKET options (libc 0.2)

nathaniel-bennett:
Added socket options, structs and other miscellaneous constants that make memory-mapped packet sockets possible in Linux (PACKET_TX_RING, PACKET_RX_RING, etc.).


nathaniel-bennett:
Additional commits were accidentally pulled in--fixing this in a new PR

PR #3533, opened 2024-01-10
Remove mach-related items

JohnTitor:
...as we recommend using mach2 instead.


bdash:
I'm confused as to why structures related to the Mach-O file format are being removed. They're not part of the Mach kernel APIs. `mach2` does not provide these types, AFAICT, and it claims to provide `/usr/include/mach`, not `/usr/include/mach-o`.

`mach_header` / `mach_header_64` are part of the definition of the Mach-O file format. A random subset of other Mach-O structures still remain: `segment_command`, `load_command`, and related constants such as `MH_MAGIC` and `LC_SEGMENT`.

PR #3534, opened 2024-01-10
Provide the same sched.h constants for musl as gnu

alyssais:
I don't understand why these constants from Linux headers aren't just defined in linux_like/mod.rs.  Surely they'd be the same for every libc?


JohnTitor:
CI fails: https://github.com/rust-lang/libc/actions/runs/7487138095/job/20379076829
It'd also be great if you could rebase onto main to trigger full CI.

tgross35:
@alyssais are you able to update this to get it over the line? If you need help figuring out the CI, let me know.

tgross35:
@alyssais ping, are you able to update this?

src/unix/linux_like/linux/mod.rs
```diff
@@ -5633,6 +5633,10 @@ pub const SCHED_FLAG_KEEP_PARAMS: c_int = 0x10;
 pub const SCHED_FLAG_UTIL_CLAMP_MIN: c_int = 0x20;
 pub const SCHED_FLAG_UTIL_CLAMP_MAX: c_int = 0x40;
 
+pub const CLONE_NEWTIME: c_int = 0x80;
+pub const CLONE_CLEAR_SIGHAND: c_ulonglong = 0x100000000;
+pub const CLONE_INTO_CGROUP: c_ulonglong = 0x200000000;
+
```
tgross35, line 5639: Actually `CLONE_NEWTIME` is defined at https://github.com/kraj/musl/blob/1b06420abdf46f7d06ab4067e7c51b8b63731852/include/sched.h#L52, but `CLONE_CLEAR_SIGHAND` and `CLONE_INTO_CGROUP` do not seem to be defined in Musl

alyssais: That's true, but given the values are passed straight to the kernel anyway, does it matter?

tgross35: We shouldn't really be peeking into what the `libc`s do with constants, i.e. whether the values go right to the kernel or if they get intercepted. Also it is a CI failure.

alyssais: As far as I can tell Glibc doesn't define `CLONE_CLEAR_SIGHAND` in a public header either — only the linux/sched.h kernel header and the internal linux/clone3.h header do. What's the difference?

tgross35: I believe glibc actually reexports the kernel headers directly, while musl redefines a specific subset that is more cross-platform compatible. Does that seem to match up?

alyssais: On Alpine at least, musl doesn't install any headers under linux/ — you're expected to separately install the kernel headers.  So it's really just a packaging difference, as I understand it.  The API is the same.

alyssais: @tgross35 so what do we do here?  I'm noticing a lot of other constants that are also defined in kernel headers yet only available for Glibc in libc.  It would be good to have a path forward for these.

tgross35: I think this is okay since we are already including `linux/sched.h` in tests. In general we are trying to move away from providing `linux/` exports and instead suggesting https://docs.rs/linux-raw-sys/latest/linux_raw_sys/, but if this is expected to work well on musl, I don't think there is too much harm in smoothing the API across platforms a bit.

Could you rebase?

alyssais:
Oh, looks like this has already been done in 3ba96dfd6cd6a9fdaa3b848efaa17ae0be8294eb without me noticing :)

PR #3535, opened 2024-01-10
Support CRT Static for FreeBSD

rellerreller:
Modify the dependency links for FreeBSD to support statically linking Rust binaries on FreeBSD.


rellerreller:
I'm going to close this pull request. I have been doing more research, and I do not believe this patch is required to build static executables on FreeBSD. I will submit another pull request if I learn otherwise. I don't want resources to be wasted until I'm more confident.

rellerreller:
Closing due to comment above.

PR #3539, opened 2024-01-12
Add initial support for hppa-unknown-linux-gnu

hdeller:
This is the followup pull-request from:
https://github.com/rust-lang/libc/pull/3032


src/unix/linux_like/linux/gnu/b32/hppa/mod.rs
```diff
@@ -0,0 +1,805 @@
...
+pub const SIGUSR2: ::c_int = 17;
+pub const SIGCONT: ::c_int = 26;
+pub const SIGSTOP: ::c_int = 24;
+pub const SIGTSTP: ::c_int = 25;
+pub const SIGURG: ::c_int = 29;
+pub const SIGIO: ::c_int = 22;
+pub const SIGSYS: ::c_int = 31;
+pub const SIGSTKFLT: ::c_int = 7;
+#[deprecated(since = "0.2.55", note = "Use SIGSYS instead")]
+pub const SIGUNUSED: ::c_int = 31;
```
JohnTitor, line 308: Could you remove this? I think we don't need to add a deprecated one initially.

JohnTitor (commented):
Thanks! Overall LGTM with one comment

src/unix/linux_like/linux/gnu/b32/hppa/mod.rs
hdeller: What is the best way I can do this without loosing this pull request ?  Delete those lines in a new additional commit and push to the branch again?

JohnTitor (approved):

glaubitz:
I guess, we could start sending such PRs for alpha and sh4 as well. No need to do it for ia64 which has already been removed from the kernel and glibc as it's on its way out.

JohnTitor:
Oh, sorry I thought this target had been already added but seems not.
We have to revert this to make CI green, could you re-submit this once the target is added on the rustc?

PR #3549, opened 2024-01-18
Fix libc-tests for loongarch64 hwcaps

heiher:
I overlooked the crucial aspect of maintaining consistency between the names in hwcaps and the C header files for libc-test. As far as I'm aware, it has not been utilized yet, and it is high time to address this issue promptly.


JohnTitor:
It'd also be great if you could update the base branch to the latest, it contains some outdated CI config.

PR #3550, opened 2024-01-18
Fix libc-tests for loongarch64 syscall

heiher:
This PR moves `SYS_fstat` and `SYS_setrlimit` from `linux.txt` to `linux-$arch.txt`. The rationale behind this change is the absence of these two system call numbers for certain architectures, such as LoongArch, which employ generic Linux syscall numbers. In these cases, [__ARCH_WANT_NEW_STAT](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/uapi/asm-generic/unistd.h?h=v6.6.12#n209) and [__ARCH_WANT_SET_GET_RLIMIT](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/uapi/asm-generic/unistd.h?h=v6.6.12#n435) are not defined.

### DIFF
--- a/libc-test/semver/linux-aarch64.txt
+++ b/libc-test/semver/linux-aarch64.txt
@@ -66,6 +66,7 @@ SKF_NET_OFF
 SO_PRIORITY
 SO_PROTOCOL
 SYS_accept
+SYS_fstat
 SYS_msgctl
 SYS_msgget
 SYS_msgrcv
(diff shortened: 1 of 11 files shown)


JohnTitor:
Thanks! It'd be great if you could merge the changes here into #3549 as it's quite similar.

PR #3538, opened 2024-01-11
Fix broken vita std compilation by adding SOMAXCONN

pheki:
std is broken on PS Vita since https://github.com/rust-lang/rust/pull/119026

- vita's SOMAXCONN discovered empirically
- ~~esp-idf's SOMAXCONN from https://github.com/espressif/esp-idf/blob/b3f7e2c8a4d354df8ef8558ea7caddc07283a57b/components/newlib/platform_include/net/if.h#L20~~ (removed as it was probably incorrect, see [comment](https://github.com/rust-lang/libc/pull/3538#issuecomment-1889107265))

~~@ivmarkov can you confirm the esp change?~~

PR #3552, opened 2024-01-19
Add SOMAXCONN to vita on 0.2 (to fix std)

pheki:
Hey, I realized we actually need it on `libc-0.2` as it needs to be released and used by `std`, so I cherry-picked the commit from `main`

Thanks a lot!

Original PR: https://github.com/rust-lang/libc/pull/3538


JohnTitor (approved):
Thanks!

PR #3553, opened 2024-01-20
Add MFD_NOEXEC_SEAL and MFD_EXEC

rusty-snake:


JohnTitor:
Could you check CI failure and squash commits into one?

rusty-snake:
glibc has them in `sys/mman.h` (dependencies), musl don't.
linux-headers has it in `linux/memfd.h`.

How to proceed.

JohnTitor:
You can take a look at https://github.com/rust-lang/libc/blob/main/libc-test/build.rs for how to include headers and ignore items (note that we use musl 1.1.24 on CI and you have to skip tests if the items are fresh. Otherwise including headers should resolve the issue).

rusty-snake:
No idea why the Mac thing failed, it does not even have a log.

JohnTitor (approved):

JohnTitor:
Seems a temporary failure, re-running resolved it. Thanks for the PR!

PR #3554, opened 2024-01-22
Backport of #3546 and update crate version to 0.2.153

GuillaumeGomez:
Backport of https://github.com/rust-lang/libc/pull/3546 and update of the crate version.

PR #3556, opened 2024-01-23
Update crate version to 0.2.153

pheki:
Hey,

This version would contain the SOMAXCONN constant for the vita (added in https://github.com/rust-lang/libc/pull/3552), required to compile std


JohnTitor:
Thanks! But this is a duplicate of #3554 and I'm going to close in favor of that.

PR #3353, opened 2023-09-14
Added all definitions from linux/ptp-clock.h

davidv1992:
This still has some trouble passing CI due to two unnamed anonymous unions in ptp_perout_request. Any guidance on how to deal with those in a better way?

PR #3559, opened 2024-01-25
Add all definitions from linux/ptp-clock.h

folkertdev:
I've taken over this change from @davidv1992 (who is my colleague). https://github.com/rust-lang/libc/pull/3353 is the original PR which can now be closed.

I'm also assuming this can/should be merged into the libc-0.2 branch (and main will be rebased on top of that at some point?).


libc-test/build.rs
```diff
@@ -4322,7 +4328,11 @@ fn test_linux(target: &str) {
         // the `ifc_ifcu` field is an anonymous union
         (struct_ == "ifconf" && field == "ifc_ifcu") ||
         // glibc uses a single array `uregs` instead of individual fields.
-        (struct_ == "user_regs" && arm)
+        (struct_ == "user_regs" && arm) ||
+        // `anonymous_1` is an anonymous union
+        (struct_ == "ptp_perout_request" && field == "anonymous_1") ||
+        // `anonymous_2` is an anonymous union
+        (struct_ == "ptp_perout_request" && field == "anonymous_2")
```
folkertdev, line 4335: I'm not sure why but both this and block and the block above excluding these fields seem required for CI to be satisfied?!

JohnTitor:
Could you resolve CI failure?
@rustbot author

folkertdev:
well that was more work that expected. A bunch of the constants and one struct are defined only in newer kernel versions (this causes problems on the 0.2 branch on musl, which uses an older kernel version)

- https://github.com/torvalds/linux/blame/8a696a29c6905594e4abf78eaafcb62165ac61f1/include/uapi/linux/ptp_clock.h#L217
- https://github.com/torvalds/linux/commit/361800876f80da3915c46e388fc682532228b2c3#diff-721f436093b6e07475c0784886e2092c2d0bc269397e0efd279fdf943ea55f27R149
- https://github.com/torvalds/linux/commit/415606588c61230b7b4f0118fc2d64a0c1c4d102#diff-721f436093b6e07475c0784886e2092c2d0bc269397e0efd279fdf943ea55f27R162-R173

and then on powerpc a bunch of the constants have different values.

src/unix/linux_like/linux/mod.rs
```diff
@@ -3638,6 +3733,58 @@ pub const HWTSTAMP_FILTER_PTP_V2_SYNC: ::c_uint = 13;
...
+cfg_if! {
+    if #[cfg(any(target_arch = "powerpc", target_arch = "powerpc64"))] {
+        pub const PTP_CLOCK_GETCAPS: ::c_uint = 0x40503d01;
+        pub const PTP_EXTTS_REQUEST: ::c_uint = 0x80103d02;
+        pub const PTP_PEROUT_REQUEST: ::c_uint = 0x80383d03;
+        pub const PTP_ENABLE_PPS: ::c_uint = 0x80043d04;
+        pub const PTP_SYS_OFFSET: ::c_uint = 0x83403d05;
+        pub const PTP_PIN_GETFUNC: ::c_uint = 0xc0603d06;
+        pub const PTP_PIN_SETFUNC: ::c_uint = 0x80603d07;
+        pub const PTP_SYS_OFFSET_PRECISE: ::c_uint = 0xc0403d08;
```
folkertdev, line 3748: these are defined with some macros. There are rust versions here https://github.com/torvalds/linux/blob/8a696a29c6905594e4abf78eaafcb62165ac61f1/rust/kernel/ioctl.rs#L13

should those be used instead?

folkertdev:
@JohnTitor this is actually waiting for review, but I don't think I can update the label? (not sure how this workflow is supposed to work?)

folkertdev:
ah, rustbot got more helpful in newer PRs 

@rustbot review

libc-test/build.rs
```diff
@@ -3457,6 +3457,8 @@ fn test_linux(target: &str) {
         "linux/netlink.h",
         // FIXME: requires Linux >= 5.6:
         [!musl]: "linux/openat2.h",
+        "linux/ptp_clock.h",
+        // FIXME: requires Linux >= 5.6:
```
tgross35, line 3461: What is this referring to? uapi `ptp_clock.h` has been around since at least 4.0, not sure if this is intended for `ptrace.h` below but that doesn't make sense either.

libc-test/build.rs
```diff
@@ -3757,6 +3762,7 @@ fn test_linux(target: &str) {
             if name == "SECCOMP_GET_NOTIF_SIZES"
                || name == "SECCOMP_FILTER_FLAG_NEW_LISTENER"
                || name == "SECCOMP_FILTER_FLAG_TSYNC_ESRCH"
+               || name == "PTP_SYS_OFFSET_EXTENDED"
```
tgross35, line 3765: Nit, alphabetical

```suggestion
            if name == "PTP_SYS_OFFSET_EXTENDED"
               || name == "SECCOMP_GET_NOTIF_SIZES"
               || name == "SECCOMP_FILTER_FLAG_NEW_LISTENER"
               || name == "SECCOMP_FILTER_FLAG_TSYNC_ESRCH"
```

libc-test/build.rs
```diff
@@ -3769,6 +3775,15 @@ fn test_linux(target: &str) {
...
                 || name.starts_with("SCM_J1939")
+                || name == "PTP_CLOCK_GETCAPS2"
+                || name == "PTP_EXTTS_REQUEST2"
+                || name == "PTP_PEROUT_REQUEST2"
+                || name == "PTP_ENABLE_PPS2"
+                || name == "PTP_SYS_OFFSET2"
+                || name == "PTP_PIN_GETFUNC2"
+                || name == "PTP_PIN_SETFUNC2"
+                || name == "PTP_SYS_OFFSET_PRECISE2"
+                || name == "PTP_SYS_OFFSET_EXTENDED2"
```
tgross35, line 3786: Nit, alphabetical

```suggestion
                || name == "PTP_CLOCK_GETCAPS2"
                || name == "PTP_ENABLE_PPS2"
                || name == "PTP_EXTTS_REQUEST2"
                || name == "PTP_PEROUT_REQUEST2"
                || name == "PTP_PIN_GETFUNC2"
                || name == "PTP_PIN_SETFUNC2"
                || name == "PTP_SYS_OFFSET2"
                || name == "PTP_SYS_OFFSET_PRECISE2"
                || name == "PTP_SYS_OFFSET_EXTENDED2"
```

src/unix/linux_like/linux/mod.rs
```diff
@@ -650,6 +650,57 @@ s! {
...
+        pub n_pins: ::c_int,
+        pub cross_timestamping: ::c_int,
+        #[cfg(target_env = "gnu")]
+        pub adjust_phase: ::c_int,
+        #[cfg(target_env = "gnu")]
+        pub max_phase_adj: ::c_int,
+        #[cfg(target_env = "gnu")]
+        pub rsv: [::c_int; 11],
+        #[cfg(any(target_env = "musl", target_env = "ohos"))]
+        pub rsv: [::c_int; 13],
```
tgross35, line 675: Could you add a note about how this is a kernel version difference that happens to be exposed by glibc / musl supported versions?

Also if you have links to relevant sources or docs, please add them in the PR description. I haven't been able to find anything relevant.

src/unix/linux_like/linux/mod.rs
```diff
@@ -650,6 +650,57 @@ s! {
...
+    pub struct ptp_extts_request {
+        pub index: ::c_uint,
+        pub flags: ::c_uint,
+        pub rsv: [::c_uint; 2],
+    }
+
+    pub struct ptp_sys_offset_extended {
+        pub n_samples: ::c_uint,
+        pub rsv: [::c_uint; 3],
+        pub ts: [[ptp_clock_time; 3]; 25],
```
tgross35, line 687: Can this make use of `PTP_MAX_SAMPLES` like the source header?

src/unix/linux_like/linux/mod.rs
```diff
@@ -889,6 +940,21 @@ s_no_extra_traits! {
         pub sched_deadline: ::__u64,
         pub sched_period: ::__u64,
     }
+
+    // linux/ptp_clock.h
+    pub struct ptp_sys_offset {
+        pub n_samples: ::c_uint,
+        pub rsv: [::c_uint; 3],
+        pub ts: [ptp_clock_time; 51],
```
tgross35, line 948: Can this make use of `PTP_MAX_SAMPLES`?

src/unix/linux_like/linux/mod.rs
```diff
@@ -920,6 +986,34 @@ cfg_if! {
     }
 }
 
+cfg_if! {
+    if #[cfg(libc_union)] {
+        s_no_extra_traits! {
```
tgross35, line 991: I think this doesn't need `cfg_if` since there is no else branch?

```suggestion
#[cfg(libc_union)]
s_no_extra_traits! {
    // ...
```

src/unix/linux_like/linux/mod.rs
```diff
@@ -1394,6 +1488,62 @@ cfg_if! {
...
+                    .finish()
+            }
+        }
+        impl ::hash::Hash for ptp_sys_offset {
+            fn hash<H: ::hash::Hasher>(&self, state: &mut H) {
+                self.n_samples.hash(state);
+                self.rsv.hash(state);
+                self.ts[..].hash(state);
+            }
+        }
```
tgross35, line 1515: Why not just move this struct from `s_no_extra_traits!` to `s!` so it gets these derived? Same for `ptp_pin_desc`

src/unix/linux_like/linux/mod.rs
tgross35: I think duplicating the macros from `ioctl.h` would be better here - these numbers are kind of difficult to verify by hand as-is.

src/unix/linux_like/linux/mod.rs
```diff
@@ -3638,6 +3788,58 @@ pub const HWTSTAMP_FILTER_PTP_V2_SYNC: ::c_uint = 13;
...
+        pub const PTP_ENABLE_PPS2: ::c_uint = 0x40043d0d;
+        pub const PTP_SYS_OFFSET2: ::c_uint = 0x43403d0e;
+        pub const PTP_PIN_GETFUNC2: ::c_uint = 0xc0603d0f;
+        pub const PTP_PIN_SETFUNC2: ::c_uint = 0x40603d10;
+        pub const PTP_SYS_OFFSET_PRECISE2: ::c_uint = 0xc0403d11;
+        pub const PTP_SYS_OFFSET_EXTENDED2: ::c_uint = 0xc4c03d12;
+    }
+}
+
+pub const PTP_PF_NONE: ::c_uint = 0;
```
tgross35, line 3838: Just for reference

```suggestion
// enum ptp_pin_function
pub const PTP_PF_NONE: ::c_uint = 0;
```

tgross35:
Left a review above, mostly nits with some follow up questions.

@rustbot author

libc-test/build.rs
folkertdev: https://github.com/rust-lang/libc/pull/3559#issuecomment-1913615212 goes into why. it's not that the file doesn't exist before 5.6, but some of the constants are only added in this later version, so for the full functionality, 5.6 is required. I believe the same is true for `netlink.h` above, a file that is also very old but its api changed over time, so that is why I copied over that comment too

src/unix/linux_like/linux/mod.rs
folkertdev: true, although that appears to be true for other declarations too (e.g. the `extra_traits` one right below), and the cfg_if right above has this too. I'm happy to change it of course, but I went with consistency. 

So, do you want to see this changed just here (and be inconsistent with the other definitions surrounding it)?

tgross35: Ah yeah good point, I guess you may as well leave it. I'll do a cleanup pass at some point...

libc-test/build.rs
tgross35: I was asking mostly about why the `FIXME` is below `ptp_clock.h` rather than above, looks like it is referring to `ptrace.h` rather than `ptp_clock.h` - but I guess this didn't show up well on the review UI. That context is probably helpful though, maybe change the comment to `some fields require Linux >= 5.6`

src/unix/linux_like/linux/mod.rs
folkertdev: I'm confused by this now. I suspect that I came to the above after some fighting with CI. Here     // https://github.com/torvalds/linux/blob/b311c1b497e51a628aa89e7cb954481e5f9dced2/include/uapi/linux/ptp_clock.h#L94
 the `max_phase_adj` field clearly is defined. But then e.g. on godbolt, for both musl and gnu, that field does not exist. `adjust_phase` does though. I'll change it back and see what CI says

tgross35: > I'm confused by this now.

At least that makes two of us :)

tgross35: Oh, I just realized this is targeting 0.2 rather than `main`. I would prefer this target `main` first since it uses some newer versions, then for the backport we can work around any kind of legacy weirdness.

Don't close this since it is still useful, but if it isn't too much trouble then it would be good if you could reapply this on top of `main` and open a new PR with that.

(If it's a PITA don't worry about it and I'll figure out how to get things in sync after)

folkertdev: worked out fine https://github.com/rust-lang/libc/pull/3865

tgross35:
Considering there are some CI failures here and https://github.com/rust-lang/libc/pull/3865 exists now, let's figure this out on `main` first then backport as applicable.

PR #3561, opened 2024-01-26
Add port for RTEMS

thesummer:
I am currently porting the rust compiler and stdlib to the RTEMS realtime operating system (https://www.rtems.org/).

This PR is the prerequisite for my changes to the *rust* repo. 
`style.sh` succeeds, the libc-test fails to compile on my machine, but since I only add things which are behind guards for `target_os = "rtems"`. I hope CI will give a 👍 .

Once this passed. I will also add a PR for *main* branch.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -35,7 +35,10 @@ const ALLOWED_CFGS: &'static [&'static str] = &[
 
 // Extra values to allow for check-cfg.
 const CHECK_CFG_EXTRA: &'static [(&'static str, &'static [&'static str])] = &[
-    ("target_os", &["switch", "aix", "ohos", "hurd", "visionos"]),
+    (
+        "target_os",
+        &["switch", "aix", "ohos", "hurd", "visionos", "rtems"],
+    ),
     ("target_env", &["illumos", "wasi", "aix", "ohos"]),
     (
         "target_arch",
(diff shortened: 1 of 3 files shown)


JohnTitor:
Thanks! Could you add the target to rustc first? This crate follows rustc's platform support for adding/removing targets.

thesummer:
Ah, ok. Will do.

tgross35:
This looks pretty reasonable for me after https://github.com/rust-lang/rust/pull/127021 merges. Could you target `main`? I've been doing backports in batches from there.

Also, are headers / docs for the added API available? Trying to get those crosslinked on all PRs so there is always a reference.

thesummer:
> This looks pretty reasonable for me after [rust-lang/rust#127021](https://github.com/rust-lang/rust/pull/127021) merges. Could you target `main`? I've been doing backports in batches from there.
> 

@tgross35  I opened a new PR https://github.com/rust-lang/libc/pull/3866 for this as I couldn't find a way to change the target branch of this PR. I also added some more reference material in the descriptions over there.

tgross35:
> @tgross35 I opened a new PR #3866 for this as I couldn't find a way to change the target branch of this PR. I also added some more reference material in the descriptions over there.

You would need to rebase (locally), force push, then click the "edit" button to the right of the title and change the dropdown. But no worries, a new PR works too.

I'll close this one so there isn't a review in two places.

PR #3567, opened 2024-01-29
Remove removed items in OpenBSD

JohnTitor:
This removes the removed items in OpenBSD.


JohnTitor:
@semarie It'd be great if you could test this against OpenBSD. If there's anything I could remove/tweak, feel free to let me know!

semarie:
the testsuite is fine with it, on OpenBSD -current (upcoming 7.5 at the time).

some remarks:
- I am unsure about removing `syscall()` at this point, as 7.5 isn't released yet. So I would maybe only mark it deprecated. But I am fine with removing it too.
- the starting point for others elements to cleanup, would be to look at `test_openbsd()`:
  - `KERN_MAXID`, `NET_RT_MAXID`  could be removed (deprecated constants, as there are changing regulary from one release to another)
  - the test for not checking `EV_SYSFLAGS` could be removed (I don't remember why it is here)
  - the test for not checking `mimmutable()` function could be removed (the function shows-up in 7.3 which is a latest supported release)

semarie:
LGTM. thanks !

PR #3569, opened 2024-01-30
Fix: time_t is 64 bits on Emscripten

hoodmane:
Since July 8, 2022:
https://github.com/emscripten-core/emscripten/pull/17401 https://github.com/emscripten-core/emscripten/blob/main/system/lib/libc/musl/arch/emscripten/bits/alltypes.h?plain=1#L83

Fixes https://github.com/rust-lang/libc/issues/3306

### DIFF
--- a/src/unix/linux_like/emscripten/mod.rs
+++ b/src/unix/linux_like/emscripten/mod.rs
@@ -16,7 +16,24 @@ pub type loff_t = i64;
 pub type pthread_key_t = ::c_uint;
 
 pub type clock_t = c_long;
-pub type time_t = c_long;
+
+// time_t was type:
+//
+//   - long   from v1.5.4  (2013/08/08)  until v3.1.11
+//   - int    from v3.1.11 (2022/05/20)  until v3.1.16
+//   - _Int64 from v3.1.16  (2022/07/14)
+//
+//  And it's still int64 today. The change from long to int was marked as a
+//  nonfunctional change since wasm64 was considered to be not working yet in
+//  v3.1.11. So it suffices to say it's 32 bit until v3.1.16 and 64 bit after.
+cfg_if! {
+    if #[cfg(emscripten_64_bit_time_t)] {
+        pub type time_t = i64;
+    } else {
+        pub type time_t = i32;
+    }
+}
+
 pub type suseconds_t = c_long;
 pub type ino_t = u64;
 pub type off_t = i64;
--- a/build.rs
+++ b/build.rs
@@ -8,6 +8,7 @@ use std::string::String;
 // make sure to add it to this list as well.
 const ALLOWED_CFGS: &'static [&'static str] = &[
     "emscripten_new_stat_abi",
+    "emscripten_64_bit_time_t",
     "freebsd10",
     "freebsd11",
     "freebsd12",
@@ -56,7 +57,13 @@ fn main() {
     }
 
     match emcc_version_code() {
-        Some(v) if (v >= 30142) => set_cfg("emscripten_new_stat_abi"),
+        Some(v) if (v >= 30142) => {
+            set_cfg("emscripten_new_stat_abi");
+            set_cfg("emscripten_64_bit_time_t");
+        }
+        Some(v) if (v >= 30116) => {
+            set_cfg("emscripten_64_bit_time_t");
+        }
         // Non-Emscripten or version < 3.1.42.
         Some(_) | None => (),
     }


alex:
At risk of asking a dumb question, how does this work with respect to backwards compatibility? What happens if one uses this patched libc with an older emscripten?

alex:
To answer my own question: I think you probably need to put a branch in `build.rs` (search for `emcc_version_code` in there) and emit a `cfg` and make this conditional on `emcc` version.

But obviously I defer to the libc maintainers.

src/unix/linux_like/emscripten/mod.rs
```diff
@@ -16,7 +16,21 @@ pub type loff_t = i64;
...
+
+// time_t was type:
+// 
+//   - long   from v1.5.4  (2013/08/08)  until v3.1.11 
+//   - int    from v3.1.11 (2022/05/20)  until v3.1.16 
+//   - _Int64 from v3.1.16  (2022/07/14)
+// 
+//  And it's still int64 today. The change from long to int was marked as a
+//  nonfunctional change since wasm64 was considered to be not working yet in
+//  v3.1.11. So it suffices to say it's 32 bit until v3.1.16 and 64 bit after.
```
hoodmane, line 28: @sbc100 does this look correct to you?

sbc100: That looks right to me yes.

alex: (Hey Sam, long time! Hope you're doing well!)

sbc100: What!  Hi.   What a pleasure to see your name pop up in my inbox.   I hope you are doing well too.

src/unix/linux_like/emscripten/mod.rs
```diff
@@ -16,7 +16,21 @@ pub type loff_t = i64;
...
+// time_t was type:
+//
+//   - long   from v1.5.4  (2013/08/08)  until v3.1.11
+//   - int    from v3.1.11 (2022/05/20)  until v3.1.16
+//   - _Int64 from v3.1.16  (2022/07/14)
+//
+//  And it's still int64 today. The change from long to int was marked as a
+//  nonfunctional change since wasm64 was considered to be not working yet in
+//  v3.1.11. So it suffices to say it's 32 bit until v3.1.16 and 64 bit after.
+#[cfg(emscripten_64_bit_time_t)]
```
devnexen, line 29: almost there 

```rust
src/unix/linux_like/emscripten/mod.rs:29: use cfg_if! and submodules instead of #[cfg]
src/unix/linux_like/emscripten/mod.rs:31: use cfg_if! and submodules instead of #[cfg]
```

hoodmane: How's it look now?

devnexen: I guess we ll find out if all pass green :) if that s the case, squashing all commits into 1 would be great.

hoodmane: Okay CI passes and I squashed the commits.

JohnTitor:
For libc v0.3, I think it's fine to bump up the minimum supported emscripten version to reduce our complexity. The release date mentioned in OP seems enough to bump to me.

hoodmane:
Is there an explicit minimum supported version somewhere? I don't find it.

JohnTitor:
No, we could define it on v0.3 if you like.

kleisauke:
FWIW, Rust currently specifies Emscripten 2.0.5 as the minimum version it supports, see:
https://github.com/rust-lang/rust/pull/107221#issuecomment-1488357314

Bumping that to version 3.1.16 would probably be fine.

JohnTitor:
Is there any policy around the supported version of Emscripten on rustc? Maybe it's good time to ask to bump on rustc as well.

benjamin-sieffert:
> Is there any policy around the supported version of Emscripten on rustc? Maybe it's good time to ask to bump on rustc as well.

Honestly, with the amount of breaking ABI changes emscripten development is doing to their libc, they can’t really ask for anyone to support a wide range of versions.

Of course, the reason they are so liberally doing those changes is because, practically speaking, C programs are statically linked against emscripten libc: It all goes into a single wasm bundle. There is normally no risk that an emscripten program ends up running against an older (or newer) version of emscripten’s libc.

On the flipside, this means there are no real reasons for anyone to be on older versions of emscripten. There is even an emscripten-equivalent to `rustup` so that developers aren’t tied to the version that their distribution has packaged. https://github.com/emscripten-core/emsdk

The attemps that were made to support different emcc versions in this repo lead to convoluted structure (not covered by automatic testing, either) and don’t even work without `-Zbuild-std`, right? Even then I think they are bugged, as some `std::fs` operations are broken for me after #3282.

Imho under these considerations it would be very reasonable for Rust to only ever support one "quite recent" version of emscripten and no others.

tgross35:
I think this can be closed since IIUC https://github.com/rust-lang/libc/pull/3962 makes this change. Please reopen if this isn't correct.

PR #3574, opened 2024-02-04
fix build for apple devices arm64. __uint128/__int128 aliases are not…

devnexen:
… available.

### DIFF
--- a/src/fixed_width_ints.rs
+++ b/src/fixed_width_ints.rs
@@ -20,7 +20,7 @@ pub type uint32_t = u32;
 pub type uint64_t = u64;
 
 cfg_if! {
-    if #[cfg(all(target_arch = "aarch64", not(target_os = "windows")))] {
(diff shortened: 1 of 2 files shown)


src/fixed_width_ints.rs
```diff
@@ -20,7 +20,7 @@ pub type uint32_t = u32;
 pub type uint64_t = u64;
 
 cfg_if! {
-    if #[cfg(all(target_arch = "aarch64", not(target_os = "windows")))] {
+    if #[cfg(all(target_arch = "aarch64", not(any(target_os = "windows", target_os = "macos", target_os = "ios", target_os = "tvos", target_os = "watchos"))))] {
```
JohnTitor, line 23: I guess we can just use vendor = apple here.

devnexen: I tried it but unfortunatly it does not build

```rust
`cfg(target_vendor)` is experimental and subject to change (see issue #29718)
```

PR #3563, opened 2024-01-27
Add fcntl OFD commands for macOS

anacrolix:
I also followed all the PR instructions and sorted the semver for apple.

PR #3575, opened 2024-02-04
CI: trying out macos 14 image (which is arm64).

devnexen:


JohnTitor:
Given that Apple would remove the Intel support, I think it's okay to switch to M1 completely.

anacrolix:
@devnexen what happened?

devnexen:
I did a common PR instead of 3 since they were interdependent

anacrolix:
Great, thanks. The PR is https://github.com/rust-lang/libc/pull/3578. I'm watching with interest as I got blocked on needing macos-14 for https://github.com/rust-lang/libc/pull/3563 too.

PR #3576, opened 2024-02-05
macOs add macho arch id to cpu_type/subtype apis.

devnexen:


JohnTitor:
CI failed, could you check it?

devnexen:
in fact, it might be better if [this one is merged first](https://github.com/rust-lang/libc/pull/3574).

PR #3578, opened 2024-02-10
macOs various updates.

devnexen:
- CI only for macOs arm64.
- Fixing build issues for macOs arm64.
- Adding macos cpu to arch api.

### DIFF
--- a/.github/workflows/full_ci.yml
+++ b/.github/workflows/full_ci.yml
@@ -31,12 +31,12 @@ jobs:
       contents: read # to fetch code (actions/checkout)
 
     name: macOS
-    runs-on: macos-13
+    runs-on: macos-14
     strategy:
       fail-fast: true
       matrix:
         target: [
-          x86_64-apple-darwin,
+          aarch64-apple-darwin,
         ]
     steps:
       - uses: actions/checkout@v4
@@ -218,10 +218,10 @@ jobs:
       max-parallel: 4
       matrix:
         target:
-          - { toolchain: stable, os: macos-13 }
-          - { toolchain: beta, os: macos-13 }
-          - { toolchain: nightly, os: macos-13 }
-          - { toolchain: 1.71.0, os: macos-13 }
+          - { toolchain: stable, os: macos-14 }
+          - { toolchain: beta, os: macos-14 }
+          - { toolchain: nightly, os: macos-14 }
+          - { toolchain: 1.71.0, os: macos-14 }
     runs-on: ${{ matrix.target.os }}
     steps:
       - uses: actions/checkout@v4
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -294,6 +294,8 @@ fn test_apple(target: &str) {
             "tcp_connection_info" => true,
             // FIXME: The size is changed in recent macOSes.
             "malloc_introspection_t" => true,
+            // sonoma changes the padding `rmx_filler` field.
+            "rt_metrics" => true,
 
             _ => false,
         }
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2217,6 +2217,12 @@ fn test_freebsd(target: &str) {
             // should've been used anywhere anyway.
             "TDF_UNUSED23" => true,
 
+            // Removed in FreeBSD 15
+            "TDF_CANSWAP" | "TDF_SWAPINREQ" => true,
+
+            // Unaccessible in FreeBSD 15
+            "TDI_SWAPPED" | "P_SWAPPINGOUT" | "P_SWAPPINGIN" => true,
+
             // Removed in FreeBSD 14 (git a6b55ee6be1)
             "IFF_KNOWSEPOCH" => true,
 
--- a/src/fixed_width_ints.rs
+++ b/src/fixed_width_ints.rs
@@ -20,7 +20,7 @@ pub type uint32_t = u32;
 pub type uint64_t = u64;
 
 cfg_if! {
-    if #[cfg(all(target_arch = "aarch64", not(target_os = "windows")))] {
+    if #[cfg(all(target_arch = "aarch64", not(any(target_os = "windows", target_os = "macos", target_os = "ios", target_os = "tvos", target_os = "watchos"))))] {
         // This introduces partial support for FFI with __int128 and
         // equivalent types on platforms where Rust's definition is validated
         // to match the standard C ABI of that platform.
@@ -92,5 +92,10 @@ cfg_if! {
 
         // static_assert_eq!(core::mem::size_of::<__uint128_t>(), _SIZE_128);
         // static_assert_eq!(core::mem::align_of::<__uint128_t>(), _ALIGN_128);
+    } else if #[cfg(all(target_arch = "aarch64", any(target_os = "macos", target_os = "ios", target_os = "tvos", target_os = "watchos")))] {
+        /// /// C `__int128_t`
+        pub type __int128_t = i128;
+        /// /// C `__uint128_t`
+        pub type __uint128_t = u128;
     }
 }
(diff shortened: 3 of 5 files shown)


Cargo.toml
```diff
@@ -106,8 +106,6 @@ targets = [
     "wasm32-unknown-emscripten",
     "wasm32-unknown-unknown",
     "wasm32-wasi",
-    "x86_64-apple-darwin",
-    "x86_64-apple-ios",
```
JohnTitor, line 110: This is for docs.rs and you don't have to touch.

src/unix/bsd/apple/mod.rs
```diff
@@ -6113,7 +6112,8 @@ extern "C" {
         out_processor_infoCnt: *mut mach_msg_type_number_t,
     ) -> ::kern_return_t;
 
-    pub static mut mach_task_self_: ::mach_port_t;
+    // FIXME: 2024 edition forbids static mut, no easy fix for both nightly/stable neither
+    //pub static mut mach_task_self_: ::mach_port_t;
```
JohnTitor, line 6116: We could just remove mach related items.

src/unix/bsd/apple/mod.rs
```diff
@@ -6228,10 +6226,16 @@ extern "C" {
...
-    mach_task_self_
+    pub fn macho_cpu_type_for_arch_name(
+        archName: *const ::c_char,
+        tpe: *mut ::cpu_type_t,
+        subtpe: *mut ::cpu_subtype_t,
+    ) -> bool;
+    pub fn macho_arch_name_for_cpu_type(
+        tpe: ::cpu_type_t,
+        subtpe: ::cpu_subtype_t,
+    ) -> *const ::c_char;
```
JohnTitor, line 6238: These should exist on the mach crate.

libc-test/build.rs
```diff
@@ -203,6 +203,7 @@ fn test_apple(target: &str) {
         "limits.h",
         "locale.h",
         "malloc/malloc.h",
+        "mach-o/utils.h",
```
JohnTitor, line 206: I think this is now unneeded, right?

JohnTitor (approved):

anacrolix:
That's an unexpected error from the merge queue.

anacrolix:
Woops, I think this got stuck again. Could we try to merge it again @JohnTitor @devnexen ?

anacrolix:
Hello again, this is ready to go! Let me know if I can help in some way?

src/fixed_width_ints.rs
```diff
@@ -92,5 +92,10 @@ cfg_if! {
 
         // static_assert_eq!(core::mem::size_of::<__uint128_t>(), _SIZE_128);
         // static_assert_eq!(core::mem::align_of::<__uint128_t>(), _ALIGN_128);
+    } else if #[cfg(all(target_arch = "aarch64", any(target_os = "macos", target_os = "ios", target_os = "tvos", target_os = "watchos")))] {
```
Amanieu, line 95: Prefer using `target_vendor = "apple"` instead.

devnexen: I think I ve tried already in the past and now I try again I get this

```rust
  error: `cfg(target_vendor)` is experimental and subject to change (see issue #29718)
     --> ../src/fixed_width_ints.rs:95:50
      |
  95  |       } else if #[cfg(all(target_arch = "aarch64", target_vendor = "apple"))] {
      |                                                    ^^^^^^^^^^^^^^^^^^^^^^^
      | 

```

Amanieu:
CI is now failing on FreeBSD 15

.github/workflows/full_ci.yml
```diff
@@ -31,12 +31,12 @@ jobs:
       contents: read # to fetch code (actions/checkout)
 
     name: macOS
-    runs-on: macos-13
+    runs-on: macos-14
     strategy:
       fail-fast: true
       matrix:
         target: [
-          x86_64-apple-darwin,
```
tgross35, line 39: We probably shouldn't lose x86 MacOS yet because it is still Tier 1, are you able to add it back?

After that I think this can be merged together with https://github.com/rust-lang/libc/pull/3794 (in https://github.com/rust-lang/libc/pull/3795)

devnexen: Note it is only for CI also because further macos images are exclusively arm64.

tgross35: We would just need separate jobs that run on `macos-13` and `macos-14`, [which is what `rustc` currently does](https://github.com/rust-lang/rust/blob/6696447f784a888446d13bb400a8d507a68331c9/src/ci/github-actions/jobs.yml#L19-L25).

Was the x86_64-apple CI broken? If not, it may be better to drop this part of the change and move it to a new PR, so this PR is only the things needed to get CI working again.

src/fixed_width_ints.rs
tgross35: Which job does this fail on? It looks like this should have been stable since 1.33 https://github.com/rust-lang/rust/pull/57465

.github/workflows/full_ci.yml
devnexen: we would need to add exceptions for x86_64 with lack of certain types updates/changes, api missing and so on (even though in reality it s just because of macos releases not architecture related at all).

devnexen: cc @Amanieu

libc-test/build.rs
```diff
@@ -2215,6 +2217,12 @@ fn test_freebsd(target: &str) {
             // should've been used anywhere anyway.
             "TDF_UNUSED23" => true,
 
+            // Removed in FreeBSD 15
+            "TDF_CANSWAP" | "TDF_SWAPINREQ" => true,
+
+            // Unaccessible in FreeBSD 15
+            "TDI_SWAPPED" | "P_SWAPPINGOUT" | "P_SWAPPINGIN" => true,
+
```
tgross35, line 2225: Do you mind moving the freebsd-specific changes to a separate commit so cherry picking is a bit easier?

src/fixed_width_ints.rs
```diff
@@ -92,5 +92,10 @@ cfg_if! {
...
+        /// C __int128_t (alternate name for [__int128][])
+        pub type __int128_t = i128;
+        /// C __uint128_t (alternate name for [__uint128][])
+        pub type __uint128_t = u128;
```
tgross35, line 99: `__int128` and `__uint128` aren't defined in this branch of the cfg.

```suggestion
        /// C `__int128_t`
        pub type __int128_t = i128;
        /// C `__uint128_t`
        pub type __uint128_t = u128;
```

.github/workflows/full_ci.yml
tgross35: Sounds good, thanks for confirming.

tgross35 (changes requested):
A couple other small changes, plus https://github.com/rust-lang/libc/pull/3578/files#r1703994194 if it works. Then I think we can merge this with the other CI fixes.

libc-test/build.rs
devnexen: you mean another PR ? usually we merge squashed commits ultimately.

src/fixed_width_ints.rs
devnexen: sure

libc-test/build.rs
tgross35: No, just separate commits. Usually only messy history needs to get squashed, and this specifically won't since it will get merged with other CI fixes.

(I'm just asking because I don't know what exactly needs to get cherry picked to the 0.2 branch to get that CI working)

devnexen: should be ok in theory,  l ll get to it soon-ish

tgross35:
This was merged via https://github.com/rust-lang/libc/pull/3797.

anacrolix:
Thank you!

PR #3585, opened 2024-02-14
Set CLONE_CLEAR_SIGHAND and CLONE_INTO_CGROUP to a large enough type

### LINKED ISSUE #3584: CLONE_INTO_CGROUP and CLONE_CLEAR_SIGHAND are both erroneously 0
Apparently the two clone flags `CLONE_INTO_CGROUP` and `CLONE_CLEAR_SIGHAND` are both larger than an i32, which results in them quietly being truncated to 0 over [here](https://github.com/rust-lang/libc/blob/7d19eedf7871923bc5092b202e84db1c5d14bcc2/src/unix/linux_like/linux/gnu/mod.rs#L1068) (or I'm misunderstanding something).

`linux/sched.h` defines them as follows:

```
#define CLONE_CLEAR_SIGHAND 0x100000000ULL
#define CLONE_INTO_CGROUP 0x200000000ULL
```

As far as I understand `clone()` cannot support these, but `clone3` should be able to.
In any way I'd argue the two flags being 0 is a bug and it would be better not having the flags then having them be wrong.

UhhhWaitWhat:
This PR prevents the flags `CLONE_CLEAR_SIGHAND` and `CLONE_INTO_CGROUP` from quietly being set to 0 by assigning them to a variable of a too-small type.

This should fix #3584. Sadly this is a breaking change (well the only thing that might break was buggy before, but I assume that still counts as breaking).

P.S.:

`ci/style.sh` did not pass on my machine, but I'm chalking this up to a shellcheck version difference or something, as I did not touch any shell scripts at all.

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -1065,8 +1065,8 @@ pub const ELFOSABI_ARM_AEABI: u8 = 64;
 
 // linux/sched.h
 pub const CLONE_NEWTIME: ::c_int = 0x80;
-pub const CLONE_CLEAR_SIGHAND: ::c_int = 0x100000000;
-pub const CLONE_INTO_CGROUP: ::c_int = 0x200000000;
+pub const CLONE_CLEAR_SIGHAND: ::c_ulonglong = 0x100000000;
+pub const CLONE_INTO_CGROUP: ::c_ulonglong = 0x200000000;
 
 // linux/keyctl.h
 pub const KEYCTL_DH_COMPUTE: u32 = 23;


JohnTitor (approved):

tgross35:
Another reevaluation: it's been a couple years, 
 https://github.com/search?q=lang%3Arust+%2Flibc.*%28CLONE_CLEAR_SIGHAND%7CCLONE_INTO_CGROUP%29%2F+NOT+is%3Afork&type=code contains more workarounds for this than code affected by it. While a backport is breaking, I think that's less bad than shipping bogus values that everyone has to work around. So:

* Deprecated the constants to give users a heads up https://github.com/rust-lang/libc/pull/5298
* Opened https://github.com/maelstrom-software/maelstrom/issues/527

I'll give a few months for things to work through, then we can backport this.

@SnoozeThis wait 2 months -> remove label stable-declined, add label stable-nominated

SnoozeThis:
(https://snoozeth.is/BhbeG7TL3Uo) I will wait until Mon, 21 Sep 2026 09:43:27 UTC and then add label stable-nominated and remove label stable-declined.

@rustbot claim.

SnoozeThis:
Resolved.

@rustbot label +stable-nominated -stable-declined.

PR #3593, opened 2024-02-23
Add group functions for Android

GuillaumeGomez:


GuillaumeGomez:
Considering it seems to not be recommended to list groups this way on android, I'll just close this PR.

PR #3594, opened 2024-02-24
libc 0.3: Replace c_void with u8 in read, write, memcpy and many other functions

safinaskar:
libc 0.3: Replace c_void with u8 in read, write, memcpy and many other functions

Because when I `write` something, I usually have `*const u8`, not `*const c_void`. So this patch will make these functions easier to use.

This is breaking change


safinaskar:
This was bad idea. Closing

Issue #1272, opened 2019-02-22
The signature of  "execv" | "execve" | "execvp" is incorrect


gnzlbg:
The type of `argv` is `*const *const c_char` but should be `*const *mut c_char`. For some reason, this is skipped in the libc-test.

gnzlbg:
@asomers it would be better to start fixing these in freebsd 

@semarie it would be good to know if their signature is also incorrect in openbsd and netbsd (e.g. by trying to remove their "skip" in `libc-test/build.rs` since we can't check that in CI.

I'll try to roll in a fix for Android.

After that is done, we can fix these on Linux, and do an update upstream, crater run, etc.

asomers:
Why do you think it would be best to start with FreeBSD?  It seems that the same problem exists on Linux.

gnzlbg:
We could try rolling this change for all targets at once as well. These APIs are available on all tier-1 targets, so I just thought that it might be "safer" to try to roll the change for less used targets first, and if there aren't any problems, only then attempt that for the tier-1 targets. 

AFAICT libstd does not use these APIs, so it is unclear to me how many people are using them (is anybody using them at all?) and how (e.g. whether changing the mutability of that pointer would be a breaking change for them or not).

asomers:
Well, Nix uses `libc::execve`.  And the proposed change will break Nix.  In fact, I think the error here is not in libc.  The error is in execve's signature as mandated by POSIX.  POSIX says "The argv[] and envp[] arrays of pointers and the strings to which those arrays point shall not be modified by a call to one of the exec functions, except as a consequence of replacing the process image.", yet it still specifies a non-const array for argv.  That's a bug.  It's probably a very old bug, dating to the days before C had a `const` keyword.  But given that "shall not be modified" requirement, I think that it's safe (and preferable) for libc to use a `*const *const c_char` argument.

https://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html

gnzlbg:
Doesn't the "except as a consequence of replacing the process image." imply that those can actually be modified ?

asomers:
I interpret that as meaning that they may be modified in the child's image, but not in the parent's image.  The constness of the pointer only matters from the perspective of the parent.

gnzlbg:
Makes sense. The question still is, why should we do something different than C here ? If the C headers have a bug, then that would be a good enough reason, but I am not sure if that's the case here.

semarie:
[...]

PR #3597, opened 2024-02-25
libc 0.3: Change prototypes for exec* function to match headers

safinaskar:
Yes, this makes the prototypes harder to use. And less intuitive.
But this makes them match headers, and thus now we can properly test them. This fixes https://github.com/rust-lang/libc/issues/1272

This is breaking change


safinaskar:
I rebased multiple times, and now the PR is ready for review

JohnTitor (approved):

asomers:
@JohnTitor did you realize that this change is backwards-incompatible, and causes downstream crates' builds to fail?

semarie:
these commits are in `main` branch, and not in `libc-0.2` branch. so I assume it is fine from libc point of vue ?

see [v0.3 Roadmap](https://github.com/rust-lang/libc?tab=readme-ov-file#v03-roadmap).

asomers:
Ahh, that makes sense.  Thanks, @semarie .

PR #3602, opened 2024-02-28
Support posix_spawn on Android

pcc:
Android exposes the same functions and constants as the other linux_like platforms, but its posix_spawnattr_t and posix_spawn_file_actions_t are opaque. See:

https://github.com/aosp-mirror/platform_bionic/blob/2215ad406b253f12e270cdd0876e19e9df2aa6d4/libc/include/spawn.h

Since the fields in the linux implementation are private, let's make it opaque on all linux_like platforms.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -1613,6 +1613,7 @@ fn test_android(target: &str) {
                "sched.h",
                "semaphore.h",
                "signal.h",
+               "spawn.h",
                "stddef.h",
                "stdint.h",
                "stdio.h",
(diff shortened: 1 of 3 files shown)


pcc:
Hmm, I think https://github.com/emscripten-core/emscripten/blob/cbb7a11e2b0ffec34b59c0a7417a1bd2a61e02d9/tools/system_libs.py#L1076 means that Emscripten doesn't actually support posix_spawn (though it does ship the header). I'll make this Android-only then.

@sbc100 FYI -- you might want to stop shipping spawn.h in Emscripten.

JohnTitor (approved):

madsmtm:
This was revered in https://github.com/rust-lang/libc/pull/3678 because exposing the fields was important to get the correct size, see https://github.com/rust-lang/libc/issues/3608.

@pcc I'd suggest you file a new PR that doesn't have this problem.

PR #3609, opened 2024-03-04
fix the size of `posix_spawn_file_actions_t` and `posix_spawnattr_t` on Linux

### LINKED ISSUE #3608: [libc-0.2] `posix_spawn_file_actions_t` cannot be used on Linux after #3602
We found an issue when updating ferrocene's libc submodule (ferrocene/ferrocene#356) to revision 947a1855124bdd13029d322ae6ab2b4f0451db9b . the only change included in our libc update was PR #3602 .

when building stage 2 of libstd to x86_64-linux the build failed to execute the build script of the `quote` crate with the following error:

``` text
thread 'main' panicked at library/std/src/sys_common/process.rs:155:17:
called Result::unwrap() on an Err value: Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" } 
```

where the mentioned line corresponds to [this revision](https://github.com/ferrocene/ferrocene/blob/46faebf340ed558b70026eaad44ca32d36d1542b/library/std/src/sys_common/process.rs#L155).

Tracking down the error led us to this usage of [the `posix_spawn` API](https://github.com/ferrocene/ferrocene/blob/46faebf340ed558b70026eaad44ca32d36d1542b/library/std/src/sys/pal/unix/process/process_unix.rs#L585-L586):

``` rust
let mut file_actions = MaybeUninit::uninit();
cvt_nz(libc::posix_spawn_file_actions_init(file_actions.as_mut_ptr()))?;
```

This code allocates a `posix_spawn_file_actions_init` type *on the stack*. before #3602, the `file_actions` stack variable had the correct size and alignment but after #3602, it now has a size and alignment of `*mut c_void` (8 bytes on x86_64) and that results in UB when `posix_spawn_file_actions_init` is called.

GLIBC implements `posix_spawn_file_actions_init` as [a memset operation](https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/spawn_faction_init.c;h=09ee703d33bd6ce9af090ef38bc3921622380783;hb=b6e3898194bbae78910bbe9cd086937014961e45#l51) that zeroes the struct. In Rust syntax, that would be more or less this code:

``` rust
#[repr(C)]
struct posix_spawn_file_actions_t {
    // fields and padding here
}

unsafe fn posix_spawn_file_actions_init(actions: *mut posix_spawn_file_actions_t) -> c_int {
    ptr::write_bytes(actions, 0, 1);
    0
}
```

Bionic implements `posix_spawn_file_actions_init` differently. It uses a heap allocation as indirection. The Rust version of the bionic implementation looks roughly like this:

``` rust
#[repr(C)]
struct __actual_posix_spawn_file_actions_t {
    // fields and padding here
}

type posix_spawn_file_actions_t = *mut c_void;

unsafe fn posix_spawn_file_actions_init(actions: *mut posix_spawn_file_actions_t) -> c_int {
    let mut alloc = Box::new(MaybeUninit::<__actual_posix_spawn_file_actions_t>::uninit());
    ptr::write_bytes(alloc.as_mut_ptr(), 0, 1);
    *actions = Box::into_raw(alloc).cast();
    0
}
```

This usage of `posix_spawn_file_actions_t` in libstd:

``` rust
let mut file_actions = MaybeUninit::uninit();
cvt_nz(libc::posix_spawn_file_actions_init(file_actions.as_mut_ptr()))?;
```

is compatible with both GLIBC and Bionic but the `libc` crate needs to provide the correct size and alignment on Linux.

So, the conditional code should produce this on Linux

``` rust
pub struct posix_spawn_file_actions_t {
    __allocated: ::c_int,
    __used: ::c_int,
    __actions: *mut ::c_int,
    __pad: [::c_int; 16],
}
```

but this on Android

``` rust
```

---

All this probably also applies to the `posix_spawnattr_t` type but I have yet to look into the details of the GLIBC code around that type.

I'll send up a fix PR after I have done some more testing.

japaric:
fixes #3608

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -3299,6 +3299,7 @@ fn test_vxworks(target: &str) {
 
 fn test_linux(target: &str) {
     assert!(target.contains("linux"));
+    assert!(!target.contains("android"));
 
     // target_env
     let gnu = target.contains("gnu");
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -48,6 +48,10 @@ pub type Elf64_Xword = u64;
 
 pub type eventfd_t = u64;
 
+// these structs sit behind a heap allocation on Android
+pub type posix_spawn_file_actions_t = *mut ::c_void;
+pub type posix_spawnattr_t = *mut ::c_void;
+
 s! {
     pub struct stack_t {
         pub ss_sp: *mut ::c_void,
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -742,6 +742,26 @@ s! {
         pub salt: [::c_uchar; TLS_CIPHER_CHACHA20_POLY1305_SALT_SIZE],
         pub rec_seq: [::c_uchar; TLS_CIPHER_CHACHA20_POLY1305_REC_SEQ_SIZE],
     }
+
+    pub struct posix_spawn_file_actions_t {
+        __allocated: ::c_int,
+        __used: ::c_int,
+        __actions: *mut ::c_int,
+        __pad: [::c_int; 16],
+    }
+
+    pub struct posix_spawnattr_t {
+        __flags: ::c_short,
+        __pgrp: ::pid_t,
+        __sd: ::sigset_t,
+        __ss: ::sigset_t,
+        #[cfg(any(target_env = "musl", target_env = "ohos"))]
+        __prio: ::c_int,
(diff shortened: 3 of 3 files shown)


pvdrz:
CI is failing due to https://github.com/rust-lang/libc/issues/3613

JohnTitor:
Could you rebase to pass CI?

coolbluewater:
@JohnTitor, is this libc release blocked waiting for a response here? How to unblock it?

japaric:
closing in favor of #3683

PR #3612, opened 2024-03-05
Support confstr on Linux

magicant:


jgarzik:
Pull request looks good to me.  More comprehensive than #3771.

Maybe a `main`-branch-first approach is ideal?

tgross35:
@magicant could you rebase this and target `main`?

This should remove `confstr` from `src/unix/linux_like/linux/gnu/mod.rs`, as was added in https://github.com/rust-lang/libc/pull/3771.

src/unix/linux_like/linux/mod.rs
```diff
@@ -1634,6 +1634,34 @@ pub const _SC_XOPEN_STREAMS: ::c_int = 246;
 pub const _SC_THREAD_ROBUST_PRIO_INHERIT: ::c_int = 247;
 pub const _SC_THREAD_ROBUST_PRIO_PROTECT: ::c_int = 248;
 
+pub const _CS_PATH: ::c_int = 0;
```
tgross35, line 1637: At least some of these should be added to other platforms, e.g. OpenBSD https://man.openbsd.org/confstr.3

src/unix/mod.rs
```diff
@@ -1423,6 +1423,18 @@ cfg_if! {
...
+    if #[cfg(not(target_os = "android"))] {
+        extern "C" {
+            #[cfg_attr(
+                all(target_os = "macos", target_arch = "x86"),
+                link_name = "confstr$UNIX2003"
+            )]
+            pub fn confstr(name: ::c_int, buf: *mut ::c_char, len: ::size_t) -> ::size_t;
+        }
+    }
+}
```
tgross35, line 1436:
```suggestion
#[cfg(not(target_os = "android"))]
extern "C" {
    #[cfg_attr(
        all(target_os = "macos", target_arch = "x86"),
        link_name = "confstr$UNIX2003"
    )]
    pub fn confstr(name: ::c_int, buf: *mut ::c_char, len: ::size_t) -> ::size_t;
}
```

No need to use `cfg_if` if there is no `else` (I know we do this elsewhere, that will be a cleanup one day...)

src/unix/linux_like/linux/mod.rs
jgarzik: Agree re additional for other platforms.

A challenge:   The constant is stable and POSIX-defined, but the _constant's value_ is not guaranteed stable across POSIX platforms.

src/unix/mod.rs
tgross35: Nevermind, ignore this one.

magicant:
Rebased.

I also re-checked the header files and changed some constant definitions that I missed before.

@rustbot ready

src/unix/linux_like/linux/mod.rs
```diff
@@ -2008,6 +2008,45 @@ pub const _SC_XOPEN_STREAMS: ::c_int = 246;
...
+pub const _CS_POSIX_V6_WIDTH_RESTRICTED_ENVS: ::c_int = 1;
+pub const _CS_GNU_LIBC_VERSION: ::c_int = 2;
+pub const _CS_GNU_LIBPTHREAD_VERSION: ::c_int = 3;
```
tgross35, line 2014: I think these _GNU_ values should go in gnu or musl only, they don't seem to exist in ulibc or newlib

tgross35:
Sorry for the conflict, one last thing to double check above and then LGTM

src/unix/linux_like/linux/mod.rs
tgross35: Musl seems to have them https://github.com/kraj/musl/blob/ffb23aef7b5339b8c3234f4c6a93c488dc873919/include/unistd.h#L439-L440 which is probably why we aren't getting test failures. But I don't think we have any need to expose it on anything but gnu.

magicant:
> 51f9da421...649154d3b

Rebased and resolved the conflicts

src/unix/linux_like/linux/mod.rs
magicant: > 649154d3b...51512d1d1

Moved back the `_CS_GNU_LIB...` definitions.

Thanks for reviewing!

tgross35 (approved):

PR #3616, opened 2024-03-09
Move testing of primitive types from std

ChrisDenton:
std shouldn't need to depend on libc for platforms that don't require it, but currently [one test](https://github.com/rust-lang/rust/blob/01d73d4041969cde4a79bf9793521ef323248a24/library/std/src/os/raw/tests.rs) means it is required.

It was suggested that this test more properly belongs in the libc crate itself, so I've made this PR for your consideration.

### DIFF
--- a/libc-test/Cargo.toml
+++ b/libc-test/Cargo.toml
@@ -85,3 +85,8 @@ harness = true
 name = "semver"
 path = "test/semver.rs"
 harness = false
+
+[[test]]
+name = "primitive_types"
+path = "test/primitive_types.rs"
+harness = true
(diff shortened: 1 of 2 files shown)


ChrisDenton:
Oh wait I just noticed an issue. This needs to be added to the libc-test Cargo.toml otherwise it's not actually being run.

tgross35:
Wow that is deceptive. We should just use `tests/` for everything that doesn't need to disable the harness, then put the no-harness tests somewhere else.

ChrisDenton:
Yeah, that sounds like a good idea. I only noticed because I looked at it in my editor and it helpfully greyed out the code. I couldn't immediately figure out why so I investigated further.

PR #3622, opened 2024-03-17
Correct the value of FAN_MARK_IGNORE

severen:
This PR aligns the value of `FAN_MARK_IGNORE` in libc with that in [`uapi/linux/fanotify.h`](https://github.com/torvalds/linux/blob/741e9d66/include/uapi/linux/fanotify.h#L87).

Previously, it was incorrectly set to the value of `FAN_MARK_MOUNT`, which can cause subtle but catastrophic bugs when one tries to mark a file as ignored.


severen:
I've also gone through and compared all of the other fanotify constants with those in `uapi/linux/fanotify.h` to see if I could spot any other mismatches, but it looks like everything else is correct.

JohnTitor (approved):
Thanks for catching!

PR #3623, opened 2024-03-17
Add missing Linux kernel header reference comment

severen:
This PR adds a comment referencing the `linux/futex.h` kernel header above the futex constants in `unix/linux_like/linux/mod.rs`.


severen:
It's a trivial commit, but while writing #3622, I noticed that the section with futex constants were missing a comment referencing the corresponding Linux kernel header, unlike everything else in that module.

Issue #3131, opened 2023-03-02
Remove "placeholder constants"


workingjubilee:
It's common for C APIs to define constant values to denote the "last" value in a set of constants, like `ELAST` or `IPPROTO_MAX`. This seems prone to weird bugs, and these constants provide _very_ marginal benefit to Rust programmers. Since we're considering some breakage anyways, we should just remove these values where we are made aware of them.

See:
- https://github.com/rust-lang/libc/issues/1896
- https://github.com/rust-lang/libc/pull/3040

JohnTitor:
👍, before 0.3, I'd like to deprecate these values.

JohnTitor:
Noticed that we have a lot of `*MAX` consts and deprecating manually is quite hard. We could write a script but it can cause a false positive, hmm. For now, I've edited the PR template to mention this concern.

Amanieu:
An alternative to deleting these placeholder constants would be to still provide them but not make any guarantees about their value moving forward. This means that we would be free to change the value of these constants without it being considered a breaking change.

CI tests should specifically exclude these, although it would be nice to make CI fail only if our value for such constants is lower than the one on the system.

dybucc:
a bunch of constants got annotated with doc comments in [^1] [^2] [^3] [^4] [^5] [^6], and the new usage guidelines mention that these symbols are likely to change without libc releasing a breaking release [^7].

i'm wondering if this issue needs rewording or otherwise repurposing?

[^1]: #5123
[^2]: #5122
[^3]: #5121
[^4]: #5120
[^5]: #5119
[^6]: #5118
[^7]: https://github.com/rust-lang/libc/blob/70c67fe2ec83eda97646b49cab2bcbd1603a65ea/src/lib.rs#L134

tgross35:
These constants have valid uses that make them not worth removing, so we're going to keep them around. We have a bit in the [usage guidelines](https://docs.rs/libc/latest/libc/#usage-guidelines) to help users avoid issues here.

PR #3628, opened 2024-03-22
Add FreeBSD's Capsicum constants

asomers:
Thanks for considering submitting a PR!

We have the [contribution guide](https://github.com/rust-lang/libc/blob/main/CONTRIBUTING.md). Please read it if you're new here!

Here's a checklist for things that will be checked during review or continuous integration.

- \[x] Edit corresponding file(s) under `libc-test/semver` when you add/remove item(s), e.g. edit `linux.txt` if you add an item to `src/unix/linux_like/linux/mod.rs`
- \[ ] Your PR doesn't contain any private or *unstable* values like `*LAST` or `*MAX` (see [#3131](https://github.com/rust-lang/libc/issues/3131))
- \[ ] If your PR has a breaking change, please clarify it
- \[ ] If your PR increments version number, it must NOT contain any other changes (otherwise a release could be delayed)
- \[x] Make sure `ci/style.sh` passes
- \[x] `cd libc-test && cargo test`
  - (this might fail on your env due to environment difference between your env and CI. Ignore failures if you are not sure)

Delete this line and everything above before opening your PR.


asomers:
The CI failures are unrelated.

asomers:
Note that the `CAP_RIGHTS_VERSION` constant violates the rule against unstable constants.  It's a version number used to tell the kernel which version of the other `CAP_*` rights the program is using.  So I think it's appropriate to include it here.  We must increment it in the future if there's a ever a new version of the rights.

JohnTitor (approved):

PR #3630, opened 2024-03-23
Fix the definition of sigevent on FreeBSD and Linux

asomers:
It was originally defined back before rust could represent C unions.  So instead of defining the union field correctly, it simply defined that union's most useful field.  Define it correctly now.

Remove traits that can't be safely implemented on a union: PartialEq, Eq, and Hash.  Define Debug, but exclude the union field.  Leave Clone in place.

This PR is a rebase of #2813 that eschews all of the backwards-compatibility parts.

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -236,20 +236,10 @@ impl ::Clone for devstat_select_mode {
 }
 
 s! {
-    pub struct aiocb {
-        pub aio_fildes: ::c_int,
-        pub aio_offset: ::off_t,
-        pub aio_buf: *mut ::c_void,
-        pub aio_nbytes: ::size_t,
-        __unused1: [::c_int; 2],
-        __unused2: *mut ::c_void,
-        pub aio_lio_opcode: ::c_int,
-        pub aio_reqprio: ::c_int,
-        // unused 3 through 5 are the __aiocb_private structure
-        __unused3: ::c_long,
-        __unused4: ::c_long,
-        __unused5: *mut ::c_void,
-        pub aio_sigevent: sigevent,
+    pub struct __c_anonymous_sigev_thread {
+        pub _function: Option<extern "C" fn(::sigval) -> *mut ::c_void>,
+        //pub _function: *mut ::c_void, // Actually a function pointer
+        pub _attribute: *mut ::pthread_attr_t,
     }
 
     pub struct jail {
@@ -1351,6 +1341,36 @@ s! {
 }
 
 s_no_extra_traits! {
+    #[cfg_attr(feature = "extra_traits", derive(Debug))]
+    pub struct __aiocb_private {
+        status: ::c_long,
+        error: ::c_long,
+        spare: *mut ::c_void,
+    }
+
+    #[cfg_attr(feature = "extra_traits", derive(Debug))]
+    pub struct aiocb {
+        pub aio_fildes: ::c_int,
+        pub aio_offset: ::off_t,
+        pub aio_buf: *mut ::c_void,
+        pub aio_nbytes: ::size_t,
+        __spare__: [::c_int; 2],
+        __spare2__: *mut ::c_void,
+        pub aio_lio_opcode: ::c_int,
+        pub aio_reqprio: ::c_int,
+        _aiocb_private: __aiocb_private,
+        pub aio_sigevent: sigevent,
+    }
+
+    // Can't correctly impl Debug for unions
+    #[allow(missing_debug_implementations)]
+    pub union __c_anonymous_sigev_un {
+        pub _threadid: ::__lwpid_t,
+        pub _sigev_thread: __c_anonymous_sigev_thread,
+        pub _kevent_flags: ::c_ushort,
+        __spare__: [::c_long; 8],
+    }
+
     pub struct utmpx {
         pub ut_type: ::c_short,
         pub ut_tv: ::timeval,
@@ -1398,12 +1418,7 @@ s_no_extra_traits! {
         pub sigev_notify: ::c_int,
         pub sigev_signo: ::c_int,
         pub sigev_value: ::sigval,
-        //The rest of the structure is actually a union.  We expose only
-        //sigev_notify_thread_id because it's the most useful union member.
-        pub sigev_notify_thread_id: ::lwpid_t,
-        #[cfg(target_pointer_width = "64")]
-        __unused1: ::c_int,
-        __unused2: [::c_long; 7],
+        pub _sigev_un: __c_anonymous_sigev_un,
     }
 
     pub struct ptsstat {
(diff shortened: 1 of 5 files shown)


JohnTitor:
Failed on musl: https://github.com/rust-lang/libc/actions/runs/8846424497/job/24292411184?pr=3630

asomers:
The Sparc failure is not due to this PR.

SteveLauC:
Considering the main branch is for version 1.0, which allows breaking changes, I think we can move this forward.

tgross35:
@asomers mind rebasing this when you get the chance? We can get this into `main`.

asomers:
Rebased, @tgross35

tgross35:
That job seems to fail a lot, I wrote up https://github.com/rust-lang/libc/issues/4112.

Thanks for the changes, I'll take a look today.

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1351,6 +1340,32 @@ s! {
...
+        // unused 3 through 5 are the __aiocb_private structure
+        __unused3: ::c_long,
+        __unused4: ::c_long,
+        __unused5: *mut ::c_void,
+        pub aio_sigevent: sigevent,
+    }
+
+    // Can't correctly impl Debug for unions
+    #[allow(missing_debug_implementations)]
+    pub union __c_anonymous_sigev_un {
```
tgross35, line 1362:
```suggestion
    pub union __c_anonymous_sigev_union {
```

Might as well be consistent with the other instance which spells out `union` in the type name but leaves calls the field `_un`

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1351,6 +1340,32 @@ s! {
...
+        pub aio_buf: *mut ::c_void,
+        pub aio_nbytes: ::size_t,
+        __unused1: [::c_int; 2],
+        __unused2: *mut ::c_void,
+        pub aio_lio_opcode: ::c_int,
+        pub aio_reqprio: ::c_int,
+        // unused 3 through 5 are the __aiocb_private structure
+        __unused3: ::c_long,
+        __unused4: ::c_long,
+        __unused5: *mut ::c_void,
```
tgross35, line 1356: Could this maybe just be put in an __aiocb_private structure for clarity?

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -236,20 +236,9 @@ impl ::Clone for devstat_select_mode {
...
-        __unused2: *mut ::c_void,
-        pub aio_lio_opcode: ::c_int,
-        pub aio_reqprio: ::c_int,
-        // unused 3 through 5 are the __aiocb_private structure
-        __unused3: ::c_long,
-        __unused4: ::c_long,
-        __unused5: *mut ::c_void,
-        pub aio_sigevent: sigevent,
+    pub struct __c_anonymous_sigev_thread {
+        pub _function: *mut ::c_void, // Actually a function pointer
```
tgross35, line 240: Could this be a function pointer?

```rust
        pub _function: Option<extern "C" fn(sigval) -> *mut c_void>
```

src/unix/linux_like/linux/gnu/mod.rs
```diff
@@ -542,6 +524,27 @@ impl siginfo_t {
...
+        pub aio_reqprio: ::c_int,
+        pub aio_buf: *mut ::c_void,
+        pub aio_nbytes: ::size_t,
+        pub aio_sigevent: ::sigevent,
+        __next_prio: *mut aiocb,
+        __abs_prio: ::c_int,
+        __policy: ::c_int,
+        __error_code: ::c_int,
+        __return_value: ::ssize_t,
+        pub aio_offset: off_t,
```
tgross35, line 541: Since this changes depending on file64, could you add this fixme so we catch it when doing that update?

```rust
        // FIXME(offset64): definition depends on oo64_t
        pub aio_offset: off_t,
```

src/unix/linux_like/linux/gnu/mod.rs
```diff
@@ -542,6 +524,27 @@ impl siginfo_t {
...
+        #[cfg(all(not(target_arch = "x86_64"), target_pointer_width = "32"))]
+        __unused1: [::c_char; 4],
```
tgross35, line 543: Doesn't make much of a difference but it looks like glibc calls this `__pad` https://github.com/bminor/glibc/blob/d899b48a30b2dd27ab25e1cd90ce28b75f7c0755/rt/aio.h#L52

src/unix/bsd/freebsdlike/freebsd/mod.rs
tgross35: Nevermind, makes more sense to keep the field name like we do for anonymous structs. I guess the other one should change.

src/unix/linux_like/mod.rs
```diff
@@ -222,6 +227,17 @@ s_no_extra_traits! {
         pub u64: u64,
     }
 
+    // Can't correctly impl Debug for unions
+    #[allow(missing_debug_implementations)]
+    pub union __c_anonymous_sigev_un {
+        #[cfg(target_pointer_width = "64")]
+        _pad: [::c_int; (64 - 2 * 4 - 8) / 4],
```
tgross35, line 234: Could you move this to a  `const SIGEV_PAD_SIZE: usize = ...` to match the source?

src/unix/linux_like/mod.rs
```diff
@@ -12,6 +12,11 @@ missing! {
 }
 
 s! {
+    pub struct __c_anonymous_sigev_thread {
+        pub _function: *mut ::c_void, // Actually a function pointer
```
tgross35, line 16:
```suggestion
        pub _function: Option<extern "C" fn(::sigval_t) -> *mut ::c_void>`
```

tgross35:
I went through line by line, everything looks correct here but I left a handful of consistency comments.

src/unix/bsd/freebsdlike/freebsd/mod.rs
asomers: Why yes it can.

asomers:
I applied your suggestions, @tgross35 , and rebased.  But I did not squash the final commit.

tgross35 (approved):
Does https://github.com/rust-lang/libc/pull/3630#discussion_r1850966501 work? If not, a `FIXME(ctest)` like the other expressions would be good.

I'll merge tomorrow either way, please squash when you get the chance.

@rustbot label stable-declined (breakage)

asomers:
> Does [#3630 (comment)](https://github.com/rust-lang/libc/pull/3630#discussion_r1850966501) work? If not, a `FIXME(ctest)` like the other expressions would be good.
> 
> I'll merge tomorrow either way, please squash when you get the chance.
> 
> @rustbot label stable-declined (breakage)

Whoops.  I missed that comment.  I've fixed it now.

tgross35:
Thanks!

PR #3631, opened 2024-03-24
Fix c_char on AIX

taiki-e:
Refs: https://github.com/rust-lang/rust/issues/122985


ecnelises:
Thanks for catching this! It makes sense to me.

taiki-e:
(Marked as ready for review because I got positive feedbacks from both target maintainers (https://github.com/rust-lang/libc/pull/3631#issuecomment-2017103370, https://github.com/rust-lang/rust/issues/122985#issuecomment-2017094844))

JohnTitor (approved):

PR #3634, opened 2024-03-27
Convert mode_t constants to octal

dtolnay:
For example the "**u**ser **r**ead **w**rite e**x**ecute" value is most appropriately represented in the following octal form:

```rust
pub const S_IRWXU: mode_t = 0o0700;
```

as opposed to what is currently found in libc, which is representations like:

https://github.com/rust-lang/libc/blob/a0f5b4b21391252fe38b2df9310dc65e37b07d9f/src/fuchsia/mod.rs#L1478

https://github.com/rust-lang/libc/blob/a0f5b4b21391252fe38b2df9310dc65e37b07d9f/src/wasi.rs#L258

The [glibc manual](https://www.gnu.org/software/libc/manual/html_node/Permission-Bits.html) consistently uses octal when describing values of mode_t, as in `0400` or `02000`.

---

This PR is generated by this script:

````rust
#!/usr/bin/env -S cargo -Zscript
```cargo
[dependencies]
anyhow = "1"
glob = "0.3"
proc-macro2 = { version = "1", features = ["span-locations"] }
syn = { version = "2", default-features = false, features = ["full", "parsing"] }
```

use anyhow::Result;
use glob::glob;
use std::fs;
use syn::{Expr, Item, Lit};

fn main() -> Result<()> {
    for path in glob("src/**/*.rs")? {
        let path = path?;
        let mut src = fs::read_to_string(&path)?;
        let syntax_tree = syn::parse_file(&src)?;

        let mut offset = 0isize;
        let mut modified = false;
        for item in syntax_tree.items {
            // Find `pub const S_[IE] = $lit`.
            let Item::Const(item) = item else {
                continue;
            };
            let name = item.ident.to_string();
            if !["S_I", "S_E", "__S_I"].iter().any(|prefix| name.starts_with(prefix)) {
                continue;
            }
            let Expr::Lit(expr) = *item.expr else {
                continue;
            };
            let Lit::Int(lit) = expr.lit else {
                continue;
            };

            // Render integer value in octal.
            let value: u64 = lit.base10_parse()?;
            let mut octal = format!("0o{:04o}", value);
            if value >= 0o10000 {
                octal.insert(octal.len() - 4, '_');
            }

            // Replace original literal.
            let range = lit.span().byte_range();
            let start = (range.start as isize + offset) as usize;
            let end = (range.end as isize + offset) as usize;
            src.replace_range(start..end, &octal);
            offset += octal.len() as isize - range.len() as isize;
            modified = true;
        }

        if modified {
            fs::write(&path, src)?;
        }
    }
    Ok(())
}
````


dtolnay:
Rebased to fix conflict with https://github.com/rust-lang/libc/pull/3654 in src/fuchsia/mod.rs.

JohnTitor (approved):
👍

PR #3635, opened 2024-03-28
android: add FUTEX_LOCK_PI2

rtzoeller:
bionic libc defines this and it appears to behave correctly on recent kernel versions.


JohnTitor (approved):

rtzoeller:
@JohnTitor do I need to put up a separate PR targeting `main` for a change like this? Or will it be automatically included at some future time?

PR #3637, opened 2024-03-30
[0.2] Add constant AT_MINSIGSTKSZ

ur4t:
https://github.com/rust-lang/rust/commit/9da004ea1997ac0eedbb882fa52960417725bd48 uses `AT_MINSIGSTKSZ` which is not present in `libc` crate. `AT_MINSIGSTKSZ` was introduced since glibc 2.35 on all architectures.


ur4t:
Some TODOs:
- [ ] cherry-pick these commits on `main` branch.
- [ ] reuse constant `libc::AT_MINSIGSTKSZ` in `std`.

Issue #3613, opened 2024-03-06
[libc-0.2] CI broken--pthread_getname_np linking failure in musl


nathaniel-bennett:
In the libc-0.2 branch, making any change (such as modifying a README) results in CI failure to to a linking error in `aarch64-unknown-linux-musl`. It appears that one of the tests is attempting to include the `Thread::get_name` method (which internally uses pthread_getn
ame_np), but the current musl CI does not have the `pthread_getname_np` function in its libc as it is too old.

```
= note: /usr/lib/gcc-cross/aarch64-linux-gnu/13/../../../../aarch64-linux-gnu/bin/ld: /rust/lib/rustlib/aarch64-unknown-linux-musl/lib/libstd-5be27651eb2c54c7.rlib(std-5be27651eb2c54c7.std.10df3d0a1be3812f-cgu.0.rcgu.o): in function `std::sys::pal::unix::thread::Thread::get_name':
          /rustc/b6d2d841bcf4b77343b159f134c7d39c2dd4ceaa/library/std/src/sys/pal/unix/thread.rs:233:(.text._ZN3std3sys3pal4unix6thread6Thread8get_name17hfd35a9649b47184eE+0x28): undefined reference to `pthread_getname_np'
          collect2: error: ld returned 1 exit status
```

chenx97:
can confirm on the main branch as well. See #3610

ChrisDenton:
Note that recent rust versions only support musl 1.2.3+. See the platform support page https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2-with-host-tools

JohnTitor:
For libc 0.2, we have to drop the CI config for newer rustc then 😢

JohnTitor:
Now it has been fixed, closing.

PR #3639, opened 2024-03-30
Temporarily disable musl builds until CI upgraded

nathaniel-bennett:
Temporarily disables musl checks (which is no longer working due to rustc requiring v1.2+) until the musl CI images can be updated.


nathaniel-bennett:
Nightly freebsd seems to be broken on `libc-0.2`, but fixing that would be outside the scope of this PR.

nathaniel-bennett:
(temporarily fixes #3613)

JohnTitor:
Thanks! Now CI works fine, closing.

PR #3646, opened 2024-04-02
Fix configuration for AIX

bzEq:
Should use `target_os` rather than `target_env`, so that correct libraries are linked.


JohnTitor (approved):
Thanks!

Issue #2669, opened 2022-02-04
Wrong structs dirent, dirent64


safinaskar:
Your structs `dirent`, `dirent64` are wrong at x86_64-unknown-linux-gnu. Their field `d_name` contains 256 bytes. But names can be bigger than 256 bytes in NTFS (in Linux). I was able to create NTFS partition in my Linux and then create file with name "ыыыы..." (250 times "ы", this is Cyrillic letter). Such name contains more than 256 bytes (if you have difficulties with reproducing, I can provide more details).

So, fix structs. I am not sure how exactly fix should be implemented. Maybe `[c_char; 512]`, but in this case we need to double check what is true upper limit. Maybe we can simply put `[c_char; 1]` or `[c_char; 0]`. Of course, it would be perfect to make `dirent` unsized, but with machine word sized reference. Unfortunately, as well as I know, such unsized structs are not possible in stable Rust

devnexen:
I m sorry to say but in fact dirent* struct are in fact correct, I invite you to look at the man pages/headers for confirmation. Note that libc is not a "smart wrapper" neither but reflects what the system libc are.

safinaskar:
@devnexen consider this code:
```rust
// This code for x86_64-unknown-linux-gnu
use std::ffi::CStr;
fn main() {
    unsafe {
        let dir = libc::opendir(b"/mnt/3" as *const u8 as *const i8);
        assert_ne!(dir, std::ptr::null_mut());
        loop {
            let ent: *mut libc::dirent = libc::readdir(dir);
            if ent == std::ptr::null_mut() {
                break;
            }
            let name = CStr::from_ptr((*ent).d_name.as_ptr());
            println!("{}", name.to_bytes().len());
            println!("{:?}", name);
        }
    }
}
```
This code correctly handles file names bigger than 256 bytes. Does it invokes undefined behavior? I. e. is reading past formal struct size undefined behavior?

safinaskar:
> getdents

getdents is not wrapped by libc crate, nor by glibc. So I will have to manually wrap it, this is error-prone

devnexen:
you can call it via syscall (the syscalls ids are present in libc) and map the linux_dirent* I think.

safinaskar:
I still think that we should research whether accessing `dirent` past its formal size is undefined behavior (possibly by reading https://rust-lang.github.io/unsafe-code-guidelines ). If yes, then we should fix or remove `readdir` function from libc, because it is impossible to use it correctly

[...]

PR #3650, opened 2024-04-04
[libc 1.0] Put a big warning to Linux's dirent docs

safinaskar:
This fixes my bug https://github.com/rust-lang/libc/issues/2669 .

We cannot change type for `d_name` field, because this will break tests. So the only thing we can do is to document `dirent` behavior.

The question remains what are exact rules for dealing with `dirent`, but this is left to Rust specification


safinaskar:
CI is failing, but it seems this is not because of my patch

JohnTitor:
We should resolve the issue instead of emitting warnings on 1.0 as we can introduce breaking changes there.

PR #3653, opened 2024-04-06
Avoid usage of thread_local on DragonFlyBSD

bjorn3:
Both __errno_location() and __error() return the same value, so the __error() function can be implemented on the Rust side as __errno_location(). This eliminates the sole use of thread_local in this crate.


src/unix/bsd/freebsdlike/dragonfly/mod.rs
```diff
@@ -1589,6 +1589,11 @@ f! {
     pub fn minor(dev: ::dev_t) -> ::c_int {
         (dev & 0xffff00ff) as ::c_int
     }
+
+    #[deprecated(since = "0.2.77", note = "Use `__errno_location()` instead")]
+    pub fn __error() -> *mut ::c_int {
+        super::__errno_location()
+    }
```
JohnTitor, line 1596: We could simply remove this function on the 1.0 release (i.e. main branch).

bjorn3: Done

JohnTitor (approved):
Thanks!

PR #3654, opened 2024-04-09
Correct types of mode bit constants on Fuchsia

djkoloski:
Fuchsia's definitions of these constants were split from the unix impls before #503, and so S_ISUID, S_ISGID, and S_ISVTX are all incorrectly typed as c_int instead of mode_t. This applies the same fix to Fuchsia's constant definitions to bring them in line with the rest of libc.

This is a breaking change since `c_int` is `i32` on Fuchsia and `mode_t` is `u32`, but it's a good breaking change.

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -1334,9 +1334,9 @@ pub const GRPQUOTA: ::c_int = 1;
 
 pub const SIGIOT: ::c_int = 6;
 
-pub const S_ISUID: ::c_int = 0x800;
-pub const S_ISGID: ::c_int = 0x400;
-pub const S_ISVTX: ::c_int = 0x200;
+pub const S_ISUID: ::mode_t = 0x800;
+pub const S_ISGID: ::mode_t = 0x400;
+pub const S_ISVTX: ::mode_t = 0x200;
 
 pub const IF_NAMESIZE: ::size_t = 16;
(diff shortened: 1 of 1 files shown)


workingjubilee:
Thanks for taking a look at this!

I think we should accept this, despite its nominal breaking change status, as this in-principle is the same change as #503 which was also breaking and also accepted-already. Code that currently builds and works on Fuchsia has to type-cast around this somehow anyways.

src/fuchsia/mod.rs
```diff
@@ -1334,9 +1334,9 @@ pub const GRPQUOTA: ::c_int = 1;
 
 pub const SIGIOT: ::c_int = 6;
 
-pub const S_ISUID: ::c_int = 0x800;
-pub const S_ISGID: ::c_int = 0x400;
-pub const S_ISVTX: ::c_int = 0x200;
+pub const S_ISUID: ::mode_t = 0x800;
+pub const S_ISGID: ::mode_t = 0x400;
+pub const S_ISVTX: ::mode_t = 0x200;
```
workingjubilee, line 1339: Not mandatory but this would eliminate an upcoming conflict with https://github.com/rust-lang/libc/pull/3634
```suggestion
pub const S_ISUID: ::mode_t = 0o4000;
pub const S_ISGID: ::mode_t = 0o2000;
pub const S_ISVTX: ::mode_t = 0o1000;
```

JohnTitor (approved):

PR #3656, opened 2024-04-13
add all android sysconf constants

fkm3:
Source:
https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/bits/sysconf.h

Hex is used instead of decimal to match the source. No entries were
removed (can be verified with `--word-diff`).


devnexen (approved):

fkm3:
The freebsd failure looks unrelated, it says `error: no member named 'inp_ppcb' in 'sockstat'`, but that type hasn't been touched.

Is there a way I can trigger a new test run?

JohnTitor (approved):

PR #3657, opened 2024-04-14
Remove linking the MSVC CRT

ChrisDenton:
This is now done in rust-lang/rust repo https://github.com/rust-lang/rust/pull/122268


JohnTitor (approved):
Thanks!

PR #3659, opened 2024-04-15
add O_EXEC and O_SEARCH on apple

TheJonny:
according to this fcntl.h,they exist: https://github.com/apple-oss-distributions/xnu/blob/main/bsd/sys/fcntl.h#L184C9-L184C47


TheJonny:
O_SEARCH = (0x40000000 | 0x00100000) does the right thing on macos 13, but on macos 12 the read permission for the directory is still checked on opening, so i think it will be ignored there. (tried on github actions, as I don't own an apple device)

if we need a compile time version check to not provide the symbol for macos 12 (older versions are end of life), I currently don't know how to do it, help would be welcome.

JohnTitor (approved):

PR #3661, opened 2024-04-18
linux_like: unify SIGEV_THREAD_ID support

christoph-heiss:
This has been supported in musl since release v1.2.2 in January 2021. See also [the commit in musl itself](https://git.musl-libc.org/cgit/musl/commit/?h=v1.2.2&id=7c71792e87691451f2a6b76348e83ad1889f1dcb).

Not sure about emscripten though - I guess without gating it, it would be enabled there too? 
As this is my first contribution, please just holler if something is off.


JohnTitor (approved):

JohnTitor:
(Should be merged after we upgrade musl header on CI).

christoph-heiss:
> (Should be merged after we upgrade musl header on CI).

For reference, should be mergeable after this PR: #3068.

tgross35 (changes requested):
We got a way to do this now: could you add `#[cfg(any(not(target_env = "musl"), musl_v1_2_3))]`?

tgross35 (approved):
Thank you!

dybucc:
Not sure if the PR author is still on this or if I missed something that
remained to be done, but I've picked up where the author left off and I believe
the patch in [^1] is finished.

I tweaked the `libc-test` build script some to ensure the SemVer test for that
one symbol only runs in musl targets when the `musl_v1_2_3` `cfg` is set. It was
not feasible to add it back to all non-glob Linux SemVer tests because that
doesn't solve the issue of musl targets failing when on a version older than
1.2.

Pinging @tgross35 as I've seen you answer back on a different stale PR.

[^1]:
    https://github.com/dybucc/libc/commit/f8982e93ee57fb365183959c7a9ecdeaf514f1b4

dybucc:
Gentle ping, @tgross35.

tgross35:
We now have a better process for this: no need to gate the constant on `musl_v1_2_3` or do anything special with semver, just add a test skip `"..." if old_musl => true` like we have a few other places.

dybucc:
> We now have a better process for this: no need to gate the constant on
> `musl_v1_2_3` or do anything special with semver, just add a test skip
> `"..." if old_musl => true` like we have a few other places.

What do you mean by a better way? I'm probably missing something but how are the
SemVer tests affected by the ctest skips? To my knowledge, those are separate
matters.

I tried out commenting out the SemVer skip, and setting up a ctest skip in the
relevant part of `test_linux`, but the tests still fail [^1]. This sample run
also had some odd behavior around `powerpc64-unknown-linux-musl`.

That target, with newer musl, is meant to find the symbol (like other targets
using newer musl.) But it doesn't. I couldn't quite figure out why that was the
case, though I believe this to be a separate matter.

[^1]: <https://github.com/dybucc/libc/actions/runs/30885076722/job/91914370701>

tgross35:
"no need to gate" was the important part, if we drop the `cfg` then the semver test problem goes away

dybucc:
Thanks. That's what I was missing. Fixed now at [^1].

[^1]: <https://github.com/dybucc/libc/commit/af6bbb3fd812813a8249478599a4db1789b8029b>

tgross35:
I think you can send that as a PR that closes this one.

Note that if you pick up somebody's work, you should set yourself as the commit author (as the person responsible for the final commit message and contents) and add them as a co-authored-by, rather than the other way around.

dybucc:
>I think you can send that as a PR that closes this one.
>
>Note that if you pick up somebody's work, you should set yourself as the commit author (as the person responsible for the final commit message and contents) and add them as a co-authored-by, rather than the other way around.

I thought it was the other way around. Done. The PR is open at #5375.

PR #3667, opened 2024-04-27
Add dns api

### LINKED ISSUE #3666: DNS configuration API on macOS
macOS has a DNS resolving API that is exported in libSystem.dylib.

https://opensource.apple.com/source/configd/configd-596.15/dnsinfo/dnsinfo_copy.c.auto.html
I will send a PR to cover the structs and functions.

aviramha:
Adding structs and functions defined in `dnsinfo.h` which is private, so no libc tests (or any idea for a workaround?)
The functions are exported via libSystem.dylib so symbol resolution should be okay.
Closes #3666

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -436,6 +436,19 @@ fn test_apple(target: &str) {
         "uuid_t" | "vol_capabilities_set_t" => true,
         _ => false,
     });
+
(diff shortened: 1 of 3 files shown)


JohnTitor:
Thank you for the PR! We, however, don't accept private items generally and would like not to introduce those.

PR #3669, opened 2024-04-28
Add htonl, htons, ntohl, ntohs

safinaskar:
Not a breaking change

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -1367,6 +1367,23 @@ extern "C" {
 
 }
 
+safe_f! {
+    // It seems htonl, etc are macros on macOS. So we have to reimplement them. So let's
+    // reimplement them for all UNIX platforms
+    pub {const} fn htonl(hostlong: u32) -> u32 {
+        u32::to_be(hostlong)
+    }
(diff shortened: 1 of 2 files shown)


src/unix/mod.rs
```diff
@@ -1367,6 +1367,23 @@ extern "C" {
 
 }
 
+safe_f! {
+    // It seems htonl, etc are macros on macOS. So we have to reimplement them. So let's
+    // reimplement them for all UNIX platforms
```
JohnTitor, line 1372: We could split the declaration for that target.

safinaskar: > We could split the declaration for that target

@JohnTitor , yes, we can. But I think this is slightly more error-prone than my approach. I. e. if we split the declaration, then it is possible we introduce bug for some platform, but not for others

tgross35: I generally agree with JohnTitor here that most things should just use the host functions if available, concerns about bugs on one platform but not others isn't much of a concern when one of those platforms uses the system functions.

However, these are also straightforward enough that it doesn't seem worth the effort to split. So I think this is okay.

PR #3670, opened 2024-04-30
Add getpid in wasi

youknowone:
Add `getpid` from wasi-libc. `getpid` in wasi itself doesn't make sense, but `wasi-libc` defines an emulated function which returns a constant value.

I found `semver/wasi.txt` doesn't exist. Do I need to create it?


JohnTitor (approved):

youknowone:
@JohnTitor I am really sorry. This one need to be reverted.

https://github.com/rust-lang/rust/pull/124554#issuecomment-2093453914

The function is not placed in `libc.a`

JohnTitor:
Alright 👍, created a revert PR: https://github.com/rust-lang/libc/pull/3679

PR #3678, opened 2024-05-03
Revert "Support posix_spawn on Android"

JohnTitor:
Reverts rust-lang/libc#3602


GuillaumeGomez:
Thanks! We just had a bug due to this in the GCC backend. Was about to send a fix for it. :)

PR #3681, opened 2024-05-05
Add wasi select, FD_SET, FD_ZERO, FD_ISSET

youknowone:
Please check if it fits in the project policy.

This is port of wasi-libc type definitions and functions. wasi-libc has different `fd_set` design to widely used one.

### DIFF
--- a/libc-test/semver/wasi.txt
+++ b/libc-test/semver/wasi.txt
@@ -1,0 +1,5 @@
+fd_set
+FD_SET
+FD_ZERO
+FD_ISSET
+select
--- a/src/wasi.rs
+++ b/src/wasi.rs
@@ -178,6 +178,11 @@ s! {
         pub st_ctim: timespec,
         __reserved: [c_longlong; 3],
     }
+
+    pub struct fd_set {
+        __nfds: usize,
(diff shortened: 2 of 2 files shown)


tgross35:
Could you start a wasi file at libc-test/semver and add these? Also needs a rebase.

I believe this is the original source: https://github.com/WebAssembly/wasi-libc/blob/230d4be6c54bec93181050f9e25c87150506bdd0/libc-bottom-half/headers/public/__fd_set.h

tgross35 (approved):

tgross35:
@youknowone it looks like this isn't working when the `rustc-dep-of-std` feature is enabled (which is how `std` is allowed to have dependencies), see here https://github.com/rust-lang/rust/pull/129226#issuecomment-2295372456. Could you put up a fix for it? You should just need to update this line to include `super::{Send, Sync, iter::Iterator}`:

https://github.com/rust-lang/libc/blob/53752600b2c5038ab3355a369fae35426e08d77b/src/wasi.rs#L4

Try the build locally, you may also need `#[allow(unused_imports)]` if Cargo complains

PR #3692, opened 2024-05-10
Remove extra traits

### LINKED ISSUE #2700: Broken PartialEq impl for sigset_t on Linux
On Linux, sigset_t is defined as a 1024-bit struct whereas only its first 64 bits are used. The sigemptyset function leaves the other 960 bits uninitialized.

The derived implementation of PartialEq for sigset_t naively compares all the bits of the struct and wrongly returns false depending on the uninitialized part.

```
$ cargo version -v
cargo 1.59.0 (49d8809dc 2022-02-10)
release: 1.59.0
commit-hash: 49d8809dc2d3e6e0d5ec634fcf26d8e2aab67130
commit-date: 2022-02-10
host: x86_64-unknown-linux-gnu
libgit2: 1.3.0 (sys:0.13.23 vendored)
libcurl: 7.80.0-DEV (sys:0.4.51+curl-7.80.0 vendored ssl:OpenSSL/1.1.1l)
os: OracleLinux 35.0.0 [64-bit]
```

```
$ cat Cargo.toml
[package]
name = "sigset-test"
version = "0.1.0"
edition = "2021"

[dependencies]
libc = { version = "0.2.119", features = ["extra_traits"] }
```

```
$ cat src/main.rs
fn main() {
    use std::mem::MaybeUninit;
    unsafe {
        let mut s1 = MaybeUninit::<libc::sigset_t>::uninit();
        let mut s2 = MaybeUninit::<libc::sigset_t>::uninit();
        libc::sigemptyset(s1.as_mut_ptr());
        libc::sigemptyset(s2.as_mut_ptr());
        let s1 = s1.assume_init();
        let s2 = s2.assume_init();
        // The two empty sets should compare equal
        assert_eq!(s1, s2);
    }
}
```

```
$ cargo r
   Compiling sigset-test v0.1.0 (/home/magicant/Desktop/sigset-test)
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/sigset-test`
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `sigset_t { __val: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] }`,
 right: `sigset_t { __val: [0, 0, 0, 0, 0, 0, 0, 0, 0, 139700099176444, 139700099131200, 139700096850408, 1736415839, 1, 139700096840524, 139700099177503] }`', src/main.rs:11:9
```

rj00a:
This PR removes the `extra_traits` feature and all implementations of `Eq`, `Hash`, and `Debug` as mentioned in #3248.

The decision to remove `Debug` impls was left as an open question, but I've gone ahead and removed them here in the interest of maintainability and consistency.

Closes #2700, #2064, #2816


ci/style.rs
```diff
@@ -140,9 +139,6 @@ fn check_style(file: &str, path: &Path, err: &mut Errors) {
         } else if line.starts_with("type ") && !in_impl {
             State::Typedefs
         } else if line.starts_with("s! {") {
-            s_macros += 1;
```
rj00a, line 143: I had to allow multiple `s!` macros in one module because the `s_no_extra_traits!` macro no longer exists.

SteveLauC:
> I would like to ask that if I could rebase your commits and develop upon them, to complete the "trait implementations" change for libc v1?

Hi, would you like to show me what these trait impl would look like in libc 1.0, this is not covered in the tracking issue and that OSPP page. Nix, as a big consumer of the libc crate, uses them a lot: https://github.com/nix-rust/nix/issues/2468

erickt:
As an alternative to getting rid of `extra_traits`, another direction could be to use or build a library like [zerocopy](https://docs.rs/zerocopy) to assert at compile time that all the fields of a union can be referenced without UB? That could be done through a test-only cfg to reduce compile times, or if the MSRV didn't support this or something.

myl7:
> what these trait impl would look like in libc 1.0

According to the "Trait implementations" section, I think `Eq` and `Hash` impl would be completely removed.
`Debug` impl may be kept, but at least future code would not impl it.

tgross35:
Hey @myl7: we still need to figure out what exactly we want to do with traits here (at this point I'm thinking we should keep `Debug` but remove the rest, but that hasn't been discussed), but any chance you would want to pick up https://github.com/rust-lang/libc/issues/4118? That is something we could merge today, and would make the diff here a bit smaller once we have a final decision.

tgross35:
We are revisiting this at [#t-libs > Questions and polls about &#96;libc&#96; 1.0](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/Questions.20and.20polls.20about.20.60libc.60.201.2E0/with/513470878) and https://github.com/rust-lang/libc/issues/3880. This is also quite out of date, so I am going to close it.

Issue #1950, opened 2020-10-22
No IN6ADDR_ANY_INIT or in6_addr_any IPV6 initializers


simias:
The libc crate currently exports the `INADDR_ANY` const to initialize a wildcard IPV4 in_addr but it doesn't provide the IPV6 counterparts as far as I can tell.

In C since in6_addr is an array instead of a single variable there are two ways to initialize a wildcard IPV6 address, the `in6_addr_any` global variable and the `IN6ADDR_ANY_INIT` macro, usually expanding to `{ { { 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0 } } }`.

I suspect that the lack of these two symbols in the libc crate is intentional since neither of them makes a lot of sense in Rust. As far as I can tell the most reasonable way to expose this feature in Rust would be:

```rust
const IN6ADDR_ANY: in6_addr = in6_addr { s6_addr: [0; 16] };
```

But then it doesn't actually match any of the C symbols exactly.

On one hand not having this feature in the crate is not the end of the world, initializing an `in6_addr` with full zeroes is not hard. On the hand people coming from C/C++ will probably be surprised not to find this symbol in the crate, especially since the IPV4 counterpart does exist.

PR #3693, opened 2024-05-10
Unix: add `IN6ADDR_ANY_INIT`, `IN6ADDR_LOOPBACK_INIT`, `in6addr_any`, `in6addr_loopback`

safinaskar:
This is not breaking change.

Note that this PR adds global variables (statics) `in6addr_any` and `in6addr_loopback`. But this is okay, because:
- We already have global variables in crate libc: https://docs.rs/libc/0.2.154/aarch64-apple-darwin/libc/index.html#statics
- `ctest2` knows how to test global variables: https://docs.rs/ctest2/0.4.8/ctest2/struct.TestGenerator.html#method.skip_static

This PR closes https://github.com/rust-lang/libc/issues/1950

Also I suggest merging this PR for ctest2: https://github.com/JohnTitor/ctest2/pull/56 , it removes annoying warnings produced by ctest2 in presence of statics. And also fixes CI for ctest2. I recommend merging https://github.com/JohnTitor/ctest2/pull/56 , but this is not required

### DIFF
--- a/libc-test/semver/unix.txt
+++ b/libc-test/semver/unix.txt
@@ -152,6 +152,8 @@ IF_NAMESIZE
 IGNBRK
 IGNCR
 IGNPAR
+IN6ADDR_ANY_INIT
+IN6ADDR_LOOPBACK_INIT
 INADDR_ANY
 INADDR_BROADCAST
 INADDR_LOOPBACK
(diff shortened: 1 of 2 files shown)


tgross35 (approved):
These are defined here https://github.com/bminor/glibc/blob/b22923abb046311ac9097a36b97b9b97342bac44/inet/netinet/in.h#L237-L240. We do also have a few statics around, e.g. https://github.com/rust-lang/libc/blob/50b4720b37526787e853b2606bc96c7609b34e93/src/solid/mod.rs#L838-L842.

LGTM, thanks!

PR #3695, opened 2024-05-11
adding a handful of linux fanotify data types.

### LINKED ISSUE #3688: Add struct `fanotify_event_info_fid`
Host triplet: `x86_64-unknown-linux-gnu`

API struct is as follows:
```c
struct fanotify_event_info_fid {
  struct fanotify_event_info_header hdr;
  __kernel_fsid_t fsid;
  unsigned char file_handle[0];
};
```

Man page for more info: https://man7.org/linux/man-pages/man7/fanotify.7.html

Source link: https://github.com/torvalds/linux/blob/dccb07f2914cdab2ac3a5b6c98406f765acab803/include/uapi/linux/fanotify.h#L166

devnexen:
close #3688

here the involved struct

[this one](https://github.com/torvalds/linux/blob/0c3836482481200ead7b416ca80c68a29cfdaabd/include/uapi/linux/fanotify.h#L153), [this one](https://github.com/torvalds/linux/blob/0c3836482481200ead7b416ca80c68a29cfdaabd/include/uapi/linux/fanotify.h#L180) and [this one](https://github.com/torvalds/linux/blob/0c3836482481200ead7b416ca80c68a29cfdaabd/include/uapi/linux/fanotify.h#L185).

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -491,6 +491,17 @@ s! {
         pub tcpi_rcv_space: u32,
         pub tcpi_total_retrans: u32,
     }
+
+    pub struct fanotify_event_info_pidfd {
+        pub hdr: ::fanotify_event_info_header,
+        pub pidfd: ::__s32,
+    }
+
+    pub struct fanotify_event_info_error {
+        pub hdr: ::fanotify_event_info_header,
+        pub error: ::__s32,
+        pub error_count: ::__u32,
+    }
 }
 
 impl siginfo_t {
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -3700,6 +3700,9 @@ fn test_linux(target: &str) {
         if musl && ty == "fanout_args" {
             return true;
         }
+        if sparc64 && ty == "fanotify_event_info_error" {
+            return true;
+        }
 
         match ty {
             // These cannot be tested when "resolv.h" is included and are tested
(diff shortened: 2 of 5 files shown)


tgross35:
I think this looks good, but can you confirm that `fanotify_event_info_pidfd` and `fanotify_event_info_error` are GNU-only? Also needs a rebase.

devnexen:
> I think this looks good, but can you confirm that `fanotify_event_info_pidfd` and `fanotify_event_info_error` are GNU-only? Also needs a rebase.

yes I confirm no presence in other libcs

tgross35:
I think this should be

@rustbot ready

If you have a link to the headers / docs that would be great, trying to get them linked in every PR so we can find them again when needed.

src/unix/linux_like/linux/align.rs
```diff
@@ -63,6 +63,13 @@ macro_rules! expand_align {
                 pub fd: ::c_int,
                 pub pid: ::c_int,
             }
+
+            #[repr(align(2))]
+            pub struct fanotify_event_info_header {
```
tgross35, line 68: https://github.com/torvalds/linux/blob/0c3836482481200ead7b416ca80c68a29cfdaabd/include/uapi/linux/fanotify.h#L153 doesn't seem to have an alignment attribute, and I think the default alignment should naturally be 16 bits. Why is it needed here?

tgross35: Also I think this would cause a problem building on ancient versions with alignment since this doesn't have a fallback

src/unix/linux_like/linux/gnu/mod.rs
```diff
@@ -453,6 +453,17 @@ s! {
...
+    pub struct fanotify_event_info_pidfd {
+        pub hdr: ::fanotify_event_info_header,
+        pub pidfd: ::__s32,
+    }
+
+    pub struct fanotify_event_info_error {
+        pub hdr: ::fanotify_event_info_header,
+        pub error: ::__s32,
+        pub error_count: ::__u32,
+    }
```
tgross35, line 466: Why are these GNU-only? I don't see them reexported in glibc, only uapi

devnexen: it s a bit of a mess on musl. with glibc it s enough to include `sys/fanotify.h` which in turn include `linux/fanotify.h` so you have all you need for this. in musl `sys/fanotify.h` which does not include `linux/fanotify.h` and if you do include both there are couple of type redefinition conflicts.

tgross35 (approved):
Lgtm, could you squash please?

PR #3698, opened 2024-05-12
adding aligned_alloc support for the following:

### LINKED ISSUE #3689: Missing function: aligned_alloc
This function is part of the [C standard](https://en.cppreference.com/w/c/memory/aligned_alloc). I don't know the full set of targets that support it.
- Linux: [glibc has it](https://www.gnu.org/software/libc/manual/html_node/Aligned-Memory-Blocks.html), and I found it in [musl](https://git.musl-libc.org/cgit/musl/tree/src/malloc/mallocng/aligned_alloc.c?h=v1.2.2), too (but that uses `malloc` + offset -- no idea how they're making sure that you can `free` the resulting pointer)
- FreeBSD also [seems to have it](https://man.freebsd.org/cgi/man.cgi?query=aligned_alloc&apropos=0&sektion=3&manpath=FreeBSD+9-current&format=html)
- macOS: available since 10.15 according to [this](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92143) (but buggy until today... but that's not libc's concern)
- Solaris seems to [have it](https://docs.oracle.com/cd/E88353_01/html/E37843/aligned-alloc-3c.html), and same for [Illumos](https://illumos.org/man/3C/aligned_alloc) (again, [buggy](https://www.illumos.org/issues/16545)...)
- The function is mentioned [here](https://android.googlesource.com/platform/bionic/+/master/docs/status.md) so I assume (recent enough) Android also has it.

devnexen:
- linux glibc/musl.
- all supported BSD implements it.
- apple.
- solarish.

close #3689

### DIFF
--- a/libc-test/semver/apple.txt
+++ b/libc-test/semver/apple.txt
@@ -1836,6 +1836,7 @@ aio_return
 aio_suspend
 aio_write
 aiocb
+aligned_alloc
 arc4random
 arc4random_buf
 arc4random_uniform
--- a/libc-test/semver/dragonfly.txt
+++ b/libc-test/semver/dragonfly.txt
@@ -1246,6 +1246,7 @@ aio_suspend
 aio_waitcomplete
 aio_write
 aiocb
+aligned_alloc
 arc4random
 arc4random_buf
 arc4random_uniform
--- a/libc-test/semver/freebsd.txt
+++ b/libc-test/semver/freebsd.txt
@@ -1703,6 +1703,7 @@ aio_suspend
 aio_waitcomplete
 aio_write
 aiocb
+aligned_alloc
 arc4random
 arc4random_buf
 arc4random_uniform
--- a/libc-test/semver/linux-gnu.txt
+++ b/libc-test/semver/linux-gnu.txt
@@ -609,6 +609,7 @@ aio_return
 aio_suspend
 aio_write
 aiocb
+aligned_alloc
 backtrace
 clock_adjtime
 copy_file_range
--- a/libc-test/semver/linux-musl.txt
+++ b/libc-test/semver/linux-musl.txt
@@ -56,6 +56,7 @@ aio_return
 aio_suspend
 aio_write
 aiocb
+aligned_alloc
 clock_adjtime
 copy_file_range
 ctermid
--- a/libc-test/semver/netbsd.txt
+++ b/libc-test/semver/netbsd.txt
@@ -1202,6 +1202,7 @@ aio_return
 aio_suspend
 aio_write
 aiocb
+aligned_alloc
 arc4random
 arc4random_buf
 arc4random_uniform
--- a/libc-test/semver/openbsd.txt
+++ b/libc-test/semver/openbsd.txt
@@ -1007,6 +1007,7 @@ __errno
 abs
 accept4
 acct
+aligned_alloc
 arc4random
 arc4random_buf
 arc4random_uniform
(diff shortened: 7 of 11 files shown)


safinaskar:
@devnexen , @JohnTitor , CI is failing in _main_, and I have some observations.

I made PR https://github.com/rust-lang/libc/pull/3693 . All checks have passed at some point in interval 11 May 00:28 - 01:29 UTC, so I assume _main_ ( 8b9d1feea295e92012c44eb17adb4c1d40a61ba0 ) was okay at that time.

But then at 11 May 18:56 UTC I noticed that very same commit in _main_ ( 8b9d1feea295e92012c44eb17adb4c1d40a61ba0 ) has failing CI with platform sparc64-unknown-linux-gnu.

So it seems some external resource was updated, and it broke CI

devnexen:
the sparc container tends to fail from time to time (e.g. timeout)

safinaskar:
@devnexen , sparc64 is broken for two days. As you can see, in all PRs for last two days sparc64 did not success.

Also, CI result in this very PR ( https://github.com/rust-lang/libc/pull/3698 ), says `qemu-system-sparc64: terminating on signal 15 from pid 1165 (timeout)`, but also says `/prog: /lib/sparc64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by /prog)`

( https://github.com/rust-lang/libc/actions/runs/9052276084/job/24869832722?pr=3698 )

tgross35:
It looks like this is available just about everywhere nowadays:

- linux glibc https://github.com/bminor/glibc/blob/b22923abb046311ac9097a36b97b9b97342bac44/stdlib/stdlib.h#L724
- linux musl https://github.com/kraj/musl/blob/2c41ee96f60074fa8ac387e1fcdedba0654fb173/include/stdlib.h#L44
- linux ulibc https://github.com/wbx-github/uclibc-ng/blob/ecbc412d0d93c8b7b77bec6539b290c9da5401cf/include/stdlib.h#L536
- newlib https://github.com/reswitched/newlib/blob/76f6d4e21ed5b621adb75fe99a69704dca831177/newlib/libc/stdlib/aligned_alloc.c#L30
- openbsd https://github.com/openbsd/src/blob/fd042ffdc70d1726d6308c4819514edac16eff53/lib/libc/stdlib/malloc.c#L2342-L2343
- netbsd https://man.netbsd.org/aligned_alloc.3
- freebsd https://man.freebsd.org/cgi/man.cgi?query=aligned_alloc&apropos=0&sektion=3&manpath=FreeBSD+9-current&format=html
- redox https://github.com/redox-os/relibc/blob/a38638542d0f7b0063da3c84833bd72c0f71da88/src/header/stdlib/mod.rs#L122
- haiku https://github.com/haiku/haiku/blob/ba4478ce60f4c0c3f00697981fc87d4e9e66b230/headers/posix/stdlib.h#L52
- aix (well z/os, but I assume they would be the same) https://www.ibm.com/docs/en/zos/2.5.0?topic=functions-aligned-alloc-allocating-aligned-memory-blocks
- nto/qnx https://www.qnx.com/developers/docs/8.0/com.qnx.doc.neutrino.lib_ref/topic/a/aligned_alloc.html

I can't find the Apple docs, but it seems like this is available in more recent versions. So could this just go in `unix/mod.rs`?

Also please target `main` and update the commit message with the changes.

@rustbot author

devnexen:
closed in favor of #3843

PR #3706, opened 2024-05-17
fix: use c_char from crate root

mekosko:
Most likely this is a typo and c_char from crate root should be used. If this is intentional and there must be an unsigned char, then it may be better to specify u8 instead of c_char.

### DIFF
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -128,7 +128,7 @@ s_no_extra_traits! {
     pub struct sockaddr_un {
         pub sun_len: u8,
         pub sun_family: sa_family_t,
-        pub sun_path: [c_char; 104]
+        pub sun_path: [::c_char; 104]
     }
 
     pub struct utsname {


tgross35:
This seems fine and it should indeed be a `char` per the docs https://man.freebsd.org/cgi/man.cgi?unix(4).

> If this is intentional and there must be an unsigned char, then it may be better to specify u8 instead of c_char

Is there any specific problem here or cases where `libc::c_char` does not line up with `std::ffi::c_char`? Usually they should coincide.

In any case, please rebase so the merge commit can be dropped.

@rustbot author

ChrisDenton:
I have a PR that intends to add a test that ensures that they are the same. https://github.com/rust-lang/libc/pull/3616

This is moved from rustc where we skip certain platforms that don't have libc in the sysroot. But we really should be testing them, hence moving the test into libc.

tgross35:
> > Is there any specific problem here or cases where `libc::c_char` does not line up with `std::ffi::c_char`? Usually they should coincide.
> 
> Yes, I was trying to port libc crate to proprietary system with network stack from bsd, this system's char is signed but for bsd it is unsigned, and they mismatched. (If I remember everything correctly, it was more than two months ago)

Meaning, `std` had the correct type but `libc` was wrong? If this is accurate, could you submit a PR here fixing that?

mekosko:
> Meaning, `std` had the correct type but `libc` was wrong? If this is accurate, could you submit a PR here fixing that?

I've added new module for that system in my libc fork, because libc does not have support for that system. In that module I've defined c_char as i8. Then I re-exported network related things from bsd module by specifying `#[path = "..."]` over submodule with dummy name and then pub use-ing everything from it. I don't think this is the right way to add new platforms, but it works fine if every c_<type> is used from crate root.

`core` crate thinks it is bare metal arm target and core::ffi::c_char = i8, it matched system's one, but it's just a coincidence :)

PR #3720, opened 2024-05-24
Add iopl/ioperm to all linuxes

espindola:


espindola:
@JohnTitor is this ok?

fslongjin (approved):
This can work well

espindola:
so, what is the process? I got an email saying @fslongjin approved this, what is next? Do we need another review?

tgross35:
These are Linux syscalls so should be available everywhere, not just x86. It should probably mirror `ioctl` and be at least in `unix/linux_like/linux/{gnu,musl,android}` (near ioctl). Please try this, CI will test the other platforms for you. The semver tests will need to be updated too.

I don't think there is any good reason to add `iopl` unless you have the xserver usecase, considering it has been deprecated since 2020 https://github.com/mkerrisk/man-pages/commit/6caf6c4380f2c6d706c3df6dcf2d8c3b9438d724.

- https://linux.die.net/man/2/ioperm
- https://man7.org/linux/man-pages/man2/iopl.2.html

@rustbot author

espindola:
iopl is used by https://github.com/pop-os/system76-firmware

I will try to add this to all platforms.

espindola:
Since you mentioned ioctl, I have rebased this on top of my other MR so that there is only one place to add these.

espindola:
> These are Linux syscalls so should be available everywhere, not just x86. It should probably mirror `ioctl` and be at least in `unix/linux_like/linux/{gnu,musl,android}` (near ioctl). Please try this, CI will test the other platforms for you. The semver tests will need to be updated too.

I just tried, and it looks like it is not available on arm or ppc. OK to leave it as x86_64/i686 only?

espindola:
so, in summary, iopl is needed, and it can't be declared next to iocl, because it is not available everywhere ioctl is.

tgross35:
Fyi you can

@rustbot review

To get the labels updated that this is ready for reveiew. I'll take another look.

tgross35 (approved):

PR #3722, opened 2024-05-25
Unify the ioctl declarations on linux

espindola:
I originally wanted to have all versions take a c_uint, since that is what the kernel expects on the other side. I then found that the test infrastructure, quite reasonably, enforces that the symbol matches what the C library provides.

So this PR is now just a small cleanup that uses the existing Ioctl type to reduce duplication.


tgross35 (approved):
This is a nice cleanup, thanks. Could you double check that the tests in `libc-test/semver` are up to date?

espindola:
> This is a nice cleanup, thanks. Could you double check that the tests in `libc-test/semver` are up to date?

I can't run the tests, since ctest2 complains that

```
  unknown os/family: x86_64-alpine-linux-musl
``

But they passed on CI and I didn't change what gets exported.

espindola:
> But they passed on CI and I didn't change what gets exported.

In fact, ioctl is listed in unix.txt, so it is already expected everywhere.

tgross35:
Thanks for confirming!

PR #3723, opened 2024-05-26
build.rs: always use freebsd12 when rustc_dep_of_std is set

RalfJung:
Currently, when rustc_dep_of_std is set, the freebsd version used depends on the return value of `which_freebsd()`:
- if it is `Some(12)`, it uses `freebsd12`
- otherwise it uses `freebsd11`

That does not seem intended? It is certainly rather odd that when building on FreeBSD 13 we'd get the freebsd11 API, but when building on FreeBSD 12 we get freebsd12. This also causes some "fun" for Miri when cross-building the freebsd std on other targets: `which_freebsd()` then returns `None`, so std uses different symbols than it does in the std that is shipped by rustup.

I assume the intention of https://github.com/rust-lang/libc/pull/3434 was to force std to use freebsd12 when rustc_dep_of_std is set (at least that's what the comments seem to say), so let's implement that.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -44,14 +44,21 @@ fn main() {
     //
     // On CI, we detect the actual FreeBSD version and match its ABI exactly,
     // running tests to ensure that the ABI is correct.
-    match which_freebsd() {
-        Some(10) if libc_ci => set_cfg("freebsd10"),
-        Some(11) if libc_ci => set_cfg("freebsd11"),
-        Some(12) if libc_ci || rustc_dep_of_std => set_cfg("freebsd12"),
-        Some(13) if libc_ci => set_cfg("freebsd13"),
-        Some(14) if libc_ci => set_cfg("freebsd14"),
-        Some(15) if libc_ci => set_cfg("freebsd15"),
-        Some(_) | None => set_cfg("freebsd11"),
+    let which_freebsd = if libc_ci {
+        which_freebsd().unwrap_or(11)
+    } else if rustc_dep_of_std {
+        12
+    } else {
+        11
+    };
+    match which_freebsd {
+        x if x < 10 => panic!("FreeBSD older than 10 is not supported"),
+        10 => set_cfg("freebsd10"),
+        11 => set_cfg("freebsd11"),
+        12 => set_cfg("freebsd12"),
+        13 => set_cfg("freebsd13"),
+        14 => set_cfg("freebsd14"),
+        _ => set_cfg("freebsd15"),
     }
 
     match emcc_version_code() {


build.rs
```diff
@@ -44,14 +44,21 @@ fn main() {
...
+        11
+    };
+    match which_freebsd {
+        10 => set_cfg("freebsd10"),
+        11 => set_cfg("freebsd11"),
+        12 => set_cfg("freebsd12"),
+        13 => set_cfg("freebsd13"),
+        14 => set_cfg("freebsd14"),
+        15 => set_cfg("freebsd15"),
+        v => panic!("unsupported FreeBSD version: {}", v),
```
asomers, line 61: Some day there will be a FreeBSD 16, so let's be forwards compatible with that.
```suggestion
        _ => set_cfg("freebsd15"),
```

asomers (changes requested):

build.rs
RalfJung: But isn't FreeBSD not-forward-compatible? It seems wrong to load the freebsd15 module on FreeBSD 16, no?

asomers: Do you  mean backwards-compatible?  It is.  It will always be safe to use a FreeBSD X ABI on a FreeBSD X+1 system.  You may be thinking of OpenBSD, which does not provide that guarantee.

RalfJung: Ah, yes I may be mixing things up. Sorry for that.

asomers (approved):

JohnTitor (approved):

RalfJung:
@JohnTitor any chance of seeing a libc release with this change any time soon? That should allow a bit of cleanup in Miri. :)

RalfJung:
Oh, I didn't realize that the branch for releases on crates.io is separate from `main`... should I just create a version of this PR against the libc-0.2 branch or how does that work?

tgross35:
The branches are unfortunately pretty divergent, so I have just been requesting everyone to always target `main` and `@rustbot label +stable-nominated` for things that should get backported. I cherry pick them in batches.

I'll get this into a release within the next couple of days.

RalfJung:
Okay cool, thanks :)

tgross35:
@RalfJung this should be available in 0.2.157 now

build.rs
```diff
@@ -44,14 +44,21 @@ fn main() {
     //
     // On CI, we detect the actual FreeBSD version and match its ABI exactly,
     // running tests to ensure that the ABI is correct.
-    match which_freebsd() {
-        Some(10) if libc_ci => set_cfg("freebsd10"),
-        Some(11) if libc_ci => set_cfg("freebsd11"),
-        Some(12) if libc_ci || rustc_dep_of_std => set_cfg("freebsd12"),
```
RalfJung, line 50: I just realized the previous logic for rustc_dep_of_std here wasn't what I thought it was...

The logic was: if the auto-detected version is 12 and `rustc_dep_of_std` is set, then use `freebsd12`. If the auto-detected version is anything else and `rustc_dep_of_std` is set, use `freebsd11`. If this was deliberate, I don't understand why, and it doesn't match the comments.

With my PR, the logic now is: if `rustc_dep_of_std` is set, don't even auto-detect, just use version 12. This matches what the comments already said before.

PR #3736, opened 2024-06-05
Update FreeBSD CI environment to 14.1

asomers:


JohnTitor (approved):

asomers:
@JohnTitor it looks like the latest nightly compiler objects to any struct with private fields that doesn't have a constructor method.  We'll have to figure out a way to suppress that.

JohnTitor:
Thank you for reporting! CI should now work.

PR #3738, opened 2024-06-05
Add missing getauxval for Linux uClibc targets

operutka:
This should fix the following error when building Rust `std` on Linux uClibc targets.

```
error[E0425]: cannot find function `getauxval` in crate `libc`
   --> /home/operutka/.rustup/toolchains/nightly-2024-06-05-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys/pal/unix/stack_overflow.rs:237:47
    |
237 |         let dynamic_sigstksz = unsafe { libc::getauxval(AT_MINSIGSTKSZ) };
    |                                               ^^^^^^^^^ not found in `libc`
```

### DIFF
--- a/src/unix/linux_like/linux/uclibc/mod.rs
+++ b/src/unix/linux_like/linux/uclibc/mod.rs
@@ -374,6 +374,7 @@ extern "C" {
     pub fn setrlimit(resource: ::__rlimit_resource_t, rlim: *const ::rlimit) -> ::c_int;
     pub fn getpriority(which: ::__priority_which_t, who: ::id_t) -> ::c_int;
     pub fn setpriority(which: ::__priority_which_t, who: ::id_t, prio: ::c_int) -> ::c_int;
+    pub fn getauxval(type_: ::c_ulong) -> ::c_ulong;
 }
 
 cfg_if! {


JohnTitor (approved):

operutka:
@JohnTitor the pull request has been removed from the merge queue due to some unrelated problems in the CI pipeline. Is there anything I can do to fix this?

JohnTitor:
Sorry! It has been fixed now.

skrap:
Hmm, I wonder what we should do for uclibc targets which do not include `getauxval`?  It's optional in some (all?) versions of uclibc, and for those of us working with unchangeable uclibc configurations, this means that we can't build rust without patching libc to include an implementation of `getauxval`.

Of course, that's not the fault of this PR; I only raise it here because the audience might have some thoughts.

One possibility: I could try to set up a `cfg` switch, similar to the `which_freebsd()` function from `build.rs`, which might be able to detect an older uclibc without a getauxval implementation and provide its own.  

OTOH I know we don't really want implementation in this crate.  Not sure where an alternative spot would be, though.

operutka:
@skrap I've also had to deal with this. The problem is that Rust std can't be built without `getauxval`, so removing (or feature-gating) it here doesn't really help. Right now I use a feature flag in my app that can enable an empty `getauxval` implementation. Something like:

```rust
#[cfg(feature = "fake-getauxval")]
#[no_mangle]
extern "C" fn getauxval(_: c_ulong) -> c_ulong {
    0
}
```

But, as you suggested, detecting it in `build.rs` and supplying an empty implementation if it isn't provided by the C library would be probably a better option.

I think the question is if Rust std should really depend on this function or if it can detect that the function isn't available. The `libc` crate could then export it only if it's actually available and Rust std wouldn't require it unless it's available. But that's probably a question for people more deeply involved in the libc and Rust std projects then I am :)

PR #3746, opened 2024-06-09
freebsd kcmp call support.

devnexen:


JohnTitor (approved):

JohnTitor (approved):

tgross35:
This doesn't work cleanly with 0.2, CI fails. Probably an easy fix but no need to do it unless there is a need. If anyone does have a need, cherry pick this per https://github.com/rust-lang/libc/blob/main/CONTRIBUTING.md#v10-roadmap and add the fix needed for CI.

@rustbot label -stable-nominated

PR #3756, opened 2024-06-21
FreeBSD: add evdev structures

Awoonyaa:


arrowd:
Some context: On FreeBSD the evdev headers are not part of the base system, but come from a package named `evdev-proto`. It seems this package should be installed to make this PR compile.

@devnexen You did some FreeBSD PRs already, could you please look at this one too?

devnexen:
> Some context: On FreeBSD the evdev headers are not part of the base system, but come from a package named `evdev-proto`. It seems this package should be installed to make this PR compile.

Just quick response here: by definition if it is not part of the system thus has no place in this crate, e.g. libutil is still part of builtin libraries. my 2 cents :)

arrowd:
Yes, but on Linux these data types are part of libc. Because of this, the consumers of this library that are using these types will fail to compile on FreeBSD. This is the case for the new release of https://github.com/mqudsi/syngesture

I think that putting these types here is the most correct solution, but I know nothing of Rust ecosystem.

arrowd:
>  nothing is stopping you define those types in the requested project

This would require patching for each project that depends on these types in libc.

> or to publish a crate dedicated to those sort of things ?

This would *probably* require patching too, although much less? Maybe it is a way to go then?

devnexen:
Regardless, to make CI work here you would need to add the related dependency. CI at the moment only install what necessaty, it appears, for rustup nothing else. The decision to merge it or not is more @JohnTitor call :)

devnexen:
I think you need to declare the appropriate evdev-proto header in `libc-test/build.rs` in the freebsd's list.

libc-test/build.rs
```diff
@@ -2066,6 +2066,7 @@ fn test_freebsd(target: &str) {
                 "sys/eui64.h",
                 "sys/event.h",
                 [freebsd13]:"sys/eventfd.h",
+                [freebsd13]:"sys/input.h",
```
devnexen, line 2069: not certain that s the right path

devnexen:
Also note you need to treat struct field such as `input_event::type_` as it is for sure not its real name, see other example how it is done in libc-test/build.rs.

Awoonyaa:
@devnexen thank you for the comments! Maybe you can advise me how to create a flag for checking file in /ust/local/include ?

Awoonyaa:
> > @devnexen thank you for the comments! Maybe you can advise me how to create a flag for checking file in /usr/local/include ?
> 
> Are you able to run tests locally ?

Yes

libc-test/build.rs
```diff
@@ -1990,6 +1990,9 @@ fn test_freebsd(target: &str) {
     cfg.define("_WITH_GETLINE", None);
     // Required for making freebsd11_stat available in the headers
     cfg.define("_WANT_FREEBSD11_STAT", None);
+    cfg.define("LIBICONV_PLUG", None);
+
+    cfg.flag("-I/usr/local/include");
```
arrowd, line 1995: Ideally, we should use the value of `sysctl -n user.localbase` here instead of hardcoding "/usr/local", but I don't know how to do it in Rust.

libc-test/build.rs
```diff
@@ -2091,6 +2095,7 @@ fn test_freebsd(target: &str) {
                 "sys/thr.h",
                 "sys/time.h",
                 [freebsd14 || freebsd15]:"sys/timerfd.h",
+                [freebsd14 || freebsd15]:"linux/input.h",
```
arrowd, line 2098: This can be just `freebsd13 || freebsd14 || freebsd15` and then remove line 2072.

tgross35:
I asked about this as a policy question in https://github.com/rust-lang/libc/issues/3839. Until then:

@rustbot blocked

jbeich:
> On FreeBSD the evdev headers are not part of the base system, but come from a package named `evdev-proto`.

`<linux/input.h>` is also available as [`<dev/evdev/input.h>`](https://github.com/freebsd/freebsd-src/blob/releng/14.1/sys/dev/evdev/input.h) (ditto [input-event-codes.h](https://github.com/freebsd/freebsd-src/blob/releng/14.1/sys/dev/evdev/input-event-codes.h) and [uinput.h](https://github.com/freebsd/freebsd-src/blob/releng/14.1/sys/dev/evdev/uinput.h)). For platform bindings it may be fine to use FreeBSD headers which are subset of Linux. This should help unblock progress here by sidestepping #3839.

CC @wulf7 who implemented evdev on FreeBSD and defined [evdev-proto policy](https://lists.freebsd.org/pipermail/freebsd-x11/2019-March/023097.html).

wulf7:
> > On FreeBSD the evdev headers are not part of the base system, but come from a package named `evdev-proto`.
> 
> `<linux/input.h>` is also available as [`<dev/evdev/input.h>`](https://github.com/freebsd/freebsd-src/blob/releng/14.1/sys/dev/evdev/input.h) (ditto [input-event-codes.h](https://github.com/freebsd/freebsd-src/blob/releng/14.1/sys/dev/evdev/input-event-codes.h) and [uinput.h](https://github.com/freebsd/freebsd-src/blob/releng/14.1/sys/dev/evdev/uinput.h)). For **static** platform bindings it may be fine to test against FreeBSD headers which are subset of Linux. This should help unblock progress here by sidestepping #3839.
> 
> CC @wulf7 who implemented evdev on FreeBSD and defined [evdev-proto policy](https://lists.freebsd.org/pipermail/freebsd-x11/2019-March/023097.html).

If this code is FreeBSD-only, you can rely on <dev/evdev/input.h>

arrowd:
@tgross35 Can we get this in? The change does not use headers from 3rd-party packages anymore, so #3839 is not unrelated.

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -280,6 +280,22 @@ s! {
...
+    }
+
+    pub struct input_absinfo {
+        pub value: c_int,
+        pub minimum: c_int,
+        pub maximum: c_int,
+        pub fuzz: c_int,
+        pub flat: c_int,
+        pub resolution: c_int,
+    }
```
tgross35, line 297: The integers can use `i32` to match https://github.com/freebsd/freebsd-src/blob/38b67578fb4bbf568f7012ca3921a4d15cfe7c5d/sys/dev/evdev/input.h#L51-L65

tgross35:
> @tgross35 Can we get this in? The change does not use headers from 3rd-party packages anymore, so #3839 is now unrelated.

Thanks for the ping, I didn't realize this was ready (for future reference, `@rustbot review` updates the labels).

@Awoonyaa one nit above and can you add these to `libc-test/semver/freebsd.txt`?

Cc @asomers

asomers:
The definitions look correct to me.  Personally I'd be tempted to expose these under an `evdev::` module rather than at the crate root.  But Linux already exposes them at the crate root, so I guess the horse has already left the barn.

tgross35:
> The definitions look correct to me. Personally I'd be tempted to expose these under an `evdev::` module rather than at the crate root. But Linux already exposes them at the crate root, so I guess the horse has already left the barn.

Thanks for taking a look. I have been thinking that modules mirroring C header paths might be worth doing for 1.0, then only reexporting a common subset for backward compatibility. Especially considering this would make our sources easier to navigate and cross reference to the system headers. But indeed we may as well be consistent here.

redwan-islam-rumman:
Hi

redwan-islam-rumman:
Where are you from?

Awoonyaa:
Can we get this in, please? Or is it something else we need to fix?

tgross35:
> Can we get this in, please? Or is it something else we need to fix?

I believe this is ready except for https://github.com/rust-lang/libc/pull/3756#discussion_r1895040761

tgross35 (approved):
Thanks!

PR #3758, opened 2024-06-22
hurd: Fix st_dev name

sthibaul:
The real GNU name is st_fsid, but the more commonly known name is st_dev, so better use it to avoid having to fix various rust package.

This also aligns on the existing struct stat field name.

### DIFF
--- a/src/unix/hurd/mod.rs
+++ b/src/unix/hurd/mod.rs
@@ -476,7 +476,7 @@ s! {
 
     pub struct stat64 {
         pub st_fstype: ::c_int,
-        pub st_fsid: __fsid_t,
+        pub st_dev: __fsid_t, /* Actually st_fsid */
         pub st_ino: __ino64_t,
         pub st_gen: ::c_uint,
         pub st_rdev: __dev_t,


JohnTitor:
Thanks! But this crate just provides raw wrappers and changing a name may bring another confusion. I think it's better it remains as-is.

sthibaul:
The  current state is *way* more than confusing: it's not coherent between `struct stat` and `struct stat64`.

Long-term wise, keeping this incoherency will be a plague. We'd *really* much better stick to only the well-known name.

So far I have not seen a rust software use the `st_fsid` name, so let's act *now* to avoid seeing them pop up.

sthibaul:
> So far I have not seen a rust software use the st_fsid name, so let's act now to avoid seeing them pop up.

@JohnTitor : On the other hand, I can quickly see ~45 rust software in debian that use the `st_dev` field name. Refusing to integrate this PR means imposing us the tedious work to go through all these to introduce an os-specific case, that this PR would avoid altogether. This is imposing us unnecessary work.

PR #3761, opened 2024-06-28
Adding constant SOMAXCONN to vxworks

biabbas:
Hi all, I'm just adding the SOMAXCONN variable that was missing in the VxWorks Rust libc. I've checked the libraries, and this is the correct value for the constant.


JohnTitor (approved):

biabbas:
Thank you. When will this be reflected in crates.io pulled crate?

PR #3764, opened 2024-07-02
Added kill to Windows

### LINKED ISSUE #3763: Missing kill function in libc on Windows
The kill function is missing in libc on Windows

chuacw:
Fixes #3763


src/windows/mod.rs
```diff
@@ -578,3 +589,16 @@ cfg_if! {
...
     }
 }
+
+pub fn kill(pid: pid_t, sig: c_int) -> ::c_int {
+    let dwDesiredAccess = PROCESS_TERMINATE;
+    let bInheritHandle: winapi::shared::minwindef::BOOL = false.into();
+    let dwProcessId = pid as _;
+    let result;
+    unsafe {
+        let hProcess = OpenProcess(dwDesiredAccess, bInheritHandle, dwProcessId);
```
devnexen, line 599: **unlikely** to be merged but wherever you implement it, I would recommend to check hProcess value before calling TerminateProcess/CloseHandle with.

chuacw: Fixed missing check for hProcess value

joshtriplett:
libc is in general not an abstraction layer attempting to provide POSIX APIs on non-POSIX platforms. While libc *does* attempt to abstract over minor differences between POSIX platforms so that callers don't have to deal with those differences, that doesn't extend to things like implementing a POSIX API layer atop Windows.

This level of abstraction would be the realm of a different crate.

PR #3769, opened 2024-07-09
macOS: add `os_sync_wait_on_address` and related definitions

joboet:
rust-lang/rust#122408 will use some of these definitions and it seems neater to add the rest as well. Unfortunately, Apple has forgotten to upload the complete documentation to their website, [here](https://github.com/apple-oss-distributions/libplatform/blob/main/include/os/os_sync_wait_on_address.h) is the complete header file for everything but [`os_clockid_t` and `OS_CLOCK_MACH_ABSOLUTE_TIME`](https://developer.apple.com/documentation/os/os_clockid_t?language=objc).


joboet:
It appears that the CI runner is on an older version of macOS. I'm not quite sure what to do here...

JohnTitor (approved):
Thanks!

PR #3771, opened 2024-07-11
Add POSIX confstr API

### LINKED ISSUE #3767: Missing POSIX confstr and _CS_* symbols on Linux.
Apple Darwin provides the POSIX standard `confstr(3)` function and associated `_CS_*` symbols.  Tested on Darwin and is working.

Requesting the same for Linux.

Quoting src/unix/bsd/apple/mod.rs:
```
// `confstr` keys (only the values guaranteed by `man confstr`).
pub const _CS_PATH: ::c_int = 1;
pub const _CS_DARWIN_USER_DIR: ::c_int = 65536;
pub const _CS_DARWIN_USER_TEMP_DIR: ::c_int = 65537;
pub const _CS_DARWIN_USER_CACHE_DIR: ::c_int = 65538;
...
    #[cfg_attr(
        link_name = "confstr$UNIX2003"
    )]
    pub fn confstr(name: ::c_int, buf: *mut ::c_char, len: ::size_t) -> ::size_t
```

The Linux version for this crate should be very similar.  The Linux man page for `confstr(3)` also helpfully lists the few key `_CS_*` symbols that the developer should expect to be available:
```
       _CS_GNU_LIBC_VERSION (GNU C library only; since glibc 2.3.2)
       _CS_GNU_LIBPTHREAD_VERSION (GNU C library only; since glibc 2.3.2)
       _CS_PATH
```

jgarzik:
Fixes #3767 

Some notes for reviewers:
* This PR was successfully tested with my Rust `getconf` util implementation at https://github.com/rustcoreutils/posixutils-rs
* It would be ideal, and seemingly appropriate, to add the `confstr(3)` API `unix`, because it is POSIX standard, just like `sysconf(3)`, and originates from libc, just like `sysconf(3)`.  This failed to due a single platform, arm-android.
* Therefore, an ideal future patch would populate other unix platforms with `confstr`.

This patch chooses the more limited path of focusing on linux-gnu, given that the symbols (not the function!) are glibc-specific.

### DIFF
--- a/libc-test/semver/linux-gnu.txt
+++ b/libc-test/semver/linux-gnu.txt
@@ -511,6 +511,9 @@ XSK_UNALIGNED_BUF_ADDR_MASK
 XDP_PKT_CONTD
 XENFS_SUPER_MAGIC
 XFS_SUPER_MAGIC
+_CS_GNU_LIBC_VERSION
+_CS_GNU_LIBPTHREAD_VERSION
+_CS_PATH
 _SC_2_C_VERSION
 _SC_BASE
 _SC_CHARCLASS_NAME_MAX
@@ -611,6 +614,7 @@ aio_write
 aiocb
 backtrace
 clock_adjtime
+confstr
 copy_file_range
 ctermid
 dlinfo
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -826,6 +826,9 @@ pub const TMP_MAX: ::c_uint = 238328;
 pub const FOPEN_MAX: ::c_uint = 16;
 pub const FILENAME_MAX: ::c_uint = 4096;
 pub const POSIX_MADV_DONTNEED: ::c_int = 4;
+pub const _CS_GNU_LIBC_VERSION: ::c_int = 2;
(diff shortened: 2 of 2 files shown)


jgarzik:
Status:  all tests _except_ `arm-linux-androidabi` are passing.  Unclear what is the best solution around this test failure.

devnexen:
```
Missing functions are either obsolete or explicitly disallowed by SELinux:

a64l/l64a
confstr
crypt/encrypt/setkey
gethostid
shm_open/shm_unlink
sockatmark
ualarm
```

[source](https://android.googlesource.com/platform/bionic/+/HEAD/docs/status.md)

jgarzik:
> fair enough, even tough I believe musl for example has it too for example but that s fair as is :)

The Platonic ideal would seem to be "confstr on all unix _except_ x, y, and z" but that is difficult to express in the semver tests, I'm guessing.

Further complicating the task is that value of `_CS_PATH` is not specified and may vary across confstr-supporting OS's.  It is a POSIX-wide API with local-libc-specific values and symbology (thus the difference in Darwin and glibc symbols _CS_*)

devnexen (approved):

JohnTitor (approved):

PR #3773, opened 2024-07-18
add `pthread_equal`

joboet:
From its [documentation](https://man7.org/linux/man-pages/man3/pthread_equal.3.html):
> The `pthread_equal()` function is necessary because thread IDs should be considered opaque: there is no portable way for applications to directly compare two `pthread_t` values.

### DIFF
--- a/libc-test/semver/unix.txt
+++ b/libc-test/semver/unix.txt
@@ -687,6 +687,7 @@ pthread_condattr_init
 pthread_condattr_t
 pthread_create
 pthread_detach
+pthread_equal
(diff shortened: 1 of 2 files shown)


tgross35:
Can you change this to target `main`? Can cherry pick to stable after merge.

tgross35 (approved):
Thanks!

tgross35:
I don't know why freebsd would have failed, it should be available https://man.freebsd.org/cgi/man.cgi?query=pthread_equal&sektion=3.

PR #3776, opened 2024-07-18
Fix CI by ignoring `rmx_state` on Apple platforms

joboet:
`rmx_state` has been removed in https://github.com/apple-oss-distributions/xnu/commit/94d3b452840153a99b38a3a9659680b2a006908e (macOS version 14.4) in favour of an expanded `rmx_filler` field. Just ignore it in CI.

### DIFF
--- a/src/wasi.rs
+++ b/src/wasi.rs
@@ -366,7 +366,7 @@ pub const _SC_PAGE_SIZE: ::c_int = _SC_PAGESIZE;
 pub const _SC_IOV_MAX: c_int = 60;
 pub const _SC_SYMLOOP_MAX: c_int = 173;
 
-#[allow(unused_unsafe)] // `addr_of!(EXTERN_STATIC)` is now considered safe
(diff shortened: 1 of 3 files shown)


JohnTitor (approved):
Thanks!

JohnTitor:
Oh, seems another failure appears.

joboet:
I think that's caused by rust-lang/rust#125834, which changed `addr_of!(EXTERN_STATIC)` to be safe.

src/wasi.rs
```diff
@@ -366,10 +366,14 @@ pub const _SC_PAGE_SIZE: ::c_int = _SC_PAGESIZE;
 pub const _SC_IOV_MAX: c_int = 60;
 pub const _SC_SYMLOOP_MAX: c_int = 173;
 
+#[allow(unused_unsafe)] // `addr_of!(EXTERN_STATIC)` is now considered safe
```
joshtriplett, line 369: Please add comments on things like this saying something like "we can remove the unsafe when our MSRV is 1.82 or newer", so that they're easy to find when updating MSRV

roblabla (approved):

joshtriplett (approved):

PR #3777, opened 2024-07-19
Add `klogctl` to linux and android

sgasse:
Those are useful to read the kernel message buffer.

References:
- https://android.googlesource.com/platform/bionic/+/master/libc/include/sys/klog.h
- https://man7.org/linux/man-pages/man2/syslog.2.html


src/unix/linux_like/linux/mod.rs
```diff
@@ -5620,6 +5620,8 @@ extern "C" {
         len: ::size_t,
         flags: ::c_uint,
     ) -> ::ssize_t;
+
+    pub fn klogctl(syslog_type: ::c_int, buf: *mut ::c_char, len: ::c_int) -> ::c_int;
```
tgross35, line 5624: Nit: Linux docs call the second arg `bufp` per https://linux.die.net/man/3/klogctl

tgross35:
Could you target `main` please? Otherwise lgtm, with one nit above.

@rustbot author

sgasse:
Thank you for the review @tgross35 ! I updated the argument name as suggested in your review. Next I changed the base of the PR to `main`, rebased locally on `upstream/main` and pushed it (hopefully not breaking anything and not generating noise due to the base change).

tgross35 (approved):
Thanks!

PR #3779, opened 2024-07-20
Add missing constants for Android

yujincheng08:


tgross35:
@maurer would you mind taking a look at this?

maurer:
This looks fine to me, the reason those constants were 64-bit only before was as a hacky form of version detection. However, now that our min API level is 21, we shouldn't need to worry about that, and making them available to 32-bit should be appropriate. The rest of this appears to be appropriate constant migration.

The only thing I'll note is that since this bumps the NDK to the latest stable (r27), we'll want to bump it in `rustc` as well before integrating this change there after it lands. (I don't think it'll be as troublesome as last time, but we'll have to do the whole song and dance with updating the mirror again.)

yujincheng08:
thanks @maurer 

r? @tgross35 could you check if it's now okay to merge it?

yujincheng08:
@tgross35 hi, would you mind taking a look at this?

src/unix/linux_like/android/b64/aarch64/mod.rs
```diff
@@ -433,6 +433,10 @@ pub const SYS_syscalls: ::c_long = 451;
 pub const PROT_BTI: ::c_int = 0x10;
 pub const PROT_MTE: ::c_int = 0x20;
 
+// From NDK's asm/auxvec.h
+pub const AT_SYSINFO_EHDR: ::c_ulong = 33;
+pub const AT_VECTOR_SIZE_ARCH: ::c_ulong = 2;
```
tgross35, line 438: Is `AT_MINSIGSTKSZ` useful? Since it is also in https://android.googlesource.com/platform/bionic/+/refs/tags/android-14.0.0_r61/libc/kernel/uapi/asm-arm64/asm/auxvec.h

src/unix/linux_like/android/b64/riscv64/mod.rs
```diff
@@ -362,5 +362,17 @@ pub const SYS_process_mrelease: ::c_long = 448;
...
+pub const AT_L1I_CACHESIZE: ::c_ulong = 40;
+pub const AT_L1I_CACHEGEOMETRY: ::c_ulong = 41;
+pub const AT_L1D_CACHESIZE: ::c_ulong = 42;
+pub const AT_L1D_CACHEGEOMETRY: ::c_ulong = 43;
+pub const AT_L2_CACHESIZE: ::c_ulong = 44;
+pub const AT_L2_CACHEGEOMETRY: ::c_ulong = 45;
+pub const AT_L3_CACHESIZE: ::c_ulong = 46;
+pub const AT_L3_CACHEGEOMETRY: ::c_ulong = 47;
+pub const AT_VECTOR_SIZE_ARCH: ::c_ulong = 9;
+
```
tgross35, line 376: These should get an API check file, could you start `libc-test/semver/android-riscv64.txt`?

Also same as aarch64, any point of `AT_MINSIGSTKSZ`?

tgross35:
Please be patient with reviews, there are a lot of open PRs.

Relevant headers for comparison:

- https://android.googlesource.com/platform/bionic/+/refs/tags/android-14.0.0_r61/libc/kernel/uapi/linux/auxvec.h
- https://android.googlesource.com/platform/bionic/+/refs/tags/android-14.0.0_r61/libc/kernel/uapi/asm-arm/asm/auxvec.h
- https://android.googlesource.com/platform/bionic/+/refs/tags/android-14.0.0_r61/libc/kernel/uapi/asm-arm64/asm/auxvec.h
- https://android.googlesource.com/platform/bionic/+/refs/tags/android-14.0.0_r61/libc/kernel/uapi/asm-riscv/asm/auxvec.h
- https://android.googlesource.com/platform/bionic/+/refs/tags/android-14.0.0_r61/libc/kernel/uapi/asm-x86/asm/auxvec.h

A few questions below then LGTM. Thanks Matthew for taking a look too.

@rustbot author

src/unix/linux_like/android/b64/riscv64/mod.rs
yujincheng08: Hi, I tried to add an API check for riscv but discovered that CI won't check it because it's tier 3. Is it still reasonable to add this since it may break without noticing because CI won't check?

src/unix/linux_like/android/b64/aarch64/mod.rs
yujincheng08: This constant is in `src/unix/linux_like/android/mod.rs`

tgross35: Ah didn't catch that it was `ifndef` in `linux/auxvec.h`, thanks

src/unix/linux_like/android/b64/riscv64/mod.rs
tgross35: It won't get run in CI, but we may as well make it available for anyone who runs tests locally. And it will give us a starting point if the target is ever promoted.

yujincheng08: sure. I will try to make one.
btw I used to make it manually, is there any tools for autogen?

tgross35: It's just manual. It would be nice if it also made note of things that aren't listed to make this easier, but I'm not sure how exactly that would work.

yujincheng08: Done

tgross35:
Oh I only meant to start with the API added here for the risc-v file. But no harm of course :)

Please rebase to drop the merge commit, then looks good to me.

yujincheng08:
Rebased

tgross35 (approved):
Thanks!

juankyc (approved):

PR #3782, opened 2024-07-22
Add MODULE_INIT_COMPRESSED_FILE definition

sinkingpoint:
This adds in the definition for MODULE_INIT_COMPRESSED_FILE so that we can use it in the init_module class of functions to load compressed modules


src/unix/linux_like/linux/mod.rs
```diff
@@ -4043,6 +4043,7 @@ pub const RTNLGRP_STATS: ::c_uint = 0x24;
 // linux/module.h
 pub const MODULE_INIT_IGNORE_MODVERSIONS: ::c_uint = 0x0001;
 pub const MODULE_INIT_IGNORE_VERMAGIC: ::c_uint = 0x0002;
+pub const MODULE_INIT_COMPRESSED_FILE: ::c_uint = 0x0004;
```
SteveLauC, line 4046: Hi, Nix kmod module is available on both Linux and Android, please add this for Android as well:)

ref: https://github.com/search?q=repo%3Aaosp-mirror%2Fplatform_bionic%20MODULE_INIT_COMPRESSED_FILE&type=code

This should be also done to #3783.

sinkingpoint: Done!

tgross35 (approved):
Source link: https://github.com/torvalds/linux/blob/d74da846046aeec9333e802f5918bd3261fb5509/include/uapi/linux/module.h#L8

Thanks!

tgross35:
It looks like there may be some platforms where this isn't available. Could you please take a look at the failures on the latest CI?

@rustbot author

nathaniel-bennett:
It looks like `MOD_INIT_COMPRESSED_FILE` was added in Linux 5.17. My guess is that the arm test is failing because it's the first test that runs on a version of Linux older than that.

tgross35:
@sinkingpoint could you rebase to get a fresh run of tests? This probably just needs something added in `libc-test/build.rs` to account for the platforms this is available on, as mentioned by @nathaniel-bennett

tgross35:
I pushed your branch with a rebase just to get fresh CI. Seems like this will need to be disabled on at least a few platforms.

tgross35:
@sinkingpoint gentle ping, are you able to update this?

PR #3783, opened 2024-07-22
Add MODULE_INIT_COMPRESSED_FILE definition

sinkingpoint:
Opening this alongside https://github.com/rust-lang/libc/pull/3782 to backport the const into libc-0.2 so that we can use it in [nix](https://github.com/nix-rust/nix)


tgross35:
We are slightly changing the workflow here - PRs should target `main`, then they will get cherry-picked to stable if applicable. So I am going to close this and stable-nominate https://github.com/rust-lang/libc/pull/3782 (which I will review once I get the chance).

PR #3784, opened 2024-07-22
Define SO_BINDTOIFINDEX on Fuchsia

ewilden:
Fuchsia supports SO_BINDTOIFINDEX as of API level 20: https://cs.opensource.google/fuchsia/fuchsia/+/main:sdk/lib/zxio/socket.cc;l=755;drc=b03121152bf13fb8898a95b58d952c95ee73cd0c


ewilden:
I assume the failed CI check for macOS is a flake? I don't think this change would have impacted macOS :shrug:

devnexen:
no I meant your change has nothing to do with the macos failure

PR #3785, opened 2024-07-25
hurd: Fix st_dev name

sthibaul:
Currently struct stat and struct stat64 are not coherent: struct stat is using st_dev, and struct stat64 is using st_fsid.

st_dev is the more commonly-known name, already used by e.g. ~45 rust software in Debian, so better fix st_fsid into st_dev, rather than having to uselessly spend time hand-patching all these software.

### DIFF
--- a/src/unix/hurd/mod.rs
+++ b/src/unix/hurd/mod.rs
@@ -481,7 +481,7 @@ s! {
 
     pub struct stat64 {
         pub st_fstype: ::c_int,
-        pub st_fsid: __fsid_t,
+        pub st_dev: __fsid_t, /* Actually st_fsid */
         pub st_ino: __ino64_t,
         pub st_gen: ::c_uint,
         pub st_rdev: __dev_t,


sthibaul:
I am re-submitting here, since my comments on https://github.com/rust-lang/libc/pull/3758 don't seem to be getting answers.

tgross35:
Brought this up for policy in https://github.com/rust-lang/libc/issues/3839, that should at least give us better guidelines.

@rustbot blocked

tgross35:
This hasn't been discussed yet, but I am inclined to just merge this to unstick things. Do you have a link to the relevant C headers?

sthibaul:
Sure, here it is: https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/mach/hurd/bits/stat.h;hb=HEAD#l38

tgross35 (approved):
Thanks, let's get this in.

SteveLauC:
Looks like this PR breaks the std, see Nix CI [log](https://github.com/nix-rust/nix/actions/runs/12033180710/job/33546728266?pr=2544):

```
error[E0609]: no field `st_fsid` on type `&stat64`
   --> /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/os/hurd/fs.rs:301:36
    |
301 |         self.as_inner().as_inner().st_fsid as u64
    |                                    ^^^^^^^ unknown field
    |
help: a field with a similar name exists
    |
301 |         self.as_inner().as_inner().st_uid as u64
    |                                    ~~~~~~
```

tgross35:
Can you just disable that test in nix for now? Some breakage was expected here (it's tier3 and this PR is authored by the Hurd target maintainer), `std` just needs to be updated.

SteveLauC:
> std just needs to be updated.

I will send a patch to the std

PR #3786, opened 2024-07-26
hurd: Add missing struct __timeval for 64bit support

sthibaul:
This is the same as linux_like/linux/gnu's __timeval for ut_tv for 64bit architectures.

### DIFF
--- a/src/unix/hurd/mod.rs
+++ b/src/unix/hurd/mod.rs
@@ -449,6 +449,11 @@ s! {
         pub tv_nsec: __syscall_slong_t,
     }
 
+    pub struct __timeval {
+        pub tv_sec: i32,
+        pub tv_usec: i32,
+    }
+
     pub struct __locale_data {
         pub _address: u8,
     }


tgross35:
Can you post a link to the relevant docs/source here? `libc-test/semver` should probably also get a new `hurd` file with this struct (more can be added later) if you are able to run the test.

@rustbot author

sthibaul:
> Can you post a link to the relevant docs/source here?

As mentioned, this is exactly the same as linux with glibc: https://github.com/rust-lang/libc/blob/main/src/unix/linux_like/linux/gnu/mod.rs#L75

for the same reason: the `ut_tv` field of `struct utmpx` uses it in the 64bit case: https://github.com/rust-lang/libc/blob/main/src/unix/linux_like/linux/gnu/mod.rs#L584 and https://github.com/rust-lang/libc/blob/main/src/unix/hurd/mod.rs#L1058

because glibc uses a 32-bit-compatible timeval for that field: https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/utmp.h;h=27cb536800c46d67c0627dddb66db4ad737d356f;hb=HEAD#l72

> libc-test/semver should probably also get a new hurd file with this struct

We don't have a hurd file there for now.

tgross35:
Thanks for the follow up, sounds reasonable to me.

PR #3788, opened 2024-07-31
[illumos] add pthread stack functions

sunshowers:
The goal of this change is to make the following improvements possible:

* https://github.com/rust-lang/stacker/pull/88
* https://github.com/rust-lang/rust/issues/128568

Add some pthread stack functions:

* [`pthread_attr_get_np`](https://illumos.org/man/3C/pthread_attr_get_np)
* [`pthread_attr_getstackaddr`](https://illumos.org/man/3C/pthread_attr_getstackaddr)
* [`pthread_attr_setstack`](https://illumos.org/man/3C/pthread_attr_setstack)
* [`pthread_attr_setstackaddr`](https://illumos.org/man/3C/pthread_attr_setstackaddr)

I've verified that these functions exist and do the right thing.

Note that there doesn't appear to be a file like `linux.txt` for illumos specifically.

Another note: `cd libc-test && cargo test` currently fails on illumos with:

```
     Running test/main.rs (/home/rain/dev/libc/target/debug/deps/main-6329db179eb2b631)
RUNNING ALL TESTS
bad PTHREAD_MUTEX_DEFAULT value at byte 0: rust: 0 (0x0) != c 8 (0x8)
thread 'main' panicked at /home/rain/dev/libc/target/debug/build/libc-test-cefbf7853a50b4e4/out/main.rs:12:21:
some tests failed
```

But it also fails on main with the same error, so this isn't a regression.

### DIFF
--- a/libc-test/semver/illumos.txt
+++ b/libc-test/semver/illumos.txt
@@ -1,0 +1,3 @@
+pthread_attr_get_np
+pthread_attr_getstackaddr
+pthread_attr_setstack
--- a/src/unix/solarish/illumos.rs
+++ b/src/unix/solarish/illumos.rs
@@ -81,6 +81,21 @@ extern "C" {
     ) -> ::c_int;
     pub fn pset_getloadavg(pset: ::psetid_t, load: *mut ::c_double, num: ::c_int) -> ::c_int;
 
+    pub fn pthread_attr_get_np(thread: ::pthread_t, attr: *mut ::pthread_attr_t) -> ::c_int;
+    pub fn pthread_attr_getstackaddr(
+        attr: *const ::pthread_attr_t,
+        stackaddr: *mut *mut ::c_void,
+    ) -> ::c_int;
+    pub fn pthread_attr_setstack(
+        attr: *mut ::pthread_attr_t,
+        stackaddr: *mut ::c_void,
+        stacksize: ::size_t,
+    ) -> ::c_int;
+    pub fn pthread_attr_setstackaddr(
+        attr: *mut ::pthread_attr_t,
+        stackaddr: *mut ::c_void,
+    ) -> ::c_int;
+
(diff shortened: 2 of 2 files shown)


sunshowers:
Oh I guess I should target the 0.2 branch. Just a sec

pfmooney (approved):
The tests are passing on illumos with this change included.

tgross35:
Could you add an `illumos.txt` file in `libc-test/semver` and add these?

There are probably other functions that should be added once the file exists, but those can come separately.

tgross35:
Also, can you target `main`? We can cherry pick from there.

sunshowers:
Thanks @tgross35! Can do both of those things. (I wasn't sure about main vs 0.2, is there documentation somewhere about that? Sorry if I missed it.)

tgross35:
> (I wasn't sure about main vs 0.2, is there documentation somewhere about that? Sorry if I missed it.)

The README actually does suggest targeting 0.2, so you correctly did what is documented. I just intend to change that so we always update `main` and cherry-pick things to stable where needed, rather than maybe forgetting to apply some things to `main` and then losing fixes when 1.0 is released.

The `stable-nominated` label can be applied with `rustbot` for any specific requests.

sunshowers:
> I just intend to change that so we always update main and cherry-pick things to stable where needed, rather than maybe forgetting to apply some things to main and then losing fixes/additions when 1.0 is released.

That makes sense! (Landing on main first and then cherry-picking is also the flow I personally recommend to folks :) )

Anyway, updated the PR. (Sorry about the rustbot ping, should have updated the base first.)

tgross35:
Thanks for the changes! Feel free to submit a cherry-pick PR yourself once this merges, otherwise I will be doing a batch of them at some point.

sunshowers:
> Thanks for the changes! Feel free to submit a cherry-pick PR yourself once this merges (updated instructions coming here #3804), otherwise I will be doing a batch of them at some point.

This is https://github.com/rust-lang/libc/pull/3807, thanks!

PR #3791, opened 2024-08-05
Rebase of "Upgrade musl supported version to 1.2.3"

### LINKED ISSUE #1848: musl: Change `time_t` definition on 32-bit targets according to "time64"

myl7:
Also discussed in [GSoC Zulip]. Related: [GSoC Idea] [OSPP Idea] [OSPP Zulip].

[GSoC Idea]: https://github.com/rust-lang/google-summer-of-code?tab=readme-ov-file#modernize-the-libc-crate
[GSoC Zulip]: https://rust-lang.zulipchat.com/#narrow/stream/421156-gsoc/topic/Idea.3A.20modernize.20the.20libc.20crate
[OSPP Idea]: https://summer-ospp.ac.cn/org/prodetail/241170528
[OSPP Zulip]: https://rust-lang.zulipchat.com/#narrow/stream/436418-ospp/topic/Project.3A.20Modernize.20the.20libc.20crate

This PR is mainly the rebase of #3068 . So closes #3068 fixes #1848 closes #2088 .
Some other changes are made:

- Fix warnings in CI by setting several `#[allow(unused_imports)]`.

For the checklist for review or CI, as the original PR should have passed it, this PR only makes sure `ci/style.sh` and `cd libc-test && cargo test` pass


semarie:
FYI, no problem on OpenBSD side

tgross35:
Could you please rebase once https://github.com/rust-lang/libc/pull/3797 lands? (Hopefully ~1 hour)

myl7:
Glad to see most pass!

The only failure of `wasm32-unknown-emscripten` is because some code generated by ctest2 does not add `struct` prefix to struct names.
However, it should be added according to:

https://github.com/rust-lang/libc/blob/fc282833449475d9e7a4a56b9caa1f3e938435b5/libc-test/build.rs#L2744

Not sure how to resolve it.

tgross35:
Are you able to just skip the type with a fixme? I think we have emscripten-specific ignores already. Also not sure why this would be happening, but I am really hoping we don't need to block this on `ctest` changes.

tgross35:
Also @kaniini  I think you had some ideas about how to do musl better here. If you get a chance, would you mind taking a look at this?

build.rs
```diff
@@ -213,3 +220,22 @@ fn set_cfg(cfg: &str) {
...
+            | "i586-unknown-linux-musl"
+            | "i686-unknown-linux-musl"
+            | "mips-unknown-linux-musl"
+            | "mipsel-unknown-linux-musl"
+            | "powerpc-unknown-linux-musl" => true,
+            _ => false,
+        },
+        Err(_) => false,
+    }
+}
```
tgross35, line 241: We probably shouldn't match literal targets here since we only care about the prefix and the suffix. Instead could we:

- Add `const TARGET_PREFIXES_32BIT: &[&str] = &[/* ... */];` that contains `arm-`, `i586-`, etc
- Add `const MUSL_SUFFIXES : &[&str] = &[/* ... */];` with `-musleabi`, `-musl`, etc
- Add `fn target_is_32bit(target: &str) -> bool` that checks if it starts if `target` starts with any of `TARGET_PREFIXES_32BIT`. Then we can reuse it with the GNU change.
- Add `fn target_is_musl(target: &str) -> bool` that does the same with the suffixes
- Add `fn target_is_linux(target: &str) -> bool` that checks if it contains `-linux-`
- Check these three functions in order to set `musl_redir_time64`/`musl_time64_abi`

libc-test/build.rs
```diff
@@ -3301,6 +3301,23 @@ fn test_linux(target: &str) {
...
+        | "armv7-unknown-linux-musleabihf"
+        | "i586-unknown-linux-musl"
+        | "i686-unknown-linux-musl"
+        | "mips-unknown-linux-musl"
+        | "mipsel-unknown-linux-musl"
+        | "powerpc-unknown-linux-musl" => {
+            cfg.cfg("musl_time64_abi", None);
+        }
+        _ => {}
+    }
```
tgross35, line 3320: Why do we need this in `build.rs` if it's set automatically by our crate's build script?

build.rs
```diff
@@ -67,6 +68,12 @@ fn main() {
         Some(_) | None => (),
     }
 
+    // Some ABIs need to redirect time related symbols to their time64
+    // equivalents. See #2088 and #1848 for more information.
+    if is_musl_time64_abi() {
+        set_cfg("musl_time64_abi");
```
tgross35, line 74: We should set two separate configs here:

- `musl_redir_time64` (`musl_time64_abi`) which corresponds to musl's _REDIR_TIME64
- `linux_time_bits64` which corresponds to `__USE_TIME_BITS64` in uapi

We need this to be split because musl and glibc share the kernel uapi

src/unix/linux_like/linux/mod.rs
```diff
@@ -322,6 +322,11 @@ s! {
     }
 
     pub struct input_event {
+        #[cfg(musl_time64_abi)]
+        pub input_event_sec: ::c_ulong,
+        #[cfg(musl_time64_abi)]
+        pub input_event_usec: ::c_ulong,
+        #[cfg(not(musl_time64_abi))]
         pub time: ::timeval,
```
tgross35, line 330: As mentioned above this should make use of `linux_time_bits64`. Specifically to match https://github.com/torvalds/linux/blob/bf9aa14fc523d2763fc9a10672a709224e8fcaf4/include/uapi/linux/input.h#L29 it could be:

```rust
#[cfg(any(target_pointer_width = "64", not(linux_time_bits64))]
pub time: ::timeval,

#[cfg(not(any(target_pointer_width = "64", not(linux_time_bits64)))]
pub input_event_sec: ::c_ulong,
#[cfg(all(not(any(target_pointer_width = "64", not(linux_time_bits64))), target_arch = "sparc64")]
pub input_event_usec: ::c_uint,
#[cfg(all(not(any(target_pointer_width = "64", not(linux_time_bits64))), target_arch = "sparc64")]
pub __pad: ::c_uint,
#[cfg(all(not(any(target_pointer_width = "64", not(linux_time_bits64))), not(target_arch = "sparc64"))]
pub input_event_usec: ::c_ulong,
```

unfortunately a bit messy but I think that's accurate

src/unix/linux_like/linux/mod.rs
```diff
@@ -5428,43 +5435,6 @@ cfg_if! {
...
-                abs_timeout: *const ::timespec,
-            ) -> ::c_int;
-            pub fn mq_getattr(mqd: ::mqd_t, attr: *mut ::mq_attr) -> ::c_int;
-            pub fn mq_setattr(
-                mqd: ::mqd_t,
-                newattr: *const ::mq_attr,
-                oldattr: *mut ::mq_attr
-            ) -> ::c_int;
-
-            pub fn pthread_mutex_consistent(mutex: *mut pthread_mutex_t) -> ::c_int;
```
tgross35, line 5467: Did these functions move because they are now available on openharmony? If so, that should go in a different commit preferably.

tgross35 (changes requested):
I didn't check all the types in depth yet but this looks good at first glance. Please address the feedback (specifically `linux_time_bits64`) and then I think we can get this in soon.

safinaskar:
Possible solution to "how to distinguish 32 and 64" problem: https://github.com/rust-lang/rfcs/pull/3716 ("[RFC] Target Modifiers")

tgross35:
@myl7 are you able to address the above comments?

src/unix/linux_like/linux/musl/b32/arm/mod.rs
```diff
@@ -15,35 +16,22 @@ s! {
...
         pub st_atime: ::time_t,
         pub st_atime_nsec: ::c_long,
+        __st_atime_nsec_padding: ::c_long,
         pub st_mtime: ::time_t,
         pub st_mtime_nsec: ::c_long,
+        __st_mtime_nsec_padding: ::c_long,
         pub st_ctime: ::time_t,
         pub st_ctime_nsec: ::c_long,
-        pub st_ino: ::ino_t,
+        __st_ctime_nsec_padding: ::c_long,
```
tammela, line 34: I believe the padding here is only considering little endian.
The big endian padding happens in between `atime` and `atime_nsec` as it's defined here:
https://git.musl-libc.org/cgit/musl/tree/include/alltypes.h.in#n44

Cc @tgross35

src/unix/linux_like/linux/mod.rs
tammela: From what I read the C struct is:
```
struct input_event {
#if (__BITS_PER_LONG != 32 || !defined(__USE_TIME_BITS64)) && !defined(__KERNEL__)
        struct timeval time;
#     define input_event_sec time.tv_sec
#     define input_event_usec time.tv_usec
#else
        __kernel_ulong_t __sec;
#     if defined(__sparc__) && defined(__arch64__)
        unsigned int __usec;
        unsigned int __pad;
#     else
        __kernel_ulong_t __usec;
#     endif
#     define input_event_sec  __sec
#     define input_event_usec __usec
#endif
        __u16 type;
        __u16 code;
        __s32 value;
};
```

```
#[cfg(any(target_pointer_width = "64"))]
pub time: crate::timeval,

#[cfg(all(target_pointer_width = "32", linux_time_bits64)]
__sec: c_kernel_ulong_t,

#[cfg(all(target_pointer_width = "32", linux_time_bits64, target_arch = "sparc64"))]
__usec: c_int,
#[cfg(all(target_pointer_width = "32", linux_time_bits64, target_arch = "sparc64"))]
__pad: c_int,

#[cfg(all(target_pointer_width = "32", linux_time_bits64, not(target_arch = "sparc64")))]
__usec: c_kernel_ulong_t,
```

Note that `kernel_ulong` can be overridden to `unsigned long long` depending on the architecture which is guaranteed 64 bits:
https://elixir.bootlin.com/linux/v6.13.4/source/arch/x86/include/uapi/asm/posix_types_x32.h#L15

martinetd: (This PR will likely need a new volunteer to rebase and  sort all this out, but while I was looking at this...)
This:
```
#[cfg(all(target_pointer_width = "64", not(linux_time_bits64)))]
pub time: crate::timeval,

#[cfg(all(target_pointer_width = "32", linux_time_bits64)]
__sec: c_kernel_ulong_t,
```
looks odd as it's not a plain if else (e.g. 64bit + linux_time_bits64 is not covered)

Since it's a `||` in C my understanding is that the top timeval should just be changed all -> any, e.g. `any(target_pointer_width = "64", not(linux_time_bits64))` as @tgross35 wrote.

Rest looks good (as `not(any(a, b))` = `all(not(a), not(b))`), it's a sensible simplification that's probably easier to read.

tgross35: Oops I probably should have closed this, but this is in progress at https://github.com/rust-lang/libc/pull/4443 and https://github.com/rust-lang/libc/pull/4463. Thanks for the comments though - help is absolutely welcome with further design, cleanup, and testing :)

tgross35:
https://github.com/rust-lang/libc/pull/4443 merged with the basic support and https://github.com/rust-lang/libc/pull/4463 is in progress to add the time fixes, so this has effectively been superseded. Thanks for the work here!

src/unix/linux_like/linux/mod.rs
martinetd: Thanks for the pointers, I'll try to find time to have a look in the coming ~~da~~weeks!

tgross35: Please also feel free to leave reviews on any of these PRs if you've already done some looking into this!

PR #3792, opened 2024-08-05
Modify QNX NTO `dl_iterate_phdr` to toke `* mut`

nyurik:
All other platforms use `* mut`, and while this is technically a breaking change, most likely noone is using it directly.

See also https://github.com/rust-lang/backtrace-rs/pull/648

r? @joshtriplett


joshtriplett:
:+1: from me as soon as we sort out CI.

tgross35:
Going to close this since it was pulled into https://github.com/rust-lang/libc/pull/3815. We will target 0.2 with a cherry pick after merge.

PR #3794, opened 2024-08-06
Remove tier 3 targets from CI

joshtriplett:
These targets seem to regularly break unrelated PRs, and in general we
shouldn't gate PRs on tier 3 targets (per the target tier policy).

### DIFF
--- a/.github/workflows/full_ci.yml
+++ b/.github/workflows/full_ci.yml
@@ -140,45 +140,6 @@ jobs:
       - name: Execute run-docker.sh
         run: LIBC_CI=1 sh ./ci/run-docker.sh ${{ matrix.target }}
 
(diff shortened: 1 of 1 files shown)


workingjubilee:
@joshtriplett This doesn't seem to remove *all* tier 3 targets, can we identify this PR as singling out the Switch and μClibc targets?

joshtriplett:
> @joshtriplett This doesn't seem to remove _all_ tier 3 targets, can we identify this PR as singling out the Switch and μClibc targets?

I had intended for it to, but clearly I've missed some that weren't labeled as such.

workingjubilee:
Maybe I'm mistaken?

tgross35:
This was merged via https://github.com/rust-lang/libc/pull/3797.

PR #3795, opened 2024-08-06
Combined CI fixes

tgross35:
Combine the following to see what else is keeping us from having a green CI:

- https://github.com/rust-lang/libc/pull/3578
- https://github.com/rust-lang/libc/pull/3794


tgross35:
We will merge https://github.com/rust-lang/libc/pull/3796 instead

PR #3798, opened 2024-08-10
Linux: add a new syscall `SYS_mseal`

devnexen:


tgross35:
Introduction: https://lore.kernel.org/lkml/20240415163527.626541-1-jeffxu@chromium.org/

Any reason to be x86-only here? This is present on quite a few architectures https://github.com/search?q=repo%3Atorvalds%2Flinux+%2F462%5Cs%2Bcommon%5Cs%2Bmseal%5Cs%2Bsys_mseal%2F&type=code. 

We apparently haven't been very diligent here with syscalls, but `libc-test/semver` should probably still get updated.

tgross35 (approved):

PR #3800, opened 2024-08-12
freebsd moving the kinfo_file type to general use.

devnexen:
but keeping the constant KINFO_FILE_SIZE for intel archs only.


tgross35:
Thanks, do you have a link to docs/headers for `KINFO_FILE_SIZE` to confirm?

This should have an update in `libc-test/semver`.

devnexen:
I m confused. It s stable nominated but it s targeting 0.2. should this is be the one being closed ?

tgross35:
Ah sorry about that, yes I did the wrong one. Reopened https://github.com/rust-lang/libc/pull/3801

PR #3803, opened 2024-08-12
vxWorks adding few errnoLib related constants.

devnexen:


tgross35:
I am going to prefer cherry picks to update the 0.2 branch rather than creating a duplicate PR, so things get crosslinked correctly. Docs are in the process of being updated here https://github.com/rust-lang/libc/pull/3804.

Opened https://github.com/rust-lang/libc/pull/3805 with a cherry-pick of the relevant commit, I will close this and track that instead.

PR #3540, opened 2024-01-12
[0.2] Add additional Linux AF_PACKET options

nathaniel-bennett:
Added socket options, structs and other miscellaneous constants that make memory-mapped packet sockets possible in Linux--PACKET_TX_RING, PACKET_RX_RING, etc.

(adding this to libc 0.2 as 0.3 is in early development)

PR #3837, opened 2024-08-16
Remove `Debug` impls from tpacket_* structs

nathaniel-bennett:
Removes `Debug` implementation from `tpacket_req_u`, `tpacket_bd_header_u` and `tpacket_block_desc` to avoid potential uninitialized memory access (noted in #3540).

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -1183,15 +1183,18 @@ s_no_extra_traits! {
         pub sched_period: ::__u64,
     }
 
+    #[allow(missing_debug_implementations)]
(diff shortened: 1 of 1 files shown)


tgross35:
Thanks, I will merge this as-is to `main`. For the backport I'm just going to change the `Debug` impls to print `...` or something without accessing the fields, so we don't actually break any API that relies on this derive.

nathaniel-bennett:
The backport is actually still pending merge (see #3540); I've gone ahead and removed the `Debug` impls there already :thumbsup:

tgross35:
Oh thanks, I'm getting getting some of the ongoing changes confused and thinking this was existing.

PR #3840, opened 2024-08-16
TEEOS: Fix octal notation for O_* constants

### LINKED ISSUE #3532: Possibly wrong constant values on any(target_os = "teeos", all(target_os = "linux", target_env = "uclibc", target_arch = "x86_64"))
Greetings,
I may have found a small set of constants whose values were included in [libc](https://github.com/rust-lang/libc) as decimal literals while in fact they were written as octal literals in C. I believe this is the case because in C an integer literal that begins with a '0' is considered an octal literal, whereas in Rust the leading '0' is silently discarded.

I first noticed this while inspecting the value of `O_NONBLOCK` of another platform for a project of mine, and I noticed it had [this declaration on Linux uClibc x86_64](https://github.com/rust-lang/libc/blob/1abc2e47107e03eab1757b8dcb24b1095f6aeeb9/src/unix/linux_like/linux/uclibc/x86_64/mod.rs#L309), which was inconsistent with all the other related values. This convinced me to investigate further, and I ran a RegExp against libc's codebase, which resulted in a small subset of constants (most are related to `fcntl.h`):
image

Before filing this issue, in order to be as sure and detailed as possible, I also went and gave a read to [uClibc's Linux `fcntl.h` file](https://github.com/gittup/uClibc/blob/9dbf00be840a15a52656a039f83d1997344ce507/libc/sysdeps/linux/x86_64/bits/fcntl.h), which uses in fact octal literals to declare the affected constants' values.

## TL;DR

The constants listed above were likely copy-pasted from their respective headers and the subtle differences in integer literal rules from Rust and C lead to "wrong" values sneaking into the library...
unless I missed something and `rustc` is able to correctly interpret those literals as octals, which I unfortunately haven't been able to test on a Linux machine at the time of writing.

nathaniel-bennett:
Resolves #3532


tgross35:
Thanks, this looks fine to me but do you have a link to the teeos headers to check against?

nathaniel-bennett:
`teeos` directly pulls in musl libc headers as its source of constants, based on what the platform support page says (https://doc.rust-lang.org/nightly/rustc/platform-support/aarch64-unknown-teeos.html) and what its builld script does (https://gitee.com/openharmony-sig/tee_tee_os_kernel/blob/master/build/build_tee.sh#L60). It uses the `O_*` constants from musl on the kernel side of things (https://gitee.com/openharmony-sig/tee_tee_os_kernel/blob/master/user/system-services/system-servers/fs_base/fs_wrapper_ops.c#L151). Based on the musl implementation, these indeed use the C octal variant of constant values:

https://github.com/bminor/musl/blob/dd1e63c3638d5f9afb857fccf6ce1415ca5f1b8b/arch/arm/bits/fcntl.h
```c
#define O_CREAT        0100
#define O_EXCL         0200
#define O_NOCTTY       0400
#define O_TRUNC       01000
#define O_APPEND      02000
#define O_NONBLOCK    04000
#define O_DSYNC      010000
#define O_SYNC     04010000
#define O_RSYNC    04010000
#define O_DIRECTORY  040000
#define O_NOFOLLOW  0100000
#define O_CLOEXEC  02000000
```

tgross35:
Thanks! Just trying to get the sources linked in all the PRs since tracing some of these APIs back can be painful.

PR #3842, opened 2024-08-16
Add RUSTC_WRAPPER support to build script

### LINKED ISSUE #3772: Build script does not respect RUSTC_WRAPPER
This came up in https://github.com/rust-lang/rust/pull/127682: the build scripts [invokes RUSTC](https://github.com/rust-lang/libc/blob/0e28c864c25d2e9b0ab082947445efccef213da4/build.rs#L123) without wrapping it in RUSTC_WRAPPER. Bypassing wrappers isn't great, so the build script should be fixed. See https://github.com/cuviper/autocfg/pull/56, https://github.com/djc/rustc-version-rs/pull/46 for similar changes in other crates, and https://github.com/rust-lang/cargo/pull/13659 for the change in cargo.

nathaniel-bennett:
Resolves #3772

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -120,8 +120,16 @@ fn rustc_minor_nightly() -> (u32, bool) {
         };
     }
 
-    let rustc = otry!(env::var_os("RUSTC"));
-    let output = Command::new(rustc)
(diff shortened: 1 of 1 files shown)


tgross35 (approved):
Thanks!

PR #3845, opened 2024-08-16
[0.2] Add RUSTC_WRAPPER support to build script

nathaniel-bennett:
(backport <https://github.com/rust-lang/libc/pull/3842>)
(cherry picked from commit 8dba4a03d23ff979eba11afac4ff0b605267cddb)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -196,8 +196,17 @@ fn rustc_minor_nightly() -> (u32, bool) {
         };
     }
 
-    let rustc = otry!(env::var_os("RUSTC"));
-    let output = Command::new(rustc)
+    let rustc = env::var_os("RUSTC").expect("Failed to get rustc version: missing RUSTC env");
+    let mut cmd = match env::var_os("RUSTC_WRAPPER").as_ref() {
+        Some(wrapper) if !wrapper.is_empty() => {
+            let mut cmd = Command::new(wrapper);
+            cmd.arg(rustc);
+            cmd
+        }
+        _ => Command::new(rustc),
+    };
+
+    let output = cmd
         .arg("--version")
         .output()
         .ok()


tgross35:
This looks fine, but any reason it doesn't use the exact same pattern as https://github.com/rust-lang/libc/pull/3842?`if let` rather than `match`

nathaniel-bennett:
While backporting to 0.2, the Windows with rust 1.19 checks failed due to the `filter()` API not being stable yet for that version of rust. Using `match` rather than `let... else` with `filter()` fixed that.

tgross35:
Makes sense, thanks for the clarification.

nlflint:
This MR broke compilation of libc when using `clippy-driver` as my RUSTC_WRAPPER. I think the root cause is clippy-driver not this MR though. The clippy-driver doesn't "wrap" rustc properly, in that its output from `--version` isn't the same as `rustc`, so it can't be parsed.

For anyone else having this problem, I am currently able to workaround it by setting using RUSTC_WORKSPACE_WRAPPER, instead of RUSTC_WRAPPER. However, this brings the question to mind, by bypassing this, am I missing out on something?

dtolnay:
Libc will need to use `$RUSTC_WRAPPER $RUSTC --rustc --version` when `$RUSTC_WRAPPER $RUSTC --version` produces output that begins with "clippy". I've need logic to handle the same thing in `rustversion` for a long time:
<https://github.com/dtolnay/rustversion/blob/1.0.17/build/build.rs>
<https://github.com/dtolnay/rustversion/blob/1.0.17/build/rustc.rs>

tgross35:
Good point, it feels like we had this same discussion somewhere else very recently. @nathaniel-bennett would you be able to put up a fix?

PR #3846, opened 2024-08-17
Apple: Add additional `pthread` APIs

### LINKED ISSUE #1618: Add posix threads support for macOS
Can we add a possibility to use POSIX threads for the macOS? It seems to be not straightforward to do myself - all the code is mixed across many files. Thanks!

P.S. Maybe it is even better if we share all the POSIX-code among several targets which support it. Right now, for example, the `pthread_setschedprio` function is defined in the `/linux-like/linux` but it shouldn't be Linux-only available.

nathaniel-bennett:
Resolves #1618

It seems like #1618 has already been mostly resolved; this PR simply adds some of the few `pthread` APIs that are still missing.


src/unix/bsd/apple/b32/mod.rs
```diff
@@ -70,7 +75,7 @@ cfg_if! {
         impl Eq for pthread_attr_t {}
         impl ::fmt::Debug for pthread_attr_t {
             fn fmt(&self, f: &mut ::fmt::Formatter) -> ::fmt::Result {
-                f.debug_struct("pthread_attr_t")
+                f.debug_struct("pthread_once_t")
```
tgross35, line 78: I think this might not have been intentional

tgross35:
Sources for future reference:

- Type `pthread_once_t` is defined as `_opaque_pthread_once_t` here https://github.com/apple-oss-distributions/libpthread/blob/d8c4e3c212553d3e0f5d76bb7d45a8acd61302dc/include/sys/_pthread/_pthread_types.h#L88-L91.
- Opaque padding `__PTHREAD_ONCE_SIZE__` is defined to 4/8 here https://github.com/apple-oss-distributions/libpthread/blob/d8c4e3c212553d3e0f5d76bb7d45a8acd61302dc/include/sys/_pthread/_pthread_types.h#L42-L52
- Constants are defined here https://github.com/apple-oss-distributions/libpthread/blob/d8c4e3c212553d3e0f5d76bb7d45a8acd61302dc/include/pthread/pthread.h#L138
- `_PTHREAD_ONCE_SIG_init` is defined here https://github.com/apple-oss-distributions/libpthread/blob/d8c4e3c212553d3e0f5d76bb7d45a8acd61302dc/include/pthread/pthread_impl.h#L48C33-L48C43
- `pthread_once` https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/pthread_once.3.html
- The remaining functions are here, some already exist or are in `unix/mod.rs` https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/pthread_attr_getinheritsched.3.html

One fix above, then could you define `__PTHREAD_ONCE_SIZE__` and make use of it in `__opaque` (rather than literals) like the definition? We already have e.g. `__PTHREAD_RWLOCKATTR_SIZE__` defined, just unfortunately don't use them consistently.

src/unix/bsd/apple/b32/mod.rs
nathaniel-bennett: Fixed--thanks for catching this!

PR #3847, opened 2024-08-17
Linux: add `getitimer()`/`setitimer()`

### LINKED ISSUE #1347: getitimer()/setitimer() are missing on Linux
```c
#include <sys/time.h>

int getitimer(int which, struct itimerval *value);
int setitimer(int which, const struct itimerval *restrict value, struct itimerval *restrict ovalue);
```

[man](http://man7.org/linux/man-pages/man2/getitimer.2.html)

This is a minimal demo:

```rust
let tval = nix::libc::itimerval {
    it_interval: nix::libc::timeval {
        tv_sec: 0,
        tv_usec: 0,
    },
    it_value: nix::libc::timeval {
        tv_sec: 1,
        tv_usec: 0,
    },
};
if unsafe { setitimer(nix::libc::ITIMER_REAL, &tval as *const _, std::ptr::null_mut()) } == -1 {
    panic!("setitimer");
}

let tval2 = nix::libc::itimerval {
    it_interval: nix::libc::timeval {
        tv_sec: 0,
        tv_usec: 0,
    },
    it_value: nix::libc::timeval {
        tv_sec: 0,
        tv_usec: 0,
    },
};
if unsafe { getitimer(nix::libc::ITIMER_REAL, &mut tval as *mut _) } == -1 {
    panic!("getitimer");
}
```

I have found these two commits: https://github.com/rust-lang/libc/commit/e976c370f840f1381f4a9bd5ad4f825af8259e5b, https://github.com/rust-lang/libc/commit/bc48202e14cb5112de5199f1ebc8b8464637bcc1 -- `getitimer` and `setitimer` we introduced and dropped for netbsd target. Those definitions seem to work fine for me on Linux.

nathaniel-bennett:
Resolves  #1347


nathaniel-bennett:
Note that this PR changes `ITIMER_REAL`, `ITIMER_VIRTUAL` and `ITIMER_PROF` from consts into variants of a named enum (which is their correct definition in glibc [1]), so this potentially breaking change shouldn't be backported to 0.2.

[1] https://github.com/bminor/glibc/blob/e7c14e542d8d858b824b5df4f4e3dc93695e6171/time/sys/time.h#L114

src/unix/linux_like/linux/mod.rs
```diff
@@ -6129,6 +6144,28 @@ extern "C" {
     pub fn ioctl(fd: ::c_int, request: ::Ioctl, ...) -> ::c_int;
 }
 
+cfg_if! {
+    if #[cfg(target_env = "musl")] {
+        extern "C" {
+            pub fn getitimer(which: ::c_int, value: *mut ::itimerval) -> ::c_int;
```
devnexen, line 6150: not sure I follow, why not using the typedef you created above ?

nathaniel-bennett: This is the result of a strange combination of issues. glibc and most other libc implementations typedef `__itimer_which_t` as an `enum` when `_GNU_SOURCE` is set, or as an `int` otherwise. `musl` unfortunately opts to just use an `int` directly within the function definitions for `getitimer` and `setitimer`. As a result of this:

- Using `__itimer_whiich_t` directly in function definitions causes `musl` CI to fail as it's not a defined type
- Removing `__itimer_which_t` altogether for `musl` builds still causes CI to fail, as there is currently no way of opting out of `enum` semver checks conditionally for `musl` (you can opt out for `struct`s, `type`s, everything else but `enum`s)

By defining `__itimer_which_t` as a `type` definition in `musl` and then opting out of it, CI passes.

A more clean solution to this may be to develop `enum` opt-out functions to the `ctest2` crate.

devnexen: Another possible approach is not to bother with enums at all.

```rust
pub type __itimer_which_t = ::c_int;
...
pub const ITIMER_REAL: __itimer_which_t = 0;
...
```

There are various places in this crate were C enums are handled as simple constants. Should pass in theory.

src/unix/linux_like/linux/mod.rs
```diff
@@ -73,6 +73,8 @@ missing! {
     pub enum fpos64_t {} // FIXME: fill this out with a struct
 }
 
+pub type __itimer_which_t = ::c_int;
```
devnexen, line 76: might work better

```rust
cfg_if! {
    if #[cfg(target_env = "musl")] {
        pub type __itimer_which_t = ::c_int;
    } else {
        pub type __itimer_which_t = ::c_uint;
    }
}

```

nathaniel-bennett: Looks like it works--thanks!

src/unix/linux_like/linux/mod.rs
```diff
@@ -73,6 +73,14 @@ missing! {
     pub enum fpos64_t {} // FIXME: fill this out with a struct
 }
 
+cfg_if! {
+    if #[cfg(target_env = "musl")] {
+        pub type __itimer_which_t = ::c_int;
+    } else {
+        pub type __itimer_which_t = ::c_uint;
+    }
+}
```
tgross35, line 82: glibc has `int` when it is not defined as an enum, and musl doesn't seem to define this type at all - why is the definition done this way?

tgross35: Ah nevermind, I see the reasoning below. I think we need to work with one of the ways that doesn't expose this on musl.

In general it is probably better to avoid Rust `enum`s to represent C types at all, given how easy it is to cause UB. So some sort of solution that just typedefs an int on glibc seems preferable to me.

PR #3854, opened 2024-08-19
WASI: fix missing `Iterator` with `rustc-dep-of-std`

youknowone:
https://github.com/rust-lang/libc/pull/3681#issuecomment-2295428691

@tgross35 How can I correctly run local test of this build? I tried `cargo build --target wasm32-wasip1 --features rustc-dep-of-std --no-default-features` but it causes multiple patch-unrelated errors

```
❯ cargo build --target wasm32-wasip1 --features rustc-dep-of-std --no-default-features
error[E0432]: unresolved import `core::clone`
  --> src/lib.rs:46:11
   |
46 | use core::clone::Clone;
   |           ^^^^^ could not find `clone` in `core`

error[E0432]: unresolved import `core::ffi`
  --> src/lib.rs:48:5
   |
48 | use core::ffi;
   |     ^^^^^^^^^ no `ffi` in the root
```

### DIFF
--- a/src/wasi.rs
+++ b/src/wasi.rs
@@ -2,6 +2,7 @@
 // `wasi-libc` project provides multiple libraries including emulated features, but we list only basic features with `libc.a` here.
 
 use super::{Send, Sync};
+use core::iter::Iterator;
 
 pub use ffi::c_void;
 


tgross35:
Thanks for the quick fix! I think you can test by patching this line https://github.com/rust-lang/libc/blob/53752600b2c5038ab3355a369fae35426e08d77b/src/lib.rs#L34 to be `extern crate core as core;`. If that doesn't work then I wouldn't worry about testing with that feature. Please make sure `wasi` still builds locally without that feature, since we don't have it in CI (we should probably add it at some point)

youknowone:
Thank you! I tested both with and without the feature and they look good. It doesn't complain about unused import.

tgross35:
Great, thanks!

PR #3855, opened 2024-08-19
fix clippy warnings

youknowone:


youknowone:
Yes, there are. I didnt fix edition related stuffs because I had no idea which edition is intended for libc

tgross35:
That is a good question that I have been meaning to bring up. I think we will target e2021 on `main`, it just hasn't been discussed anywhere yet. So good to leave those lints as-is until we have a policy.

tgross35:
I was going to backport this but almost all of this doesn't work on our oldest version, so I guess there isn't much need.

PR #3766, opened 2024-07-06
backport: add stabilized HermitOS interface

stlankes:
HermitOS is a unikernel and its interface to the kernel is provided by https://crates.io/crates/hermit-abi. In the meantime parts of the interface is stabilized and we want to integrated it into libc. Unstable version will be still provided by hermit-abi.

This PR is a backport of #3724 and doesn't have breaking changes. It just provides new functions to enter the kernel.

PR #3858, opened 2024-08-19
add missing error numbers for HermitOS

stlankes:
HermitOS is a unikernel and its interface to the kernel is provided by https://crates.io/crates/hermit-abi. In the meantime parts of the interface is stabilized and we want to integrated it into libc. Unstable version will be still provided by hermit-abi.

Error numbers are missing in the `main` branch.


stlankes:
See also rust-lang/libc#3766

tgross35:
The numbers all looks correct to me. However, the linked source at https://github.com/hermit-os/hermit-rs/blob/5a4d7efe1ab5727cf4d5299943aa8deaa098a603/hermit-abi/src/errno.rs looks like it always uses `i32`, not `c_int` that is used in this PR. Probably better to just use `i32` here as well?

Also, could you start a file `libc-test/semver/hermit.txt` and add these? This won't be checked in CI but just run the tests locally and make sure they pass.

stlankes:
`c_int` is defined as i32 (https://github.com/rust-lang/libc/blob/main/src/hermit.rs#L15). Hm, it irritates a little. We thought that this easier to understand because all other OS are using `c_int`. We can change it, I am open for it.

> Also, could you start a file libc-test/semver/hermit.txt and add these? This won't be checked in CI but just run the tests locally and make sure they pass.

I will do it

tgross35:
> `c_int` is defined as i32 (https://github.com/rust-lang/libc/blob/main/src/hermit.rs#L15). Hm, it irritates a little. We thought that this easier to understand because all other OS are using `c_int`. We can change it, I am open for it.

Regardless of how the type aliases are defined, prefer being consistent with the relevant headers / source files over being consistent with other platforms. Doesn't really matter here since `c_int` is almost always `i32`, but it makes it easier to verify that we don't quietly introduce bugs.

Looks like this is already done elsewhere, e.g. `SO_REUSEADDR` is a `c_int` on most platforms but an `i32` on Hermit.

stlankes:
Ik, I thought about it. I changed it to i32.

tgross35:
Thanks! Can you add `libc-test/semver/hermit.txt`?

stlankes:
I just added `libc-test/semver/hermit.txt`.

tgross35:
Thanks!

tgross35 (approved):

PR #3860, opened 2024-08-19
backport: add missing error numbers for HermitOS

stlankes:
HermitOS is a unikernel and its interface to the kernel is provided by https://crates.io/crates/hermit-abi. In the meantime parts of the interface is stabilized and we want to integrated it into libc. Unstable version will be still provided by hermit-abi.

Error numbers are missing in the `libc-0.2` branch.


tgross35:
No worries 😆 I'll try to get to it today

PR #3863, opened 2024-08-21
Revise network definitions for HorizonOS

Meziu:
@ian-h-chamberlain @FenrirWolf @adryzz

Following some discussion in https://github.com/rust3ds/ctru-rs/issues/174.

The changes include:
- Update to the `hostent` definition for HorizonOS (`armv6k-nintendo-3ds`) following upstream [changes](https://github.com/devkitPro/libctru/commit/6714c04806c1164eef0718191060ee7a38e64d77).

- A step back on the decision to remove the trailing zero bytes from `sockaddr_in` that was made all the way back in https://github.com/rust-lang/libc/pull/2725. This change will require a little work in `std`, but we came to understand it's only right to keep the definitions true to the upstream versions.

Also, should this PR be flagged for release in 0.2.x as well as the 1.0 milestone? Work on tier-3 targets isn't clearly talked about in the new contribution guidelines (not that this PR is of any relevance to the release process).

### DIFF
--- a/src/unix/newlib/mod.rs
+++ b/src/unix/newlib/mod.rs
@@ -53,6 +53,21 @@ cfg_if! {
     }
 }
 
+cfg_if! {
+    if #[cfg(not(target_os = "horizon"))] {
+        s!{
+            pub struct hostent {
+                pub h_name: *mut ::c_char,
+                pub h_aliases: *mut *mut ::c_char,
+                pub h_addrtype: ::c_int,
+                pub h_length: ::c_int,
+                pub h_addr_list: *mut *mut ::c_char,
+                pub h_addr: *mut ::c_char,
+            }
+        }
+    }
+}
+
 s! {
     // The order of the `ai_addr` field in this struct is crucial
     // for converting between the Rust and C types.
@@ -87,16 +102,7 @@ s! {
     }
 
     pub struct in_addr {
-            pub s_addr: ::in_addr_t,
-    }
-
-    pub struct hostent {
-            pub h_name: *mut ::c_char,
-            pub h_aliases: *mut *mut ::c_char,
-            pub h_addrtype: ::c_int,
-            pub h_length: ::c_int,
-            pub h_addr_list: *mut *mut ::c_char,
-            pub h_addr: *mut ::c_char,
+        pub s_addr: ::in_addr_t,
     }
 
     pub struct pollfd {
--- a/src/unix/newlib/horizon/mod.rs
+++ b/src/unix/newlib/horizon/mod.rs
@@ -21,6 +21,14 @@ pub type sbintime_t = ::c_longlong;
 pub type sigset_t = ::c_ulong;
 
 s! {
+    pub struct hostent {
+        pub h_name: *mut ::c_char,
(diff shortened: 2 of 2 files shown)


tgross35:
Thanks for the fix! Looks like this is defined here https://libctru.devkitpro.org/in_8h_source.html. Can you verify that tests pass with `./ci/run.sh relevant-target` when checking Horizon?

> Also, should this PR be flagged for release in 0.2.x as well as the 1.0 milestone? Work on tier-3 targets isn't clearly talked about in the new contribution guidelines (not that this PR is of any relevance to the release process).

You are welcome to stable nominate anything (the label works via rustbot). I don't mind backporting as long as it's not overly conflicty.

src/unix/newlib/mod.rs
```diff
@@ -87,16 +87,27 @@ s! {
...
+
+        #[cfg(not(target_os = "horizon"))]
+        pub h_length: ::c_int,
+        #[cfg(target_os = "horizon")]
+        pub h_length: u16,
+
+        pub h_addr_list: *mut *mut ::c_char,
+
+        #[cfg(not(target_os = "horizon"))]
+        pub h_addr: *mut ::c_char,
```
tgross35, line 110: Could you instead put this struct in a `cfg_if!` with the current definition under `if #[cfg(not(target_os = "horizon"))] { ... }`, then put the new definition in `horizon/mod.rs`? It's a bit tough to read this way.

tgross35:
Also, do you have a link to the relevant headers where `hostent` is defined? Trying to get these crosslinked in all PRs.

@rustbot author

FenrirWolf:
`hostent` is in [netdb.h](https://libctru.devkitpro.org/netdb_8h_source.html)

Meziu:
> Thanks for the fix! Looks like this is defined here https://libctru.devkitpro.org/in_8h_source.html. Can you verify that tests pass with `./ci/run.sh relevant-target` when checking Horizon?

The test (rightfully) panics at [this line](https://github.com/rust-lang/libc/blob/cdf68968a2e40a93027000fb07688ea54f021e2a/libc-test/build.rs#L65) because our target has no specified tests. I will add that our target isn't a host target (it's basically an embedded device) and that cross-compiling and running tests is a problem we've had for quite a long time due to the very weird setup with the required toolchain libraries.

Is it necessary to implement these tests? I *might* be able to cook up something that checks out the headers (assuming the host has all the libraries installed), but it would serve little purpose as the toolchain used is (and will always be) fundamentally experimental.

src/unix/newlib/mod.rs
Meziu: I thought not to do that originally because there is no way to use `cfg_if!` within `s!`, but I just separated the struct from the rest and made a new definition in `horizon/mod.rs`.

tgross35 (approved):

tgross35:
Thanks for the change, looks good now! Could you please squash?

Regarding the tests, no need to do anything here. At some point in the future you are welcome to add a similar function, but it doesn't need to happen now.

ian-h-chamberlain (approved):

PR #3864, opened 2024-08-21
Fix tests on Solaris

psumbera:
Tested (libc-test) on both Solaris and Illumos (with latest stable Omnios). All tests are passing.

Illumos changes can be checked against https://github.com/illumos/illumos-gate
Solaris changes can be checked only with installed Solaris CBE release (https://blogs.oracle.com/solaris/post/announcing-the-first-oracle-solaris-114-cbe) or using Solaris 11.4 machine at Cfarm (https://portal.cfarm.net/machines/list/).


src/unix/solarish/mod.rs
```diff
@@ -489,25 +510,35 @@ s! {
...
         pub ut_session: ::c_int,
+        #[cfg(target_os = "illumos")]
         pub ut_pad: [::c_int; _UTX_PADSIZE],
+        #[cfg(target_os = "solaris")]
+        pub pad: [::c_int; 5],
         pub ut_syslen: ::c_short,
+        #[cfg(target_os = "illumos")]
         pub ut_host: [::c_char; _UTX_HOSTSIZE],
+        #[cfg(target_os = "solaris")]
+        pub ut_host: [::c_char; 257],
```
pfmooney, line 541: Rather than this `cfg()` soup (here and for some of the other modified structs), it might make more sense to move to separate definitions in `illumos.rs` and `solaris.rs`

src/unix/solarish/mod.rs
```diff
@@ -2154,6 +2201,9 @@ pub const PTHREAD_RWLOCK_INITIALIZER: pthread_rwlock_t = pthread_rwlock_t {
 pub const PTHREAD_MUTEX_NORMAL: ::c_int = 0;
 pub const PTHREAD_MUTEX_ERRORCHECK: ::c_int = 2;
 pub const PTHREAD_MUTEX_RECURSIVE: ::c_int = 4;
+#[cfg(target_os = "illumos")]
+pub const PTHREAD_MUTEX_DEFAULT: ::c_int = 8;
```
pfmooney, line 2205: Until we (illumos) make a decision about what the minimum support illumos version is, this change to PTHREAD_MUTEX_DEFAULT cannot be made

src/unix/solarish/mod.rs
```diff
@@ -2182,11 +2232,22 @@ pub const PORT_SOURCE_MQ: ::c_int = 6;
...
 pub const _UTX_HOSTSIZE: usize = 257;
+    }
+}
+cfg_if! {
+    if #[cfg(target_os = "solaris")] {
+pub const _UTMP_USER_LEN: usize = 32;
+pub const _UTMP_LINE_LEN: usize = 32;
+pub const _UTMP_ID_LEN: usize = 4;
+    }
+}
```
pfmooney, line 2250: Here and elsewhere: Move to the OS specific file, rather than a bunch of cfg_if stuff? (At least when it's huge blocks of stuff, vs one-off things in a largely shared list)

psumbera:
@tgross can you please help me with the buildbot failure? I have no idea what is wrong and how can I fixt it...

tgross:
Hi @psumbera! I think you meant @tgross35 (which is funny because @pfmooney and @jclulow who are at-mentioned here are old colleagues!) :grinning: 

Not really my area of expertise at all but from context it looks like the `epoll_event` might be missing an implementation of the `Debug` trait?

src/unix/mod.rs
```diff
@@ -1423,12 +1424,20 @@ cfg_if! {
...
-                     target_os = "nto")))] {
+                     target_os = "nto",
+                     target_os = "solaris")))] {
         extern "C" {
             pub fn adjtime(delta: *const timeval, olddelta: *mut timeval) -> ::c_int;
         }
     }
 }
+cfg_if! {
+    if #[cfg(target_os = "solaris")] {
```
tgross35, line 1435:
```suggestion
    } else if #[cfg(target_os = "solaris")] {
```

I think this could be combined with the existing `cfg_if`

src/unix/solarish/solaris.rs
```diff
@@ -48,13 +67,96 @@ s_no_extra_traits! {
...
+        pub ut_session: ::c_int,
+        pub pad: [::c_int; 5],
+        pub ut_syslen: ::c_short,
+        pub ut_host: [::c_char; 257],
+    }
+
+    pub struct xrs_t {
+        pub xrs_id: ::c_ulong,
+        pub xrs_ptr: *mut ::c_char,
+    }
```
tgross35, line 88: I think this may be the source of the error

src/unix/solarish/x86_common.rs
```diff
@@ -63,3 +65,5 @@ pub const AT_386_FPINFO_NONE: u32 = 0;
 pub const AT_386_FPINFO_FXSAVE: u32 = 1;
 pub const AT_386_FPINFO_XSAVE: u32 = 2;
 pub const AT_386_FPINFO_XSAVE_AMD: u32 = 3;
+    }
+}
```
tgross35, line 69: Can you fix the indentation?

src/unix/solarish/x86_64.rs
```diff
@@ -1,3 +1,9 @@
+cfg_if! {
+    if #[cfg(target_os = "solaris")] {
+        use xrs_t;
+    }
+}
+
```
tgross35, line 6: Could you just use the full path to `xrs_t` where it is used, rather than having a single `cfg`ed import here?

tgross35:
> @tgross can you please help me with the buildbot failure? I have no idea what is wrong and how can I fixt it...

I left a comment above but I think `xrs_t` may be missing implementations. If it can just go in `s!` to get the automatic derives, rather than `s_no_extra_traits!`, that is probably easiest. (I'm not actually sure why any of these types are in `s_no_extra_traits!` when their implementations look near identical to the derives, some of them could move to `s!`. We have a big cleanup of these traits coming in the near future.)

Are tests currently failing on these platforms? It would be great to add `x86_64-unknown-illumos` to CI since it is tier 2 with host tools.

tgross35:
> Hi @psumbera! I think you meant @tgross35 (which is funny because @pfmooney and @jclulow who are at-mentioned here are old colleagues!) 😀

Out of all the similar GitHub usernames 😄 that really is an unbelievable coincidence

> Not really my area of expertise at all but from context it looks like the `epoll_event` might be missing an implementation of the `Debug` trait?

Thanks for taking a look :) I think it's a different struct though, this crate is unfortunately pretty macro-heavy so the error messages are a bit obscure.

src/unix/solarish/x86_64.rs
psumbera: Well, not sure what is proper fix. Now I use:
```
cfg_if! {
    if #[cfg(target_os = "solaris")] {
        use unix::solarish::solaris;
    }
}
..
pub uc_xrs: solaris::xrs_t,
```

psumbera:
> > Hi @psumbera! I think you meant @tgross35 (which is funny because @pfmooney and @jclulow who are at-mentioned here are old colleagues!) 😀
> 
> Out of all the similar GitHub usernames 😄 that really is an unbelievable coincidence

Sorry about that!

psumbera:
Now it fails somewhere else and I don't know how to proceed again...

tgross35:
Sorry, slipped under my radar. I think the error is just saying that `door_desc_t__d_data__d_desc` needs to move to `s_no_extra_traits!` rather than `s!`, before a certain version somewhere in 1.6x you can't derive `Debug` on packed traits (I suspect you are just testing with a more recent rustc version locally).

tgross35:
I think this looks okay but I am not able to test on the relevant systems. @pfmooney would you be able to double check this?

tgross35 (approved):
Sorry I didn't merge this already, I think it should be good with a rebase.

tgross35:
Hm, not sure why freebsd is failing but it seems to be happening on everything

psumbera:
> Hm, not sure why freebsd is failing but it seems to be happening on everything

Have no idea. Doesn't seem to be related.

src/unix/solarish/illumos.rs
```diff
@@ -58,12 +184,68 @@ pub const SOL_FILTER: ::c_int = 0xfffc;
 
 pub const MADV_PURGE: ::c_int = 9;
 
-pub const POSIX_FADV_NORMAL: ::c_int = 0;
-pub const POSIX_FADV_RANDOM: ::c_int = 1;
-pub const POSIX_FADV_SEQUENTIAL: ::c_int = 2;
-pub const POSIX_FADV_WILLNEED: ::c_int = 3;
-pub const POSIX_FADV_DONTNEED: ::c_int = 4;
-pub const POSIX_FADV_NOREUSE: ::c_int = 5;
```
sunshowers, line 66: It looks like these, plus POLLRDHUP above, were accidentally deleted maybe in a bad merge. I'll roll a fix in for these in #3999.

psumbera: Sorry about that.

PR #3867, opened 2024-08-24
adding new illumos ptsname_r call.

devnexen:
while at it, fixing PTHREAD_MUTEX_DEFAULT which differs from solaris.

[man page](https://github.com/illumos/illumos-gate/blob/f1131be9159d3bbfbff616a59d26504ef672b1f2/usr/src/man/man3c/ptsname.3c#L100)

### DIFF
--- a/src/unix/solarish/illumos.rs
+++ b/src/unix/solarish/illumos.rs
@@ -100,4 +100,6 @@ extern "C" {
     pub fn pwritev(fd: ::c_int, iov: *const ::iovec, iovcnt: ::c_int, offset: ::off_t)
         -> ::ssize_t;
     pub fn getpagesizes2(pagesize: *mut ::size_t, nelem: ::c_int) -> ::c_int;
+
+    pub fn ptsname_r(fildes: ::c_int, name: *mut ::c_char, namelen: ::size_t) -> ::c_int;
 }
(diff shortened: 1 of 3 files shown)


src/unix/solarish/illumos.rs
```diff
@@ -68,6 +68,8 @@ pub const B4000000: ::speed_t = 31;
 // sys/systeminfo.h
 pub const SI_ADDRESS_WIDTH: ::c_int = 520;
 
+pub const PTHREAD_MUTEX_DEFAULT: ::c_int = 8;
```
pfmooney, line 71: This define cannot be changed in libc so soon.  It was only changed in illumos in February 2024.  Any running systems which predate the change (which there are many) would potentially malfunction if newly-compiled rust binaries use this newer value.

devnexen: so we are ok if unit tests do not necessarily pass then ?

tgross35:
The above change was reverted as suggested, @pfmooney does this look reasonable to you?

Also @devnexen do you have links to relevant headers / manpages?

devnexen:
description updated

tgross35 (approved):
Thanks, I haven't been very successful finding illumos headers but that looks like a good source.

This lgtm but I'll give pfmooney a couple more days to double check.

tgross35:
@pfmooney any chance you could double check this with the changes?

tgross35:
This looks straightforward enough to me. I won't backport it for a few days at least, @pfmooney feel free to leave an ack/nack at any time.

PR #3869, opened 2024-08-27
Re-enable testing of WASI on CI

alexcrichton:
This commit updates CI to resume testing WASI. This updates the container and testing scripts from historical processes to more modern ones, e.g. downloading wasi-sdk instead of compiling a custom toolchain. This should make it easier to update in the future and keep it in sync with rust-lang/rust as well.

This also required a few minor fixes such as:

* The `S_IFIFO` and `S_IFMT` constants had incorrect values.
* The `CLOCK_*` definitions cause `ctest2`'s parsing to panic to they're skipped with a new `#[cfg]`.
* A new `langinfo.h` header was added to the list to include.
* Some historically skipped checks were removed since they're no longer necessary.
* Checks for `__errno_location` are disabled since that doesn't actually exist in headers.
* Checks for `select` are disabled because the Rust definition got the `const`-ness swapped for the final `timeval` argument.

It's intended that this is not a breaking change for existing users. I also wanted to ask, is there a reasonable location that I can record my name for being cc'd for any WASI-related issues? Happy to help out!


tgross35:
Yay! Thanks, everything looks reasonable to me. Do you have a link to the headers that define the changed `S_*` and `CLOCK_*` constants? Trying to get these linked in the PRs so there is always a reference when digging through the history.

> I also wanted to ask, is there a reasonable location that I can record my name for being cc'd for any WASI-related issues? Happy to help out!

Great to have you back :) There is the O-wasi label, github lets you watch specific labels in this setting (check `Issues` then it lets you select)

image

Or maybe this can be set up through triagebot too? https://github.com/rust-lang/libc/blob/72cb7aaf50bf70b5d360b1a64b0d52d3ed3fbf15/triagebot.toml you should at least set up a ping for yourself in `src/wasi` there.

alexcrichton:
For the headers https://github.com/WebAssembly/wasi-libc/blob/230d4be6c54bec93181050f9e25c87150506bdd0/libc-top-half/musl/include/sys/stat.h is the source of `S_*` and https://github.com/WebAssembly/wasi-libc/blob/230d4be6c54bec93181050f9e25c87150506bdd0/libc-bottom-half/headers/public/__header_time.h#L17-L20 is the `CLOCK_*` bits.

Also holy cow I had no idea you could subscribe to just labels, I didn't know that github added that! I've subscribed myself to O-wasi/O-wasm in this repo. I'll do the same for rust-lang/rust too.

Looks like the same setting isn't applied for PRs though (it looks to be all-or-nothing?), so if I don't appear on wasm-related PRs feel free to cc me

tgross35:
> For the headers https://github.com/WebAssembly/wasi-libc/blob/230d4be6c54bec93181050f9e25c87150506bdd0/libc-top-half/musl/include/sys/stat.h is the source of `S_*` and https://github.com/WebAssembly/wasi-libc/blob/230d4be6c54bec93181050f9e25c87150506bdd0/libc-bottom-half/headers/public/__header_time.h#L17-L20 is the `CLOCK_*` bits.

Awesome, thanks!

> Also holy cow I had no idea you could subscribe to just labels, I didn't know that github added that! I've subscribed myself to O-wasi/O-wasm in this repo. I'll do the same for rust-lang/rust too.
> 
> Looks like the same setting isn't applied for PRs though (it looks to be all-or-nothing?), so if I don't appear on wasm-related PRs feel free to cc me

Ah yeah it is pretty new and ime somewhat buggy, so don't rely on it :) better than nothing though. You should definitely make yourself a mention entry for the wasi directory once you create it for wasip2.

Issue #3789, opened 2024-08-01
libc crate triggers warnings when compiled with `default-features = false, features = ['rustc-dep-of-std']`


felixwrt:
# Summary

- Trigger: I wanted to make some changes to the `libc` used in rustc and therefore patched rustc to use `libc` from a local path. This made the compilation fail.
- Root cause:
  - building the `libc` crate for `x86_64-unknown-linux-gnu` (and probably other targets) produces warnings when using `default-features = false, features = ['rustc-dep-of-std']` (like for example the rust standard library does)
  - rustc usually pulls `libc` from crates.io and a [heuristic in Cargo](https://github.com/rust-lang/cargo/issues/14337) suppresses warnings from it in that case
  - when I changed the `libc` dependency from crates.io to a local folder, the warnings weren't suppressed anymore
  - rustc is compiled using `-D warnings` which makes the warnings in the `libc` crate fail the entire build

# Steps to reproduce

Clone Rust and libc:

```
# clone rust
git clone https://github.com/rust-lang/rust.git --depth 1
# clone libc 0.2.155 (latest released version)
git clone --branch 0.2.155 https://github.com/rust-lang/libc.git
```

Add a path override for libc in rustc's Cargo workspace:

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 178a5ab9..51b4664e 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -122,6 +122,8 @@ codegen-units = 1
 # lto = true

 [patch.crates-io]
+libc = { path = "../libc" }
+
 # See comments in `library/rustc-std-workspace-core/README.md` for what's going on
 # here
 rustc-std-workspace-core = { path = 'library/rustc-std-workspace-core' }
```

Try building the compiler (stage 0 std is enough to trigger the problem):

```
./configure
./x.py build --stage 0 library/std
```

Compilation fails with the following output:

```
...
Building stage0 library artifacts {std} (x86_64-unknown-linux-gnu)
    Updating crates.io index
    Updating crates.io index
     Locking 1 package to latest compatible version
      Adding libc v0.2.155 (/home/me/pg/warnings_issue_tmp/libc)
  Downloaded compiler_builtins v0.1.114
  Downloaded object v0.36.2
  Downloaded 2 crates (516.4 KB) in 2.27s
   Compiling cc v1.0.99
   Compiling core v0.0.0 (/home/me/pg/warnings_issue_tmp/rust/library/core)
[...]

PR #3892, opened 2024-08-30
Add wasm32-wasip2 definitions necessary for std::net support

nickrum:
This moves the bindings necessary for https://github.com/rust-lang/rust/pull/129638 to libc. Unfortunately, I wasn't able to test this with the standard library due to #3789. Is there any workaround for that issue?

Header sources:
[sa_family_t](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-bottom-half/headers/public/__typedef_sa_family_t.h)
[in_port_t](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-bottom-half/headers/public/__typedef_in_port_t.h)
[in_addr_t](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-bottom-half/headers/public/__typedef_in_addr_t.h)
[socklen_t](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-bottom-half/headers/public/__typedef_socklen_t.h)
[sockaddr](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-bottom-half/headers/public/__struct_sockaddr.h)
[in_addr](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-bottom-half/headers/public/__struct_in_addr.h)
[sockaddr_in](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-bottom-half/headers/public/__struct_sockaddr_in.h)
[in6_addr](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-bottom-half/headers/public/__struct_in6_addr.h)
[sockaddr_in6](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-bottom-half/headers/public/__struct_sockaddr_in6.h)
[sockaddr_storage](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-bottom-half/headers/public/__struct_sockaddr_storage.h)
[addrinfo](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-top-half/musl/include/netdb.h)
[ipv6_mreq](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-top-half/musl/include/netinet/in.h)
[[SHUT_RD, SHUT_WR, SHUT_RDWR, MSG_NOSIGNAL, MSG_PEEK, SO_REUSEADDR, SO_ERROR, SO_RCVTIMEO, SO_SNDTIMEO, SOCK_DGRAM, SOCK_STREAM, SOL_SOCKET, AF_INET, AF_INET6]](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-bottom-half/headers/public/__header_sys_socket.h)
[[IPPROTO_IP, IPPROTO_TCP, IPPROTO_IPV6]](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-bottom-half/headers/public/__header_netinet_in.h)
[[IP_TTL, IP_MULTICAST_TTL, IP_MULTICAST_LOOP, IP_ADD_MEMBERSHIP, IP_DROP_MEMBERSHIP, IPV6_MULTICAST_LOOP, IPV6_JOIN_GROUP, IPV6_LEAVE_GROUP, IPV6_V6ONLY, IPV6_ADD_MEMBERSHIP, IPV6_DROP_MEMBERSHIP]](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-top-half/musl/include/netinet/in.h)
[TCP_NODELAY](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-top-half/musl/include/netinet/tcp.h)
[EAI_SYSTEM](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-top-half/musl/include/netdb.h)
[[socket, connect, bind, listen, accept, getsockname, getpeername, sendto, recvfrom, getsockopt, setsockopt]](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-top-half/musl/include/sys/socket.h)
[[getaddrinfo, freeaddrinfo, gai_strerror]](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-top-half/musl/include/netdb.h)

There is one problem that I just now realized when I collected all the header sources: The `ip_mreq` struct and the `SO_BROADCAST` constant aren't actually defined in wasi-libc, but they are needed since the standard library uses them in `sys_common`. Could this be an oversight given that `ipv6_mreq` and most of the other `SO_*` constants are in fact defined? I assume this means I have to add them to wasi-libc before this PR can be merged?

cc @alexcrichton


alexcrichton:
Thanks for this @nickrum! For the missing constants/types mind filing an issue on [wasi-libc itself](https://github.com/webAssembly/wasi-libc)? I think it's ok to omit those here and handle libstd by either:

* If possible remove code that references them and add comments about how to fill it in when support is added referencing the wasi-libc issue.
* Otherwise invent dummy definitions in libstd where you already have definitions and leave a comment pointing to the wasi-libc issues as how they're not part of the public api just yet but will be soon

nickrum:
@alexcrichton Thanks for having a look! It seems like the `ip_mreq` struct is gated behind `_GNU_SOURCE`. Since the [list of predefined macros](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/expected/wasm32-wasip2/predefined-macros.txt) does contain `#define _GNU_SOURCE 1`, would you know if this does in fact mean that I can assume that the layout of the struct is stable and I can include it here?
If this is the case, then it is just `SO_BROADCAST` that is missing. I think it should be reasonable to take the value for this constant from [here](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-top-half/musl/include/sys/socket.h), which is where the values for the other `SO_*` constants are coming from. I'll file an issue on wasi-libc just for the constant 👍

alexcrichton:
I don't know a huge amount about that but I think it should be fine to include yeah. I think if you edit `libc-test/build.rs` it should already enable that for some other platforms and adding that for WASI seems reasonable too.

Thanks for filing an issue!

nickrum:
Ah, sorry! Looks like `_GNU_SOURCE` is already defined for the WASI targets. I got confused by the `cfg.define("_GNU_SOURCE", None)` syntax.

Now I'm running into a strange issue, where the tests succeed no matter what's in the `p2` module as long as the module declaration is inside the `cfg_if` block. If it is moved outside the block, things are properly tested.

libc-test/build.rs
```diff
@@ -1455,6 +1455,7 @@ fn test_dragonflybsd(target: &str) {
 
 fn test_wasi(target: &str) {
     assert!(target.contains("wasi"));
+    let p2 = target.contains("p2");
```
tgross35, line 1458:
```suggestion
    let p2 = target.contains("wasip2");
```

Just to make this a bit more clear

tgross35:
> Now I'm running into a strange issue, where the tests succeed no matter what's in the `p2` module as long as the module declaration is inside the `cfg_if` block. If it is moved outside the block, things are properly tested.

Do you mean locally or in CI? It sounds a bit like the `p2` target_env isn't being set - either the target is showing up as `p1` somehow, or the compiler doesn't properly set `target_env` for wasi?. You can put a `compile_error!("whatever")` in that cfg_if block to make sure that block does actually get enabled.

alexcrichton:
I believe that `ctest2`, which is used to generate the crate-to-test, has a [hard-coded list of default cfgs](https://github.com/JohnTitor/ctest2/blob/80f9cc420ed2739631f1604f44744107ad10d9a0/src/lib.rs#L1045) which doesn't include wasip2. For that it might make sense to manually specify `cfg("target_env", "p2")` for now?

nickrum:
This was happening locally and only for the `ctest2`-based tests. @alexcrichton got it exactly right. This was due to the hard-coded list of default cfgs and setting `target_env` manually solved the issue.

tgross35:
Okay, great! I didn't take an in-depth look at everything here but @alexcrichton if this looks good to you, I am happy to merge.

src/wasi/p2.rs
```diff
@@ -0,0 +1,174 @@
...
+        name = "c",
+        kind = "static",
+        modifiers = "-bundle",
+        cfg(target_feature = "crt-static")
+    )
+)]
+#[cfg_attr(
+    feature = "rustc-dep-of-std",
+    link(name = "c", cfg(not(target_feature = "crt-static")))
+)]
```
alexcrichton, line 123: This is already present a level above this so it should be ok to not need to duplicate it here too.

alexcrichton (approved):
Looks good to me!

src/wasi/p2.rs
nickrum: Ah, okay! I thought this had to be duplicated since those two extern blocks aren't nested within each other and two separate extern blocks could in theory be linked to two different libraries. I removed it 👍

tgross35:
Thanks for authoring Nicola & thanks for reviewing Alex

PR #3893, opened 2024-08-31
Handle version output correctly when `clippy-driver` used

### LINKED ISSUE #1544: Allow clippy-driver maybe?
## The Dream

So I'm playing around and testing how hard it would be to integrate clippy-driver into `cargo build` and `cargo check` by exporting `RUSTC=clippy-driver`. For most crates this works fine and it builds and outputs just as if it were rustc but also you get clippy warnings.

## The Problem

Turns out libc's build.rs file doesn't like clippy very much, and when you try to build you get the following error.

Screenshot from 2019-10-10 15-10-29

Which seems to be caused by this section of the build.rs file

https://github.com/rust-lang/libc/blob/master/build.rs#L87-L96

## The Proposal

I'm not sure if this is crazy talk but I'd like it if libc could play nicely with clippy-driver. I know old versions will still miscompile so iunno how important helpful fixing this in libc is, but I figured I'd report it anyways

I'm not an expert but I think using `rustc -vV` instead of `rustc --version` might be better and more consistent with how the rest of the ecosystem gets version info from the compiler (thinking about cargo and sccache here). Here is the comparable outputs of those two invocations on rustc and clippy-driver

Screenshot from 2019-10-10 15-09-50

nathaniel-bennett:
Checks to see if the output of `$RUSTC --version` is actually a `clippy-driver` version. If so, reruns with the appropriate flags to correctly output the rust version from `clippy-driver`.

Resolves #1544.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,5 +1,6 @@
 use std::env;
-use std::process::Command;
+use std::ffi::{OsStr, OsString};
+use std::process::{Command, Output};
 use std::str;
 
 // List of cfgs this build script is allowed to set. The list is needed to support check-cfg, as we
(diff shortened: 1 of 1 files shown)


tgross35:
Thanks for the fix!

Further context: https://github.com/rust-lang/libc/pull/3842 added RUSTC_WRAPPER support. https://github.com/rust-lang/libc/pull/3845 is the backport, it was reported on that thread that using `RUSTC_WRAPPER=clippy-driver` got broken from the change. This should fix it.

tgross35 (approved):

PR #3894, opened 2024-09-01
Fix alignment of uc_mcontext in ucontext_t on arm64 android

### LINKED ISSUE #3655: The offset of mcontext_t inside the libc::ucontext_t struct is incorrect
Cpp code:

```cpp
DEBUG("off: 0x%x\n", (size_t)(&((ucontext_t *) data)->uc_mcontext) - (size_t)data);
```

output:

```text
off: 0xB0
```

Rust code:

```rust
let ucontext = data as *const libc::ucontext_t;
let start = ucontext as usize;
let uc_mcontext_ref = &(*ucontext).uc_mcontext;
let uc_mcontext_ptr = uc_mcontext_ref as *const _;
let end = uc_mcontext_ptr as usize;
log::info!("off:0x{:x}",end -star);
```

output:

```text
off:0x30
```

build for target aarch64-linux-android

CUB3D:
On ARM64 Android there should be padding between `uc_sigmask` and `uc_mcontext` in `libc::ucontext_t`, see  [here](https://cs.android.com/android/platform/superproject/main/+/main:prebuilts/runtime/mainline/runtime/sdk/android/arm64/include/bionic/libc/kernel/uapi/asm-arm64/asm/ucontext.h;l=15;drc=9f2b634c25c4af1a26f6a92263fef48377f0ddec)

Now `core::mem::offset_of!(libc::ucontext_t, uc_mcontext)` is the expected value of `0xB0`

fixes #3655

### DIFF
--- a/src/unix/linux_like/android/b64/aarch64/align.rs
+++ b/src/unix/linux_like/android/b64/aarch64/align.rs
@@ -12,6 +12,7 @@ s! {
         pub uc_link: *mut ucontext_t,
         pub uc_stack: ::stack_t,
         pub uc_sigmask: ::sigset_t,
+        pub __pad: [u8; 1024 / 8 - core::mem::size_of::<::sigset_t>()],
         pub uc_mcontext: mcontext_t,
     }
 


tgross35:
Thanks! Does `[u8; 1024 / 8 - size_of::<sigset_t>()]` work here to mirror the source? The field also needs to be public, otherwise this isn't possible to construct (I suspect there will be a CI error)

CUB3D:
> Thanks! Does `[u8; 1024 / 8 - size_of::<sigset_t>()]` work here to mirror the source? The field also needs to be public, otherwise this isn't possible to construct (I suspect there will be a CI error _edit: nevermind, we don't run aarch64 android_).

Apparently it does! Also fixed the visibility, although you still won't be able to construct a `ucontext_t` because of the `sigset_t`

tgross35 (approved):
Thanks!

Since this is a fix to a tier 2 target, it should be okay to backport.

tgross35:
Hey @maurer would you mind sanity checking this? I was going to backport it but figure it is worth double checking since this is a breaking change.

maurer:
tl;dr: Change is fine, but I think we have this bug for `aarch64` on base Linux, and need to update `riscv` as well in the same places, am I missing something?

This both seems correct and should be safe to backport, but I think it's actually incomplete.

For Android, this is also the case in [riscv](https://cs.android.com/android/platform/superproject/main/+/main:prebuilts/runtime/mainline/runtime/sdk/android/arm64/include/bionic/libc/kernel/uapi/asm-riscv/asm/ucontext.h;l=15;drc=9f2b634c25c4af1a26f6a92263fef48377f0ddec) - we should probably add it to the [rust-side riscv](https://github.com/rust-lang/libc/blob/7f2e466936fe4857df7351a40c1b49e27f53ad6b/src/unix/linux_like/android/b64/riscv64/align.rs)

However, I think this may be even worse - I checked the Linux [headers for an unmodified upstream kernel](https://github.com/torvalds/linux/blob/master/arch/arm64/include/uapi/asm/ucontext.h#L28) - they also have this padding, but our [`aarch64-linux` implementation](https://github.com/rust-lang/libc/blob/7f2e466936fe4857df7351a40c1b49e27f53ad6b/src/unix/linux_like/linux/gnu/b64/aarch64/align.rs#L10-L16) currently doesn't.

Issue #3615, opened 2024-03-07
Signal types missing for some embedded targets


madsmtm:
I'm just getting started with embedded programming, and found that compiling the standard library using rust-analyzer with `-Zbuild-std` fails because of [this line](https://github.com/rust-lang/rust/blob/9c3ad802d9b9633d60d3a74668eb1be819212d34/library/test/src/test_result.rs#L102).

I'm not entirely sure, but it seems like this is entirely valid code, and that the problem lies with `libc` missing definitions for `SIGABRT` (and other signal handler #defines).

As far as I can tell, these [are defined in `newlib`'s `sys/signal.h` header](https://sourceware.org/git/?p=newlib-cygwin.git;a=blob;f=newlib/libc/include/sys/signal.h;h=96bf9781ae9313e4181d0b3c65df1105f1ac79df;hb=HEAD#l258), so perhaps the solution could be as simple as adding these to `libc`'s `newlib` module?

My target triple is `riscv32imc-esp-espidf`.

This would resolve the following issues:
https://github.com/esp-rs/esp-idf-svc/issues/366
https://github.com/rust-lang/rust-analyzer/issues/16552

devnexen:
I would suggest to go for the PR :)

Tevz-Beskovnik:
Hello, are you still planning on PRing this into libc? Otherwise I want to take a crack at it

madsmtm:
I'm fine with you doing it

Vollbrecht:
you initially linked cygwin-newlib implementation. Please keep in mind that esp's uses its own fork of newlib. So if you want to cross reference what is there and what not have a look [here](https://github.com/espressif/newlib-esp32) . If i understand correctly it should cover both newlib functions found in esp rom and the once that get compiled into the binary.

PR #3895, opened 2024-09-02
Change signal constants to c_int on espidf

saethlin:
The espidf targets cannot build libtest because the types of these signal constants disagree with all the other cfg(unix) targets. Since they're of course macros on the C side https://github.com/espressif/newlib-esp32/blob/7cdcb0b0848a5d909e906834fe81903ba9466c90/newlib/libc/include/sys/signal.h#L259-L269 I think `c_int` is a more reasonable type to use. Without this change, attempting to build libtest will produce this error:
```
   --> /home/ben/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/test/src/test_result.rs:108:18
    |
107 |         None => match status.signal() {
    |                       --------------- this expression has type `std::option::Option<i32>`
108 |             Some(libc::SIGABRT) => TestResult::TrFailed,
    |                  ^^^^^^^^^^^^^ expected `i32`, found `usize`
```
It's not entirely clear to me why the conversation on https://github.com/rust-lang/libc/issues/3615 and the issues from there petered out. I'm a bit concerned that the users just configured their editors to suppress the error and moved on. That's certainly what it looks like.

I've checked that this fixes the problem by pointing my rust-lang/rust checkout to my libc branch and running
```
MIRI_LIB_SRC=/home/ben/rust/library cargo +nightly miri setup --target riscv32imafc-esp-espidf
```
or
```
x check library/test --target riscv32imafc-esp-espidf
```

### DIFF
--- a/src/unix/newlib/espidf/mod.rs
+++ b/src/unix/newlib/espidf/mod.rs
@@ -89,14 +89,14 @@ pub const MSG_EOR: ::c_int = 0x08;
 
 pub const PTHREAD_STACK_MIN: ::size_t = 768;
 
-pub const SIGABRT: ::size_t = 1;
-pub const SIGFPE: ::size_t = 1;
-pub const SIGILL: ::size_t = 1;
-pub const SIGINT: ::size_t = 1;
-pub const SIGSEGV: ::size_t = 1;
-pub const SIGTERM: ::size_t = 1;
-pub const SIGHUP: ::size_t = 1;
-pub const SIGQUIT: ::size_t = 1;
+pub const SIGABRT: ::c_int = 1;
+pub const SIGFPE: ::c_int = 1;
+pub const SIGILL: ::c_int = 1;
+pub const SIGINT: ::c_int = 1;
+pub const SIGSEGV: ::c_int = 1;
+pub const SIGTERM: ::c_int = 1;
+pub const SIGHUP: ::c_int = 1;
+pub const SIGQUIT: ::c_int = 1;
 pub const NSIG: ::size_t = 2;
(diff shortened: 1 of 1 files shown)


saethlin:
cc @Tevz-Beskovnik because you wrote https://github.com/rust-lang/libc/pull/3658, just in case you want to tell me that this PR is a bad idea for some target-specific reason :see_no_evil:

tgross35 (approved):
This looks reasonable to me since the `sigaction` functions still use the default `int` https://github.com/espressif/newlib-esp32/blob/7cdcb0b0848a5d909e906834fe81903ba9466c90/newlib/libc/include/sys/signal.h#L184 (our version  https://github.com/rust-lang/libc/blob/cdf68968a2e40a93027000fb07688ea54f021e2a/src/unix/mod.rs#L1533), I don't know where else these would be used. But I'll give another day or two for Tevz-Beskovnik to respond, just in case.

saethlin:
Also I would rather like to have this in the rustc repo so that the relevant targets can build again. I think that means this needs to go to the 0.2 branch as well? Do you just take care of that?

tgross35:
Yeah, if you just put the stable-nominated label then I'll get it in the next round of cherry picks before my release this week (I already added it here, and it works with rustbot too)

tgross35 (approved):
I want to do a release today so I'm just going to merge this. @Tevz-Beskovnik feel free to leave any feedback if there is a reason this shouldn't be done. We can always change it back if needed since this is tier 3.

PR #3908, opened 2024-09-09
Add missing musl utmpx.h constants

andrewliebenow:


andrewliebenow:
See https://git.musl-libc.org/cgit/musl/tree/include/utmpx.h?h=v1.1.24.

I'm not sure if adding these constants would be considered a breaking change (https://doc.rust-lang.org/cargo/reference/semver.html#item-new).

libc-test/semver/linux-musl.txt
```diff
@@ -1,4 +1,14 @@
 # TODO: musl.
+# linux-gnu.txt is alphabetized, but this file is not
```
tgross35, line 2: Sorted in https://github.com/rust-lang/libc/pull/3954, could you rebase once that merges and sort these in?

andrewliebenow: Done

tgross35 (approved):
Thanks!

PR #3910, opened 2024-09-10
Simplify the RUSTC_WRAPPER check

tgross35:
This should be compatible with older versions of rustc, to get the branches more in sync.


FrankReh:
build-tmp.rs added intentionally?

tgross35:
Ah! Thanks, removed in https://github.com/rust-lang/libc/pull/3930

PR #3914, opened 2024-09-10
Add optional zerocopy crate derives

joshlf:
This is not ready for merging until zerocopy publishes a stable 0.8 release (currently it only has 0.8 alpha releases). I'm putting this up now to test out CI and get the ball rolling in case there are fixes required.


tgross35:
As mentioned in https://github.com/rust-lang/libc/pull/3924#issuecomment-2356572073, we can't take on new large features for 0.2. Thanks for the PR, but it is better to focus both author and reviewer effort for such things into something that targets `main` - it seems like https://github.com/rust-lang/libc/pull/3407 does that.

PR #2135, opened 2021-03-31
Unified all Linux sockopts definitions by arch

zonyitoo:
- ref rust-lang/socket2#213

Fixes #2133

PR #3917, opened 2024-09-11
Add SO_PREFER_BUSY_POLL and SO_BUSY_POLL_BUDGET

tammela:
Remove the comment of these socket options.
Reference: https://elixir.bootlin.com/linux/latest/source/include/uapi/asm-generic/socket.h


tammela:
From the CI results (i686-unknown-linux-musl) it seems like the machine does not have a uapi header with the constant.
They were added in 5.11 and the sanitized headers are in 4.19: https://github.com/rust-lang/libc/blob/main/ci/install-musl.sh#L68

devnexen:
for the time being, need to add those constants as exception, as if musl then ignore those two.

tammela:
@devnexen Got it. I was wondering if someone tried to update the kernel headers for the musl CI.
Alpine uses musl as well and they don't seem to use sabotage-linux headers but rather roll their own (currently targeting 6.6 soon to be 6.10):
https://git.alpinelinux.org/aports/tree/main/linux-headers?h=master

tammela:
I updated the CI to use the alpine headers here: https://github.com/tammela/libc/commit/17af7ba240f5deda442b126ae56b4a2937e6f8bc

Seems to be passing the CI scripts, let me know if it's worth the shot and I can open up a PR

devnexen:
nice, it might be best to create a PR for this musl update alone tough. cc @tgross35

tgross35 (approved):
Huh, no clue why those were commented out but looks like it has been that way since https://github.com/rust-lang/libc/pull/2135. Anyway lgtm, thanks for including the link.

tammela:
Needs this one to pass CI: https://github.com/rust-lang/libc/pull/3921

tgross35 (changes requested):
https://github.com/rust-lang/libc/pull/3921 should be merged soon, would you mind rebasing to resolve conflicts?

Also, could you add these to the semver tests at `libc-test/semver/linux.txt`?

tgross35 (approved):

tammela:
I will double check CI

tammela:
So musl for some reason hard codes the socket constants instead of fetching them from the OS:
https://git.musl-libc.org/cgit/musl/tree/include/sys/socket.h#n293

I will look for a way to test these constants only on gnu libc

tgross35:
It looks like CI is passing - is this ready to be merged or is there something else to be done? (I've held off because of the labels)

tammela:
> It looks like CI is passing - is this ready to be merged or is there something else to be done? (I've held off because of the labels)

I gave in and gated the new constants for !musl. So it should be good from my side.
Musl made the decision to hardcode the SO_ constants instead of inherit them from the OS via `#include <asm-generic/socket.h>`.
Essentially updating the kernel headers is not enough, you have to hope a recent musl version kept up with the new constants...
The next musl version update will likely fill in the gap (till 70), but linux already added a bunch more :)

libc-test/semver/TODO-linux.txt
```diff
@@ -31,6 +31,7 @@ SO_ATTACH_REUSEPORT_EBPF
 SO_BINDTOIFINDEX
 SO_BPF_EXTENSIONS
 SO_BSDCOMPAT
+SO_BUSY_POLL_BUDGET
```
tgross35, line 34: Could you instead add these to `linux-gnu.txt` or `linux-gnu-x86_64.txt` instead?

Since there is something left to do, maybe leave a `// FIXME(musl)` in the `.rs` file rather than using the TODO semver.

tammela: Good catch, will do!

tgross35: I think it needs to go in one of the `linux-gnu*` files rather than `linux-{arch}` since these aren't yet available on musl

tammela: I added to the x86 one, but it should be fine for everything except musl, but it doesn't seem to have such a file

src/unix/linux_like/linux/arch/generic/mod.rs
```diff
@@ -126,10 +131,10 @@ cfg_if! {
         pub const SO_RCVTIMEO_NEW: c_int = 66;
         pub const SO_SNDTIMEO_NEW: c_int = 67;
         pub const SO_DETACH_REUSEPORT_BPF: c_int = 68;
+        pub const SO_PREFER_BUSY_POLL: c_int = 69;
+        pub const SO_BUSY_POLL_BUDGET: c_int = 70;
     }
 }
-// pub const SO_PREFER_BUSY_POLL: c_int = 69;
-// pub const SO_BUSY_POLL_BUDGET: c_int = 70;
```
tgross35, line 132: Sorry, this slipped under the radar. I don't understand what the change is here; it looks like the net effect is changing `SO_PREFER_BUSY_POLL` and `SO_BUSY_POLL_BUDGET` from being available on all platforms to only being available on non-musl platforms. Why is this needed? Musl does seem to define them https://github.com/kraj/musl/blob/1880359b54ff7dd9f5016002bfdae4b136007dde/include/sys/socket.h#L292-L293.

tammela: At the time they didn't have this constants, let me push without them and see if it works now!

tammela: So I went a bit further in the upgrading musl rabbit hole.

Musl 1.2.x now defines time to be 64 bits on 32 bit systems. Unfortunately, that's the version that has the constants we are looking for.
Time in 64 bits changes a bunch of struct definitions and even syscalls so it's a big change [1].
I can open up a PR changing some low hanging fruits to kickstart the transition.
One deficiency we have is that two new types are being used to help the transition: `__kernel_ulong_t` and `__kernel_long_t`. These can be redefined on 32bits to be `long long` variants [1], so far x86 is the only arch expanding them.

The syscall change is more challenging, the tests fail with:
```
bad recvmmsg function pointer: rust: 139377735 (0x84ebc47) != c 139259307 (0x84cedab)
bad gettimeofday function pointer: rust: 139376444 (0x84eb73c) != c 139362562 (0x84e8102)
bad adjtimex function pointer: rust: 139375084 (0x84eb1ec) != c 139209963 (0x84c2ceb)
bad clock_adjtime function pointer: rust: 139375189 (0x84eb255) != c 139210036 (0x84c2d34)
```

Not sure how to this fix those as `SYS_recvmmsg` works on 64-bit but on 32-bit you would need `SYS_recvmmsg_time64` + something else?

[1] https://git.musl-libc.org/cgit/musl/tree/src/network/recvmmsg.c#n21
[2] https://elixir.bootlin.com/linux/v6.13.4/source/arch/x86/include/uapi/asm/posix_types_x32.h#L15

tgross35: If the SO_ constants are useful then we can add them but just skip tests on musl in libc-test/build.rs, separate from the musl upgrade.

The musl upgrade is unfortunately tricky, but very much needs to happen. https://github.com/rust-lang/libc/pull/3791 does most of the changes (and I left some review there), that PR is a cleaned up version of https://github.com/rust-lang/libc/pull/3068. Cc https://github.com/rust-lang/libc/pull/4148 which made linux uapi configurable - we'd probably want something similar for musl so we can unstably test before 1.0.

(If you're considering working on this, feel free to ask me questions on [zulip](https://rust-lang.zulipchat.com/))

tammela: Oh great I didn't know about this PR.
I did a quick run with my patch applied on top of it and it works just fine!

tgross35: To be clear, that PR cannot be merged in its current state. This doesn't need to block the change here though, you can just skip the constants in libc-test/build.rs.

tammela: > To be clear, that PR cannot be merged in its current state
Oh, what is missing?
> This doesn't need to block the change here though, you can just skip the constants on musl in libc-test/build.rs.
Will do

tgross35: > Oh, what is missing?

Not that there is anything fundamentally wrong with the change, but it is pretty out of date and there are a lot of open review comments. I haven't heard from the author in a while so unfortunately it likely needs somebody else to pick it up.

libc-test/build.rs
```diff
@@ -3964,6 +3964,12 @@ fn test_linux(target: &str) {
             if loongarch64 && (name == "MFD_NOEXEC_SEAL" || name == "MFD_EXEC") {
                 return true;
             }
+            // FIXME: Requires musl >= 1.2
```
tgross35, line 3967: Could you make this `FIXME(musl)`? Trying to keep all our fixmes greppable since we have so many of them

tgross35 (approved):
One nit but lgtm

tgross35:
Thanks!

PR #3921, opened 2024-09-13
ci: update musl headers to 6.6

tammela:
Update the musl headers in CI to use alpine-linux instead of sabotage-linux.
Alpine also uses musl but follows the linux stable releases, providing more up-to-date headers.

I also took the opportunity to clean up libc-test/build.rs a bit.

### DIFF
--- a/ci/install-musl.sh
+++ b/ci/install-musl.sh
@@ -64,13 +64,63 @@ esac
 cd ..
 rm -rf "$musl"
 
-# Download, configure, build, and install musl-sanitized kernel headers:
-kernel_header_ver="4.19.88"
-curl --retry 5 -L \
-    "https://github.com/sabotage-linux/kernel-headers/archive/v${kernel_header_ver}.tar.gz" |
-    tar xzf -
+# Download, configure, build, and install musl-sanitized kernel headers.
+
+# Alpine follows stable kernel releases, 3.20 uses Linux 6.6 headers.
+alpine_version=3.20
+alpine_git=https://gitlab.alpinelinux.org/alpine/aports
+
+# This routine piggybacks on: https://git.alpinelinux.org/aports/tree/main/linux-headers?h=3.20-stable
+git clone -n --depth=1 --filter=tree:0 -b "${alpine_version}-stable" "$alpine_git"
 (
-    cd "kernel-headers-${kernel_header_ver}"
-    make ARCH="${kernel_arch}" prefix="/musl-${musl_arch}" install -j4
+    cd aports
+    git sparse-checkout set --no-cone main/linux-headers
+    git checkout
+
+    cd main/linux-headers
+    cp APKBUILD APKBUILD.vars
+    cat <<- EOF >> APKBUILD.vars
+        echo "\$source" > alpine-source
+        echo "\$_kernver" > alpine-kernver
+        echo "\$pkgver" > alpine-pkgver
+        echo "\$sha512sums" > alpine-sha512sums
+EOF
+
+    # Retrieve all the variables
+    sh APKBUILD.vars
+
+    cat APKBUILD.vars
+
+    kernel_version=$(tr -d "[:space:]" < alpine-kernver)
+    pkg_version=$(tr -d "[:space:]" < alpine-pkgver)
+
+    urls=$(grep -o 'https.*' alpine-source)
+    kernel=""
+    patch=""
+    for url in $urls; do
+        base=$(basename "$url")
+        curl --retry 5 -L "$url" > "$base"
+        case $base in
+            linux-*) kernel=$base;;
+            patch-*) patch=$base;;
+        esac
+        # Check if file is known
+        grep -o "$base" alpine-sha512sums
+    done
+
+    # Double check checksums
+    sha512sum -c alpine-sha512sums
+
+    # Extract, apply patches, compile and install headers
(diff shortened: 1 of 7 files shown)


ci/install-musl.sh
```diff
@@ -64,13 +64,50 @@ esac
...
+   for url in $urls; do
+      base=$(basename "$url")
+      curl --retry 5 -L "$url" > "$base"
+      case $base in
+         linux-*) kernel=$base;;
+         patch-*) patch=$base;;
+      esac
+   done
+   bash APKBUILD.sha512 | grep "$kernel" >> sha-check
+   bash APKBUILD.sha512 | grep "$patch" >> sha-check
```
tgross35, line 100: I think this gets a bit cleaner with heredoc and writing the outputs to files rather than filtering:

```suggestion
#!/bin/sh

    # Create a version of APKBUILD to extract variables
    cp APKBUILD APKBUILD.vars
    cat <<- 'EOF' >> APKBUILD.vars
    printf "$source" > alpine-source
    printf "$_kernver" > alpine-kernver
    printf "$pkgver" > alpine-pkgver
    printf "$sha512sums" > alpine-sha512sums
EOF

    # Execute the build script to write the output files
    sh APKBUILD.vars

    alpine_source="$(cat alpine-source)"    
    kernel_version="$(cat alpine-kernver)"
    pkg_version="$(cat alpine-pkgver)"

    kernel=""
    patch=""
    urls="$(echo "$alpine_source" | grep -o 'https.*')"
    
    # Download 
    for url in $urls; do
        base=$(basename "$url")
        curl --retry 5 -L "$url" > "$base"
        case "$base" in
            linux-*) kernel="$base";;
            patch-*) patch="$base";;
            *)
                echo "Unexpected basename $base"
                exit 1
                ;;
        esac
    done

    grep "$kernel" alpine-sha512sums >> sha-check
    grep "$patch" alpine-sha512sums >> sha-check
```

(I haven't tested this)

tgross35: Also, does `CARCH` need to be set before running `APKBUILD`?

tgross35 (commented):
Sorry this hasn't been on my radar. Is the switch to Alpine just because of better maintenance?

Suggested some ways to clean up the shell script, also I think there are a couple other musl dockerfiles that will need to be updated now (loongarch?). `libc-test/build.rs` is probably pretty out of date, if you'd like we can merge the musl changes first and then clean up `build.rs` in a followup.

ci/install-musl.sh
```diff
@@ -64,13 +64,62 @@ esac
...
+   cp APKBUILD APKBUILD.vars
+   {
+   echo "printf \"\$source\" > alpine-source"
+   # shellcheck disable=SC2028
+   echo "printf \"\$_kernver\n\" > alpine-kernver"
+   # shellcheck disable=SC2028
+   echo "printf \"\$pkgver\n\" > alpine-pkgver"
+   # shellcheck disable=SC2028
+   echo "printf \"\$sha512sums\n\" > alpine-sha512sums"
+   } >> APKBUILD.vars
```
tgross35, line 90: Could this just use heredoc rather than the nested quoting? Which shouldn't hit the shellcheck warnings

ci/install-musl.sh
```diff
@@ -64,13 +64,62 @@ esac
...
+   kernel=""
+   patch=""
+   for url in $urls; do
+      base=$(basename "$url")
+      curl --retry 5 -L "$url" > "$base"
+      case $base in
+         linux-*) kernel=$base;;
+         patch-*) patch=$base;;
+      esac
+   done
```
tgross35, line 106: Does the following work instead?

```suggestion
   urls=$(grep -o 'https.*' alpine-source)
   for url in $urls; do
      base=$(basename "$url")
      
      # We only need to download `linux` and the patch archive, not individual
      # patches.
      case "$base" in
         linux-*) kernel="$base";;
         patch-*) patch="$base";;
         *) continue;;
      esac

      curl --retry 5 -L "$url" -o "$base"
      grep "$base" alpine-sha512sums >> sha-check
   done
```

Then remove the `grep "$kernel/patch" alpine-sha512sums >> sha-check` lines below. This makes sure we don't download/use anything that we don't verify the sha of.

If we do actually need to download all the files (`revert-broken-uapi.patch` and `0003-remove-inclusion-of-sysinfo.h-in-kernel.h.patch`), then remove `*) continue` and `>> sha-check`, instead just `sha512sum -c alpine-sha512sums`.

tgross35 (changes requested):
> Sabotage is pretty much stuck on 4.19.88 for years now. Alpine follows the stable releases from [www.kernel.org](http://www.kernel.org) and uses musl. As you saw, this is pretty much a hack to piggy back on Alpine's patches for the kernel headers. I wouldn't claim it's better maintenance over sabotage, but rather it's a better fit since we want to test and add the shiniest new features :)

Thanks for the context, that makes a lot of sense 👍

Two small suggestions for the script then this looks good to me. One other nit - it seems like your bash is indented by three spaces, mind changing to 4 like the rest of the scripts?

ci/install-musl.sh
tammela: We need the latter. I added an explicit check to avoid downloading files that do not contain a checksum.

tgross35 (approved):
Great, thanks for getting this working!

PR #3922, opened 2024-09-13
Linux: epoll: add busy polling parameters

tammela:
In Linux 6.9 a new ioctl for epoll was added:
https://man.archlinux.org/man/ioctl_eventpoll.2.en

It's used for busy polling sockets

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -1036,6 +1036,15 @@ s! {
         pub get_args: __u16,
         pub name: [c_char; ::IFNAMSIZ],
     }
+
+    // #include <linux/eventpoll.h>
+
+    pub struct epoll_params {
+        pub busy_poll_usecs: u32,
+        pub busy_poll_budget: u16,
+        pub prefer_busy_poll: u8,
+        pub __pad: u8, // Must be zero
+    }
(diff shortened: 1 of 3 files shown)


src/unix/linux_like/linux/mod.rs
```diff
@@ -1036,6 +1036,12 @@ s! {
         pub get_args: __u16,
         pub name: [c_char; ::IFNAMSIZ],
     }
+    pub struct epoll_params {
+        busy_poll_usecs: __u32,
+        busy_poll_budget: __u16,
+        prefer_busy_poll: __u8,
+        __pad: __u8,
+    }
```
tgross35, line 1044: The source seems to use uint32_t/uint16_t/uint8_t so these could probably just be u32/u16/u8. I think we use the dunder integers for the others just to be consistent.

Also, any reason these aren't pub?

tgross35:
Could you please update the relevant files in `libc-test/semver`?

src/unix/linux_like/linux/mod.rs
tammela: Sure, my bad. I will export all but `__pad` and use the Rust integers.

tgross35 (approved):

PR #3929, opened 2024-09-17
Feat(Apple): Add the LOCAL_PEERTOKEN socket option

PaulDance:
Dear maintainers,

This proposes to add the Darwin-only `LOCAL_PEERTOKEN` constant as a new API. It is a `getsockopt` option variant that enables retrieving the connected peer's [audit token](https://developer.apple.com/documentation/endpointsecurity/es_process_t/3228975-audit_token?language=objc), which enables uniquely identifying the process with respect to the underlying system. It is taken from XNU's [`sys/un.h`](https://github.com/apple-oss-distributions/xnu/blob/94d3b452840153a99b38a3a9659680b2a006908e/bsd/sys/un.h#L93) header.

Regarding a justification of a use case, this would be for reliably authenticating a peer process that attempts to communicate with the current one after connecting to one of its opened sockets. However, due to it being a simple constant, this is not strictly-speaking necessary. This should therefore be seen as more of an upstreaming and contributing to complete the already-existing APIs as all the other `LOCAL_PEER*` constants are already included.

One thing this also does in addition is to add the small descriptions from the header as documentation comments in the Rust module, since I could see it was done from time to time, although not always. It is also done for the other similar constants, for completion's sake. If needed, this can be restricted to only `LOCAL_PEERTOKEN` or even removed entirely.

Cheers,
Paul.


PaulDance:

seems appropriate?

tgross35 (approved):
LGTM, thanks!

PaulDance:
Cool, thanks! Should I open a PR for the backport to 0.2 myself or will it be done soon anyway?

tgross35:
Before doing a release I go through and cherry pick the backports, nothing for you to do :)

PR #3938, opened 2024-09-22
linux: Make elf name type constants always available

SteveLauC:
### What does this PR do

Adds some register set values for Linux/musl

Linux header: https://github.com/torvalds/linux/blob/88264981f2082248e892a706b2c5004650faac54/include/uapi/linux/elf.h#L378-L382

-----

We can port this patch to the `libc-0.2` branch.


tgross35:
Thanks for the PR! Are these also available through glibc? If so they should probably just go to `linux/mod.rs` (if not then lgtm).

SteveLauC:
> Are these also available through glibc?

Yeah, we already have them in `gnu/mod.rs`, considering that they are directly defined in the kernel, I guess they are available on all the libcs (Not quite sure, let's test this with our CI).

> If so they should probably just go to linux/mod.rs

Let me give this a try:)

SteveLauC:
> Let me give this a try:)

Done:)

tgross35 (approved):
Thanks!

tgross35:
Docs for reference https://man7.org/linux/man-pages/man5/elf.5.html

PR #3939, opened 2024-09-23
Fix "struct stat" on 32-bit FreeBSD 12+

asomers:
It has always been broken.  But we never noticed because we don't run CI on 32-bit FreeBSD, and until #3723 our users always used a FreeBSD 11 ABI.

Fixes https://github.com/rust-lang/rust/issues/130677


tgross35:
Thanks for the fix, can this target `main` or is it conflict-y? Also, do you have a link to the relevant headers?

asomers:
> Thanks for the fix, can this target `main` or is it conflict-y? Also, do you have a link to the relevant headers?

We'll need to merge it to both branches, but there's a slight conflict with main.  As for headers:
https://github.com/freebsd/freebsd-src/blob/ee3da604dd016439850dae77366796313e60f0e0/sys/sys/stat.h#L122
https://github.com/freebsd/freebsd-src/blob/ee3da604dd016439850dae77366796313e60f0e0/sys/sys/stat.h#L157

tgross35:
Thanks for the links, easy enough.

For future reference I've been preferring PRs to target `main` first so that branch can't wind up missing anything, then I do periodic backports. But since this is a fix we might as well get it in.

@rustbot label +main-needs-cherry-pick

PR #3940, opened 2024-09-24
android: add mallinfo2

SamB:
It's actually just mallinfo in a funny hat, just like in C; see <https://android-review.googlesource.com/c/platform/bionic/+/1910571> for the review of when they added this to Bionic's malloc.h header.

Since it's actually just mallinfo in dsguise, there is no minimum SDK level (just like for mallinfo).

Unlike in the header, I had to copy and paste the struct, because that's easier than building/borrowing a Rust macro to duplicate the fields. (Sometimes worse really is better.)

### DIFF
--- a/libc-test/semver/android.txt
+++ b/libc-test/semver/android.txt
@@ -3454,6 +3454,7 @@ madvise
 major
 makedev
 mallinfo
+mallinfo2
 malloc
 malloc_usable_size
 mcontext_t
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -112,6 +112,20 @@ s! {
         pub keepcost: ::size_t,
     }
 
+    // This is identical to struct mallinfo on Android
+    pub struct mallinfo2 {
+        pub arena: ::size_t,
+        pub ordblks: ::size_t,
+        pub smblks: ::size_t,
+        pub hblks: ::size_t,
+        pub hblkhd: ::size_t,
+        pub usmblks: ::size_t,
+        pub fsmblks: ::size_t,
+        pub uordblks: ::size_t,
+        pub fordblks: ::size_t,
+        pub keepcost: ::size_t,
+    }
+
     pub struct flock {
         pub l_type: ::c_short,
         pub l_whence: ::c_short,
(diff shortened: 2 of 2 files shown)


SamB:
Hmm, I was really hoping there was Android CI because there are a lot of funny little failures when I run "cargo test" on my phone (and that's after skipping the tests for sem_open/sem_close/sem_unlink that Termux `#define`s to names not found in libc.so).

Lucky for me, it looks like more checks materialized while I was writing this comment.

Well, the dynamic linker looks a bit grouchy, but the tests themselves appear to pass.

See e.g. 
<https://github.com/rust-lang/libc/actions/runs/11006374551/job/30560642282?pr=3940#step:4:972> for the linker complaints:

```
linker: /checkout/target/x86_64-linux-android/debug/deps/main-5f4c44db87887603: unsupported flags DT_FLAGS_1=0x8000001
```

No clue why the dynamic linker on my phone is happy with the same flags ...

tgross35:
This looks fine to me, @maurer mind double checking?

maurer:
Seems fine from an Android POV, that function is indeed available.

That said:
1. Maybe we should use a type alias rather than a separate struct? Not a strong opinion here though.
2. What are you looking for `mallinfo2` for? `malloc_info` is more reliable and supported from API level 23, which accounts for 98.6% of existing devices. (It is true though, that `rustc` still supports back to API level 21.)

tgross35:
Labeled waiting on author since the type alias makes sense here as Matthew mentioned (`pub type mallinfo2 = mallinfo`). Just comment `@rustbot review` once that is done.

> What are you looking for `mallinfo2` for? `malloc_info` is more reliable and supported from API level 23, which accounts for 98.6% of existing devices. (It is true though, that `rustc` still supports back to API level 21.)

It would be good to confirm that this is actually useful as well (easy enough if you are supporting an old API).

tgross35:
@SamB could you change this to make use of type aliases? Also confirm that this is for situations where `malloc_info` isn't available.

tgross35:
@SamB I am going to close this since it has been inactive for a little while. Feel free to reopen or submit a PR if you make the requested changes, or just switch to `malloc_info` as noted above.

PR #3942, opened 2024-09-24
Add Trusty OS support

randomPoison:
This PR adds support for the [Trusty secure operating system](https://source.android.com/docs/security/features/trusty). This upstreams [the patch that we have been using](https://android.googlesource.com/platform/external/rust/crates/libc/+/refs/heads/main/patches/trusty.patch) internally, cleaned up to meet the project's coding conventions.

This is a revival of https://github.com/rust-lang/libc/pull/2987.

Relevant headers:

* `PROT_READ`, `PROT_WRITE`, and `MAP_FAILED` come from [`mman.h`](https://cs.android.com/android/platform/superproject/main/+/main:external/trusty/musl/include/sys/mman.h).
* `CLOCK_BOOTTIME` comes from [`time.h`](https://cs.android.com/android/platform/superproject/main/+/main:external/trusty/musl/include/time.h).
* `STDOUT_FILENO` and `STDERR_FILE_NO` come from [`unistd.h`](https://cs.android.com/android/platform/superproject/main/+/main:external/trusty/musl/include/unistd.h).
* `AT_PAGESZ` comes from [`elf.h`](https://cs.android.com/android/platform/superproject/main/+/main:external/trusty/musl/include/unistd.h).


randomPoison:
Addressing a comment from the original PR, @JohnTitor asked:

> I wonder if we could move the module under linux_like or somewhere else similar to that. I'm not familiar with Trusty but is it completely different from Linux-like OS (i.e. exposing linux-like items on Trusty is harmful)?

Yeah, Trusty is not linux-like, nor is it a Unix. It has some overlap with those platforms, but isn't part of the same family of systems. I think the right thing to do is have Trusty be its own module.

randomPoison:
I think this would also be a good candidate for merging over to the 0.2 branch. The diff should apply cleanly, I've tested it against both branches.

randomPoison:
What should I do about the unknown `target_os = "trusty"` cfg in CI? The targets aren't yet in the stable compiler. I've got this `Cargo.toml` snippet I've been using locally

```toml
[lints.rust.unexpected_cfgs]
level = "warn"
check-cfg = [
    'cfg(target_os, values("trusty"))',
]
```

Is that appropriate to add in the PR as a temporary way to suppress the lint? Or is there a better way to handle this?

randomPoison:
@tgross35 We're using musl as our libc implementation, so those definitions come from the musl headers.

* `PROT_READ`, `PROT_WRITE`, and `MAP_FAILED` come from [`mman.h`](https://cs.android.com/android/platform/superproject/main/+/main:external/trusty/musl/include/sys/mman.h).
* `CLOCK_BOOTTIME` comes from [`time.h`](https://cs.android.com/android/platform/superproject/main/+/main:external/trusty/musl/include/time.h).
* `STDOUT_FILENO` and `STDERR_FILE_NO` come from [`unistd.h`](https://cs.android.com/android/platform/superproject/main/+/main:external/trusty/musl/include/unistd.h).
* `AT_PAGESZ` comes from [`elf.h`](https://cs.android.com/android/platform/superproject/main/+/main:external/trusty/musl/include/unistd.h).

Also it looks like a rebase to the latest `main` fixed the unknown platform error, so I didn't need to update `build.rs`.

randomPoison:
oh I forgot to say the magic word

@rustbot review

tgross35 (approved):
Added the headers to the PR description so they make it into git history. Thanks!

PR #3946, opened 2024-09-27
Fix the definition of struct stat on 32-bit FreeBSD 12+

asomers:
This is a rebase #3939 onto the main branch.


tgross35 (approved):
Ah sorry for the delay, this lgtm but needs a rebase now

PR #3948, opened 2024-09-28
Fix tests on FreeBSD x86

asomers:
This PR fixes the tests on FreeBSD 15 x86.  Many of the symbols were wrong for all 32-bit architectures.  All of these symbols are fairly obscure, so no real users are known to have encountered problems with them.  The discrepancies were never caught because we do CI only on FreeBSD x86_64.

Note: this PR must be rebased after #3946 merges.  That PR is much more important since it involves "struct stat", whereas this PR only concerns more obscure symbols.


asomers:
@tgross35 this is a follow up to the other PRs you've reviewed for me lately.

tgross35:
This looks fine at first glance. Could you add links to the relevant headers in the PR description? At least for the few variables that have different values per-arch.

asomers:
@tgross35 I added links to the header files in all of the commit messages and force-pushed.

tgross35 (approved):
Thanks for the perfect commit history, I didn't look in detail but everything seems reasonable here.

PR #3950, opened 2024-09-29
Fix CI for FreeBSD 15

asomers:
It was failing for two reasons:

* 87fbd9fc7[^1] removed the TCP_MAXPEAKRATE symbol.
* 3458bbd39[^2] changed the value of RLIM_NLIMITS

Fixes #3947

[^1]: https://github.com/freebsd/freebsd-src/commit/87fbd9fc7fc5f8d79fe5e3dcd13ad02b11a67ef0
[^2]: https://github.com/freebsd/freebsd-src/commit/3458bbd397783f3bb62713c54ae87f19eeb98dc0

PR #3952, opened 2024-09-29
Fix CI for FreeBSD 15, on libc-0.2 branch

### LINKED ISSUE #3947: CI fails on FreeBSD 15 for TCP_MAXPEAKRATE
CI is currently failing on FreeBSD 15 because TCP_MAXPEAKRATE is unknown.  That's because it was recently removed by this commit:
https://github.com/freebsd/freebsd-src/commit/87fbd9fc7fc5f8d79fe5e3dcd13ad02b11a67ef0 .  That, and other constants removed by the same commit, should be removed from libc's main branch and deprecated in the libc-0.2 branch.

asomers:
It was failing for two reasons:

* 87fbd9fc71[^1] removed the TCP_MAXPEAKRATE symbol.
* 3458bbd392[^2] changed the value of RLIM_NLIMITS

Fixes #3947

[^1]: https://github.com/freebsd/freebsd-src/commit/87fbd9fc7fc5f8d79fe5e3dcd13ad02b11a67ef0
[^2]: https://github.com/freebsd/freebsd-src/commit/3458bbd397783f3bb62713c54ae87f19eeb98dc0

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -3843,6 +3843,7 @@ pub const TCP_INFO: ::c_int = 32;
 pub const TCP_CONGESTION: ::c_int = 64;
 pub const TCP_CCALGOOPT: ::c_int = 65;
 pub const TCP_MAXUNACKTIME: ::c_int = 68;
+#[deprecated(since = "0.2.160", note = "Removed in FreeBSD 15")]
 pub const TCP_MAXPEAKRATE: ::c_int = 69;
 pub const TCP_IDLE_REDUCE: ::c_int = 70;
 pub const TCP_REMOTE_UDP_ENCAPS_PORT: ::c_int = 71;
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2338,6 +2338,9 @@ fn test_freebsd(target: &str) {
             // base system anyway.
             "CTL_MAXID" | "KERN_MAXID" | "HW_MAXID" | "USER_MAXID" => true,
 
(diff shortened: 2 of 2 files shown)


src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -3843,6 +3843,8 @@ pub const TCP_INFO: ::c_int = 32;
 pub const TCP_CONGESTION: ::c_int = 64;
 pub const TCP_CCALGOOPT: ::c_int = 65;
 pub const TCP_MAXUNACKTIME: ::c_int = 68;
+#[doc(hidden)]
+#[deprecated(since = "0.2.160", note = "Removed in FreeBSD 15")]
 pub const TCP_MAXPEAKRATE: ::c_int = 69;
```
tgross35, line 3848: I think it would be better to remove `#[doc(hidden)]` for now so it doesn't disappear entirely. We can add it back in a couple versions.

tgross35:
Could you amend the message to mention 590d78d2a8359067387c0d6b79ec5adc80d90a95 (commit from https://github.com/rust-lang/libc/pull/3950#event-14446527227)? The branches are just a bit easier to manage if similar commits are crosslinked somehow, even though this isn't exactly a cherry pick.

src/unix/bsd/freebsdlike/freebsd/mod.rs
asomers: Ok.  Though note that the sockopt was never actually implemented within the kernel, so no program that tried to use it could've possibly worked.

tgross35 (approved):
Thanks!

PR #3956, opened 2024-10-01
update/add missing AF_XDP structs & constants

arctic-alpaca:
# Description

This PR adds missing structs/constants from the [if_xdp.h header](https://github.com/torvalds/linux/blob/v6.11/include/uapi/linux/if_xdp.h) and updates the `xdp_umem_reg` struct. Due to the added field in `xdp_umem_reg`, this is a breaking change from the 0.2 branch.

Ideally, I would like to include all the changes in the 0.2 branch. Would it be possible to add a `xdp_umem_reg_v3` to not break the current `xdp_umem_reg`?
If not, how would deprecating a struct to change it work? Just add a deprecation notice, wait for X amount of time, add the field?

# Sources

AF_XDP documentation: https://docs.kernel.org/networking/af_xdp.html
Header file: https://github.com/torvalds/linux/blob/v6.11/include/uapi/linux/if_xdp.h

# Checklist

I could not test the the changes locally due to the error described in https://github.com/rust-lang/libc/pull/3865#issuecomment-2316180504


tgross35:
Thanks for the PR, I'm taking a look at this. Where is `xsk_tx_metadata_completion` defined? I can't seem to find it anywhere.

> Ideally, I would like to include all the changes in the 0.2 branch. Would it be possible to add a `xdp_umem_reg_v3` to not break the current `xdp_umem_reg`?

The kernel removed the old interface because it was buggy, right? If the kernel did a breaking change here then I think we can just do the same.

arctic-alpaca:
>I'm taking a look at this.

Thank you!

>Where is `xsk_tx_metadata_completion` defined?

[`xsk_tx_metadata_completion`](https://github.com/torvalds/linux/blob/v6.11/include/uapi/linux/if_xdp.h#L147-L150) and [`xsk_tx_metadata_request`](https://github.com/torvalds/linux/blob/v6.11/include/uapi/linux/if_xdp.h#L138-L145) are the two possible variants of the [anonymous union](https://github.com/torvalds/linux/blob/v6.11/include/uapi/linux/if_xdp.h#L137-L151) in [`xsk_tx_metadata`](https://github.com/torvalds/linux/blob/v6.11/include/uapi/linux/if_xdp.h#L134C8-L152). 

>The kernel removed the old interface because it was buggy, right?

Correct, the issue was initially reported [here](https://lore.kernel.org/bpf/2d6ff64a-5e2c-4078-a8d1-84f1ff3361ce@arctic-alpaca.de/T/#u) with the fix removing the `xdp_umem_reg_v2` struct. The current situation is either the new (`xdp_umem_reg`) struct is used or the kernel falls back to using `xdp_umem_reg_v1`, which silently ignores the `flags` field ([source](https://github.com/torvalds/linux/blob/v6.12-rc6/net/xdp/xsk.c#L1360-L1367)). Silently ignoring the `flags` field being a breaking change in the kernel.

src/unix/linux_like/linux/gnu/mod.rs
```diff
@@ -433,6 +434,15 @@ s! {
         pub options: ::__u32,
     }
 
+    pub struct xsk_tx_metadata_completion {
```
tgross35, line 437: Everything added here is uapi right? It should probably go in `src/unix/linux_like/linux/mod.rs` rather than duplicating into gnu and musl. (The existing xdp types could probably move as well? Not in this PR though.)

tgross35 (changes requested):
I did a bit of looking and it seems like `rustix` is about the only consumer of the specific type that would break. I asked about possible fallout here https://github.com/bytecodealliance/rustix/issues/1221, if they say they can work around us breaking things then that is much easier than figuring out the best mitigation here.

In any case we can get this merged to `main`, left a comment about deduplicating gnu and musl.

@rustbot author

src/unix/linux_like/linux/gnu/mod.rs
arctic-alpaca: > Everything added here is uapi right?

Yes.

> The existing xdp types could probably move as well?

I think so, I wasn't sure where to put them when I initially added them. I'll create a PR after this one is merged to move the rest too.

arctic-alpaca:
Everything that was added in this PR (besides the field addition) is moved to `src/unix/linux_like/linux/mod.rs` now.

@rustbot review

tgross35 (approved):

tgross35:
Thanks for moving things. I'll merge this now but hold off for a few more days on the backport to hear back from rustix (I am leaning toward just doing the backport, we can't reasonably do a major release in every case where the API we bind breaks).

PR #3909, opened 2024-09-10
Initial support for NuttX

no1wudi:
Define the essential types that for NuttX OS.

NuttX has beed added to Rust by https://github.com/rust-lang/rust/pull/127755, this patch is try to introduce basic libc support for NuttX, see the discussion in NuttX community: 
https://github.com/apache/nuttx/issues/12960
https://github.com/apache/nuttx/pull/13245

PR #3961, opened 2024-10-10
Initial support for NuttX (#3909)

no1wudi:
Define the essential types that for NuttX OS.

This is picked from https://github.com/rust-lang/libc/pull/3909


tgross35:
I got this backport in a group https://github.com/rust-lang/libc/pull/3968. Usually I do them periodically and catch the stable-nominated issues. All the same, thanks for putting this up!

PR #3962, opened 2024-10-11
emscripten: Upgrade emsdk to 3.1.68

kleisauke:
In line with commit https://github.com/rust-lang/rust/commit/2c38ecfc9077000db7de14bf501fa5294eeecedd.

Resolves: #3306.
Supersedes: #3569.
Cherry-picks a couple of changes from: #3307.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2779,31 +2779,21 @@ fn test_emscripten(target: &str) {
                "stdio.h",
                "stdlib.h",
                "string.h",
-               "sys/epoll.h",
-               "sys/eventfd.h",
                "sys/file.h",
                "sys/ioctl.h",
                "sys/ipc.h",
                "sys/mman.h",
                "sys/mount.h",
                "sys/msg.h",
-               "sys/personality.h",
-               "sys/prctl.h",
-               "sys/ptrace.h",
-               "sys/quota.h",
-               "sys/reboot.h",
                "sys/resource.h",
                "sys/sem.h",
                "sys/shm.h",
-               "sys/signalfd.h",
                "sys/socket.h",
                "sys/stat.h",
                "sys/statvfs.h",
-               "sys/swap.h",
                "sys/syscall.h",
                "sys/sysinfo.h",
                "sys/time.h",
-               "sys/timerfd.h",
                "sys/times.h",
                "sys/types.h",
                "sys/uio.h",
@@ -2828,8 +2818,6 @@ fn test_emscripten(target: &str) {
         match ty {
             // Just pass all these through, no need for a "struct" prefix
             "FILE" | "fd_set" | "Dl_info" | "DIR" => ty.to_string(),
-
-            "os_unfair_lock" => "struct os_unfair_lock_s".to_string(),
 
             // LFS64 types have been removed in Emscripten 3.1.44+
             // https://github.com/emscripten-core/emscripten/pull/19812
@@ -2867,6 +2855,10 @@ fn test_emscripten(target: &str) {
             // FIXME: is this necessary?
             "sighandler_t" => true,
 
+            // No epoll support
+            // https://github.com/emscripten-core/emscripten/issues/5033
+            ty if ty.starts_with("epoll") => true,
+
             // FIXME: The size has been changed due to musl's time64
             "time_t" => true,
 
@@ -2892,6 +2884,11 @@ fn test_emscripten(target: &str) {
 
             // FIXME: The size has been changed when upgraded to musl 1.2.2
             "pthread_mutex_t" => true,
+
+            // No epoll support
+            // https://github.com/emscripten-core/emscripten/issues/5033
+            ty if ty.starts_with("epoll") => true,
+            ty if ty.starts_with("signalfd") => true,
 
             // FIXME: Lowered from 16 to 8 bytes in
             // llvm/llvm-project@d1a96e9
@@ -2929,9 +2926,29 @@ fn test_emscripten(target: &str) {
             // FIXME: emscripten uses different constants to constructs these
             n if n.contains("__SIZEOF_PTHREAD") => true,
 
+            // No epoll support
+            // https://github.com/emscripten-core/emscripten/issues/5033
+            n if n.starts_with("EPOLL") => true,
+
+            // No ptrace.h
+            // https://github.com/emscripten-core/emscripten/pull/17704
+            n if n.starts_with("PTRACE_") => true,
+
+            // No quota.h
+            // https://github.com/emscripten-core/emscripten/pull/17704
+            n if n.starts_with("QIF_") => true,
+            "USRQUOTA" | "GRPQUOTA" | "Q_GETFMT" | "Q_GETINFO" | "Q_SETINFO" | "Q_SYNC"
+            | "Q_QUOTAON" | "Q_QUOTAOFF" | "Q_GETQUOTA" | "Q_SETQUOTA" => true,
+
             // FIXME: `SYS_gettid` was removed in
             // emscripten-core/emscripten@6d6474e
             "SYS_gettid" => true,
+
+            // No personality.h
+            // https://github.com/emscripten-core/emscripten/pull/17704
+            "ADDR_NO_RANDOMIZE" | "MMAP_PAGE_ZERO" | "ADDR_COMPAT_LAYOUT" | "READ_IMPLIES_EXEC"
+            | "ADDR_LIMIT_32BIT" | "SHORT_INODE" | "WHOLE_SECONDS" | "STICKY_TIMEOUTS"
+            | "ADDR_LIMIT_3GB" => true,
 
             // FIXME: These values have been changed
             | "POSIX_MADV_DONTNEED" // to 4
@@ -2973,17 +2990,6 @@ fn test_emscripten(target: &str) {
         // musl names this __dummy1 but it's still there
         // FIXME: is this necessary?
         (struct_ == "glob_t" && field == "gl_flags") ||
-        // musl seems to define this as an *anonymous* bitfield
-        // FIXME: is this necessary?
-        (struct_ == "statvfs" && field == "__f_unused") ||
-        // sigev_notify_thread_id is actually part of a sigev_un union
-        (struct_ == "sigevent" && field == "sigev_notify_thread_id") ||
-        // signalfd had SIGSYS fields added in Linux 4.18, but no libc release has them yet.
-        (struct_ == "signalfd_siginfo" && (field == "ssi_addr_lsb" ||
-                                           field == "_pad2" ||
-                                           field == "ssi_syscall" ||
-                                           field == "ssi_call_addr" ||
-                                           field == "ssi_arch")) ||
         // FIXME: After musl 1.1.24, it have only one field `sched_priority`,
         // while other fields become reserved.
         (struct_ == "sched_param" && [
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2739,7 +2739,6 @@ fn test_emscripten(target: &str) {
     cfg.define("_GNU_SOURCE", None); // FIXME: ??
 
     headers! { cfg:
-               "aio.h",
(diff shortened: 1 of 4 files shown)


hoodmane:
Thanks @kleisauke it would be nice to see time_t finally get fixed...

juntyr:
Could/should this PR keep the emcc version check to give a nice error at build time if a now-no-longer supported Emscripten is used?

hoodmane:
I wouldn't want rust to generate an error when building with Emscripten 3.1.58 since that is the version that Pyodide is using with Python 3.12 and it would prevent us from upgrading. Though I suppose that it would just require us to use a patched `rust/libc` and we are already doing that, it would be nice to avoid additional patches.

hoodmane:
But having a version check would be fine as long as there is an override.

juntyr:
I was thinking about the previous check for 3.1.42 when the ABI change happened, which previous libc versions still supported (now there would be an ABI break in the opposite direction).

I definitely agree that other than that pretty much any emcc version has to be allowed since there is no current indicator which versions are ABI compatible and which aren’t

tgross35:
How backwards compatible is emsdk? I think putting this on the 0.2 stable branch should be okay, but would like to know if there is any expected breakage here.

kleisauke:
Updated now that PR https://github.com/rust-lang/rust/pull/131533 was landed. This is ready for review.

> How backwards compatible is emsdk?

AFAIK, there were no ABI-breaking changes after Emscripten 3.1.42 (released at 2023/06/23), which was addressed in libc by commit 63b0d673eaf2a177ff208c5c50999738bdd1bf0d (but only when building with `-Zbuild-std`).

kleisauke:
> Could/should this PR keep the emcc version check to give a nice error at build time if a now-no-longer supported Emscripten is used?

I could address this within a follow-up PR, but perhaps its severity should be a warning instead? Simple programs that don't use filesystem operations will likely still work with Emscripten versions prior to 3.1.42. On the other hand, Emscripten 3.1.42 is already more than a year old.

hoodmane:
> AFAIK, there were no ABI-breaking changes after Emscripten 3.1.42

@kleisauke it would be really great if we could make some sort of tool to do ABI compatibility checks. For a static test, we could compare the contents of `emscripten/src/struct_info_generated.json`. Perhaps we could use `emscripten/tools/maint/gen_struct_info.py` to generate info about the layout of more standard library structs than just the ones that are in the default struct_info. For a dynamic test, maybe take the unistd test suite and build the tests as side modules with one emcc version then load them and call into them with a second emcc version. 

This probably wouldn't be perfect, but it would be a lot better than the current method where we don't really know anything.

src/unix/linux_like/emscripten/mod.rs
```diff
@@ -867,22 +863,28 @@ pub const SHM_UNLOCK: ::c_int = 12;
 pub const SHM_HUGETLB: ::c_int = 0o4000;
 pub const SHM_NORESERVE: ::c_int = 0o10000;
 
-pub const QFMT_VFS_OLD: ::c_int = 1;
-pub const QFMT_VFS_V0: ::c_int = 2;
+// No quota.h
+// https://github.com/emscripten-core/emscripten/pull/17704
+/*pub const QFMT_VFS_OLD: ::c_int = 1;
+pub const QFMT_VFS_V0: ::c_int = 2;*/
```
tgross35, line 869: It is awesome to have the emscripten PRs linked, but could you put them all into the commit message and delete the commented-out code? Since these are things we want to keep in history but it isn't relevant to the code anymore.

tgross35:
I don't have any major concerns here but I'll defer to @hoodmane's review (I'll merge once you give it the checkmark)

src/unix/linux_like/emscripten/mod.rs
kleisauke: Good idea! Split into commit 476bf225658143f5689fa2986e6fb491c8081eaf and 99f8c8751f94d6149e55c6d238f310303694ca4c.

libc-test/build.rs
```diff
@@ -2730,7 +2730,6 @@ fn test_emscripten(target: &str) {
     cfg.define("_GNU_SOURCE", None); // FIXME: ??
 
     headers! { cfg:
-               "aio.h",
```
hoodmane, line 2733: Most of these went away in https://github.com/emscripten-core/emscripten/pull/17704/ but I don't see `aio.h` or `sys/sysctl.h`. Were they removed in a separate PR?

src/unix/linux_like/emscripten/align.rs
```diff
@@ -37,10 +37,12 @@ macro_rules! expand_align {
                 size: [u8; ::__SIZEOF_PTHREAD_COND_T],
             }
 
+            // Lowered from 16 to 8 bytes in
+            // https://github.com/llvm/llvm-project/commit/d1a96e906cc03a95cfd41a1f22bdda92651250c7
             #[allow(missing_debug_implementations)]
-            #[repr(align(16))]
+            #[repr(align(8))]
```
hoodmane, line 43: (Irrelevant remark.) The stack still has to be aligned to 16 bytes when you call a function even though maxalign is only 8.  Discovering this wasted a lot of my time.

libc-test/build.rs
```diff
@@ -2922,23 +2896,34 @@ fn test_emscripten(target: &str) {
...
-                => true,
+            // No personality.h
+            // https://github.com/emscripten-core/emscripten/pull/17704
+            "ADDR_NO_RANDOMIZE" | "MMAP_PAGE_ZERO" | "ADDR_COMPAT_LAYOUT" | "READ_IMPLIES_EXEC"
+            | "ADDR_LIMIT_32BIT" | "SHORT_INODE" | "WHOLE_SECONDS" | "STICKY_TIMEOUTS"
+            | "ADDR_LIMIT_3GB" => true,
 
-            // LFS64 types have been removed in Emscripten 3.1.44+
+            // FIXME: https://github.com/emscripten-core/emscripten/pull/14883
+            "SIG_IGN" => true,
```
hoodmane, line 2924: What does this FIXME mean? Did that PR break this test? Why?

hoodmane (approved):
Thanks @kleisauke generally looks good to me.

libc-test/build.rs
kleisauke: This was removed via PR https://github.com/emscripten-core/emscripten/pull/18058. Hence, I've split this removal into commit 476bf225658143f5689fa2986e6fb491c8081eaf.

kleisauke: Oh, indeed `sys/sysctl.h` has been removed via https://github.com/emscripten-core/emscripten/pull/18863. Let me split that change as well.

libc-test/build.rs
kleisauke: That PR did break the libc test, see e.g.:
```
RUNNING ALL TESTS
bad SIG_IGN value at byte 0: rust: 1 (0x1) != c 254 (0xfe)
bad SIG_IGN value at byte 1: rust: 0 (0x0) != c 255 (0xff)
bad SIG_IGN value at byte 2: rust: 0 (0x0) != c 255 (0xff)
bad SIG_IGN value at byte 3: rust: 0 (0x0) != c 255 (0xff)
thread 'main' panicked at /checkout/target/wasm32-unknown-emscripten/debug/build/libc-test-890b1e2d4768b06a/out/main.rs:12:21:
some tests failed
```

However, this `FIXME` can probably not fixed according to 
https://github.com/emscripten-core/emscripten/pull/14883/files#r892912370.

I'll change this to:
```suggestion
            // `SIG_IGN` has been changed to -2 since 1 is a valid function address
            // https://github.com/emscripten-core/emscripten/pull/14883
            "SIG_IGN" => true,
```

Note that other implementations have a similar issue, see e.g:
https://github.com/rust-lang/libc/blob/e4c7accc68b1a376eb1f538d7b991802793afe96/libc-test/build.rs#L1198

build.rs
```diff
@@ -212,28 +205,6 @@ fn which_freebsd() -> Option<i32> {
...
-    // Some Emscripten versions come with `-git` attached, so split the
-    // version string also on the `-` char.
-    let mut pieces = version.trim().split(['.', '-']);
-
-    let major = pieces.next().and_then(|x| x.parse().ok()).unwrap_or(0);
-    let minor = pieces.next().and_then(|x| x.parse().ok()).unwrap_or(0);
-    let patch = pieces.next().and_then(|x| x.parse().ok()).unwrap_or(0);
-
-    Some(major * 10000 + minor * 100 + patch)
-}
```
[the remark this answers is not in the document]
kleisauke: This change was made based on the assumption that Rust requires Emscripten 3.1.42 or later. However, upon review, it appears that the compatibility note in https://github.com/rust-lang/rust/issues/131689 indicates that versions prior 3.1.42 are still supported.

Let me drop commit a35f2120b95413a787edff15411d789416c5bbbf for now.

libc-test/build.rs
kleisauke: Fixed with commit fbeecf3e812b05a4357d0e2ef663766f8cbe2b2b.

libc-test/build.rs
hoodmane: Maybe there should be a constant somewhere that indicates what the value of SIG_IGN is on the platform... But that would presumably be dealt with best in a followup.

src/unix/linux_like/emscripten/align.rs
```diff
@@ -37,10 +37,12 @@ macro_rules! expand_align {
                 size: [u8; ::__SIZEOF_PTHREAD_COND_T],
             }
 
+            // Lowered from 16 to 8 bytes in
+            // https://github.com/llvm/llvm-project/commit/d1a96e906cc03a95cfd41a1f22bdda92651250c7
```
tgross35, line 41: This can just go in the commit message

src/unix/linux_like/emscripten/mod.rs
```diff
@@ -777,9 +746,9 @@ pub const POSIX_MADV_WILLNEED: ::c_int = 3;
 
 pub const AT_EACCESS: ::c_int = 0x200;
 
-pub const S_IEXEC: mode_t = 0o0100;
-pub const S_IWRITE: mode_t = 0o0200;
-pub const S_IREAD: mode_t = 0o0400;
+pub const S_IEXEC: mode_t = 64;
+pub const S_IWRITE: mode_t = 128;
+pub const S_IREAD: mode_t = 256;
```
tgross35, line 751: These look like they are octal in the musl source, could they stay that way?

tgross35 (changes requested):
Just a couple small nits before merge. Thanks @hoodmane for taking a look.

tgross35:
One last thing: could you add the following section to README.md after "Rust version support" so we have this noted somewhere? I wanted to document everything we interface with at some point, might as well start here.

```markdown
## Interface version support

`libc` targets being able to interface with the following minimum library
versions:

- Emscripten 3.1.42, released 2023-06-23
```

src/unix/linux_like/emscripten/align.rs
kleisauke: Split into commit 52fa133223a5a338494f1abab42e2a47a464e0d6 and moved the `time_t` description to the commit message as well.
https://github.com/rust-lang/libc/compare/b6a957f2ec9f24a843d04ec4313f3a4634944072..a37fa9d267ddb3c7cbc9504ac730a7ed6fc11f59

src/unix/linux_like/emscripten/mod.rs
kleisauke: I reverted this part, I can't remember why I did this in the first place.

kleisauke:
@rustbot ready

> One last thing: could you add the following section to README.md after "Rust version support" so we have this noted somewhere? [...]

I think this should be done as a follow-up given that this PR does not raise the minimum Emscripten version (at least when building with `-Zbuild-std`), see:
https://github.com/rust-lang/libc/pull/3962#discussion_r1832494361

PR #3963, opened 2024-10-14
pthread constants and functions for VxWorks

biabbas:
Changes made are supporting changes for https://github.com/iddm/thread-priority/pull/45
1. Changes name of sched_param in VxWorks so that it matches unix platforms
2. Add pthread constants and functions required for thread-priority.

Requesting for stable nomination.

r? @tgross35 

Thank you,
B I Mohammed Abbas

### DIFF
--- a/src/vxworks/mod.rs
+++ b/src/vxworks/mod.rs
@@ -253,7 +253,7 @@ s! {
     }
 
     // b_struct__Sched_param.h
-    pub struct _Sched_param {
+    pub struct sched_param {
         pub sched_priority: ::c_int, /* scheduling priority */
         pub sched_ss_low_priority: ::c_int,    /* low scheduling priority */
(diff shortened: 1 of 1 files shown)


tgross35:
Thanks for the PR, could you provide links to headers or manpages? Or at least copy+paste the relevant bits of the headers you are testing against to the PR description.

tgross35:
Do you have links to headers / docs to check against?

Also FYI the commit summary only says "VxWorks"

biabbas:
> Do you have links to headers / docs to check against?
> 
> Also FYI the commit summary only says "VxWorks"

I am using original sources to check these against. Unfortunately, I can't share them.

tgross35:
Alright, they match linux so I guess that is straightforward enough. It would be good if we could find somebody who can crosscheck these in the future if sources aren't available.

Thanks for updating the commit message.

PR #3714, opened 2024-05-19
Add RTF_*, RTA_*, RTAX_*, RTM_* definitions on BSDs

tesuji:
Grep from:

* https://github.com/freebsd/freebsd-src/raw/main/sys/net/route.h
* https://github.com/DragonFlyBSD/DragonFlyBSD/raw/master/sys/net/route.h
* https://github.com/NetBSD/src/raw/trunk/sys/net/route.h
* https://github.com/openbsd/src/raw/master/sys/net/route.h

Closes #3711.

@rustbot review

PR #3970, opened 2024-10-15
unbreak OpenBSD after #3714 by properly define RTF_FMASK

semarie:
Fix `RTF_FMASK` on OpenBSD. It seems the value used is due to some bad parser.

```c
/* mask of RTF flags that are allowed to be modified by RTM_CHANGE */
#define RTF_FMASK       \
    (RTF_LLINFO | RTF_PROTO1 | RTF_PROTO2 | RTF_PROTO3 | RTF_BLACKHOLE | \
     RTF_REJECT | RTF_STATIC | RTF_MPLS | RTF_BFD)
```


semarie:
the freebsd-15 CI failure is unrelated (`Agent is not responding!`)

tgross35:
@tesuji could you double check this since you authored https://github.com/rust-lang/libc/pull/3714?

tesuji:
> It seems the value used is due to some bad parser.

I used regex that greps for one-line only. Thanks for noticing and fixing this.

PR #3973, opened 2024-10-15
Add getentropy to RTEMS

thesummer:
# Description

With the newest updates in the compiler the `getentropy` method is needed and not exposed for RTEMS.
This PR adds the function prototype to the rtems module

# Sources

* Having the `getentropy` method is mandatory for all RTEMS board support packages according to the [RTEMS BSP documentation](https://docs.rtems.org/branches/master/bsp-howto/getentropy.html).
* The method itself is declared in `unistd.h`

# Checklist


tgross35 (approved):
Thanks!

Issue #3967, opened 2024-10-15
Lots of failures on Cirrus CI / nightly freebsd-15


tgross35:
Almost everything has been failing at least once on freebsd-15, e.g.:

- https://github.com/rust-lang/libc/pull/3966/checks?check_run_id=31530590009
- https://github.com/rust-lang/libc/pull/3963/checks?check_run_id=31497998977
- https://github.com/rust-lang/libc/pull/3960/checks?check_run_id=31394381764

I don't know why this would be, considering freebsd-13 and freebsd-14 seem to work fine.

Cc @tesuji (since you did the CI matrix recently) @asomers

asomers:
Something must be broken in the gcp image. I'll take a look tomorrow, if I have the time. In the mean time, we should pin CI to the last known good image.

tgross35:
Any chance you know which one would be good, or even what the options are? I'm not sure why we seem to be using an image with snap, but I can't find the image list anywhere.

tesuji:
I don't know the root cause of this, but I think sometimes CirrusCI fails to download the GCE image.
Maybe we are on free CirrusCI tier so the build image agents are delayed or we hit the rate limits 🤷 .
The freeebsd15-snap image is listed here: <https://cirrus-ci.org/guide/FreeBSD/#list-of-available-image-families>.
Also if you have a gcloud account, you could check what freebsd images there are with
```
gcloud compute images list --no-standard-images --project=freebsd-org-cloud-dev
```
I don't have any gcloud accounts to check so I'm sorry I can't help.

asomers:
tesuji's command is the right one.  The linked PR tries several different images as an experiment.

asomers:
So due to the libmd .so version bump, even old 15.0-CURRENT images aren't working.  And the newest images don't work with a different error message, possibly for a different reason.  At this point I think the best action for libc to take would be to temporarily disable CI on FreeBSD 15.
Upstream ticket: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=282112

tgross35:
I asked on Zulip, @syphar was able to run the command to get the full list:

```text
NAME                                        PROJECT                FAMILY             DEPRECATED  STATUS
freebsd-13-2-stable-amd64-v20231228         freebsd-org-cloud-dev  freebsd-13-2-snap              READY
freebsd-13-2-stable-amd64-v20240104         freebsd-org-cloud-dev  freebsd-13-2-snap              READY
[...]

PR #3974, opened 2024-10-15
Temporarily disable CI on FreeBSD 15

asomers:
FreeBSD 15 is the unstable development release.  Currently its GCE images available to Cirrus CI don't work because the solib version of libmd was just bumped, and the package builders haven't yet caught up.

Issue #3967

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -7,9 +7,13 @@ task:
     - name: nightly freebsd-14
       freebsd_instance:
         image: freebsd-14-1-release-amd64-ufs
-    - name: nightly freebsd-15
-      freebsd_instance:
-        image_family: freebsd-15-0-snap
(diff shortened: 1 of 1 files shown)


tgross35 (approved):
Thanks for putting this up, I guess this is about the best we can do here for now. Could you add `FIXME` to the comment so it is greppable?

asomers:
@tgross35 comment added.  BTW, there is another option.  If we revert to an image from a couple of weeks ago we can symlink libmd.so.6 -> libmd.so.7.  It's a hack, but it works in this case. Or, we can wait a few days/weeks until newer packages are available and then revert this change.

tgross35:
I'll merge this now just so we have a workaround. A symlink hack sounds fine to me if you care to follow this up, otherwise I don't think we need to worry about it unless this winds up sticking around for more than a couple weeks.

PR #3976, opened 2024-10-16
Enable statx on musl-libc

neuschaefer:
# Description

This PR adds the statx syscall on Linux/musl, which is available since musl 1.2.5.

# Sources

- https://musl.libc.org/releases.html
- https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/stat.h

# Checklist


tgross35:
We now have the config `musl_v1_2_3` that you could gate this under

src/unix/linux_like/mod.rs
```diff
@@ -267,7 +267,12 @@ cfg_if! {
                 pub stx_mnt_id: crate::__u64,
                 pub stx_dio_mem_align: crate::__u32,
                 pub stx_dio_offset_align: crate::__u32,
-                __statx_pad3: [crate::__u64; 12],
+                pub stx_subvol: crate::__u64,
+                pub stx_atomic_write_unit_min: crate::__u32,
+                pub stx_atomic_write_unit_max: crate::__u32,
+                pub stx_atomic_write_segments_max: crate::__u32,
+                __statx_pad2: [crate::__u32; 1],
+                __statx_pad3: [crate::__u64; 9],
```
neuschaefer: I'm dropping the new constant for now, as they seem likely to cause more headaches

PR #3977, opened 2024-10-16
semver/linux-musl: Remove outdated "TODO: musl" comment

neuschaefer:
# Description

The comment was there since the file was otherwise empty, but it has outlived its purpose.

# Sources

_not applicable_

# Checklist


tgross35 (approved):
Sounds about right, thanks for the cleanup!

PR #3978, opened 2024-10-16
linux_like: Unify statx definitions

neuschaefer:
# Description

The statx system call and corresponding constants are defined by the Linux kernel and don't depend on the libc or architecture. The only difference is whether a libc exports the statx syscall wrapper or not.

We can thus unify the statx definitions for all Linux "like" platforms: GNU (glibc), Android (bionic), and (in a later commit) musl.

The statx struct is in a separate file because the style check doesn't allow multiple s! macros in one file, and #[cfg()] doesn't work in s!.

Plain u64 (or uint64_t in C) can't be used for the statx fields because bionic defines them as __u64, and provides incompatible definitions of uint64_t and __u64.

# Sources

_none_

# Checklist

  especially relevant for platforms that may not be checked in CI (failed due to unrelated error, linux/errqueue.h)

### DIFF
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -205,6 +205,45 @@ s! {
     }
 }
 
+cfg_if! {
+    if #[cfg(any(target_env = "gnu", target_os = "android"))] {
+        s! {
+            pub struct statx {
+                pub stx_mask: ::__u32,
+                pub stx_blksize: ::__u32,
+                pub stx_attributes: ::__u64,
+                pub stx_nlink: ::__u32,
+                pub stx_uid: ::__u32,
+                pub stx_gid: ::__u32,
+                pub stx_mode: ::__u16,
+                __statx_pad1: [::__u16; 1],
+                pub stx_ino: ::__u64,
+                pub stx_size: ::__u64,
+                pub stx_blocks: ::__u64,
+                pub stx_attributes_mask: ::__u64,
+                pub stx_atime: statx_timestamp,
+                pub stx_btime: statx_timestamp,
+                pub stx_ctime: statx_timestamp,
+                pub stx_mtime: statx_timestamp,
+                pub stx_rdev_major: ::__u32,
+                pub stx_rdev_minor: ::__u32,
+                pub stx_dev_major: ::__u32,
+                pub stx_dev_minor: ::__u32,
+                pub stx_mnt_id: ::__u64,
+                pub stx_dio_mem_align: ::__u32,
+                pub stx_dio_offset_align: ::__u32,
+                __statx_pad3: [::__u64; 12],
+            }
+
+            pub struct statx_timestamp {
+                pub tv_sec: ::__s64,
+                pub tv_nsec: ::__u32,
+                __statx_timestamp_pad1: [::__s32; 1],
+            }
+        }
+    }
+}
+
 s_no_extra_traits! {
     #[cfg_attr(
         any(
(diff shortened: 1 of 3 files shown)


src/unix/linux_like/mod.rs
```diff
@@ -1886,6 +1886,14 @@ cfg_if! {
     }
 }
 
+// The statx syscall, available on some libcs.
+cfg_if! {
+    if #[cfg(any(target_env = "gnu", target_os = "android"))] {
+        mod linux_statx;
+        pub use self::linux_statx::*;
```
tgross35, line 1893: This looks fine to me, but could you inline the definitions here rather than creating an API-specific module?

tgross35:
Also cc @maurer regarding android changes

tgross35:
@rustbot author for the above.

Just comment `@rustbot review` once changed

src/unix/linux_like/mod.rs
neuschaefer: I tried, but I wasn't able to use `s!` and apply `cfg!` conditions

neuschaefer:
Thanks for reviewing!

@rustbot review

src/unix/linux_like/mod.rs
tgross35: Hm, what exactly is the error? This should work okay, see e.g. https://github.com/rust-lang/libc/blob/c3bc4068ad51d91cfe973e62a51169dbb96bce9c/src/unix/linux_like/linux/mod.rs#L1041-L1043

maurer:
Unifying this should be fine, as our `statx` definition in Android is directly derived from kernel headers.

There are also at least four new members on this struct, but they ate the padding so what's written here is still right.

src/unix/linux_like/mod.rs
neuschaefer: @tgross35 not sure what I previously did wrong, but it _seems_ to work now. I'll update the PR, thanks!

neuschaefer: Ah, the [problems are back](https://github.com/rust-lang/libc/actions/runs/11826922892/job/32953981082?pr=3978), the style check isn't happy:

```
src/unix/linux_like/mod.rs:1892: constant found after extern function when it belongs before
src/unix/linux_like/mod.rs:1933: struct found after constant when it belongs before
src/unix/linux_like/mod.rs:1933: multiple s! macros in one module
```

tgross35: I wound up disabling that `multiple s!` check because it's kind of false-positive-y, so this should be good with a rebase after https://github.com/rust-lang/libc/pull/4107 lands (~1 hour)

Sorry for all the conflicts, been doing a lot of branch cleanup work.

neuschaefer: However, the other two ordering errors still occur. Perhaps I could move the constants/structs to the other constants/structs

tgross35: It is required that things are ordered typedefs->types->functions->consts, it looks like your `cfg_if` block is reversed.

It it still complains after you correct the ordering, I think you might just need to split into three invocations of `cfg_if!` and sort it with the rest of the file.

tgross35 (approved):

neuschaefer:
Hmm,

```json
    "test_tier2": {
      "result": "cancelled",
      "outputs": {}
    },
```

apparently that's what lead to this PR getting kicked out of the merge queue

tgross35:
Probably a timeout, sometimes the android jobs get stuck. You can see the relevant run under “view details” where GH says it removed from the queue.

Added it back.

PR #3937, opened 2024-09-20
Add `fnmatch.h`

epilys:
Add `fnmatch()` function and `FNM_*` constants.

Signed-off-by: Manos Pitsidianakis <manos@pitsidianak.is>

PR #3983, opened 2024-10-17
unbreak OpenBSD after #3937

semarie:
fix `FNM_PATHNAME` and `FNM_NOESCAPE` values.


tgross35 (approved):
Thanks for the quick fix

PR #3989, opened 2024-10-17
RTEMS export arc4random

thesummer:
# Description

Export ´arc4random_buf´ for RTEMS.
Seems to be preferred to the `getentropy` method (see here: https://github.com/rust-lang/rust/pull/131774).

# Sources

https://docs.rtems.org/branches/master/bsp-howto/getentropy.html

# Checklist


tgross35 (approved):
Easy enough!

PR #3991, opened 2024-10-19
haiku: add getentropy

devnexen:
no man page so far but
https://github.com/haiku/haiku/blob/e4a557f372b21b348e0c6a2ae7157c1b73e0d738/src/system/libroot/posix/unistd/getentropy.c#L13

### DIFF
--- a/src/unix/haiku/mod.rs
+++ b/src/unix/haiku/mod.rs
@@ -2093,6 +2093,8 @@ extern "C" {
         length: ::size_t,
         locale: ::locale_t,
     ) -> ::c_int;
+
+    pub fn getentropy(buf: *mut ::c_void, buflen: ::size_t) -> ::c_int;
 }
 
 #[link(name = "bsd")]


tgross35:
Looks straightforward enough to me, @nielx could you double check this?

tgross35 (approved):

nielx:
@devnexen No objections to adding `getentropy()`, however I think the place is incorrect. The commit you cite actually moved this out of libbsd into the general libroot. However, in this change, you place it in the block that links to `bsd`. While this will not cause any problems in practise, I suggest it is placed outside of this block to match Haiku.

nielx:
@tgross35 looks good to me!

tgross35:
Thanks for reviewing! I am just holding off on merging anything until FreeBSD comes back (rustc broke, should be fixed in this nightly or the next one)

devnexen:
> Thanks for reviewing! I am just holding off on merging anything until FreeBSD comes back (rustc broke, should be fixed in this nightly or the next one)
Seems [working now](https://github.com/rust-lang/libc/pull/3996/commits/ca2f10910879548cd3fcd2c4ef4a235b08ccaed4).

PR #3992, opened 2024-10-19
hurd: fix definition of utsname struct

pinotree:
- drop the `domainname` field, as it is not actually used
- add a private `_UTSNAME_LENGTH` constant matching the helper libc one, to ease declaring the struct
- bump the size of the other fields to `_UTSNAME_LENGTH`

### DIFF
--- a/src/unix/hurd/mod.rs
+++ b/src/unix/hurd/mod.rs
@@ -872,12 +872,11 @@ s! {
     }
 
     pub struct utsname {
-        pub sysname: [::c_char; 65],
-        pub nodename: [::c_char; 65],
-        pub release: [::c_char; 65],
-        pub version: [::c_char; 65],
-        pub machine: [::c_char; 65],
-        pub domainname: [::c_char; 65]
+        pub sysname: [::c_char; _UTSNAME_LENGTH],
+        pub nodename: [::c_char; _UTSNAME_LENGTH],
+        pub release: [::c_char; _UTSNAME_LENGTH],
+        pub version: [::c_char; _UTSNAME_LENGTH],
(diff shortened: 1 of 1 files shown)


pinotree:
@sthibaul this should fix crashes when using this struct (by calling `uname()`); because of this, I'm nominating this for a backport in `libc-0.2`, even if "breaking".

pinotree:
For reference:
- `sys/utsname.h`: https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/sys/utsname.h;h=c6884255e04595996dbbc337ac667d0defd035f3;hb=ab564362d0470d10947c24155ec048c4e14a009d
  - the general definition of the `utsname` struct
- `bits/utsname.h`: https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/utsname.h;h=c5c57e840397d9efa6ad6a3b614569d094d26c21;hb=ab564362d0470d10947c24155ec048c4e14a009d
  - the actual bits used on Hurd, setting 1024 as fields size and not enabling `domainname`

sthibaul:
agreed!

tgross35:
Thanks for the sources, agreed that we can backport this (thanks for the ack @sthibaul).

Could you add `const _UTSNAME_LENGTH: usize = 1024;` and use that to make it a bit more consistent and greppable?

pinotree:
> Could you add `const _UTSNAME_LENGTH: usize = 1024;` and use that to make it a bit more consistent and greppable?

Makes sense, done so -- thanks for the suggestion!

tgross35 (approved):
Thanks for the fix!

PR #3993, opened 2024-10-21
Update esp-idf time flag

SergioGasquez:
# Description

This PR updates the ESP-IDF time flag, instead of having a `espidf_time64`, we will have a `espidf_time32` flag. `espidf_time64` was enabling support for ESP-IDF 5 (latest stable), while `espidf_time32` enables support for ESP-IDF 4 (which is now end of life). So making ESP-IDF 4 the opt-in and having ESP-IDF 5 enabled by default makes sense.

See https://github.com/esp-rs/esp-idf-sys/discussions/306#discussioncomment-9537457 for more information.

cc: @ivmarkov 

# Checklist

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -7,7 +7,7 @@ use std::str;
 // make sure to add it to this list as well.
 const ALLOWED_CFGS: &'static [&'static str] = &[
     "emscripten_new_stat_abi",
-    "espidf_time64",
+    "espidf_time32",
     "freebsd10",
     "freebsd11",
     "freebsd12",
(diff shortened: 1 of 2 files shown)


tgross35 (approved):
The change makes sense. Do you want this in 0.2 stable, considering it is a breaking change?

SergioGasquez:
> Do you want this in 0.2 stable, considering it is a breaking change?

Yes, our idea is to get it into 0.2 stable. 

cc: @ivmarkov

PR #3848, opened 2024-08-17
Add ucontext_t for freebsd-aarch64

nathaniel-bennett:
Resolves #3217

PR #3994, opened 2024-10-21
Fix the build on armv7-unknown-freebsd

asomers:
PR #3848 broke the build on avm7-unknown-freebsd by defining a field to be of an unknown type.  Use the correct type name `usize` instead of `::size_t`.


asomers:
cc @tgross35

tgross35 (approved):
Thanks for responding and for the fix.

PR #3996, opened 2024-10-23
freebsd: add CLOSE_RANGE_CLOEXEC flag

devnexen:
[reference](https://github.com/freebsd/freebsd-src/blob/91b7b225b2ce577381aae989c361b9b5b6dd1500/sys/sys/unistd.h#L210)


tgross35 (approved):
Could you squash please?

PR #3997, opened 2024-10-25
Add i686-unknown-freebsd to CI

asomers:
Add i686-unknown-freebsd to CI.  Run it using 32-bit emulation in a 64-bit environment, with the nightly compiler only.
    
So as to avoid a repeat of https://github.com/rust-lang/rust/issues/130677


tgross35:
It looks like the freebsd CI is failing on all PRs. Any idea why?

I wonder if something broke in a recent nightly.

PR #4004, opened 2024-11-06
Fix definition of FNM_CASEFOLD for Illumos/Solaris

psumbera:
Illumos (Solaris) reference: https://github.com/illumos/illumos-gate/blob/aaceae985c2e78cadef76bf0b7b50ed887ccb3a6/usr/src/head/fnmatch.h#L41

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -321,8 +321,18 @@ pub const ATF_PUBL: ::c_int = 0x08;
 pub const ATF_USETRAILERS: ::c_int = 0x10;
 
 pub const FNM_PERIOD: c_int = 1 << 2;
-pub const FNM_CASEFOLD: c_int = 1 << 4;
 pub const FNM_NOMATCH: c_int = 1;
+
+cfg_if! {
+    if #[cfg(any(
+        target_os = "illumos",
+        target_os = "solaris",
+    ))] {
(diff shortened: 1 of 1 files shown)


src/unix/mod.rs
```diff
@@ -321,9 +321,19 @@ pub const ATF_PUBL: ::c_int = 0x08;
...
 pub const FNM_PERIOD: c_int = 1 << 2;
-pub const FNM_CASEFOLD: c_int = 1 << 4;
 pub const FNM_NOMATCH: c_int = 1;
 
+cfg_if! {
+    if #[cfg(any(
+        target_os = "illumos",
+        target_os = "solaris",
+    ))] {
+        pub const FNM_CASEFOLD: c_int = 0x08;
```
tgross35, line 331:
```rust
        pub const FNM_CASEFOLD: c_int = 1 << 3;
```

To be consistent, unless the headers say differently (got a link by the way?)

psumbera: Illumos (Solaris) reference: https://github.com/illumos/illumos-gate/blob/aaceae985c2e78cadef76bf0b7b50ed887ccb3a6/usr/src/head/fnmatch.h#L41

tgross35: Thanks! Added to the PR description so it gets in history.

tgross35 (approved):

PR #4010, opened 2024-11-06

sunshowers:


sunshowers:
Oh sorry, meant to open this against my own branch. Closing for now.

tgross35:
Fyi there was a work in progress to add testing for another OS via `vmactions`, which also has Illumos. That would probably be preferable to adding another CI provider https://github.com/rust-lang/libc/pull/3965

tgross35:
I stand corrected, they don't have Illumos. Does Cirrus? As I understand it, they have everything that is available via GCP

sunshowers:
Thanks, I'll look at that. OmniOS might work!

tgross35:
You know, we could use a target maintainer for Illumos (or at least a platform support page). It seems like you have been working with the platform a lot, any chance you would be willing to sign up for that? It's the zero commitment, just means that we have somebody specific to ping in the rare cases there are platform-specific questions.

If yes, just submit a rust-lang/rust PR. Copy TEMPLATE in https://github.com/rust-lang/rust/tree/master/src/doc/rustc/src/platform-support to illumos.md, update it (doesn't have to be very detailed), then update the table.

Solaris could use a platform page as well in case there is overlap there.

Cc @jclulow @pfmooney in case you might also be okay with being listed for illumos and/or solaris.

jclulow:
> Thanks, I'll look at that. OmniOS might work!

We'll want to be careful using OmniOS, as they ship some extensions that are not in the base illumos bits.

jclulow:
> You know, we could use a target maintainer for Illumos (or at least a platform support page). It seems like you have been working with the platform a lot, any chance you would be willing to sign up for that? It's the zero commitment, just means that we have somebody specific to ping in the rare cases there are platform-specific questions.
> 
> If yes, just submit a rust-lang/rust PR. Copy TEMPLATE in https://github.com/rust-lang/rust/tree/master/src/doc/rustc/src/platform-support to illumos.md, update it (doesn't have to be very detailed), then update the table.
> 
> Cc @jclulow @pfmooney in case you might also be okay with being listed for illumos and/or solaris.

Patrick and I have been acting in this capacity for years, but were only recently informed about the maintainer page stuff.  There's a PR open right now to add a page: rust-lang/rust#132477

> Solaris could use a platform page as well in case there is overlap there.

We've been separate OSes for 14 years, so there's not a great deal of overlap per se.  I suspect folks from the Solaris team at Oracle would be best placed to handle their own page.

tgross35:
> We'll want to be careful using OmniOS, as they ship some extensions that are not in the base illumos bits.

It looks like OmniOS was added by request https://github.com/vmactions/solaris-vm/issues/34 maybe worth asking for vanilla Illumos for the ecosystem as a whole (even if we wind up not using it here).

> Patrick and I have been acting in this capacity for years, but were only recently informed about the maintainer page stuff. There's a PR open right now to add a page: [rust-lang/rust#132477](https://github.com/rust-lang/rust/pull/132477)

Timing :) glad to hear it is in the works, thanks for the link! I think we really only started trying to fill out those pages in the past few months, most didn't exist until the past year.

> > Solaris could use a platform page as well in case there is overlap there.
> 
> We've been separate OSes for 14 years, so there's not a great deal of overlap per se. I suspect folks from the Solaris team at Oracle would be best placed to handle their own page.

👍 I just know the changes often come together to this repo but didn't know the background, thanks for the background.

sunshowers:
We've had a closely look, and unfortunately OmniOS is different enough from upstream illumos that we don't feel confident. (For example, OmniOS has an inotify implementation that upstream doesn't, and it would be pretty easy to accidentally introduce inotify-related symbols to the illumos build.)

We'd like to not just throw illumos CI over the wall, and instead would prefer to fully support it and address issues in a timely manner. Buildomat is the system we use for our own CI at Oxide, so using that aligns with us.

Based on this, we believe that Buildomat is the way to go. What do you think?

sunshowers:
Put up a PR using Buildomat: https://github.com/rust-lang/libc/pull/4023

PR #3999, opened 2024-10-28
[illumos] add some recently-added constants

sunshowers:
# Description

Constants added from the commit linked below.

# Sources

https://github.com/illumos/illumos-gate/commit/0250c53ad267726f2438e3c6556199a0bbf588a2 -- this commit to illumos-gate also includes documentation.

# Checklist

- [x] Relevant tests in `libc-test/semver` have been updated
- [x] No placeholder or unstable values like `*LAST` or `*MAX` are
  included (see [#3131](https://github.com/rust-lang/libc/issues/3131))
- [x] Tested locally (`cd libc-test && cargo test --target mytarget`);
  especially relevant for platforms that may not be checked in CI

PR #4017, opened 2024-11-07
[0.2] Backports

tgross35:
Backport the following Solaris changes:

- https://github.com/rust-lang/libc/pull/3864
- https://github.com/rust-lang/libc/pull/3999
- https://github.com/rust-lang/libc/pull/4004

Also cherry picks the backport of https://github.com/rust-lang/libc/pull/3634 so this can start before https://github.com/rust-lang/libc/pull/4016 merges.


tgross35:
Including https://github.com/rust-lang/libc/pull/4019 so this can merge without conflict

tgross35:
It was not mentioned specifically but Solaris does not have `flock`, so having that exposed was a bug. The relevant PR is https://github.com/rust-lang/libc/pull/3864.

pfmooney:
But removing the `flock`-related defines from illumos would still be breaking.

tgross35:
They should not have been removed from illumos right? https://github.com/rust-lang/libc/blob/0f9f8c91adb0365c0d13b34ec27ac8019818cf1f/src/unix/mod.rs#L1573

pfmooney:
Sorry, I misread and through the `LOCK_`* bits had gotten lost in the shuffle.

PR #4022, opened 2024-11-10
Add missing sysctl net types for mac

GuillaumeGomez:
Thanks to https://github.com/GuillaumeGomez/sysinfo/issues/1378, I discovered that mac's API was actually 32 bits for network information, even if it's not obvious. To go around this limitation, we need to use a different kind of `sysctl` value and `struct`. Adding them here.

## Sources

- https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/net/if_mib.h
- https://fergofrog.com/code/codebrowser/xnu/bsd/net/if_mib.h.html

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -1218,6 +1218,51 @@ s! {
     pub struct host_cpu_load_info {
         pub cpu_ticks: [::natural_t; CPU_STATE_MAX as usize],
     }
+
+    // net/if_mib.h
+    pub struct ifmibdata {
+        /// Name of interface
+        pub ifmd_name: [::c_char; ::IFNAMSIZ],
+        /// Number of promiscuous listeners
+        pub ifmd_pcount: ::c_uint,
+        /// Interface flags
+        pub ifmd_flags: ::c_uint,
+        /// Instantaneous length of send queue
+        pub ifmd_snd_len: ::c_uint,
+        /// Maximum length of send queue
+        pub ifmd_snd_maxlen: ::c_uint,
+        /// Number of drops in send queue
+        pub ifmd_snd_drops: ::c_uint,
+        /// For future expansion
+        pub ifmd_filler: [::c_uint; 4],
+        /// Generic information and statistics
+        pub ifmd_data: if_data64,
+    }
+
+    pub struct ifs_iso_8802_3 {
+        pub dot3StatsAlignmentErrors: u32,
+        pub dot3StatsFCSErrors: u32,
+        pub dot3StatsSingleCollisionFrames: u32,
+        pub dot3StatsMultipleCollisionFrames: u32,
+        pub dot3StatsSQETestErrors: u32,
+        pub dot3StatsDeferredTransmissions: u32,
+        pub dot3StatsLateCollisions: u32,
+        pub dot3StatsExcessiveCollisions: u32,
+        pub dot3StatsInternalMacTransmitErrors: u32,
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Ah I see, the `net/if_mib.h` is not included for mac's `libc-test`. Fixing that.

GuillaumeGomez:
And implemented traits on the new types as well.

GuillaumeGomez:
CI passed. \o/

src/unix/bsd/apple/mod.rs
```diff
@@ -1580,6 +1580,50 @@ s_no_extra_traits! {
         pub ifr_name: [::c_char; ::IFNAMSIZ],
         pub ifr_ifru: __c_anonymous_ifr_ifru6,
     }
+
+    pub struct ifmibdata {
```
tgross35, line 1584: Could these just go in `s!` rather than `s_no_extra_traits!` so the trait implementations don't need to be handwritten?

tgross35 (commented):
One cleanup suggestion but otherwise lgtm. Also added source links to the PR description.

src/unix/bsd/apple/mod.rs
GuillaumeGomez: Ah I was wondering why I needed to implement these traits myself. Completely forgot about the macro... ^^'

GuillaumeGomez:
Not sure to understand why linux is failing considering I didn't change linux's code... Weird `rustc-llvm` error.

tgross35 (approved):
Yeah the CI failure is unrelated https://github.com/rust-lang/rust/issues/133035, I'll merge after the next nightly drops.

tgross35:
Hey @GuillaumeGomez where did the definition for `if_family_id` come from? I only found it at https://fergofrog.com/code/codebrowser/xnu/bsd/net/if_mib.h.html but it doesn't seem to exist in xnu https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/net/if_mib.h. These tests are failing on my machine, I'm not sure how they are passing in CI.

GuillaumeGomez:
It's the second header: https://fergofrog.com/code/codebrowser/xnu/bsd/net/if_mib.h.html#if_family_id

tgross35:
I know that site shows it, but where does that site's source come from? It's not in the official xnu repo

GuillaumeGomez:
Absolutely no clue. Seems to be a common header for BSD-based systems from what I can understand. It's horribly difficult to find apple's headers. When I tested locally on my (2015) macbook, there was no issue so I assumed it was the right definition.

tgross35:
Mine is a bit newer, I have the header but I don't have that specific struct (matches xnu). Maybe it existed at some point but was removed? For reference `﻿﻿xcodebuild -showsdks` shows

```
macOS SDKs:
        macOS 15.0                      -sdk macosx15.0
        macOS 15.0                      -sdk macosx15.0
```

It's not a problem to have if they're useful, just wondering if there is a test config that should be set so mine don't fail :)

GuillaumeGomez:
Ah. If it was removed then we have a big issue. OS versions don't mix well at all with APIs.

tgross35:
It looks like CI has the macOS 14.5 SDKs, so I'm going to guess that it isn't available in the newest APIs. Did you need `if_family_id` specifically or would you mind if I just removed that type?

GuillaumeGomez:
I don't need it so go ahead. :)

PR #4023, opened 2024-11-11

sunshowers:
Add continuous integration for the x86_64-unknown-illumos target using the [buildomat] CI service, and fix up issues such that CI passes.

## What is buildomat?

Buildomat is a CI service, similar in spirit to services like GitHub Actions. Buildomat is open source, built and maintained by us at Oxide, and is used by our own CI.

Here's an example run with this PR:

* [GitHub](https://github.com/sunshowers/libc/runs/32829121549)
* [Buildomat](https://buildomat.eng.oxide.computer/wg/0/details/01JCEH2XM6JWQMFZRWP20XM7D8/NBYYTAsj3lylUuK0VrofLaby5WaNremha5hZrvuWRjw7fYqY/01JCEH37JR7KQXJZMVFX1E6KN1)

As part of this, at Oxide we're committing to maintaining the buildomat support. This includes:

* Providing compute resources.
* Maintaining and fixing issues in a timely manner.

In order to have buildomat CI working, the rust-lang project would need to [install the buildomat app] (allowlisted to libc if desired).

## Why another CI system?

The main alternative to buildomat would be something like [vmactions].

While vmactions provides an OmniOS build, OmniOS is actually a downstream distribution of illumos that includes several APIs not present in illumos. For example, OmniOS has an [inotify implementation] but upstream illumos doesn't. It would be easy to accidentally add inotify APIs to libc even though that wouldn't build on other illumos distributions.

More importantly, at Oxide we don't feel confident providing support for a third-party service. We'd like to not just throw CI support over the wall but instead are committed to the long-term health of Rust on illumos. Since we use Buildomat ourselves, we're strongly incentivized to maintain it in a way that wouldn't be true for other solutions.

[Buildomat]: https://github.com/oxidecomputer/buildomat
[install the buildomat app]: https://github.com/apps/buildomat
[vmactions]: https://github.com/vmactions
[inotify implementation]: https://man.omnios.org/7/inotify

### DIFF
--- a/ci/verify-build.sh
+++ b/ci/verify-build.sh
@@ -69,6 +69,11 @@ test_target() {
     $cmd --no-default-features
     $cmd --no-default-features --features extra_traits
 
+    # Ensure the crate will build when used with `std`
+    if [ "$rust" = "nightly" ]; then
+        $cmd --no-default-features --features rustc-dep-of-std
+    fi
+
     # For tier 2 freebsd targets, check with the different versions we support
     # if on nightly or stable
     case "$rust-$target" in
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,6 +20,7 @@
 #![cfg_attr(libc_deny_warnings, deny(warnings))]
 // Attributes needed when building as part of the standard library
 #![cfg_attr(feature = "rustc-dep-of-std", feature(link_cfg, no_core))]
+#![cfg_attr(feature = "rustc-dep-of-std", allow(internal_features))]
 // Enable extra lints:
 #![cfg_attr(feature = "extra_traits", deny(missing_debug_implementations))]
 #![deny(missing_copy_implementations, safe_packed_borrows)]
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -70,13 +70,13 @@ macro_rules! prelude {
         mod prelude {
             // Exports from `core`
             #[allow(unused_imports)]
-            pub(crate) use core::clone::Clone;
+            pub(crate) use ::core::clone::Clone;
             #[allow(unused_imports)]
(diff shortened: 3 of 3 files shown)


tgross35:
Thanks for the PR, I'll ask infra about installing the app.

jclulow:
> Thanks for the PR, I'll ask infra about installing the app.

Thanks!  I'm the lead on buildomat at Oxide, and am happy to join a meeting or a chat thing or answer any questions folks might have!

tgross35:
I don't expect many concerns but the thread is here https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Buildomat.20app.20for.20.60libc.60.

I guess I have a question though, is there a list of which platforms are supported? I'm wondering if we could use buildomat for more T2 targets that aren't supported by GitHub.

jclulow:
> I don't expect many concerns but the thread is here https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Buildomat.20app.20for.20.60libc.60.

Awesome.  I am not extremely Zulip-literate but I believe I have subscribed to it!

> I guess I have a question though, is there a list of which platforms are supported? I'm wondering if we could use buildomat for more T2 targets that aren't supported by GitHub.

As a niche OS person I am always happy to help out where I can!  The software is all Rust, at least, so if it's a platform that supports Rust already it's probably going to be pretty easy to get _something_ going.  What sort of platforms are you considering?

sunshowers:
Hmm, didn't mean to close this, sorry about the accidental push. Will put up again.

sunshowers:
Re-done (and also updated against main) in https://github.com/rust-lang/libc/pull/4172

PR #4028, opened 2024-11-12
musl: add more syscall constants

wolfv:
I hit a compilation issue where the `SYS_rseq` constant is not defined for some musl targets. For this PR I compared with upstream `syscall.h.in` files and added constants that I found where missing.

I wanted to double check if this is something we can easily adjust. If a PR like this is accepted, I can also see to adjust more of the remaining files that I haven't touched yet, and potentially write a small script for the conversion. 

The changes in s390x are a little bigger because the integers were in relatively random order.

References:

- [`riscv64`](https://github.com/kraj/musl/blob/ffb23aef7b5339b8c3234f4c6a93c488dc873919/arch/riscv64/bits/syscall.h.in)
- [`x86_64`](https://github.com/kraj/musl/blob/ffb23aef7b5339b8c3234f4c6a93c488dc873919/arch/x86_64/bits/syscall.h.in)
- [`powerpc64`](https://github.com/kraj/musl/blob/ffb23aef7b5339b8c3234f4c6a93c488dc873919/arch/powerpc64/bits/syscall.h.in)
- [`aarch64`](https://github.com/kraj/musl/blob/ffb23aef7b5339b8c3234f4c6a93c488dc873919/arch/aarch64/bits/syscall.h.in)
- [`s390x`](https://github.com/kraj/musl/blob/ffb23aef7b5339b8c3234f4c6a93c488dc873919/arch/s390x/bits/syscall.h.in)


tgross35 (approved):
LGTM, thanks! Current nightly has a bug on s390x & sparc so I can't merge now, should be fixed tomorrow.

tgross35 (approved):

wolfv:
@tgross35 I can come up with a Python script to make these lists more consistent either with MUSL upstream or the Linux sources upstream. What do you think about that?

tgross35:
> @tgross35 I can come up with a Python script to make these lists more consistent either with MUSL upstream or the Linux sources upstream. What do you think about that?

Thanks for offering! I am in favor of better comparison against sources but I'm not sure what would be best here. A script to compare syscalls would be somewhat limited, but it would be great to do something like compare against bindgen's output.

Happy to discuss something more in an issue if you're up for helping out, our tests could definitely be improved in a variety of ways.

wolfv:
@tgross35 I had a Python script that looked like this and used the `musl` source code, but I am wondering if it would be more appropriate to use the Linux headers? I guess these files are the definite source: https://github.com/torvalds/linux/blob/4a5df37964673effcd9f84041f7423206a5ae5f2/arch/x86/entry/syscalls/syscall_64.tbl

```py
import re
from pathlib import Path

def convert_to_rust(input_file, output_file):
    # Updated pattern to capture the entire value/expression
    pattern = r'#define\s+__NR_(\w+)\s+(\(?\s*(?:0x[0-9a-fA-F]+)?\s*\+?\s*\d+\s*\)?|\d+)'

    # Read input file and convert to Rust format
    with open(input_file, 'r') as infile, open(output_file, 'w') as outfile:
        for line in infile:
            match = re.match(pattern, line.strip())
            if match:
                syscall_name, value = match.groups()
                # Keep the value exactly as is, just remove any extra whitespace
                value = ''.join(value.split())
                # Convert snake_case to the Rust SYS_ format
                rust_line = f'pub const SYS_{syscall_name}: ::c_long = {value};\n'
                outfile.write(rust_line)


def run(input_file):
    print("Running for file: ", input_file)
    output_file = input_file.parent / 'syscalls.rs'
    convert_to_rust(input_file, output_file)
    print(f"Successfully converted {input_file} to {output_file}")
    print("===============================\n\n\n")

    
def main():
    paths = Path('.').rglob('**/syscall.h.in')
    for path in paths:
        print(path)
        run(path)


if __name__ == '__main__':
    main()
```

PR #4029, opened 2024-11-12
hurd: cleanups

sthibaul:
# Description

This cleans a couple of details in the hurd port: ssize_t and align are currently not actually used

# Checklist

### DIFF
--- a/src/unix/hurd/align.rs
+++ b/src/unix/hurd/align.rs
@@ -1,1 +1,0 @@
-// Placeholder file
--- a/src/unix/hurd/mod.rs
+++ b/src/unix/hurd/mod.rs
@@ -72,7 +72,6 @@ pub type __ptrdiff_t = __sword_type;
 pub type __socklen_t = __u32_type;
 pub type __sig_atomic_t = ::c_int;
 pub type __time64_t = __int64_t;
-pub type ssize_t = __ssize_t;
 pub type wchar_t = ::c_int;
 pub type wint_t = ::c_uint;
 pub type gid_t = __gid_t;
--- a/src/unix/hurd/mod.rs
+++ b/src/unix/hurd/mod.rs
@@ -4671,9 +4671,6 @@ safe_f! {
     }
 }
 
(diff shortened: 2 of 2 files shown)


tgross35:
Could this also remove `align.rs`? Also, isn't `ssize_t` publicly exported?

sthibaul:
> Could this also remove align.rs?

Sure, done so.

> Also, isn't ssize_t publicly exported?

I have to admit I am completely at a loss at what platforms are supposed to export, I have not found a documentation that explains how this is organized. The Linux platform doesn't seem to be exporting `ssize_t`, so I assumed that it's not actually being in use in the wild? We don't use it for hurd-specific code at least.

tgross35:
`ssize_t` is actually exported in unix https://github.com/rust-lang/libc/blob/db364574e79f74eaa8faf60e8d2982a5796ccab9/src/unix/mod.rs#L23 (I am surprised this wasn't already conflicting somehow). I'm not positive about the exact rules either but it seems like currently the typedef is `ssize_t` ->` __ssize_t` -> `__sword_type` -> `c_int` on 32-bit or `c_long`  on 64-bit.

So AIUI, this change is making `ssize_t` always `isize` (signed pointer-sized type) rather than being manually defined to `i32`/`i64`. This seems preferable to me since it will be consistent with other platforms - as long as that seems reasonable to you then I think we can merge this.

(cc @sunfishcode since I see you crosslinked an issue)

sthibaul:
> as long as that seems reasonable to you

Yes, this is reasonable, GNU/Hurd has nothing specific here compared to other unices.

sthibaul:
Since it's needed for rustix, a backport to 0.2 will be useful.

tgross35 (approved):

tgross35:
SGTM, thanks for the follow up

> Since it's needed for rustix, a backport to 0.2 will be useful.

Feel free to just nominate anything that makes sense for a backport, the rustbot command works for everyone.

PR #4034, opened 2024-11-13
fix: value of `PTHREAD_MUTEX_DEFAULT` on illumos

Berrysoft:
# Description

The value on illumos is different to the one on solaris.

# Sources

https://github.com/illumos/illumos-gate/blob/master/usr/src/head/pthread.h

# Checklist


Berrysoft:
Fair enough:)

PR #4038, opened 2024-11-15
Use `#[derive]` for `Copy`/`Clone` in `s!` and friends

bossmc:
# Description

Extracted from #3261 as a pre-requisite change.

Allows use of `#[cfg]` to skip items in an `s!` block without creating invalid, orphaned `impl` blocks.

# Checklist

- [ ] ~~Relevant tests in `libc-test/semver` have been updated~~
- [ ] ~~No placeholder or unstable values like `*LAST` or `*MAX` are included (see [#3131](https://github.com/rust-lang/libc/issues/3131))~~

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -4503,11 +4503,11 @@ fn test_linux(target: &str) {
             true
         }
 
-        // The `inotify_event` and `cmsghdr` types contain Flexible Array Member fields (the
-        // `name` and `data` fields respectively) which have unspecified calling convention.
-        // The roundtripping tests deliberately pass the structs by value to check "by value"
-        // layout consistency, but this would be UB for the these types.
+        // The following types contain Flexible Array Member fields which have unspecified calling
+        // convention. The roundtripping tests deliberately pass the structs by value to check "by
+        // value" layout consistency, but this would be UB for the these types.
         "inotify_event" => true,
+        "fanotify_event_info_fid" => true,
         "cmsghdr" => true,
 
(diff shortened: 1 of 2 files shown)


tgross35 (approved):
Ideally I'd like to backport this but that can't happen until we bump the MSRV of libc-0.2 (hopefully soon).

Also can't merge this right away due to a CI failure in recent nightly, LGTM though.

@rustbot label +stable-blocked

bossmc:
How does this change impact the MSRV?  (I tried building with the currently declared 1.13.0 but that is _completely_ broken, it can't even resolve the Cargo.toml since some versions of the `cc` crate have a feature named after a dependency)

tgross35:
For 0.2 we test down to 1.19 - do the `derive`s work on 1.19? If so then we can backport with no problem, I was just assuming they wouldn't since we don't have this already, but could be wrong there.

There is no problem for `main` since that only goes down to 1.63.

bossmc:
Deriving `Copy` and `Clone` was present in Rust 1.0, my theory for why they're manually implemented in `s!` is to ensure that the generated code for `Clone` is a memcpy even in debug builds (the unoptimized implementation of `Clone`  is a recursive call into `Clone` for each field, even if those fields are all `Copy`).  As discussed in the other MR, if that is a genuine problem for developers they can chose to compile `libc` with optimizations even if they're compiling the rest of their crate(s) in debug mode.

I'm pretty sure this change can be ported to the 0.2 branch with the above caveat if you wished.

PR #4043, opened 2024-11-15
flock for vxworks using ioctl

jackorobot:
# Description

Added an implementation for `flock` for vxworks target. Vxworks only supplies `ioctl`, but it does have all `flock` related `#define`s available.

# Sources

Vxworks headers are copyrighted, however this is based on: `ioLib.h` and `sys/fcntlcom.h`

# Checklist

Note that testing seems to not be available for the vxworks target

### DIFF
--- a/src/vxworks/mod.rs
+++ b/src/vxworks/mod.rs
@@ -413,6 +413,14 @@ s! {
         pub mq_flags:   ::c_long,
         pub mq_curmsgs: ::c_long,
     }
+
+    pub struct flock {
+        pub l_type: ::c_short,
+        pub l_whence: ::c_short,
+        pub l_start: ::c_long,
+        pub l_len: ::c_long,
+        pub l_pid: ::c_long,
+    }
 }
 
 s_no_extra_traits! {
@@ -944,6 +952,9 @@ pub const F_GETLK: ::c_int = 7;
 pub const F_SETLK: ::c_int = 8;
 pub const F_SETLKW: ::c_int = 9;
 pub const F_DUPFD_CLOEXEC: ::c_int = 14;
+pub const F_RDLCK: ::c_int = 1;
+pub const F_WRLCK: ::c_int = 2;
+pub const F_UNLCK: ::c_int = 3;
 
 // signal.h
 pub const SIG_DFL: sighandler_t = 0 as sighandler_t;


src/vxworks/mod.rs
```diff
@@ -2013,6 +2029,35 @@ pub unsafe fn posix_memalign(
     }
 }
 
+pub unsafe fn flock(fd: ::c_int, operation: ::c_int) -> ::c_int {
```
tgross35, line 2032: We prefer not to have any polyfill implementations in `libc`, only interfaces as much as possible. It would be best to drop this function from `libc` and instead add this implementation wherever this would be called - for `std` this is at https://github.com/rust-lang/rust/blob/76fd47124b8f8037b6187169b2cdf39139466952/library/std/src/sys/pal/unix/fs.rs#L1268-L1276.

tgross35 (changes requested):
I don't think we can add the polyfill function but everything else looks fine once the public fields are added. Cc @biabbas for vxworks review

src/vxworks/mod.rs
jackorobot: Thanks for informing me about that preference. I understand you want to keep this library mostly to bindings. I will propose an update to the `std` library for a similar implementation using the contents of this merge request.

tgross35 (approved):
Thanks for the changes, this looks good to me. I'll give another day or so for someone more familiar with vxworks to take a look before merging.

Please squash when you get the chance.

src/vxworks/mod.rs
```diff
@@ -412,6 +412,15 @@ s! {
         pub mq_flags:   ::c_long,
         pub mq_curmsgs: ::c_long,
     }
+
+    pub struct flock {
+        pub l_type: ::c_short,
+        pub l_whence: ::c_short,
+        pub l_start: ::c_long,
+        pub l_len: ::c_long,
+        pub l_pid: ::c_short,
```
biabbas, line 421: This is of type long in fnctcom.h

src/vxworks/mod.rs
```diff
@@ -412,6 +412,15 @@ s! {
...
+        pub l_xxx: ::c_short,
```
biabbas, line 422: This is not declared in fnctcom.h.

biabbas (changes requested):
There are five fields in struct flock: 
`l_type`(short), `l_whence`(short), `l_start`(long long), `l_len`(long long) and `l_pid`(long).

Regards,

tgross35:
> I would propose to implement your suggestions regardless, since on versions where only the lower 16-bits are used this should still result in the same behaviour.

Is this saying there is a change to make here? Just `@rustbot review` when it's ready for a review again, either with the change or if I misunderstood the comment.

@rustbot author

src/vxworks/mod.rs
jackorobot: Changed to `::c_long`

biabbas (approved):

tgross35 (approved):

tgross35:
Thanks @jackorobot for authoring and @biabbas for taking a look.

I don't know much about vxworks but is it possible you are looking at headers for 32-bit vs 64-bit? The two mentioned definitions are compatible for the lower four fields on LP64, but not on ILP32.

tgross35:
I'm going to hold off on backporting actually until confirming the above. Would you both mind mentioning what target and vxworks versions the headers you're referencing are from?

If it's a difference in version then we should just add the latest, but I don't want to give 32-bit users unsound API if that's what the difference is.

biabbas:
I referred the latest vxworks sources, I found only one header file with this definition for for all targets. I also checked the previous vxworks 7 versions but the definitions remained same. I'll check again tomorrow and confirm.

biabbas:
>There are five fields in struct flock:
> l_type(short), l_whence(short), l_start(long long), l_len(long long) and l_pid(long).

I checked on different targets with the latest source, the struct flock declaration in fcntlcom.h remained the same.

tgross35:
Thanks for checking, this makes no sense. As a last effort to explain things, is it possible to build a program both for i686 and x86_64 and print that? Something like

```
int main() {
    struct flock f;
    printf(
        "flock %zu, l_type %zu, l_whence %zu, l_start %zu, "
        "l_len %zu, l_pid %zu, long %zu, long long %zu\n",
        sizeof(struct flock),
        sizeof(f.l_type),
        sizeof(f.l_whence),
        sizeof(f.l_start),
        sizeof(f.l_len),
        sizeof(f.l_pid),
        sizeof(long),
        sizeof(long long)
    );
}
```

VxWorks doesn't do API versioning separate from OS version, do they?

In any case I'll backport this as long as neither of you have objections, but will add a `// FIXME(vxworks): ...` since we definitely need to get a better answer here at some point.

biabbas:
@tgross35 , I realize now that long may be 4bytes on 32 bit targets. The same is consistent with vxworks. On 32 bit itl_generic target I get output:
```
[vxWorks *]# ./flock.vxe
flock 24
, l_type 2
, l_whence 2
, l_start 8
, l_len 8
, l_pid 4
, long 4
, long long 8
```
On 64 bits long is 8 bytes. This would cause an issue for 32 bit targets.

biabbas:
@tgross35, 
On 64 bit targets long long and long are both 8 bytes. So our changes here won't cause any problems as c_long is also 8 bytes.
```
[vxWorks *]# ./flock.vxe
flock 32
, l_type 2
, l_whence 2
, l_start 8
, l_len 8
, l_pid 8
, long 8
, long long 8
```
But on 32 bits long long is 8 bytes and long is 4 bytes. c_long is 4 bytes for 32 bit targets. This would indeed cause problems for 32 bit targets.
```
[vxWorks *]# ./flock.vxe
flock 24
, l_type 2
, l_whence 2
, l_start 8
, l_len 8
, l_pid 4
, long 4
, long long 8
```
I don't think we should be releasing these changes.

biabbas:
@jackorobot Could you use `c_longlong` for l_start and l_len.

tgross35:
Are there more specific version numbers? Aiui vxworks 7 has been around since 2010

biabbas:
I have access to almost all VxWorks versions. I checked the early VxWorks 7versions and VxWorks 6 versions too, but the definition was the same as in the latest 24.03 version.
@jackorobot Do you know the version of the VxWorks workbench you are using?

PR #4046, opened 2024-11-15
Add struct and constants for mount_setattr syscall

pcc:
# Description

Adds struct and constants for mount_setattr syscall.

# Sources

https://github.com/torvalds/linux/blob/master/include/uapi/linux/mount.h

# Checklist


tgross35 (approved):

tgross35 (changes requested):
Ah, you might need to update build.rs if this isn't available on musl

tgross35:
@pcc we just updated the linux headers on musl, could you rebase? The tests might start to pass.

pcc:
> @pcc we just updated the linux headers on musl, could you rebase? The tests might start to pass.

Done

tgross35 (approved):
Thanks!

PR #4047, opened 2024-11-15
Add the TCP_FUNCTION_BLK and TCP_FUNCTION_ALIAS socket options

asomers:
For FreeBSD only

# Description

These two socket options are used to get and set a socket's TCP stack, on FreeBSD only.

# Sources

https://man.freebsd.org/cgi/man.cgi?query=tcp
https://github.com/freebsd/freebsd-src/blob/976e1d36be688166e140cbaded01d433c61e6d47/sys/netinet/tcp.h#L229

# Checklist


asomers:
The error looks unrelated to my change.  For one thing, it was a Linux build that failed.  For another, this is the error:
```
rustc-LLVM ERROR: Cannot select: 0x7fe87eb15000: f32,ch = load<(load (s16) from %ir.0), anyext from f16> 0x7fe87eb15a10, 0x7fe87eaf6540, undef:i64, src/fmt/float.rs:238:30
  0x7fe87eaf6540: i64,ch = CopyFromReg 0x7fe87eabb468, Register:i64 %0
    0x7fe87eb15af0: i64 = Register %0
  0x7fe87eb155b0: i64 = undef
In function: _ZN4core3fmt5float50_$LT$impl$u20$core..fmt..Debug$u20$for$u20$f16$GT$3fmt17h9203c7e9bc0be64fE
error: could not compile `core` (lib)
```

tgross35 (approved):
> The error looks unrelated to my change. For one thing, it was a Linux build that failed. For another, this is the error:
> 
> ```
> rustc-LLVM ERROR: Cannot select: 0x7fe87eb15000: f32,ch = load<(load (s16) from %ir.0), anyext from f16> 0x7fe87eb15a10, 0x7fe87eaf6540, undef:i64, src/fmt/float.rs:238:30
>   0x7fe87eaf6540: i64,ch = CopyFromReg 0x7fe87eabb468, Register:i64 %0
>     0x7fe87eb15af0: i64 = Register %0
>   0x7fe87eb155b0: i64 = undef
> In function: _ZN4core3fmt5float50_$LT$impl$u20$core..fmt..Debug$u20$for$u20$f16$GT$3fmt17h9203c7e9bc0be64fE
> error: could not compile `core` (lib)
> ```

That was just a bug on the most recent nightly, I think it should be fixed with the one that just released.

PR #4050, opened 2024-11-16
ci: trim trailing whitespace

tgross35:


tgross35:
Seems like this wound up being fixed somehow.

PR #4051, opened 2024-11-16
[0.2] ci: Remove tests with rust < 1.63

tgross35:
Since the increase to MSRV in [1], we no longer need to test with old
versions of Rust.

[1]: https://github.com/rust-lang/libc/pull/4040


tgross35:
Includes https://github.com/rust-lang/libc/pull/4048 and https://github.com/rust-lang/libc/pull/4050 just so I can get this going without conflicts.

PR #4056, opened 2024-11-16
[0.2] Require rust >= 1.24 and drop libc_const_size_of conditional

tgross35:
[ resolve conflicts - Trevor ]

Applied from https://github.com/rust-lang/libc/pull/2845


tgross35:
This will include https://github.com/rust-lang/libc/pull/4055 to save queuing one item.

PR #4063, opened 2024-11-16
[0.2] Drop the `libc_const_extern_fn` conditional

tgross35:
Additionally deprecate the `const-extern-fn` feature. This is possible since the MSRV was increased to 1.63.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ Raw FFI bindings to platform libraries like libc.
 """
 
 [package.metadata.docs.rs]
-features = ["const-extern-fn", "extra_traits"]
+features = ["extra_traits"]
 default-target = "x86_64-unknown-linux-gnu"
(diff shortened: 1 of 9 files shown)


SteveLauC:
Hi, Looks like this is not working? See Nix PR https://github.com/nix-rust/nix/pull/2544, we tried [using the git version of libc](https://github.com/nix-rust/nix/pull/2544#discussion_r1849629364), then found that the `CMSG_SPACE` is not const with the git version of libc (See the CI errors)

tgross35:
Indeed, that was reported in https://github.com/rust-lang/libc/issues/4115. Will try to fix it soon (I think I want to just remove the feature and instead use an environment variable to indicate we need ctest hacks)

SteveLauC:
Could you please cc me when you have this fixed 🫣

PR #4074, opened 2024-11-17
[0.2] Eliminate uses of `struct_formatter`

tgross35:
This pattern was previously used with `cfg` fields but is no longer needed.


tgross35:
No port to stable, will handle that separately.

PR #4088, opened 2024-11-17
openbsd: skip ATF_* constants in CI

semarie:
# Description

CI is failing on OpenBSD (due to OpenBSD change).

Some constants (`ATF_*`) were removed in OpenBSD 7.7 by https://github.com/openbsd/src/commit/ff46e7d6ebc305e25e40b31e65a50463cfa5dc30

they were unused since 1991.

# Checklist


tgross35 (approved):

PR #4095, opened 2024-11-18
[0.2]  apple: Mark mach_task_self as deprecated

tgross35:
These were already removed on `main` in https://github.com/rust-lang/libc/commit/56d665c9b3f7687eb206c3a1d2f943852df65f54 ("macOs various updates").


tgross35:
Oh, I think these were in https://github.com/rust-lang/libc/pull/3578 because they are part of the mach2 crate. So I think I should close this and instead mark them `deprecated` on libc-0.2.

@devnexen does that sound right?

devnexen:
yes

PR #4107, opened 2024-11-18
ci: use some tricks to format macro bodies

tgross35:
We have a lot of syntax contained within macro bodies, which `rustfmt` doesn't touch. Add a hack that replaces relevant macro invocations with functions, formats, then resets.


tgross35:
This still isn't ideal, I really wish `rustfmt` handled something like this out of the box even if unstable. In any case, this is significantly nicer than just having the majority of this codebase unformatted because everything needs a macro wrapper.

Sorry everyone, conflicts ahead 🚢

PR #4119, opened 2024-11-20
ci: Change Docker images to ubuntu:24.10

tgross35:
23.10 is EOL so update to the latest stable version.


tgross35:
NOT FUN 24.04 is also when they did the change to 64-bit time_t which we are not ready for...

tgross35:
Superseded by https://github.com/rust-lang/libc/pull/4120

PR #4128, opened 2024-11-20
add `ptp_clock_caps`

folkertdev:
# Description

The final missing pieces of https://github.com/rust-lang/libc/pull/3865

# Sources

# Checklist

r? @tgross35

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -3885,7 +3885,8 @@ fn test_linux(target: &str) {
                 return true;
             }
             // FIXME: Requires >= 5.4 kernel headers
-            if name == "PTP_ENABLE_PPS2" 
+            if name == "PTP_CLOCK_GETCAPS2"
+                || name == "PTP_ENABLE_PPS2" 
                 || name == "PTP_EXTTS_REQUEST2"
                 || name == "PTP_PEROUT_REQUEST2"
                 || name == "PTP_PIN_GETFUNC2"
@@ -4333,7 +4334,11 @@ fn test_linux(target: &str) {
         // `anonymous_1` is an anonymous union
         (struct_ == "ptp_perout_request" && field == "anonymous_1") ||
         // `anonymous_2` is an anonymous union
-        (struct_ == "ptp_perout_request" && field == "anonymous_2")
+        (struct_ == "ptp_perout_request" && field == "anonymous_2") ||
+        // FIXME(linux): `adjust_phase` requires >= 5.7 kernel headers
+        // FIXME(linux): `max_phase_adj` requires >= 5.19 kernel headers
+        // the rsv field shrunk when those fields got added, so is omitted too
+        (struct_ == "ptp_clock_caps" && (loongarch64 || sparc64) && (["adjust_phase", "max_phase_adj", "rsv"].contains(&field)))
     });
 
     cfg.volatile_item(|i| {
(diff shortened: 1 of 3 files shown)


src/unix/linux_like/linux/mod.rs
```diff
@@ -1206,6 +1206,81 @@ s! {
     }
 }
 
+cfg_if! {
```
folkertdev, line 1209: so yeah, this looks like the best I can do...

Defining a `PTP_CLOCK_CAPS_RSV_LEN` const just does not seem to work. not with cfg_if, a vanilly block with an if expression, defining a `const fn` and calling that: they all run into unreachable code in garando.

Next it tusrns out that `loongarch` is weird and special (or just has a later version maybe?), so that further complicates the situation. 

It looks like I can bring it down to three branches by just making the conditionals more complex (that almost adds as many lines as it saves...) Any other ideas?

tgross35: Where do all these definitions even come from? I don't see it in glibc or musl source so I assume it should be uapi source, but I only see the one definition with no conditions https://github.com/torvalds/linux/blob/43fb83c17ba2d63dfb798f0be7453ed55ca3f9c2/include/uapi/linux/ptp_clock.h#L94-L107

src/unix/linux_like/linux/mod.rs
```diff
@@ -1195,6 +1195,19 @@ s! {
         pub rsv: [::c_uint; 5],
     }
 
+    pub struct ptp_clock_caps {
```
tgross35, line 1198: Nit, trying to do a bit more tagging of where everything comes from (for the consts too)

```rust
    // ptp_clock.h
    pub struct ptp_clock_caps {
```

libc-test/build.rs
```diff
@@ -4473,7 +4474,11 @@ fn test_linux(target: &str) {
         // `anonymous_1` is an anonymous union
         (struct_ == "ptp_perout_request" && field == "anonymous_1") ||
         // `anonymous_2` is an anonymous union
-        (struct_ == "ptp_perout_request" && field == "anonymous_2")
+        (struct_ == "ptp_perout_request" && field == "anonymous_2") ||
+        // FIXME: `adjust_phase` requires >= 5.7 kernel headers
+        // FIXME: `max_phase_adj` requires >= 5.19 kernel headers
```
tgross35, line 4479: I know we don't do this very many places but I'm trying to make sure all our fixmes are greppable to something specific

```suggestion
        // FIXME(linux): `adjust_phase` requires >= 5.7 kernel headers
        // FIXME(linux): `max_phase_adj` requires >= 5.19 kernel headers
```

libc-test/build.rs
```diff
@@ -4473,7 +4474,11 @@ fn test_linux(target: &str) {
...
+        // the rsv field shrunk when those fields got added, so is omitted too
+        (struct_ == "ptp_clock_caps" && (musl || loongarch64 || sparc64) && (["adjust_phase", "max_phase_adj", "rsv"].contains(&field)))
```
tgross35, line 4481: Musl kernel header updates are in the queue right now https://github.com/rust-lang/libc/pull/3921, mind just waiting until that lands then dropping the `musl`?

src/unix/linux_like/linux/mod.rs
tgross35: _Oh_ all of this was just accounting for what we happen to test in CI rather than actual sources? Why didn't you just say so :) we should definitely only implement the latest and skip tests on platforms where we can explain the failure.

If this explains the platform-specific config for the other PTP structs (I think there was something for `sparc`), mind putting up a PR to drop that as well?

I suggested a few changes here but think this should mostly be good.

src/unix/linux_like/linux/mod.rs
folkertdev: it's on line 1181 already, I think that should do? I've assumed that if the comment is missing it's from the same file as the preceding definition?

libc-test/build.rs
folkertdev: sounds good, I'll squash and rebase in a bit then

src/unix/linux_like/linux/mod.rs
tgross35: Oh I missed that, thanks that's good 👍

tgross35 (approved):

src/unix/linux_like/linux/mod.rs
folkertdev: the rest of the ptp structs are clean (no cfg's or anything like that). For the ioctl const functions there is some powerpc/sparc/mips-specific logic, but that really is because they define constants with different values. There are links to those values there so I think it's all good now.

tgross35: SGTM, I must have been thinking of something else. Thanks for the followup.

folkertdev:
thanks for the help here @tgross35, that was a journey!

PR #4129, opened 2024-11-20
Fix the tests on riscv64gc-unknown-freebsd

asomers:
The fpregs was defined with two fields having the wrong sign, and one field having the wrong name.

# Description

Fixes one struct that was defined incorrectly for FreeBSD riscv64

# Sources

https://github.com/freebsd/freebsd-src/blob/7a3af393d8ac2dfe72d24fe401344b60c4f87866/sys/riscv/include/ucontext.h#L50

# Checklist


tgross35 (approved):

asomers:
@tgross35 tests pass.  Go ahead and merge to libc-0.2.

tgross35:
Thanks, I'll get it in the next round.

Issue #4061, opened 2024-11-16
Do something about the `sparc` CI job


tgross35:
Most CI jobs take about 5 minutes but `sparc64-unknown-linux-gnu` takes about 25 minutes. I think this is because it is the only job that uses qemu-system vs qemu-user.

It would be nice to somehow get this job's runtime closer to the other jobs. It might be possible to replace qemu-system somehow, or even just do something like only run the job once the PR is in the merge queue (i.e. don't require sparc to complete in PR CI).

folkertdev:
to try to debug this, it would be really useful if the script could be run locally, but I'm running into compile failures

```
>  sudo sh ci/run-docker.sh sparc64-unknown-linux-gnu
... stuff ...
+ docker run --rm --user 0:0 --env LIBC_CI --env LIBC_CI_ZBUILD_STD --env CARGO_HOME=/cargo --env CARGO_TARGET_DIR=/checkout/target --volume .:/cargo --volume /home/folkertdev/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu:/rust:ro --volume /home/folkertdev/rust/libc:/checkout:ro --volume /home/folkertdev/rust/libc/target:/checkout/target --init --workdir /checkout libc-sparc64-unknown-linux-gnu sh -c HOME=/tmp PATH=$PATH:/rust/bin exec ci/run.sh sparc64-unknown-linux-gnu
+ mirrors_url=https://ci-mirrors.rust-lang.org/libc
+ target=sparc64-unknown-linux-gnu
+ [ -n  ]
+ [ sparc64-unknown-linux-gnu = s390x-unknown-linux-gnu ]
+ cargo test --no-default-features --manifest-path libc-test/Cargo.toml --target sparc64-unknown-linux-gnu
   Compiling quote v1.0.21
   Compiling dirs-sys v0.3.7
   Compiling rustc_version v0.4.0
   Compiling cc v1.0.83
   Compiling libc v1.0.0-alpha.1 (/checkout)
   Compiling dirs v2.0.2
   Compiling term v0.6.1
   Compiling syn v1.0.99
   Compiling serde_derive v1.0.143
   Compiling serde v1.0.143
   Compiling garando_pos v0.1.0
   Compiling serde_json v1.0.83
   Compiling garando_errors v0.1.0
   Compiling garando_syntax v0.1.0
   Compiling ctest2 v0.4.8
error[E0061]: this function takes 3 arguments but 2 arguments were supplied
[...]

PR #4133, opened 2024-11-21
use `qemu-sparc64` to run sparc64 tests

folkertdev:
fixes https://github.com/rust-lang/libc/issues/4061

# Description

uses `qemu-sparc64` to run the sparc64 tests. That is (for some reason) much faster than the `test-runner-linux` script that uses `qemu-system-sparc64` under the hood. It runs the test suite locally in ~3 minutes, similar to the other targets.

I added some things to make this work, I'm not sure if other things should be removed that are now no longer needed.

# Sources

# Checklist

r? @tgross35


folkertdev:
on CI it's `4:45` ish, still pretty good, and no longer the bottleneck for CI.

tgross35 (approved):
Oh this is a wonderful improvement, thanks for working on it!

tgross35:
Merged in 15 minutes, awesome improvement

Issue #4115, opened 2024-11-19
CMSG_SPACE is no longer a const function


asomers:
`CMSG_SPACE` is supposed to be a const function.  It was, until PR #4105.  Now, nix fails to build with the libc-0.2 branch.  It fails for both x86_64-unknown-linux-gnu and x86_64-unknown-freebsd.  I bisected the regression to verify that it really was #4105.

Steps to reproduce:
```
$ git clone git@github.com:nix-rust/nix.git
$ cd nix
$ sed -I "" 's!version = "0.2.164"!git = "https://github.com/rust-lang/libc.git", rev = "ae052d3b6a045d58b7ef3187a94de81695e2591e"!' Cargo.toml
$ cargo check --all-features
    Updating crates.io index
    Updating git repository `https://github.com/rust-lang/libc.git`
     Locking 54 packages to latest compatible versions
      Adding sysctl v0.4.6 (available: v0.6.0)
   Compiling autocfg v1.4.0
   Compiling libc v0.2.164 (https://github.com/rust-lang/libc.git?rev=ae052d3b6a045d58b7ef3187a94de81695e2591e#ae052d3b)
   Compiling cfg_aliases v0.2.1
    Checking cfg-if v1.0.0
    Checking pin-utils v0.1.0
    Checking bitflags v2.6.0
   Compiling nix v0.29.0 (/tmp/nix)
   Compiling memoffset v0.9.1
error[E0015]: cannot call non-const fn `CMSG_SPACE` in constant functions
   --> src/sys/socket/mod.rs:652:14
    |
652 |     unsafe { libc::CMSG_SPACE(mem::size_of::<T>() as libc::c_uint) as usize }
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: calls in constant functions are limited to constant functions, tuple structs and tuple variants

For more information about this error, try `rustc --explain E0015`.
error: could not compile `nix` (lib) due to 1 previous error
```

tgross35:
Thanks for catching this, I guess the `const-extern-fn` feature (that we need for `ctest`) isn't getting enabled. It should be on by default except for when `ctest` is looking at things.

tgross35:
Fyi @SteveLauC this should be resolved

tamird:
[...]

PR #4134, opened 2024-11-22
fix: make sure the `const-extern-fn` feature is enabled

tgross35:
This was dropped in fa554bc4f8 on `main` and 674cc1f47f on `libc-0.2` ("Drop the libc_const_extern_fn conditional"). Unfortunately, this meant that functions were incorrectly never getting marked `const`, which showed up with `CMSG_SPACE` https://github.com/rust-lang/libc/issues/4115.

Instead of the fix here, I attempted to just use `cfg(not(libc_ctest))` instead of a Cargo feature to enable `const` extern functions; however, this seemed extremely problematic with `ctest` for some reason [2]. Instead, leave the feature as-is and just make it enabled by default.

Fixes: https://github.com/rust-lang/libc/issues/4115

[2]: https://github.com/rust-lang/libc/pull/4134

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -135,7 +135,7 @@ cargo-args = ["-Zbuild-std=core"]
 rustc-std-workspace-core = { version = "1.0.0", optional = true }
 
 [features]
-default = ["std"]
+default = ["const-extern-fn", "std"]
 std = []
 rustc-dep-of-std = ["rustc-std-workspace-core"]
 extra_traits = []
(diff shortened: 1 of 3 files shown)


tgross35:
This makes no sense, `ctest2` seems to be parsing the cfg-if backwards. I have to make the `const` version available under `cfg(not(libc_ctest))` to get `ctest` to not complain, but doing that makes the function not const by default. Changing it to `cfg(libc_ctest)` makes `ctest` look at the _other_ branch of `cfg_if` and fail because it sees `const extern "C"`.

tgross35:
I was going to ask before doing a release, thanks for confirming :)

PR #4145, opened 2024-11-25
solarish update stat type with st_fstype field.

devnexen:
ref illumos: https://illumos.org/man/2/stat
ref solaris: https://docs.oracle.com/cd/E36784_01/html/E36872/stat-2.html


tgross35 (approved):
The change looks fine but please include header links.

Cc @psumbera for Solaris

PR #4132, opened 2024-11-21
Migrate to Rust edition 2021

eduardosm:
# Description

Migrates the libc crate to Rust edition 2021 with an automated script (after some initial manual adjustments).

Since ctest2 does not support `crate::` in paths, a hack has been implemented in `libc-test/build.rs`, which makes a copy of `src` replacing every `crate::` with `::`.

cc https://github.com/rust-lang/libc/issues/3898

# Sources

# Checklist

- [x] Relevant tests in `libc-test/semver` have been updated
- [x] No placeholder or unstable values like `*LAST` or `*MAX` are
  included (see [#3131](https://github.com/rust-lang/libc/issues/3131))
- [ ] Tested locally (`cd libc-test && cargo test --target mytarget`);
  especially relevant for platforms that may not be checked in CI

PR #4148, opened 2024-11-25
Add new cfg linux_time_bits64 corresponding to __USE_TIME_BITS64

snogge:
# Description

Add a new cfg `linux_time_bits64` -- corresponding to the kernels `__USE_TIME_BITS64` --  with some associated changes as a base for 64bit `time_t` support for `glibc` and `musl`

# Sources

https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/socket.h#L157
https://github.com/torvalds/linux/blob/master/arch/mips/include/uapi/asm/socket.h#L164
https://github.com/torvalds/linux/blob/master/arch/powerpc/include/uapi/asm/socket.h

https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/resource.h#L58
https://github.com/torvalds/linux/blob/master/arch/mips/include/uapi/asm/resource.h#L31

https://github.com/torvalds/linux/blob/master/include/uapi/linux/input.h#L28

https://github.com/torvalds/linux/blob/master/include/uapi/linux/time.h#L17

# Checklist


tgross35:
Thanks! I'll need to double check the definitions but at first glance this looks good.

Could you add an internal env `RUST_LIBC_UNSTABLE_LINUX_TIME_BITS64` to control this to set this config? Then here https://github.com/rust-lang/libc/blob/42e5708d3c0cfcf68f6ffea45eed7e1732dbd7ec/ci/verify-build.sh#L67 add

```sh
if [ "$os" = "linux" ]; then
    RUST_LIBC_UNSTABLE_LINUX_TIME_BITS64=1 $cmd
fi
```

So we make sure the build is correct in CI

src/unix/linux_like/linux/arch/generic/mod.rs
```diff
@@ -37,10 +37,6 @@ pub const SO_PASSCRED: ::c_int = 16;
 pub const SO_PEERCRED: ::c_int = 17;
 pub const SO_RCVLOWAT: ::c_int = 18;
 pub const SO_SNDLOWAT: ::c_int = 19;
-pub const SO_RCVTIMEO: ::c_int = 20;
-pub const SO_SNDTIMEO: ::c_int = 21;
-// pub const SO_RCVTIMEO_OLD: ::c_int = 20;
-// pub const SO_SNDTIMEO_OLD: ::c_int = 21;
```
tgross35, line 43: I think maybe we could define both the `_OLD` and `_NEW` values to be consistent with the headers (maybe don't make them public if they're not in the docs?) and then make use of them within the config block.

build.rs
```diff
@@ -15,6 +15,7 @@ const ALLOWED_CFGS: &'static [&'static str] = &[
     "freebsd15",
     "libc_deny_warnings",
     "libc_ctest",
+    "linux_time_bits64",
```
tgross35, line 18:
```suggestion
    // Corresponds to `__USE_TIME_BITS64` in UAPI
    "linux_time_bits64",
```

tgross35:
Also FYI there is a large edition refactor coming https://github.com/rust-lang/libc/pull/4132 so this will probably wind up with conflicts in the near future.

src/unix/mod.rs
```diff
@@ -68,7 +68,12 @@ s! {
 
     pub struct timeval {
         pub tv_sec: time_t,
+        #[cfg(not(linux_time_bits64))]
         pub tv_usec: suseconds_t,
+        // For 64 bit time on 32 bit linux glibc, suseconds_t is still
+        // a 32 bit type.  Using suseconds_t here will break the tests, use i64 instead
+        #[cfg(linux_time_bits64)]
+        pub tv_usec: i64,
```
tgross35, line 76: How does this work in the kernel - is `suseconds_t` getting defined somewhere else? I only see https://github.com/torvalds/linux/blob/9f16d5e6f220661f73b36a4be1b21575651d8833/include/uapi/asm-generic/posix_types.h#L41 and something for sparc64.

tgross35: Ah, I see this is redefined in glibc https://github.com/bminor/glibc/blob/d92a5e1dad39ab0ac09ab1beeb055c1f1d981cd4/time/bits/types/struct_timeval.h#L8-L17. If the tests still pass, could you remove the changes to this type and leave a `// FIXME(time): ...` to adjust when we do the glibc updates?

src/unix/linux_like/linux/mod.rs
```diff
@@ -324,7 +324,19 @@ s! {
...
+        #[cfg(all(target_pointer_width = "32", linux_time_bits64))]
+        pub input_event_sec: ::c_ulong,
+
+        #[cfg(any(target_pointer_width = "64", not(linux_time_bits64)))]
+        pub input_event_usec: ::suseconds_t,
+        #[cfg(all(target_pointer_width = "32", linux_time_bits64))]
+        pub input_event_usec: ::c_ulong,
+
+        #[cfg(target_arch = "sparc64")]
+        _pad1: ::c_int,
```
tgross35, line 338: I think that we should leave the `pub time: ::timeval` for `#[cfg(any(target_pointer_width = "64", not(linux_time_bits64)))]` to minimize breakage. Maybe it's better to leave `time` or `__sec`+`__usec` as the fields, then make `input_event_u?sec` methods on the struct.

Also I believe `sparc64` needs `__usec`/`input_event_usec` to be a `c_uint` rather than `c_ulong`.

snogge: I choose the current struct based on the discussion starting here https://github.com/rust-lang/libc/pull/3175#pullrequestreview-2068891747 . The `input_event_X` members are the platform independent way of referring to these members, even if I cannot find any documentation for this. 
We can exclude the `input_event` change from this PR for now if we need to discuss this further,

src/unix/linux_like/linux/arch/generic/mod.rs
snogge: The `_OLD` and `_NEW` variants are not mentioned in `socket(7)` where the other constants are listed, so I added them as non-public.  Some of them are mentioned in `semver/TODO-linux.txt`.

src/unix/mod.rs
snogge: Done

snogge:
I think I've addressed most of your comments.  I did not change the `input_event` struct as I think the current version is the most compatible one.

src/unix/linux_like/linux/mod.rs
snogge: And `suseconds_t` is `i32` on `sparc64` for `glibc`.  As far as I can tell none of the other `libc`'s support `sparc64`.

snogge:
The PR has been rebased on main, including the mentioned refactor in #4132 .

tgross35:
Thanks for the changes, I'll double check this within the next couple of days (the recent MSRV bump and edition change have been pretty time consuming in this repo).

src/unix/linux_like/linux/mod.rs
tgross35: Thanks for the information, that reasoning makes sense. Agreed that we should make this change but I would like to be able to backport this PR (so users can start experimenting before commiting to 1.0) so I don't think we should make this change now. Would it be possible to:

1. Leave `time` with the cfg mentioned in https://github.com/rust-lang/libc/pull/4148#discussion_r1857511742
2. Just put the conflicting `input_event_*` in a block comment with a `FIXME(1.0): ...` so we make this change before that release

I think we should mark `time` deprecated at some point but I am hesitant to do that without a migration path, so this can be revisited later.

tgross35:
One last request to defer the `time` field change until later and then this looks great to me.

snogge:
I've updated input_event as I think you meant.

Issue #4149, opened 2024-11-26
`libc` crate tests do not run (or even compile) on linux targets in CI


tamird:
https://github.com/rust-lang/libc/pull/1536 added the ability to mark functions as const, along with a test in https://github.com/rust-lang/libc/blob/42e5708d3c0cfcf68f6ffea45eed7e1732dbd7ec/tests/const_fn.rs. Unfortunately this file is completely dead due to the combination of `#[cfg(target_os = "linux")]` in it and the behavior introduced in 3eb4a48b1b63909a8898aa37b5f610a63d776f4d which skips building the libc crate when `docker: true`, which [is the case for all linuxes](https://github.com/rust-lang/libc/blob/42e5708d3c0cfcf68f6ffea45eed7e1732dbd7ec/.github/workflows/ci.yaml#L52-L57).

This all culminated in the fact that https://github.com/rust-lang/libc/pull/4105 broke const functions and https://github.com/rust-lang/libc/pull/4134 failed to fix it in the `--no-default-features` case. This remains broken.

@tgross35 

EDIT: all releases after 0.2.163 which do not contain a fix for this should be yanked.

tgross35:
Your `--no-default-features` configuration is not covered, but isn't this test being run by https://github.com/rust-lang/libc/blob/42e5708d3c0cfcf68f6ffea45eed7e1732dbd7ec/ci/run.sh#L83-L92?

tamird:
No, it is not, because of `exec`: https://github.com/rust-lang/libc/blob/42e5708d3c0cfcf68f6ffea45eed7e1732dbd7ec/ci/run.sh#L80

tgross35:
The branch with `exec` would only be executed when `QEMU` is set, which shouldn't be the case for anything x86 (though a command followed by `exit` is better style than relying on `exec`). I also don't think `QEMU` is set for any of the qemu-system tests, since those were failing when I added the test branch.

tamird:
> The branch with `exec` would only be executed when `QEMU` is set, which shouldn't be the case for anything x86 (though a command followed by `exit` is better style than relying on `exec`, and that grep seems fragile). I also don't think `QEMU` is set for any of the qemu-system tests, since those were failing when I added the test branch.

Yes, I think you're right.

> `$cmd` is not the same at lines 91 and 123, it gains `--manifest-path libc-test/Cargo.toml` in between. This should probably just be `--all` on platforms where unit tests can run successfully.

[...]

PR #4151, opened 2024-11-26
Ensure all tests in the workspace are run and that `const extern fn` is always enabled

tgross35:
In [1] this conditional was dropped in favor of a Cargo feature, which was turned on by default in [2]. However, this did not help the case where `--no-default-features` is passed.

Unfortunately we still can't drop this config entirely since `ctest` cannot parse the syntax, so change back to useing a `cfg` to control constness rather than a Cargo feature.

Fixes: https://github.com/rust-lang/libc/issues/4149

[1]: https://github.com/rust-lang/libc/pull/4105
[2]: https://github.com/rust-lang/libc/pull/4134

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -167,40 +167,31 @@ macro_rules! e {
     )*);
 }
 
-// This is a pretty horrible hack to allow us to conditionally mark
-// some functions as 'const', without requiring users of this macro
-// to care about the "const-extern-fn" feature.
+// This is a pretty horrible hack to allow us to conditionally mark some functions as 'const',
+// without requiring users of this macro to care "libc_const_extern_fn".
 //
-// When 'const-extern-fn' is enabled, we emit the captured 'const' keyword
-// in the expanded function.
+// When 'libc_const_extern_fn' is enabled, we emit the captured 'const' keyword in the expanded
+// function.
 //
-// When 'const-extern-fn' is disabled, we always emit a plain 'pub unsafe extern fn'.
+// When 'libc_const_extern_fn' is disabled, we always emit a plain 'pub unsafe extern fn'.
 // Note that the expression matched by the macro is exactly the same - this allows
-// users of this macro to work whether or not 'const-extern-fn' is enabled
+// users of this macro to work whether or not 'libc_const_extern_fn' is enabled
 //
 // Unfortunately, we need to duplicate most of this macro between the 'cfg_if' blocks.
 // This is because 'const unsafe extern fn' won't even parse on older compilers,
-// so we need to avoid emitting it at all of 'const-extern-fn'.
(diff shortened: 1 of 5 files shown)


tgross35:
@tamird could you check this resolves https://github.com/rust-lang/libc/issues/4149 for you?

src/macros.rs
```diff
@@ -199,7 +199,7 @@ macro_rules! e {
 // cfg completely.
 // FIXME(ctest): ctest can't handle `$(,)?` so we use `$(,)*` which isn't quite correct.
 cfg_if! {
-    if #[cfg(feature = "const-extern-fn")] {
```
tamird, line 202: consider removing the feature from the default features (in this commit, so it is backported) and also updating the comment in `src/macros.rs` in this commit (so it no longer mentions this feature, also so it is backported).

tamird:
> @tamird could you check this resolves #4149 for you?

The test failures before the fix are pretty convincing.

src/macros.rs
tgross35: Reasonable 👍 changed, thanks for taking a look.

tgross35:
Backport in https://github.com/rust-lang/libc/pull/4152

src/macros.rs
```diff
@@ -167,40 +167,31 @@ macro_rules! e {
...
-// Note that users of this macro need to place 'const' in a weird position
-// (after the closing ')' for the arguments, but before the return type).
-// This was the only way I could satisfy the following two requirements:
-// 1. Avoid ambiguity errors from 'macro_rules!' (which happen when writing '$foo:ident fn'
-// 2. Allow users of this macro to mix 'pub fn foo' and 'pub const fn bar' within the same
-// 'f!' block
+// Specifically, moving the 'cfg_if' into the macro body will *not* work. Doing so would cause the
+// '#[cfg(libc_const_extern_fn)]' to be emitted into user code. The 'cfg' gate will not stop Rust
+// from trying to parse the 'pub const unsafe extern fn', so users would get a compiler error even
+// when the 'libc_const_extern_fn' feature is disabled.
```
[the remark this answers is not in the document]
tgross35: Thanks for noticing, but I'm not worried about it - this comment is preexisting and will hopefully go away in the near future once ctest gets fixed.

tamird:
Thanks for fixing. Please remember to yank the broken releases.

tgross35:
There was some other unintentional breakage in the huge cleanup refactoring, it's already yanked.

tamird:
0.2.164 remains unyanked.

tgross35:
That version didn't update this feature https://github.com/rust-lang/libc/compare/0.2.163...0.2.164

tamird:
Ah, thanks for clarifying. I thought the regression was first noticed after release but it was before.

PR #4152, opened 2024-11-26
[0.2]: Ensure all tests in the workspace are run and that const extern fn is always enabled

tgross35:
Backport https://github.com/rust-lang/libc/pull/4151, excluding the last commit that removes the `const-extern-fn` feature.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -167,40 +167,31 @@ macro_rules! e {
     )*);
 }
 
-// This is a pretty horrible hack to allow us to conditionally mark
-// some functions as 'const', without requiring users of this macro
-// to care about the "const-extern-fn" feature.
+// This is a pretty horrible hack to allow us to conditionally mark some functions as 'const',
+// without requiring users of this macro to care "libc_const_extern_fn".
 //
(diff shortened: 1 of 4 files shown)


src/macros.rs
```diff
@@ -167,40 +167,31 @@ macro_rules! e {
...
-// Note that users of this macro need to place 'const' in a weird position
-// (after the closing ')' for the arguments, but before the return type).
-// This was the only way I could satisfy the following two requirements:
-// 1. Avoid ambiguity errors from 'macro_rules!' (which happen when writing '$foo:ident fn'
-// 2. Allow users of this macro to mix 'pub fn foo' and 'pub const fn bar' within the same
-// 'f!' block
+// Specifically, moving the 'cfg_if' into the macro body will *not* work. Doing so would cause the
+// '#[cfg(libc_const_extern_fn)]' to be emitted into user code. The 'cfg' gate will not stop Rust
+// from trying to parse the 'pub const unsafe extern fn', so users would get a compiler error even
+// when the 'libc_const_extern_fn' feature is disabled.
```
tamird, line 187: it's no longer a feature.

PR #4158, opened 2024-11-27
Fix usage with `rustc-dep-of-std`

tgross35:
Since https://github.com/rust-lang/libc/pull/4038 we use derive macros rather than manually implementing `Clone`
and `Copy`. However, this caused the build in `std` to start failing since the `core` prelude is not available. This provides the derive macros as well as `derive` itself.

Resolve this by using complete paths. Additionally allow `internal_features` to suppress the warning using `link_cfg`.

Includes a test that verifies build with `rustc-dep-of-std` enabled.

### DIFF
--- a/ci/verify-build.sh
+++ b/ci/verify-build.sh
@@ -69,6 +69,11 @@ test_target() {
     $cmd --no-default-features
     $cmd --no-default-features --features extra_traits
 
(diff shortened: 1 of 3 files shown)


tgross35:
I'm going to wait for https://github.com/rust-lang/libc/pull/4161 and then do the remaining fixes on top of that.

PR #4159, opened 2024-11-27
ci: Check various FreeBSD versions

tgross35:
Since we suport multiple versions and this is tier 2, we should make sure that we can build with a couple versions. This does not run tests.

This introduces an environment variable for an easy way to override the version for testing.

### DIFF
--- a/ci/verify-build.sh
+++ b/ci/verify-build.sh
@@ -28,6 +28,7 @@ if [ "$TOOLCHAIN" = "nightly" ] ; then
     rustup component add rust-src
 fi
 
+# Run the tests for a specific target
 test_target() {
     target="${1}"
     no_dist="${2:-0}"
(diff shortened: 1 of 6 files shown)


tgross35:
@asomers could you take a look at this? I don't think we should check all versions with all architectures on CI but we can at least check them all with the T1 i686 and x86_64.

ci/verify-build.sh
```diff
@@ -67,8 +68,31 @@ test_target() {
...
     $cmd --no-default-features --features extra_traits
+
+    # For freebsd targets, check with the different versions we support if on
+    # nightly or stable
+    case "$rust-$target" in
+        stable-*freebsd*) do_freebsd_checks=1 ;;
+        nightly-*freebsd*) do_freebsd_checks=1 ;;
+    esac
+    
+    if [ "${do_freebsd_checks:-}" = "1" ]; then
```
asomers, line 79: I would write it like this, but that's just a matter of style. 
```suggestion
    if [ -n "$do_freebsd_checks" ]; then
```

ci/verify-build.sh
```diff
@@ -67,8 +68,31 @@ test_target() {
...
+            export RUST_LIBC_UNSTABLE_FREEBSD_VERSION="$version"
+            $cmd
+            $cmd --no-default-features
+        done
+    fi
 }
 
+freebsd_versions="\
+11 \
+12 \
```
asomers, line 90: This will test 12 twice, once at line 69-70 and once at 82-83.

asomers (approved):
This looks good to me.  There are a few other FreeBSD arches too, but you got the most important ones.

ci/verify-build.sh
tgross35: Do you have any idea how best to avoid this? It wouldn't be that bad if there isn't anything straightforward.

ci/verify-build.sh
tgross35: Ah it still needs the fallback since it may be unset, but I changed to `-n`

ci/verify-build.sh
asomers: The default version of "12" is set in build.rs.  So you can simply leave 12 off of the list.  But then it would have to be manually updated in case the default ever changes.  So if object files are retained between runs, then running `cargo check` twice should be fast enough, I think.

ci/verify-build.sh
asomers: The quotes are the key to making it work when the variable is unset.  You don't actually need the `:-`

ci/verify-build.sh
tgross35: That's about the same conclusion I came to, it should be fast enough that it's better to not worry about forgetting to update the default.

Thanks for taking a look at this!

PR #4157, opened 2024-11-27
fix(wasi): Add back unsafe block for clockid_t static variables

DarumaDocker:
The MSRV is still far before 1.82.0

# Description

Version 0.2.166 can not be built for the wasm32-wasi target caused by the removed `unsafe` block.
This commit is taking them back.

# Checklist

- [x] Relevant tests in `libc-test/semver` have been updated
- [x] No placeholder or unstable values like `*LAST` or `*MAX` are
  included (see [#3131](https://github.com/rust-lang/libc/issues/3131))
- [x] Tested locally (`cd libc-test && cargo test --target mytarget`);
  especially relevant for platforms that may not be checked in CI

PR #4160, opened 2024-11-27
[0.2] Backports

tgross35:
Backport https://github.com/rust-lang/libc/pull/4157


tgross35:
The first CI run is me applying only the patch adding tests, that fails as expected without the fix https://github.com/rust-lang/libc/actions/runs/12046642238/job/33587533778?pr=4160.

PR #4161, opened 2024-11-27
Create an internal prelude

tgross35:
When building with `rustc-dep-of-std`, we don't get the core types imported by default (`Clone`, `Copy`, `Option`). In order to avoid needing to import these individually, introduce a prelude that includes them.

This also includes the C numeric types that are frequently used.

### DIFF
--- a/src/fuchsia/aarch64.rs
+++ b/src/fuchsia/aarch64.rs
@@ -1,4 +1,5 @@
-use crate::{c_int, c_long, c_uint, c_ulong, c_ulonglong, c_ushort, off_t, size_t};
+use crate::off_t;
+use crate::prelude::*;
 
 pub type c_char = u8;
 pub type __u64 = c_ulonglong;
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -3,7 +3,7 @@
 //! More functions and definitions can be found in the more specific modules
 //! according to the platform in question.
 
-use crate::c_void;
+use crate::prelude::*;
 
 // PUB_TYPE
 
@@ -1073,8 +1073,8 @@ cfg_if! {
             }
         }
         impl Eq for sysinfo {}
-        impl crate::fmt::Debug for sysinfo {
-            fn fmt(&self, f: &mut crate::fmt::Formatter) -> crate::fmt::Result {
+        impl fmt::Debug for sysinfo {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
                 f.debug_struct("sysinfo")
                     .field("uptime", &self.uptime)
                     .field("loads", &self.loads)
@@ -1093,8 +1093,8 @@ cfg_if! {
                     .finish()
             }
         }
-        impl crate::hash::Hash for sysinfo {
-            fn hash<H: crate::hash::Hasher>(&self, state: &mut H) {
+        impl hash::Hash for sysinfo {
+            fn hash<H: hash::Hasher>(&self, state: &mut H) {
                 self.uptime.hash(state);
                 self.loads.hash(state);
                 self.totalram.hash(state);
@@ -1123,16 +1123,16 @@ cfg_if! {
             }
         }
         impl Eq for sockaddr_un {}
-        impl crate::fmt::Debug for sockaddr_un {
-            fn fmt(&self, f: &mut crate::fmt::Formatter) -> crate::fmt::Result {
+        impl fmt::Debug for sockaddr_un {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
                 f.debug_struct("sockaddr_un")
                     .field("sun_family", &self.sun_family)
                     // FIXME: .field("sun_path", &self.sun_path)
                     .finish()
             }
         }
-        impl crate::hash::Hash for sockaddr_un {
-            fn hash<H: crate::hash::Hasher>(&self, state: &mut H) {
+        impl hash::Hash for sockaddr_un {
+            fn hash<H: hash::Hasher>(&self, state: &mut H) {
                 self.sun_family.hash(state);
                 self.sun_path.hash(state);
             }
@@ -1150,8 +1150,8 @@ cfg_if! {
             }
         }
         impl Eq for sockaddr_storage {}
-        impl crate::fmt::Debug for sockaddr_storage {
-            fn fmt(&self, f: &mut crate::fmt::Formatter) -> crate::fmt::Result {
+        impl fmt::Debug for sockaddr_storage {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
                 f.debug_struct("sockaddr_storage")
                     .field("ss_family", &self.ss_family)
                     .field("__ss_align", &self.__ss_align)
@@ -1159,8 +1159,8 @@ cfg_if! {
                     .finish()
             }
         }
-        impl crate::hash::Hash for sockaddr_storage {
-            fn hash<H: crate::hash::Hasher>(&self, state: &mut H) {
+        impl hash::Hash for sockaddr_storage {
+            fn hash<H: hash::Hasher>(&self, state: &mut H) {
                 self.ss_family.hash(state);
                 self.__ss_align.hash(state);
                 self.__ss_pad2.hash(state);
@@ -1196,8 +1196,8 @@ cfg_if! {
             }
         }
         impl Eq for utsname {}
-        impl crate::fmt::Debug for utsname {
-            fn fmt(&self, f: &mut crate::fmt::Formatter) -> crate::fmt::Result {
+        impl fmt::Debug for utsname {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
                 f.debug_struct("utsname")
                     // FIXME: .field("sysname", &self.sysname)
                     // FIXME: .field("nodename", &self.nodename)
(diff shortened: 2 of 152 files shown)


tgross35:
@tammela fyi, I guess the reason we use `crate`'s `Option` `Clone` and `Copy` was for compatibility when we don't have the prelude as a dependency of `core`. So this is just moving those imports to a prelude so we don't need to think about it.

tgross35:
@eduardosm I still have a bit to tweak here but would you mind giving this a second set of eyes?

<details>
<summary>Current script</summary>

```sh
#!/bin/bash

set -eux

files=()

# Types either defined in this crate or in `core`
prelude_types=(
    c_char
    c_double
    c_float
    c_int
    c_longlong
    c_long
    c_short
    c_uchar
    c_uint
    c_ulonglong
    c_ulong
    c_ushort
    c_void
    intptr_t
    size_t
    ssize_t
    Clone
    Copy
    Option
)

# Reexports from core
prelude_modules=(
    fmt
    hash
    iter
    mem
)

# Everything in the prelude
prelude=( "${prelude_types[@]}" "${prelude_modules[@]}" )

# Generate a list of all files excluding `lib.rs` (since the prelude being
# defined there makes string matching weird).
while IFS= read -r -d '' file; do
    files+=("$file")
done < <(find src -name '*.rs' -not -name '*lib.rs' -not -name '*macros.rs' -print0)

# files=(src/unix/linux_like/linux/gnu/b64/aarch64/lp64.rs)

for file in "${files[@]}"; do
    needs_prelude=0

    # If the file already has some sort of glob import, skip it
    if rg --pcre2 -q 'use crate::(?!prelude).*\*' "$file"; then
        continue
    fi

    # Core types always require the prelude to handle rustc-dep-of-std
    if rg --pcre2 -q "\b(?<!\.)(Option|Clone|Copy)\b" "$file"; then
        echo NEEDS PRELUDE WITH CORE
        needs_prelude=1
    fi

    # If we use any types that are specified in the prelude then we will import it
    for ty in "${prelude[@]}"; do
        # If the type is defined in the current module, we don't need it from the prelude
        if rg -q "type $ty =" "$file"; then
            continue
        fi

        if rg -qU "(crate::$ty|use crate::(\{\n){0,2}.*$ty)" "$file"; then
            echo NEEDS PRELUDE WITH TYPE $ty
            needs_prelude=1
        fi
    done

    # Check if the prelude is needed and does not already exist; if so, add it
    if [ "$needs_prelude" = "1" ] && ! rg -q "use crate::prelude::\*" "$file"; then
        echo ADDING PRELUDE
        perl -pi -0777 -e 's/(use crate::)/use crate::prelude::*;\n$1/' "$file"
    fi

    for ty in "${prelude[@]}"; do
        export TY="$ty" # env for perl to use

        # Remove simple imports `use crate::ty;`
        perl -pi -0777 -e 's/use crate::($ENV{TY});//g' "$file"

        # Remove the type if it is part of a group import
        perl -pi -0777 -e 's/(use crate::\{?(.*|(\n.*){0,2}))\b$ENV{TY}\b,? ?/$1/g' "$file"

        # Replace pathed `crate::ty`
        perl -pi -0777 -e 's/crate::($ENV{TY})\b/$1/g' "$file"
    done

    # For some reason, rustfmt doesn't trim leading newlines. Do so manually here.
    perl -pi -0777 -e 's/\A\n+//' "$file"

    rustfmt "$file"
done

./ci/style.sh
```

</details>

eduardosm:
Script looks good to me and produces identical results in my machine.

However I'm seeing some new compilation failures on some targets (Tier 3, I'm not sure how important it is):

aarch64-unknown-trusty:
```
error[E0432]: unresolved import `crate::intptr_t`
   --> src/macros.rs:83:57
    |
83  |                 c_ulong, c_ulonglong, c_ushort, c_void, intptr_t, size_t, ssize_t,
    |                                                         ^^^^^^^^ no `intptr_t` in the root
    |
   ::: src/lib.rs:112:9
    |
112 |         prelude!();
    |         ---------- in this macro invocation
    |
    = note: this error originates in the macro `prelude` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Maybe `intptr_t` (and `uintptr_t`) could be added to that target?

I believe they should be the same as `ssize_t` / `size_t`:

https://cs.android.com/android/platform/superproject/main/+/main:external/trusty/musl/include/alltypes.h.in;l=1-5

riscv32imac-unknown-xous-elf:
```
error: no rules expected `full`
   --> src/lib.rs:136:18
    |
136 |         prelude!(full);
    |                  ^^^^ no rules expected this token in macro call
    |
   ::: src/macros.rs:65:1
    |
65  | macro_rules! prelude {
    | -------------------- when calling this macro
    |
    = note: while trying to match end of macro
```

Changing `prelude!(full);` to `prelude!();` fixes it.

eduardosm:
Another failure on powerpc64-unknown-openbsd:
```
error[E0433]: failed to resolve: use of undeclared crate or module `mem`
 --> src/unix/bsd/netbsdlike/openbsd/powerpc64.rs:5:39
  |
5 | pub(crate) const _ALIGNBYTES: usize = mem::size_of::<c_long>() - 1;
  |                                       ^^^ use of undeclared crate or module `mem`
  |
help: consider importing one of these modules
  |
1 + use crate::prelude::mem;
  |
1 + use core::mem;
```

The script changed `crate::mem::size_of` to `mem::size_of` but did not import the prelude.

Probably because this line will not insert the prelude import unless the file has already any other import:
 
```sh
perl -pi -0777 -e 's/(use (crate|super)::)/use crate::prelude::*;\n$1/' "$file"
```

tgross35:
Great catches, thank you for taking a look! Updated to add `intptr_t` and `uintptr_t` to Trusty (cc @randomPoison), fixed the `prelude!(full)`, and adjusted the script to use your awk commands from https://github.com/rust-lang/libc/pull/4132 rather than adding the import via substitution (next script will be in python or rustscript...).

I'll make sure to verify builds for tier3 once this is closer to ready for a release which will probably be tomorrow.

Current script:

<details>
<summary>Current script</summary>

```sh
#!/bin/bash

set -eux

files=()

# Types either defined in this crate or in `core`
prelude_types=(
    c_char
    c_double
    c_float
    c_int
    c_longlong
    c_long
    c_short
    c_uchar
    c_uint
    c_ulonglong
    c_ulong
    c_ushort
    c_void
    intptr_t
    size_t
    ssize_t
    Clone
    Copy
    Option
    Send
    Sync
)

# Reexports from core
prelude_modules=(
    fmt
    hash
    iter
    mem
)

# Everything in the prelude
prelude=( "${prelude_types[@]}" "${prelude_modules[@]}" )

# Generate a list of all files excluding `lib.rs` (since the prelude being
# defined there makes string matching weird).
while IFS= read -r -d '' file; do
    files+=("$file")
done < <(find src -name '*.rs' -not -name '*lib.rs' -not -name '*macros.rs' -not -name 'fixed_width_ints.rs' -print0)

for file in "${files[@]}"; do
    needs_prelude=0

    # If the file already has some sort of glob import, skip it
    if rg --pcre2 -q 'use (crate|super)::(?!prelude).*\*' "$file"; then
        continue
    fi

    # Core types always require the prelude to handle rustc-dep-of-std
    if rg --pcre2 -q '\b(?<!\.)(Option|Clone|Copy|Send|Sync|fmt|hash|iter|mem)\b' "$file"; then
        needs_prelude=1
    fi

    # If we use any types that are specified in the prelude then we will import it
    for ty in "${prelude[@]}"; do
        # If the type is defined in the current module, we don't need it from the prelude
        if rg -q "type $ty =" "$file"; then
            continue
        fi

        if rg -qU '((crate|super)::'"$ty"'|use (crate|super)::(\{\n){0,2}.*'"$ty)" "$file"; then
            needs_prelude=1
        fi
    done

    # Check if the prelude is needed and does not already exist; if so, add it
    if [ "$needs_prelude" = "1" ] && ! rg -q 'use crate::prelude::\*' "$file"; then
        # Split the file into two parts: module-level attributes and rest
        # Imports will be added after module-level attributes

        attrs=$(awk '/^#!|^\/\/!/ {found=NR} {lines[NR]=$0} END {for (i=1; i<=found; i++) print lines[i]}' "$file")
        rest=$(awk '/^#!|^\/\/!/ {found=NR} END {if (found) {for (i=found+1; i<=NR; i++) print lines[i]} else {for (i=1; i<=NR; i++) print lines[i]}} {lines[NR]=$0}' "$file")


        printf "%s\n" "$attrs" > "$file"
        printf "\n%s\n\n" "use crate::prelude::*;" >> "$file"
        printf "%s" "$rest" >> "$file"
    fi

    for ty in "${prelude[@]}"; do
        export TY="$ty" # env for perl to use

        # Remove simple imports `use crate::ty;`
        perl -pi -0777 -e 's/use ((crate|super)::)?($ENV{TY});//g' "$file"

        # Remove the type if it is part of a group import
        perl -pi -0777 -e 's/(use (crate|super)::\{?(.*|(\n.*){0,2}))\b$ENV{TY}\b,? ?/$1/g' "$file"

        # Replace pathed `crate::ty`
        perl -pi -0777 -e 's/(crate|super)::($ENV{TY})\b/$2/g' "$file"
    done

    # For some reason, rustfmt doesn't trim leading newlines. Do so manually here.
    perl -pi -0777 -e 's/\A\n+//' "$file"

    rustfmt "$file"
done

./ci/style.sh
```

</details>

randomPoison:
The Trusty change looks good 👍 Thanks for adding those.

PR #4163, opened 2024-11-27
move AF_XDP structs and constants to `linux/mod.rs`

arctic-alpaca:
# Description

As discussed in https://github.com/rust-lang/libc/pull/3956, the `AF_XDP` structs and constants are all uapi and as such should go into `src/unix/linux_like/linux/mod.rs`. This PR moves all `AF_XDP` structs and constants there.

The `_v1` versions are for backwards compatibility with older kernel versions. With v1.0, they could probably be removed, since the newer versions will work (with new fields not being taken into account by the kernel). If that is desirable, I can crate a follow-up PR to remove them or remove them here.

# Sources

- All structs and constants not explicitly listed are defined in [`if_xdp.h`](https://github.com/torvalds/linux/blob/master/include/uapi/linux/if_xdp.h).
- `xdp_ring_offset_v1` & `xdp_mmap_offsets_v1` in [`xsk.h`](https://github.com/torvalds/linux/blob/master/net/xdp/xsk.h).
- `xdp_umem_reg_v1` & `xdp_statistics_v1`  in [`xsk.c`](https://github.com/torvalds/linux/blob/master/net/xdp/xsk.c).

# Checklist

I could not test the the changes locally due to the error described in https://github.com/rust-lang/libc/pull/3865#issuecomment-2316180504


tgross35 (approved):
Thanks!

If the `_v1` structs aren't intended to be used anymore, could you mark them as deprecated here? Then this can go to 0.2, and they can be removed in a separate PR that only goes to `main`.

arctic-alpaca:
If the CI failure is related to the changes, I cannot tell what caused it.

src/unix/linux_like/linux/mod.rs
```diff
@@ -1213,6 +1213,100 @@ s! {
...
+    pub struct xdp_mmap_offsets {
+        pub rx: xdp_ring_offset,
+        pub tx: xdp_ring_offset,
+        pub fr: xdp_ring_offset,
+        pub cr: xdp_ring_offset,
+    }
+
+    #[deprecated(
+        since = "0.2.167",
+        note = "We consider removing this as the newer version of this struct will work with older kernel versions. If you're using it, please comment on https://github.com/rust-lang/libc/issues/4168"
```
tgross35, line 1241: Slight rewording to suggest an alternative, and to rewrap to 100 characters. Could you update the others as well?

```suggestion
        note = "Use `xdp_ring_offset` instead, which will still work with old kernel versions. If \
                you still need this version, please comment at \
                https://github.com/rust-lang/libc/issues/4168"
```

After this please squash and we should be able to merge.

tgross35:
Nah CI is just Android getting stuck, that happens for some reason

arctic-alpaca:
I just realized that I made a mistake with the deprecation. The older versions might be used to check sizes for when another version gets added. Basically checking the length the kernel returns against `foo_v1`, `foo_v2` and if it's larger than both, going with `foo`. Currently there are no `_v2` structs, but that's only the case because of the bug mentioned in https://github.com/rust-lang/libc/pull/3956. So while the up-to-date structs would work, there would be no way to check which version the kernel used if/when `_v2` versions get added.

I'm not sure whether you would prefer to keep them in, or to only support the latest version of the Linux Kernel.

Sorry for getting this wrong. If it's fine for you, I'd remove the deprecation info and close the issue so that this PR only contains the move to the `src/unix/linux_like/linux/mod.rs` file.

tgross35:
I think I need to double check what the kernel headers/docs say to get a better understanding. Feel free to drop deprecation from this PR, we can revisit later.

arctic-alpaca:
Thanks, sorry again for the trouble.

tgross35:
No trouble at all :)

tgross35:
Looks like I missed labeling this for backport, feel free to ping me in the future if something gets excluded!

PR #4171, opened 2024-12-03
[solaris/illumos] add SIGRTMIN and SIGRTMAX

sunshowers:
# Description

Add these functions, similar to the Linux ones. Also add tests.

# Sources

For illumos, the source code is at [1] and documentation is at [2]. Blame suggests that Solaris also supports the same calls.

[1]: https://github.com/illumos/illumos-gate/blame/27ecbff00d8c86a2647d6fe325cacb220d712115/usr/src/uts/common/sys/iso/signal_iso.h#L100-L101
[2]: https://illumos.org/man/3HEAD/signal.h

The comment about "no placeholder values" doesn't really apply in this case, just like on Linux, because `SIGRTMAX` is a dynamic function call.

# Checklist


jclulow (approved):
Looks good, thanks!

tgross35 (approved):

sunshowers:
Thanks for the quick review :D

PR #4176, opened 2024-12-07
fix: make Debug impl for unions opaque

### LINKED ISSUE #4118: Make the `Debug` implementation for `union`s always opaque
We currently have a lot of `unsafe` reads of unions fields which is source of easy unsoundness. Let's make things easier while avoiding this completely:

1. Change `s_no_extra_traits` to do an opaque `Debug` implementation on unions, here https://github.com/rust-lang/libc/blob/2f931d9d5789039fdde0419df79eee92c29b1270/src/macros.rs#L122-L129
2. Delete all existing `Debug` implementations on unions (which will now conflict).

The implementation is easy, this will just print the types as `some_union { .. }`:

```rust
#[cfg(feature = "extra_traits")]
impl ::core::fmt::Debug for $i {
    fn fmt(&self, f: &mut ::core::fmt::Formatter<'_>) -> ::core::fmt::Result {
        f.debug_struct(stringify!($i)).finish_non_exhaustive()
    }
}
```

We can backport this to `libc-0.2` since it fixes some soundness issues without breaking any API.

zzau13:
Fixes #4118

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -159,6 +159,13 @@ macro_rules! s_no_extra_traits {
             $(#[$attr])*
             pub union $i { $($field)* }
         }
+
+        #[cfg(feature = "extra_traits")]
+        impl ::core::fmt::Debug for $i {
+            fn fmt(&self, f: &mut ::core::fmt::Formatter<'_>) -> ::core::fmt::Result {
+                f.debug_struct(::core::stringify!($i)).finish_non_exhaustive()
+            }
+        }
     );
 
     (it: $(#[$attr:meta])* pub struct $i:ident { $($field:tt)* }) => (
--- a/src/unix/aix/mod.rs
+++ b/src/unix/aix/mod.rs
@@ -575,14 +575,6 @@ cfg_if! {
             }
         }
         impl Eq for __sigaction_sa_union {}
-        impl fmt::Debug for __sigaction_sa_union {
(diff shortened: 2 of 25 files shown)


tgross35:
Thanks for working on this!

I'll take a more in-depth look soon, but why did this add `core::cmp` to `Eq` and `PartialEq`? If it is related to `rustc-dep-of-std` then that can be disregarded, we don't need to support both `rustc-dep-of-std` and `extra-traits` being enabled at the same time.

ci/run.sh
```diff
@@ -122,4 +122,5 @@ else
     $cmd --no-default-features
     $cmd
     $cmd --features extra_traits
+    $cmd --all-features
```
tgross35, line 125: Ah, I see this answers my question. We already run checks with `extra_traits` and `rustc-dep-of-std` independently here https://github.com/rust-lang/libc/blob/149d3922e3493f14ec03f0bdd7968e0fe85b5ddf/ci/verify-build.sh#L64-L75, making them work together isn't a concern. So you can drop this part of the change and anything that was added to fix it.

tgross35:
Fyi the current error comes from a new lint in a recent nightly, I'll have it fixed soon https://github.com/rust-lang/libc/pull/4177

tgross35:
Feel free to cherry pick 99f4dd920b580ca3d201d948153cffe54158e4b8 so you get CI results now, I can rebase once things turn green.

tgross35 (approved):
Thank you for taking care of this!

PR #4180, opened 2024-12-07
ci: Upload artifacts created by libc-test

tgross35:
This gives us something that is easier to look at when the automatically generated tests fail.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -96,12 +96,22 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           key: ${{ matrix.target }}
+
       - name: Run natively
         if: "!matrix.docker"
         run: ./ci/run.sh ${{ matrix.target }}
       - name: Run in Docker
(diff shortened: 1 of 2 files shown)


tgross35:
@asomers it seems like the FreeBSD 15 runner is giving us an error that `CAP_UNUSED0_44` is undeclared. Do you know if there is a recent change that we should adjust or skip this for?

asomers:
Yep.  https://github.com/freebsd/freebsd-src/commit/b165e9e3ea4e327fc421d81c2a89242bd8720780 deleted that definition.  I wonder if we should just delete all of the CAP_UNUSED constants, for the same reason that we don't allow constants like XXX_MAX or RLIM_NLIMITS?

asomers:
@tgross35 see #4183 and #4184 for the `CAP_UNUSED0_44` fix.

tgross35:
I was going to say it's entirely up to you, it seemed like capsicum was the only user based on a quick search. Appreciate the fast PRs.

PR #4181, opened 2024-12-07
ci: Switch to the latest Windows runners

tgross35:


tgross35:
Looks like even though the actions repo has windows-2025, it is not yet available.

PR #4182, opened 2024-12-07
fix: use core::mem::offset_of

zzau13:


tgross35:
Thanks for taking a look at this, but indeed we can't use `offset_of` until our MSRV eventually reaches 1.77.

PR #4183, opened 2024-12-08
[0.2] Deprecate FreeBSD's CAP_UNUSED* and CAP_ALL* constants

asomers:
They aren't stable across OS versions, and they don't have any legitimate use in applications.

# Description

Deprecate constants that are unstable and not useful to our consumers

# Sources

This commit, for example, changed some of these constants in FreeBSD 15.

https://github.com/freebsd/freebsd-src/commit/b165e9e3ea4e327fc421d81c2a89242bd8720780

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2618,6 +2618,11 @@ fn test_freebsd(target: &str) {
             // Added in FreeBSD 14.0
             "TCP_FUNCTION_ALIAS" if Some(14) > freebsd_ver => true,
 
+            // These constants may change or disappear in future OS releases, and they probably
+            // have no legitimate use in applications anyway.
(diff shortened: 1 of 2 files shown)


asomers:
I audited all of the consumers of the capsicum-rs crate.  None of them use these constants, and they all build OK with a patched libc.  For the main branch, I think we can simply delete these constants.

asomers:
CI is failing because this branch needs https://github.com/rust-lang/libc/pull/4177 .

tgross35 (approved):
Thanks for looking into this. I’ll probably have to pick it and merge together with the other fix.

tgross35:
Aapplied via https://github.com/rust-lang/libc/pull/4187. Not sure why GH didn't close this PR but this should be all set.

PR #4184, opened 2024-12-08
Remove FreeBSD's CAP_UNUSED* and CAP_ALL* constants

asomers:
They aren't stable across OS versions and don't have any legitimate use in applications.

See https://github.com/rust-lang/libc/pull/4183 for the corresponding change to the libc-0.2 branch.

# Description

Remove unstable and non-useful definitions

# Sources

This commit, for example, changed some of these constants in FreeBSD 15.

https://github.com/freebsd/freebsd-src/commit/b165e9e3ea4e327fc421d81c2a89242bd8720780

I was unable to find any applications using these constants with Github code search.  Only libraries like this one that supplied bindings.  I also audited every open source consumer of the capsicum-rs crate to ensure that none were using these constants.

# Checklist


asomers:
I think the failure on aarch64-linux-android is unrelated.

tgross35 (approved):
Thanks for the quick fix

PR #4185, opened 2024-12-08
Lift IFA_ -constants from linux/gnu to linux

jtt:
# Description

Lift IFA_ -constants from linux/gnu to linux/mod.rs 

# Sources

Constants are defined in [linux/if_addr.h](https://github.com/torvalds/linux/blob/master/include/uapi/linux/if_addr.h)

# Checklist

- [ x] No placeholder or unstable values like `*LAST` or `*MAX` are
- [x ] Tested locally (`cd libc-test && cargo test --target mytarget`);


tgross35:
Since these are uapi, could you just lift the existing gnu definitions to Linux/mod.rs?

jtt:
Lifted to constants to `linux/mod.rs` and updated the wording of this PR to reflect that. I would also like to nominate this change to be included in stable branch. @rustbot label stable-nominated

tgross35 (approved):
LGTM, thanks!

PR #4191, opened 2024-12-09
triagebot: Remove JohnTitor from the review rotation

tgross35:
I think you may appreciate some reduced notifications :) let me know if you prefer just to be put in vacation mode.

r? @JohnTitor


JohnTitor (approved):
Yeah, thank you!

PR #4193, opened 2024-12-10
freebsd add more socket TCP stack constants.

devnexen:
[ref](https://man.freebsd.org/cgi/man.cgi?query=tcp)


tgross35 (approved):
LGTM, @asomers could you double check this?

asomers (approved):
These are pretty obscure, but they all look correct to me.  There are dozens more obscure TCP socket options, too.

PR #4194, opened 2024-12-11
Add sockaddr_vm definition for Fuchsia

jamesr:
# Description

This adds a definition of `struct sockaddr_vm` for Fuchsia

# Sources

`man vsock` documents this type on most platforms. The Fuchsia definition matches
Linux except that we do not (currently) have support for flags.

# Checklist

libc-test does not have support for Fuchsia currently.

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -357,6 +357,14 @@ s! {
         pub sin6_scope_id: u32,
     }
 
+    pub struct sockaddr_vm {
+        pub svm_family: sa_family_t,
+        pub svm_reserved1: c_ushort,
+        pub svm_port: crate::in_port_t,
+        pub svm_cid: c_uint,
+        pub svm_zero: [u8; 4],
+    }
+
     pub struct addrinfo {
         pub ai_flags: c_int,
         pub ai_family: c_int,
--- a/libc-test/semver/fuchsia.txt
+++ b/libc-test/semver/fuchsia.txt
@@ -1446,6 +1446,7 @@ sigwait
 sigwaitinfo
 sockaddr_ll
 sockaddr_nl
+sockaddr_vm
 splice
 spwd
 srand


src/fuchsia/mod.rs
```diff
@@ -357,6 +357,14 @@ s! {
         pub sin6_scope_id: u32,
     }
 
+    pub struct sockaddr_vm {
+        pub svm_family: sa_family_t,
+        pub svm_reserved: u16,
+        pub svm_port: crate::in_port_t,
+        pub svm_cid: u32,
```
tgross35, line 364: Usually we use the `c_*` types if that is how they are defined in the source.

```suggestion
        pub svm_reserved: c_short,
        pub svm_port: crate::in_port_t,
        pub svm_cid: c_int,
```

Do you have a link to relevant header for fuchsia?

tgross35:
This should also get added to `libc-test/semver/fuchsia.txt`

tgross35:
This lgtm but I'm not sure what to compare it against. Is the relevant fuchsia header public?

jamesr:
> This lgtm but I'm not sure what to compare it against. Is the relevant fuchsia header public?

https://cs.opensource.google/fuchsia/fuchsia/+/main:src/paravirtualization/lib/vsock/vm_sockets.h;drc=4050ee185f9701a852d764e120bc0886917e3c1a is the most relevant header today. It's public but it is not in the best location - we are still debating exactly where to put it. Regardless of where it is, I don't expect the definition to change (and if it does, I'll update the bindings here to match).

(I tried to reply inline but may have failed - I'm not very practiced with GitHub-ing)

tgross35:
> > https://cs.opensource.google/fuchsia/fuchsia/+/main:src/paravirtualization/lib/vsock/vm_sockets.h;drc=4050ee185f9701a852d764e120bc0886917e3c1a is the most relevant header today. It's public but it is not in the best location - we are still debating exactly where to put it. Regardless of where it is, I don't expect the definition to change (and if it does, I'll update the bindings here to match).
> 
> Thanks, that works for me :) Just trying to keep everything crosslinked since it can can be difficult to find sources for everything down.
> 
> > (I tried to reply inline but may have failed - I'm not very practiced with GitHub-ing)
> 
> I assume it put your comments in a review and wants you to submit them in a batch - It seems to do this sometimes but I have no clue what decides that should happen...

tgross35:
Oh wait, one more thing - it looks like the field is called `svm_reserved1` in Linux and in the headers, but this has it as `svm_reserved`. Should this be updated?

jamesr:
> Oh wait, one more thing - it looks like the field is called `svm_reserved1` in Linux and in the headers, but this has it as `svm_reserved`. Should this be updated?

That seems like a good idea - done. In practice it's rare to ever type this field but consistent is better than inconsistency.

PR #4195, opened 2024-12-13
Update ESP-IDF `c_char` type

SergioGasquez:
# Description

Updates the `c_chart` type for ESP-IDF after https://github.com/rust-lang/rust/pull/132975

# Checklist

  - Still not tested, hence the PR is still in draft

### DIFF
--- a/src/unix/newlib/espidf/mod.rs
+++ b/src/unix/newlib/espidf/mod.rs
@@ -1,7 +1,7 @@
 use crate::prelude::*;
 
 pub type clock_t = c_ulong;
-pub type c_char = i8;
+pub type c_char = u8;
 pub type wchar_t = u32;
 
 pub type c_long = i32;


tgross35 (approved):
Any reason this is marked as a draft?

SergioGasquez:
@ivmarkov wanted to do a small test, that's why I marked it as draft until we confirm that it works, but if you think the change is small enough I can mark it as ready

tgross35:
Happy to have tests! There probably won't be another release until next week anyway.

ivmarkov:
> Happy to have tests! There probably won't be another release until next week anyway.

Tests passed successfully. 
TL;DR: `libc`'s `c_char` and `core::ffi::c_char` should both be `u8`, for all `*-espidf` targets (i.e. both `riscv` as well as `xtensa`).

@SergioGasquez if you could "un-draft" the PR?

Long explanation:
- I checked the Espressif's GCC toolchains (both `xtensa` and `riscv`) and both report the `char` type as **unsigned** by default (`limits.h`'s `CHAR_MAX` is 255).
- I also checked Espressif's `clang` toolchains and there:
  - `clang` with `riscv` also agrees that `CHAR_MAX` is 255 (i.e. **unsigned**)
  - `clang` with `xtensa` still reports `char` as signed, but this is only becausde the Espressif's LLVM+clang `xtensa` fork is not yet updated with the upstream `clang` `xtensa` fix which marks `char` as **unsigned**. There is an open MR to fix the fork to treat `char` as unsigned (just like the upstream code), which will be merged soon

tgross35:
Thanks for confirming!

ivmarkov:
@tgross35 somehow this PR did fall out from the merge queue. Is it something we need to address? Weird.

tgross35:
Thanks for the reminder, I tried adding it back. We just have some tests that are a bit flaky.

ivmarkov:
@tgross35 Sorry to bother - but I think we need this in the 0.2 branch as well.

tgross35:
Indeed, it has the label so it will be in the next release. I was originally planning to do a release next week - but I guess ideally you need it as soon as possible?

Assuming so, I'll try to get one out within the next couple days instead.

ivmarkov:
> Indeed, it has the label so it will be in the next release. I was originally planning to do a release next week - but I guess ideally you need it as soon as possible?
> 
> Assuming so, I'll try to get one out within the next couple days instead.

Yes, ASAP would be deeply appreciated as we are broken since... some time. :-)

Thank you so much!

PR #4196, opened 2024-12-16
freebsd adding further TCP stack related constants.

devnexen:
ref: https://man.freebsd.org/cgi/man.cgi?query=tcp


src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -3778,6 +3778,30 @@ pub const TCP_BBR_USEDEL_RATE: c_int = 1079;
...
+pub const TCP_BBR_TMR_PACE_OH: c_int = 1096;
+pub const TCP_RACK_DO_DETECTION: c_int = 1097;
+pub const TCP_BBR_RACK_RTT_USE: c_int = 1098;
+pub const TCP_BBR_RETRAN_WTSO: c_int = 1099;
+pub const TCP_DATA_AFTER_CLOSE: c_int = 1100;
+pub const TCP_BBR_PROBE_RTT_GAIN: c_int = 1101;
+pub const TCP_BBR_PROBE_RTT_LEN: c_int = 1102;
+pub const TCP_BBR_SEND_IWND_IN_TSO: c_int = 1103;
+pub const TCP_BBR_USE_RACK_RR: c_int = 1104;
+pub const TCP_BBR_USE_RACK_CHEAT: c_int = TCP_BBR_USE_RACK_RR;
```
tgross35, line 3796: We probably only need one of these. It looks like `TCP_BBR_USE_RACK_RR` was added more recently https://github.com/freebsd/freebsd-src/commit/e570d231f48957dcf0757b7b716330f3bd64e362#diff-c354fa9fe15a3e5c90079f38792a8e5458d76677437a53ee7537a62fbc72e100R255.

tgross35:
> ref: https://man.freebsd.org/cgi/man.cgi?query=tcp

Looks like this is the wrong link, these aren't mentioned there. https://github.com/freebsd/freebsd-src/blob/main/sys/netinet/tcp.h is the header file.

One nit but the rest looks fine to me. @asomers could you double check this?

tgross35 (approved):

asomers:
The definitions look correct to me, but they're awfully obscure.  These constants are so obscure that there's a risk of them being deemed unstable private APIs.  That's why I generally don't like to add such things to libc unless there's a real need.

For example, constants added in 9ab890d42b8435517369320ae6a259386049917a later had to be deprecated in ab577960b8545839fad516ee6ae64221a3a57224 .  Do you have an actual need for these constants, or are you just adding them for completeness's sake?

devnexen:
> The definitions look correct to me, but they're awfully obscure. These constants are so obscure that there's a risk of them being deemed unstable private APIs. That's why I generally don't like to add such things to libc unless there's a real need.
> 
> For example, constants added in [9ab890d](https://github.com/rust-lang/libc/commit/9ab890d42b8435517369320ae6a259386049917a) later had to be deprecated in [ab57796](https://github.com/rust-lang/libc/commit/ab577960b8545839fad516ee6ae64221a3a57224) . Do you have an actual need for these constants, or are you just adding them for completeness's sake?

I m mostly interested by BBR constants even tough not all values have necessarily an immediate usefulness and a little bit of completeness yes.

PR #4199, opened 2024-12-17
Fix c_char on various targets

taiki-e:
# Description

This was originally included in the list as a TODO for https://github.com/rust-lang/rust/pull/131319 and should have been done at the same time the c_char definition was updated, just as it was done for https://github.com/rust-lang/rust/pull/122986.

Fixes the following targets:

- aarch64-kmc-solid_asp3
- armv7a-kmc-solid_asp3-eabi
- armv7a-kmc-solid_asp3-eabihf
- riscv64-linux-android
- x86_64-unknown-l4re-uclibc
- armv7-sony-vita-newlibeabihf
- riscv32imac-unknown-nuttx-elf
- riscv32imafc-unknown-nuttx-elf
- riscv32imc-unknown-nuttx-elf
- riscv64gc-unknown-nuttx-elf
- riscv64imac-unknown-nuttx-elf
- thumbv6m-nuttx-eabi
- thumbv7em-nuttx-eabi
- thumbv7em-nuttx-eabihf
- thumbv7m-nuttx-eabi
- thumbv8m.base-nuttx-eabi
- thumbv8m.main-nuttx-eabi
- thumbv8m.main-nuttx-eabihf
- aarch64-unknown-redox
- aarch64-unknown-illumos
- riscv32-wrs-vxworks
- riscv64-wrs-vxworks

# Sources

# Checklist

### DIFF
--- a/src/solid/aarch64.rs
+++ b/src/solid/aarch64.rs
@@ -1,4 +1,4 @@
-pub type c_char = i8;
+pub type c_char = u8;
 pub type wchar_t = u32;
 pub type c_long = i64;
 pub type c_ulong = u64;
--- a/src/solid/arm.rs
+++ b/src/solid/arm.rs
@@ -1,4 +1,4 @@
-pub type c_char = i8;
+pub type c_char = u8;
 pub type wchar_t = u32;
 pub type c_long = i32;
 pub type c_ulong = u32;
--- a/src/unix/linux_like/android/b64/riscv64/mod.rs
+++ b/src/unix/linux_like/android/b64/riscv64/mod.rs
@@ -1,7 +1,7 @@
 use crate::off64_t;
 use crate::prelude::*;
 
-pub type c_char = i8;
+pub type c_char = u8;
 pub type wchar_t = u32;
 pub type greg_t = i64;
 pub type __u64 = c_ulonglong;
--- a/src/unix/linux_like/linux/uclibc/x86_64/mod.rs
+++ b/src/unix/linux_like/linux/uclibc/x86_64/mod.rs
@@ -6,7 +6,7 @@ use crate::prelude::*;
 pub type blkcnt_t = i64;
 pub type blksize_t = i64;
 pub type clock_t = i64;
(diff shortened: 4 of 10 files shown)


taiki-e:
Ideally we should have a top-level c_char definition, but I'm not sure if that is backport friendly.

tgross35:
> Ideally we should have a top-level c_char definition, but I'm not sure if that is backport friendly.

I have https://github.com/rust-lang/libc/pull/4198 in the queue, after which it should be possible to just `pub use crate::arch::c_char_def as c_char` in the relevant `mod.rs`, rather than having the config in a handful of places. Would you be able to rebase this and use that after it lands?

tgross35:
We can't use `core::ffi::c_char` if that is what you meant, since our MSRV is 1.63 but those got stabilized in 1.64. We could make https://github.com/rust-lang/libc/pull/4198 public though; the only reason I added it this way is to do piecemeal updates. After the recently changed targets get fixed, the rest can get double checked and cleaned up.

taiki-e:
I meant to  *define* c_char at the top-level instead of re-exporting.

tgross35:
How would that be different from making https://github.com/rust-lang/libc/pull/4198 public (without `mod arch` and with `c_char_def` -> `c_char`)?

taiki-e:
It should mean the same thing (assuming “making https://github.com/rust-lang/libc/pull/4198 public” means “making c_char_def public as c_char at top-level, remove per-target c_char definitions, and import top-level c_char as needed).

taiki-e:
Closing in favor of https://github.com/rust-lang/libc/pull/4202, which is doing the approach I said I think is ideal in https://github.com/rust-lang/libc/pull/4199#issuecomment-2548226721.

taiki-e:
Re-opening -- Since this is a non-aggressive change, we might put this before the #4202, as pointed out in https://github.com/rust-lang/libc/pull/4201#issuecomment-2548317669.

tgross35 (approved):

PR #4201, opened 2024-12-17
Re-export c_char_def as c_char instead of defining c_char per target

taiki-e:
# Description

Replaces https://github.com/rust-lang/libc/pull/4199 by doing https://github.com/rust-lang/libc/pull/4199#issuecomment-2548178664.

Closes https://github.com/rust-lang/libc/pull/4199.

# Sources

# Checklist

- [ ] Relevant tests in `libc-test/semver` have been updated
- [x] No placeholder or unstable values like `*LAST` or `*MAX` are
  included (see [#3131](https://github.com/rust-lang/libc/issues/3131))
- [ ] Tested locally (`cd libc-test && cargo test --target mytarget`);
  especially relevant for platforms that may not be checked in CI

PR #4202, opened 2024-12-17
Define c_char at top-level and remove per-target c_char definitions

taiki-e:
# Description

Replaces https://github.com/rust-lang/libc/pull/4201

This is the approach I said I think is ideal in https://github.com/rust-lang/libc/pull/4199#issuecomment-2548226721.

It does not have the CI problem https://github.com/rust-lang/libc/pull/4201 has.

# Sources

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -600,11 +600,6 @@ fn test_openbsd(target: &str) {
         }
     });
 
-    cfg.skip_type(move |ty| {
-        // `c_char_def` is always public but not always reexported.
-        ty == "c_char_def"
-    });
-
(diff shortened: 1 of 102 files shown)


taiki-e:
I was concerned about whether this was backport friendly or not, but it turns out that it is not really a problem: #4203

tgross35 (approved):
I double checked everything here, looks like there are no further discrepancies. Pushed a commit that removes the last remaining `c_char_def` skip, which was added in https://github.com/rust-lang/libc/pull/4210.

Thanks for the cleanup!

Issue #3759, opened 2024-06-24
Unnecessary `unsafe` for `libc::major` and `libc::minor`? Or is `libc::makedev` missing unsafe?


VorpalBlade:
* The `libc::major` and `libc::minor` functions are unsafe. The reverse operation `libc::makedev` is not.
* There is no safety invariants on docs.rs for these, this seems to be the norm for libc, and I can see why (you should refer to the man page for the C safety invariants probably).
* However, there is nothing in my (Linux Glibc) man page that describes why they would have safety invariants. Nor do the FreeBSD, NetBSD, OpenBSD or Illumos man pages have anything along those lines.
* Most implementations are in pure rust and use no unsafe operations. The sole exception seems to be Illumos where C functions are being called. Here `makedev` is also unsafe, unlike other platforms.
* Illumos can apparently return NODEV, but that is a safe failure (with a corresponding errno being set). 

I have not been able to check the man page of every single Unix system, so maybe there is one where this makes sense. Otherwise I think this is an oversight.

I think one or both of the following should be done:
* If it is not an oversight, since it is safe on the majority of Unices, it would be good to add a documentation string that explains why it is unsafe on some.
* Also, I would expect some consistency between `makedev` and `major`/`minor` in how unsafe vs not unsafe is handled across platforms. 
   * Is the idea that functions should only be unsafe on platforms where a C function is called? Then `major`/`minor` are wrong.
   * Is the idea that functions should be unsafe if it they call C functions on any platforms? Then `makedev` is wrong. I can see the argument to not get lints about unnecessary unsafe blocks, but that only works if it is done consistently.

I also very much doubt there is *any* Unix where any of these functions *could* reasonably be unsafe. After all, even the C function will only do some arithmetic. 

Minimum working example (not that I think it makes sense here): 

```rust
/// To be called with value from std::os::unix::fs::MetadataExt::rdev
fn split_to_major_minor(rdev: u64) -> (u64, u64) {
    // SAFETY: I don't think there *actually* are any safety invariants here
    (unsafe { libc::major(rdev) } as u64, unsafe { libc::minor(rdev) } as u64)
}
```

Target triplet: x86_64-unknown-linux-gnu (and several other ones I checked)

PR #4208, opened 2024-12-18
Make `makedev`, `major`, `minor` const

igankevich:
# Description

This PR marks all `makedev`, `major`, `minor` re-implementations (except Solarish) as `const fn` and tests `major` and `minor` against their C counterparts (macros). This PR closely follows the work that was done in this [commit](https://github.com/rust-lang/libc/commit/5e6d9c4a92).

The motivation behind this PR is to make Rust code that uses these three functions consistent across different platforms. To give a concrete example: currently one has to wrap `major` and `minor` in `unsafe` blocks on MacOS, whereas on Linux these blocks are not needed. Unnecessary `unsafe` blocks produce a warning that one silences with `#[allow(unused_unsafe)]`. All of that confuses code reviewers and creates unnecessary noise. `makedev`, `major` and `minor` are macros on most Unixes, there is no need for them to be unsafe.

As far as I understand these changes do not break existing code: changing unsafe functions to const only produce warnings but not compilation errors.

# Sources

None.

# Checklist

### DIFF
--- a/libc-test/test/makedev.rs
+++ b/libc-test/test/makedev.rs
@@ -1,5 +1,49 @@
-//! Compare libc's makdev function against the actual C macros, for various
+//! Compare libc's makedev, major, minor functions against the actual C macros, for various
 //! inputs.
+
+#[cfg(any(target_os = "solaris", target_os = "illumos"))]
+mod ret {
+    pub type MajorRetType = libc::major_t;
+    pub type MinorRetType = libc::minor_t;
+}
+
+#[cfg(any(
+    target_os = "linux",
+    target_os = "l4re",
+    target_os = "emscripten",
+    target_os = "fuchsia",
+    target_os = "aix",
+    target_os = "nto",
+    target_os = "hurd",
+    target_os = "openbsd",
+))]
+mod ret {
+    pub type MajorRetType = libc::c_uint;
+    pub type MinorRetType = libc::c_uint;
+}
+
+#[cfg(any(
+    target_os = "android",
+    target_os = "dragonfly",
+    target_os = "netbsd",
+    target_os = "freebsd",
+))]
+mod ret {
+    pub type MajorRetType = libc::c_int;
+    pub type MinorRetType = libc::c_int;
+}
+
+#[cfg(any(
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "tvos",
+    target_os = "watchos",
+    target_os = "visionos"
+))]
+mod ret {
+    pub type MajorRetType = i32;
+    pub type MinorRetType = i32;
+}
 
 #[cfg(any(
     target_os = "android",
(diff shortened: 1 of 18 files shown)


igankevich:
After testing with the CI I found and fixed a bug in emscriten's `major` and `minor`: bit shifts and masking were done in the reverse order. The original macros are [here](https://github.com/emscripten-core/emscripten/blob/main/system/lib/libc/musl/include/sys/sysmacros.h), the same link is in the crate's source code.

tgross35:
This also changes from unsafe to safe, which should be fine and fixes https://github.com/rust-lang/libc/issues/3759.

@hoodmane could you double check the emscripten changes?

libc-test/test/makedev.rs
```diff
@@ -16,11 +16,17 @@ mod t {
...
 
     fn compare(major: c_uint, minor: c_uint) {
-        let expected = unsafe { makedev_ffi(major, minor) };
-        assert_eq!(libc::makedev(major, minor), expected);
+        let dev = unsafe { makedev_ffi(major, minor) };
+        assert_eq!(libc::makedev(major, minor), dev);
+        let major = unsafe { major_ffi(dev) };
+        assert_eq!(libc::major(dev), major as _);
+        let minor = unsafe { minor_ffi(dev) };
+        assert_eq!(libc::minor(dev), minor as _);
```
tgross35, line 29: Could you name the types here rather than using `_`?

src/unix/linux_like/emscripten/mod.rs
```diff
@@ -1455,41 +1455,41 @@ f! {
     pub fn CPU_EQUAL(set1: &cpu_set_t, set2: &cpu_set_t) -> bool {
         set1.bits == set2.bits
     }
+}
+
+safe_f! {
+    pub {const} fn makedev(major: c_uint, minor: c_uint) -> crate::dev_t {
```
hoodmane, line 1461: Emscripten has actual functions for these in addition to macros so it would be possible to call out to them. But it would be a lot slower than this.

hoodmane (approved):
Maybe it would be nice to keep the bitwise or terms of the rust code in the same order as they appear in musl for easier validation?

hoodmane:
Otherwise it looks good to me.

libc-test/test/makedev.rs
igankevich: I wish I could, but the return type can be `c_int`, `c_uint`, `major_t`, `minor_t` depending on the platform.

igankevich:
> Maybe it would be nice to keep the bitwise or terms of the rust code in the same order as they appear in musl for easier validation?

Thanks for pointing this out! I've changed the order. Kindly re-requesting the review :pray:

libc-test/test/makedev.rs
tgross35: Sorry for the late review, but could you change both the LHS and RHS to be `as i64`? At least that gives a concrete size so we don't have any chance of wrapping (in case some obscure platform has a `major_t` smaller than `int`).

tgross35 (approved):
I have one request but otherwise lgtm

libc-test/test/makedev.rs
igankevich: I tried casting them to `i64` and it fails on FreeBSD. Ended up casting to proper target-specific types.

tgross35:
Thanks for the updates, that's even better. Please squash and this should be ready to merge.

igankevich:
> Thanks for the updates, that's even better. Please squash and this should be ready to merge.

Done.

Issue #4209, opened 2024-12-18
OpenBSD is broken due to char_def


semarie:
My daily CI on OpenBSD (x86_64) failed (testing 5f41b7a3b3)

As several PR regarding `char_def` was merged, I am unsure which one caused the regression.

Cc: @taiki-e @tgross35 

```
Wed Dec 18 08:00:11 CET 2024: run_test: build and run the testsuite
+ mirrors_url=https://ci-mirrors.rust-lang.org/libc
+ target=x86_64-unknown-openbsd
+ export RUST_BACKTRACE=1
+ [ -n  ]
+ cmd=cargo test --target x86_64-unknown-openbsd 
+ cmd=cargo test --target x86_64-unknown-openbsd  --workspace
+ [ x86_64-unknown-openbsd = s390x-unknown-linux-gnu ]
+ cargo test --target x86_64-unknown-openbsd --workspace --no-default-features
    Updating crates.io index
     Locking 42 packages to latest compatible versions
   Compiling proc-macro2 v1.0.92
   Compiling unicode-ident v1.0.14
   Compiling libc v0.2.168
   Compiling serde v1.0.216
   Compiling memchr v2.7.4
   Compiling cfg-if v0.1.10
   Compiling dirs-sys v0.3.7
   Compiling semver v1.0.24
   Compiling serde_json v1.0.133
   Compiling dirs v2.0.2
   Compiling ryu v1.0.18
   Compiling term v0.6.1
   Compiling quote v1.0.37
   Compiling itoa v1.0.14
   Compiling unicode-xid v0.2.6
   Compiling libc v1.0.0-alpha.1 (/data/semarie/repos/rust/libc-rs)
   Compiling syn v2.0.90
   Compiling aho-corasick v1.1.3
   Compiling bitflags v1.3.2
   Compiling regex-syntax v0.8.5
   Compiling shlex v1.3.0
   Compiling serde_derive v1.0.216
   Compiling log v0.4.22
   Compiling cc v1.2.4
   Compiling regex-automata v0.4.9
   Compiling rustc_version v0.4.1
   Compiling regex v1.11.1
   Compiling garando_pos v0.1.0
   Compiling garando_errors v0.1.0
   Compiling garando_syntax v0.1.1
   Compiling ctest2 v0.4.9
   Compiling libc-test v0.1.0 (/data/semarie/repos/rust/libc-rs/libc-test)
The following warnings were emitted during compilation:

[...]

PR #4210, opened 2024-12-18
Skip `c_char_def` on OpenBSD

tgross35:
Fixes https://github.com/rust-lang/libc/issues/4209


semarie:
it fixes the build on OpenBSD.

PR #4212, opened 2024-12-18
Declare `setdomainname` and `getdomainname` on Android.

sunfishcode:
# Description

Declare `setdomainname` and `getdomainname` on Android.

# Sources

Android [supports] `setdomainname` and `getdomainname` in API level 26.

[supports] https://github.com/aosp-google/bionic/blob/28f9101d76b709febe25977f98530d77580387d1/libc/include/unistd.h#L236

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2081,6 +2081,9 @@ fn test_android(target: &str) {
             // Added in API level 26, but some tests use level 24.
             "endgrent" => true,
 
+            // Added in API level 26, but some tests use level 24.
+            "getdomainname" | "setdomainname" => true,
+
             // FIXME: bad function pointers:
             "isalnum" | "isalpha" | "iscntrl" | "isdigit" | "isgraph" | "islower" | "isprint"
             | "ispunct" | "isspace" | "isupper" | "isxdigit" | "isblank" | "tolower"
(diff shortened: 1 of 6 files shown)


tgross35:
Changes lgtm, but could you instead add this to `src/unix/linux_like/mod.rs` and remove them from `linux` and `emscripten`?

@maurer could you double check this?

maurer:
LGTM, these two functions are in fact available since API level 26. Not aware of any special caveats for Android here other than that apps don't have `CAP_SYS_ADMIN`, so only platform code would be able to actually call the `set` side of this.

tgross35 (approved):

tgross35:
Actually sorry, one request - since all `linux_like` OSs have this now, could you instead add the functions `linux_like/mod.rs` and remove from `linux_like/linux/mod.rs` and `linux_like/emscripten/mod.rs`? Same changes for the semver files.

sunfishcode:
Which semver file corresponds to "linux_like"? Is it linux.txt? I can't tell whether that's used for Android or not.

tgross35:
Ah you're right, we don't have one for `linux_like`. Leaving the existing `.txt` files unchanged and adding to `android.txt`, like you currently have, is probably the correct thing to do for the semver tests.

(this test setup needs improvement at some point...)

tgross35 (approved):
Thanks!

sunfishcode:
This failed on the arch64-linux-android build, and it's not clear what's wrong.

tgross35:
I wish GH would send a notification for failed CI after merge. It was probably our spurious android failure, I'm not sure why it happens.

PR #4213, opened 2024-12-19
Apply modulo 256 to BSD WEXITSTATUS

krobelus:
wait(2p)[^1] says

> WEXITSTATUS(status)
>         If WIFEXITED(status) is true, evaluates to the low-order 8 bits
>         of the argument passed to _exit(2) or exit(3) by the child.

meaning WEXITSTATUS(status) is an 8-bit value.  We accidentally return
too many bits. For example WEXITSTATUS(-1) returns -1 instead of 255.

Fix it, matching the C library and our other WEXITSTATUS implementations.

[^1] https://manpage.me/index.cgi?apropos=0&q=wait&sektion=2&manpath=FreeBSD+12-CURRENT+and+Ports&arch=default&format=html

Originally reported at https://github.com/fish-shell/fish-shell/issues/10919

### DIFF
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -639,7 +639,7 @@ safe_f! {
     }
 
     pub {const} fn WEXITSTATUS(status: c_int) -> c_int {
-        status >> 8
+        (status >> 8) & 0x00ff
     }
 
     pub {const} fn WCOREDUMP(status: c_int) -> bool {


tgross35 (approved):
Thanks, relevant headers for NetBSD https://github.com/NetBSD/src/blob/75dc4ba6f0d86305c83dc5c10ec2f0d96eafe807/sys/sys/wait.h#L67 and FreeBSD https://github.com/freebsd/freebsd-src/blob/16f0d01f9ca1e28bede9a493329c5d66e317d88f/contrib/bmake/wait.h#L48.

faho:
@tgross35 The auto-merge seems to have failed here for what appears to be unrelated reasons (adb daemon not starting on android). Could you kick it off again?

tgross35:
CI is temporarily blocked due to https://github.com/rust-lang/libc/pull/4254 and an upstream issue with FreeBSD. Everything that has a checkmark will get merged once that is resolved.

PR #4203, opened 2024-12-17
v0.2: Backport c_char fixes

taiki-e:
# Description

Backports the following to libc-0.2:
- https://github.com/rust-lang/libc/pull/4195
- https://github.com/rust-lang/libc/pull/4198
- https://github.com/rust-lang/libc/pull/4200
- https://github.com/rust-lang/libc/pull/4202

4200 is not actually c_char-related PR, but 4202 depend on import change made in 4200.

# Sources

# Checklist

- [ ] Relevant tests in `libc-test/semver` have been updated
- [x] No placeholder or unstable values like `*LAST` or `*MAX` are
  included (see [#3131](https://github.com/rust-lang/libc/issues/3131))
- [ ] Tested locally (`cd libc-test && cargo test --target mytarget`);
  especially relevant for platforms that may not be checked in CI

PR #4214, opened 2024-12-19
[0.2] Define c_char at top-level and remove per-target c_char definitions

tgross35:
Backport https://github.com/rust-lang/libc/pull/4202

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -644,11 +644,6 @@ fn test_openbsd(target: &str) {
         }
     });
 
-    cfg.skip_type(move |ty| {
-        // `c_char_def` is always public but not always reexported.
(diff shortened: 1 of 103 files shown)


taiki-e:
As for 0.2, we also need to remove c_char from src/psp.rs, like https://github.com/rust-lang/libc/pull/4203/files#diff-0abdf4e9d0e64b993ee63247b3ed855cbf4c6e795b7982a371e9576b78f61c50.

tgross35:
Thanks, added 50f82adc3e79f1aed1d319d268986df1cf1b6cf0 to address this.

tgross35:
Merge already, or at least notify me when you don't want to...

PR #4216, opened 2024-12-21
android: use proper types for dirent.d_ino, dirent.d_off, stat.st_mode and stat64.st_mode

thunder-coding:
This bug was noticed in https://github.com/termux/termux-packages/pull/22609

# Description

Fixes types for `dirent.d_ino`, `dirent.d_off`, `stat.st_mode` and `stat64.st_mode` on Android.

The hardcoded types are incorrect and this PR changes them to how they are declared for Linux. Also fixes the incorrect type errors in https://github.com/termux/termux-packages/pull/22609, where I found the bug (The builds are still failing due to missing `posix_spawn*` support here, but that's a different issue.)

This patch should likely also be backported to 0.2.x release, I have the patchset already applied on my other branch. Will link another PR soon.

# Sources

Not needed?

# Checklist

  I tried building the tests locally, it seems like the tests are broken for Android. But I was able to fix some of the issues in https://github.com/termux/termux-packages/pull/22609


tgross35:
Thanks for the detailed sources. Is the following delta accurate?

- `unsigned int` -> `unsigned short` (32 -> 16) for `mode_t` on 32 bit
- `unsigned long` -> `uint64_t` (32 -> 64) for `ino_t` on 32 bit
- `long` -> `int32_t` (32 -> 32, nop) for `off_t` on 32 bit

These issues seem reasonably severe so I am surprised it has not been noticed before (@maurer do you know of anything?). I guess the other fields at least keep the relevant fields here correctly aligned?

> ~This patch should likely also be backported to 0.2.x release, I have the patchset already applied on my other branch. Will link another PR soon.~ It seems that this patch applies cleanly to libc-0.2 branch, I was backporting to an older release earlier. `@rustbot label stable-nominated` should be it according to CONTRIBUTING.md

For reference, the command can't be in backticks :)

src/unix/linux_like/android/mod.rs
```diff
@@ -528,16 +526,16 @@ s_no_extra_traits! {
     }
 
     pub struct dirent {
-        pub d_ino: u64,
-        pub d_off: i64,
+        pub d_ino: crate::ino_t,
```
maurer, line 529: This is not correct, `ino_t` is [only used when `LP64`](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/dirent.h;l=61-69?q=d_ino&ss=android%2Fplatform%2Fsuperproject%2Fmain:bionic%2Flibc%2Finclude%2F&start=1) is enabled (int is 32-bit, long int is 64-bit, pointers are 64-bit).

In all other cases (mostly 32-bit CPUs), this is an *explicit* `u64`, not an `ino_t`.

We could split `dirent` by architecture if we really wanted to use `ino_t` here, but on LP64 platforms, `ino_t` happens to be `u64` which is why this was working.

src/unix/linux_like/android/mod.rs
```diff
@@ -528,16 +526,16 @@ s_no_extra_traits! {
...
+        pub d_off: crate::off_t,
```
maurer, line 530: This either needs to remain unmodified or become `off64_t` - it is [unconditionally 64-bits on android](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/dirent.h;l=69?q=d_ino&ss=android%2Fplatform%2Fsuperproject%2Fmain:bionic%2Flibc%2Finclude%2F&start=1)

src/unix/linux_like/android/mod.rs
```diff
@@ -528,16 +526,16 @@ s_no_extra_traits! {
...
         pub d_reclen: c_ushort,
         pub d_type: c_uchar,
         pub d_name: [c_char; 256],
     }
 
     pub struct dirent64 {
-        pub d_ino: u64,
-        pub d_off: i64,
+        pub d_ino: crate::ino64_t,
```
maurer, line 537: Same comment applies here - `dirent64` and `dirent` [are the same struct](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/dirent.h;l=74-77?q=d_ino&ss=android%2Fplatform%2Fsuperproject%2Fmain:bionic%2Flibc%2Finclude%2F&start=1) on Android.

maurer (changes requested):
Fixing up `mode_t`/`ino_t` might be worth doing, but the changes you're making to these structs will actively break things on certain platforms, as I've enumerated.

You can't just assume that `dirent`/`stat` & co will have the same layout they have in other environments.

You can't reset `ino_t` to be `u64` because it got represented as `u64` in `dirent` - it's used in other places, e.g. `FTSENT` where it must have its original value. While `libc` doesn't bind this struct, other crates do, and they use `libc`'s `ino_t` type assuming that it matches the C variant.

src/unix/linux_like/android/b32/mod.rs
```diff
@@ -52,7 +54,7 @@ s! {
         pub st_dev: c_ulonglong,
         __pad0: [c_uchar; 4],
         __st_ino: crate::ino_t,
-        pub st_mode: c_uint,
+        pub st_mode: crate::mode_t,
```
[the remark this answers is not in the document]
thunder-coding: I made this mode_t to allow code using rust libc to just use the time mode_t to store the type like on other *nix platforms instead of having to deal with Android in other way.

thunder-coding:
> You can't reset ino_t to be u64 because it got represented as u64 in dirent - it's used in other places, e.g. FTSENT where it must have its original value. While libc doesn't bind this struct, other crates do, and they use libc's ino_t type assuming that it matches the C variant.

Oh yeah, I didn't look into other uses of ino_t in FTSENT. I don't think I am capable of taking this PR further. This is just way too complicated for me. Feel free to take over this

robertkirkman:
Unfortunately, it turned out that this PR did not successfully work at run-time in a real-world 32-bit Android situation:

- https://github.com/termux/termux-packages/issues/24741

My personal belief is that in this case, the original errors (in the Fish project) should actually be considered the responsibility of the Fish side of the code, and that the correct location to apply patches is in the Fish code, so if, for example, the Fish project does not officially intend to support Android, then termux-packages should apply downstream patches to the Fish-specific code, not to the rust-lang/libc cargo crate.

PR #4220, opened 2024-12-25
Rewrite `style.rs` using `syn`

### LINKED ISSUE #4109: Rewrite `ci/style.rs` using `syn`
The [style checking script](https://github.com/rust-lang/libc/blob/f92da8e27f9c9d3554dc97a3b2a583209074f939/ci/style.rs) has been about the same since it was introduced in 2015, but its text-based parsing seems to make it get in the way a lot of the time.

Ideally it would be rewritten using [`syn`](https://docs.rs/syn/latest/syn/), which should be more accurate and not have problems e.g. handling newlines or recognizing when we are within `cfg_if`. It could become part of the `libc-test` crate, so `cargo run libc-test` checks style.

rmehri01:
# Description

- Use `syn` instead of text-based parsing since it's more accurate and easier to work with
- Move `style.rs` to `libc-test` and make it run as a part of `cargo test`, as well as cleaning it up a bit and adding tests for the style checker implementation itself
- Use a visitor that will walk over all the items in the file, including within `cfg_if!` macros, and tries to apply the same logic that the previous `style.rs` script did
- Fix some lints that came up as a result of rewriting the script
- Added some new rules for dealing with `s!` macros

Closes #4109 

# Sources

# Checklist


tgross35:
Huge thanks for working on this!! I’ll take a look in a couple of days :)

libc-test/test/style.rs
```diff
@@ -0,0 +1,357 @@
...
+        //     self.s_macros += 1;
+        //     (self.on_err)(line, "multiple s! macros in one module");
+        // }
+
+        self.state = new_state;
+    }
+
+    /// Visit the items inside the [CfgExprIf] and return the final
+    /// state after, which is conservatively the minimum of all branches
+    /// since the branches could have diverged in state.
```
rmehri01, line 175: Not sure what the best way to handle the state while branching is, the old `style.rs` script didn't consider branching at all but only got away with it here because it thought it was in an `impl` due to this comment:

https://github.com/rust-lang/libc/blob/48692735ab221c30e63e413ede29272b406dfd91/src/fixed_width_ints.rs#L45

I think the two options are to keep the same behavior as the old script which would mean this `cfg_if!` block would have to be separated into two (duplicating the `#[cfg(...)]` or to do some sort of join with the min or max of the states of all the branches.

libc-test/test/style.rs
```diff
@@ -0,0 +1,357 @@
...
+        }
+    };
+}
+
+#[test]
+fn check_style() {
+    if env::var("LIBC_CI").is_ok() {
+        // we already run this in the style.sh script
+        return;
+    }
```
rmehri01, line 54: CI is failing since this env var doesn't seem to be set. It seems like it might be specific to android, running it locally I get:

```
WARNING: linker: /data/local/tmp/style-d96f44504d869edb: unsupported flags DT_FLAGS_1=0x8000001
```

libc-test/test/style.rs
tgross35: Would it be possible to treat each arm as if each item has the `#[cfg(...)]` (or `#[cfg(not(...))]` for the default directly applied?

I think if you save the state before entering each `cfg_if` branch and then restore it after, things will probably work out okay.

We might want to adjust this behavior later so as long as existing files pass, I'm not too concerned.

tgross35 (commented):
Sorry that took a little longer than expected. Overall this looks great, most of the comments I left relate to things we can probably upgrade now since the original is from 2016.

One meta thing - maybe it's better to put this into a new module of `libc-test/src/` and then only run it in `libc-test/test`. Then you could add a new test file to check this implementation itself, just have a bunch of examples in strings and assert they pass/fail.

libc-test/test/style.rs
tgross35: Also in general, I'm not really concerned about whether configuration happens via `cfg_if` or `cfg`. The original checker was pretty strict about this, but IMO I'd rather see `#[cfg(...)]` if it's a one-time thing.

The more important thing is have it yell at you and suggest `cfg_if` if the same `cfg` is repeated, either in multiple `#[cfg(...)]`s or in multiple `cfg_if`s. But that's something that could be added later.

rmehri01: > Would it be possible to treat each arm as if each item has the #[cfg(...)] (or #[cfg(not(...))] for the default directly applied?

Sorry, what do you mean by this, what is the default that needs to be directly applied?

> I think if you save the state before entering each `cfg_if` branch and then restore it after, things will probably work out okay.
> 
> We might want to adjust this behavior later so as long as existing files pass, I'm not too concerned.

Yeah makes sense, let's do that for now

> The more important thing is have it yell at you and suggest `cfg_if` if the same `cfg` is repeated, either in multiple `#[cfg(...)]`s or in multiple `cfg_if`s. But that's something that could be added later.

Sounds good, I'd be happy to help out with adding some of these new rules that are easier with `syn` but I just want to work on getting the basics down first :)

ci/style.sh
```diff
@@ -9,7 +9,7 @@ if [ -n "${CI:-}" ]; then
     check="--check"
 fi
 
-rustc ci/style.rs && ./style src
+cargo test --manifest-path libc-test/Cargo.toml --test style
```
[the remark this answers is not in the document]
rmehri01: Good point, though I think this conflicts with the root directory arg that is passed in. Is there anyway we can have both or maybe we should just always run from `src`?

https://github.com/rust-lang/libc/blob/55476a052cd37b1855b05767333ecc569cf87ee9/libc-test/test/style.rs#L56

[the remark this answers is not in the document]
tgross35: I don't think there is any need to retain the ability to pass a path, especially if it can figure out the path relative to `CARGO_MANIFEST_DIR`. Wouldn't the test harness be eating that arg anyway

libc-test/test/style.rs
tgross35: > > Would it be possible to treat each arm as if each item has the #[cfg(...)] (or #[cfg(not(...))] for the default directly applied?
> 
> Sorry, what do you mean by this, what is the default that needs to be directly applied?

I missed a `)`, it should have been `(or #[cfg(not(...))] for the default)`referring to the last `else` condition. But I'm realizing this doesn't make a whole lot of sense anyway, so just ignore that comment for now.

> > The more important thing is have it yell at you and suggest `cfg_if` if the same `cfg` is repeated, either in multiple `#[cfg(...)]`s or in multiple `cfg_if`s. But that's something that could be added later.
> 
> Sounds good, I'd be happy to help out with adding some of these new rules that are easier with `syn` but I just want to work on getting the basics down first :)

This is much appreciated :) the tooling here has been in need of a refresh for a while!

ci/style.sh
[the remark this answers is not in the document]
rmehri01: Sounds good! It seems like getting the args in the test takes priority over the test harness so it could be useful for just checking a subdir but I don't think this is very common anyways

libc-test/test/style.rs
rmehri01: Nice, that seems to work! :partying_face:

rmehri01:
> One meta thing - maybe it's better to put this into a new module of `libc-test/src/` and then only run it in `libc-test/test`. Then you could add a new test file to check this implementation itself, just have a bunch of examples in strings and assert they pass/fail.

I couldn't quite get it to work with `libc-test/src/` since there were some linking errors, I think having to do with using `cc` in the build script. For now I just moved the common functionality into `tests/style/mod.rs` similar to what is suggested here so that we can test the style checker implementation as well: https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html.

libc-test/test/style.rs
```diff
@@ -0,0 +1,357 @@
...
+            "lib.rs" | "macros.rs" => continue,
+
+            _ => {}
+        }
+
+        let mut contents = String::new();
+        t!(t!(fs::File::open(&path)).read_to_string(&mut contents));
+
+        let file = t!(syn::parse_file(&contents));
+        StyleChecker::new(|line, msg| err.error(&path, line, msg)).visit_file(&file);
```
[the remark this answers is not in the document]
rmehri01: For the bonus points I added `annotate_snippets` for nicer error reporting, it adds a dependency but I think it's much nicer to read:

image

rmehri01:
Sorry for cramming so much into one pr but I think I came up with some decent rules for doing the `#[cfg(...)]` checking, basically just checking if there are duplicates and standalone positive `#[cfg(...]` attributes instead of requiring that `cfg_if!` is used everywhere:

https://github.com/rust-lang/libc/blob/60c840e039166c7340954c2ac58e35a041fe5ea4/libc-test/test/style/mod.rs#L22-L26

Here are some examples, mostly taken from what you mentioned in the issue/review comments:

https://github.com/rust-lang/libc/blob/60c840e039166c7340954c2ac58e35a041fe5ea4/libc-test/test/style_tests.rs#L135-L260

libc-test/test/style/mod.rs
```diff
@@ -0,0 +1,465 @@
...
+/// as the condition instead of [syn::Expr].
+struct ExprCfgIf {
+    _attr: syn::Attribute,
+    then_branch: Vec<syn::Item>,
+    else_branch: Option<Box<ExprCfgElse>>,
+}
+
+enum ExprCfgElse {
+    Block(Vec<syn::Item>),
+    If(ExprCfgIf),
```
tgross35, line 85: Just some suggested docs

```suggestion
    /// Final block with no condition `else { /* ... */ }`.
    Block(Vec<syn::Item>),
    /// `else if { /* ... */ }` block.
    If(ExprCfgIf),
```

ci/run.sh
```diff
@@ -101,25 +102,25 @@ if [ "$target" = "s390x-unknown-linux-gnu" ]; then
     passed=0
     until [ $n -ge $N ]; do
         if [ "$passed" = "0" ]; then
-            if $cmd --no-default-features; then
+            if eval "$cmd --no-default-features -- $test_flags"; then
```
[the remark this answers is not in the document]
rmehri01: I think stylecheck isn't happy about it:

```
In ci/run.sh line 105:
            if $cmd --no-default-features -- $test_flags; then
                                             ^---------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
            if $cmd --no-default-features -- "$test_flags"; then

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

rmehri01:
> Sorry this took a little while, usually I sweep this repo about once a week. Overall, looks great! I left a handful of mostly stylistic suggestions, nothing major.
> 
> After you make any changes please just clean up the history and we should be able to get this in.

No problem at all! I just addressed the comments and cleaned up the history :)

ci/run.sh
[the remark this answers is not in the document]
tgross35: Using `eval` always resplits so it doesn't really sidestep any possible unquoted arg issues :) but there shouldn't be any here, it's fine to just `# shellcheck disable=SC2086` on the relevant lines.

(I wish vanilla sh had arrays...)

tgross35:
I'm not quite sure why emscripten is failing, seems like it's failing to run the C build? I don't know why that would be, seems like it should be unrelated.

tgross35 (approved):
In any case lgtm, thanks for getting this done! I'll try to figure out if our emscripten CI just broke somehow, unless you have any ideas.

rmehri01:
> I'm not quite sure why emscripten is failing, seems like it's failing to run the C build? I don't know why that would be, seems like it should be unrelated.

Yeah I wasn't sure about this either, it was passing earlier and I don't think addressing the review comments should have touched anything related to the build script.

rmehri01:
Just reran it and it passes now so maybe it was a flake?

tgross35:
> Just reran it and it passes now so maybe it was a flake?

That's lucky 🎉. Thanks again for making this happen!

PR #4221, opened 2024-12-25
Haiku bsd

nielx:
# Description

Haiku implements various parts of the (non-POSIX) BSD API. This moves it to a separate file, for easier future maintenance. It also adds some missing functionality that is part of other platforms and that may be useful for other crates.

Note that this synchronizes with Haiku R1 Beta 5, which is the latest relaase.

# Sources

See https://github.com/haiku/haiku/tree/r1beta5/headers/compatibility/bsd/ for all the headers that match up with the ones in the `bsd.rs` file.

# Checklist

Note: `semver` is not used by Haiku (yet), so N/A. The test currently fails, but in a panic! in the garando_syntax crate. The panic is the same before and after the changes.

### DIFF
--- a/src/unix/haiku/bsd.rs
+++ b/src/unix/haiku/bsd.rs
@@ -22,6 +22,17 @@ s! {
         pub sl_cur: size_t,
     }
 
+    // sys/event.h
+    pub struct kevent {
+        pub ident: crate::uintptr_t,
+        pub filter: c_short,
+        pub flags: c_ushort,
+        pub fflags: c_uint,
+        pub data: i64,
+        pub udata: *mut c_void,
+        pub ext: [u64; 4],
+    }
+
     // sys/link_elf.h
     pub struct dl_phdr_info {
         pub dlpi_addr: crate::Elf_Addr,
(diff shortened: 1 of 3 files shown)


src/unix/haiku/bsd.rs
```diff
@@ -0,0 +1,151 @@
+use crate::prelude::*;
+
+// This file contains the BSD APIs available in Haiku. It corresponds to the
+// header files in `headers/compatibility/bsd`.
+//
+// Note that Haiku's BSD compatibility is a combination of system APIs and
+// utility libraries. There should only be system APIs in `libc`. When you are
+// trying to determine whether something should be included in this file, the
+// best indicator is whether it also exists in the BSD-specific definitions in
+// this libc crate.
```
tgross35, line 10: Would you mind turning this into `//!` module doc comments since it's moving anyway?

tgross35 (approved):
One suggested nit, LGTM otherwise. Thanks for doing some cleanup here.

src/unix/haiku/bsd.rs
nielx: I am not sure if that would show up? The contents of the bsd.rs file is imported into the main module, so I am not sure if rustdoc would pick this up.

tgross35: Technically it will show up with `--document-private-items` and in LSP for the module, but yeah that doesn't come up very much. I just asked for style/consistency with the rest of the code, all the other files with module-level docs have them as doc comments.

tgross35: @nielx mind making this small consistency change? Should be otherwise ready to go.

PR #4224, opened 2024-12-28
fix: typos

chloefeal:
I hope my corrections will help you. Thank you for your work.

### DIFF
--- a/src/unix/nto/x86_64.rs
+++ b/src/unix/nto/x86_64.rs
@@ -69,7 +69,7 @@ s! {
         reserved2: [u8; 224],
     }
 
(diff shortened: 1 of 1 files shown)


tgross35:
Thanks for the PR, could you provide a link to the headers where these are defined?

chloefeal:
Sorry, maybe I am disunderstand.

tgross35:
The correction is definitely welcome if it makes things more consistent with NTO's sources, but if they have that typo then I think we unfortunately need to keep the same typo.

chloefeal:
@tgross35 Agree with you. It's better to make fix to upstream.

PR #4228, opened 2025-01-05
linux/android: deprecate obsolete kernel module syscalls

devnexen:
they were functional up 2.6.

[create_module](https://man7.org/linux/man-pages/man2/create_module.2.html) [query_module](https://man7.org/linux/man-pages/man2/query_module.2.html) [get_kernel_syms](https://man7.org/linux/man-pages/man2/get_kernel_syms.2.html)


maurer (approved):
Fine from Android PoV - a quick spot check shows these are in fact all syscalls removed in 2.6, and even the earliest Android kernels were past 2.6.

tgross35 (approved):

PR #4238, opened 2025-01-12
feat: add syslog macros LOG_MASK/LOG_UPTO

SteveLauC:
# Description

Adds syslog C macros to all UNIX targets

1. LOG_MASK
2. LOG_UPTO

They are re-implemented as Rust functions, and tests are added to ensure our Rust implementations have the same behavior as the C macro.

# Sources

Link to the musl impl: https://github.com/bminor/musl/blob/61399d4bd02ae1ec03068445aa7ffe9174466bfd/include/syslog.h#L23-L24

# Checklist


SteveLauC:
CI failure seems unrelated, it is due to a failed network call

SteveLauC:
Ok, now, the CI errors are really related to my patch, the wasi CIs failed because there is no `sys/syslog.h` header file, I am no WebAssembly expert, should I just simply disable the cc part and tests under these targets?

tgross35:
> Ok, now, the CI errors are really related to my patch, the wasi CIs failed because there is no `sys/syslog.h` header file, I am no WebAssembly expert, should I just simply disable the cc part and tests under these targets?

That is probably pretty reasonable. Does it work if `cc::Build::new().file("src/syslog.c").compile("syslog");` moves one block up, within the `linux || android || emscripten || fuchsia || bsd` block?

tgross35:
@SteveLauC I'm taking a look through older-ish PRs, any chance you could do that small change so we can merge this?

tgross35:
It's been over a year and this is out of date, so I'm going to close it. Please feel free to reopen or resubmit if you get back to it!

Issue #3880, opened 2024-08-29
Discussion about removing `extra_traits` in 1.0


tgross35:
It has been proposed to remove the `Eq`, `Hash`, and `Debug` implementations as part of 1.0 https://github.com/rust-lang/libc/issues/3248. The reason for this is that for anything with padding or unions, it is not easy to ensure these are sound. An example is at https://github.com/rust-lang/libc/issues/2816.

I think the problem with this isn't always that they are unsound, but that their usage leads to unsoundness. For example, https://github.com/rust-lang/libc/issues/2700: they definitely shouldn't have relied on `sigemptyset` to fully initialize the struct, but there isn't anything that looks incorrect at first glance.

At this time, a very rough search shows about 850 `Cargo.toml` files that make use of this feature, compared to ~50k without it (https://github.com/search?q=lang%3Atoml+%2F%5Elibc%5Cs%2B%3D.*%22extra_traits%22%2F&type=code vs https://github.com/search?q=lang%3Atoml+%2F%5Elibc%5Cs%2B%3D%2F&type=code). This suggests there isn't a whole lot of ecosystem use that will experience breakage, but it also isn't negligible.

It seems Nix may be influenced by this more directly https://github.com/nix-rust/nix/issues/2468.

Personally I think we could leave a `Debug` implementation but just make non-primitive types nonexhaustive (e.g. `sigset_t { .. }`). This would at least allow `#[derive(Debug)]` on anything that wraps these types.

There is a proposed PR up: https://github.com/rust-lang/libc/pull/3692

tgross35:
Trying to move discussion here from https://github.com/nix-rust/nix/issues/2468. @asomers said:

> Extra unused fields might be unhelpful when implementing certain traits, but how could they make it unsafe? And shouldn't the compiler be able to handle unaligned fields?

For padding fields or reserved fields, reading the data might not be a problem - but we have no clue what the kernel might put there so a derived `PartialEq` is totally meaningless. Even without padding fields, having the derived implementation makes it easy to assume that the kernel/c library will set all fields to meaningful values which certainly isn't always true.

It is just difficult to justify keeping around with these kind of footguns. Plus having a pointer in the struct makes the equality check meaningless, or at the very least unintuitive. Making sure that everyone only compares the fields they need and know are initialized seems like a pretty reasonable tradeoff.

However, it would be good to see some code examples to better understand how exactly this is currently being used.

[...]

PR #4239, opened 2025-01-12
adding ethhdr type for linux/android for proper packet filtering.

devnexen:


src/unix/linux_like/android/mod.rs
```diff
@@ -517,6 +517,11 @@ s! {
         pub ifr6_prefixlen: u32,
         pub ifr6_ifindex: c_int,
     }
+
+    // is __be16 __bitwise __u16
+    pub struct __c_anonymous___be16 {
```
maurer, line 522: Why `__c_anonymous_`? This isn't an anonymous type, it's name is `__be16`.

Unless I'm mistaken, `__c_anonymous_fieldname` is normally used for inline structs/unions attached to a field of that name - this is a named type, and that's not the field name.

src/unix/linux_like/android/mod.rs
```diff
@@ -1020,6 +1034,31 @@ cfg_if! {
                     .finish()
             }
         }
+
+        impl Eq for ethhdr {}
+
+        impl PartialEq for ethhdr {
```
maurer, line 1040: This `PartialEq` definition appears to lack a check on the `h_proto` field.

src/unix/linux_like/android/mod.rs
```diff
@@ -1020,6 +1034,31 @@ cfg_if! {
...
+            fn eq(&self, other: &ethhdr) -> bool {
+                self.h_dest
+                    .iter()
+                    .zip(other.h_dest.iter())
+                    .all(|(a, b)| a == b)
```
maurer, line 1045: Do you not want them to compare equal if they have the same CStr value, but different null trailing bytes?

src/unix/linux_like/android/mod.rs
maurer, line 522: Consider deriving Eq/PartialEq, adding a method to access/set it as host `u16`, and adding a `Debug` based on the host representation?

If you do that, your later stuff on `ethhdr` can just be a derive.

src/unix/linux_like/android/mod.rs
```diff
@@ -638,6 +643,15 @@ s_no_extra_traits! {
         pub ifc_len: c_int,
         pub ifc_ifcu: __c_anonymous_ifc_ifcu,
     }
+
+    // linux/if_ether.h
+
+    #[repr(C, align(1))]
```
maurer, line 649: This struct is [packed](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/kernel/uapi/linux/if_ether.h;l=128;bpv=1;bpt=1) and not align-1.

src/unix/linux_like/android/mod.rs
```diff
@@ -1020,6 +1034,31 @@ cfg_if! {
...
+                        .all(|(a, b)| a == b)
+            }
+        }
+
+        impl fmt::Debug for ethhdr {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                f.debug_struct("ethhdr")
+                    .field("h_dest", &self.h_dest)
+                    .field("h_source", &self.h_source)
+                    .finish()
```
maurer, line 1059: Missing proto field

src/unix/linux_like/linux/mod.rs
```diff
@@ -1319,6 +1319,11 @@ s! {
         pub propagation: crate::__u64,
         pub userns_fd: crate::__u64,
     }
+
+    // is __be16 __bitwise __u16
+    pub struct __c_anonymous___be16 {
```
maurer, line 1324: As above re: __c_anonymous - it also might be helpful just to have a type like this higher up in the hierarchy to be shared, I find the idea that only Linux/Android have bigendian u16s unlikely.

maurer (commented):
The main thing I see missing here is a use of `ethhdr`. A struct type with no usage by libc might not actually belong here, even though it is on the UAPI interface - this is a bindings crate for `libc`, not for the Linux kernel.

Is there a function in `libc` that you intend to call that uses the `ethhdr` struct? If so, layering on a commit binding that on top of this one would make a better PR IMO.

src/unix/linux_like/android/mod.rs
devnexen: I know. was trying to solve few architectures build failures.

devnexen:
> The main thing I see missing here is a use of `ethhdr`. A struct type with no usage by libc might not actually belong here, even though it is on the UAPI interface - this is a bindings crate for `libc`, not for the Linux kernel.
> 

There is, for packet filters purpose. sendto/recvfrom can use any buffer e.g. ethhdr for this case. since AF_PACKET family and ETH_P* constants are already there, I m just trying to finish it off.

src/unix/linux_like/android/mod.rs
devnexen: I did not forget but until I solve the alignment issue in some platforms and as I was playing with the h_proto type I do not bother with it just yet (still a draft).

src/unix/linux_like/android/mod.rs
devnexen: note that those fields represent mac addresses not as strings (note they re unsigned char).

maurer: Ah, my bad, you're correct.

tgross35:
Since this is marked as a draft:

@rustbot author

(just update the labels if that isn't accurate)

tgross35:
This will need a rebase but I think it's ready for review so updating the labels

@rustbot review

libc-test/build.rs
```diff
@@ -3918,6 +3920,10 @@ fn test_linux(target: &str) {
         if sparc64 && (ty == "Elf32_Rela" || ty == "Elf64_Rela") {
             return true;
         }
+        // FIXME: alignment issue with this arch
+        if loongarch64 && ty == "ethhdr" {
+            return true;
+        }
```
tgross35: Any idea why? There doesn't seem to be anything loongarch-specific at https://github.com/torvalds/linux/blob/e53642b87a4f4b03a8d7e5f8507fc3cd0c595ea6/include/uapi/linux/if_ether.h#L172-L178. If it was just a bug then maybe it's worth removing to retry.

libc-test/build.rs
```diff
@@ -2301,6 +2301,8 @@ fn test_android(target: &str) {
         (struct_ == "sigaction" && field == "sa_sigaction") ||
         // signalfd had SIGSYS fields added in Android 4.19, but CI does not have that version yet.
         (struct_ == "signalfd_siginfo" && field == "ssi_call_addr") ||
+        // FIXME: `h_proto` is of type __be16 big endian version of __u16
+        (struct_ ==  "ethhdr" && field == "h_proto") ||
```
tgross35: What doesn't it like here? As far as I can tell, the `__be16` type just gets the `bitwise` attribute which isn't picked up by GCC https://github.com/torvalds/linux/blob/e53642b87a4f4b03a8d7e5f8507fc3cd0c595ea6/include/uapi/linux/types.h#L27-L41.

src/unix/linux_like/android/mod.rs
```diff
@@ -1025,6 +1035,33 @@ cfg_if! {
...
+        impl fmt::Debug for ethhdr {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                f.debug_struct("ethhdr")
+                    .field("h_dest", &self.h_dest)
+                    .field("h_source", &self.h_source)
+                    .field("h_proto", &{ self.h_proto })
+                    .finish()
+            }
+        }
     }
```
tgross35: `Debug` isn't needed anymore, and the other two can be omitted for now. If we bump MSRV to 1.67 so we can derive them then they can be added, but `extra_traits` is likely on the way out anyway.

tgross35: I take that back, doesn't the `derive(PartialEq)` work for these too?

devnexen: the simpler, the better, I was unaware of extra_traits obsolescence and the rest of it.

libc-test/build.rs
devnexen: yes, going to ..

src/unix/linux_like/android/mod.rs
tgross35: I think just having them in `s!` may work here, at least it's worth a try if you haven't yet. Otherwise, yeah - what exactly to do with `extra_traits` still isn't certain yet, but we do know for sure that we don't want to add any more handwritten impls.

There is a bit more at https://github.com/rust-lang/libc/issues/3880

src/unix/linux_like/android/mod.rs
```diff
@@ -857,15 +867,6 @@ cfg_if! {
...
         }
-
-        impl PartialEq for prop_info {
-            fn eq(&self, other: &prop_info) -> bool {
-                self.__name == other.__name
-                    && self.__serial == other.__serial
-                    && self.__value == other.__value
-            }
-        }
-        impl Eq for prop_info {}
```
tgross35: I think this was removed accidentally

libc-test/build.rs
tgross35: @devnexen what are the errors with this one?

devnexen: i forgot since, gonna try removing it

devnexen: I recall just now it was for some archs only.

tgross35: Could you paste the actual error here? I still don't understand how this could be failing.

devnexen: It happened only on CI but seems it does not happen anymore.

maurer (approved):

tgross35:
Sorry, guess I didn't see this was ready. Needs a rebase now

@rustbot author

PR #4242, opened 2025-01-14
Fix missing SO_* values 

snogge:
Some identifiers for #4148 was missing, and this was not detected because `verify-build` does not rebuild when 
RUST_LIBC_UNSTABLE_LINUX_TIME_BITS64=1 is set.

Fix the `verify-build` task by using a separate `target-dir` and add the missing identifiers

This is probably not the best way of structuring the CI steps, it could also be set up using a matrix in the `verify-build` task of `ci.yml`.

# Sources

https://github.com/torvalds/linux/blob/master/arch/mips/include/uapi/asm/socket.h#L125
https://github.com/torvalds/linux/blob/master/arch/powerpc/include/uapi/asm/socket.h
https://github.com/torvalds/linux/blob/master/tools/include/uapi/asm-generic/socket.h

# Checklist


ci/verify-build.sh
```diff
@@ -73,7 +72,7 @@ test_target() {
 
     if [ "$os" = "linux" ]; then
         # Test with the equivalent of __USE_TIME_BITS64
-        RUST_LIBC_UNSTABLE_LINUX_TIME_BITS64=1 $cmd
+        RUST_LIBC_UNSTABLE_LINUX_TIME_BITS64=1 $cmd --target-dir "target/$target-time64"
```
tgross35, line 75: Instead of the changes to this file, could you add `println!("cargo:rerun-if-env-changed=RUST_LIBC_UNSTABLE_LINUX_TIME_BITS64")` to build.rs?

tgross35: Do you just mean the right place within `build.rs`? If so, I would put it immediately before/after the env is read at `let linux_time_bits64 = env::var("RUST_LIBC_UNSTABLE_LINUX_TIME_BITS64").is_ok();` so it's easy to see they go together.

LGTM otherwise, note that CI will fail until a rust-lang/rust fix gets into nightly.

snogge: That was my question.  I moved things around a bit to collect the linux_time_bits64 stuff in one place as much as possible.

tgross35 (approved):
Thanks for following up, this looks good to merge as soon as CI gets fixed.

PR #4243, opened 2025-01-14
emscripten: Assume version is at least 3.1.42

kleisauke:
This revises commit 63b0d673eaf2a177ff208c5c50999738bdd1bf0d to assume that Emscripten 3.1.42 or later is being used whenever `emcc` is not available. Since Emscripten 3.1.42 was released on June 23, 2023, the majority of users are expected to have upgraded to a more recent version.

Resolves: https://github.com/rust-lang/rust/issues/131467.

### DIFF
--- a/ci/emscripten.sh
+++ b/ci/emscripten.sh
@@ -3,7 +3,7 @@
 set -eux
 
 # Note: keep in sync with:
-# https://github.com/rust-lang/rust/blob/master/src/ci/docker/scripts/emscripten.sh
+# https://github.com/rust-lang/rust/blob/master/src/doc/rustc/src/platform-support/wasm32-unknown-emscripten.md#requirements
 emsdk_version=3.1.68
 
 git clone https://github.com/emscripten-core/emsdk.git /emsdk-portable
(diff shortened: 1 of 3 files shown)


ci/emscripten.sh
```diff
@@ -2,8 +2,6 @@
 
 set -eux
 
-# Note: keep in sync with:
-# https://github.com/rust-lang/rust/blob/master/src/ci/docker/scripts/emscripten.sh
```
tgross35, line 6: Is there something else we should be referencing for version to test with? Maybe the platform docs.

tgross35:
The changes look fine to me, I think we might even be fine emitting a deprecation warning for old versions in `build.rs` so we can remove support. Keeping the old version around doesn't cost us anything though, so there is no urgency for that.

@hoodmane could you take a look at this as well?

ci/emscripten.sh
kleisauke: I've updated this link to https://github.com/rust-lang/rust/blob/master/src/doc/rustc/src/platform-support/wasm32-unknown-emscripten.md#requirements instead.

tgross35 (approved):
LGTM but I'll wait for hoodmane to ack. CI is broken with the current nightly, I'll rerun once that is fixed.

hoodmane (approved):
Thanks @kleisauke

PR #4244, opened 2025-01-16
feat: Wasm Linux support `wasm32-wali`

arjunr2:
# Description

Retargetting support to `main` (from #3778) for `wasm32-wali` according to MCP  rust-lang/compiler-team#797 

# Sources

WALI documentation: https://github.com/arjunr2/WALI

# Checklist


src/unix/linux_like/linux/musl/b64/wasm32/mod.rs
tgross35, line 1: Same question as in https://github.com/rust-lang/libc/pull/3778, why is there a `b64` module when the arch is `wasm32`?

src/unix/linux_like/linux/musl/b64/wasm32/wali.rs
tgross35, line 1: As mentioned in https://github.com/rust-lang/libc/pull/3778, please drop the syscall shims

tgross35 (changes requested):

src/lib.rs
```diff
@@ -1,6 +1,8 @@
 //! libc - Raw FFI bindings to platforms' system libraries
 #![crate_name = "libc"]
 #![crate_type = "rlib"]
+#![feature(c_variadic)]
+#![feature(concat_idents)]
```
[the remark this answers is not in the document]
arjunr2: This will not be necessary if we remove the syscall shims.

src/unix/linux_like/linux/musl/b64/wasm32/wali.rs
arjunr2: The syscall shims are rather integral to portability of many applications. What is the best direction forward to get that into the ecosystem?

libc-test/build.rs
```diff
@@ -4527,12 +4528,1011 @@ fn test_linux(target: &str) {
     test_linux_like_apis(target);
 }
 
+// Wasm32 Linux tests
+// Skips the following subcomponents:
+// * Linux headers
+// * Architecture-specific features
+// * Dynamic linking (until Wasm gains support)
+// Currently only supports the 'musl' environment
+fn test_wasm32_linux(target: &str) {
```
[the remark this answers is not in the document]
tgross35: I'm okay to add testing and think it's probably a good idea if more API is planned, but that is up to you as the target maintainer. Reasonably small changes in `test_linux` are just preferable over having the same config in two places.

src/unix/linux_like/linux/musl/b64/wasm32/wali.rs
tgross35: The part I am missing is, why doesn't wali just add a varidic `syscall(long, ...)` symbol that we could bind to? If portability is important then this seems relevant beyond Rust's libc, and it would be better to just not have the number->signature mapping responsibility in this repo.

If that doesn't work then I can help rework the macro to not rely on unstable features, and we could probably start by mapping a subset like e.g. whatever std will need.

In any case I think it's easiest to just drop this portion for now, get the rest merged, and then revisit.

src/unix/linux_like/linux/musl/b64/wasm32/mod.rs
tgross35: Also, please make sure the unneeded comments are cleaned up once this is ready.

src/unix/linux_like/linux/musl/b64/wasm32/wali.rs
arjunr2: The `wali-libc` does provide musl's variadic `syscall` that can be bound too. The problem I encountered though is that most invocations of `syscall` used by many libraries don't adhere to the exact type signature for some arguments (using i32 instead of i64 for instance). This would not be type-checked by Rust

This causes a unique issue for variadic argument parsing in C. It has not been an issue for hardware ISAs in musl till date since they just pass arguments to registers. However, for Wasm as a virtual ISA target, variadic calls (as per clang compiler today) are supported through arguments packed into a memory buffer -- passing incorrect types will lead to wrong offset parsing.

It is currently unclear what it handle this problem in the compiler, or if it even needs to be done. The easiest reliable fix is to type-cast all arguments accordingly in Rust, which is what the shim provides.

arjunr2: That being said, I'm on board for dropping the shim for the first merge, and we can start by increasing support as you suggested.

tgross35: Interesting; do varidic functions aside from `syscall` work cross-language, or is that one special? Rust's `...` should work with the exact same ABI as C's `...` and if not, it is a bug. Shimming to work around an ABI mismatch isn't a great solution.

You should open an issue in rust-lang/rust to investigate further, but if Rust's varidics aren't compatible with C's varidics then that just needs to be fixed.

arjunr2: This is not a Rust variadics issue since this also happens when invoking syscall from C. It's an LLVM compiler issue in marshalling variadic arguments and should be fixable.

src/unix/linux_like/linux/musl/b64/wasm32/mod.rs
arjunr2: wasm32 linux uses all the definitions in `b64`. I will have to copy all of that into `b32/mod` with  a `cfg_if` then. Is that preferable?

tgross35: Interesting, it's fine to leave as-is but could you add a note to the doc comment?

tgross35 (approved):
I haven't reviewed the bindings in depth since you are the maintainer, but changes LGTM. Could you add a doc comment in `b64/wasm32/mod.rs` saying why that is in the b64 module rather than b32 for wasm32? Since that's a bit confusing at first glance.

Also please squash, GH doesn't let me do that with merge queues.

arjunr2:
Squashed and added the doc comment. Merging the changes in

arjunr2:
@tgross35 Are the current failures an issue with the CI system post recent merges? I just rebased my changes (previously passing) over commits since March 1

tgross35:
I have no clue what is going on with android but it's pretty flaky (https://github.com/rust-lang/libc/issues/4297). Restarted the job, ping me if it fails again.

Issue #4249, opened 2025-01-23
[AIX] Use sa_sigaction instead of sa_union.__su_sigaction


xingxue-ibm:
The current `libc` crate implementation for AIX defines the `struct sigaction` as containing a union `sa_union` with two members: `__su_handler` and `__su_sigaction`. This mirrors the `struct sigaction` definition in the AIX system header. Consequently, any reference to `sa_sigaction` in Rust code, whether in shipped crates or user code, must be replaced with `sa_union.__su_sigaction` for AIX. Additionally, the types of these two union members are declared as function pointers rather than `sighandler_t` (i.e., `usize`), as is the case on other platforms. This discrepancy causes compiler errors when `sighandler_t` is used as the type for signal handlers.

The POSIX standard defines `struct sigaction` as containing `sa_handler` and `sa_sigaction` members, with their storage allowed to overlap. Other operating systems, such as Linux, also define a union for `sa_handler` and `sa_sigaction`. However, the `libc` crate implementations on these platforms define `sa_sigaction` directly as a member of `struct sigaction` rather than as part of a union. This is because the storage of `sa_handler` and `sa_sigaction` overlaps, and only `sa_sigaction` is typically referenced in Rust.

We propose modifying the `libc` crate implementation for AIX to define `sa_sigaction` as a direct member of `struct sigaction`, aligning it with implementations for other platforms. We will also update affected crates and test cases to reflect this change.
 
target=powerpc64-ibm-aix

xingxue-ibm:
Created PR [[AIX] Use sa_sigaction instead the union](https://github.com/rust-lang/libc/pull/4250).

xingxue-ibm:
- [x] PR for the `libc` crate [[AIX] Use sa_sigaction instead of the union #4250](https://github.com/rust-lang/libc/pull/4250)
- [ ] PR for the `nix` crate [Use sa_sigaction instead of sa_union.__su_sigaction for AIX](https://github.com/nix-rust/nix/pull/2614)
- [ ] PR for the `jobserver` crate [Bump libc to 0.2.171 to fix sigaction](https://github.com/rust-lang/jobserver-rs/pull/109)

tgross35:
I'll close this since AIUI https://github.com/rust-lang/libc/pull/4250 resolved the issues in `libc`, feel free to reopen if that isn't accurate.

PR #4250, opened 2025-01-23
[AIX] Use sa_sigaction instead of the union

xingxue-ibm:
# Description
The current `libc` crate implementation for AIX defines `struct sigaction` as containing a union `sa_union` with two members: `__su_handler` and `__su_sigaction`. This mirrors the `struct sigaction` definition in the AIX system header. Consequently, any reference to `sa_sigaction` in Rust code, whether in shipped crates or user code, must be replaced with `sa_union.__su_sigaction` for AIX. Additionally, the types of these two union members are declared as function pointers rather than `sighandler_t` (i.e., `usize`), as is the case on other platforms. This discrepancy causes compiler errors when `sighandler_t` is used as the type for signal handlers. Other operating systems, such as Linux, also define a union for `sa_handler` and `sa_sigaction` in `struct sigaction`. However, the `libc` crate implementations on these platforms define `sa_sigaction` directly as a member of `struct sigaction` rather than as part of a union. This patch modifies the `libc` crate implementation for AIX to define `sa_sigaction` as a direct member of `struct sigaction`, aligning it with implementations for other similar platforms. We will also update affected crates and test cases to reflect this change.

# Sources

# Checklist

### DIFF
--- a/src/unix/aix/mod.rs
+++ b/src/unix/aix/mod.rs
@@ -533,20 +533,15 @@ s! {
         pub it_interval: crate::timespec,
         pub it_value: crate::timespec,
     }
-}
-
-s_no_extra_traits! {
-    pub union __sigaction_sa_union {
-        pub __su_handler: extern "C" fn(c: c_int),
-        pub __su_sigaction: extern "C" fn(c: c_int, info: *mut siginfo_t, ptr: *mut c_void),
-    }
 
     pub struct sigaction {
-        pub sa_union: __sigaction_sa_union,
+        pub sa_sigaction: crate::sighandler_t, // FIXME(union): this field is actually a union
         pub sa_mask: sigset_t,
         pub sa_flags: c_int,
     }
+}
 
+s_no_extra_traits! {
     pub union __poll_ctl_ext_u {
         pub addr: *mut c_void,
         pub data32: u32,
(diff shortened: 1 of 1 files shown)


xingxue-ibm:
close [GH-4249](https://github.com/rust-lang/libc/issues/4249)

daltenty (approved):
LGTM from the AIX perspective. This makes the implementation common with what is done on other targets

src/unix/aix/mod.rs
```diff
@@ -533,20 +533,15 @@ s! {
...
-
-s_no_extra_traits! {
-    pub union __sigaction_sa_union {
-        pub __su_handler: extern "C" fn(c: c_int),
-        pub __su_sigaction: extern "C" fn(c: c_int, info: *mut siginfo_t, ptr: *mut c_void),
-    }
 
     pub struct sigaction {
-        pub sa_union: __sigaction_sa_union,
+        pub sa_sigaction: crate::sighandler_t, //actually a union with sa_handle
```
tgross35, line 538: Is this something that should be changed once we are able to use unions (1.0)? If so, could you label the comment `FIXME(union)`?

xingxue-ibm: Changed as suggested, thanks so much!

tgross35 (approved):
Needs a rebase but the change LGTM. @xingxue-ibm @daltenty do you want this in a stable release since it is a breaking change?

xingxue-ibm:
> Needs a rebase but the change LGTM. @xingxue-ibm @daltenty do you want this in a stable release since it is a breaking change?

Done rebasing. Yeah, it would be great if this change could be included in a stable release. What do you think, @daltenty?

tgross35:
I'll wait for @daltenty to ack before backporting. Just add the label back (`@rustbot label +stable-nominated`) so I pick it up if you do want it in stable.

daltenty:
Thanks! Yep, it'd help a lot of crates to have this in stable as well.

PR #4252, opened 2025-01-24
freebsd add basic ethernet frames support

devnexen:
[reference](https://github.com/freebsd/freebsd-src/blob/stable/14/sys/net/ethernet.h)


src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -2541,6 +2553,53 @@ cfg_if! {
                     .finish()
             }
         }
+
```
tgross35, line 2556: Could you add a

```rust
// FIXME(msrv): `derive` on packed structs cannot be used below 1.67
```

Just so we clean it up later?

tgross35 (approved):
One nit but this looks good to me. @asomers could you take a look?

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -2541,6 +2553,55 @@ cfg_if! {
...
+        // FIXME(msrv): `derive` on packed structs cannot be used below 1.67
```
asomers, line 2557: That sounds to me like a compelling reason to raise the main branch's MSRV.  Why is it currently set to 1.63?

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -4957,6 +5028,24 @@ f! {
     pub fn PROT_MAX_EXTRACT(x: c_int) -> c_int {
         (x >> 16) & (crate::PROT_READ | crate::PROT_WRITE | crate::PROT_EXEC)
     }
+
+    pub {const} fn ETHER_IS_MULTICAST(addr: *mut u_char) -> bool {
+        (*addr.wrapping_add(0)) & 0x01 != 0x00
```
asomers, line 5033: I'm surprised that Clippy doesn't complain about the uselessness of `wrapping_add(0)`.

src/unix/bsd/freebsdlike/freebsd/mod.rs
tgross35: 1.63 is what Debian stable currently ships so it is a pretty common baseline in the ecosystem. In a few months Trixie will be released 1.83 (or higher) so we will probably be able to do another bump (not that high, but at least above 1.63).

tgross35: Stumbled across this again because considering a MSRV bump: what exactly happened at 1.67? `Derive` on packed structs seems to work all the way down to 1.0 https://rust.godbolt.org/z/qqPsT46zr.

`Derive` on packed structs without `Copy` is a newer feature as of 1.69, but everything we have is `Copy` anyway

PR #4258, opened 2025-02-05
fix: Revert vita's c_char back to i8

pheki:
# Description

Hi,

https://github.com/rust-lang/libc/pull/4199 changed the definition of `c_char` from i8 to u8 for most ARM targets. While that would usually be correct, [VITASDK uses signed chars by default](https://github.com/vitasdk/buildscripts/blob/master/patches/gcc/0001-gcc-10.patch#L33-L34).

This will surely conflict with #4256, but I can rebase it after that one is merged. I opened that one first because in the off-chance it was just accepted, this fix would only be needed on rustc. I'll also create a rustc PR.

# Sources

https://github.com/vitasdk/buildscripts/blob/master/patches/gcc/0001-gcc-10.patch#L33-L34

# Checklist

- [X] Relevant tests in `libc-test/semver` have been updated (API didn't change)

### DIFF
--- a/src/primitives.rs
+++ b/src/primitives.rs
@@ -21,6 +21,7 @@ cfg_if! {
     if #[cfg(all(
         not(windows),
         not(target_vendor = "apple"),
+        not(target_os = "vita"),
         any(
             target_arch = "aarch64",
             target_arch = "arm",


tgross35:
Thanks for the PR, but could you do the rustc change first? That's the better place to discuss this change and we can ping a few maintainers, `libc` will just follow whatever the decision is there.

tgross35:
Also, does Clang do the right thing? This might need to be an upstream issue as well.

pheki:
> Thanks for the PR, but could you do the rustc change first? That's the better place to discuss this change and we can ping a few maintainers, libc will just follow whatever the decision is there.

Sure can, will do that.

> Also, does Clang do the right thing? This might need to be an upstream issue as well.

It [does not](https://github.com/llvm/llvm-project/blob/08bda1cc6b0d2f1d31a89a76b5c154a11086c420/clang/lib/Driver/ToolChains/Clang.cpp#L1347), but I'm not too worried as for now Clang is not really supported by the vita programming community / `VITASDK`. The pre-compiled libraries are all compiled using vita's `gcc` and [we set `TARGET_CC` and `TARGET_CXX`](https://github.com/vita-rust/cargo-vita/blob/d564a132cbd43947118c0d6d0ebfbea7d1dd7fa7/src/commands/build.rs#L230) in `cargo vita` for build scripts using `cc`. Regardless I'll take a look into fixing Clang once the Rust side is fixed.

cuviper:
> Thanks for the PR, but could you do the rustc change first? That's the better place to discuss this change and we can ping a few maintainers, `libc` will just follow whatever the decision is there.

We're in agreement in rust-lang/rust#136667, so I think you can proceed here.

src/lib.rs
```diff
@@ -50,6 +50,7 @@ cfg_if! {
             target_os = "watchos",
             target_os = "visionos",
         )),
+        not(target_os = "vita"),
```
tgross35, line 53: Nit, this can be in the existing group

```suggestion
            target_os = "watchos",
            target_os = "vita",
        )),
```

pheki: The reason I put it separately was due to an organization purpose: in my understanding, that group has all of the apple OSes, which is what the FIXME above applies to. Otherwise, `windows` would also be in that same group.

1. Do you still want me to make this change? 
2. Should I also do the same for windows? e.g. 
```
not(any(
    windows,
    // FIXME(ctest): just use `target_vendor` = "apple"` once `ctest` supports it
    target_os = "macos",
    target_os = "ios",
    target_os = "tvos",
    target_os = "watchos",
    target_os = "visionos",
    target_os = "vita",
)),
```

tgross35: Nope you're right, good as-is. Thanks for the clarification!

tgross35 (approved):

tgross35:
As you mentioned above, this just needs a rebase

tgross35 (approved):

pheki:
Is it possible to retry the job in some way? The failure seems unrelated, the test for aarch64-linux-android hung

@rustbot ready

tgross35:
Yeah, that isn't this PR's fault, restarted the job.

PR #4262, opened 2025-02-13
Update Redox's signal constants

daxtens:
Redox changed its signal ABI last year:
https://gitlab.redox-os.org/redox-os/relibc/-/commit/63509e75ce127fdb05c07a464e975687e8ef2977

The old definitions (unsurprisingly) cause weird crashes. Use the new ones from Redox's relibc.

# Description
Redox changed its signal ABI last year. The old definitions (unsurprisingly) cause weird crashes. Use the new ones from Redox's relibc.

# Sources

https://gitlab.redox-os.org/redox-os/relibc/-/commit/63509e75ce127fdb05c07a464e975687e8ef2977

# Checklist

I'm still working out how to run tests inside of redox, so I'm leaving this as a draft until I figure that out.

- [X] Relevant tests in `libc-test/semver` have been updated (No update required)

### DIFF
--- a/src/unix/redox/mod.rs
+++ b/src/unix/redox/mod.rs
@@ -645,14 +645,14 @@ pub const SIGPWR: c_int = 30;
 pub const SIGSYS: c_int = 31;
 pub const NSIG: c_int = 32;
 
-pub const SA_NOCLDSTOP: c_ulong = 0x00000001;
-pub const SA_NOCLDWAIT: c_ulong = 0x00000002;
-pub const SA_SIGINFO: c_ulong = 0x00000004;
-pub const SA_RESTORER: c_ulong = 0x04000000;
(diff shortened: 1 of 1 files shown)


tgross35:
@rustbot author since this is a draft (comment `@rustbot review` when it's ready for a look)

Were you able to figure the tests out? `cargo t` does the basics, following or directly running the ci scripts should get everything else.

daxtens:
I am working on actually getting the libc tests compiled for redox and installed into a redox image - once I do I will update the PR and change the label, but it might take a while as this is very much a spare-time thing.

tgross35:
Fyi it looks like https://github.com/rust-lang/libc/pull/4426 is making this same change.

daxtens:
Fantastic, let’s go with that change then. They’ve made much better progress with testing too

Issue #4261, opened 2025-02-13
`README.md` is outdated, references `ci/build.sh` which doesn't exist anymore


cipherself:
https://github.com/rust-lang/libc/blob/2258bf0fb96767bcffbe3ed09b29a31ee54b549b/README.md?plain=1#L68

tgross35:
Ah yeah, it should be ci/run.sh. Feel free to submit a PR if you're up for it.

cipherself:
> Ah yeah, it should be ci/verify-build.sh now. Feel free to submit a PR if you're up for it.

Sure thing, here we go https://github.com/rust-lang/libc/pull/4263

PR #4263, opened 2025-02-13
Fix reference to build file with guaranteed build platforms.

cipherself:
# Description

refs https://github.com/rust-lang/libc/issues/4261

This PR fixes a reference in `README.md` that referenced `ci/build.sh` which no longer exists, instead it now references `ci/verify-build.sh`

# Sources

# Checklist


tgross35 (approved):
Thanks!

PR #4265, opened 2025-02-14
NetBSD: fix getmntinfo for NetBSD

0-wiz-0:
# Description

This patch fixes libc::getmntinfo on NetBSD.

# Sources

The problem is that NetBSD provides binary compatibility in a way that was not considered when initially providing the implementation.
Once a symbol name has been used, it stays the same, so the original `getmntinfo()` in NetBSD, using `struct statfs`, will always be available as `getmntinfo`. When `getmntinfo` was changed to provide `struct statvfs` (note the 'v') a long time ago, a new symbol was introduced, `__getmntinfo13` and the system headers rename access to `getmntinfo` to call `__getmntinfo13` instead.
The rust implementation called the old symbol (providing `struct statfs`) but parsed the returned information as `struct statvfs`, leading to wrong data and segfaults.

# Checklist

The change has been tested on NetBSD 9, 10, and -current (all supported versions).
No compatibility for NetBSD 8 or older is required, since these versions are not supported any longer; the symbol rename happened before NetBSD 9.

### DIFF
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -2849,7 +2849,7 @@ extern "C" {
         ntargets: size_t,
         hint: *const c_void,
     ) -> c_int;
-
+    #[link_name = "__getmntinfo13"]
     pub fn getmntinfo(mntbufp: *mut *mut crate::statvfs, flags: c_int) -> c_int;
     pub fn getvfsstat(buf: *mut statvfs, bufsize: size_t, flags: c_int) -> c_int;
 }


0-wiz-0:
This should be pulled up, so

tgross35 (approved):
Makes sense to me. Our [`statvfs`](https://github.com/rust-lang/libc/blob/41f1fdba0ea7a46b4f4899abaa0a20deaabdf1ab/src/unix/bsd/netbsdlike/netbsd/mod.rs#L853) is actually [`statvfs90`](https://github.com/NetBSD/src/blob/ea69897761f4bc39da7326529c06cc55f32a6d1c/sys/compat/sys/statvfs.h#L37) then? (Based on the call here https://github.com/NetBSD/src/blob/ea69897761f4bc39da7326529c06cc55f32a6d1c/lib/libc/compat/gen/compat___getmntinfo13.c#L52-L58)

0-wiz-0:
Exactly. We could switch to the new statvfs with the one additional field, but it's a bigger change and we'd need to differentiate between NetBSD 9 and 10+, so i'd leave that for another time.

tgross35:
You should open an issue for changing the struct, that is probably something we could do for 1.0.

0-wiz-0:
> You should open an issue for changing the struct, that is probably something we could do before 1.0.

I've filed for this, which works on NetBSD 10, but I need some guidance on how to best support NetBSD 9. Thanks!

PR #1408, opened 2019-06-26
Support Vxworks

BaoshanPang:
Define the C interface to VxWorks

PR #4268, opened 2025-02-15
Fix size of time_t in vxworks

hax0kartik:
# Description

This fixes the size of time_t in VxWorks which was incorrectly set to c_long. This broke compatibility with 32-bit targets.

Requesting stable-nomination.

# Sources

# Checklist

- [X] Checked locally on VxWorks.

### DIFF
--- a/src/vxworks/mod.rs
+++ b/src/vxworks/mod.rs
@@ -34,7 +34,7 @@ pub type ino_t = c_ulong;
 
 pub type rlim_t = c_ulong;
 pub type suseconds_t = c_long;
-pub type time_t = c_long;
+pub type time_t = c_longlong;
 
 pub type errno_t = c_int;
 


tgross35:
Cc @jackorobot @biabbas would you be able to double check this?

I know there were some surprising differences on vxworks, from https://github.com/rust-lang/libc/pull/4043.

biabbas:
@tgross35 
Yes, this is correct. To avoid the problems like year 2038 problem, the default size of time_t on 32 bits VxWorks targets is 64 bits.

biabbas (approved):

tgross35 (approved):
Thanks for confirming. Looks like this has existed since https://github.com/rust-lang/libc/pull/1408 (cc @BaoshanPang)

PR #4269, opened 2025-02-15
adding SUN_LEN macro to linux glibc/musl

devnexen:
[glibc ref](https://github.com/bminor/glibc/blob/1b6f868625403d6b7683af840e87d2b18d5d7731/socket/sys/un.h#L41) [musl ref](https://github.com/bminor/musl/blob/c47ad25ea3b484e10326f933e927c0bc8cded3da/include/sys/un.h#L24)


tgross35:
I'm just going to mark this blocked since I don't yet know what to do without `offset_of!`. I don't think we want to add a new config specific for this macro, but maybe something like a `recent-msrv` cargo feature is possible.

@rustbot blocked

tgross35:
Added an `offset_of!` macro in 18083938f28c39908f487624bd96ed9a6c0c24e0, this should be able to go forward with that.

@rustbot author

src/unix/linux_like/linux/mod.rs
```diff
@@ -5877,6 +5877,17 @@ f! {
     pub fn ELF64_R_INFO(sym: Elf64_Xword, t: Elf64_Xword) -> Elf64_Xword {
         sym << (32 + t)
     }
+
+    #[cfg(target_env = "gnu")]
+    #[allow(unused_unsafe)]
+    pub fn SUN_LEN(s: crate::sockaddr_un) -> usize {
+        offset_of!(crate::sockaddr_un, sun_path) + crate::strlen(s.sun_path.as_ptr())
+    }
```
tgross35, line 5625: Could you add `#[allow(unused_unsafe)]` in the macro instead? There isn't any `unsafe` here.

src/macros.rs
```diff
@@ -401,7 +401,7 @@ macro_rules! offset_of {
         // SAFETY: computed address is inbounds since we have a stack alloc for T
         let fptr = unsafe { core::ptr::addr_of!((*ptr).$field) };
         let off = (fptr as usize).checked_sub(ptr as usize).unwrap();
-        assert!(off <= core::mem::size_of::<$Ty>());
+        core::assert!(off <= core::mem::size_of::<$Ty>());
```
tgross35, line 426: If you could split macro changes to a separate commit that would be ideal (distinct changes in separate commits makes the backporting easier)

tgross35: @devnexen the first commit still has some macro changes, and the second commit undoes part of the first. It should be split so that one commit does all the changes in `macros.rs`, then another commit does the rest.

(I'm happy to take care of this for you if you'd like, just don't like touching commit contents without asking)

devnexen: please proceed :)

tgross35 (approved):
Should be all set, thanks!

PR #4276, opened 2025-02-19
Split out the stat struct for gnu/b32/mips

snogge:
# Description

The stat struct is significantly different for GNU 32 bit mips.  Split it out to make future work on `_FILE_OFFSET_BITS=64` and `_TIME_BITS=64` easier. 

# Sources

# Checklist

There is no way to test mips builds at the moment.  This change does not add, change or remove any constants, structs or functions.


tgross35 (approved):
Seems reasonable to me, uapi defines these separately anyway.

PR #4279, opened 2025-02-22
Add cygwin support

Berrysoft:
# Description
https://github.com/rust-lang/rust/pull/134999

Most changes are ported from https://github.com/ookiineko-cygport/libc

# Sources
https://github.com/cygwin/cygwin/tree/main/winsup/cygwin/include

# Checklist

I'm not sure how to test a no_std target. Actually I need these changes to implement std for cygwin target.


Berrysoft:
Stable compilers failed because `cygwin` target is only supported by nightly ones nowadays. Are there some ways to solve this problem?

tgross35:
> Stable compilers failed because `cygwin` target is only supported by nightly ones nowadays. Are there some ways to solve this problem?

If it's just check-cfg failing, then it needs an update in https://github.com/rust-lang/libc/blob/3695157ec1ca33c1da8ed8214e09b9ed25acafca/build.rs#L25

src/unix/mod.rs
```diff
@@ -136,9 +136,18 @@ s! {
     }
 
     pub struct hostent {
```
tgross35, line 138: `hostent` and `linger` seem to have pretty different definitions between Cygwin and the rest of Unix. Would it be possible to move the existing structures into a `cfg(not(target_os = "cygwin"))` block, then redefine them in `cygwin/mod.rs` instead?

src/unix/cygwin/mod.rs
```diff
@@ -0,0 +1,3137 @@
+use crate::prelude::*;
+use crate::*;
+
+pub type c_char = i8;
+pub type c_long = i64;
+pub type c_ulong = u64;
```
tgross35, line 6: These shouldn't be needed, the types are now defined in `primitives.rs`

src/unix/cygwin/mod.rs
```diff
@@ -0,0 +1,3137 @@
...
+
+cfg_if! {
+    if #[cfg(doc)] {
+        pub(crate) type Ioctl = c_int;
+    } else {
+        #[doc(hidden)]
+        pub type Ioctl = c_int;
+    }
+}
```
tgross35, line 15: Why is this config needed?

src/unix/cygwin/mod.rs
```diff
@@ -0,0 +1,3137 @@
...
+pub type useconds_t = c_ulong;
+
+#[cfg_attr(feature = "extra_traits", derive(Debug))]
+pub enum timezone {}
+impl Copy for timezone {}
+impl Clone for timezone {
+    fn clone(&self) -> timezone {
+        *self
+    }
+}
```
tgross35, line 51: Can this go in an `e!` macro so it gets these derived? Same for `enum sem`.

src/unix/cygwin/mod.rs
```diff
@@ -0,0 +1,3137 @@
...
+    pub struct flock {
+        pub l_type: c_short,
+        pub l_whence: c_short,
+        pub l_start: off_t,
+        pub l_len: off_t,
+        pub l_pid: pid_t,
+    }
+}
+
+s_no_extra_traits! {
```
tgross35, line 280: Do all of the items here need to be in `s_no_extra_traits`? It looks like at least a few of them don't have an alignment requirement or contain a union so could probably go in `s!`.

src/unix/cygwin/mod.rs
```diff
@@ -0,0 +1,3137 @@
...
+            fn hash<H: hash::Hasher>(&self, state: &mut H) {
+                self.uc_mcontext.hash(state);
+                self.uc_link.hash(state);
+                self.uc_sigmask.hash(state);
+                self.uc_stack.hash(state);
+                self.uc_flags.hash(state);
+            }
+        }
+
+        impl PartialEq for __c_anonymous_ifr_ifru {
```
tgross35, line 952: We aren't doing any traits that read values for unions anymore, or anything that contains them, since the implementations are too often unsound (as is the case here, possible uninitialized data read). So please drop these, the unions should still get a default debug impl from https://github.com/rust-lang/libc/blob/268e1b3810ac07ed637d9005bc1a54e49218c958/src/macros.rs#L163-L168.

src/unix/cygwin/mod.rs
```diff
@@ -0,0 +1,3137 @@
...
+// intentionally not public, only used for fd_set
+cfg_if! {
+    if #[cfg(target_pointer_width = "32")] {
+        const ULONG_SIZE: usize = 32;
+    } else if #[cfg(target_pointer_width = "64")] {
+        const ULONG_SIZE: usize = 64;
+    } else {
+        // Unknown target_pointer_width
+    }
+}
```
tgross35, line 2486: Could these use `core::mem::size_of::<c_ulong>()` instead?

src/unix/cygwin/mod.rs
```diff
@@ -0,0 +1,3137 @@
...
+
+    pub fn FD_ZERO(set: *mut fd_set) {
+        for slot in (*set).fds_bits.iter_mut() {
+            *slot = 0;
+        }
+    }
+
+    pub fn CPU_ALLOC_SIZE(count: c_int) -> size_t {
+        let _dummy: cpu_set_t = core::mem::zeroed();
+        let size_in_bits = 8 * core::mem::size_of_val(&_dummy.bits[0]);
```
tgross35, line 2515: `core::mem::zeroed()` is advised against nowadays. Could this just be `cpu_set_t { bits: [0; 16] }`?

Applies a couple of other places.

src/unix/cygwin/mod.rs
```diff
@@ -0,0 +1,3137 @@
...
+    }
+
+    pub struct siginfo_t {
+        pub si_signo: c_int,
+        pub si_code: c_int,
+        pub si_pid: pid_t,
+        pub si_uid: uid_t,
+        pub si_errno: c_int,
+        __pad: [u32; 32],
+    }
```
tgross35, line 382: Do you have a link to the header this is defined in? It doesn't seem to match https://github.com/cygwin/cygwin/blob/4bcc6adec765ee8354bfc9e6c060c9d1c0c7bf46/newlib/libc/include/sys/signal.h#L69-L73.

tgross35 (commented):
Most of this looks pretty good but there are a handful of changes to get this more in line with the other OSs.

Does cygwin only ever use newlib, or are there other libc implementations?

src/unix/cygwin/mod.rs
Berrysoft: The other targets use this way of definition.

src/unix/cygwin/mod.rs
Berrysoft: The other targets use this kind of definition.

src/unix/cygwin/mod.rs
Berrysoft: The linux targets use `zeroed`.

src/unix/cygwin/mod.rs
Berrysoft: https://github.com/cygwin/cygwin/blob/main/winsup/cygwin/include/cygwin/signal.h#L158-L203

src/unix/cygwin/mod.rs
Berrysoft: Seems that there are many code in this repo still impl Eq for unions:)

src/unix/cygwin/mod.rs
Berrysoft: The macro `e!` also impls Eq, Hash, PartialEq. We don't want those traits.

mati865:
AFAIK Nnwlib is the only one ported.

Berrysoft:
Newlib is the only libc on cygwin, but some definitions are replaced, e.g., signal-related structures.

src/unix/cygwin/mod.rs
tgross35: That probably predates const `size_of`. We don't need to support versions that old anymore so this can use the new version, the others will get cleaned up at some point.

src/unix/cygwin/mod.rs
```diff
@@ -0,0 +1,2449 @@
...
+
+    pub {const} fn CMSG_SPACE(length: c_uint) -> c_uint {
+        (CMSG_ALIGN(length as usize) + CMSG_ALIGN(::core::mem::size_of::<cmsghdr>())) as c_uint
+    }
+
+    pub fn CMSG_FIRSTHDR(mhdr: *const msghdr) -> *mut cmsghdr {
+        if (*mhdr).msg_controllen as usize >= core::mem::size_of::<cmsghdr>() {
+            (*mhdr).msg_control as *mut cmsghdr
+        } else {
+            0 as *mut cmsghdr
```
tgross35, line 1899: Prefer `core::ptr::nul_mut()` rather than casting from 0 (applies a few places).

src/unix/cygwin/mod.rs
```diff
@@ -0,0 +1,2449 @@
...
+        CMSG_ALIGN(::core::mem::size_of::<cmsghdr>()) as c_uint + length
+    }
+
+    pub {const} fn CMSG_SPACE(length: c_uint) -> c_uint {
+        (CMSG_ALIGN(length as usize) + CMSG_ALIGN(::core::mem::size_of::<cmsghdr>())) as c_uint
+    }
+
+    pub fn CMSG_FIRSTHDR(mhdr: *const msghdr) -> *mut cmsghdr {
+        if (*mhdr).msg_controllen as usize >= core::mem::size_of::<cmsghdr>() {
+            (*mhdr).msg_control as *mut cmsghdr
```
tgross35, line 1897: Prefer `.cast::<cmsghdr>()` over `as` casting with `.cast_mut()` if needed in order to avoid silently changing mutability. Applies a few places.

tgross35 (changes requested):
I did a more thorough read through of everything here, I did not check the definitions for correctness but everything looks pretty good for implementation. I have a handful of small requests then this should be good to merge.

Please squash after these changes are done.

libc-test/build.rs
```diff
@@ -642,6 +644,173 @@ fn test_openbsd(target: &str) {
...
+            "FIOCLEX" | "SA_NOCLDWAIT" => true,
+
+            _ => false,
+        }
+    });
+
+    cfg.skip_signededness(move |c| match c {
+        n if n.starts_with("pthread") => true,
+
+        "sighandler_t" => true,
```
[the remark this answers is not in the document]
Berrysoft: Yes. It is defined as `size_t` on unix, while it's actually a function ptr on cygwin.

libc-test/semver/cygwin.txt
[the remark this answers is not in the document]
Berrysoft: I'll leave this to someone others then:)

libc-test/build.rs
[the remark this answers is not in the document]
tgross35: Mind adding a comment?

otherwise this all lgtm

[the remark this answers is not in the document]
Berrysoft: Comment added.

Berrysoft:
The windows CI failed because the self-update command of rustup failed. I have also met this problem in my other repos, and it's not caused by this PR.

tgross35 (approved):
Thanks for the updates, everything looks good. Looks like CI is going again.

Berrysoft:
Thank you! I don't understand why the arm64 android CI failed. Maybe a rerun would be fine?

tgross35:
Android is just flaky https://github.com/rust-lang/libc/issues/4297, restarted

Berrysoft:
I see. It reached the timeout (25 minutes) and was cancelled again.

tgross35:
> The windows CI failed because the self-update command of rustup failed. I have also met this problem in my other repos, and it's not caused by this PR.

Fwiw things still seem broken, passing `--no-self-update` corrected the issue.

tgross35:
This is in the latest release fyi https://github.com/rust-lang/libc/releases/tag/0.2.171

PR #4283, opened 2025-02-22
Merge the `ctest2` repository into `libc`

tgross35:
`libc`'s testing is pretty tightly coupled to [`ctest2`](https://crates.io/crates/ctest2), which is a reasonably simple crate that maps Rust syntax to C syntax and creates tests based off of that. `ctest2` itself is a fork of [`ctest`](https://crates.io/crates/ctest).

Unfortunately, relying on `ctest(2)` has been somewhat of a problem since it supports only a very outdated version of Rust's syntax, and there are a handful of other bugs. There have been some GSoC proposals to modernize `libc` and its test crates; moving `ctest` to a rust-lang repo makes that easier and should come with some maintenance benefits.

On Zulip, @alexcrichton confirmed ownership of the original `ctest` crate could be transferred to rust-lang so we can continue publishing under that name, for other repos that make use of `ctest`/`ctest2`.

This pull only brings over the repo source and history as it currently exists. I will need to follow up with integration of the CI and release workflows.


tgross35:
@JohnTitor would you be okay with this libc absorbing ctest2, or do you have any concerns? You can continue to maintain it here if you wish of course.

tgross35:
Also, does anyone know of a better way to merge unrelated histories? The blame works fine but the log appears empty.

alexcrichton:
I think I've done something [like this](https://stackoverflow.com/questions/16450423/git-rewrite-history-move-folders-and-files-up-to-parent-folders) in the past using `git filter-branch` which rewrites the history as-if it all happened in a subfolder which should help with log/blame by ensuring the filenames don't change abruptly

JohnTitor (approved):
Yeah, no objection. Thank you!

JohnTitor:
Ah, note that some crates use ctest2 in their tests, so maybe I should add a deprecation notice and encourage to migrate in the `ctest2`.

tgross35:
Merged! 🎉 https://github.com/rust-lang/libc/tree/d5278fd7a746572c0c0b31d4a3d74366fbf5243e/ctest. Thanks Alex and Yuki for the help here.

> I think I've done something [like this](https://stackoverflow.com/questions/16450423/git-rewrite-history-move-folders-and-files-up-to-parent-folders) in the past using `git filter-branch` which rewrites the history as-if it all happened in a subfolder which should help with log/blame by ensuring the filenames don't change abruptly

I wound up using `git filter-repo` (external tool recommended by git) rather than `git filter-branch` which worked out quite nice because it also rewrites shas mentioned in commit messages (e.g. reverts) and has a nice UI. The resulting history and blame look correct.

For future reference since I'm sure I'll need to do this again at some point:

```sh
# This was done on a fresh clone from ctest2

# Move to a subdirectory
git filter-repo --to-subdirectory-filter ctest

# Prefix all tags with `ctest-`
git filter-repo --tag-rename ':ctest-'

# The default merge messages are "merge pull request #nnn from user/branch"
# GH links these incorrectly in the new repo, so rename from `#nnn` to
# `JohnTitor/ctest2#nnn` (doesn't work for anything from the original `ctest`
# repo, but close enough).
echo 'regex:(^|\s)(#\d+)==>\1JohnTitor/ctest2\2' > messages.txt
git filter-repo --replace-message messages.txt

# After configuring a new remote (mine), push everything
git push --force --tags --prune tg-ctest2

# Now in the `libc` directory...

# Add that remote, fetch to update, etc
git remote add origin tg-ctest2 ...

# Merge, put something reasonable in the message
git merge tg-ctest2/... --allow-unrelated-histories

# Add whatever fixup commits are needed so CI passes,
# then push to `main` or use a PR.

# Tags didn't come through, so I just pushed them directly to rust-lang/libc
git push upstream tag 'ctest-*'
```

PR #4285, opened 2025-02-25
openbsd: add devname(3)

tobhe:
# Description

I'm planning to add OpenBSD support to the drm crate, for that I need [devname(3)](https://man.openbsd.org/devname.3)  to replace devname_r which is used on FreeBSD but not available on OpenBSD.

# Sources

https://man.openbsd.org/devname.3
https://github.com/openbsd/src/blob/master/include/stdlib.h#L283
https://github.com/openbsd/src/blob/master/lib/libc/gen/devname.c#L73

# Checklist

### DIFF
--- a/libc-test/semver/apple.txt
+++ b/libc-test/semver/apple.txt
@@ -1831,6 +1831,7 @@ cpu_type_t
 ctime
 ctime_r
 ctl_info
+devname
 difftime
 dirfd
 dirname
--- a/libc-test/semver/dragonfly.txt
+++ b/libc-test/semver/dragonfly.txt
@@ -1303,6 +1303,7 @@ cpuctl_cpuid_count_args_t
 cpuctl_msr_args_t
 cpuctl_update_args_t
 daemon
+devname
 devname_r
 difftime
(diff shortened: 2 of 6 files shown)


src/unix/bsd/netbsdlike/openbsd/mod.rs
tgross35, line 1: It looks like this may be available on all the BSDs we support; OpenBSD (your link), NetBSD https://man.netbsd.org/devname.3, FreeBSD https://man.freebsd.org/cgi/man.cgi?query=devname&sektion=3&n=1, and Apple https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/devname.3.html. Would you be able to try moving this definition to `unix/bsd/mod.rs` so they all get it?

tobhe: Should be fixed now

semarie:
LGTM. thanks

tgross35 (approved):
Awesome, thank you!

PR #4286, opened 2025-02-25
Add SysV semaphore constants

### LINKED ISSUE #4264: Add missing constants for System V semaphores
* Architecture tested: `x86_64-linux-unknown-gnu`, though to the best of my knowledge all linuxes should apply
* https://www.man7.org/linux/man-pages/man7/sysvipc.7.html

Note that the extern function definitions for the relevant APIs `semget` `semop` are already there, but the constants are not, making them harder to use. I'm currently using bindgen to work around this problem, the constants appear to be as follows:

```rust
pub const SEM_UNDO: i32 = 4096;
pub const GETPID: i32 = 11;
pub const GETVAL: i32 = 12;
pub const GETALL: i32 = 13;
pub const GETNCNT: i32 = 14;
pub const GETZCNT: i32 = 15;
pub const SETVAL: i32 = 16;
pub const SETALL: i32 = 17;
pub const SEM_STAT: i32 = 18;
pub const SEM_INFO: i32 = 19;
pub const SEM_STAT_ANY: i32 = 20;
```

These are all found in the `<sys/sem.h>` header.

bertptrs:
# Description

This patch adds missing constants required to use the SysV semaphore system, fixes #4264. Arguably, most of these constants should be `c_short` rather than `c_int`, as that is how they are used, but they already exist for other UNIXes as `c_int` so I opted to be consistent with those.

# Sources

- https://man.archlinux.org/man/semget.2.en
- https://man.archlinux.org/man/semctl.2.en
- https://man.archlinux.org/man/semop.2.en

Definitions: https://github.com/torvalds/linux/blob/cc8a0934d099b8153fc880a3588eec4791a7bccb/include/uapi/linux/sem.h#L7-L22

# Checklist


src/unix/linux_like/linux/mod.rs
```diff
@@ -2979,6 +2979,19 @@ pub const MSG_NOERROR: c_int = 0o10000;
 pub const MSG_EXCEPT: c_int = 0o20000;
 pub const MSG_ZEROCOPY: c_int = 0x4000000;
 
+pub const SEM_UNDO: c_int = 4096;
```
tgross35, line 2982: Could you change the literal to 0x1000 to match the Linux header? Please squash, otherwise lgtm

tgross35 (approved):
Thanks!

PR #4288, opened 2025-02-27
linux: Add new netlink flags

jbaublitz:
# Description

This change adds some new constants for the DELETE and ACK subset of netlink commands.

# Sources

[The documentation](https://www.man7.org/linux/man-pages/man7/netlink.7.html) appears to be out of date, but the header is `/usr/include/linux/netlink.h` and includes the following:

```
/* Modifiers to DELETE request */
#define NLM_F_NONREC    0x100   /* Do not delete recursively    */
#define NLM_F_BULK      0x200   /* Delete multiple objects      */
 
/* Flags for ACK message */
#define NLM_F_CAPPED    0x100   /* request was capped */
#define NLM_F_ACK_TLVS  0x200   /* extended ACK TVLs were included */
```

Header: https://github.com/torvalds/linux/blob/0fed89a961ea851945d23cc35beb59d6e56c0964/include/uapi/linux/netlink.h#L82-L87

# Checklist


tgross35 (changes requested):
Are these available on Sparc? CI is throwing some errors, but this may just be because our headers are somewhat old.

If that is the case, `libc-test/build.rs` will need to be updated to exclude them.

jbaublitz:
@tgross35 Thanks for getting back to me on this. I'm willing to investigate and see which constants are available on what platform in recent versions of the headers but I'm not quite sure what the best way to look into this is. What would you recommend?

tgross35:
Actually just try rebasing, the sparc headers in CI were updated recently so maybe things fixed themselves.

tgross35:
> [The documentation](https://www.man7.org/linux/man-pages/man7/netlink.7.html) appears to be out of date, but the header is `/usr/include/linux/netlink.h` and includes the following:

A permalink to the header is preferable over docs links anyway, since things change. I updated the top post 👍 

Thanks!

PR #4290, opened 2025-03-02
closefrom: add NetBSD, OpenBSD, DragonflyBSD

tobhe:
# Description

closefrom is actually widely adopted in *BSD. The only one that seems to not support it at this point is macOS.
This effectively adds closefrom for NetBSD, OpenBSD and DragonFly.

# Sources

https://leaf.dragonflybsd.org/cgi/web-man?command=closefrom
https://man.netbsd.org/closefrom.3
https://man.openbsd.org/closefrom.2

# Checklist

### DIFF
--- a/libc-test/semver/dragonfly.txt
+++ b/libc-test/semver/dragonfly.txt
@@ -1296,6 +1296,7 @@ clock_getcpuclockid
 clock_getres
 clock_nanosleep
 clock_settime
+closefrom
 cmsgcred
 cmsghdr
 cpuctl_cpuid_args_t
--- a/libc-test/semver/netbsd.txt
+++ b/libc-test/semver/netbsd.txt
@@ -1252,6 +1252,7 @@ clearerr
 clock_getres
 clock_nanosleep
 clock_settime
+closefrom
 cmsghdr
 consttime_memequal
 daemon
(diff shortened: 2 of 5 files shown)


semarie:
the return type of `closefrom` is `c_int` on OpenBSD, NetBSD and DragonflyBSD (it is `void` only on FreeBSD).

so the tests are failing on OpenBSD : 
```
  cargo:warning=/data/semarie/repos/rust/build-libc/build/x86_64-unknown-openbsd/debug/build/libc-test-6990a9227fb77c31/out/main.c:16166:24: error: incompatible function pointer types returning 'int (int)' from a function with result type
'void (*)(int)' [-Werror,-Wincompatible-function-pointer-types]
  cargo:warning=                return closefrom;
  cargo:warning=                       ^~~~~~~~~
  cargo:warning=1 error generated.
```

tobhe:
Absolutely correct @semarie, thanks! No idea how i didn't catch that but it should be fixed now.

semarie:
thanks, it passes the test now. the i686-linux-android fails seems unrelated.

tgross35 (approved):
Thanks! Yeah, android CI is very flaky.

Issue #3305, opened 2023-07-17
Unable to run tests in libc-test using musl (Alpine Linux 3.18.2)


CuriouslyCurious:
I am unable to run the tests in `libc-test` using musl on Alpine Linux.

To reproduce:
1. Clone the libc repo.
2. `cd libc-test && cargo test`

This fails with:
```sh
  ...
  cargo:rerun-if-changed=../src/lib.rs
  cargo:rerun-if-changed=../src/macros.rs
  cargo:rerun-if-changed=../src/fixed_width_ints.rs
  cargo:rerun-if-changed=../src/unix/mod.rs
  cargo:rerun-if-changed=../src/unix/linux_like/mod.rs
  cargo:rerun-if-changed=../src/unix/linux_like/linux/mod.rs
  cargo:rerun-if-changed=../src/unix/linux_like/linux/musl/mod.rs
  cargo:rerun-if-changed=../src/unix/linux_like/linux/musl/lfs64.rs
  cargo:rerun-if-changed=../src/unix/linux_like/linux/musl/b64/mod.rs
  cargo:rerun-if-changed=../src/unix/linux_like/linux/musl/b64/x86_64/mod.rs
  cargo:rerun-if-changed=../src/unix/linux_like/linux/arch/mod.rs
  cargo:rerun-if-changed=../src/unix/linux_like/linux/arch/generic/mod.rs
  cargo:rerun-if-changed=../src/unix/linux_like/linux/align.rs
  OPT_LEVEL = Some("0")
  HOST = Some("x86_64-unknown-linux-musl")
  cargo:rerun-if-env-changed=CC_x86_64-unknown-linux-musl
  CC_x86_64-unknown-linux-musl = None
  cargo:rerun-if-env-changed=CC_x86_64_unknown_linux_musl
  CC_x86_64_unknown_linux_musl = None
  cargo:rerun-if-env-changed=HOST_CC
  HOST_CC = None
  cargo:rerun-if-env-changed=CC
  CC = None
  cargo:rerun-if-env-changed=CFLAGS_x86_64-unknown-linux-musl
  CFLAGS_x86_64-unknown-linux-musl = None
  cargo:rerun-if-env-changed=CFLAGS_x86_64_unknown_linux_musl
  CFLAGS_x86_64_unknown_linux_musl = None
  cargo:rerun-if-env-changed=HOST_CFLAGS
  HOST_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS
[...]

Issue #3149, opened 2023-03-12
Missing `tls.h` bindings


JonathanWoollett-Light:
Bindings are missing for TLS specifically [`include/uapi/linux/tls.h`](https://github.com/torvalds/linux/blob/master/include/uapi/linux/tls.h)

howjmay:
May I try this?

JohnTitor:
@howjmay Great, go ahead!

PR #4296, opened 2025-03-04
linux: add missing tls bindings

mbyx:
# Description
Adds missing TLS bindings for linux as mentioned in #3149. Tests should pass as long as the kernel used for testing implements them.

# Sources
Previously only a part of the header had been implemented, now everything has been bound except for placeholder values (__TLS_INFO_MAX, TLS_INFO_MAX).
https://github.com/torvalds/linux/blob/99fa936e8e4f117d62f229003c9799686f74cebc/include/uapi/linux/tls.h

# Checklist

I was unable to test locally due to errors similar to #3305.


tgross35:
Permalink to the headers in their current state https://github.com/torvalds/linux/blob/99fa936e8e4f117d62f229003c9799686f74cebc/include/uapi/linux/tls.h (you can get these from the triple dot menu at the top right of the file view).

The CI failure just needs you to sort `libc-test/semver/linux.txt`, FreeBSD is spurious.

mbyx:
The test that failed gives the same errors that I experienced when testing my commit locally. Is it the CI system or something wrong with my implementation? It also cancelled the rest, so I can't check if the problem will occur for them all.

tgross35 (approved):
LGTM and the CI problem should be fixed with https://github.com/rust-lang/libc/pull/4311. Could you rebase and squash?

tgross35:
Hm, are these headers just not available on loongarch musl for some reason? Cc @heiher in case you know something.

heiher:
> Hm, are these headers just not available on loongarch musl for some reason? Cc @heiher in case you know something.

Thanks for your CC. I have pushed a PR (#4320) for that, and I believe it has fixed this issue.

tgross35:
This should work now, @mbyx could you squash please?

mbyx:
Sorry, but this is my first time having to squash, I think I'm not doing it properly, could you please guide me?

tgross35:
The squash looks good, thanks!

PR #4299, opened 2025-03-05
linux: Add devmem constants and structs

tammela:
# Description

Adds devmem constants and structs. Devmem was merged recently (6.12).
I had also add all the other missing `SO_` constants for completion.
As usual, musl is way behind (even on 1.2) so all of these will be skipped there.

# Sources

# Checklist


tgross35 (approved):
Implementation LGTM. Could you add permalinks to the headers in the PR description and fix the style failure?

tgross35 (changes requested):
Are `SO_RESERVE_MEM` and similar available on sparc? CI seems to think they aren't. If they are more recent, the test crate's build script will need to be updated.

tammela:
My guess is that there will one more architecture that this PR will be break.
Anyways I posted this: https://github.com/rust-lang/libc/pull/4317
Should fix for sparc and reactivate testing on s390x.

tgross35 (approved):
Looks like things are passing now, thanks for the CI fix!

PR #4308, opened 2025-03-07
Fix build & link on cygwin

Berrysoft:
It's a fix for #4279 . My fault:(


Berrysoft:
Let me explain this PR:
* It fixes the usage of `f!`. The signatures of functions should have return types, even they are `-> ()`.
* It fixes the link of `strerror_r`.

tgross35 (approved):
The macros are unfriendly :) thanks!

PR #4311, opened 2025-03-10
Change the range syntax that is giving `ctest` problems

tgross35:
`ctest` is iffy about whether or not it accepts `..=` syntax, and I can't figure out what makes it decide whether or not to accept it and sometimes random changes seem to make things fail, so just replace the syntax.

This is simpler anyway, and closer matches the upstream definition [1].

Link: https://github.com/torvalds/linux/blob/80e54e84911a923c40d7bee33a34c1b4be148d7a/Makefile#L1316 [1]


tgross35:
Also removing the i586 MSVC target https://github.com/rust-lang/rust/pull/137957

PR #4316, opened 2025-03-10
linux_like: add F_SEAL_EXEC

lucab:
# Description

This adds the `F_SEAL_EXEC` flag for `fcntl(2)`, which has been introduced in Linux kernel 6.3.

# Sources

https://github.com/torvalds/linux/commit/6fd7353829cafc4067aad9eea0dc95da67e7df16

# Checklist


tgross35 (approved):
Double checked that this is on Bionic https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/kernel/uapi/linux/fcntl.h;l=27;drc=65552ba5db8f299144172d4464350f1363a95304.

Thanks!

PR #4317, opened 2025-03-10
ci: Fixes for sparc and s390x

tammela:
# Description

Fixes the 'cannot find libc' error plaguing s390x and sparc. A side effect is that now these CIs will be tested on much more recent kernel headers (6.11).

# Sources

# Checklist

### DIFF
--- a/ci/docker/sparc64-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/sparc64-unknown-linux-gnu/Dockerfile
@@ -1,11 +1,4 @@
-# FIXME(sparc): newer versions of Ubuntu get the following errors
-# ```
-# /prog: /lib/sparc64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by /prog)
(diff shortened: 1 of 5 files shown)


tgross35 (approved):
Awesome, thank you!

Do you know why the "`GLIBC_2.39' not found (required by /prog)" was showing up before for similar changes but no longer is?

tgross35:
This seems to have doubled s390x CI time from ~5 minutes to ~10 minutes. Not too much of a concern since it doesn't increase our total CI time by much, but I'm not sure why this would have changed.

tammela:
> This seems to have doubled s390x CI time from ~5 minutes to ~10 minutes. Not too much of a concern since it doesn't increase our total CI time by much, but I'm not sure why this would have changed.

So it turns out s390x was not running at all, it was failing 5 times with the libc error then silently succeeding

tgross35:
Oh wtf, great catch. We can probably clean up https://github.com/rust-lang/libc/blob/1c65dda4fa8d0e9bc2f3a4d4dcadda2ebacbf33b/ci/run.sh#L96-L124 at some point then.

PR #4320, opened 2025-03-11
ci: install musl from source for loongarch64

heiher:
# Description

This patch switches to installing musl from source for the LoongArch64 target to ensure compatibility between kernel headers and Rust bindings.

# Sources

None.

# Checklist


tgross35:
The change itself looks fine to me, but will this cause any problems with time_t? Since the change from 32- to 64-bit time happened at 1.2.0.

tgross35 (approved):
Nevermind, this only affects a 64-bit arch that didn't have the time change. Hopefully this won't cause any problems using a different version than the rest of the platforms, but I don't have any reason not to try it.

Issue #4323, opened 2025-03-11
Export ioctl request generator macros


Mr2Brainless:
Currently the macros `_IO`, `_IOW`, `_IOR`, `_IOWR` are not being exported. To get driver specific ioctl request codes which probably do not belong in `libc`, like the DRM ioctls, it would be useful to have them exported.

PR #4114, opened 2024-11-19
add various `PTP_` consts

folkertdev:
The constants from https://github.com/rust-lang/libc/pull/3865, hopefully.

# Description

# Sources

# Checklist

- [x] Relevant tests in `libc-test/semver` have been updated
- [x] No placeholder or unstable values like `*LAST` or `*MAX` are
  included (see [#3131](https://github.com/rust-lang/libc/issues/3131))
- [ ] Tested locally (`cd libc-test && cargo test --target mytarget`);
  especially relevant for platforms that may not be checked in CI

r? @tgross35

PR #4325, opened 2025-03-12
linux: Export ioctl request generator macros

Mr2Brainless:
# Description

This exports the existing macros / const fns `_IO`, `_IOW`, `_IOR`, `_IOWR`. Related issue #4323.

# Sources

No new source. Old code is from #4114.

# Checklist

Fixes: https://github.com/rust-lang/libc/issues/4323


tgross35 (approved):
Makes sense, thanks!

PR #4327, opened 2025-03-12
Add missing `filedesc` and `fdescenttbl` in FreeBSD

GuillaumeGomez:
Types I need for sysinfo to allow to know how many open files a FreeBSD process has.

Item declarations can be found there:
 * https://github.com/freebsd/freebsd-src/blob/58164dcb55d62ca73b5e550b8344bf61e2d8a47a/sys/sys/_lock.h
 * https://github.com/freebsd/freebsd-src/blob/58164dcb55d62ca73b5e550b8344bf61e2d8a47a/sys/sys/filedesc.h
 * https://github.com/freebsd/freebsd-src/blob/58164dcb55d62ca73b5e550b8344bf61e2d8a47a/sys/sys/_sx.h

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -1355,6 +1355,47 @@ s! {
         pub strchange_instrms: u16,
         pub strchange_outstrms: u16,
     }
+
+    pub struct filedesc {
+        pub fd_files: *mut fdescenttbl,
+        pub fd_map: *mut c_ulong,
+        pub fd_freefile: c_int,
+        pub fd_refcnt: c_int,
+        pub fd_holdcnt: c_int,
+        fd_sx: sx,
+        fd_kqlist: kqlist,
+        pub fd_holdleaderscount: c_int,
+        pub fd_holdleaderswakeup: c_int,
+    }
+
+    pub struct fdescenttbl {
+        pub fdt_nfiles: c_int,
+        fdt_ofiles: [*mut c_void; 0],
+    }
+
+    // FIXME: Should be private.
+    #[doc(hidden)]
+    pub struct sx {
+        lock_object: lock_object,
+        sx_lock: crate::uintptr_t,
+    }
+
+    // FIXME: Should be private.
+    #[doc(hidden)]
+    pub struct lock_object {
+        lo_name: *const c_char,
+        lo_flags: c_uint,
+        lo_data: c_uint,
+        // This is normally `struct  witness`.
+        lo_witness: *mut c_void,
+    }
+
+    // FIXME: Should be private.
+    #[doc(hidden)]
+    pub struct kqlist {
+        tqh_first: *mut c_void,
+        tqh_last: *mut *mut c_void,
+    }
 }
 
 s_no_extra_traits! {
--- a/src/unix/bsd/freebsdlike/freebsd/freebsd11/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/freebsd11/mod.rs
@@ -51,9 +51,8 @@ s! {
         // This is normally "struct vnode".
         /// Pointer to executable file.
         pub ki_textvp: *mut c_void,
-        // This is normally "struct filedesc".
         /// Pointer to open file info.
-        pub ki_fd: *mut c_void,
+        pub ki_fd: *mut crate::filedesc,
         // This is normally "struct vmspace".
         /// Pointer to kernel vmspace struct.
         pub ki_vmspace: *mut c_void,
--- a/src/unix/bsd/freebsdlike/freebsd/freebsd12/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/freebsd12/mod.rs
@@ -59,9 +59,8 @@ s! {
         // This is normally "struct vnode".
         /// Pointer to executable file.
         pub ki_textvp: *mut c_void,
-        // This is normally "struct filedesc".
         /// Pointer to open file info.
-        pub ki_fd: *mut c_void,
+        pub ki_fd: *mut crate::filedesc,
         // This is normally "struct vmspace".
         /// Pointer to kernel vmspace struct.
         pub ki_vmspace: *mut c_void,
--- a/src/unix/bsd/freebsdlike/freebsd/freebsd13/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/freebsd13/mod.rs
@@ -69,9 +69,8 @@ s! {
         // This is normally "struct vnode".
         /// Pointer to executable file.
         pub ki_textvp: *mut c_void,
-        // This is normally "struct filedesc".
         /// Pointer to open file info.
-        pub ki_fd: *mut c_void,
+        pub ki_fd: *mut crate::filedesc,
         // This is normally "struct vmspace".
         /// Pointer to kernel vmspace struct.
         pub ki_vmspace: *mut c_void,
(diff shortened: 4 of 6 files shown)


tgross35:
Cc @asomers for FreeBSD

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1355,6 +1357,66 @@ s! {
         pub strchange_instrms: u16,
         pub strchange_outstrms: u16,
     }
+
+    pub struct filedesc {
+        pub fd_files: *mut fdescenttbl,
+        pub fd_map: *mut c_void,
```
asomers, line 1363:
```suggestion
        pub fd_map: *mut c_ulong,
```

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1355,6 +1357,66 @@ s! {
...
+    }
+
+    pub struct fdescenttbl {
+        pub fdt_nfiles: c_int,
+        pub fdt_ofiles: [filedescent; 0],
+    }
+
+    pub struct filedescent {
+        // This is normally `struct file`.
+        pub fde_file: *mut c_void,
```
asomers, line 1380: Are you defining it as "c_void" instead of `struct file` just so you won't have to add a definition for the latter, because you don't need it?  In that case, I suggest you make the field private.  That way it will be easier to change the field later.

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1355,6 +1357,66 @@ s! {
...
         pub strchange_outstrms: u16,
     }
+
+    pub struct filedesc {
+        pub fd_files: *mut fdescenttbl,
+        pub fd_map: *mut c_void,
+        pub fd_freefile: c_int,
+        pub fd_refcnt: c_int,
+        pub fd_holdcnt: c_int,
+        pub fd_sx: sx,
```
asomers, line 1367: You should make this field private and opaque.  It isn't used by userspace.

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1355,6 +1357,66 @@ s! {
...
+        pub fd_kqlist: kqlist,
```
asomers, line 1368: This field too isn't used by userspace.

src/unix/bsd/freebsdlike/freebsd/mod.rs
GuillaumeGomez: Good catch, followed `NDSLOTTYPE` wrongly.

src/unix/bsd/freebsdlike/freebsd/mod.rs
GuillaumeGomez: Yes, gonna do that.

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1355,6 +1357,66 @@ s! {
...
+        pub fd_holdleaderscount: c_int,
+        pub fd_holdleaderswakeup: c_int,
+    }
+
+    pub struct fdescenttbl {
+        pub fdt_nfiles: c_int,
```
GuillaumeGomez, line 1374: So all I need is this file. But if I understand correctly, I still need `filedescent` definition because of the `fdt_ofiles` field.

src/unix/bsd/freebsdlike/freebsd/mod.rs
GuillaumeGomez: I made them private, but wdym by opaque? To get the size and alignment right, we still need to define the inner types, or do you mean something else?

asomers: You should be able to get the size and alignment correct with a suitably constructed array of bytes, longs, etc.  But if that's too much work then it's ok to define the real structs instead.

src/unix/bsd/freebsdlike/freebsd/mod.rs
asomers: I think you can define fdt_ofiles as a `[*mut c_void; 0]` instead to get the correct alignment.

src/unix/bsd/freebsdlike/freebsd/mod.rs
GuillaumeGomez: I find it preferrable to have definitions matching C as it makes it easier for us to update it if changed but really as you prefer.

src/unix/bsd/freebsdlike/freebsd/mod.rs
GuillaumeGomez: If so that'd make things MUCH simpler. Let's give it a try.

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1355,6 +1357,52 @@ s! {
...
+        fdt_ofiles: [*mut c_void; 0],
+    }
+
+    #[doc(hidden)]
+    pub struct sx {
+        lock_object: lock_object,
+        sx_lock: crate::uintptr_t,
+    }
+
+    pub struct lock_object {
```
asomers, line 1384: Should make this one hidden, too.  Or better yet, just make all of these structures private.

asomers (changes requested):

src/unix/bsd/freebsdlike/freebsd/mod.rs
GuillaumeGomez: The macro doesn't allow it sadly.

src/unix/bsd/freebsdlike/freebsd/mod.rs
tgross35: For 1.0 we are going to make all opaque / padding fields `MaybeUninit`, that could be an option here

tgross35:
FreeBSD13 CI has a spurious segfault. I have no idea why, haven't had a chance to look into it.

Retried the job

GuillaumeGomez:
CI passed! \o/

GuillaumeGomez:
@tgross35 Is there anything else to be done here?

GuillaumeGomez:
Rebased on `main` branch.

tgross35:
I think this looks fine to me with one note that there should be a comment about why some of these are `#[doc(hidden)]`. I'm not sure about the change to `kinfo_proc`; that is breaking so I'm not sure @asomers would want to backport it, in which case that needs to be split to a separate commit so I can backport the rest.

Could you add a permalink to the relevant headers in the PR descripton?

GuillaumeGomez:
About the changes to `kinfo_proc`: you cannot access the data pointed by the pointer as is because it's a kernel pointer, you need to send this pointer to `kvm_read` to get the actual struct.

So although the pointer type changes, is it really an issue considering you cannot access the data as is?

GuillaumeGomez:
Oh also I added the permalinks in the first comment.

tgross35:
> About the changes to `kinfo_proc`: you cannot access the data pointed by the pointer as is because it's a kernel pointer, you need to send this pointer to `kvm_read` to get the actual struct.
> 
> So although the pointer type changes, is it really an issue considering you cannot access the data as is?

Good point, it's probably not a problem in practice then. I'm fine with backporting everything as long as Alan okays it.

GuillaumeGomez:
Thanks, that'd allow me to remove a lot of code from `sysinfo`. :)

asomers:
I'm ok with backporting the kinfo_proc change to the 0.2 branch.

tgross35:
@asomers was there anything else you wanted here or is this good to merge?

asomers:
> @asomers was there anything else you wanted here or is this good to merge?

I would prefer that types like `sx` weren't declared, since they will never be used in userland.  It would be better to make those fields just an opaque byte array.  But if you like it this way instead, I'm ok with it.

GuillaumeGomez:
I'd prefer to keep the struct declaration. In future version, we can make the macro allow to have private types.

PR #3116, opened 2023-02-11
prctl SME flags for gnu arm64.

devnexen:

PR #4328, opened 2025-03-12
Clean up some test skips on Linux

tammela:
# Description

Since I added a bunch of new skips I decided to remove some of the stale ones

# Sources

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -3919,11 +3919,6 @@ fn test_linux(target: &str) {
             return true;
         }
 
-        // FIXME(linux): sparc64 CI has old headers
-        if sparc64 && (ty == "uinput_ff_erase" || ty == "uinput_abs_setup") {
-            return true;
-        }
-
         // FIXME(#1558): passing by value corrupts the value for reasons not understood.
         if (gnu && sparc64) && (ty == "ip_mreqn" || ty == "hwtstamp_config") {
             return true;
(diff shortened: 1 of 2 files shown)


src/unix/linux_like/linux/gnu/b64/aarch64/mod.rs
```diff
@@ -631,10 +631,6 @@ pub const PR_PAC_APGAKEY: c_ulong = 1 << 4;
 
 pub const PR_SME_SET_VL: c_int = 63;
 pub const PR_SME_GET_VL: c_int = 64;
-pub const PR_SME_VL_LEN_MAX: c_int = 0xffff;
-
-pub const PR_SME_SET_VL_INHERIT: c_ulong = 1 << 17;
-pub const PR_SME_SET_VL_ONE_EXEC: c_ulong = 1 << 18;
```
tgross35, line 637: What is the reason for removing these? I can't actually find them defined anywhere, @devnexen do you remember what these are from in https://github.com/rust-lang/libc/pull/3116?

tgross35: If it is something that makes sense to remove, I'd rather only deprecate it here so I can backport the PR. We'll go through and do a sweep of deprecated items before 1.0.

tammela:
@tgross35 I'm unavailable currently and will be for some time. Trevor (or anyone else), feel free to continue this PR if you can!

tgross35:
I updated a lot of these in dd44cb603e77e5bca7d9de0af2ee5c6e189e08d8, so I don't think there is much remaining in this PR that applies. Thank you for getting the ball rolling here! I'm sure there is still more work to be done if you ever circle back to it.

Issue #4045, opened 2024-11-15
Remove solarish compat for `openpty` and `forkpty`


tgross35:
We currently polyfill these functions for solarish in https://github.com/rust-lang/libc/blob/3a0b0444d7206aa3feb30297c005597a960d0b33/src/unix/solarish/compat.rs, but it would be better to phase these out so we're not providing any implementations in `libc`.

It looks like Solaris supports `openpty` and `forkpty` https://docs.oracle.com/cd/E88353_01/html/E37843/openpty-3c.html, and Illumos has had them for at least a few years https://www.illumos.org/issues/5386, so at this point our implementation could be replaced with bindings.

devnexen:
Correction, those are not available on illumos, either never been accepted or reverted ; most likely the former.

tgross35:
Oh that's interesting, thanks for the context. Hm, I'm not sure what to do here in that case - this is about the only case of polyfill in this crate and ideally we wouldn't have any. I guess maybe we can remove these with 1.0.

pfmooney:
> I guess maybe we can remove these with 1.0.

Doing so would break us (illumos) in the nix crate, for which they were added.  In the fullness of time, I do hope/expect we'll add implementations to our native libc, but even once that happens, it will be some time until we can expect downstream distros to have it rolled out to a suitable cross section of the user base.

PR #4329, opened 2025-03-13
solarish:  Restrict openpty and forkpty polyfills to illumos, replace Solaris implementation with FFI

yuvraj-wale:
# Description

- This PR modifies the openpty and forkpty implementations in the Solarish layer:
    - Restricts the existing polyfill implementations to illumos using cfg(illumos).
    - Removes the Solaris-specific polyfill implementations.
    - Replaces them with FFI definitions where necessary.

-  Related to #4045 

# Sources

- Relevant discussion : #4045 

# Checklist

### DIFF
--- a/src/unix/solarish/compat.rs
+++ b/src/unix/solarish/compat.rs
@@ -53,6 +53,7 @@ unsafe fn bail(fdm: c_int, fds: c_int) -> c_int {
     return -1;
 }
 
+#[cfg(target_os = "illumos")]
 pub unsafe fn openpty(
     amain: *mut c_int,
     asubord: *mut c_int,
@@ -123,6 +124,7 @@ pub unsafe fn openpty(
     0
 }
 
+#[cfg(target_os = "illumos")]
 pub unsafe fn forkpty(
     amain: *mut c_int,
     name: *mut c_char,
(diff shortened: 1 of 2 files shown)


pfmooney (changes requested):
illumos still relies on these implementations.  If you've confirmed that they are not necessary for Solaris (in which case wouldn't defining them as FFI symbols be necessary?), you should put them behind a `cfg(illumos)` guard rather than outright removing them.

yuvraj-wale:
Yes my bad, I’ll update it right away. Thanks!

src/unix/solarish/compat.rs
```diff
@@ -53,6 +53,147 @@ unsafe fn bail(fdm: c_int, fds: c_int) -> c_int {
...
+        winp: *const crate::winsize,
+    ) -> c_int;
+
+    pub fn forkpty(
+        amain: *mut c_int,
+        name: *mut c_char,
+        termp: *const crate::termios,
+        winp: *const crate::winsize,
+    ) -> crate::pid_t;
+}
```
pfmooney, line 195: These should go in `solaris.rs`, where they don't require the guard.

tgross35:
@psumbera would you mind double checking this for Solaris?

psumbera:
As can be seen from failed CI test. Solaris don't define these with 'const`.

So it's just:
```
       int openpty(int *amaster, int *aslave, char *name,
            struct termios *termp, struct winsize *winp);

       pid_t forkpty(int *amaster, char *name, struct termios *termp,
            struct winsize *winp);
```

yuvraj-wale:
Updated, thanks for the help! 

cc @tgross35

tgross35 (approved):
LGTM, please rebase and squash. Also the title & PR description should be updated.

tgross35:
Thanks!

Issue #4322, opened 2025-03-11
Some utmp symbols still missing on musl


Ecordonnier:
This is a follow-up of https://github.com/rust-lang/libc/issues/3190

The function utmpxname is missing from the musl implementation ( src/unix/linux_like/linux/musl/mod.rs ), even though this function is implemented (as a stub doing nothing) in musl (see https://git.musl-libc.org/cgit/musl/tree/include/utmpx.h ).

As reference, see this [commit](https://github.com/rust-lang/libc/commit/98fd3188e7d46db84d321475344f2f82d958b0c7) which added the function for solaris, as well as the man page: https://linux.die.net/man/3/endutxent

The symbols "libc::__UT_HOSTSIZE", "libc::__UT_LINESIZE", "libc::__UT_NAMESIZE" and "libc::[ACCOUNTING](https://github.com/kraj/musl/blob/kraj/master/include/utmp.h#L10)" are also missing.

PR #4332, opened 2025-03-15
Add missing utmpx apis for linux musl

Ecordonnier:
# Description

Add the utmp APIs which were missing from musl.
Close https://github.com/rust-lang/libc/issues/4322

# Sources

This is based on https://git.musl-libc.org/cgit/musl/tree/include/utmpx.h
Note that musl implements only stub functions for utmp functionalities, because they consider it unsafe.

# Checklist

- [x] Tested locally (`cd libc-test && cargo test --target mytarget`);  (unrelated error about #include <linux/errqueue.h>)
  
Signed-off-by: Etienne Cordonnier <ecordonnier@snap.com>

### DIFF
--- a/libc-test/semver/linux-musl.txt
+++ b/libc-test/semver/linux-musl.txt
@@ -1,3 +1,4 @@
+ACCOUNTING
 AF_IB
 AF_MPLS
 AF_XDP
@@ -37,6 +38,9 @@ RWF_HIPRI
 RWF_NOWAIT
 RWF_SYNC
 USER_PROCESS
+UT_HOSTSIZE
+UT_LINESIZE
+UT_NAMESIZE
 _CS_V6_ENV
 _CS_V7_ENV
 adjtimex
@@ -82,3 +86,4 @@ reallocarray
 setutxent
 tcp_info
 timex
+utmpxname
--- a/src/unix/linux_like/linux/musl/mod.rs
+++ b/src/unix/linux_like/linux/musl/mod.rs
@@ -656,7 +656,7 @@ pub const INIT_PROCESS: c_short = 5;
 pub const LOGIN_PROCESS: c_short = 6;
 pub const USER_PROCESS: c_short = 7;
 pub const DEAD_PROCESS: c_short = 8;
-// musl does not define ACCOUNTING
+pub const ACCOUNTING: c_short = 9;
 
 pub const SFD_CLOEXEC: c_int = 0x080000;
 
@@ -888,6 +888,10 @@ pub const _CS_V7_ENV: c_int = 1149;
 
 pub const CLONE_NEWTIME: c_int = 0x80;
 
+pub const UT_HOSTSIZE: usize = 256;
+pub const UT_LINESIZE: usize = 32;
+pub const UT_NAMESIZE: usize = 32;
+
 cfg_if! {
     if #[cfg(target_arch = "s390x")] {
         pub const POSIX_FADV_DONTNEED: c_int = 6;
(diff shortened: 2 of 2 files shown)


Ecordonnier:
It compiles and I tested it by building uutils-coreutils with it as dependency, however I haven't managed to run cargo test on ubuntu 24.04:

```
~/dev/libc/libc-test$ cargo test --target x86_64-unknown-linux-musl
   Compiling proc-macro2 v1.0.94
   Compiling unicode-ident v1.0.18
...
   Compiling libc-test v0.1.0 (/home/ecordonnier/dev/libc/libc-test)
The following warnings were emitted during compilation:

warning: libc-test@0.1.0: src/errqueue.c:2:10: fatal error: linux/errqueue.h: No such file or directory
warning: libc-test@0.1.0:     2 | #include <linux/errqueue.h>
warning: libc-test@0.1.0:       |          ^~~~~~~~~~~~~~~~~~
warning: libc-test@0.1.0: compilation terminated.
```

Do you know what the issue is? I do have the required header-file:
```
~/dev/libc/libc-test$ ls /usr/include/linux/errqueue.h 
/usr/include/linux/errqueue.h
```

Ecordonnier:
Please also consider this for backporting to 0.2

Ecordonnier:
The header-file error seems to be a generic error when compiling with musl-gcc: https://stackoverflow.com/questions/42128091/linking-to-musl-with-other-headers-from-usr-include

tgross35 (changes requested):
Considering musl's stance here, what failures motivate this addition? It seems like the API consumer may be better off disabling utmpx functionality on musl rather than having surprising untested nop behavior, and I'm wondering if we wouldn't be better off leaving these undefined to discourage running into that footgun.

If you still think this is worth having, it would be preferable to move the existing definitions from `linux/gnu/mod.rs` to `linux/mod.rs` so glibc, musl, and uclibc all get covered.

Ecordonnier:
@tgross35 I understand your concern. I was thinking about this as well before the implementation and was hesitating as well. However my reasoning for including it in libc is the following:

- IMO the libc crate should provide exactly what musl is providing, and not make its own decisions on excluding certain functions, because this introduces even more complexity for developers (then they need to know not only the implementation of musl, but the implementation of the libc crate as well). The musl library took the decision to publish those functions as stub.

- my motivation for this addition is that uutils-coreutils does not compile for musl because of those missing definitions ( https://github.com/uutils/coreutils/issues/1361 ). It is easily possible to exclude some utilities for musl at install time if desired (the coreutils utillities using utmpx are "pinky", "uptime", "users", "who"), but it is desirable to be able to compile everything with musl, since the goal of the uutils coreutils project is to have 100% parity with the C coreutils project, and thus have the exact same behavior than GNU coreutils (written in C) compiled with musl (there is code in uutils coreutils handling the errors of the utmpx functions and printing some error messages, so being able to compile those functions would avoid adding "[cfg(not musl)]" lines above each call to a utmpx function).

Regarding your request to make those generic, please note that since constants are prefixed with "__" in glibc, but not in musl, the constant is e.g. "UT_HOSTSIZE" in musl, but "__UT_HOSTSIZE" in glibc. Do you want me to move a subset to the generic file?

tgross35:
I think bug-for-bug tools like uutils are probably somewhat on the rare side. Maybe let's do this: we can add the functions but immediately mark them deprecated so the user is likely to get a warning. That way they are available if needed but less likely to be used by complete accident, and `uutils` or anyone who expects the behavior can just silence the lint on a case-by-case/function/module basis.

```rust
// NB: not actually intended for removal; this is a user warning that the function may not
// behave as expected
#[deprecated(since="0.2.172",
    note="musl provides `utmp` as stubs and an alternative should be preferred; see \
    https://wiki.musl-libc.org/faq.html"
)]
```

If that sounds reasonable, would you mind adding the same attribute to other functions that musl has stubbed out?

> Regarding your request to make those generic, please note that since constants are prefixed with "__" in glibc, but not in musl, the constant is e.g. "UT_HOSTSIZE" in musl, but "__UT_HOSTSIZE" in glibc. Do you want me to move a subset to the generic file?

Didn't realize that, nope it's fine!

Ecordonnier:
I think it is a good idea. I have added the deprecation warnings using a macro. I've also tested that the warnings are printed when compiling uutils-coreutils:
```
ecordonnier@LC49NTQ3:~/dev/coreutils$ cargo build --target  x86_64-unknown-linux-musl -p uu_users
   Compiling libc v1.0.0-alpha.1 (/home/ecordonnier/dev/libc)
   Compiling uucore v0.0.30 (/home/ecordonnier/dev/coreutils/src/uucore)
warning: use of deprecated function `libc::endutxent`: musl provides `utmp` as stubs and an alternative should be preferred; see https://wiki.musl-libc.org/faq.html
  --> src/uucore/src/lib/features/utmpx.rs:43:15
   |
43 | pub use libc::endutxent;
   |               ^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default
   ...
```

Ecordonnier:
This is ready to be merged IMO.

tgross35 (approved):
Sorry this fell off my radar, thanks for the changes. LGTM

PR #4336, opened 2025-03-17
Correct the definition of `pthread_t` on Darwin

### LINKED ISSUE #2903: Incorrect `pthread_t` definition on Darwin (should be a raw pointer, not an `uintptr_t`)
This is a bug, but not one we can probably fix without breaking too much (it would break [`std`'s reexport of `pthread_t`](https://doc.rust-lang.org/nightly/std/os/unix/raw/type.pthread_t.html)). Aside from breaking code that treats it directly as an integer, it would mean `pthread_t` would suddenly become `!Send` and `!Sync`, and I wouldn't be surprised if that's a problem for some.

Anyway, on Darwin-based OSes, `pthread_t` should be a pointer, but is instead [an integer](https://github.com/rust-lang/libc/blob/master/src/unix/bsd/mod.rs#L6). I noticed this when trying to pass `ptr::null_mut()` into a function that Apple documents as having special behavior when passed a `NULL` for `pthread_t`.

However, it's worth noting that the structure it points to is *not* opaque (at least not fully), and low level code on Darwin could potentially need to access it (at the moment this is likely not possible to do correctly on stable Rust, due to the lack of stable `"C-unwind"`). Specifically it has a list of functions that may be manipulated by code that wants to have cleanup functions run on thread cancellation. This list is directly manipulated by the `pthread_cleanup_push` and `pthread_cleanup_pop` function-style C macros, but this would be done directly by Rust code, which cannot use such macros. This is pretty niche, but is a case that's slightly relevant to https://github.com/rust-lang/rust/issues/95496.

The correct definition on Darwin would look more like:

```rs
s! {
    pub struct __darwin_pthread_handler_rec {
        pub __routine: Option<unsafe extern "C" fn(arg1: *mut ::c_void)>,
        pub __arg: *mut ::c_void,
        pub __next: *mut __darwin_pthread_handler_rec,
    }
    pub struct _opaque_pthread_t {
        pub __sig: ::c_long,
        pub __cleanup_stack: *mut __darwin_pthread_handler_rec,
        pub __opaque: [::c_char; __PTHREAD_SIZE__],
    }
}

pub type __darwin_pthread_t = *mut _opaque_pthread_t;

pub type pthread_t = __darwin_pthread_t;
```

I suppose it's fine for code that cares about this to maintain its own bindings, though.

highjeans:
# Description

This PR Fixes rust-lang/libc#2903, however I am not sure if the change to `build.rs` is the right way to do it. MacOS has the `_opaque_pthread_t` type, however in `build.rs` it gets classified as a `type` rather than a struct, resulting in the test case for it using `_opaque_pthread_t` rather than `struct _opaque_pthread_t`. Please let me know if there is a better way to test `_opaque_pthread_t` without having to modify the build.rs file.

# Sources

https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/EXTERNAL_HEADERS/sys/_pthread/_pthread_types.h#L103
https://github.com/apple-oss-distributions/libpthread/blob/main/include/sys/_pthread/_pthread_t.h#L31

# Checklist

### DIFF
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -5,7 +5,6 @@ pub type useconds_t = u32;
 pub type blkcnt_t = i64;
 pub type socklen_t = u32;
 pub type sa_family_t = u8;
-// pub type pthread_t = crate::uintptr_t;
 
 cfg_if! {
     if #[cfg(any(
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -5,7 +5,36 @@ pub type useconds_t = u32;
 pub type blkcnt_t = i64;
 pub type socklen_t = u32;
 pub type sa_family_t = u8;
-pub type pthread_t = crate::uintptr_t;
+// pub type pthread_t = crate::uintptr_t;
+
+cfg_if! {
+    if #[cfg(any(
+            target_os = "macos",
+            target_os = "ios",
+            target_os = "tvos",
+            target_os = "watchos",
+            target_os = "visionos",
+        ))] {
+        s! {
+            pub struct __darwin_pthread_handler_rec {
+                __routine: Option<unsafe extern "C" fn(*mut c_void)>,
+                __arg: *mut c_void,
+                __next: *mut __darwin_pthread_handler_rec,
+            }
+            pub struct _opaque_pthread_t {
+                __sig: c_long,
+                __cleanup_stack: *mut __darwin_pthread_handler_rec,
+                __opaque: [c_char; __PTHREAD_SIZE__],
+            }
+        }
+        pub type __darwin_pthread_t = *mut _opaque_pthread_t;
+        pub type pthread_t = __darwin_pthread_t;
+    }
+    else {
+        pub type pthread_t = crate::uintptr_t;
+    }
+}
+
 pub type nfds_t = c_uint;
 pub type regoff_t = off_t;
 
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -9,12 +9,14 @@ pub type sa_family_t = u8;
 
 cfg_if! {
     if #[cfg(any(
-            target_os = "macos",
-            target_os = "ios",
-            target_os = "tvos",
-            target_os = "watchos",
-            target_os = "visionos",
-        ))] {
+        target_os = "macos",
+        target_os = "ios",
+        target_os = "tvos",
+        target_os = "watchos",
+        target_os = "visionos",
+    ))] {
+        pub type __darwin_pthread_t = *mut _opaque_pthread_t;
+        pub type pthread_t = __darwin_pthread_t;
         s! {
             pub struct __darwin_pthread_handler_rec {
                 __routine: Option<unsafe extern "C" fn(*mut c_void)>,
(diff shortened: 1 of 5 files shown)


src/unix/bsd/mod.rs
```diff
@@ -5,7 +5,34 @@ pub type useconds_t = u32;
...
-pub type pthread_t = crate::uintptr_t;
+
+cfg_if! {
+    if #[cfg(any(
+        target_os = "macos",
+        target_os = "ios",
+        target_os = "tvos",
+        target_os = "watchos",
+        target_os = "visionos",
+    ))] {
```
tgross35, line 16: This can be simplified to `cfg(target_vendor = "apple")`

src/unix/bsd/mod.rs
```diff
@@ -5,7 +5,34 @@ pub type useconds_t = u32;
...
+                __next: *mut __darwin_pthread_handler_rec,
+            }
+            pub struct _opaque_pthread_t {
+                __sig: c_long,
+                __cleanup_stack: *mut __darwin_pthread_handler_rec,
+                __opaque: [c_char; __PTHREAD_SIZE__],
+            }
+        }
+    } else {
+        pub type pthread_t = crate::uintptr_t;
```
tgross35, line 32: Could you add a comment `// FIXME(1.0): verify other BSDs`? Glancing around, other BSDs also look like a pointer https://github.com/freebsd/freebsd-src/blob/5bffa1d2069a05c8346eb34e17a39085fe0bf09b/include/signal.h#L68 https://github.com/openbsd/src/blob/ac71a6695016645a6726c964da2a0f509d63c2c8/include/pthread.h#L112

tgross35 (changes requested):
Sorry I never looked at this! Looks pretty good to me, just two small requests.

src/unix/bsd/mod.rs
```diff
@@ -5,7 +5,34 @@ pub type useconds_t = u32;
...
+            pub struct __darwin_pthread_handler_rec {
+                __routine: Option<unsafe extern "C" fn(*mut c_void)>,
+                __arg: *mut c_void,
+                __next: *mut __darwin_pthread_handler_rec,
+            }
+            pub struct _opaque_pthread_t {
+                __sig: c_long,
+                __cleanup_stack: *mut __darwin_pthread_handler_rec,
+                __opaque: [c_char; __PTHREAD_SIZE__],
+            }
```
tgross35, line 29: One other thing, could you mark these `#[non_exhaustive]`? Doesn't really matter here since they are all private fields, but we are starting to do that for everything new

src/unix/bsd/mod.rs
```diff
@@ -5,7 +5,34 @@ pub type useconds_t = u32;
...
+        }
+    } else {
+        pub type pthread_t = crate::uintptr_t;
+    }
+}
```
tgross35, line 34: Circling back, I think this could be simpler. Could you: instead update the definition to this?

```rust
extern_ty! {
    enum _opaque_pthread_t {}
}

type __darwin_pthread_t = *mut _opaque_pthread_t;
pub type pthread_t = __darwin_pthread_t;
```

We don't actually need the internal details here.

tgross35 (changes requested):

PR #4338, opened 2025-03-18
android: Add getauxval for 32-bit targets

taiki-e:
# Description

[`getauxval` has been added in Android 4.3 (API level 18)](https://github.com/aosp-mirror/platform_bionic/blob/d3ebc2f7c49a9893b114124d4a6b315f3a328764/libc/include/sys/auxv.h#L49). ([Since Rust 1.68, std requires API level 19+](https://blog.rust-lang.org/2023/01/09/android-ndk-update-r25.html))

(On Android, getauxval is currently only provided for 64-bit targets https://github.com/rust-lang/libc/pull/1991)

# Sources

# Checklist

### DIFF
--- a/libc-test/semver/android.txt
+++ b/libc-test/semver/android.txt
@@ -3329,6 +3329,7 @@ fwrite_unlocked
 gai_strerror
 genlmsghdr
 getaddrinfo
+getauxval
 getchar
 getchar_unlocked
 getcwd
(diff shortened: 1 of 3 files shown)


maurer (approved):
LGTM for Android

tgross35 (approved):

PR #4339, opened 2025-03-18
Remove core ffi primitive redefinitions

### LINKED ISSUE #4257: Remove the `libc` version of `core::ffi` primitives
Now that these exist in `core`, there isn't any reason to duplicate the definitions here.

highjeans:
# Description

This PR removed the duplicate definitions of the c primitive types and re-exports the definitions from rust's `core::ffi` module.
Closes #4257

# Sources

No API changes as the definitions of c primitive types are now coming from rust's `core::ffi` module.

# Checklist

### DIFF
--- a/src/primitives.rs
+++ b/src/primitives.rs
@@ -6,70 +6,70 @@
 //!
 //! The fixed-width integer aliases are deprecated: use the Rust types instead.
 
-pub type c_schar = i8;
-pub type c_uchar = u8;
-pub type c_short = i16;
-pub type c_ushort = u16;
-
-pub type c_longlong = i64;
-pub type c_ulonglong = u64;
-
-pub type c_float = f32;
-pub type c_double = f64;
-
-cfg_if! {
-    if #[cfg(all(
-        not(windows),
-        // FIXME(ctest): just use `target_vendor` = "apple"` once `ctest` supports it
-        not(any(
-            target_os = "macos",
-            target_os = "ios",
-            target_os = "tvos",
-            target_os = "watchos",
-            target_os = "visionos",
-        )),
-        not(target_os = "vita"),
-        any(
-            target_arch = "aarch64",
-            target_arch = "arm",
-            target_arch = "csky",
-            target_arch = "hexagon",
-            target_arch = "msp430",
-            target_arch = "powerpc",
-            target_arch = "powerpc64",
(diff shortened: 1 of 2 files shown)


highjeans:
Seems like the failure is related to [#94501](https://github.com/rust-lang/rust/issues/94501). I will look into and see if I can fix the issue.

highjeans:
Turns out `core::ffi:c_*` types were unstable in rust 1.63.0 and was only stabilized in 1.64.0. @tgross35 Would you have any ideas on what to do about this?

tgross35:
This will have to hold off until we do our next MSRV bump, which will likely be later this year. Technically the `main` branch doesn't need to test down to our current MSRV since 1.0 will require a higher version, but I also don't want to change this CI job in the interest of keeping our branches as similar as possible (for my sanity doing backports).

We do want this so feel free to get it ready, but for now:

@rustbot blocked

src/primitives.rs
```diff
@@ -186,3 +186,10 @@ cfg_if! {
         pub type __uint128_t = u128;
     }
 }
+
+// Re-export rust core::ffi c perimitive types
+#[allow(unused_imports)]
+pub use core::ffi::{
+    c_char, c_double, c_float, c_int, c_long, c_longlong, c_schar, c_short, c_uchar, c_uint,
+    c_ulong, c_ulonglong, c_ushort,
+};
```
tgross35, line 195: I think we could actually go a step further here; we won't want to publicly reexport these types at all anymore since users have access to `core::ffi` directly, so this block can be removed. Instead, https://github.com/rust-lang/libc/blob/5d549b27b1b1b005903210be00af11da91990aa4/src/macros.rs#L83-L86 can be adjusted to reexport `core::ffi` for our internal use.

highjeans: Ah yes this is a good point! I will add this right now.

highjeans: Turns out doing this will break the libc tests as those rely of the re-exports through libc rather than directly from core::ffi. Should I see if I can get the tests to use core::ffi instead of libc for these types?

tgross35: That seems reasonable to me 👍. It could even be a separate PR to merge sooner since the rest of this is blocked

tgross35:
Thank you for the work here!

PR #4341, opened 2025-03-19
Use unsafe callbacks

### LINKED ISSUE #2198: libc::clone() should take an unsafe callback
On Linux, [clone](https://docs.rs/libc/0.2.95/libc/fn.clone.html) takes a callback function pointer. The function pointer type is not marked `unsafe`, but there is basically no way to implement this callback with a safe function since it takes its context parameter by pointer.

It seems to me that `unsafe` should be added to the function pointer type, e.g.:

<pre>
    pub fn clone(
        cb: <b>unsafe</b> extern "C" fn(*mut ::c_void) -> ::c_int,
        child_stack: *mut ::c_void,
        flags: ::c_int,
        arg: *mut ::c_void,
        ...
    ) -&gt; ::c_int;
</pre>

I'm not certain, but I don't think this would be a breaking change.

highjeans:
# Description

This PR changes the callback function type to be unsafe rather than safe in (i believe) all functions that need a callback function, as per the discussion in #2198.

Closes #2198 

# Sources

As no APIs were changes (other than making the callback type unsafe), and safe vs unsafe functions are not documented in the linux documentation/header files, there are no sources.

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -4064,8 +4064,7 @@ fn test_linux(target: &str) {
             "epoll_params" => true,
 
             // FIXME(linux): Requires >= 6.12 kernel headers.
-            "dmabuf_cmsg" |
-            "dmabuf_token" => true,
+            "dmabuf_cmsg" | "dmabuf_token" => true,
 
             _ => false,
         }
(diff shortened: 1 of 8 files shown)


tgross35:
These look good to me, but there seem to be a few more cases. Would you mind getting those as well? https://github.com/search?q=repo%3Arust-lang%2Flibc%20%2F%3A.*fn%5C(%2F&type=code for a rough search, e.g. the aix module, solaris, espidf, and a few others.

We can't backport this since it is breaking unfortunately, but it will be great to have for 1.0:

@rustbot label +stable-declined

tgross35:
@highjeans gentle ping, would you be able to take a look at the above suggestion?

Issue #4340, opened 2025-03-18
Deprecate `__int128`, `__uint128`, `__int128_t`, `__uint128_t`


tgross35:
These types are only available on aarch64, but similar to `int8_t`, `int16_t`, etc. they don't have much purpose. We should deprecate them and remove in 1.0, replacing the single in-crate use with `i128`.

sanstzu:
Hi @tgross35, I would like to attempt to do this issue.

tgross35:
Assigned you, thanks! It should be pretty easy.

chengr4:
I think this issue can be closed?

tgross35:
Indeed, thanks!

PR #4343, opened 2025-03-19
Deprecate `__int128`, `__uint128`, `__int128_t`, `__uint128`

sanstzu:
# Description

Deprecates `__int128`, `__uint128`, `__int128_t`, `__uint128_t` and replace them with `u128`/`i128` types, as per (#4340)

# Sources

# Checklist


src/primitives.rs
```diff
@@ -87,102 +87,3 @@ pub type uint16_t = u16;
...
-        // progress on platforms where this type is important.
-        //
-        // The list of supported architectures and OSes is intentionally very restricted,
-        // as careful work needs to be done to verify that a particular platform
-        // has a conformant ABI.
-        //
-        // [0]: https://github.com/rust-lang/rust/issues/54341
-
-        /// C `__int128` (a GCC extension that's part of many ABIs)
-        pub type __int128 = i128;
```
tgross35, line 132: This PR seems to remove the types, but we don't want to do that at this time. Instead, add a `#[deprecated(...)]` attribute to each type alias, like the `int*_t` types higher up in this file.

sanstzu: Fixed it!

tgross35 (changes requested):
Could you set the version to 0.2.172? We can backport the deprecation, 1.0 will just have the removal. You can also add back the changes to use `i128`/`u128` internally, that is a non-breaking change.

Those two things and a squash, then lgtm!

sanstzu:
> Could you set the version to 0.2.172? We can backport the deprecation, 1.0 will just have the removal. You can also add back the changes to use `i128`/`u128` internally, that is a non-breaking change.
> 
> Those two things and a squash, then lgtm!
> 
> @rustbot label +stable-nominated

Thanks for the review Trevor! Squashed my commits already.

tgross35:
Okay I might have an idea here, could you play around with setting a `.typ_name` override for the types that are failing? Something like https://github.com/rust-lang/libc/blob/8c10293dc971b97fbe8559c4cc73da23a9d0bcf9/libc-test/build.rs#L706-L725, the config for Linux and Android will need to be updated.

sanstzu:
> Okay I might have an idea here, could you play around with setting a `.typ_name` override for the types that are failing? Something like
> 
> https://github.com/rust-lang/libc/blob/8c10293dc971b97fbe8559c4cc73da23a9d0bcf9/libc-test/build.rs#L706-L725
> 
> , the config for Linux and Android will need to be updated.



> Okay I might have an idea here, could you play around with setting a `.typ_name` override for the types that are failing? Something like
> 
> https://github.com/rust-lang/libc/blob/8c10293dc971b97fbe8559c4cc73da23a9d0bcf9/libc-test/build.rs#L706-L725
> 
> , the config for Linux and Android will need to be updated.

Sorry for the slow reply, wasn't on GitHub for the past few days and will work on it soon!

tgross35:
I've commandeered this patch since we'd like to have it done (you are still listed as an author). Thank you for getting this started!

PR #4344, opened 2025-03-20
hermit: add/fix `AF_*` values

mkroening:
# Description

This PR

- adds `AF_UNSPEC`,
- adds `AF_VSOCK`, and
- makes `AF_INET = 3`

on Hermit.

# Sources

See [hermit-os/kernel/src/syscalls/socket.rs#L27-L31](https://github.com/hermit-os/kernel/blob/1ed6594b342336db2b5bb87c363361f2df11be64/src/syscalls/socket.rs#L27-L31) for reference and https://github.com/hermit-os/kernel/pull/1640 for details on the `AF_INET` change (POSIX requires `AF_UNSPEC = 0` and distinct values for `AF_*`).

# Checklist


tgross35 (approved):
Thanks!

PR #4346, opened 2025-03-20
musl: enable `getrandom` on all musl platforms

tamird:
# Description

The existing bindings were added in #1399 and limited to targets where
rustc used musl version >= 1.1.20 which was not all musl targets at that
time. Since https://github.com/rust-lang/rust/pull/107129 all musl
targets use musl 1.2.3. Hence, move the binding to the module root so it
is available for all musl targets.

# Checklist


tgross35 (approved):

PR #4348, opened 2025-03-21
temporarily define O_DIRECT and SIGINFO for Solaris

psumbera:
This is temporary fix before new `Nix` crate version is released. This version will have following commit:
https://github.com/nix-rust/nix/commit/0f4559386404f58803b27807201b4dd6d1411f50

The fix is waiting there for new version for several months now. And meanwhile build of `Rust` on `Solaris` requires patching.

This should allow proceed with following commit to `Rust`:
https://github.com/rust-lang/rust/pull/138699

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -998,6 +998,13 @@ fn test_solarish(target: &str) {
     cfg.define("__EXTENSIONS__", None);
     cfg.define("_LCONV_C99", None);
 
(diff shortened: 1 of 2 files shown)


tgross35:
Would you mind marking the comments `FIXME(nix)` or `FIXME(solaris)` or something so we remember to clean them up at some point? Otherwise LGTM.

tgross35 (approved):
Thanks!

PR #4355, opened 2025-03-27
Add HelenOS support

mvolfik:
# Description

Add support for [HelenOS](https://www.helenos.org). I would like to get this backported to 0.2, because I also have a stdlib support ready, which depends on libc 0.2. I'll be sending that patch to rust-lang/rust rather soon (this or next week), I hope.

# Sources

All HelenOS source code is at https://github.com/HelenOS/helenos/. There are comments in this patch attributing each section to a specific header file. Files from `common/` and `abi/` get included in HelenOS libc by the build system there, from `uspace/lib` we use `c`, `inet` and `posix`, so that's what we link. `libstartfiles` is HelenOS' version of `crt*` files, because they don't distribute them along with the compiler toolchain, but instead as a library.

# Checklist

_Tested locally as `cargo test` (without the `--target` flag), testing for HelenOS targets will require support [here in ctest2](https://github.com/JohnTitor/ctest2/blob/05634716fbe6f9ea4bf090ce85776269740b1a58/src/lib.rs#L1091), I hope for a tier3 target, it is ok to go without these tests for now?_


mvolfik:
Ugh, the only error is missing HelenOS target... I planned to merge the compiler and stdlib support at once. But this makes it seem like I will need to first merge no-std support for HelenOS in compiler, then this libc patch can go through, and only then the stdlib PR makes sense?

tgross35:
Is there a proposal for adding the target to Rust? The changes look pretty reasonable to me, but getting it at least okayed for rustc is a prerequisite for `libc` changes.

> Ugh, the only error is missing HelenOS target... I planned to merge the compiler and stdlib support at once. But this makes it seem like I will need to first merge no-std support for HelenOS in compiler, then this libc patch can go through, and only then the stdlib PR makes sense?

It is usually preferable to do this anyway; the compiler changes are usually pretty easy and can get merged as soon as the proposal is accepted. Then library changes can build up from there and things don't get slowed down by having a large diff to review at once.

tgross35:
Per the above

@rustbot blocked

tgross35:
Compiler support merged, so

@rustbot author

Note that we now also have the `src/new` module which is meant to mirror the header structure that a user can `#include`. If you're up for moving the HelenOS-specific bits there then that would be ideal (though not strictly necessary). The POSIX APIs could stay since they are going to get commonized separately.

src/helenos.rs
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    // uspace/lib/inet/include/inet/addr.h
+    #[repr(u32)]
+    pub enum ip_ver_t {
+        ip_any,
+        ip_v4,
+        ip_v6,
+    }
+}
```
tgross35: I know this is a WIP but as a driveby comment, use `c_enum!` rather than `e!` which we're getting rid of (closer matches C enums by allowing you to have extra variants without causing problems)

mvolfik: yep, I eventually figured that out from broken tests :) thanks

tgross35:
> hm, seems like musl website is down, which causes the CI to fail :(

Coincidentally I was working on getting a mirror set up, but currently can't because I can't download the source :)

> in the meantime, @tgross35 - I could rewrite this for the new structure, but I'm not exactly sure how - most of the systems have the `new` folders still empty, so I can't really use them as example.
>
> Also, I'm not adding only HelenOS `libc` - some of the functionality required for standard library interfaces is also in `libinet` and `libposix`, + some definitions used in libc come from a bit weird locations in the HelenOS source tree, `abi/include` and `common/include`.1 Do you have any advice how this should be reflected for this project? (Or is it even right to include these other, albeit essential, libraries in the libc crate?)

First off it's pretty new, so if anything feels too weird then feel free to leave it as-is and I'll clean it up when things are more established.

For structure the main thing is we'd like to make things a bit like what is `#include`able. So if the user can `#include <sys/socket>` then there should be a `sys::socket` module, and we reexport it here https://github.com/rust-lang/libc/blob/83bd32f8ac10bf20bba9509b06de5f242ca7c6ad/src/new/mod.rs#L172-L204. 

The secondary thing is to match the source structure. And then reexport that in a way that matches the `#include`s, kind of echoing what the build system is doing. I can explain a bit more but I did that with glibc https://github.com/rust-lang/libc/blob/83bd32f8ac10bf20bba9509b06de5f242ca7c6ad/src/new/glibc/mod.rs, the nonpublic modules are source directories that don't exist in the final build (not available to `#include` from). Apple is another example, there are a few different repos that it pulls from.

Again, don't worry if this doesn't really work out - it's all pretty new :)
 
I can look a little closer at the specific API later, but libposix seems fine. However, is libinet needed for networking? (I'm not familiar with it)

mvolfik:
hmm, I hoped I would test this restructured patch with my rust stdlib patch, but I realized this isn't the 0.2 branch supported by libc.. and backporting this for 0.2 seems like quite a rewrite. What is the best path forward? Maybe make libstd depend on libc github master branch (for HelenOS only, obviously)?

and musl is down again :( it worked for me this morning >:(

mvolfik:
hm, almost won the CI roulette this time. Is the segfault on freebsd something caused by me?

mvolfik:
> Backporting shouldn't be a problem, is your libc-0.2 branch up to date

ah, my `upstream/libc-0.2` was up to date, but I was cherrypicking over the local `libc-0.2` :D thanks for the nudge

all tested and should be ready for merge from my side. Could you please point me where in the stdlib I should add the `#[link()]` attributes to produce correct builds? Thanks for all the assistance!

src/new/helenos/fibril.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+}
+
+extern_ty! {
+    pub enum fibril_t {}
+}
+
+extern "C" {
+    pub fn fibril_create_generic(
+        func: extern "C" fn(*mut c_void) -> errno_t,
```
tgross35: Callbacks should take an `unsafe extern "C" fn`

src/new/helenos/dirent_mod.rs
```diff
@@ -0,0 +1,29 @@
...
+extern_ty! {
+    pub enum DIR {}
+}
+
+extern "C" {
+    pub fn opendir(name: *const c_char) -> *mut DIR;
+    pub fn readdir(dir: *mut DIR) -> *mut dirent;
+    pub fn closedir(dir: *mut DIR) -> c_int;
+    pub fn rewinddir(dir: *mut DIR);
+
```
tgross35: Nit, extra space

src/new/helenos/fibril_synch.rs
```diff
@@ -0,0 +1,49 @@
...
+    pub struct fibril_mutex_t {
+        pub oi: fibril_owner_info_t,
+        pub counter: c_int,
+        pub waiters: list_t,
+    }
+
+    pub struct fibril_condvar_t {
+        pub waiters: list_t,
+    }
+}
```
tgross35, line 30: Do you also want rwlock for use in `RwLock`? Noblocking of course, I'm not sure whether you plan to use these APIs or the fallback in std anyway.

src/new/helenos/fibril.rs
```diff
@@ -0,0 +1,36 @@
+//! HelenOS fibrils API
+//!
+//! * Header file: <https://github.com/HelenOS/helenos/tree/master/uspace/lib/c/include/fibril.h>
+
+pub use crate::time::*;
+use crate::{
+    c_void,
+    errno_t,
+    size_t,
+};
```
tgross35: `use crate::prelude::*` is a bit preferred to import the ints since most refactoring assumes they are in scope, and it brings a few other useful things.

(Applies to a few of these files)

src/new/helenos/fibril_synch.rs
```diff
@@ -0,0 +1,49 @@
...
+
+f! {
+    pub fn fibril_mutex_initialize(fm: *mut fibril_mutex_t) -> () {
+        let fm = &mut *fm;
+        fm.oi.owned_by = core::ptr::null_mut();
```
tgross35: Nit: If you use the prelude, `core::ptr` will be in scope

src/new/helenos/fibril_synch.rs
```diff
@@ -0,0 +1,49 @@
...
+        fm.counter = 1;
+        list_initialize(&mut fm.waiters);
+    }
```
tgross35: https://github.com/HelenOS/helenos/blob/144fafda7cb6d43dfda1c93b8e46b8d51635e0c2/uspace/lib/c/include/fibril_synch.h#L55-L66 seems to effectively return a definition. Any reason not to do similar here and directly return a `fibril_mutex_t`, rather than using a pointer?

You may want to uppercase to match the definitions too.

src/new/helenos/fibril_synch.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    pub struct fibril_condvar_t {
+        pub waiters: list_t,
+    }
+}
+
+f! {
+    pub fn fibril_mutex_initialize(fm: *mut fibril_mutex_t) -> () {
+        let fm = &mut *fm;
```
tgross35: Maybe take a `&mut fibril_mutex_t` as the arg and make the function safe? Allows the caller to avoid a bit of `unsafe` if they have a known good reference.

src/new/helenos/fibril_synch.rs
tgross35: Ah, I now see it has a self-referential list. I guess this is about as good as it gets, unfortunately.

src/new/helenos/inet/tcp.rs
```diff
@@ -0,0 +1,76 @@
...
+s_no_extra_traits! {
+    pub struct tcp_cb_t {
+        pub connected: extern "C" fn(conn: *mut tcp_conn_t),
+        pub conn_failed: extern "C" fn(conn: *mut tcp_conn_t),
+        pub conn_reset: extern "C" fn(conn: *mut tcp_conn_t),
+        pub data_avail: extern "C" fn(conn: *mut tcp_conn_t),
+        pub urg_data: extern "C" fn(conn: *mut tcp_conn_t),
+    }
+    pub struct tcp_listen_cb_t {
+        pub new_conn: extern "C" fn(listener: *mut tcp_listener_t, conn: *mut tcp_conn_t),
```
tgross35: These callbacks should be `unsafe`

tgross35 (changes requested):
This looks _awesome_, thank you making everything so well organized and providing links. I just have a few small requests.

src/new/helenos/inet/tcp.rs
tgross35: Noticing now that they should also be `Option<...>`

tgross35:
Just another friendly ping here @mvolfik; no pressure but this is 98% of the way done and I'd love to get it closed out :) if you're still busy, perhaps another HelenOS contributor could pick up the patches?

src/new/helenos/fibril_synch.rs
mvolfik: i actually had no rwlock support for HelenOS (because it wasn't really needed for the subset of stdlib I wanted to get working), but good point, added it.

mvolfik:
@tgross35 thanks for the nudge.

I rebased it, and pushed an LLM to do the mechanical chances you pointed out. It also made updates to reflect the updated conventions and macro usages - `enum FILE {}` -> `type FILE;` and similar. I reviewed the diff and all seems fine, but pls check that it really matches the libc conventions.

Also added the rwlock API, as I commented above.

Tomorrow, I hope to rebase my pending stdlib changes (which is also not going to be fun, since they're doing some restructuring in the pal/sys internals (: ), and test that everything still works over a year later, to give a green-light on merging this. Again, thanks for the assistance.

tgross35 (approved):
Thanks for the update, LGTM! Could you please clean up commit history? (most rust-lang repos use rebase or merge rather than squash, so they actually make it to the repo's history)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/4355/1a8e71f33b1d6ea1e072210e7fc994417bbb2e34..f83186fb3c887bfc168c5823ac97132d7fa81473)*

mvolfik:
ah, I didn't realize that (we do squash merges at work, so I forgot this is a thing). rebased & squashed (I forgot one commit in the original force-push, here's the correct [range diff](https://triagebot.infra.rust-lang.org/gh-range-diff/rust-lang/libc/1a8e71f33b1d6ea1e072210e7fc994417bbb2e34..f83186fb3c887bfc168c5823ac97132d7fa81473/2eda1e52fe2efdf93c1e5c65e346343a530d8bc3..1ed7f095c9d0daf3db6dea8e2a5ca468f53d4296))

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/4355/2eda1e52fe2efdf93c1e5c65e346343a530d8bc3..1ed7f095c9d0daf3db6dea8e2a5ca468f53d4296)*

tgross35:
> ah, I didn't realize that (we do squash merges at work, so I forgot this is a thing)

If GH let you pick then I'd have squashed it here as well. But we need the option to preserve history and unfortunately if merge queues are enabled, it doesn't let you pick a different strategy per PR :(

mvolfik:
:tada: thank you!

@tgross35 what do I need to do to get this backported to 0.2, so I could carry on with the stdlib PR?

tgross35:
No action needed, I do backports in batches right before a release. Probably in about a week or so until then

PR #4359, opened 2025-04-01
link README.md

Hack666r:
fix old link to a new one


tgross35:
LGTM, but could you please update the commit message to be more descriptive? Something like "Replace obsolete docs.rs badges with shields.io".

tgross35:
User's history looks inauthentic and the last comment makes no sense. Seems to be a bot PR.

PR #4361, opened 2025-04-02
Merge the `ctest2` repository into `libc`

tgross35:
`libc`'s testing is pretty tightly coupled to [`ctest2`](https://crates.io/crates/ctest2), which is a reasonably simple crate that maps Rust syntax to C syntax and creates tests based off of that. `ctest2` itself is a fork of [`ctest`](https://crates.io/crates/ctest).

Unfortunately, relying on `ctest(2)` has been somewhat of a problem since it supports only a very outdated version of Rust's syntax, and there are a handful of other bugs. There have been some GSoC proposals to modernize `libc` and its test crates; moving `ctest` to a rust-lang repo makes that easier and should come with some maintenance benefits.

On Zulip, Alex confirmed ownership of the original `ctest` crate could be transferred to rust-lang so we can continue publishing under that name, for other repos that make use of `ctest`/`ctest2`.

This pull only brings over the repo source and history as it currently exists. I will need to follow up with integration of the CI and release workflows.


tgross35:
Recreated from https://github.com/rust-lang/libc/pull/4361 which GH closed on me

PR #4377, opened 2025-04-03
Fix makedev function signature on Apple targets

nohajc:
# Description

Fix the makedev function signature according to actual implementation:
https://github.com/apple-oss-distributions/xnu/blob/main/bsd/sys/types.h#L154
https://developer.apple.com/documentation/kernel/1547212-makedev

# Checklist


tgross35:
This lgtm, but any reason not to also fix `major` and `minor`?

nohajc:
@tgross35 Oh, right. They're not even using `dev_t` there. Let me fix it.

nohajc:
Well, I just noticed the functions are only compiled for C++. C still uses those macros which are more loosely defined. Not sure that's reason enough to deviate from the type signatures even if they're questionable.

tgross35 (changes requested):
Going through old PRs, sorry I missed this one. I think this change is worth making even if they use untyped macros, given this makes Apple match all the other platforms. We still can't backport unfortunately due to breakage, but this would be good to have for 1.0.

Would you be able to rebase?

src/unix/bsd/apple/mod.rs
```diff
@@ -3,10 +3,7 @@
 //! This covers *-apple-* triples currently
 
 use crate::prelude::*;
-use crate::{
-    cmsghdr,
-    off_t,
-};
+use crate::{cmsghdr, off_t};
```
tgross35: Looks like this may have accidentally snuck in

nohajc: Sorry, auto-formatting.

PR #4379, opened 2025-04-03
if_tun.h ioctls for android

jjanowsk:
# Description

Add missing constants from linux/if_tun.h header on android platform. Mainly ioctl operation codes

# Sources

https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/kernel/uapi/linux/if_tun.h;drc=180edefbd287c39caeb9d48784a9a10ac35f3636

# Checklist


src/unix/linux_like/android/b32/mod.rs
```diff
@@ -232,6 +232,12 @@ pub const UT_HOSTSIZE: usize = 16;
 pub const SIGSTKSZ: size_t = 8192;
 pub const MINSIGSTKSZ: size_t = 2048;
 
+// linux/if_tun.h
+// Ioctl operation codes
+pub const TUNATTACHFILTER: c_int = 0x400854d5;
```
maurer, line 237: These should be declared via `_IOW`/`_IOR`, as they are in the [source of truth](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/kernel/uapi/linux/if_tun.h;l=29?q=if_tun.h).

We have this functionality [in libc already](https://github.com/rust-lang/libc/blob/30f03b2a990dd529cf4b2c433c738f4a8e366417/src/unix/linux_like/linux/mod.rs#L6023), so we should use it rather than pre-expanding things like this.

src/unix/linux_like/android/b64/mod.rs
```diff
@@ -288,6 +288,12 @@ pub const UT_LINESIZE: usize = 32;
 pub const UT_NAMESIZE: usize = 32;
 pub const UT_HOSTSIZE: usize = 256;
 
+// linux/if_tun.h
+// Ioctl operation codes
+pub const TUNATTACHFILTER: c_int = 0x401054d5;
```
maurer, line 293: As above, IOW/R

src/unix/linux_like/android/mod.rs
```diff
@@ -2646,6 +2647,36 @@ pub const TUN_F_TSO_ECN: c_uint = 0x08;
 pub const TUN_F_UFO: c_uint = 0x10;
 pub const TUN_F_USO4: c_uint = 0x20;
 pub const TUN_F_USO6: c_uint = 0x40;
+// Protocol info prepended to the packets (when IFF_NO_PI is not set)
```
maurer, line 2650: IOW/R

src/unix/linux_like/android/mod.rs
```diff
@@ -2623,7 +2623,7 @@ pub const SND_CNT: usize = SND_MAX as usize + 1;
 pub const UINPUT_VERSION: c_uint = 5;
 pub const UINPUT_MAX_NAME_SIZE: usize = 80;
 
-// bionic/libc/kernel/uapi/linux/if_tun.h
+// linux/if_tun.h
```
maurer, line 2626: This should ideally remain sourced from bionic, as these are libc bindings, not kernel bindings.

maurer (commented):
Main thing is we should probably use IOW/R to mimic what's in the bionic header. This will help us make sure things are synced up + structs people expect are available.

src/unix/linux_like/android/mod.rs
jjanowsk: Reverted

src/unix/linux_like/android/b32/mod.rs
jjanowsk: Thanks, I was not aware we already had this functionality. I've updated the PR.

jjanowsk:
@maurer Updated. Please re-review

maurer (approved):

src/unix/linux_like/linux/mod.rs
```diff
@@ -5940,88 +5940,28 @@ pub const SCHED_FLAG_ALL: c_int = SCHED_FLAG_RESET_ON_FORK
...
-        | ((size as u32) << _IOC_SIZESHIFT)
-}
-
-/// Build an ioctl number for an argumentless ioctl.
-pub const fn _IO(ty: u32, nr: u32) -> u32 {
-    _IOC(_IOC_NONE, ty, nr, 0)
+ /// Build an ioctl number for an argumentless ioctl.
+ pub const fn _IO(ty: u32, nr: u32) -> u32 {
+    super::_IOC(super::_IOC_NONE, ty, nr, 0)
 }
```
tgross35, line 5950: Could these just move to `src/unix/linux_like/mod.rs` next to `_IOC`?

jjanowsk: They are not strictly duplicates. They differ when it comes to the return type. Of course I could put them in some non public module (e.g. `mod internal`) inside of `linux_like/mod.rs` and then reexport it in e.g. linux and wrap in android to cast the returned value. I thought about it but in the end I thought that that's just cleaner to duplicate those two small pieces. Let me know what you prefer. I can do it as I described here.

BTW using those `_IO*` macros are nice because the remove duplication when it comes IOCTL definitions on different architectures. I could use them for `if_tun.h` ioctl numbers in linux tree as well in separate PR if you're interested.

tgross35: I missed the return type difference, as-is is fine then 👍

Updating any hardcoded ioctl values would be appreciated if you are interested in doing that. A lot of these definitions come from before we had `const fn` available, it would be great to clean that up.

tgross35 (approved):
LGTM, thanks for the PR and thanks @maurer for reviewing

PR #4381, opened 2025-04-04
FreeBSD: Deprecate TCP_PCAP_OUT and TCP_PCAP_IN

tgross35:
FreeBSD removed these upstream in [1], so deprecate them here. This resolves a recent CI failure.

The constants were originally added in [2].

[1]: https://github.com/freebsd/freebsd-src/commit/6e76489098c6dc415ac3f2ae084154c3c22558ec
[2]: https://github.com/rust-lang/libc/pull/1151

Cc @asomers

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2783,6 +2783,9 @@ fn test_freebsd(target: &str) {
             // Added in FreeBSD 14.0
             "TCP_FUNCTION_ALIAS" if Some(14) > freebsd_ver => true,
 
+            // FIXME(freebsd): Removed in FreeBSD 15, deprecated in libc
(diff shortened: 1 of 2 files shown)


libc-test/build.rs
```diff
@@ -2783,6 +2783,9 @@ fn test_freebsd(target: &str) {
             // Added in FreeBSD 14.0
             "TCP_FUNCTION_ALIAS" if Some(14) > freebsd_ver => true,
 
+            // Removed in FreeBSD, deprecated in libc
```
asomers, line 2786: You might want to add the release number to this comment

asomers (approved):

libc-test/build.rs
tgross35: Good idea, added

PR #4382, opened 2025-04-06
Define more ioctl codes on riscv32gc-unknown-linux-gnu

sunfishcode:
# Description

Define ioctl codes including `FICLONE` and `FS_IOC32_GETVERSION` on riscv32gc-unknown-linux-gnu.

# Sources

The values match the values in the linux-raw-sys crate, which is autogenerated from Linux sources:

https://github.com/sunfishcode/linux-raw-sys/blob/21a3b3b1fada619179810ca92404690d45890f29/src/riscv32/ioctl.rs#L248

# Checklist

### DIFF
--- a/src/unix/linux_like/linux/arch/generic/mod.rs
+++ b/src/unix/linux_like/linux/arch/generic/mod.rs
@@ -1,5 +1,5 @@
 use crate::prelude::*;
-use crate::Ioctl;
+use crate::{Ioctl, _IOR, _IOW};
 
 s! {
     pub struct termios2 {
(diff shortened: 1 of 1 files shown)


kraj (approved):
lgtm

tgross35 (approved):
Any idea why these are gated in the first place? All of the constants are in https://github.com/torvalds/linux/blob/bec7dcbc242c6c087cede1a6fdfaeb5d6eaf25bf/include/uapi/linux/fs.h, I don't know why any platforms wouldn't have them unless they have no fs support.

In any case LGTM

tgross35:
> they all do exist but values are different sadly

Oh, from differences in sizes passed to `_IOR` and `_IOW`? We actually have these functions now, the `cfg` could probably just be deleted and replaced with calls https://github.com/rust-lang/libc/blob/30f03b2a990dd529cf4b2c433c738f4a8e366417/src/unix/linux_like/linux/mod.rs#L5991-L6025.

tgross35:
Actually @sunfishcode would you mind just changing the relevant blocks to make use of those instead? I think we may as well use this as an excuse to get rid of some pre-`const fn` cruft.

sunfishcode:
@tgross35 Done.

kraj (approved):
LGTM

tgross35 (approved):
Great cleanup, thanks!

PR #4385, opened 2025-04-09
Pin cc to `<1.2.18`

tgross35:
Work around a `cc` issue that doesn't have a fix released yet [1], which is causing CI failures for musl targets.

[1]: https://github.com/rust-lang/cc-rs/issues/1452


tgross35:
Reverted before it needed to merge to stable

PR #4389, opened 2025-04-10
Add MS_NOSYMFOLLOW flag

tedbrandston:
# Description

Adds the `MS_NOSYMFOLLOW` flag from linux/mount.h

MS_NOSYMFOLLOW ([since Linux 5.10](https://github.com/torvalds/linux/commit/dab741e0e02bd3c4f5e2e97be74b39df2523fc6e)) disallows automatic following of symlinks. 

# Sources

https://github.com/torvalds/linux/blob/ab59a8605604f71bbbc16077270dc3f39648b7fc/include/uapi/linux/mount.h

# Checklist

### DIFF
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -654,6 +654,7 @@ pub const MS_SYNCHRONOUS: c_ulong = 0x10;
 pub const MS_REMOUNT: c_ulong = 0x20;
 pub const MS_MANDLOCK: c_ulong = 0x40;
 pub const MS_DIRSYNC: c_ulong = 0x80;
+pub const MS_NOSYMFOLLOW: c_ulong = 0x100;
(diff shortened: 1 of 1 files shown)


tgross35 (changes requested):
The change looks fine but `semver/linux.txt` needs to be updated. Any reason this is marked as a draft?

tedbrandston:
The template said to mark it as a draft if you weren't sure about the checklist items. 

I'll take a look at updating the semver, but won't have a chance to test locally for a while. Is local testing necessary for a change like this?

tgross35:
It's usually good to run locally if possible but don't worry about it here, CI covers testing for this platform. Just update the semver file and squash and this should be good to go.

tedbrandston:
Looks like the freebsd failures are present on other recent PRs. I've updated linux.txt and run the tests locally (they pass).

PR #4400, opened 2025-04-12
Add constants for Memory-Deny-Write-Execute prctls

rusty-snake:
# Description

This PR adds constants for Memory-Deny-Write-Execute prctls added in Linux 6.3 (`PR_MDWE_NO_INHERIT` Linux 6.6).

# Sources

https://github.com/torvalds/linux/blob/3bde70a2c82712f05c7220b8b94fc2cbdf7fbfe0/include/uapi/linux/prctl.h#L287-L292

# Checklist

---


rusty-snake:
I'm not sure if the CI failure is because architecture or libc.

tgross35:
It's probably just that everything in CI running musl is very outdated, I assume they should be available since they are uapi. Cc @maurer in case that is not accurate.

Assuming it's just a CI thing, you can update libc-test/build.rs to skip these on the relevant platform.

@rustbot author

src/unix/linux_like/linux/mod.rs
```diff
@@ -3303,6 +3303,11 @@ pub const PR_GET_CHILD_SUBREAPER: c_int = 37;
 pub const PR_SET_NO_NEW_PRIVS: c_int = 38;
 pub const PR_GET_NO_NEW_PRIVS: c_int = 39;
 
+pub const PR_SET_MDWE: c_int = 65;
+pub const PR_GET_MDWE: c_int = 66;
+pub const PR_MDWE_REFUSE_EXEC_GAIN: c_uint = 1;
+pub const PR_MDWE_NO_INHERIT: c_uint = 2;
```
tgross35, line 3309: Tiny nit: could you make these `1 << 0` and `1 << 1` to match the headers?

tgross35:
I double checked and Android does look like it defines these https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/kernel/uapi/linux/prctl.h;l=175-178;drc=7ac54f5c391bf3aeb9fec84a158939421fd8f505

tgross35 (approved):
Thank you!

Issue #4297, opened 2025-03-04
Android CI jobs spurious failure


tgross35:
It seems like all of them fail occasionally but i686-linux-android is pretty bad, with an 11% failure rate per GH's metrics. Some sample messages:

```text
adb: error: failed to copy '/checkout/target/i686-linux-android/debug/deps/cmsg-930d5ec3aabcc80f' to '/data/local/tmp/cmsg-930d5ec3aabcc80f': remote secure_mkdirs failed: Read-only file system
```

```rust
thread 'main' panicked at /tmp/runtest.rs:49:9:
failed to find successful test run
stack backtrace:
status: exit status: 127
stdout ---

stderr ---
/system/bin/sh: /data/local/tmp/linux_fcntl-d0799fab8ac010c3: not found

   0: std::panicking::begin_panic
             at /rustc/e16a049adbf94d610787430b6efdf31d896dc5b6/library/std/src/panicking.rs:767:5
   1: runtest::main
   2: core::ops::function::FnOnce::call_once
```

From https://github.com/rust-lang/libc/actions/runs/13617856722/job/38063309102 and https://github.com/rust-lang/libc/actions/runs/13659560732/job/38187528080

tgross35:
Sample failure on aarch64 https://github.com/rust-lang/libc/actions/runs/13663057122/job/38198631635

PR #4401, opened 2025-04-12
Update statx constants and types (Linux 6.8-6.14)

rusty-snake:
# Description

This PR updates statx constants and types for features added in Linux 6.8 to 6.14.

# Sources

- https://github.com/torvalds/linux/blob/3bde70a2c82712f05c7220b8b94fc2cbdf7fbfe0/include/uapi/linux/stat.h#L168-L186
- https://github.com/torvalds/linux/blob/3bde70a2c82712f05c7220b8b94fc2cbdf7fbfe0/include/uapi/linux/stat.h#L213-L217
- https://github.com/torvalds/linux/blob/3bde70a2c82712f05c7220b8b94fc2cbdf7fbfe0/include/uapi/linux/stat.h#L253

# Checklist

---


rusty-snake:
No idea why CI is failing.

src/unix/linux_like/mod.rs
```diff
@@ -1617,6 +1622,10 @@ cfg_if! {
         pub const STATX_ALL: c_uint = 0x0fff;
         pub const STATX_MNT_ID: c_uint = 0x1000;
         pub const STATX_DIOALIGN: c_uint = 0x2000;
+        pub const STATX_MNT_ID_UNIQUE: c_uint = 0x4000;
+        pub const STATX_SUBVOL: c_uint = 0x8000;
```
tgross35, line 1626: Small nit, do you mind adding the leading `0x0000` here to keep these in sync with the upstream source?

tgross35: LGTM with that and a squash

tgross35 (approved):
LGTM as long as CI passes

tgross35 (changes requested):
Looks like some of this may not be available on arm32 Android, can you check this against the bionic headers? https://cs.android.com/android/platform/superproject/main, cc @maurer

src/unix/linux_like/mod.rs
```diff
@@ -1617,6 +1622,10 @@ cfg_if! {
         pub const STATX_ALL: c_uint = 0x0fff;
         pub const STATX_MNT_ID: c_uint = 0x1000;
         pub const STATX_DIOALIGN: c_uint = 0x2000;
+        pub const STATX_MNT_ID_UNIQUE: c_uint = 0x4000;
+        pub const STATX_SUBVOL: c_uint = 0x_0000_8000;
+        pub const STATX_WRITE_ATOMIC: c_uint = 0x_0001_0000;
+        pub const STATX_DIO_READ_ALIGN: c_uint = 0x_0002_0000;
```
maurer, line 1628: This constant is not yet present in bionic, likely because it was introduced to Linux in 6.14, and the latest Android kernel is 6.12.

maurer (commented):
tl;dr:
* Remove support for `statx_dio_read_align` on Android; it won't arrive for at least a year.
* Bump the NDK version to 28, and the rest of your stuff should work.

This is going to be compatible from the Android PoV - only `STATX_DIO_READ_ALIGN` is not yet present in bionic, and that's likely because the latest kernel version Android uses is 6.12 right now.

The reason you're getting an error from `offsetof` for `stx_subvol` is not because of the architecture, it's because NDK 27 (current LTS NDK) does not yet have that field. NDK 28 (stable and released, but not LTS) does.

Probably (subject to @tgross35 agreement) we should [switch to NDK 28b](https://github.com/rust-lang/libc/blob/main/ci/android-install-ndk.sh#L5) in this PR.

Support in bionic for `statx_dio_read_align` is probably more than a year out, because Android needs to release a kernel >= 6.14, and then needs to have an NDK release. I don't anticipate any differences between our support and anyone else's, but no Android device will exist that supports it for at least a year, so it'd be a mistake to try to use it. I'd suggest configuring out `statx_dio_read_align` both field and constant.

tgross35:
> Probably (subject to @tgross35 agreement) we should [switch to NDK 28b](https://github.com/rust-lang/libc/blob/main/ci/android-install-ndk.sh#L5) in this PR.

I have no objections to bumping the version. Is there any chance you, or anyone else from the android team / more familiar, might be able to do it in a separate PR? We have been having some pretty consistent intermittent failures (https://github.com/rust-lang/libc/issues/4297), it would be good for somebody more familiar with the platform to double check that the way we are setting up and testing actually makes sense.

libc-test/build.rs
```diff
@@ -4220,6 +4220,51 @@ fn test_linux(target: &str) {
...
+            | "STATX_WRITE_ATOMIC" // Added in Linux 6.11
+            | "STATX_ATTR_WRITE_ATOMIC" // Added in Linux 6.11
+            | "STATX_DIO_READ_ALIGN" // Added in Linux 6.14
+            => true,
+
+            // FIXME(linux): Parts of netfilter/nfnetlink*.h require more recent kernel headers:
+            | "RTNLGRP_MCTP_IFADDR" // linux v5.17+
+            | "RTNLGRP_TUNNEL" // linux v5.18+
+            | "RTNLGRP_STATS" // linux v5.18+
+                => true,
```
tgross35: Did this come through the rebase? Most of these constants are unrelated to what is added here

src/unix/linux_like/mod.rs
```diff
@@ -299,7 +299,12 @@ cfg_if! {
                 pub stx_mnt_id: crate::__u64,
                 pub stx_dio_mem_align: crate::__u32,
                 pub stx_dio_offset_align: crate::__u32,
-                __statx_pad3: Padding<[crate::__u64; 12]>,
+                pub stx_subvol: crate::__u64,
+                pub stx_atomic_write_unit_min: crate::__u32,
+                pub stx_atomic_write_unit_max: crate::__u32,
+                pub stx_atomic_write_segments_max: crate::__u32,
+                pub stx_dio_read_offset_align: crate::__u32,
+                __statx_pad3: [crate::__u64; 9],
```
tgross35: The last field should keep `Padding`

libc-test/build.rs
```diff
@@ -4384,6 +4384,31 @@ fn test_linux(target: &str) {
...
+            | "CANXL_MAX_MTU"
+            | "CANXL_MIN_DLC"
+            | "CANXL_MIN_DLEN"
+            | "CANXL_MIN_MTU"
+            | "CANXL_MTU"
+            | "CANXL_PRIO_BITS"
+            | "CANXL_PRIO_MASK"
+            | "CANXL_SEC"
+            | "CANXL_XLF"
+             => true,
```
tgross35: These don't seem to be touched here?

tgross35 (changes requested):

libc-test/semver/linux-gnu.txt
```diff
@@ -418,20 +418,25 @@ STATX_ATTR_IMMUTABLE
...
 STATX_MNT_ID
+STATX_MNT_ID_UNIQUE
 STATX_MODE
 STATX_MTIME
 STATX_NLINK
 STATX_SIZE
+STATX_SUBVOL
 STATX_TYPE
 STATX_UID
+STATX_WRITE_ATOMIC
```
tgross35, line 439: These are added to linux-gnu, but should be possible to have in linux-like

tgross35:
@rustbot author

CI still isn't passing here, you'll also need to skip the struct fields with `if gnu`. This is because the 32-bit glibc tests run with an older version docker image than the rest, since we (unfortunately) had to pin them.

libc-test/build.rs
```diff
@@ -4391,6 +4391,14 @@ fn test_linux(target: &str) {
             // Not present in glibc
             "PR_SME_VL_LEN_MAX" | "PR_SME_SET_VL_INHERIT" | "PR_SME_SET_VL_ONE_EXEC" if gnu => true,
 
+            "STATX_DIOALIGN" // Added in  linux 6.1
+            | "STATX_MNT_ID_UNIQUE" // Added in Linux 6.8
+            | "STATX_SUBVOL" // Added in Linux 6.10
+            | "STATX_WRITE_ATOMIC" // Added in Linux 6.11
+            | "STATX_ATTR_WRITE_ATOMIC" // Added in Linux 6.11
+            | "STATX_DIO_READ_ALIGN" // Added in Linux 6.14
+            => true,
```
tgross35, line 4400: Add `if gnu` and something like `// FIXME(ci): pinned glibc tests` so we can clean these up once CI is resolved

rusty-snake:
I have no time to work on this anymore.

PR #4403, opened 2025-04-13
adding pid_type enum for Linux.

devnexen:
[ref](https://github.com/torvalds/linux/blob/7cdabafc001202de9984f22c973305f424e0a8b7/include/linux/pid_types.h#L5)

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -92,6 +92,16 @@ e! {
     }
 }
 
+c_enum! {
+    pid_type {
+        PIDTYPE_PID,
+        PIDTYPE_TGID,
+        PIDTYPE_PGID,
+        PIDTYPE_SID,
+        PIDTYPE_MAX,
+    }
+}
+
 s! {
     pub struct glob_t {
         pub gl_pathc: size_t,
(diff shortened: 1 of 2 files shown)


tgross35:
We should probably avoid adding any more Rust `enum` s because of the trivial unsoundness - I wrote this up a bit more in https://github.com/rust-lang/libc/issues/4419. To work around this, I'm adding a `c_enum!` macro that produces constants rather than a Rust enum https://github.com/rust-lang/libc/pull/4420. Would you be able to try using that here?

devnexen:
Oh I see yes I ll try it out a bit later, so by quick glancing it, we can t choose the underlying type of the values right ?

tgross35:
The macro for this should expand to

```rust
pub type pid_type = c_uint;
pub const PIDTYPE_PID: pid_type = 0;
pub const PIDTYPE_TGID: pid_type = 1
// etc;
```

I made it accept `#[repr(u16)]` to change the type alias if that is needed. I think the default `c_uint` should be fine here though, right?

src/unix/linux_like/linux/mod.rs
```diff
@@ -92,6 +92,16 @@ e! {
     }
 }
 
+c_enum! {
+    e {
```
tgross35, line 96: This identifier is the name of the type alias, so should be `pid_type` here. I should add some docs...

devnexen: of course ... thanks :)

tgross35 (approved):
Thanks!

PR #4405, opened 2025-04-14
chore: minor cleanup `mode_t` usage

nyurik:
`crate::mode_t` is almost never needed because it is part of prelude.  Moreover, in many cases `mode_t` was already used without the prefix - thus confusing if they are different or not. Keeping it the same helps readability.


tgross35 (approved):
Thanks!

PR #4406, opened 2025-04-14
chore: consolidate workspace settings [MSRV 1.64+]

nyurik:
* Move common workspace settings to the `[workspace]` sections in the root.
* Add common clippy lint configuration

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,7 +50,7 @@ jobs:
     name: Verify build
     strategy:
       matrix:
-        toolchain: [stable, nightly, 1.63.0]
+        toolchain: [stable, nightly, 1.64.0]
         os: [ubuntu-24.04, macos-14, windows-2022]
         include:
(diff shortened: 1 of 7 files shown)


tgross35 (changes requested):
We don't want to do a MSRV bump at this time - is this required?

The other cleanup looks good but I'd rather defer anything that doesn't work in 1.63.

nyurik:
@tgross35 sadly we can't use workspace settings inheritence until 1.64, which stabilized `workspace-inheritance` :(

tgross35:
Ah, I was thinking it worked for some fields before then. We can land changes that don't depend on that like removing `build = "build.rs"`, but organization improvements in this repo aren't enough to justify a user-facing new MSRV. We'll likely be bumping it sometime later this year, but until then

@rustbot blocked

tgross35:
@nyurik could you rebase this? We have a MSRV bump in the queue

Issue #3998, opened 2024-10-25
CI segfaults on FreeBSD with the latest cargo


asomers:
For the last 2 days CI has been failing on FreeBSD due to a segfault in Cargo.  Upstream issue: https://github.com/rust-lang/cargo/issues/14731

tgross35:
The upstream problem still exists but this hasn't been relevant on our CI for a while.

PR #4408, opened 2025-04-14
chore: cleanup Cargo.toml and rm perl

nyurik:
This PR is a part of the #4406 per @tgross35 suggestion

* Remove Cargo settings that are identical to the defaults (per Cargo [recommendations](https://doc.rust-lang.org/cargo/reference/manifest.html))
* Fix perl-based method to extract MSRV
* Note that the order of the values is such as to make it easier to switch to the inherited ones from workspace once MSRV is bumped

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -255,7 +255,7 @@ jobs:
     steps:
     - uses: actions/checkout@master
     - run: |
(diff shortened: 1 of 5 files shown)


tgross35 (approved):

tgross35:
Thanks!

tgross35:
FreeBSD 13 has a spurious segfault, unrelated to this change of course. It may be related to https://github.com/rust-lang/libc/issues/3998 but I'm not sure why it isn't consistent.

PR #4413, opened 2025-04-15
chore: lint `ctest-test` crate

nyurik:
run `cargo clippy --all-targets` on `ctest-test`, and fix all default issues. Also added a generic


ctest-test/src/t1.rs
```diff
@@ -87,6 +87,7 @@ extern "C" {
     pub static T1static: c_uint;
 }
 
+#[expect(clippy::eq_op)]
```
tgross35, line 90: Can we just ignore clippy in this crate's src/lib.rs? `tests/` and `build.rs` have real source but the "source" is mostly dummy files.

tgross35 (changes requested):
Side note, `ctest` isn't on the `libc-0.2` branch at all so there are no backports needed for these changes.

ctest-test/src/t1.rs
nyurik: ok, disabled

tgross35 (approved):
LGTM, please squash

nyurik:
thx, done.  Why isn't libc using the github squash command?

tgross35 (approved):
Thanks!

Annoyingly with merge queues you can't choose the strategy on a case-by-case basis, and we do need to preserve commits in some cases. Sometimes I can squash and force push via the gh UI, but better if the author can do it so they don't get out of sync.

PR #4417, opened 2025-04-15
chore: use auto-derived copy/clone impls

nyurik:
Unless there is something really magical, it doesn't seem to make sense to have manual impl for the most commonly used traits

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -81,30 +81,16 @@ pub type rlim_t = c_ulonglong;
 // FIXME(fuchsia): why are these uninhabited types? that seems... wrong?
 // Presumably these should be `()` or an `extern type` (when that stabilizes).
 #[cfg_attr(feature = "extra_traits", derive(Debug))]
+#[derive(Clone, Copy)]
 pub enum timezone {}
-impl Copy for timezone {}
-impl Clone for timezone {
-    fn clone(&self) -> timezone {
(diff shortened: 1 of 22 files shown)


tgross35:
I'm not sure if it would work but you can try adding `::core::prelude::v1::derive` to the prelude at https://github.com/rust-lang/libc/blob/0fe1d452daefd36c30e56952349641b6d19d51e5/src/macros.rs#L65, since that isn't in scope by default when building as part of std.

nyurik:
@tgross35 thx, but I am still getting a really strange warning in the [style check](https://github.com/rust-lang/libc/actions/runs/14476558988/job/40603497387?pr=4417) -- what is an `s!` macro, and why rust/clippy suggests it?

```
error: impl Copy and Clone manually
  --> /home/runner/work/libc/libc/libc-test/../src/fuchsia/mod.rs:84:3
   |
84 | #[derive(Clone, Copy)]
   |   ^^^^^^^^^^^^^^^^^^^ found manual implementation of Copy and/or Clone
   |
   = help: use one of the s! macros instead
```

tgross35:
Most of the binding structs are wrapped in `s!` which adds the trait implementations. For enums, `e!` should do the same.

@rustbot author

tgross35:
This has been resolved through a series of cleanup patches, so isn't relevant anymore. Thanks for the work here!

PR #4422, opened 2025-04-16
Add AT_EXECVE_CHECK

rusty-snake:
# Description

Add `AT_EXECVE_CHECK`

# Sources

https://github.com/torvalds/linux/blob/c62f4b82d57155f35befb5c8bbae176614b87623/include/uapi/linux/fcntl.h#L159

# Checklist

---


tgross35 (approved):

src/unix/linux_like/linux/mod.rs
```diff
@@ -1,13 +1,7 @@
...
 
 use crate::prelude::*;
-use crate::{
-    sock_filter,
-    _IO,
-    _IOR,
-    _IOW,
-    _IOWR,
-};
+use crate::{sock_filter, _IO, _IOR, _IOW, _IOWR};
```
tgross35: Looks like this may have snuck in when rebasing

tgross35:
@rusty-snake another gentle ping here: I think this just needs a rebase, and for the unrelated part of the diff to be dropped.

src/unix/linux_like/linux/mod.rs
rusty-snake: That's caused by automatic rustfmt run when saving the file.

tgross35 (approved):
Thank you!

PR #4423, opened 2025-04-18
linux updating the remain e! macro to c_enum!

devnexen:


tgross35 (approved):
Thanks!

@rustbot label +stable-declined

Issue #4369, opened 2025-04-03
ctest: return a result


tgross35:
The [`generate` method](https://docs.rs/ctest/0.4.8/ctest/struct.TestGenerator.html#method.generate) returns nothing, but it would occasionally be useful to have non-panicking handling of errors. It should instead return some result type.

Owen-CH-Leung:
I'd like to take up this issue. I'll file a PR soon.

PR #4424, opened 2025-04-19
Revise generate function to return Result

Owen-CH-Leung:
# Description

As per #4369 , the generate function should return a Result instead of returning nothing

# Checklist


ctest-test/build.rs
```diff
@@ -48,7 +49,62 @@ fn main() {
...
+
+    assert!(err.is_err(), "Expected error with non-existent header");
+
+    // Test invalid include path
+    let err = ctest::TestGenerator::new()
+        .header("t1.h")
+        .include("nonexistent_directory")
+        .generate("src/t1.rs", "invalid_include_gen.rs");
+
+    assert!(err.is_err(), "Expected error with invalid include path");
```
tgross35, line 107: `build.rs` works for the test that runs the whole crate, but these `assert!` tests should be able to work on their own. Could you move them to `ctest-test/tests/all.rs`?

ctest/src/lib.rs
```diff
@@ -793,9 +795,9 @@ impl TestGenerator {
     /// let mut cfg = TestGenerator::new();
     /// cfg.generate("../path/to/libfoo-sys/lib.rs", "all.rs");
     /// ```
-    pub fn generate<P: AsRef<Path>>(&mut self, krate: P, out_file: &str) {
+    pub fn generate<P: AsRef<Path>>(&mut self, krate: P, out_file: &str) -> Result<PathBuf> {
```
tgross35, line 798: It would be best if we could do this in a way that doesn't break semver, just because this repo currently has a tricky branch setup. Could `generate` keep its signature, but move the body to a new `fn try_generate(...) -> Result`? The existing `generate` can just call `try_generate` and then panic if there is a failure.

ctest/src/lib.rs
tgross35, line 1: I think it would be easiest to switch our `Error` to [`anyhow`](https://docs.rs/anyhow/latest/anyhow/)'s `Error`. This gives us `.context()` and `.with_context()` which should make using `?` a lot easier.

ctest/src/lib.rs
```diff
@@ -851,25 +853,54 @@ impl TestGenerator {
...
-            .compile(&format!("lib{stem}.a"));
+        let stem = out
+            .file_stem()
+            .ok_or_else(|| Error::from("Output file has no stem"))?
+            .to_str()
+            .ok_or_else(|| Error::from("Output filename is not valid UTF-8"))?;
+
+        let parent = out
+            .parent()
+            .ok_or_else(|| Error::from("Output file has no parent directory"))?;
```
tgross35, line 864: After switching to `anyhow` this will be a bit easier to use `.context` and `?`

ctest/src/lib.rs
```diff
@@ -851,25 +853,54 @@ impl TestGenerator {
...
+
+        cfg.target(&target).out_dir(parent);
+
+        match catch_unwind(|| cfg.compile(&format!("lib{stem}.a"))) {
```
tgross35, line 868: Unfortunately not quite :) we want to change all internal functions to return a `Result` as well so it's easier to know what might fail and for simpler testing.

It should be pretty easy to find+replace any `.unwrap()` with `?`, then update the function signatures anywhere it complains.

tgross35 (changes requested):
Thank you for picking this up!

The results of the new tests look good, but unfortunately the internal functions also need to be updated to return `Result` rather than just catching panics.

ctest-test/build.rs
Owen-CH-Leung: Thanks for your review! I'll file a commit to move them to `ctest-test/tests/all.rs` soon.

ctest/src/lib.rs
Owen-CH-Leung: Sure. I'll keep the `generate` function signature and move them to `try_generate` instead

ctest/src/lib.rs
Owen-CH-Leung: Ah ok. Yeah I think using anyhow's Error is a more idiomatic way to handle error. I'll add anyhow as part of dependencies and make the code change soon

Owen-CH-Leung:
Sorry - bad rebase here. Will close this MR and create a new one

PR #4426, opened 2025-04-21
Update Redox SA_ constants.

4lDO2:
# Description

Redox completely redesigned its signal handling code last summer, but this wasn't upstreamed to the libc crate, making crates like signal-hook segfault.

It would be appreciated if a 0.2 minor version could be released with these changes.

Closes: https://github.com/rust-lang/libc/pull/4262

# Sources

https://gitlab.redox-os.org/redox-os/relibc/-/blob/master/src/header/signal/redox.rs?ref_type=heads#L43-50

underlying diff

https://gitlab.redox-os.org/redox-os/relibc/-/merge_requests/480/diffs#d174caa1d6ee478ceffe421ab58c6f035a7e7285_72_39

# Checklist

  included
- [x] Tested locally using https://gitlab.redox-os.org/redox-os/acid/-/blob/master/src/proc.rs?ref_type=heads#L649-667 with and without this patch, respectively

### DIFF
--- a/src/unix/redox/mod.rs
+++ b/src/unix/redox/mod.rs
@@ -661,14 +661,14 @@ pub const SIGPWR: c_int = 30;
 pub const SIGSYS: c_int = 31;
 pub const NSIG: c_int = 32;
 
-pub const SA_NOCLDSTOP: c_ulong = 0x00000001;
-pub const SA_NOCLDWAIT: c_ulong = 0x00000002;
-pub const SA_SIGINFO: c_ulong = 0x00000004;
-pub const SA_RESTORER: c_ulong = 0x04000000;
-pub const SA_ONSTACK: c_ulong = 0x08000000;
-pub const SA_RESTART: c_ulong = 0x10000000;
-pub const SA_NODEFER: c_ulong = 0x40000000;
-pub const SA_RESETHAND: c_ulong = 0x80000000;
+pub const SA_NOCLDWAIT: c_ulong = 0x0000_0002;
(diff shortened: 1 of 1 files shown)


src/unix/redox/mod.rs
```diff
@@ -661,14 +661,14 @@ pub const SIGPWR: c_int = 30;
...
-pub const SA_NOCLDSTOP: c_ulong = 0x00000001;
-pub const SA_NOCLDWAIT: c_ulong = 0x00000002;
-pub const SA_SIGINFO: c_ulong = 0x00000004;
-pub const SA_RESTORER: c_ulong = 0x04000000;
-pub const SA_ONSTACK: c_ulong = 0x08000000;
-pub const SA_RESTART: c_ulong = 0x10000000;
-pub const SA_NODEFER: c_ulong = 0x40000000;
-pub const SA_RESETHAND: c_ulong = 0x80000000;
+pub const SA_NOCLDWAIT: c_ulong = 0x0000_0002;
+pub const SA_RESTORER: c_ulong = 0x0000_0004; // TODO: remove
```
tgross35, line 665: Could you change this to something like `FIXME(redox): will be removed in a future version`? Since the TODO doesn't say much relevant to this crate, and we try to use tagged fixmes anyway.

Otherwise LGTM!

tgross35 (changes requested):

src/unix/redox/mod.rs
4lDO2: Done

tgross35 (approved):
Thanks for the change!

PR #4430, opened 2025-04-26
ctest: Revise generate function to return Result

Owen-CH-Leung:
# Description

As per rust-lang/libc#4369 , the generate function should return a Result instead of returning nothing. 

This PR is a continuation of rust-lang/libc#4424 which was closed due to bad rebase.

# Checklist

### DIFF
--- a/ctest/src/lib.rs
+++ b/ctest/src/lib.rs
@@ -13,6 +13,9 @@
 #![recursion_limit = "256"]
 #![deny(missing_docs)]
 
+use anyhow::{anyhow, Context, Result};
+use garando_syntax as syntax;
+use indoc::writedoc;
 use std::collections::{HashMap, HashSet};
 use std::env;
 use std::fs::File;
@@ -20,9 +23,6 @@ use std::io::prelude::*;
 use std::io::BufWriter;
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
-
-use garando_syntax as syntax;
-use indoc::writedoc;
 use syntax::abi::Abi;
 use syntax::ast;
 use syntax::ast::{Attribute, Name};
@@ -40,9 +40,6 @@ use syntax::parse::{self, ParseSess};
 use syntax::ptr::P;
 use syntax::util::small_vector::SmallVector;
 use syntax::visit::{self, Visitor};
-
-type Error = Box<dyn std::error::Error>;
-type Result<T, E = Error> = std::result::Result<T, E>;
 
 /// Programming language
 #[derive(Debug)]
@@ -773,6 +770,17 @@ impl TestGenerator {
         self
     }
 
+    /// Generate all tests and panic on any errors.
+    ///
+    /// This function is a convenience wrapper around `try_generate` that panics instead of returning
+    /// errors.
+    ///
+    /// See `try_generate` for the error-handling version of this function.
+    pub fn generate<P: AsRef<Path>>(&mut self, krate: P, out_file: &str) {
+        self.try_generate(krate, out_file)
+            .unwrap_or_else(|e| panic!("Failed to generate tests: {e}"));
+    }
+
     /// Generate all tests.
     ///
     /// This function is first given the path to the `*-sys` crate which is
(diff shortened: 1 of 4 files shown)


tgross35 (changes requested):

ctest/src/lib.rs
```diff
@@ -13,16 +13,18 @@
...
+use anyhow::{Context, Result};
+use garando_syntax as syntax;
+use indoc::writedoc;
+use std::any::Any;
 use std::collections::{HashMap, HashSet};
 use std::env;
 use std::fs::File;
 use std::io::prelude::*;
 use std::io::BufWriter;
+use std::panic::{catch_unwind, AssertUnwindSafe};
```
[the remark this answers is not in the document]
Owen-CH-Leung: Thanks. I've removed the remaining `catch_unwind` logic.

ctest/src/lib.rs
```diff
@@ -851,25 +859,58 @@ impl TestGenerator {
...
+        let parent = out
+            .parent()
+            .context("Output file has no parent directory")?;
+
+        cfg.target(&target).out_dir(parent);
+
+        match cfg.try_compile(&format!("lib{stem}.a")) {
+            Ok(_) => Ok(out),
+            Err(e) => Err(anyhow::anyhow!("Compilation failed").context(format!("Error: {}", e))),
+        }
```
tgross35, line 877: I think the error handling could be a bit simplified here:

```rust
let name = format!("lib{stem}.a");
cfg.try_compile(&name)
    .context("failed to compile `{name}`")?
    .map(|()| out)
```

ctest/src/lib.rs
```diff
@@ -851,25 +859,58 @@ impl TestGenerator {
...
+        for header in &self.headers {
+            // Check if header exists in any of the include paths
+            for include_path in &self.includes {
+                let header_path = include_path.join(header);
+                if !header_path.exists() {
+                    return Err(anyhow::anyhow!("Header file not found"))
+                        .with_context(|| format!("Path {} does not exist", header_path.display()));
+                }
+            }
+        }
```
tgross35, line 898: Where do the header files actually get read? There is a TOCTOU error here (albeit mild). It would be better to just add an error message wherever they are opened, and this block can be removed.

ctest/src/lib.rs
```diff
@@ -851,25 +859,58 @@ impl TestGenerator {
...
-            .unwrap_or_else(|| PathBuf::from(env::var_os("OUT_DIR").unwrap()));
+            .or_else(|| env::var_os("OUT_DIR").map(PathBuf::from))
+            .context("Neither out_dir nor OUT_DIR environment variable is set")?;
+
+        if !out_dir.is_dir() {
+            return Err(anyhow::anyhow!(
+                "Output directory does not exist: {}",
+                out_dir.display()
+            ));
+        }
```
tgross35, line 912: Similar to the above, we shouldn't really need this check.

ctest/src/lib.rs
```diff
@@ -878,18 +919,27 @@ impl TestGenerator {
...
         }
 
-        // Parse the libc crate
-        let krate = parse::parse_crate_from_file(krate, &sess).ok().unwrap();
+        // Convert DiagnosticBuilder -> Error so the `?` works
+        let krate = parse::parse_crate_from_file(krate, &sess).map_err(|mut d| {
+            let diag_info = format!("{:?}", d);
+            // Emit the diagnostic to properly handle it and show error to the user
+            d.emit();
+            anyhow::anyhow!("failed to parse crate").context(format!("Diagnostic: {}", diag_info))
```
tgross35, line 941:
```rust
            d.emit();
            anyhow!("failed to parse crate: {d:?}")
```

No need to put the diagnostic in context since it goes with the error, or to `format!` it separately. Also you can import `anyhow::anyhow` since it's used a few times.

tgross35: Actually, why `.emit()` the diagnostic at all? Does that give different output compared to printing it with either `Debug` or `Display`?

tgross35 (commented):
Looks much better! I left a few more comments.

ctest/src/lib.rs
Owen-CH-Leung: the `emit()` is necessary becoz without the emit call, the `DiagnosticBuilder` will panic on drop: 

https://github.com/JohnTitor/garando/blob/master/garando_errors/src/diagnostic_builder.rs#L157

ctest/src/lib.rs
Owen-CH-Leung: Sure & removed

ctest/src/lib.rs
Owen-CH-Leung: Indeed. I've removed this block. Thanks!

Owen-CH-Leung:
Looks like a flaky test ? 

https://github.com/rust-lang/libc/pull/4430/checks?check_run_id=41604457695

This `unwrap()` line is causing panic:

https://github.com/rust-lang/libc/blob/main/ctest/src/lib.rs#L168

Owen-CH-Leung:
Looks like this is a flaky test : 

https://github.com/rust-lang/libc/runs/41604457695

I re-trigger the CI and it passes

Owen-CH-Leung:
> LGTM! We can update more `.unwrap()` to proper errors later. Please squash.

Done :)

tgross35 (approved):
Thanks for the work here!

PR #4434, opened 2025-04-30
Add constants from linux/cn_proc.h and linux/connector.h

Bben01:
# Description

Add constants from linux/cn_proc.h and linux/connector.h

# Sources

https://github.com/torvalds/linux/blob/b6ea1680d0ac0e45157a819c41b46565f4616186/include/uapi/linux/cn_proc.h
https://github.com/torvalds/linux/blob/b6ea1680d0ac0e45157a819c41b46565f4616186/include/uapi/linux/connector.h

# Checklist

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -4694,6 +4694,51 @@ pub const RTNLGRP_MCTP_IFADDR: c_uint = 0x22;
 pub const RTNLGRP_TUNNEL: c_uint = 0x23;
 pub const RTNLGRP_STATS: c_uint = 0x24;
 
+// linux/cn_proc.h
+c_enum! {
+    proc_cn_mcast_op {
+        PROC_CN_MCAST_LISTEN = 1,
+        PROC_CN_MCAST_IGNORE = 2,
+    }
+}
+
+c_enum! {
+    proc_cn_event {
+        PROC_EVENT_NONE = 0x00000000,
+        PROC_EVENT_FORK = 0x00000001,
+        PROC_EVENT_EXEC = 0x00000002,
+        PROC_EVENT_UID = 0x00000004,
+        PROC_EVENT_GID = 0x00000040,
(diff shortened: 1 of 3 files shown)


Bben01:
I think that the CI is failing on i686-unknown-linux-gnu because it didn't find the header files. What should I do to fix that?

src/unix/linux_like/linux/mod.rs
```diff
@@ -4708,6 +4708,43 @@ pub const RTNLGRP_MCTP_IFADDR: c_uint = 0x22;
 pub const RTNLGRP_TUNNEL: c_uint = 0x23;
 pub const RTNLGRP_STATS: c_uint = 0x24;
 
+// linux/cn_proc.h
+pub const PROC_CN_MCAST_LISTEN: c_int = 1;
+pub const PROC_CN_MCAST_IGNORE: c_int = 2;
+
+pub const PROC_EVENT_NONE: c_int = 0x00000000;
```
tgross35, line 4715: There is a `c_enum!` macro for representing enums as a type plus constants, could you use that here?

tgross35:
If these are more recent than our kernel versions in CI, you will need to update `libc-test/build.rs` to skip them with a FIXME. If you do this, just make sure libc-test passes locally without the skips (assuming you have a newer version locally).

Bben01:
I don't know why the style check fails. It says to replace the `=` with `:` in the `c_enum`, but the macro cannot accept `:`

```
help: replace equals symbol with a colon
     |
4760 -         PROC_EVENT_EXIT = 0x80000000,
4760 +         PROC_EVENT_EXIT: 0x80000000,
```

I don't know if the segfault in freebsd is linked to my PR, but if it is, I don't know what I can do to fix it

@rustbot ready

Bben01:
I've added `c_enum` to the ignore list in `.ci/style.sh` because rustfmt cannot handle the case when values are specified. If you have a better solution, I will implement it instead

Bben01:
@tgross35 do you know when the next release is coming out?

PR #4438, opened 2025-05-01
musl: fix test build with musl 1.2.0+

xbjfk:
# Description
Since musl 1.2.0, the utmpx.ut_session type changed from long to int (with padding). For now, skip the test for this field. This fixes rust-lang/libc#3305. This change type can probably be changed and this commit reverted when rust-lang/libc#3068 is merged.

# Sources

[diff](https://github.com/bminor/musl/commit/1e7f0fcd7ff2096904fd93a2ee6d12a2392be392)
[permalink](https://github.com/bminor/musl/blob/c47ad25ea3b484e10326f933e927c0bc8cded3da/include/utmpx.h#L30)

# Checklist

- [X] Relevant tests in `libc-test/semver` have been updated
- [X] No placeholder or unstable values like `*LAST` or `*MAX` are
  included (see [rust-lang/libc#3131](https://github.com/rust-lang/libc/issues/3131))
- [X] Tested locally (`cd libc-test && cargo test --target mytarget`);
  especially relevant for platforms that may not be checked in CI

@rustbot label +stable-nominated

PR #4433, opened 2025-04-28
Gnu time bits 64

snogge:
# Description

Add a new cfg `gnu_time_bits64` corresponding to
`-D_TIME_BITS=64`.
The cfg can also be enabled by setting the environment variable
`RUST_LIBC_UNSTABLE_GNU_TIME_BITS64=1`. 

This is the third part of what was rust-lang/libc#3175 .

# Sources

Sources are listed listed in each commit message.

`struct stat64`:
 ARM:
  https://github.com/bminor/glibc/blob/77930e0447e0b37a129db0e13c6c6f5e60a3019e/sysdeps/unix/sysv/linux/arm/bits/struct_stat.h
 MIPS:
  https://github.com/bminor/glibc/blob/77930e0447e0b37a129db0e13c6c6f5e60a3019e/sysdeps/unix/sysv/linux/mips/bits/struct_stat.h
 POWERPC:
  https://github.com/bminor/glibc/blob/77930e0447e0b37a129db0e13c6c6f5e60a3019e/sysdeps/unix/sysv/linux/powerpc/bits/struct_stat.h
 SPARC:
  https://github.com/bminor/glibc/blob/77930e0447e0b37a129db0e13c6c6f5e60a3019e/sysdeps/unix/sysv/linux/sparc/bits/struct_stat.h
 x86:
  https://github.com/bminor/glibc/blob/77930e0447e0b37a129db0e13c6c6f5e60a3019e/sysdeps/unix/sysv/linux/x86/bits/struct_stat.h
 Common definition for _TIME_BITS=64
  https://github.com/bminor/glibc/blob/77930e0447e0b37a129db0e13c6c6f5e60a3019e/sysdeps/unix/sysv/linux/bits/struct_stat_time64_helper.h

`struct stat`:
 ARM:
  https://github.com/bminor/glibc/blob/77930e0447e0b37a129db0e13c6c6f5e60a3019e/sysdeps/unix/sysv/linux/arm/bits/struct_stat.h
 MIPS:
  https://github.com/bminor/glibc/blob/77930e0447e0b37a129db0e13c6c6f5e60a3019e/sysdeps/unix/sysv/linux/mips/bits/struct_stat.h
 POWERPC:
  https://github.com/bminor/glibc/blob/77930e0447e0b37a129db0e13c6c6f5e60a3019e/sysdeps/unix/sysv/linux/powerpc/bits/struct_stat.h
 SPARC:
[...]

PR #4439, opened 2025-05-01
musl: fix CPU_SETSIZE

xbjfk:
# Description
musl upstream has changed the size from 128 to 1024 in upstream commit
bc695a5, which is in versions 1.2.4+. I see this was fixed for loongarch in this repository, but this change applies to all architectures.

# Sources
[loongarch commit that changes this](https://github.com/rust-lang/libc/commit/bfdc3708196b02b854648b5f245f1304e8875029)
[upstream change](https://github.com/bminor/musl/commit/bc695a5ac1d7929e5c1ad5297eb47e146cccd157)
[upstream permalink](https://github.com/bminor/musl/blob/c47ad25ea3b484e10326f933e927c0bc8cded3da/include/sched.h#L126)
# Checklist

I'm not sure if this can be nominated for 0.2. I would appreciate if someone could confirm.

NOTE: review/merge after rust-lang/libc#4438 - it includes a commit from that PR.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -4795,7 +4795,9 @@ fn test_linux(target: &str) {
         (struct_ == "statvfs" && field == "__f_spare") ||
         (struct_ == "statvfs64" && field == "__f_spare") ||
         // the `xsk_tx_metadata_union` field is an anonymous union
-        (struct_ == "xsk_tx_metadata" && field == "xsk_tx_metadata_union")
+        (struct_ == "xsk_tx_metadata" && field == "xsk_tx_metadata_union") ||
+        // FIXME(musl): After musl 1.2.0, the type becomes `int` instead of `long`.
+        (struct_ == "utmpx" && field == "ut_session")
     });
 
     cfg.skip_roundtrip(move |s| match s {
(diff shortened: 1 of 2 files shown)


xbjfk:
Hmmm, looks like the CI uses a version of musl from 6 years ago (1.1.24). Maybe the version should be upgraded? It feels awkward essentially keeping musl bugs around longer than they need to be to appease the CI which tests against an ancient version. `RLIM_NLIMITS` has the same issue.

xbjfk:
Since it seems the minimum supported musl version by rust has been 1.2.3 for 2 years already, I think it is safe to update the CI version. I will look into this soon, and combine this with #3068 and #4438, as well as fix the other issues I am having when running tests (such as `RLIM_NLIMITS` being 16 now)

tgross35:
Musl indeed really needs to be updated, it needs some help to get over the line (thank you for offering!).

Unfortunately we can't easily change things all at once because there are some user-visible breaks. I think the best solution is probably to add the new interfaces bit-by-bit under a new `cfg(musl_v1_2)`, which can be set via an unstable environment variable for experimentation before the final switch (which will probably be libc's 1.0).

This is similar to what is being done for glibc's time, https://github.com/rust-lang/libc/pull/4433.

xbjfk:
I agree that there should be a new config option - I think the priority would be to get tests fully working on 1.2.3, and focus on making those changes compatible with libc 0.2. There are a couple of constant changes that will be non-breaking

xbjfk:
Closing in favor of #4443

PR #4441, opened 2025-05-02
redox: make CMSG_ALIGN, CMSG_LEN, and CMSG_SPACE const functions

jackpot51:
# Description

This makes CMSG_ALIGN, CMSG_LEN, and CMSG_SPACE const functions, which matches other platforms and fixes issues in rustix.

# Sources

The function implementations came from https://gitlab.redox-os.org/redox-os/relibc/-/blob/master/src/header/sys_socket/mod.rs#L72

# Checklist


tgross35 (approved):
Thanks!

PR #4443, opened 2025-05-03
Start update to musl v1.2.3

xbjfk:
# Description

This is effectively a partial continuation of rust-lang/libc#3068 (no time64 yet), however done behind a flag so people can test changes in 0.2. Some changes are not behind the musl_v1_2_3 config flag, and these are: 
 - The change to `RLIM_NLIMITS`. This one is definitely safe, as it has had a deprecation warning saying it could change any time with OS version for >5 years
 - The change to `O_LARGEFILE` is likely safe.
 - Same with the change to the `tcp_info` struct, as these only add fields while keeping ABI. However, if manually initializing all fields this could cause breakage.

Let me know if any of m assumptions are incorrect and I am happy to gate those changes behind the cfg as well.

I've also added rust-lang/libc#4437, as this seems like a good chance to rename the field while the cfg option is in place.

# Sources

O_LARGEFILE change: https://github.com/bminor/musl/commit/b8b729bd22c28c9116c2fce65dce207a35299c26
RLIM_NLIMITS change: https://github.com/bminor/musl/commit/2507e7f5312e79620f6337935d0a6c9045ccba09
utmpx change: https://github.com/bminor/musl/commit/1e7f0fcd7ff2096904fd93a2ee6d12a2392be392
tcp_info changes: https://github.com/bminor/musl/commit/d4f298149c5e36ac5521888d645bfb65f4d027fc, https://github.com/bminor/musl/commit/5e0c9f246cf7c0d174b7b8fbd9e02d4729ff66d2

# Checklist

- [ ] (rest of) time64 API - this can be done in another PR after this one is merged
- [ ] `CPU_SETSIZE` change (not urgent as changed after 1.2.3)


xbjfk:
hmm, looks like I need to allow deprecation for CI to pass

src/unix/linux_like/linux/musl/mod.rs
```diff
@@ -457,31 +442,26 @@ s_no_extra_traits! {
...
+        )]
         pub ut_session: c_long,
 
-        #[cfg(target_env = "musl")]
-        #[cfg(target_arch = "loongarch64")]
-        pub ut_session: c_int,
-
-        #[cfg(target_env = "musl")]
-        #[cfg(target_arch = "loongarch64")]
+        #[cfg(any(target_arch = "loongarch64", target_env = "ohos", musl_v1_2_3))]
```
tgross35, line 455: Maybe it's worth setting `musl_v1_2_3` via build.rs if the target is loongarch64? Since we only support/test 1.2.3+ on that target.

ohos might fall into this same boat based on config.

tgross35 (changes requested):
Thank you for working on this! A handful of small comments, most importantly we don't want to stop testing the current musl version because that needs to continue working until we do the final version bump. Also looks like there is a CI failure.

src/unix/linux_like/linux/musl/mod.rs
xbjfk: Yeah, this is definitely a lot cleaner. I was thinking about the time64 changes, which are not on loongarch64 yet - but these can be controlled by a separate cfg, which can only be enabled through env when `musl_v1_2_3` is.

tgross35:
> ⚠️ **Warning** ⚠️
> 
>     * There are issue links (such as `#123`) in the commit messages of the following commits.
>       _Please remove them as they will spam the issue with references to the commit._
>       
>       * [5f03e36](https://github.com/rust-lang/libc/commit/5f03e36a51907a467127c9a722039842ff1fd066)

Disregard this, that's not a good warning

libc-test/build.rs
```diff
@@ -3648,15 +3648,21 @@ fn test_linux(target: &str) {
...
     let x86_64 = target.contains("x86_64");
-    let aarch64_musl = aarch64 && musl;
     let gnueabihf = target.contains("gnueabihf");
     let x86_64_gnux32 = target.contains("gnux32") && x86_64;
     let riscv64 = target.contains("riscv64");
     let loongarch64 = target.contains("loongarch64");
     let wasm32 = target.contains("wasm32");
     let uclibc = target.contains("uclibc");
 
+    let musl_v1_2_3 = env::var("RUST_LIBC_UNSTABLE_MUSL_1_2_3").is_ok();
```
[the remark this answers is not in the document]
xbjfk: done, renamed the variable to `RUST_LIBC_UNSTABLE_MUSL_V1_2_3`

ci/install-musl.sh
```diff
@@ -10,7 +10,7 @@ case ${1} in
         musl_version=1.2.5
         ;;
     *)
-        musl_version=1.1.24
+        musl_version=1.2.3
```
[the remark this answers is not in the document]
xbjfk: done, CI will test both with and without RUST_LIBC_UNSTABLE_MUSL_V1_2_3, and this will change the version in install-musl.sh

src/unix/linux_like/linux/musl/mod.rs
tgross35: I'm not sure I follow - what changes to time are there on loongarch64 since that is a 64-bit platform? And what would this separate config be?

ci/install-musl.sh
```diff
@@ -10,7 +10,7 @@ case ${1} in
         musl_version=1.2.5
         ;;
     *)
-        musl_version=1.1.24
+        test -z RUST_LIBC_UNSTABLE_MUSL_V1_2_3 && musl_version=1.1.24 || musl_version=1.2.3
```
tgross35, line 13: Is a `$` missing? (looks like it's checking the literal `RUST_LIBC_UNSTABLE_MUSL_V1_2_3`)

build.rs
```diff
@@ -85,6 +86,12 @@ fn main() {
...
 
+    let musl_v1_2_3 = env::var("RUST_LIBC_UNSTABLE_MUSL_V1_2_3").is_ok();
+    println!("cargo:rerun-if-env-changed=RUST_LIBC_UNSTABLE_MUSL_V1_2_3");
+    // loongarch64 and ohos have already updated
+    if musl_v1_2_3 || target_os == "loongarch64" || target_env == "ohos" {
+        set_cfg("musl_v1_2_3");
+    }
     let linux_time_bits64 = env::var("RUST_LIBC_UNSTABLE_LINUX_TIME_BITS64").is_ok();
     println!("cargo:rerun-if-env-changed=RUST_LIBC_UNSTABLE_LINUX_TIME_BITS64");
     if linux_time_bits64 {
```
tgross35, line 97: If musl_v1_2_3 is enabled, should `linux_time_bits64` also be set for `__USE_TIME_BITS64` in UAPI? It's fine if you plan to do the needed time changes later, but it should be a FIXME.

ci/install-musl.sh
xbjfk: Oops, this was a silly omission - will fix on next push

src/unix/linux_like/linux/musl/mod.rs
```diff
@@ -557,6 +535,7 @@ cfg_if! {
         }
 
         impl PartialEq for utmpx {
+            #![allow(deprecated)]
```
tgross35, line 538: This should be an inner attribute

tgross35:
This looks pretty good as a first step to me, I think we can merge it once you get CI passing.

@rustbot author

build.rs
xbjfk: added a FIXME

src/unix/linux_like/linux/musl/mod.rs
xbjfk: good catch!

xbjfk:
Thanks for the review! Looks like the CI passes, except for freebsd (which seems to have had failures previously)

I've started working on the time64 changes, expect a PR for those in a few days if all goes to plan :)

@rustbot ready

tgross35 (approved):
Thank you!

xbjfk:
Yes, that is correct - I'll aim to make a new PR within the next few days :)

xbjfk:
@tgross35 took me a while but I opened #4463 :).

tgross35:
@xbjfk bit of a throwback here, do you remember why you picked `musl_v1_2_3` for the flag name rather than `musl_v1_2`? Were there breaking changes in between 1.2 and 1.2.3?

xbjfk:
I don't think there were any breaking changes, it should technically work but I believe I picked it because Rust itself jumped from 1.1.24 to 1.2.3 directly in Rust 1.71.0: https://github.com/rust-lang/rust/pull/107129/

tgross35:
Ah, that sounds right. There was just a bit of confusion so I think I may rename this cfg in the near future.

PR #4447, opened 2025-05-06
Add MADV_SOFT_OFFLINE definition for RISC-V musl targets

mnissler-rivos:
# Description

This adds missing MADV_SOFT_OFFLINE definitions for RISC-V musl targets. It is analogous to https://github.com/rust-lang/libc/pull/2391 where the same definitions were added for RISC-V glibc targets. The patch addresses a rustix build issue: https://github.com/bytecodealliance/rustix/issues/1462

# Sources

https://github.com/bminor/musl/blob/c47ad25ea3b484e10326f933e927c0bc8cded3da/include/sys/mman.h#L99

# Checklist

  -- I have added MADV_SOFT_OFFLINE to libc-test/semver/linux-riscv64gc.txt for good measure
- [x] Tested locally (`cd libc-test && cargo test --target riscv64gc-unknown-linux-{gnu,musl}`);
  -- I've tested the `riscv64gc-unknown-linux-musl` target against a riscv-gnu-toolchain musl build - there are actually a bunch of other issues with struct layout, field name, and value inconsistencies, but these are unrelated to my change.


mnissler-rivos:
There's one consideration I want to bring up: Instead of adding the missing definitions to the RISC-V musl targets, we could also opt for a single definition at `linux_like` level, with cfg guards to exempt mips platforms. The definition was actually more centralized, but https://github.com/rust-lang/libc/commit/1cbc523e50d7dd37f13330d3ef5b80d84b76705e changed it to the approach that is currently used, and I don't know what the rationale/preference is. Let me know if you'd prefer a conditionally compiled definition in `linux_like` and I'll happily update the pull request.

tgross35:
> There's one consideration I want to bring up: Instead of adding the missing definitions to the RISC-V musl targets, we could also opt for a single definition at `linux_like` level, with cfg guards to exempt mips platforms. The definition was actually more centralized, but [1cbc523](https://github.com/rust-lang/libc/commit/1cbc523e50d7dd37f13330d3ef5b80d84b76705e) changed it to the approach that is currently used, and I don't know what the rationale/preference is. Let me know if you'd prefer a conditionally compiled definition in `linux_like` and I'll happily update the pull request.

I don't like this repo's layout :) we'll be reorganizing things soon, don't worry about it for now.

tgross35 (approved):

PR #4451, opened 2025-05-10
adding SO_SPLICE socket option support for freebsd >= 14.2

devnexen:
[ref](https://github.com/freebsd/freebsd-src/blob/d3f15bc2a51d1822795135d9ad4627dc1c7f2b18/sys/sys/socket.h#L175) and
[ref](https://github.com/freebsd/freebsd-src/blob/d3f15bc2a51d1822795135d9ad4627dc1c7f2b18/sys/sys/socketvar.h#L76)

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2781,6 +2781,9 @@ fn test_freebsd(target: &str) {
             // FIXME(freebsd): Removed in FreeBSD 15, deprecated in libc
             "TCP_PCAP_OUT" | "TCP_PCAP_IN" => true,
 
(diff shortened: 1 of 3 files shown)


tgross35:
@devnexen could you please add a link to the headers in the description? I can't find anything from a quick search.

@asomers could you review this?

asomers (approved):

tgross35 (approved):
> This LGTM except that the change to src/windows/mod.rs is unrelated and needs to be reverted.

... that explains why I couldn't find `get_tzname` for FreeBSD. Thanks for the catch.

PR #4452, opened 2025-05-11
linux: add new flags for pwritev2/preadv2

the8472:
# Description

RWF_* flags supported by newer linux kernels.

- RWF_NOAPPEND 
- RWF_ATOMIC
- RWF_DONTCACHE (doesn't have a manpage entry yet)

# Sources

https://github.com/torvalds/linux/blob/3ce9925823c7d6bb0e6eb951bf2db0e9e182582d/include/uapi/linux/fs.h#L336-L343

# Checklist


the8472:
Hrm, I guess it's testing against CI kernel and its headers aren't new enough?

src/unix/linux_like/linux/gnu/mod.rs
tgross35, line 1: Since these are uapi, they can go in `linux_like/linux/mod.rs` to cover both glibc and musl (and others)

tgross35 (approved):
Pushed to squash since I can't do it with the merge queue API

Issue #4449, opened 2025-05-09
Missing strftime on Illumos


extrowerk:
Reference: https://github.com/fish-shell/fish-shell/issues/11099#issuecomment-2619613196

Illumos strftime manpage: https://smartos.org/man/3C/strftime

Fish git master build output:

```
user@build:~/projekt/fish-shell% cargo build
warning: fish@4.1.0-alpha0: Cannot find sphinx-build to build man pages.
warning: fish@4.1.0-alpha0: If you install it now you need to run `cargo clean` and rebuild, or set $FISH_BUILD_DOCS=1 explicitly.
   Compiling fish v4.1.0-alpha0 (/export/home/user/projekt/fish-shell)
error[E0432]: unresolved import `libc::strftime`
    --> src/history.rs:1451:17
     |
1451 |             use libc::strftime;
     |                 ^^^^^^^^^^^^^^ no `strftime` in the root

error[E0433]: failed to resolve: could not find `resource` in `sys`
  --> src/builtins/ulimit.rs:18:66
   |
18 |     let resource = unsafe { std::mem::transmute::<i32, nix::sys::resource::Resource>(resource) };
   |                                                                  ^^^^^^^^ could not find `resource` in `sys`

error[E0433]: failed to resolve: could not find `resource` in `sys`
  --> src/builtins/ulimit.rs:19:15
   |
19 |     nix::sys::resource::getrlimit(resource)
   |               ^^^^^^^^ could not find `resource` in `sys`

error[E0433]: failed to resolve: could not find `resource` in `sys`
  --> src/builtins/ulimit.rs:27:66
   |
27 |     let resource = unsafe { std::mem::transmute::<i32, nix::sys::resource::Resource>(resource) };
   |                                                                  ^^^^^^^^ could not find `resource` in `sys`

error[E0433]: failed to resolve: could not find `resource` in `sys`
  --> src/builtins/ulimit.rs:28:15
   |
28 |     nix::sys::resource::setrlimit(resource, rlim_cur, rlim_max)
   |               ^^^^^^^^ could not find `resource` in `sys`

error[E0425]: cannot find function `mkostemp` in crate `libc`
    --> src/fallback.rs:126:24
     |
126  |         unsafe { libc::mkostemp(name, O_CLOEXEC) }
     |                        ^^^^^^^^ help: a function with a similar name exists: `mkstemp`
     |
[...]

PR #4453, opened 2025-05-11
Add strftime* for illumos.

devnexen:
[ref](https://smartos.org/man/3C/strftime)

Fixes: https://github.com/rust-lang/libc/issues/4449

### DIFF
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -3202,6 +3202,21 @@ extern "C" {
     pub fn arc4random_uniform(upper_bound: u32) -> u32;
 
     pub fn secure_getenv(name: *const c_char) -> *mut c_char;
+
+    #[cfg_attr(target_os = "solaris", link_name = "__strftime_xpg7")]
+    pub fn strftime(
+        s: *mut c_char,
+        maxsize: size_t,
+        format: *const c_char,
+        timeptr: *const crate::tm,
+    ) -> size_t;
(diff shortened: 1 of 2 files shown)


psumbera:
Any chance to get this added also for Solaris? This should be common for both Illumos and Solaris.

tgross35 (changes requested):
Looks like there is a Solaris test failure, does this need a different link name or something?

>  bad strftime function pointer: rust: 140736400656704 (0x7fffbf2b0940) != c 140736400656800 (0x7fffbf2b09a0)

devnexen:
Seems like it, I ll have a look later today.

src/unix/solarish/mod.rs
```diff
@@ -3202,6 +3202,21 @@ extern "C" {
     pub fn arc4random_uniform(upper_bound: u32) -> u32;
 
     pub fn secure_getenv(name: *const c_char) -> *mut c_char;
+
+    #[link_name = "__xpg4_strftime"]
```
pfmooney, line 3206: Doing this unconditionally for both Solaris and illumos is going to break the latter, I believe.

devnexen: in fact I m just trying to see if it fixes solaris case :) thus this PR being a draft for now

tgross35 (approved):
Assuming this is good on Illumos since `link_name` isn't overridden. Thank you!

PR #4458, opened 2025-05-22
Correct the function signatures for getpriority/setpriority

chriswailes:
The `__priority_which_t` type was incorrectly exposed as a type alias and defined as unsigned when they should be signed.

# Description

This CL removes the `__priority_which_t` type alias from the `unix::linux_like::linux::gnu` module.  It is not used for the same purpose as in the C++ source (enum value definition) and it is not supposed to be exposed to the user (see `man setpriority`).  The type signatures for `getpriority` and `setpriority` are then updated to use the correct `c_int` type.

# Sources

Libc source references:
 *https://github.com/bminor/glibc/blob/4052d99ead880797cf271309fd87ddd2b95bd353/resource/sys/resource.h#L105
 * https://github.com/bminor/glibc/blob/4052d99ead880797cf271309fd87ddd2b95bd353/sysdeps/unix/sysv/linux/bits/resource.h#L187
 
C++ Standard Reference:
* Default underlying type for an enum is `int`: https://en.cppreference.com/w/cpp/language/enum

# Checklist

Compilation error when running `cargo test --target x86_64-unknown-linux-gnu`:

```
error[E0061]: this function takes 3 arguments but 2 arguments were supplied
    --> ctest/src/lib.rs:2496:25
     |
2496 |                 Rc::new(macro_rules::compile(self.parse_sess, item)),
     |                         ^^^^^^^^^^^^^^^^^^^^                  ---- argument #2 of type `&RefCell<Features>` is missing
     |
note: function defined here
    --> /usr/local/google/home/chriswailes/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/garando_syntax-0.1.0/src/ext/tt/macro_rules.rs:172:8
     |
172  | pub fn compile(sess: &ParseSess, features: &RefCell<Features>, def: &ast::Item) -> SyntaxExtension {
     |        ^^^^^^^
help: provide the argument
     |
2496 |                 Rc::new(macro_rules::compile(self.parse_sess, /* &RefCell<Features> */, item)),
     |
```

This does not appear to be caused by this PR.

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -2,7 +2,6 @@ use crate::off64_t;
 use crate::prelude::*;
 
 pub type pthread_t = c_ulong;
-pub type __priority_which_t = c_uint;
 pub type __rlimit_resource_t = c_uint;
 pub type Lmid_t = c_long;
 pub type regoff_t = c_int;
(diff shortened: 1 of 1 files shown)


chriswailes:
I apparently looked at the wrong portion of `libc-test/semvar`.  I'm attempting to fix this now.

Issue #4080, opened 2024-11-17
See if `non_exhaustive` would be useful for 1.0


tgross35:
There are a handful of structs that users should be able to access but possibly not construct in case fields change in the future. Currently we often do this by having a single private field, but maybe `#[non_exhaustive]` would be better suited here.

tgross35:
We talked about this at rustweek; we would like to use `#[non_exhaustive]` by default, with some very minor exceptions (e.g. `iovec`)

tgross35:
For how to implement this: the `s!` and `s_no_extra_traits!` macros should be updated to emit `#[non_exhaustive]` by default. Then the macro should accept something like `#[not_non_exhaustive]` to opt out.

dybucc:
Coming from #1324, and seeing how we've had the new usage guidelines for a few
weeks now, is this still necessary? I'm guessing the `Default` impl still is,
but what about the `non_exhaustive` attribute annotation?

dybucc:
If it's worth it, I think I've got the `non_exhaustive` stuff done here [^1].
The `Default` impl I'm still looking into.

Edit: tweaked it some and realized the solution was crap. Everything is still a
WIP.

Edit 2: I finally did get an attribute muncher working on `s`. It should be
fairly simple to extend it to the other `s_*` macro variants.

[^1]: <https://github.com/dybucc/libc/commit/bbe85f40596ea3566dc2c4b5339f1478c411958c>

tgross35:
See https://github.com/rust-lang/libc/pull/5257 for the `Default` bit. I think a `not_non_exhaustive` macro will be easier after that lands because of how it restructures the macro and does some attribute munching already.

Actually using `non_exhaustive` is blocked on https://github.com/rust-lang/rust/issues/132699, but we could add a `__non_exhaustive: ()`.

tgross35:
Trying to think of a reasonable migration path here, maybe something like:

1. Update the macros to accept `#[exhaustive]`. For now `#[exhaustive]` does nothing, otherwise a field `__non_exhaustive: ()` gets added. (If that Rust issue is fixed, we can change it to just be `#[non_exhaustive]`.)
2. In some semi-automated way (your CLI tool, perhaps?) add `#[exhaustive]` to all structs that don't have a private field (i.e. those for which adding `__non_exhaustive` would be technically API-breaking)
3. Merge and backport that
4. Immediately revert the commit adding all the `#[exhaustive]` on the 1.0 branch

[...]

PR #4463, opened 2025-05-26
musl: time64

xbjfk:
# Description
This change includes time64 support for applicable architectures (x86, arm, mips and powerpc). This is based on the previous PRs to this repo as well as the musl changelog from 1.1.24 -> 1.2. It can be enabled with the environment variable 
`RUST_LIBC_UNSTABLE_MUSL_TIME64` only when musl_v1_2_3 is enabled and the architecture is supported. 

A lot of structures, especially ones with mixed endian became excessively complicated, so I used `cfg_if` to separate them. It looks like you can only nest `s! {}` in `cfg_if! {}`, but not vice versa.

As a note, I'm considering removing `musl_not_time64`, and just keeping the old logic of allowing deprecated for function definitions involving `time_t` as it introduces necessary complexity.

When libc 1.0 is released, I believe the best path would be to remove the `musl_v1_2_3` feature, making it unconditionally enabled, keeping `musl_time64` which will expand to `(musl && time64_arch)`.

Tested through QEMU for all architectures.

# Sources
Sources are located on each commit, in the form of upstream commits

# Checklist


xbjfk:
Since #4433 seems close to merging, I can rebase on top of it once merged.

xbjfk:
Rebased on top of the GNU changes :)

build.rs
```diff
@@ -90,10 +94,20 @@ fn main() {
...
+    {
         set_cfg("musl_v1_2_3");
+        if musl_time64 && MUSL_TIME64_ARCHS.contains(&target_arch.as_str()) {
+            set_cfg("musl_time64");
+            set_cfg("linux_time_bits64");
+        } else {
+            set_cfg("musl_not_time64");
+        }
+    } else if target_env == "musl" || target_env == "ohos" {
+        set_cfg("musl_not_time64");
```
tgross35: Why is `musl_not_time64` needed? `not(musl_time64)` should be usable everywhere.

ci/run-docker.sh
```diff
@@ -45,6 +45,8 @@ run() {
         --env LIBC_CI_ZBUILD_STD \
         --env RUST_LIBC_UNSTABLE_GNU_FILE_OFFSET_BITS \
         --env RUST_LIBC_UNSTABLE_GNU_TIME_BITS \
+	--env RUST_LIBC_UNSTABLE_MUSL_V1_2_3 \
+	--env RUST_LIBC_UNSTABLE_MUSL_TIME64 \
         --env CARGO_HOME=/cargo \
         --env CARGO_TARGET_DIR=/checkout/target \
```
tgross35, line 71: Nit: this should match the indentation

build.rs
tgross35, line 1: I think `musl_time64` as it is used should be named something like `musl_time64_on_32`. We probably don't need an env to set this independently, `build.rs` should just set it when `musl_time64` is set and `target_ptr_width` is 32.

src/unix/linux_like/linux/mod.rs
```diff
@@ -2850,8 +2850,14 @@ pub const IPC_NOWAIT: c_int = 0o4000;
 
 pub const IPC_RMID: c_int = 0;
 pub const IPC_SET: c_int = 1;
+#[cfg(musl_time64)]
+pub const IPC_STAT: c_int = 0x102;
+#[cfg(not(musl_time64))]
 pub const IPC_STAT: c_int = 2;
```
tgross35: Style nit: these should be writeable as `pub const IPC_STAT: c_int = if cfg!(musl_time64) { 0x102 } else { 2 }` to get the `else` fallback and reduce the amount of code that is `cfg`ed out on some platforms.

tgross35 (changes requested):
Thanks for working on this! Left a few comments here.

For the style check, are the changes it is suggesting possible? If not, that will probably have to be updated unfortunately (hopefully we will be able to reorganize things at some point so it's less annoying...)

xbjfk:
Thanks for the review! I'll get around to implementing the changes hopefully in the next few days.

build.rs
xbjfk: Yeah, this is kind of a hack to ignore the deprecation warning for musl but not time64. I'll get rid of it by allowing deprecation for any `not(musl_time64)`

build.rs
xbjfk: Sure, I think I will rename it to something like `musl_time64_on_32`, `musl32_time64` or maybe `musl_redir_time64` as you suggested in the previous PR. I don't think there is a practical use for a separate `musl_time64` variable.

src/unix/mod.rs
```diff
@@ -71,10 +71,14 @@ s! {
...
     pub struct timespec {
         pub tv_sec: time_t,
+        #[cfg(all(musl_time64, target_endian = "big"))]
+        __pad0: u32,
         #[cfg(all(target_arch = "x86_64", target_pointer_width = "32"))]
         pub tv_nsec: i64,
         #[cfg(not(all(target_arch = "x86_64", target_pointer_width = "32")))]
         pub tv_nsec: c_long,
+        #[cfg(all(musl_time64, target_endian = "little"))]
+        __pad0: u32,
```
martinetd: Playing with this PR, this makes constructing timespec structs directly as was done in std fail:
```
error: cannot construct `timespec` with struct literal syntax due to private fields
   --> library/std/src/sys/pal/unix/thread.rs:251:30
    |
251 |                 let mut ts = libc::timespec {
    |                              ^^^^^^^^^^^^^^
    |
    = note: ...and other private field `__pad0` that was not provided
```

I'm not quite sure what to suggest here -- making these pub doesn't really make sense, and even if we did I believe we'd need something like `..Default::default()` to make it work so users of libc::timespec (including the compiler) will need to be fixed?... :/

At this build step there are only 4 (as of rust-lang/rust@255aa220821c05c3eac7605fce4ea1c9ab2cbdb4) so this is probably not too bad, but here goes my hopes to see this backported in 0.2 :p
(for reference:
 - library/std/src/sys/pal/unix/thread.rs:251:30
 - library/std/src/sys/pal/unix/time.rs:10:5
 - library/std/src/sys/pal/unix/time.rs:199:14
 - library/std/src/sys/fs/unix.rs:1507:24
)

martinetd: FWIW, just reporting it's working as expected here by working around this issue by making the fields `pub` here and patching std as follow:
<details><summary>rust-lang/rust diff</summary>

```diff
diff --git a/Cargo.lock b/Cargo.lock
index 1a619096d34c..6f91779999af 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2046,8 +2046,6 @@ checksum = "9fa0e2a1fcbe2f6be6c42e342259976206b383122fc152e872795338b5a3f3a7"
 [[package]]
 name = "libc"
 version = "0.2.174"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1171693293099992e19cddea4e8b849964e9846f4acee11b3948bcc337be8776"
 
 [[package]]
 name = "libdbus-sys"
diff --git a/Cargo.toml b/Cargo.toml
index c4d2a06f4cb1..12757101281e 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,3 +89,5 @@ codegen-units = 1
 # FIXME: LTO cannot be enabled for binaries in a workspace
 # <https://github.com/rust-lang/cargo/issues/9330>
 # lto = true
+[patch.crates-io.libc]
+path = "/mnt/rust/libc"
diff --git a/library/Cargo.lock b/library/Cargo.lock
index 34012d6943b7..b112d7ab9ac6 100644
--- a/library/Cargo.lock
+++ b/library/Cargo.lock
@@ -142,8 +142,6 @@ dependencies = [
 [[package]]
 name = "libc"
 version = "0.2.174"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1171693293099992e19cddea4e8b849964e9846f4acee11b3948bcc337be8776"
 dependencies = [
  "rustc-std-workspace-core",
 ]
diff --git a/library/Cargo.toml b/library/Cargo.toml
index 35480b9319d7..955ae17cca35 100644
--- a/library/Cargo.toml
+++ b/library/Cargo.toml
@@ -51,3 +51,5 @@ rustc-std-workspace-core = { path = 'rustc-std-workspace-core' }
 rustc-std-workspace-alloc = { path = 'rustc-std-workspace-alloc' }
 rustc-std-workspace-std = { path = 'rustc-std-workspace-std' }
 compiler_builtins = { path = "compiler-builtins/compiler-builtins" }
+[patch.crates-io.libc]
+path = "/mnt/rust/libc"
diff --git a/library/std/src/sys/fs/unix.rs b/library/std/src/sys/fs/unix.rs
index dc278274f00f..db1fc8a527f0 100644
--- a/library/std/src/sys/fs/unix.rs
+++ b/library/std/src/sys/fs/unix.rs
@@ -1504,7 +1504,12 @@ pub fn set_times(&self, times: FileTimes) -> io::Result<()> {
                 io::ErrorKind::InvalidInput,
                 "timestamp is too small to set as a file time",
             )),
-            None => Ok(libc::timespec { tv_sec: 0, tv_nsec: libc::UTIME_OMIT as _ }),
+            None => Ok(libc::timespec {
+                tv_sec: 0,
+                tv_nsec: libc::UTIME_OMIT as _,
+                #[cfg(target_pointer_width = "32")]
+                __pad0: 0,
+            }),
         };
         cfg_if::cfg_if! {
             if #[cfg(any(target_os = "redox", target_os = "espidf", target_os = "horizon", target_os = "nuttx"))] {
diff --git a/library/std/src/sys/pal/unix/thread.rs b/library/std/src/sys/pal/unix/thread.rs
index d8b189413f4a..20002a4caf16 100644
--- a/library/std/src/sys/pal/unix/thread.rs
+++ b/library/std/src/sys/pal/unix/thread.rs
@@ -251,6 +251,8 @@ pub fn sleep(dur: Duration) {
                 let mut ts = libc::timespec {
                     tv_sec: cmp::min(libc::time_t::MAX as u64, secs) as libc::time_t,
                     tv_nsec: nsecs,
+                    #[cfg(target_pointer_width = "32")]
+                    __pad0: 0,
                 };
                 secs -= ts.tv_sec as u64;
                 let ts_ptr = &raw mut ts;
diff --git a/library/std/src/sys/pal/unix/time.rs b/library/std/src/sys/pal/unix/time.rs
index 0074d7674741..262f2d70fb18 100644
--- a/library/std/src/sys/pal/unix/time.rs
+++ b/library/std/src/sys/pal/unix/time.rs
@@ -6,8 +6,12 @@
 const NSEC_PER_SEC: u64 = 1_000_000_000;
 pub const UNIX_EPOCH: SystemTime = SystemTime { t: Timespec::zero() };
 #[allow(dead_code)] // Used for pthread condvar timeouts
-pub const TIMESPEC_MAX: libc::timespec =
-    libc::timespec { tv_sec: <libc::time_t>::MAX, tv_nsec: 1_000_000_000 - 1 };
+pub const TIMESPEC_MAX: libc::timespec = libc::timespec {
+    tv_sec: <libc::time_t>::MAX,
+    tv_nsec: 1_000_000_000 - 1,
+    #[cfg(target_pointer_width = "32")]
+    __pad0: 0,
+};
 
 // This additional constant is only used when calling
 // `libc::pthread_cond_timedwait`.
@@ -199,6 +203,8 @@ pub fn to_timespec(&self) -> Option<libc::timespec> {
         Some(libc::timespec {
             tv_sec: self.tv_sec.try_into().ok()?,
             tv_nsec: self.tv_nsec.as_inner().try_into().ok()?,
+            #[cfg(target_pointer_width = "32")]
+            __pad0: 0,
         })
     }
```
</details>

made our axum-based server run properly after changing the date past 2038.

(I'd be curious if there's a better way to test this though, perhaps with `-Zbuild-std` ?, but this is good enough for high level testing)

xbjfk: Hmm, yeah - I'm not sure what the best path forward for this is. I believe the GNU changes will also result in a similar issue. I guess I will make the migration process to 1.0 more painful. I think I can make another PR to `rust` in that case. I think consumers will need to use `..Default::default()`. Another problem I can think of is people naively not including that when compiling for 64bit locally breaking things for time64 users.

xbjfk:
Thanks for waiting - I've fixed most of the points, however I haven't done the style check just yet. It also looks like the `cfg!` change broke test compilation - there is a panic in `garando_syntax`.

src/unix/mod.rs
martinetd: > Another problem I can think of is people naively not including that when compiling for 64bit locally breaking things for time64 users.

Yeah, if we go the default way I think we'll want to make `..Default::default()` mandatory -- I 've just tried and it's possible with a zero-length field like PhantomData (...except I guess we can't use that in libc, so we'll need another way to express it)
For the transition we could add the default trait so someone could add `..Default::default()` now even if not required (... but that triggers a clippy warning...)

(Oh, and there's also deconstruction by pattern matching that'll need a ` , ..`...)

At this point if we're breaking API though there might be a better way, implicit cast through a well defined struct with the current fields? ... or maybe there's even a way to specify the physical representation of the fields without adding explicit padding? (but if there is I sure don't know :grin: )

Anyway, yeah, this will probably make the transition to 1.0 a bit painful so it's probably worth spending a bit of time now to figure out the path of least resistance...

xbjfk: I'm not certain, but wouldn't `PhantomData` be fine if it is a member? I know ZST for FFI is generally not allowed - but the C code does not directly access the zero sized type? I just tested it and can see that rust does not report `improper_ctypes` if a struct contains `PhantomData` is used as an `extern "C"` argument - it *does* report it however if the only member of the struct is `PhantomData`.

We should probably also start some migration documentation of 1.0 changes, and how to fix them.

xbjfk:
Okay, I have made a bit more progress. I ended up adding a whitelist for the "positive cfg" error - and whitelisted `musl32_time64` as well as `target_endian`.

The duplicate `s!` macro seems to be erroneously counting top-level/unconditional `s!` macros as the the same as `s!` macros within `cfg_if`. I might try find a way around this some other way if I can't figure out how to resolve the issue.

I have investigated the style crash when using `cfg!` and found that returning `Ok(None)` in `resolve_invoc` in `ctest` [here](https://github.com/rust-lang/libc/blob/90a377b3f48b0c8d29305c5173a9ffb052cbcd67/ctest/src/lib.rs#L1743) causes the issue. Always returning `Err(Determinacy::Determined)` seems to not crash. Looking at [garando](https://github.com/JohnTitor/garando/blob/5db92fce92f61fe0ace4011a39658ff82a2b58c3/garando_syntax/src/ext/expand.rs#L270), it seems to assume - `resolve_invoc` returns either `Err`, `Ok(...)` or be `InvocationKind::Attr`.

@alexcrichton Sorry for the ping but potentially you could help here? I know it's been a while since you wrote the `ctest` code referenced.

If we don't manage to resolve this, I think I will just revert back to separate statements, or maybe using `cfg_if` and adding a fixme.

alexcrichton:
Heh yes sorry at this point I'm no more an expert in ctest than anyone else!

src/unix/mod.rs
xbjfk: I thought about this - I think I will open another PR that will not be stable-nominated specifically for this.

xbjfk:
Pushed changes. Happy to wait for the ctest rewrite while we review :). I might need some help with the duplicate `s!` macro style checker, and how it marks `else` branches as duplicate with `s!` macros not in `cfg_if`.

src/unix/mod.rs
martinetd: My first reaction was that it makes sense to split the thorny part out for more discussion/refining, but thinking back it's probably wrong to use time64 variant functions with a mismatched type (as in will appear to work but fail subtly because ns are not where the callee thinks they are), so I think this has to stay together?
Even if the cost is that this never makes it to stable, I don't see how we could make a "partial time64 support" work properly.

I'll try to play some with struct shaping next week to see if we can do better than relying on default; I've never heard of anything that we could use to shape the struct without affecting the API but it can't hurt to spend a bit more time on it...

xbjfk: Actually, I just realized - we don't need to use `PhantomData` hacks at all - we can just mark it as `#[non_exhaustive]` for all other architectures - this will have essentially the same effect, forcing people to use `..Default::default()` and such. Also, I think the idea is to fully enable time64 support and remove 32 bit time support, at least on `musl`.

You are right regarding the types - I meant to split making `timespec` `non_exhaustive` globally out.

xbjfk: I created #4508 for visibility - let me know if you find a better alternative!

tgross35: In 1.0 we'll be making everything `non_exhaustive`, we more or less want everyone to go via `MaybeUninit::zeroed` and write individual fields like you have to do with C. Don't worry about it here while the flags are unstable.

See https://github.com/rust-lang/libc/issues/4080 for a bit more.

xbjfk:
Rebased on top of main :)

xbjfk:
Hey @tgross35, sorry to ping/nag you - but I would be grateful if you could take a look at this :)

build.rs
```diff
@@ -90,10 +93,15 @@ fn main() {
...
+    println!("cargo:rerun-if-env-changed=RUST_LIBC_UNSTABLE_MUSL_TIME64");
     // loongarch64 and ohos have already updated
-    if musl_v1_2_3 || target_arch == "loongarch64" || target_env == "ohos" {
-        // FIXME(musl): enable time64 api as well
+    if ((musl_v1_2_3 || target_arch == "loongarch64") && target_env == "musl") || target_env == "ohos" {
         set_cfg("musl_v1_2_3");
+        if musl32_time64 && MUSL_TIME64_ARCHS.contains(&target_arch.as_str()) {
+            set_cfg("musl32_time64");
+            set_cfg("linux_time_bits64");
+        }
```
tgross35: (first commit) don't check for architectures explicitly, use `target_pointer_width` but opt out on specific arches if needed (similar to the GNU config)

src/unix/mod.rs
```diff
@@ -71,10 +71,14 @@ s! {
     #[cfg(not(target_env = "gnu"))]
     pub struct timespec {
```
tgross35, line 73: Commit "musl: time64: adjust struct timespec definition": lgtm

src/unix/linux_like/mod.rs
```diff
@@ -13,6 +13,46 @@ missing! {
...
+                pub sched_priority: c_int,
+
+                #[deprecated(since = "0.2.173", note = "This field has been removed upstream and we'll follow that change in a future release.")]
+                pub sched_ss_low_priority: c_int,
+                #[deprecated(since = "0.2.173", note = "This field has been removed upstream and we'll follow that change in a future release.")]
+                pub sched_ss_repl_period: crate::timespec,
+                #[deprecated(since = "0.2.173", note = "This field has been removed upstream and we'll follow that change in a future release.")]
+                pub sched_ss_init_budget: crate::timespec,
+                #[deprecated(since = "0.2.173", note = "This field has been removed upstream and we'll follow that change in a future release.")]
+                pub sched_ss_max_repl: c_int,
```
tgross35: Simplify these messages to just "This field has been removed upstream"

src/unix/linux_like/mod.rs
```diff
@@ -13,6 +13,46 @@ missing! {
...
+        s! {
+            pub struct sched_param {
+                pub sched_priority: c_int,
+
+                __reserved1: c_int,
+                //#[cfg(musl32_time64)]
+                __reserved2: [c_long; 4],
+                //#[cfg(not(musl32_time64))]
+                //__reserved2: [crate::timespec; 2],
+                __reserved3: c_int,
```
tgross35: To match the source, add a `struct __c_anon_sched_param__reserved2 { time_t, long }` for this field and keep the configs.

build.rs
```diff
@@ -24,6 +24,7 @@ const ALLOWED_CFGS: &[&str] = &[
     // Corresponds to `__USE_TIME_BITS64` in UAPI
     "linux_time_bits64",
     "musl_v1_2_3",
+    "musl32_time64",
```
tgross35, line 32: Add a note about `_REDIR_TIME64` similar to the comments above

build.rs
```diff
@@ -44,6 +45,8 @@ const CHECK_CFG_EXTRA: &[(&str, &[&str])] = &[
     ),
 ];
 
+const MUSL_TIME64_ARCHS: &[&str] = &["arm", "mips", "powerpc", "x86"];
```
tgross35: Add a comment

`` /// Musl architectures that set `#define _REDIR_TIME64 1`. ``

I'd also call it `MUSL_REDIR_TIME64_ARCHES` to keep the lingo

build.rs
```diff
@@ -90,10 +93,15 @@ fn main() {
...
     // loongarch64 and ohos have already updated
-    if musl_v1_2_3 || target_arch == "loongarch64" || target_env == "ohos" {
-        // FIXME(musl): enable time64 api as well
+    if ((musl_v1_2_3 || target_arch == "loongarch64") && target_env == "musl") || target_env == "ohos" {
         set_cfg("musl_v1_2_3");
+        if musl32_time64 && MUSL_TIME64_ARCHS.contains(&target_arch.as_str()) {
+            set_cfg("musl32_time64");
+            set_cfg("linux_time_bits64");
+        }
     }
```
tgross35: Some of this is preexisting, but I think this would be cleaner as:

```rust
// OpenHarmony uses a fork of the musl libc
let musl = target_env == "musl" || target_env == "ohos";

// loongarch64 and ohos only exist with recent musl
if target_arch == "loongarch64" || target_env == "ohos" {
    musl_v1_2_3 = true;
}

if musl && musl_v1_2_3 {
    set_cfg("musl_v1_2_3");

    // Set the equivalent of `_REDIR_TIME64`, which is used for 64-bit time_t on 32-bit
    if MUSL_TIME64_ARCHS.contains(...) {
        set_cfg("musl32_time64");
        set_cfg("linux_time_bits64");
    }
}
```

build.rs
```diff
@@ -90,10 +93,15 @@ fn main() {
 
     let musl_v1_2_3 = env::var("RUST_LIBC_UNSTABLE_MUSL_V1_2_3").is_ok();
     println!("cargo:rerun-if-env-changed=RUST_LIBC_UNSTABLE_MUSL_V1_2_3");
+    let musl32_time64 = env::var("RUST_LIBC_UNSTABLE_MUSL_TIME64").is_ok();
+    println!("cargo:rerun-if-env-changed=RUST_LIBC_UNSTABLE_MUSL_TIME64");
```
tgross35: Do we actually need these to be distinct setting? If `musl_v1_2_3` is set then 64-bit time_t is implied on applicable targets. I don't think we really need to account for the 1.2.0..1.2.3 releases.

src/unix/linux_like/linux/musl/b32/arm/mod.rs
```diff
@@ -4,6 +4,11 @@ use crate::prelude::*;
 pub type wchar_t = u32;
 
 s! {
+    pub struct __c_anonymous_timespec32 {
```
tgross35: This doesn't need to be public. I'll change the macros to accept non-pub items in https://github.com/rust-lang/libc/pull/4783.

Also small org nit, put these anon structs after the structs that contain them.

(applies a few places here)

src/unix/linux_like/linux/musl/b32/arm/mod.rs
```diff
@@ -39,12 +59,27 @@ s! {
...
         pub st_blksize: crate::blksize_t,
         pub st_blocks: crate::blkcnt_t,
+        #[cfg(musl32_time64)]
+        __st_atim32: __c_anonymous_timespec32,
+        #[cfg(musl32_time64)]
+        __st_mtim32: __c_anonymous_timespec32,
+        #[cfg(musl32_time64)]
+        __st_ctim32: __c_anonymous_timespec32,
+        #[cfg(musl32_time64)]
+        pub st_ino: crate::ino_t,
```
tgross35: I can't comment on the right lines, but isn't `struct stat64` identical here? If so, I think you could delete the struct and turn it into a type alias.

src/unix/linux_like/linux/mod.rs
```diff
@@ -1,5 +1,8 @@
 //! Linux-specific definitions for linux-like values
 
+use core::cfg;
+use core::mem::size_of;
```
tgross35: `size_of` should be in the prelude. Isn't `cfg` already available?

src/unix/linux_like/linux/musl/b32/mips/mod.rs
```diff
@@ -562,6 +562,13 @@ pub const SIGTSTP: c_int = 24;
...
 pub const SIGIO: c_int = 22;
 pub const SIGSYS: c_int = 12;
+pub const SIGEMT: c_int = 7;
+#[cfg(not(musl_v1_2_3))]
+#[deprecated(
+    since = "0.2.173",
+    note = "This signal actually corresponds to SIGEMT was incorrectly named in musl.
+        As it does not exist on mips it will be dropped entirely in a future release"
+)]
 pub const SIGSTKFLT: c_int = 7;
```
tgross35: Just delete it, mips is tier3

libc-test/build.rs
```diff
@@ -4826,7 +4826,9 @@ fn test_linux(target: &str) {
         // the `xsk_tx_metadata_union` field is an anonymous union
         (struct_ == "xsk_tx_metadata" && field == "xsk_tx_metadata_union") ||
         // After musl 1.2.0, the type becomes `int` instead of `long`.
-        (old_musl && struct_ == "utmpx" && field == "ut_session")
+        (old_musl && struct_ == "utmpx" && field == "ut_session") ||
+        // FIXME(linux): this is changed to separate sec/usec fields when time64 is enabled
+        (struct_ == "input_event" && field == "time")
```
tgross35: This is musl-specific right? If so, it can get `musl &&` as part of the condition.

ci/run-docker.sh
```diff
@@ -45,6 +45,8 @@ run() {
         --env LIBC_CI_ZBUILD_STD \
         --env RUST_LIBC_UNSTABLE_GNU_FILE_OFFSET_BITS \
         --env RUST_LIBC_UNSTABLE_GNU_TIME_BITS \
+        --env RUST_LIBC_UNSTABLE_MUSL_V1_2_3 \
```
tgross35, line 67: Also trying to fix this in https://github.com/rust-lang/libc/pull/4775

libc-test/tests/style_lib/mod.rs
```diff
@@ -38,6 +38,7 @@ use syn::visit::{self, Visit};
 use syn::Token;
 
 const ALLOWED_REPEATED_MACROS: &[&str] = &["s", "s_no_extra_traits", "s_paren"];
+const ALLOWED_POSITIVE_S_CFGS: &[&str] = &["musl32_time64"];
```
tgross35: Would you mind adding the gnu cfgs here too?

src/unix/linux_like/linux/musl/b32/mips/mod.rs
xbjfk: I see that `SIGEMT` was already added in de7e184784a5e7fcacf9c6f90b5cf6331c44cbbd

And `SIGSTKFLT` deleted in bf0ea47:
> For one MIPS architecture, we define it as 7 for unknown reasons,
but 7 is already SIGEMT

Because it was an upstream bug :) https://github.com/bminor/musl/commit/cabc369

tgross35:
Feel free to just add a commit deleting the `handle_s_macro_no_attrs` check, I don't have a good idea how to fix it to not complain in `cfg_if`

src/unix/linux_like/mod.rs
xbjfk: The problem with this (and the reason I kept these commented) is that this will be a breaking change as it will remove access to the `sched_ss_*` fields - but this sounds good if it stays in the 1.0 branch

src/unix/mod.rs
xbjfk: Thanks!

src/unix/linux_like/mod.rs
xbjfk: Done!

build.rs
xbjfk: Yeah, that's a lot cleaner - done!

src/unix/linux_like/mod.rs
tgross35: I'm not sure I follow; why would changing the type of these private fields affect access of public fields under a different cfg?

xbjfk: Oh, sorry - I meant in the larger context as this struct is under the `#[cfg(musl32_time64)]` `cfg_if` - without the config, we have to keep `sched_ss_*` for semver compatibility. But, now that I think of it, I'm not sure why I gated this struct under `musl32_time64`  and not `musl_v1_2_3`.. I will change it and add re-add the configs

build.rs
xbjfk: Yeah, I think you are right - deleting..

src/unix/linux_like/linux/musl/b32/arm/mod.rs
xbjfk: Yeah - I checked the musl source code and I see that `stat64` is defined as `stat` - I'll double check each arch in musl and merge as necessary

xbjfk: Yeah - I checked the musl source code and I see that `stat64` is defined as `stat` - I'll double check each arch in musl and merge as necessary. I noticed that __pad1 and __pad2 are pub in loongarch64, I'm not sure if making these private is acceptably breaking or not. Also noticed that in `riscv64` and `riscv32`, both __pad1 and __pad2 are public in both structs..

ci/run-docker.sh
xbjfk: I see that it's merged - since we aren't using `RUST_LIBC_UNSTABLE_MUSL_TIME64`, I'll just delete the commit

src/unix/linux_like/linux/mod.rs
xbjfk: Yeah, no idea how this got in - deleting..

ci/run-docker.sh
xbjfk: done

src/unix/linux_like/linux/musl/b32/arm/mod.rs
xbjfk: done!

src/unix/linux_like/mod.rs
xbjfk: Oh yeah, I just remembered why I did this - `musl_v1_2_3` will enable on `ohos` as well meaning an ABI break for OHOS. I made the changes but will leave this unresolved for now so we can confirm

src/unix/linux_like/linux/musl/b32/arm/mod.rs
xbjfk: I added a new commit to make those private - if that's not acceptable breakage, let me know and I can mark it deprecated instead

libc-test/tests/style_lib/mod.rs
xbjfk: Done!

libc-test/build.rs
xbjfk: Actually, I don't think it is - not sure why test failures weren't shown on GNU, will need to investigate further..

src/unix/linux_like/linux/mod.rs
xbjfk: Actually, cfg is not available already, it seems like it has to be explicit otherwise build fails..

xbjfk:
Okay, I've addressed most comments - unfortunately, it seems like there is an issue with compiling for hexagon - but it looks like it still appears on `main`.

src/unix/linux_like/mod.rs
```diff
@@ -12,6 +12,53 @@ extern_ty! {
...
+                #[deprecated(since = "0.2.173", note = "This field has been removed upstream")]
+                pub sched_ss_repl_period: crate::timespec,
+                #[deprecated(since = "0.2.173", note = "This field has been removed upstream")]
+                pub sched_ss_init_budget: crate::timespec,
+                #[deprecated(since = "0.2.173", note = "This field has been removed upstream")]
+                pub sched_ss_max_repl: c_int,
+            }
+        }
+    }
+}
```
tgross35: Commit "musl: time64: update struct sched_param": to keep things a bit cleaner, move these musl definitions to the `musl` module. Then keep the existing `struct sched_param` in this module but add the deprecated notes and move it to a `cfg_if` block with the `not(any(musl, emscripten, ohos))` bit.

Also could you put this `cfg_if!` after the main `s!` macro with the others? Around line 252 looking at this file currently.

tgross35: Actually if you wouldn't mind, the musl-specific bits could move directly to the new source structure a `src/new/musl/sched.rs`. Add `pub(crate) mod sched;` to `src/new/musl/mod.rs` and reexport its contents in the musl block at the bottom of `new/mod.rs`

src/unix/linux_like/linux/musl/b32/arm/mod.rs
```diff
@@ -19,15 +19,35 @@ s! {
...
         pub st_mtime: crate::time_t,
         pub st_mtime_nsec: c_long,
+        #[cfg(musl32_time64)]
+        __st_mtim_pad: u32,
         pub st_ctime: crate::time_t,
         pub st_ctime_nsec: c_long,
+        #[cfg(musl32_time64)]
+        __st_ctim_pad: u32,
+        #[cfg(not(musl32_time64))]
         pub st_ino: crate::ino_t,
```
tgross35: Where do these new padding fields come from? They don't seem to be at https://github.com/bminor/musl/blob/1b76ff0767d01df72f692806ee5adee13c67ef88/arch/arm/bits/stat.h

src/unix/linux_like/linux/musl/b32/mips/mod.rs
```diff
@@ -5,6 +5,104 @@ pub type wchar_t = c_int;
 
 pub type stat64 = stat;
 
+cfg_if! {
+    if #[cfg(target_endian = "big")] {
```
tgross35: Instead of using cfg-if on big/little and then configuring on time_t inside, could you configure for `musl32_time64` on the outside and configure big/little on each of the fields? This closer matches source and makes it easier to delete the non-64-bit version when that time comes.

Also, org nit: please put any `cfg_if!` invocations containing `s!` macros after the non-cfg `s!` macro (this is done in most files)

src/unix/linux_like/linux/mod.rs
```diff
@@ -2767,9 +2768,9 @@ pub const IPC_NOWAIT: c_int = 0o4000;
 
 pub const IPC_RMID: c_int = 0;
 pub const IPC_SET: c_int = 1;
-pub const IPC_STAT: c_int = 2;
+pub const IPC_STAT: c_int = if cfg!(musl32_time64) { 0x102 } else { 2 };
 pub const IPC_INFO: c_int = 3;
-pub const MSG_STAT: c_int = 11;
+pub const MSG_STAT: c_int = if cfg!(musl32_time64) { 0x10b } else { 11 };
```
tgross35: I think this can be ` (11 | (IPC_STAT & 0x100))` to match the source https://github.com/bminor/musl/blob/1b76ff0767d01df72f692806ee5adee13c67ef88/include/sys/msg.h#L28

src/unix/linux_like/linux/mod.rs
```diff
@@ -2786,9 +2787,9 @@ pub const GETNCNT: c_int = 14;
 pub const GETZCNT: c_int = 15;
 pub const SETVAL: c_int = 16;
 pub const SETALL: c_int = 17;
-pub const SEM_STAT: c_int = 18;
+pub const SEM_STAT: c_int = if cfg!(musl32_time64) { 0x112 } else { 18 };
 pub const SEM_INFO: c_int = 19;
-pub const SEM_STAT_ANY: c_int = 20;
+pub const SEM_STAT_ANY: c_int = if cfg!(musl32_time64) { 0x114 } else { 20 };
```
tgross35: Ditto, may as well define them in terms of IPC_STAT https://github.com/bminor/musl/blob/1b76ff0767d01df72f692806ee5adee13c67ef88/include/sys/sem.h#L32-L34

libc-test/build.rs
```diff
@@ -4558,7 +4558,9 @@ fn test_linux(target: &str) {
         // `frames` is a flexible array member
         (struct_ == "bcm_msg_head" && field == "frames") ||
         // FAM
-        (struct_ == "af_alg_iv" && field == "iv")
+        (struct_ == "af_alg_iv" && field == "iv") ||
+        // FIXME(linux): this is changed to separate sec/usec fields when time64 is enabled
+        (struct_ == "input_event" && field == "time")
```
tgross35: This isn't defined in musl, why is the skip needed as part of this series? Or if the definition is wrong, shouldn't it be changed?

src/unix/linux_like/linux/musl/b32/arm/mod.rs
xbjfk: the e.g. `st_mtime` and `st_mtime_nsec` fields are defined as `struct timespec` in the source, and `struct timespec` has these padding values in time64. I'm not sure why they are flattened here and not just timespec structs - maybe that's another change for version 1.0

src/unix/linux_like/mod.rs
xbjfk: Done

src/unix/linux_like/linux/musl/b32/mips/mod.rs
xbjfk: Done!

src/unix/linux_like/linux/musl/b64/riscv64/mod.rs
```diff
@@ -22,10 +22,10 @@ s! {
         pub st_uid: crate::uid_t,
         pub st_gid: crate::gid_t,
         pub st_rdev: crate::dev_t,
-        pub __pad1: crate::dev_t,
+        __pad1: crate::dev_t,
         pub st_size: off_t,
         pub st_blksize: crate::blksize_t,
-        pub __pad2: c_int,
+        __pad2: c_int,
```
tgross35: Could you wrap the newly-private `__pad` fields in `Padding<...>`?

src/unix/linux_like/linux/musl/b64/riscv64/mod.rs
```diff
@@ -44,10 +44,10 @@ s! {
         pub st_uid: crate::uid_t,
         pub st_gid: crate::gid_t,
         pub st_rdev: crate::dev_t,
-        pub __pad1: crate::dev_t,
+        __pad1: crate::dev_t,
         pub st_size: off64_t,
         pub st_blksize: crate::blksize_t,
-        pub __pad2: c_int,
+        __pad2: c_int,
```
tgross35: Ditto here

src/unix/mod.rs
```diff
@@ -72,10 +72,14 @@ s! {
...
         pub tv_sec: time_t,
+        #[cfg(all(musl32_time64, target_endian = "big"))]
+        __pad0: u32,
         #[cfg(all(target_arch = "x86_64", target_pointer_width = "32"))]
         pub tv_nsec: i64,
         #[cfg(not(all(target_arch = "x86_64", target_pointer_width = "32")))]
         pub tv_nsec: c_long,
+        #[cfg(all(musl32_time64, target_endian = "little"))]
+        __pad0: u32,
```
tgross35: Same note about padding

src/new/musl/sched.rs
```diff
@@ -0,0 +1,38 @@
...
+            pub struct sched_param {
+                pub sched_priority: c_int,
+
+                __reserved1: c_int,
+                #[cfg(musl32_time64)]
+                __reserved2: [c_long; 4],
+                #[cfg(not(musl32_time64))]
+                __reserved2: [__c_anon_sched_param__reserved2; 2],
+                __reserved3: c_int,
+            }
```
tgross35, line 20: `__reserved` should be wrapped in `Padding` as well

src/unix/linux_like/linux/musl/b32/mips/mod.rs
```diff
@@ -195,7 +195,6 @@ cfg_if! {
                 #[cfg(target_endian = "big")]
                 __msg_ctime_lo: Padding<c_ulong>,
 
-
                 #[cfg(target_endian = "little")]
                 __msg_stime_lo: Padding<c_ulong>,
```
tgross35, line 173: The final commit "libc-test: delete handle_s_macro_no_attrs" seems to apply formatting to `src/unix/linux_like/linux/musl/b32/mips/mod.rs`. Mind moving that bit to the commit where it was changed? ("musl: time64: change time_t type and structs")

src/new/mod.rs
```diff
@@ -59,6 +59,7 @@ cfg_if! {
     } else if #[cfg(target_os = "emscripten")] {
         mod emscripten;
         pub(crate) use emscripten::*;
+        pub use musl::sched::sched_util;
```
tgross35: For the test failure: drop this bit...

src/unix/linux_like/mod.rs
```diff
@@ -109,16 +109,9 @@ s! {
...
         pub sched_priority: c_int,
-        #[cfg(any(target_env = "musl", target_os = "emscripten", target_env = "ohos"))]
-        pub sched_ss_low_priority: c_int,
-        #[cfg(any(target_env = "musl", target_os = "emscripten", target_env = "ohos"))]
-        pub sched_ss_repl_period: crate::timespec,
-        #[cfg(any(target_env = "musl", target_os = "emscripten", target_env = "ohos"))]
-        pub sched_ss_init_budget: crate::timespec,
-        #[cfg(any(target_env = "musl", target_os = "emscripten", target_env = "ohos"))]
-        pub sched_ss_max_repl: c_int,
     }
```
tgross35: ... and restore these fields, but with only `#[cfg(target_os = "emscripten")]` (update the `cfg(not)` of course.

Or create a `src/new/emscripten/sched.rs` file and copy the deprecated version from musl there.

src/unix/linux_like/linux/musl/b32/arm/mod.rs
tgross35: Huh, that's a wtf. Would you mind adding a commit before this one that makes this conditional on `musl_v1_2_3`? So the fields are inline on `not(musl_v1_2_3)` but just use `struct timespec` otherwise. Then this commit can drop that bit.

libc-test/build.rs
tgross35: Could you try dropping this commit and seeing what fails in CI?

This may interact with https://github.com/rust-lang/libc/pull/4654.

tgross35:
Sorry for the conflicts here - more padding changes in https://github.com/rust-lang/libc/pull/4862, and a build config update in https://github.com/rust-lang/libc/commit/5c4255f8259298ac6ad159d1dd3a27484844d454.

src/unix/linux_like/linux/musl/b32/arm/mod.rs
xbjfk: Done - I see it seems to be the same for glibc as well..

src/new/emscripten/sched.rs
```diff
@@ -0,0 +1,38 @@
...
+                #[deprecated(since = "0.2.173", note = "This field has been removed upstream")]
+                pub sched_ss_repl_period: crate::timespec,
+                #[deprecated(since = "0.2.173", note = "This field has been removed upstream")]
+                pub sched_ss_init_budget: crate::timespec,
+                #[deprecated(since = "0.2.173", note = "This field has been removed upstream")]
+                pub sched_ss_max_repl: c_int,
+            }
+        }
+    }
+}
```
tgross35, line 38: This cfg isn't ever set on emscripten right? It just needs the old defs with `deprecated`

martinetd:
Great, thanks for confirming!

I did a build with this branch back in June and it worked fine (at least I didn't notice any weird behavior, and my server didn't break when manually setting the date past 2038), but I'll definitely make time to retest at various parts of the process (soonish on master as a couple of things changed since, and when it gets backported etc), feel free to ping me if you want testing on later PRs and you remember (no harm either way).

PR #4466, opened 2025-06-01
Add even more netfilter constants

vgskye:
# Description

Adds missing nftables definitions, mostly `NFTA_` enums.

# Sources

https://github.com/torvalds/linux/blob/7d4e49a77d9930c69751b9192448fda6ff9100f1/include/uapi/linux/netfilter/nf_tables.h

# Checklist


tgross35:
@vgskye this has been a draft for a while, what is its status?

If you would like to move this forward, I'd prefer for it to be added to the new `src/new` module under `src/new/linux_uapi/linux/netfilter/nf_tables.rs`, following the pattern of others in that directory. Existing definitions that are part of the same enums can move as well. `c_enum!` should be used rather than handwriting the types.

@rustbot author

tgross35:
Closing since this has been inactive for about a year. Please feel free to reopen if you're ready to push it forward!

PR #4467, opened 2025-06-01
Add ctest-next stub and expected dependencies

mbyx:
# Description

As part of setup, I have added a new library called ctest-next for the new version of ctest. Additionally I have already added two dependencies which will most likely be needed immediately after I start, and added the library as a workspace member.

# Checklist


Cargo.toml
```diff
@@ -138,7 +138,7 @@ extra_traits = []
 
 [workspace]
 members = [
-    "ctest",
+    "ctest", "ctest-next",
```
tgross35: Newline please

ctest-next/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "ctest-next"
+version = "0.1.0"
+edition = "2021"
+rust-version = "1.77"
```
tgross35, line 5: This should also get:

```toml
license = "MIT OR Apache-2.0"
repository = "https://github.com/rust-lang/libc"
publish = false
```

ctest-next/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
```
tgross35: This can be deleted

PR #4477, opened 2025-06-04
ctest: Add type alias extraction logic

mbyx:
# Description

This PR adds a helper function for calling cargo expand, as well as the start of constructing a `SymbolTable` that extracts all required items from the codebase (for now only type aliases).

It also has two tests, although more are to be added.

It also adds new dependencies:
- quote for easily testing small snippets of code, as well as converting the parsed ast back into a token stream (and therefore the original code), which is useful for when we don't want to do further parsing but extract the rest as is (for example in a type alias we don't care about whether the type is a pointer or a path, we just want to know what the aliased type is).
- proc_macro2 for using `TokenStream` in struct fields.

# Sources
N/A

# Checklist


tgross35:
Right now the tests aren't very useful, they're just testing syn's ability to roundtrip tokens. If you can add functionality to print `TypeAlias` as a C `typedef` and test that instead, I'd be happy to merge this with `#[expect(dead_code)]` anywhere relevant to pass CI.

mbyx:
Now that I think about it, you're right the tests aren't very useful. For the conversion to C part I want to keep that separate from the extraction logic, so it'll belong to a different struct. I'll add support for extracting all the other types that we need tomorrow, remove the tests for extraction and add a simple stub for a rust to C converter and try to test that.

My intial reasoning for the tests was because locally I wasn't really sure if the visitor was visiting every item regardless of whether it was inside a function, top level etc. This could mean that some symbols would accidentally be missed when extracting and it wouldn't be easy to figure out which ones.

ctest-next/src/ir/constant.rs
```diff
@@ -0,0 +1,37 @@
...
+        &self.ident
+    }
+
+    pub fn ty(&self) -> &TokenStream {
+        &self.ty
+    }
+
+    pub fn value(&self) -> &TokenStream {
+        &self.value
+    }
```
tgross35, line 36: For our public API, we probably don't want to expose `syn`/`pm2` types to give us more flexibility in the future. So for now you can change any `-> TokenStream` methods to `pub(crate)` and mark them `#[expect(unused)]` if needed. and then `ident` should return a `String`. Or also save an `ident_string` field so we can return an `&str`. 

(applies to all the types in `ir`)

ctest-next/src/ir/mod.rs
tgross35, line 1: As a nit, this is more of an `ast` than `ir` - `ast` is just structure, `ir` typically has more information associated with it.

ctest-next/src/skip.rs
```diff
@@ -0,0 +1,39 @@
...
+/// To be used as `SkipItem::Struct((|s: Struct| !s.public()).into())`
+pub enum SkipItem {
+    Constant(Predicate<Constant>),
+    Function(Predicate<Function>),
+    Static(Predicate<Static>),
+    TypeAlias(Predicate<TypeAlias>),
+    Struct(Predicate<Struct>),
+    Field(Predicate<Field>),
+    Union(Predicate<Union>),
+}
```
tgross35, line 39: Maybe this should be reversed; in AST add a `pub enum Item { Constant(Constant), ... }` and then `Predicate` can wrap that.

ctest-next/src/lib.rs
```diff
@@ -1,14 +1,10 @@
-pub fn add(left: usize, right: usize) -> usize {
-    left + right
-}
+#![allow(dead_code)]
```
tgross35, line 1: This will need to be removed before merge. Add `#[expect(dead_code)]` to specific types/functions instead.

ctest-next/src/lib.rs
tgross35, line 1: Could you add `#![warn(unreachable_pub)]` and `#![warn(missing_docs)]`? This will make it a bit more obvious what API is user public vs. crate public. Even a small docstring is helpful for public items so we don't forget, that can be expanded later.

ctest-next/src/symbol_table.rs
```diff
@@ -0,0 +1,190 @@
...
+use quote::ToTokens;
+use syn::visit::Visit;
+
+use crate::ir::{Constant, Field, Function, Parameter, Static, Struct, TypeAlias, Union};
+
+type Abi = String;
+
+/// A `SymbolTable` represents a collected set of top-level Rust items
+/// relevant to FFI generation or analysis, including foreign functions/statics,
+/// type aliases, structs, unions, and constants.
```
tgross35, line 12: Two small nits:

1. Docs don't need to restate the type, rustdoc handles the association for you
2. The first paragraph needs to be 1-2 lines max since it gets used as a summary

So something like this:

```rust
/// Represents a collected set of top-level Rust items relevant to FFI generation or analysis.
///
/// Includes foreign functions/statics, type aliases, structs, unions, and constants.
```

ctest-next/src/symbol_table.rs
```diff
@@ -0,0 +1,190 @@
...
+        self.structs()
+            .iter()
+            .any(|structure| structure.ident().to_string() == ident)
+    }
+
+    pub fn contains_union(&self, ident: &str) -> bool {
+        self.unions()
+            .iter()
+            .any(|structure| structure.ident().to_string() == ident)
+    }
```
tgross35, line 45: `.ident()` without `.to_string()` should be fine here, `Ident` implements `PartialEq` for everything `AsRef<str>` https://docs.rs/proc-macro2/latest/proc_macro2/struct.Ident.html#impl-PartialEq%3CT%3E-for-Ident

ctest-next/src/ir/mod.rs
```diff
@@ -0,0 +1,17 @@
...
+mod union;
+
+pub use constant::Constant;
+pub use field::Field;
+pub use function::Function;
+pub use parameter::Parameter;
+pub use static_variable::Static;
+pub use structure::Struct;
+pub use type_alias::TypeAlias;
+pub use union::Union;
```
tgross35, line 17: Could you keep these names consistent with `syn`'s https://docs.rs/syn/latest/syn/enum.Item.html which are consistent with `rustc`? So `Constant`->`Const`, `Function` -> `Fn`, `TypeAlias` -> `Type`.

I actually prefer the names you have now, but consistency with the ecosystem is useful.

ctest-next/src/symbol_table.rs
```diff
@@ -0,0 +1,190 @@
...
+
+use crate::ir::{Constant, Field, Function, Parameter, Static, Struct, TypeAlias, Union};
+
+type Abi = String;
+
+/// A `SymbolTable` represents a collected set of top-level Rust items
+/// relevant to FFI generation or analysis, including foreign functions/statics,
+/// type aliases, structs, unions, and constants.
+#[derive(Debug)]
+pub struct SymbolTable {
```
tgross35, line 14: Do you have any ideas for a different name than `SymbolTable`? A "symbol table" is typically a list of the exported symbols from a library which doesn't really apply here.

Maybe something like `ItemTable` or `MappableItems` (mappable from C->Rust)?

tgross35 (changes requested):
Left some review for `ir`, I still need to take a look at `translation`.

ctest-next/src/symbol_table.rs
```diff
@@ -0,0 +1,190 @@
...
+
+fn is_visible(vis: &syn::Visibility) -> bool {
+    // Assume that if the visibility is restricted we are not meant to access it.
+    match vis {
+        syn::Visibility::Public(_) => true,
+        syn::Visibility::Inherited | syn::Visibility::Restricted(_) => false,
+    }
+}
+
+fn collect_fields(fields: &syn::punctuated::Punctuated<syn::Field, syn::Token![,]>) -> Vec<Field> {
```
tgross35, line 80: `syn::punctuated::Punctuated` is a long path, you can just import it

ctest-next/src/translation/translator.rs
```diff
@@ -0,0 +1,206 @@
...
+                }
+            }
+            s => self.overrides.type_name(s, &self.table),
+        }
+    }
+
+    pub fn translate_ptr_type(&self, mut ty: &str) -> String {
+        if ty == "&str" {
+            return "char*".to_string();
+        }
```
tgross35, line 82: This isn't correct, `&str` isn't a C-safe type

ctest-next/src/translation/translator.rs
```diff
@@ -0,0 +1,206 @@
...
+            s,
+            "typedef {} {};",
+            // TODO: Add support for more complex types such as crate::c_int.
+            self.translate_ptr_type(&type_alias.ty().to_string()),
+            type_alias.ident().to_string()
+        )?;
+        Ok(s)
+    }
+
+    pub fn translate_primitive_type(&self, ty: &str) -> String {
```
tgross35, line 51: These functions should be taking `syn` types, syn's `&str` roundtrip isn't reliable. I can help with that in a follow up.

mbyx:
I'm a little confused on the last part, do you mean that `simple.rs` and `hierarchy/` are two different fake files used for testing expansion of a single file and expansion of a crate respectively? And `basic.rs` is then the actual test file?

Then in the translation PR `basic.rs` would be updated to also translate the expansion to c and assert that they are the same?

And then the same thing for the skip PR?

ctest-next/src/symbol_table.rs
mbyx: How about `FfiItems`?

ctest-next/src/translation/translator.rs
mbyx: `translate_primitive_type` would have to remain `&str` since `syn` only supports `Path` representation once all other parts of a type have been stripped from it. The other functions and those in the future will take in `syn` types.

tgross35:
@mbyx can this PR be closed? I think everything here has been superseded by now

PR #4478, opened 2025-06-05
Support `renameat2` and `statx` under musl libc

frc4533-lincoln:
# Description

This PR enables `statx` and adds `renameat2` under musl libc.

# Sources

* `renameat2`: https://github.com/bminor/musl/blob/ae3a8c93a663b553e65f096498937083dad210d2/include/stdio.h#L166
* `statx`: https://github.com/bminor/musl/blob/ae3a8c93a663b553e65f096498937083dad210d2/include/sys/stat.h#L108

# Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -98,6 +98,12 @@ jobs:
             artifact-tag: offset-bits64
             env:
               RUST_LIBC_UNSTABLE_GNU_FILE_OFFSET_BITS: 64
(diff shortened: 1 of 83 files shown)


tgross35 (changes requested):
Statx was added in https://github.com/rust-lang/libc/commit/ef3c046a1a839d3eef28d69de6b397e83ce9b94a. Could you drop the statx changes and do something similar with that config for `renameat2`?

PR #4479, opened 2025-06-09
Separate L4Re from Linux code, add aarch64 and enable tests

farao:
The L4Re code was previously attached to the Linux code which was not correct in many ways. This commit separates the L4Re code and enables the libc-tests and includes the fixes for the failing tests. Aarch64 is added as a second supported architecture (more to come).

# Sources

L4Re-adapted version of uclibc: https://github.com/kernkonzept/l4re-core/tree/master/uclibc/lib.

# Checklist


farao:
Hi @tgross35, this is the refactoring of the recent L4Re PR: https://github.com/rust-lang/libc/pull/4383 (which is kept around just in case).

I think the failure of the freebsd nightly checks is not my fault. The rest succeeds now.

I did most of what you requested in your comment on the old PR, I just kept the linux/mod.rs file around since there is a massive part of code that is linux-only and is not supported by L4Re and it did not seem to make sense to put it all in shared.rs.

Also, I put some more code that's shared from the sub modules (emscripten, android, linux, l4re) up in linux_like/mod.rs. In theory, I think there's potential to do that with more code, that's just the one that I would have put in shared.rs but realized that it could go into linux_like/mod.rs instead.

ci/run-docker.sh
tgross35, line 1: I assume this isn't directly related to l4re, mind splitting it to a separate commit? (same PR is fine of course)

libc-test/build.rs
```diff
@@ -3990,6 +4007,10 @@ fn test_linux(target: &str) {
     });
 
     cfg.skip_struct(move |ty| {
+        if ty.starts_with("l4_") {
```
tgross35: Are these anonymous on l4re?

libc-test/build.rs
```diff
@@ -4169,6 +4190,12 @@ fn test_linux(target: &str) {
     });
 
     cfg.skip_const(move |name| {
+        // L4Re requires a min stack size of 64k; that isn't defined in uClibc, but
+        // somewhere in the core libraries. uClibc wants 16k, but that's not enough.
+        if name == "PTHREAD_STACK_MIN" {
```
tgross35: This should be `l4re && ...` right?

libc-test/build.rs
```diff
@@ -3716,11 +3728,12 @@ fn test_linux(target: &str) {
                "libgen.h",
                "limits.h",
                "link.h",
-               "linux/sysctl.h",
+               [uclibc]: "linux/if_ether.h",
+               [linux]: "linux/sysctl.h",
```
tgross35: I think it is probably safer to flip the logic here. That is, rather than `[linux]`, make these `[!l4re]` so any other OS that happens to reuse some of this function gets the same defaults.

libc-test/build.rs
```diff
@@ -4920,7 +4949,9 @@ fn test_linux(target: &str) {
 
     cfg.generate(src_hotfix_dir().join("lib.rs"), "main.rs");
 
-    test_linux_like_apis(target);
+    if linux {
```
tgross35: Similarly, `if !l4re` here

src/unix/linux_like/l4re/uclibc/mod.rs
```diff
@@ -0,0 +1,450 @@
...
+        #[repr(C)]
+        struct siginfo_si_value {
+            _si_signo: c_int,
+            _si_errno: c_int,
+            _si_code: c_int,
+            _si_timerid: c_int,
+            _si_overrun: c_int,
+            si_value: crate::sigval,
+        }
+        (*(self as *const siginfo_t as *const siginfo_si_value)).si_value
```
tgross35: I know this is preexisting, but mind updating to the slightly safer newer casting APIs? So `ptr::from_ref(self).cast::<siginfo_sigfault>()`

src/unix/linux_like/l4re/uclibc/x86_64/mod.rs
```diff
@@ -0,0 +1,415 @@
...
+        pub c_cflag: crate::tcflag_t,
+        pub c_lflag: crate::tcflag_t,
+        pub c_line: crate::cc_t,
+        pub c_cc: [crate::cc_t; crate::NCCS],
+        pub c_ispeed: crate::speed_t,
+        pub c_ospeed: crate::speed_t,
+    }
+
+    // FIXME(1.0): this is actually a union
+    pub struct sem_t {
```
tgross35: ulibc was broken before this right? If so, might as well make it correct here.

src/unix/linux_like/shared.rs
```diff
@@ -0,0 +1,2127 @@
+//! Shared definitions between Linux and L4Re
+
+// TODO: check if there is more overlap with emscripten and android
```
tgross35: `FIXME(linux)` if this comment is meant to be left in the codebase

tgross35 (commented):
Sorry this has taken a while to get to, but thank you for all the changes here! The shape of this one looks much better. I have a handful of small comments but will need to take a deeper look at the big refactor portions again.

farao:
Thanks for the comments! Unfortunately, I'll only be able to get to them beginning of next week but I'll make it a priority then!

ci/run-docker.sh
farao: Done

libc-test/build.rs
farao: Actually turned out to be not necessary. I removed this and similar lines.

libc-test/build.rs
farao: Yes, done.

libc-test/build.rs
farao: True, that's better! I changed it.

libc-test/build.rs
farao: Done

src/unix/linux_like/l4re/uclibc/mod.rs
farao: Done

src/unix/linux_like/l4re/uclibc/x86_64/mod.rs
farao: Yes, that was broken before. I fixed it now.

src/unix/linux_like/shared.rs
farao: I used the FIXME. There are actually a lot of cases where the same struct that is shared between l4re and linux is also shared with one of the other targets (android or emscripten) and one might think about adding all these cases to linux_like/mod.rs instead (with the exception of the differing target as #[cfg..].

src/unix/linux_like/l4re/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+pub const RLIMIT_AS: crate::__rlimit_resource_t = 9;
+pub const RLIMIT_LOCKS: crate::__rlimit_resource_t = 10;
+pub const RLIMIT_SIGPENDING: crate::__rlimit_resource_t = 11;
+pub const RLIMIT_MSGQUEUE: crate::__rlimit_resource_t = 12;
+pub const RLIMIT_NICE: crate::__rlimit_resource_t = 13;
+pub const RLIMIT_RTPRIO: crate::__rlimit_resource_t = 14;
+pub const RLIMIT_RTTIME: crate::__rlimit_resource_t = 15;
+#[allow(deprecated)]
+#[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
+pub const RLIMIT_NLIMITS: crate::__rlimit_resource_t = RLIM_NLIMITS;
```
tgross35: Since this is l4re-only, it can be removed

src/unix/linux_like/mod.rs
```diff
@@ -212,10 +213,91 @@ s! {
...
+        all(
+            not(target_env = "musl"),
+            not(target_env = "ohos"),
+            not(target_env = "uclibc"),
+            target_pointer_width = "64"
+        ),
+        repr(align(8))
+    )]
+    #[cfg(not(target_os = "android"))]
+    pub struct pthread_rwlockattr_t {
```
tgross35: emscripten also needs 4-alignment right? Same for the other structs in this hunk

src/unix/linux_like/mod.rs
```diff
@@ -349,7 +595,7 @@ cfg_if! {
                 u64.hash(state);
             }
         }
-
+    } else if #[cfg(feature = "extra_traits")] {
```
tgross35: This isn't equivalent right? These used to be emitted unconditionally, but now are only if the `all(feature = "extra_traits", not(target_os = "l4re"))` condition above did not match.

src/unix/linux_like/mod.rs
```diff
@@ -448,6 +803,19 @@ cfg_if! {
...
+    size: [0; crate::__SIZEOF_PTHREAD_MUTEX_T],
+};
+#[cfg(not(target_os = "android"))]
+pub const PTHREAD_COND_INITIALIZER: pthread_cond_t = pthread_cond_t {
+    size: [0; crate::__SIZEOF_PTHREAD_COND_T],
+};
+#[cfg(not(any(target_os = "android", target_env = "uclibc")))]
+pub const PTHREAD_RWLOCK_INITIALIZER: pthread_rwlock_t = pthread_rwlock_t {
+    size: [0; crate::__SIZEOF_PTHREAD_RWLOCK_T],
+};
```
tgross35: Weren't these emitted on Android before? Also, looks like they moved up in the file (used to be around 2717)

src/unix/linux_like/shared.rs
```diff
@@ -0,0 +1,2127 @@
+//! Shared definitions between Linux and L4Re
```
tgross35: Maybe call this file `linux_l4re_shared` to match the description?

src/unix/linux_like/shared.rs
```diff
@@ -0,0 +1,2127 @@
...
+
+// FIXME(linux): check if there is more overlap with emscripten and android
+
+use core::mem::size_of;
```
tgross35: Should already come via the prelude

src/unix/linux_like/shared.rs
```diff
@@ -0,0 +1,2127 @@
...
+    }
+
+    pub struct __c_anonymous_ifru_map {
+        pub mem_start: c_ulong,
+        pub mem_end: c_ulong,
+        pub base_addr: c_ushort,
+        pub irq: c_uchar,
+        pub dma: c_uchar,
+        pub port: c_uchar,
+    }
```
tgross35: This was in `s!` rather than `s_no_extra_traits` before right?

src/unix/linux_like/mod.rs
```diff
@@ -434,6 +681,114 @@ cfg_if! {
...
+                    && self.d_off == other.d_off
+                    && self.d_reclen == other.d_reclen
+                    && self.d_type == other.d_type
+                    && self
+                        .d_name
+                        .iter()
+                        .zip(other.d_name.iter())
+                        .all(|(a, b)| a == b)
+            }
+        }
```
tgross35: Why is this trait impl (and the others below it) in a separate module from the structs they go with?

src/unix/linux_like/shared.rs
```diff
@@ -0,0 +1,2127 @@
...
+#[cfg(target_env = "uclibc")]
+const base: crate::nl_item = 768;
+
+pub const ABDAY_1: crate::nl_item = base;
+pub const ABDAY_2: crate::nl_item = base + 0x1;
+pub const ABDAY_3: crate::nl_item = base + 0x2;
+pub const ABDAY_4: crate::nl_item = base + 0x3;
+pub const ABDAY_5: crate::nl_item = base + 0x4;
+pub const ABDAY_6: crate::nl_item = base + 0x5;
+pub const ABDAY_7: crate::nl_item = base + 0x6;
```
tgross35: These were gated on `gnu || musl || ohos` before. Is that not needed anymore?

src/unix/linux_like/l4re/mod.rs
farao: True. I removed the deprecation warning instead, since L4Re's uclibc configuration has the define and I did not find a deprecation warning there.

src/unix/linux_like/mod.rs
farao: Oh, I missed that, thanks. I added it.

src/unix/linux_like/mod.rs
farao: You're right. I split it up into two blocks.

src/unix/linux_like/mod.rs
farao: They are present on android (in b32/mod.rs and b64/mod.rs) but implemented differently, therefore I left them out here. Before, they were in another file (emscripten/mod.rs and linux/mod.rs) afaics.

src/unix/linux_like/shared.rs
farao: Done

src/unix/linux_like/shared.rs
farao: removed

src/unix/linux_like/shared.rs
farao: True, fixed.

src/unix/linux_like/mod.rs
farao: This is because the types of the fields of dirent are different by os (but still they can share the trait implementation). But maybe that's overshooting the mark for code deduplication? Let me know if I shall put them together with their struct implementations again.

src/unix/linux_like/shared.rs
farao: yes, that's not needed anymore. uclibc has them too, just starting at a different number. So I introduced the base variable which is still different between uclibc and not uclibc.

farao:
@tgross35 Btw all the force pushes are no changes to the PR but only the rebases + conflict resolution to the current stat in the main branch. So the PR is almost the same as end of july.

So in the end, all that is missing is a second look that my corrections due to your review requests are ok and as you said this PR would be good to go then.

We are really eagerly waiting for this to be submitted since without it we can't get aarch64 support for L4Re upstream to the main rust repository.

src/unix/linux_like/android/mod.rs
maurer, line 1: Why are we deleting a bunch of types from Android?

I took a brief scan through the rest of the changes, and it doesn't look like they're getting re-imported somewhere else, they're just being deleted?

maurer: My bad, the diff view had chosen not to render the adjustment to `linux_like/mod.rs`, so I didn't see them being added back.

libc-test/build.rs
```diff
@@ -3525,18 +3527,26 @@ fn config_gnu_bits(target: &str, cfg: &mut ctest::TestGenerator) {
...
+    match (linux, gnu, musl, uclibc) {
+        (true, true, false, false) => (),
+        (true, false, true, false) => (),
+        (true, false, false, true) => (),
+        (false, false, false, true) => (),
+        (_, _, _, _) => panic!(
+            "{} target lib is gnu: {gnu}, musl: {musl}, uclibc: {uclibc}",
+            if linux { "linux" } else { "l4re" }
+        ),
     }
```
tgross35: Nit: match `l4re` rather than linux and invert the first field (since it's the default)

libc-test/build.rs
```diff
@@ -3571,6 +3581,9 @@ fn test_linux(target: &str) {
         .define("__GLIBC_USE_DEPRECATED_SCANF", None);
 
     config_gnu_bits(target, &mut cfg);
+    if l4re {
+        cfg.flag("-Wno-unused-function");
+    }
```
tgross35, line 3727: Add a note about why this is special cased

libc-test/build.rs
```diff
@@ -3525,18 +3527,26 @@ fn config_gnu_bits(target: &str, cfg: &mut ctest::TestGenerator) {
 }
 
 fn test_linux(target: &str) {
```
tgross35, line 3665: Could you add a note here about how this covers l4re as well since they have similar API surface?

libc-test/build.rs
```diff
@@ -3994,6 +4014,13 @@ fn test_linux(target: &str) {
 
     cfg.skip_const(move |constant| {
         let name = constant.ident();
+
+        // L4Re requires a min stack size of 64k; that isn't defined in uClibc, but
+        // somewhere in the core libraries. uClibc wants 16k, but that's not enough.
+        if l4re && name == "PTHREAD_STACK_MIN" {
+            return true;
+        }
```
tgross35, line 4159: Restructure this as

```rust
if l4re {
    match name {
        "PTHREAD_STACK_MIN" => true,
        _ => false,
    }
}
```

Which will make it easier to add more in the future if needed

src/unix/linux_like/l4re/uclibc/aarch64/mod.rs
tgross35, line 1: Can aarch64 support be trivially split to a separate commit? If so, that would be ideal (or even a separate PR) since it is orthogonal to the refactoring.

src/unix/mod.rs
```diff
@@ -1803,13 +1817,17 @@ cfg_if! {
...
                 dirfd: c_int,
                 pathname: *const c_char,
                 buf: *mut c_char,
                 bufsiz: size_t,
             ) -> ssize_t;
+            #[cfg(not(target_os = "l4re"))]
             pub fn fmemopen(buf: *mut c_void, size: size_t, mode: *const c_char) -> *mut FILE;
+            #[cfg(not(target_os = "l4re"))]
+            #[cfg(not(target_os = "l4re"))]
             pub fn open_memstream(ptr: *mut *mut c_char, sizeloc: *mut size_t) -> *mut FILE;
```
tgross35, line 1844: Duplicate cfg

src/unix/linux_like/l4re/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+        pub __inheritsched: c_int,
+        pub __scope: c_int,
+        pub __guardsize: size_t,
+        pub __stackaddr_set: c_int,
+        pub __stackaddr: *mut c_void, // better don't use it
+        pub __stacksize: size_t,
+        // L4Re specifics
+        pub affinity: l4_sched_cpu_set_t,
+        pub create_flags: c_uint,
+    }
```
tgross35, line 57: Most of these fields are dunder—should they be nonpublic? We don't have the most consistent record here unfortunately, but it's better to not expose more than we need to if the user isn't expected to look at these fields.

src/unix/linux_like/l4re/uclibc/mod.rs
```diff
@@ -0,0 +1,450 @@
...
+    si_status: c_int,
+    si_utime: c_long,
+    si_stime: c_long,
+}
+impl Copy for sifields_sigchld {}
+impl Clone for sifields_sigchld {
+    fn clone(&self) -> sifields_sigchld {
+        *self
+    }
+}
```
tgross35: `s_no_extra_traits!` (and `s!`) now accepts non-`pub` structs so you don't need to write out these traits. I'll be updating the others that do this soon.

src/unix/linux_like/l4re/uclibc/mod.rs
```diff
@@ -0,0 +1,450 @@
...
+
+cfg_if! {
+    if #[cfg(target_arch = "x86_64")] {
+        mod x86_64;
+        pub use self::x86_64::*;
+    } else if #[cfg(target_arch = "aarch64")] {
+        mod aarch64;
+        pub use self::aarch64::*;
+    } else {
+        pub use unsupported_target;
```
tgross35: Bit of an unusual pattern we seem to have

```suggestion
        // unsupported target
```

src/unix/linux_like/l4re/uclibc/mod.rs
```diff
@@ -0,0 +1,450 @@
...
+pub const O_TMPFILE: c_int = 0o20000000 | O_DIRECTORY;
+pub const PACKET_MR_UNICAST: c_int = 3;
+pub const PF_NFC: c_int = 39;
+pub const PF_VSOCK: c_int = 40;
+pub const RTLD_NOLOAD: c_int = 0x00004;
+pub const RUSAGE_THREAD: c_int = 1;
+pub const SHM_EXEC: c_int = 0o100000;
+pub const SOCK_DCCP: c_int = 6;
+#[deprecated(since = "0.2.70", note = "AF_PACKET must be used instead")]
+pub const SOCK_PACKET: c_int = 10;
```
tgross35: This can just be removed outright

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -0,0 +1,2103 @@
...
+pub type off64_t = i64;
+pub type blkcnt64_t = i64;
+pub type rlim64_t = u64;
+pub type nfds_t = c_ulong;
+pub type nl_item = c_int;
+pub type idtype_t = c_uint;
+pub type loff_t = c_longlong;
+pub type pthread_key_t = c_uint;
+pub type pthread_once_t = c_int;
+pub type pthread_spinlock_t = c_int;
```
tgross35: These fundamental types are fine to keep duplicated in their separate modules. Structs/functions usually make sense to factor out since different platforms try to keep the API the same, but I assume it's more or less coincidence that this subset of types happens to line up.

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -0,0 +1,2103 @@
...
+
+pub type iconv_t = *mut c_void;
+
+cfg_if! {
+    if #[cfg(not(target_env = "gnu"))] {
+        extern_ty! {
+            pub enum fpos64_t {} // FIXME(linux): fill this out with a struct
+        }
+    }
+}
```
tgross35, line 41: For anything that has more specific config, that's fine to leave in Linux and just duplicate in l4re. E.g. I don't expect we will be getting a l4re-sparc target soon, or l4re-gnu.

src/unix/linux_like/mod.rs
tgross35: Yeah, keeping traits with their modules probably makes more sense to keep things robust against added fields or other changes. However, most of these will be going away in https://github.com/rust-lang/libc/pull/4806

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -0,0 +1,2103 @@
...
+        /// Size of buffer
+        pub ifc_len: c_int,
+        pub ifc_ifcu: __c_anonymous_ifc_ifcu,
+    }
+}
+
+#[cfg(not(target_env = "uclibc"))]
+const base: crate::nl_item = 0x20000;
+#[cfg(target_env = "uclibc")]
+const base: crate::nl_item = 768;
```
tgross35: Give this a name like DATE_BASE so its less conflict prone. Also `if cfg!(...)` is preferable over `#[cfg(...)]` where possible, so it gets checked on more targets.

libc-test/build.rs
farao: done

libc-test/build.rs
farao: done

libc-test/build.rs
farao: done

libc-test/build.rs
farao: Unfortunately I can't since this is an early exit (involving a return statement). I could "return true" on the return case and do nothing on the catchall case, but I think it would be more clear code if we just left it as is?

src/unix/linux_like/l4re/uclibc/aarch64/mod.rs
farao: yes, that's true, I put it in a separate commit for now. please tell if you want it in another PR instead.

src/unix/mod.rs
farao: removed

src/unix/linux_like/l4re/mod.rs
farao: yes probably better. I removed the pubs from the prefixed fields

src/unix/linux_like/l4re/uclibc/mod.rs
farao: done

src/unix/linux_like/l4re/uclibc/mod.rs
farao: done

src/unix/linux_like/l4re/uclibc/mod.rs
farao: Hm, it is however still present in our include files. I removed the deprecation notice instead since that's linux specific

src/unix/linux_like/linux_l4re_shared.rs
farao: ok, I put them back into the separate modules!

src/unix/linux_like/linux_l4re_shared.rs
farao: true, but many other configs actually might happen and it seems to me as if this is a good code deduplication anyways since I'd really only copy the whole entry to two places and remove the config in one of them. If it's not all too important to you, I'd prefer to keep them in here?

src/unix/linux_like/linux_l4re_shared.rs
farao: done (both)

src/unix/linux_like/l4re/mod.rs
```diff
@@ -0,0 +1,195 @@
+use crate::prelude::*;
+
+/// L4Re specifics
+/// This module contains definitions required by various L4Re libc backends.
+/// Some of them are formally not part of the libc, but are a dependency of the
+/// libc and hence we should provide them here.
+
+pub type l4_umword_t = c_ulong; // Unsigned machine word.
```
tgross35, line 7: This doc comment is getting associated with the `l4_umword_t` typedef. Switch it to `//!` and put it at the top of the file

src/unix/linux_like/mod.rs
```diff
@@ -234,10 +235,95 @@ s! {
...
+            target_arch = "loongarch64",
+            target_os = "emscripten"
+        )),
+        repr(align(8))
+    )]
+    #[cfg(not(target_os = "android"))]
+    pub struct pthread_mutexattr_t {
+        #[doc(hidden)]
+        size: [u8; crate::__SIZEOF_PTHREAD_MUTEXATTR_T],
+    }
```
tgross35: For this PR, I'd like to avoid commonizing things that aren't already commonized, i.e. pulling them out of android/emscripten up to linux_like (thinking both about conflicts for the backport and changes that aren't simple). Could you restore changes to the android/emscripten files? This means some things will be a bit duplicated in the l4re module, but I'd rather clean it up later than dealing with >2 platforms here.

The `pthread_*` types stand out in particular since they have such a complex definition.

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -0,0 +1,2095 @@
+//! Shared definitions between Linux and L4Re
+
+// FIXME(linux): check if there is more overlap with emscripten and android
```
tgross35: This can be dropped, we're starting to build out the new structure.

src/unix/linux_like/emscripten/mod.rs
```diff
@@ -343,6 +337,12 @@ s! {
         pub d_name: [c_char; 256],
     }
 
+    #[cfg_attr(target_pointer_width = "32", repr(align(4)))]
+    #[cfg_attr(target_pointer_width = "64", repr(align(8)))]
+    pub struct pthread_cond_t {
+        size: [u8; crate::__SIZEOF_PTHREAD_COND_T],
+    }
+
```
tgross35: In the diff, looks like this got moved up from below

src/unix/linux_like/android/mod.rs
```diff
@@ -3649,20 +3643,6 @@ extern "C" {
...
-        buflen: size_t,
-        result: *mut *mut passwd,
-    ) -> c_int;
-    pub fn getpwuid_r(
-        uid: crate::uid_t,
-        pwd: *mut passwd,
-        buf: *mut c_char,
-        buflen: size_t,
-        result: *mut *mut passwd,
-    ) -> c_int;
```
tgross35, line 3617: For these functions and if_nameindex*, I don't mind too much that they get commonized - but could you put any commonization in a separate commit? Since it's distinct from l4re support.

(having orthogonal commits just makes life easier with the backports)

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -0,0 +1,2395 @@
...
+            target_arch = "riscv32",
+            target_arch = "loongarch64",
+            target_os = "emscripten"
+        )),
+        repr(align(8))
+    )]
+    pub struct pthread_mutexattr_t {
+        #[doc(hidden)]
+        size: [u8; crate::__SIZEOF_PTHREAD_MUTEXATTR_T],
+    }
```
tgross35: Don't try to share these pthread_* types between linux and l4re, they're a mess and most of the config isn't relevant for l4re. Just redefine them in the l4re module with something that matches the source headers.

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -0,0 +1,2395 @@
...
+        }
+
+        impl Eq for pthread_barrier_t {}
+
+        impl hash::Hash for pthread_barrier_t {
+            fn hash<H: hash::Hasher>(&self, state: &mut H) {
+                self.size.hash(state);
+            }
+        }
+    }
```
tgross35: These handwritten impls have been removed

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -0,0 +1,2395 @@
...
+        cpusetsize: size_t,
+        cpuset: *mut crate::cpu_set_t,
+    ) -> c_int;
+    pub fn pthread_setaffinity_np(
+        thread: crate::pthread_t,
+        cpusetsize: size_t,
+        cpuset: *const crate::cpu_set_t,
+    ) -> c_int;
+    #[cfg(not(target_env = "uclibc"))]
+    pub fn pthread_setschedprio(native: crate::pthread_t, priority: c_int) -> c_int;
```
tgross35: Is this meant to be unsupported on linux-uclibc also? It seems to be available at https://github.com/kraj/uClibc/blob/ca1c74d67dd115d059a875150e10b8560a9c35a8/libpthread/nptl/sysdeps/pthread/pthread.h#L427-L429

tgross35: Similar for a few other API that is disabled on uclibc, e.g. pthread_mutexattr_getprotocol

farao: Oh, true, sorry, I think that was a mistake. I see it's fixed now that they are in "new"

src/unix/linux_like/linux_l4re_shared.rs
farao: sorry, removed again

src/unix/linux_like/linux_l4re_shared.rs
farao: ok, done

src/unix/linux_like/android/mod.rs
farao: I put them in a separate commit

src/unix/linux_like/emscripten/mod.rs
farao: fixed

src/unix/linux_like/linux_l4re_shared.rs
farao: done

src/unix/linux_like/mod.rs
farao: done

src/unix/linux_like/l4re/mod.rs
farao: done

src/unix/linux_like/linux/gnu/mod.rs
```diff
@@ -645,7 +645,6 @@ pub const BUFSIZ: c_uint = 8192;
 pub const TMP_MAX: c_uint = 238328;
 pub const FOPEN_MAX: c_uint = 16;
 pub const FILENAME_MAX: c_uint = 4096;
-pub const POSIX_MADV_DONTNEED: c_int = 4;
```
tgross35, line 648: Nit: this movement should go in the commonization commit

src/unix/linux_like/linux/mod.rs
```diff
@@ -1133,7 +824,8 @@ s! {
             target_arch = "aarch64",
             target_arch = "riscv64",
             target_arch = "riscv32",
-            target_arch = "loongarch64"
+            target_arch = "loongarch64",
+            target_os = "emscripten"
         ),
         repr(align(4))
```
tgross35, line 827: Why emscripten config being adjusted here? Happens a few other places in this file

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -0,0 +1,2141 @@
...
+                target_arch = "sparc",
+                target_arch = "x86_64",
+                target_arch = "x86"
+            ))
+        ),
+        repr(align(8))
+    )]
+    pub struct pthread_barrier_t {
+        size: [u8; crate::__SIZEOF_PTHREAD_BARRIER_T],
+    }
```
tgross35: My comments at https://github.com/rust-lang/libc/pull/4479#discussion_r2542055478 and https://github.com/rust-lang/libc/pull/4479#discussion_r2544012185 were meant for all `pthread_*` types, not just the specific ones I happened to comment on. Please do not move them into `linux_l4re_shared`: the config is nonsensical and l4re has no need for it.

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -0,0 +1,2141 @@
...
+pub type Elf64_Rel = __c_anonymous_elf64_rel;
+
+cfg_if! {
+    if #[cfg(not(target_arch = "sparc64"))] {
+        pub type Elf32_Rela = __c_anonymous_elf32_rela;
+        pub type Elf64_Rela = __c_anonymous_elf64_rela;
+    }
+}
+
+pub type iconv_t = *mut crate::c_void;
```
tgross35: The addition of `crate::` isn't needed here, you're already importing the prelude

src/unix/linux_like/linux/mod.rs
```diff
@@ -306,18 +214,6 @@ s! {
...
-        #[cfg(all(target_pointer_width = "32", not(target_arch = "x86_64")))]
-        bits: [u32; 32],
-        #[cfg(not(all(target_pointer_width = "32", not(target_arch = "x86_64"))))]
-        bits: [u64; 16],
-    }
-
-    pub struct if_nameindex {
-        pub if_index: c_uint,
-        pub if_name: *mut c_char,
-    }
```
tgross35, line 317: This struct seems to disappear in commit "Separate L4Re from Linux code..." and reappear in "linux_like: communize some structs". The deletion should just move to that last commit so things still pass at this point

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -0,0 +1,2141 @@
...
+        pub ifru_addr: crate::sockaddr,
+        pub ifru_dstaddr: crate::sockaddr,
+        pub ifru_broadaddr: crate::sockaddr,
+        pub ifru_netmask: crate::sockaddr,
+        pub ifru_hwaddr: crate::sockaddr,
+        pub ifru_flags: c_short,
+        pub ifru_ifindex: c_int,
+        pub ifru_metric: c_int,
+        pub ifru_mtu: c_int,
+        pub ifru_map: crate::__c_anonymous_ifru_map,
```
tgross35: Nit: it's in this module, `crate::` isn't needed

src/unix/linux_like/linux/mod.rs
```diff
@@ -1778,531 +1263,19 @@ s_no_extra_traits! {
...
 pub const F_SEAL_FUTURE_WRITE: c_int = 0x0010;
 pub const F_SEAL_EXEC: c_int = 0x0020;
 
 pub const IFF_LOWER_UP: c_int = 0x10000;
 pub const IFF_DORMANT: c_int = 0x20000;
 pub const IFF_ECHO: c_int = 0x40000;
 
+pub const CLONE_CLEAR_SIGHAND: c_ulonglong = 0x100000000;
+pub const CLONE_INTO_CGROUP: c_ulonglong = 0x200000000;
+
```
tgross35: Nit, looks like these moved around but didn't need to

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -0,0 +1,2141 @@
...
+    }
+
+    pub fn CPU_COUNT_S(size: usize, cpuset: &cpu_set_t) -> c_int {
+        let mut s: u32 = 0;
+        let size_of_mask = mem::size_of_val(&cpuset.bits[0]);
+        for i in &cpuset.bits[..(size / size_of_mask)] {
+            s += i.count_ones();
+        }
+        s as c_int
+    }
```
tgross35, line 1545: Nit: this move adds `mem::` to `size_of_val`, but that is no longer needed (it's in the prelude)

src/unix/linux_like/linux/mod.rs
```diff
@@ -6121,70 +4202,14 @@ extern "C" {
...
-        buflen: size_t,
-        result: *mut *mut passwd,
-    ) -> c_int;
-    pub fn getpwuid_r(
-        uid: crate::uid_t,
-        pwd: *mut passwd,
-        buf: *mut c_char,
-        buflen: size_t,
-        result: *mut *mut passwd,
-    ) -> c_int;
```
tgross35, line 6166: Unless I'm missing something, it looks like these disappear without showing up in another module?

tgross35: Ah, it looks like this is another thing that got deleted here then added back in a later commit. That should be avoided, ideally things keep passing between commits.

If you want to put the commonization commit into a separate PR I'm happy to merge that right away. Though of course, the rest of this shouln't be too far behind.

src/unix/linux_like/l4re/mod.rs
```diff
@@ -0,0 +1,194 @@
+///! L4Re specifics
+///! This module contains definitions required by various L4Re libc backends.
+///! Some of them are formally not part of the libc, but are a dependency of the
+///! libc and hence we should provide them here.
+use crate::prelude::*;
+
+pub type l4_umword_t = c_ulong; // Unsigned machine word.
+pub type pthread_t = *mut c_void;
```
tgross35, line 8: Where does this come from? From https://github.com/search?q=repo%3Akernkonzept%2Fl4re-core+%2Ftypedef.*pthread_t%3B%2F&type=code, it seems like they are all `c_ulong`, like upstream uclibc. (Not that it really matters too much.)

src/unix/linux_like/l4re/uclibc/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+    pub struct pthread_condattr_t {
+        #[doc(hidden)]
+        size: [u8; crate::__SIZEOF_PTHREAD_CONDATTR_T],
+    }
+
+    #[repr(align(4))]
+    pub struct pthread_barrierattr_t {
+        #[doc(hidden)]
+        size: [u8; crate::__SIZEOF_PTHREAD_BARRIERATTR_T],
+    }
```
tgross35: I'm looking at the definitions here https://github.com/kernkonzept/l4re-core/blob/e1bf5beeb2933af404b03ca1571e82f33c3385c2/libc/uclibc-ng/libpthread/include/bits/pthreadtypes.h and they don't seem to match up at all - is that the wrong place?

src/unix/linux_like/linux/gnu/mod.rs
farao: done

src/unix/linux_like/linux/mod.rs
farao: ah, that was a leftover of the commonization that I reversed. fixed

src/unix/linux_like/linux_l4re_shared.rs
farao: oh, sorry, missed those because they were in another block. I moved them.

src/unix/linux_like/linux_l4re_shared.rs
farao: fixed

src/unix/linux_like/linux/mod.rs
farao: done

src/unix/linux_like/linux/mod.rs
farao: done

src/unix/linux_like/linux_l4re_shared.rs
farao: done

src/unix/linux_like/linux/mod.rs
farao: fixed it. I think since we're anyways almost done with the overall PR, I'll keep the commit in here.

src/unix/linux_like/l4re/mod.rs
farao: I will try to investigate (since changing it brings me a warning). I'd like to leave it for another PR if that's ok.

src/unix/linux_like/l4re/uclibc/mod.rs
farao: No, the definitions are different. I thought it might have been a concious design decision in the linux part of the code to commonize the pthread structs by just putting an array of bytes into them so I kept this (especially because some time in the near future we might support other libcs). But I'm also happy to choose the actual struct definitions if you prefer this?

maurer (approved):

src/unix/linux_like/l4re/uclibc/mod.rs
tgross35: Oh no it wasn't a deviant decision, that's actually how they're defined in the linux libs https://github.com/bminor/glibc/blob/f9e61cd446d45016e20b6fe85ab87364ebdbec1b/sysdeps/nptl/bits/pthreadtypes.h#L67-L72. I would indeed prefer to match the struct definitions (with all fields private). It's much easier to make sure the fields line up than to add manually compute the size/align, and I don't know that the sizes are likely to always stay the same across platforms anyway.

src/unix/linux_like/linux/mod.rs
```diff
@@ -1521,20 +1144,22 @@ s! {
...
+                target_arch = "x86",
+                target_os = "emscripten",
             ))
         ),
         repr(align(8))
     )]
     pub struct pthread_mutex_t {
         #[doc(hidden)]
-        size: [u8; crate::__SIZEOF_PTHREAD_MUTEX_T],
+        pub size: [c_char; crate::__SIZEOF_PTHREAD_MUTEX_T],
```
tgross35: Please no!

src/unix/linux_like/linux/mod.rs
```diff
@@ -1398,96 +1079,37 @@ s! {
...
-            not(any(target_env = "musl", target_env = "ohos")),
+            not(any(target_env = "musl", target_env = "ohos", target_os = "emscripten")),
             not(target_arch = "x86")
         ),
         repr(align(8))
     )]
     pub struct pthread_cond_t {
         #[doc(hidden)]
-        size: [u8; crate::__SIZEOF_PTHREAD_COND_T],
+        pub size: [u8; crate::__SIZEOF_PTHREAD_COND_T],
```
tgross35: Please no!

src/unix/linux_like/linux/mod.rs
```diff
@@ -1398,96 +1079,37 @@ s! {
...
     )]
     #[cfg_attr(
         all(
-            not(any(target_env = "musl", target_env = "ohos")),
+            not(any(target_env = "musl", target_env = "ohos", target_os = "emscripten")),
             not(target_arch = "x86")
         ),
         repr(align(8))
     )]
     pub struct pthread_cond_t {
```
tgross35, line 1108: Could you please make sure to self-review these pthread diffs? Again this has some changes for emscripten that aren't relevant.

src/unix/linux_like/linux/mod.rs
```diff
@@ -1503,7 +1125,8 @@ s! {
                 target_arch = "powerpc",
                 target_arch = "sparc",
                 target_arch = "x86_64",
-                target_arch = "x86"
+                target_arch = "x86",
+                target_os = "emscripten",
             )
         ),
         repr(align(4))
```
tgross35, line 1129: Ditto, drop this

src/unix/linux_like/linux/mod.rs
```diff
@@ -1521,20 +1144,22 @@ s! {
...
     }
 
     #[cfg_attr(
         all(
             target_pointer_width = "32",
             any(
+                target_os = "emscripten",
                 target_arch = "mips",
                 target_arch = "mips32r6",
                 target_arch = "arm",
```
tgross35, line 1160: Ditto

src/unix/linux_like/linux/mod.rs
```diff
@@ -1553,6 +1178,7 @@ s! {
         any(
             target_pointer_width = "64",
             not(any(
+                target_os = "emscripten",
                 target_arch = "mips",
                 target_arch = "mips32r6",
                 target_arch = "arm",
```
tgross35, line 1178: Ditto

src/unix/linux_like/linux/mod.rs
```diff
@@ -1567,7 +1193,7 @@ s! {
         repr(align(8))
     )]
     pub struct pthread_rwlock_t {
-        size: [u8; crate::__SIZEOF_PTHREAD_RWLOCK_T],
+        pub size: [u8; crate::__SIZEOF_PTHREAD_RWLOCK_T],
     }
 
     #[cfg_attr(
```
tgross35, line 1193: Ditto

tgross35 (changes requested):
Still some leftovers in the pthread definitions, and I would like the uclibc pthread defs to match source rather than using the array. But after this, LGTM!

src/unix/linux_like/l4re/uclibc/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+    pub struct pthread_barrier_t {
+        pub __ba_lock: _pthread_fastlock,
+        pub __ba_required: c_int,
+        pub __ba_present: c_int,
+        pub __ba_waiting: _pthread_descr,
+    }
+
+    pub struct pthread_barrierattr_t {
+        pub __pshared: c_int,
+    }
```
tgross35, line 279: The types are intended to be opaque, so please make the fields private. (that's generally the rule for `__`-prefixed types/fields anyway).

Last thing, after that I'm happy to merge!

PR #4480, opened 2025-06-09
Add SECBIT_ constants from securebits.h

chriswailes:
# Description

This PR adds the `SECBIT_` constants for use with the [`PR_GET_SECUREBITS`](https://github.com/rust-lang/libc/blob/2d96246ca5bff3494cf38124aac25268354ac835/src/unix/linux_like/linux/mod.rs#L3073) and [`PR_SET_SECUREBITS`](https://github.com/rust-lang/libc/blob/2d96246ca5bff3494cf38124aac25268354ac835/src/unix/linux_like/linux/mod.rs#L3074) `prctl` operations.

# Sources

[linux/securebits.h](https://github.com/torvalds/linux/blob/19272b37aa4f83ca52bdf9c16d5d81bdd1354494/include/uapi/linux/securebits.h)

# Checklist


tgross35 (changes requested):
The values look good to me, but in the interest of matching the source code would you mind adding a `const fn issecure_mask` to do the shifts? Something like

```rust
const SECURE_NOROOT: c_int = 0;
pub const SECBIT_NOROOT: c_int = issecure_mask(SECURE_NOROOT);

// ...

const fn issecure_mask(x: c_int) -> c_int {
    1 << x
}
```

chriswailes:
> The values look good to me, but in the interest of matching the source code would you mind adding a `const fn issecure_mask` to do the shifts?

Sure thing.  Consistency with source is always great for debugging and verification.

tgross35 (approved):
Sorry, didn't see the change. Thanks!

PR #4482, opened 2025-06-10
FreeBSD: Support kernel-offloaded TLS TCP

kostikbel:
# Description

The PR adds definitions for kernel-offloaded TLS TCP session information sysctls.  It is used by the utility to list kTLS connections https://github.com/kostikbel/ktcplist

# Sources

https://github.com/freebsd/freebsd-src/blob/4078e0d1d6f8189bff44fcad04df256220035f76/sys/sys/ktls.h
https://github.com/freebsd/freebsd-src/commit/c9e9a0fe5b0f88561f55fb2f6f5354fbbd96dd5d

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -1760,6 +1760,38 @@ s_no_extra_traits! {
         pub inc_fibnum: u16,
         pub inc_ie: crate::in_endpoints,
     }
+
+    pub struct xktls_session_onedir {
+        pub gennum: u64,
+        _rsrv1: [u64; 8],
+        _rsrv2: [u32; 8],
+        pub iv: [u8; 32],
+        pub cipher_algorithm: i32,
+        pub auth_algorithm: i32,
+        pub cipher_key_len: u16,
+        pub iv_len: u16,
+        pub auth_key_len: u16,
+        pub max_frame_len: u16,
+        pub tls_vmajor: u8,
+        pub tls_vminor: u8,
+        pub tls_hlen: u8,
+        pub tls_tlen: u8,
+        pub tls_bs: u8,
+        pub flags: u8,
+        pub drv_st_len: u16,
+        pub ifnet: [u8; 16],
+    }
+
+    pub struct xktls_session {
+        pub tsz: u32,
+        pub fsz: u32,
+        pub inp_gencnt: u64,
+        pub so_pcb: kvaddr_t,
+        pub coninf: crate::in_conninfo,
+        pub rx_vlan_id: c_ushort,
+        pub rcv: crate::xktls_session_onedir,
+        pub snd: crate::xktls_session_onedir,
+    }
 }
 
 cfg_if! {
--- a/src/unix/bsd/freebsdlike/freebsd/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/mod.rs
@@ -11,6 +11,9 @@ pub type fixpt_t = __fixpt_t;
 pub type __lwpid_t = i32;
 pub type lwpid_t = __lwpid_t;
 pub type blksize_t = i32;
+pub type ksize_t = u64;
+pub type inp_gen_t = u64;
+pub type so_gen_t = u64;
 pub type clockid_t = c_int;
 pub type sem_t = _sem;
 pub type timer_t = *mut __c_anonymous__timer;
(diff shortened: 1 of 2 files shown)


kostikbel:
For nightly freebsd-15 ci task to success, you would need to update the ci freebsd env with the new headers.

I have no idea why freebsd-13 is failing.

asomers:
> If there is a version bump available, feel freeto submit a PR updating it.

Our CI config uses the latest FreeBSD 15 snapshot on GCE.  So we don't need to do anything.  If we just wait, GCE will have the latest release soon enough.

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1720,6 +1723,72 @@ s_no_extra_traits! {
...
         __spare__: [c_int; 4],
     }
+
+    pub struct xinpgen {
+        pub xig_len: ksize_t,
+        pub xig_count: u32,
+        pub _xig_spare32: u32,
+        pub xig_gen: inp_gen_t,
+        pub xig_sogen: so_gen_t,
+        pub _xig_spare64: [u64; 4],
```
asomers: The convention used elsewhere in libc is to make spare and padding fields private.  The forces consumers to initialize with `mem::zeroed`, but it also makes it easier for future versions of libc to replace those spare fields with no backwards-compatibility problems.
```suggestion
        _xig_spare32: u32,
        pub xig_gen: inp_gen_t,
        pub xig_sogen: so_gen_t,
        _xig_spare64: [u64; 4],
```

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1720,6 +1723,72 @@ s_no_extra_traits! {
...
+    }
+
+    pub struct in_addr_4in6 {
+        pub ia46_pad32: [u32; 3],
```
asomers:
```suggestion
        ia46_pad32: [u32; 3],
```

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1720,6 +1723,72 @@ s_no_extra_traits! {
...
+        pub ifnet: [u8; 16],
+    }
+
+    pub struct xktls_session {
+        pub tsz: u32,
+        pub fsz: u32,
+        pub inp_gencnt: u64,
+        pub so_pcb: kvaddr_t,
+        pub coninf: crate::in_conninfo,
+        pub rx_vlan_id: c_short,
```
asomers:
```suggestion
        pub rx_vlan_id: c_ushort,
```

asomers (approved):

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -1722,6 +1725,72 @@ s_no_extra_traits! {
         pub uc_flags: c_int,
         __spare__: [c_int; 4],
     }
+
+    pub struct xinpgen {
```
asomers, line 1730: I think this will fix the latest test failure on i386.
```suggestion
    #[repr(align(8)]
    pub struct xinpgen {
```

kostikbel:
> I rebased this to pick up a new set of CI runs since one thing was fixed. The FreeBSD-15 and FreeBSD-13 x86-64 jobs are spurious, but the FreeBSD-13 i686 "bad xinpgen align: rust: 4 (0x4) != c 8 (0x8)" message looks legit. Is one of these fields actually a different size on 32-bit?

Hm, no. The reason for the message is the attribute of the C structure, declaring 8-bytes alignment.
See https://github.com/freebsd/freebsd-src/blob/cd0169c9379c400ec75b77e87ca770e37f964276/sys/netinet/in_pcb.h#L296

Would it be enough to add
#[repr(C, align(8))]``
before the structure?

tgross35:
You don't need to include `C` since that's added by default for everything in the macro, but `#[repr(align(8))]` should work!

src/unix/bsd/freebsdlike/freebsd/mod.rs
kostikbel: Right, can I still update the branch, or should the change be a fixup?

kostikbel:
> Thanks! No need to squash everything to one but if you combine the align fix with whatever added the struct, that would make a bit more clean cherry pick.

Squashed just the fix.

tgross35:
Thanks!

PR #4484, opened 2025-06-10
Add macro expansion and ctest entrypoint.

mbyx:
# Description

This PR adds a function to expand a crate (consisting of either a single file or a folder) using rustc, a stub entrypoint for test generation, and a simple test that checks whether it can expand both types of crates.

# Sources

# Checklist


ctest-next/src/generation/generator.rs
```diff
@@ -0,0 +1,20 @@
+use std::{error::Error, path::Path};
+
+use crate::expand;
+
+pub struct TestGenerator;
```
tgross35: Make this `#[non_exhaustive] struct TestGenerator {}` so building this has to go through the constructor

ctest-next/src/generation/mod.rs
tgross35, line 1: Can you remove the module nesting and just move `src/generation/generator.rs` to `src/generator.rs`? I don't think we're going to have more than one file here.

ctest-next/src/lib.rs
tgross35, line 1: Enforce some rules from the get go

```rust
#![warn(missing_docs)]
#![warn(unreachable_pub)]
```

ctest-next/tests/basic.rs
```diff
@@ -0,0 +1,13 @@
+use std::error::Error;
+
+use ctest_next::TestGenerator;
+
+#[test]
+fn test_entrypoint() -> Result<(), Box<dyn Error>> {
+    let generator = TestGenerator::new();
+
+    generator.generate("./tests/sources/hierarchy/lib.rs", "hierarchy_out.rs")?;
+    generator.generate("./tests/sources/simple.rs", "simple_out.rs")?;
```
tgross35: `?` -> `.unwrap()` so we get the information about which line failed. The test can drop the `-> Result<...>` return type.

Could you also split these test sources into two different `#[test]` functions?

ctest-next/tests/sources/hierarchy/lib.rs
tgross35, line 1:
```rust
//! Ensure that our crate is able to handle definitions spread across many files
```

ctest-next/src/generation/generator.rs
```diff
@@ -0,0 +1,20 @@
...
+impl TestGenerator {
+    pub fn new() -> Self {
+        Self
+    }
+
+    pub fn generate<P: AsRef<Path>>(
+        &self,
+        crate_path: P,
+        _output_file_path: P,
+    ) -> Result<(), Box<dyn Error>> {
```
tgross35: Could you add

```rust
type Error = Box<dyn std::error::Error>;
type Result<T, E = Error> = std::result::Result<T, E>;
```

And then import those? Rather than retyping the definitions everywhere.

tgross35 (changes requested):
Looks pretty good, some small requests. Please also add a test file next to `simple.rs` that puts some code in a macro.

ctest-next/src/error.rs
tgross35, line 1: Anything tiny like this can exist in the parent module, no need for a separate file. It can move if we wind up with more error types but we might not.

ctest-next/tests/sources/macro.rs
```diff
@@ -0,0 +1,13 @@
...
+            Ok(e) => e,
+            Err(e) => panic!("{} failed with {}", stringify!($e), e),
+        }
+    };
+}
+
+pub fn write_buffer() {
+    let buf = String::new();
+    t!(writeln!(buf, "#define A 1"));
+}
```
tgross35: Could the macro generate structs or functions that will be visible in the C interface? Since that is what we'll want to test.

ctest-next/tests/sources/simple.rs
```diff
@@ -0,0 +1,11 @@
+pub type Byte = u8;
+
+pub struct Person {
```
tgross35: `#[repr(C)]` since we'll want to have it exported

ctest-next/tests/sources/macro.rs
tgross35, line 1: Thinking about it some, could you rename the `sources` directory to `input`? Probably makes things a bit more obvious.

tgross35 (commented):
A few remaining nits then please squash and this should be good to merge

tgross35:
CI also needs to be fixed, probably an update here https://github.com/rust-lang/libc/blob/2d96246ca5bff3494cf38124aac25268354ac835/ci/run.sh#L97-L99

mbyx:
The `expand` function seemed to not fail even if the source file had incorrect syntax or could only be partially expanded due to an error, so I fixed that.

tgross35:
> The `expand` function seemed to not fail even if the source file had incorrect syntax or could only be partially expanded due to an error, so I fixed that.

Could you add a test with invalid syntax in `tests/input` and assert that it fails in `basic.rs`?

Please squash the commits once this is all set, everything else lgtm

PR #4485, opened 2025-06-13
Add extraction of relevant types.

mbyx:
# Description

Adds support for extracting relevant types (such as Structs, Unions, Aliases, etc.) from Rust code.

# Sources

# Checklist

### DIFF
--- a/ctest-next/src/ast/constant.rs
+++ b/ctest-next/src/ast/constant.rs
@@ -1,0 +1,41 @@
+/// Represents a constant variable defined in Rust.
+pub struct Const {
+    public: bool,
+    ident: syn::Ident,
+    ty: syn::Type,
+    expr: syn::Expr,
+}
+
+impl Const {
+    /// Creates a new constant.
+    pub fn new(public: bool, ident: syn::Ident, ty: syn::Type, expr: syn::Expr) -> Self {
+        Self {
+            public,
+            ident,
+            ty,
+            expr,
+        }
+    }
+
+    /// Return whether the constant is visible to other crates.
+    pub fn public(&self) -> bool {
+        self.public
+    }
+
+    /// Return the identifier of the constant as a string.
+    pub fn ident(&self) -> String {
+        self.ident.to_string()
+    }
+
+    /// Return the type of the constant.
+    #[expect(unused)]
+    pub(crate) fn ty(&self) -> &syn::Type {
+        &self.ty
+    }
+
+    /// Return the expression stored as a constant.
+    #[expect(unused)]
+    pub(crate) fn expr(&self) -> &syn::Expr {
+        &self.expr
+    }
+}
--- a/ctest-next/src/ast/field.rs
+++ b/ctest-next/src/ast/field.rs
@@ -1,0 +1,29 @@
+/// Represents a field in a struct or union defined in Rust.
+pub struct Field {
+    public: bool,
+    ident: Option<syn::Ident>,
+    ty: syn::Type,
+}
+
+impl Field {
+    /// Creates a new field.
+    pub fn new(public: bool, ident: Option<syn::Ident>, ty: syn::Type) -> Self {
+        Self { public, ident, ty }
+    }
+
+    /// Return whether the field is visible to other crates.
+    pub fn public(&self) -> bool {
+        self.public
+    }
+
+    /// Return the identifier of the field as a string if it exists.
+    pub fn ident(&self) -> Option<String> {
+        self.ident.as_ref().map(|i| i.to_string())
+    }
+
+    /// Return the type of the field.
+    #[expect(unused)]
+    pub(crate) fn ty(&self) -> &syn::Type {
+        &self.ty
+    }
+}
--- a/ctest-next/src/ast/item.rs
+++ b/ctest-next/src/ast/item.rs
@@ -1,0 +1,20 @@
+use crate::{Const, Fn, Static, Struct, Type, Union};
+
+/// Things that can appear directly inside of a module or scope.
+///
+/// This is not an exhaustive list and only contains variants directly useful
+/// for our purposes.
+pub enum Item {
+    /// Represents a constant defined in Rust.
+    Const(Const),
+    /// Represents a function defined in Rust.
+    Fn(Fn),
+    /// Represents a static variable defined in Rust.
+    Static(Static),
+    /// Represents a type alias defined in Rust.
+    Type(Type),
+    /// Represents a struct defined in Rust.
+    Struct(Struct),
+    /// Represents a union defined in Rust.
+    Union(Union),
+}
--- a/ctest-next/src/ast/mod.rs
+++ b/ctest-next/src/ast/mod.rs
@@ -1,0 +1,19 @@
+mod constant;
+mod field;
+mod function;
+mod item;
+mod parameter;
+mod static_variable;
+mod structure;
+mod type_alias;
+mod union;
+
+pub use constant::Const;
+pub use field::Field;
+pub use function::Fn;
+pub use item::Item;
+pub use parameter::Parameter;
+pub use static_variable::Static;
+pub use structure::Struct;
+pub use type_alias::Type;
+pub use union::Union;
--- a/ctest-next/src/ffi_items.rs
+++ b/ctest-next/src/ffi_items.rs
@@ -1,0 +1,209 @@
+use std::{collections::HashMap, ops::Deref};
+
+use syn::{punctuated::Punctuated, visit::Visit};
+
+use crate::{Const, Field, Fn, Parameter, Static, Struct, Type, Union};
+
+/// The ABI as defined by the extern block.
+type Abi = String;
+
+/// Represents a collected set of top-level Rust items relevant to FFI generation or analysis.
+///
+/// Includes foreign functions/statics, type aliases, structs, unions, and constants.
+pub struct FfiItems {
+    aliases: Vec<Type>,
+    structs: Vec<Struct>,
+    unions: Vec<Union>,
+    constants: Vec<Const>,
+    foreign_functions: HashMap<Abi, Vec<Fn>>,
+    foreign_statics: HashMap<Abi, Vec<Static>>,
+}
+
+impl Default for FfiItems {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl FfiItems {
+    /// Creates a new blank FfiItems.
+    pub fn new() -> Self {
+        Self {
+            aliases: Vec::new(),
+            structs: Vec::new(),
+            unions: Vec::new(),
+            constants: Vec::new(),
+            foreign_functions: HashMap::new(),
+            foreign_statics: HashMap::new(),
+        }
+    }
+
+    /// Return whether the type has parsed a struct with the given identifier.
+    pub fn contains_struct(&self, ident: &str) -> bool {
+        self.structs()
+            .iter()
+            .any(|structure| structure.ident() == ident)
+    }
+
+    /// Return whether the type has parsed a union with the given identifier.
+    pub fn contains_union(&self, ident: &str) -> bool {
+        self.unions().iter().any(|union| union.ident() == ident)
+    }
+
+    /// Return a list of all type aliases found.
+    pub fn aliases(&self) -> &Vec<Type> {
+        &self.aliases
+    }
+
+    /// Return a list of all structs found.
+    pub fn structs(&self) -> &Vec<Struct> {
+        &self.structs
+    }
+
+    /// Return a list of all unions found.
+    pub fn unions(&self) -> &Vec<Union> {
+        &self.unions
+    }
+
+    /// Return a list of all constants found.
+    pub fn constants(&self) -> &Vec<Const> {
+        &self.constants
+    }
+
+    /// Return a list of all foreign functions found mapped by their ABI.
+    pub fn foreign_functions(&self) -> &HashMap<Abi, Vec<Fn>> {
+        &self.foreign_functions
+    }
+
+    /// Return a list of all foreign statics found mapped by their ABI.
+    pub fn foreign_statics(&self) -> &HashMap<Abi, Vec<Static>> {
+        &self.foreign_statics
+    }
+}
+
+/// Determine whether an item is visible to other crates.
+///
+/// This function assumes that if the visibility is restricted then it is not
+/// meant to be accessed.
+fn is_visible(vis: &syn::Visibility) -> bool {
+    match vis {
+        syn::Visibility::Public(_) => true,
+        syn::Visibility::Inherited | syn::Visibility::Restricted(_) => false,
+    }
+}
+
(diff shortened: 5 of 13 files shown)


mbyx:
Would it be a good idea to add tests at this point for extraction as well? By just counting up how many aliases, structs, unions etc. `FfiItems` could find in the test inputs and if it matches previous runs.

ctest-next/src/ffi_items.rs
```diff
@@ -0,0 +1,209 @@
...
+/// Represents a collected set of top-level Rust items relevant to FFI generation or analysis.
+///
+/// Includes foreign functions/statics, type aliases, structs, unions, and constants.
+pub struct FfiItems {
+    aliases: Vec<Type>,
+    structs: Vec<Struct>,
+    unions: Vec<Union>,
+    constants: Vec<Const>,
+    foreign_functions: HashMap<Abi, Vec<Fn>>,
+    foreign_statics: HashMap<Abi, Vec<Static>>,
```
tgross35: This is a bit unusual, you don't really need a "mapping" here; update `Fn` to store an `Abi` so the `HashMap` isn't needed. You can filter/retain the vec contents based on ABI when that part is needed.

Also, why does this need to be public?

ctest-next/src/ffi_items.rs
```diff
@@ -0,0 +1,209 @@
+use std::{collections::HashMap, ops::Deref};
+
+use syn::{punctuated::Punctuated, visit::Visit};
+
+use crate::{Const, Field, Fn, Parameter, Static, Struct, Type, Union};
+
+/// The ABI as defined by the extern block.
+type Abi = String;
```
tgross35: So you don't always need to do string comparisons, an enum would probably be good here

```rust
enum Abi {
    C,
    Rust,
    Other(String)
}
```

ctest-next/src/ast/field.rs
```diff
@@ -0,0 +1,29 @@
+/// Represents a field in a struct or union defined in Rust.
+pub struct Field {
+    public: bool,
+    ident: Option<syn::Ident>,
```
tgross35: Is this accounting for tuple structs? If so, I think we can probably ignore those and get rid of the `Option` since they can't be mapped to C anyway.

ctest-next/src/ffi_items.rs
```diff
@@ -0,0 +1,209 @@
...
+pub struct FfiItems {
+    aliases: Vec<Type>,
+    structs: Vec<Struct>,
+    unions: Vec<Union>,
+    constants: Vec<Const>,
+    foreign_functions: HashMap<Abi, Vec<Fn>>,
+    foreign_statics: HashMap<Abi, Vec<Static>>,
+}
+
+impl Default for FfiItems {
```
tgross35: You can probably just derive `Default` here, then call it in new

ctest-next/src/ast/constant.rs
```diff
@@ -0,0 +1,41 @@
+/// Represents a constant variable defined in Rust.
+pub struct Const {
```
tgross35, line 5: For basically every struct (public and not), it would be good to `#[derive(Clone, Debug)]`

ctest-next/src/ast/constant.rs
```diff
@@ -0,0 +1,41 @@
...
+    public: bool,
+    ident: syn::Ident,
```
tgross35: I'm changing my mind a bit here; I think identifiers should be stored as strings rather than `Ident` after all, so the `.ident()` method doesn't need to keep cloning it.

Add a `type BoxStr = Box<str>;` to `lib.rs` and use that for the `ident` fields, you can call `.to_string().into_boxed_str()` from `Ident`.

ctest-next/src/ast/constant.rs
```diff
@@ -0,0 +1,41 @@
...
+    ty: syn::Type,
+    expr: syn::Expr,
+}
+
+impl Const {
+    /// Creates a new constant.
+    pub fn new(public: bool, ident: syn::Ident, ty: syn::Type, expr: syn::Expr) -> Self {
```
tgross35: `new` doesn't need to be crate-public does it?

tgross35: Honestly I think you could probably delete `new` and all accessors except `.ident()` (which I assume will be needed for filters) for all ast types, and make the fields `pub(crate)`. Should be a bit easier to work around, especially when these functions don't do any logic yet.

ctest-next/src/ast/item.rs
tgross35, line 1: Just put this in `ast/mod.rs`, the files don't need to be split up quite so much

tgross35:
> Would it be a good idea to add tests at this point for extraction as well? By just counting up how many aliases, structs, unions etc. `FfiItems` could find in the test inputs and if it matches previous runs.

A test wouldn't be bad. I would just make a single `tests.rs` unit test file and provide an input string of Rust code that includes one of each item, parse that to `syn::parse_file`, get it to a `FfiItems` instance, and assert that the items you expect are in there (just checking the ident is fine).

ctest-next/src/ast/constant.rs
```diff
@@ -0,0 +1,20 @@
...
+    pub(crate) ident: BoxStr,
+    #[expect(unused)]
+    pub(crate) ty: syn::Type,
+    #[expect(unused)]
+    pub(crate) expr: syn::Expr,
+}
+
+impl Const {
+    /// Return the identifier of the constant as a string.
+    pub fn ident(&self) -> &BoxStr {
```
tgross35: Return `&str` rather than `&BoxStr`, just so we don't expose the storage type

ctest-next/src/ffi_items.rs
```diff
@@ -0,0 +1,215 @@
...
+
+use syn::{punctuated::Punctuated, visit::Visit};
+
+use crate::{Abi, Const, Field, Fn, Parameter, Static, Struct, Type, Union};
+
+/// Represents a collected set of top-level Rust items relevant to FFI generation or analysis.
+///
+/// Includes foreign functions/statics, type aliases, structs, unions, and constants.
+#[derive(Default)]
+pub struct FfiItems {
```
tgross35: This struct shouldn't need to be crate-public right? It's only used for internal representations.

ctest-next/src/lib.rs
```diff
@@ -7,13 +7,19 @@
...
+pub use ffi_items::FfiItems;
 pub use generator::TestGenerator;
 pub use macro_expansion::expand;
 
 /// A possible error that can be encountered in our library.
 pub type Error = Box<dyn std::error::Error>;
 /// A type alias for `std::result::Result` that defaults to our error type.
 pub type Result<T, E = Error> = std::result::Result<T, E>;
+/// A boxed string for representing identifiers.
+pub type BoxStr = Box<str>;
```
tgross35: This shouldn't need to be public

ctest-next/src/ffi_items.rs
```diff
@@ -0,0 +1,215 @@
+use std::ops::Deref;
+
+use syn::{punctuated::Punctuated, visit::Visit};
+
+use crate::{Abi, Const, Field, Fn, Parameter, Static, Struct, Type, Union};
+
+/// Represents a collected set of top-level Rust items relevant to FFI generation or analysis.
+///
+/// Includes foreign functions/statics, type aliases, structs, unions, and constants.
+#[derive(Default)]
```
tgross35: + Clone, Debug

ctest-next/src/lib.rs
tgross35, line 1: Could you add `#![warn(missing_debug_implementations)]` so we make sure to catch any new ones?

ctest-next/tests/extraction.rs
```diff
@@ -0,0 +1,19 @@
...
+
+    let mut ffi_items = FfiItems::new();
+    ffi_items.visit_file(&ast);
+
+    assert_eq!(ffi_items.aliases().len(), 1);
+    assert_eq!(ffi_items.constants().len(), 1);
+    assert_eq!(ffi_items.foreign_functions().len(), 1);
+    assert_eq!(ffi_items.foreign_statics().len(), 1);
+    assert_eq!(ffi_items.structs().len(), 1);
+    assert_eq!(ffi_items.unions().len(), 1);
```
tgross35: Don't just check the length since it's possible to pick up the wrong thing. You should be able to do something like

```rust
    assert_eq!(ffi_items.aliases().map(|x| x.ident()).collect::<Vec<_>>(), ["Foo"]);
```

ctest-next/tests/input/all_items.rs
```diff
@@ -0,0 +1,23 @@
+use std::os::raw::c_void;
+
+mod level1 {
```
tgross35: Huh... we're not checking module visibility are we. Could you add it here if it's trivial?

If not don't worry about it, can be fixed later.

ctest-next/src/ffi_items.rs
mbyx: I was expecting it to also be used for filtering, like skipping fields for example.

tgross35: Mind sketching what kind of API you had in mind?

ctest-next/tests/input/all_items.rs
mbyx: Right now visibility for anything isn't being checked, just stored. For module visibility all the non foreign methods in `FfiItems` would have to turn into functions and be called from `visit_item_mod` instead, and the `is_visible` function would be modified. They would also need to be called from `visit_item` as well and duplication removed somehow.

tgross35: Alright, this is fine as-is then

ctest-next/src/ffi_items.rs
mbyx: The TestGenerator would have a method called `skip_item` that takes in a closure that takes in an `Item`, since everything like visibility and field names etc. are available on it there would be no need for any other skip_X methods. That's why `Item` is needed.

It would have a method called `skip_test` that takes in an enum with all tests we do and skips them accordingly.

Then for translation from Rust to C ctest2 originally had fields that stored closures for translating type names. One of these requires knowledge of whether a named type is a struct or not:

```rust
type_name: Box::new(|f, is_struct, is_union| {
    if is_struct {
        format!("struct {}", f)
    } else if is_union {
        format!("union {}", f)
    } else {
        f.to_string()
    }
}),
```

And this requires access to a list of structs and unions that have been extracted (instead of two booleans), so `FfiItems` would be public.

tgross35: We don't need to expose this combined type though, right? We can keep something that is more or less drop in with the existing API, only passing type references rather than always `&str`.

```rust
impl TestGenerator {
    fn skip_struct(&mut self, f: impl Fn(&Struct) -> bool);
    fn skip_field(&mut self, f: impl Fn(&Struct, &Field) -> bool);
    fn skip_fn(&mut self, f: impl Fn(&Function) -> bool);
}
```

mbyx: That is also possible.

ctest-next/src/ast/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+}
+
+/// Things that can appear directly inside of a module or scope.
+///
+/// This is not an exhaustive list and only contains variants directly useful
+/// for our purposes.
+#[derive(Debug, Clone)]
+pub enum Item {
```
tgross35: This also shouldn't be crate-public right?

ctest-next/src/ffi_items.rs
tgross35: As far as how to store it; we'll still need the `Item` / `FfiItems` types but just keep them internal. Something like this:

```rust
/// Inputs needed to rename / skip a field
enum MapInput {
    Struct(Struct),
    Function(Function),
    Field {
        struct_: Struct,
        field: Field,
    }
}

pub struct TestGenerator {
    name_output: Vec<Box<dyn Fn(&MapInput) -> Option<String>>>,
    skips: Vec<Box<dyn Fn(&MapInput) -> bool>>,
}

impl TestGenerator {
    pub fn struct_name(&mut self, f: impl Fn(&Struct) -> String + 'static) {
        self.name_output.push(Box::new(move |item| {
            if let MapInput::Struct(s) = item {
                Some(f(s))
            } else {
                None
            }
        }))
    }

    pub fn union_name(&mut self, f: impl Fn(&Struct) -> String + 'static);

    pub fn skip_struct(&mut self, f: impl Fn(&Struct) -> bool + 'static) {
        self.skips.push(Box::new(move |item| {
            if let MapInput::Struct(s) = item {
                f(s)
            } else {
                false
            }
        }))
    }

    pub fn skip_field(&mut self, f: impl Fn(&Struct, &Field) -> bool + 'static) {
        self.skips.push(Box::new(move |item| {
            if let MapInput::Field { struct_, field } = item {
                f(struct_, field)
            } else {
                false
            }
        }))
    }
}
```

ctest-next/src/ast/mod.rs
mbyx: Should I then make both `FFiItems` and `Items` `pub(crate)` if we follow the design for the API you laid out above?

tgross35: Maybe just do that for now, if we wind up needing to make them public that's easy to do

mbyx: Actually the extraction test currently uses the `FfiItems` directly, since `generate` would eventually do more than is needed for that test (we only need the extraction part there). So should I keep FfiItems `pub` then and only change `Item`?

tgross35: Sorry if I made that confusing at https://github.com/rust-lang/libc/pull/4485#issuecomment-2974810735 - I meant this should be a unit test so it doesn’t need to be public, i.e. `#[cfg(test)] mod tests;` in lib.rs

mbyx:
There seems to be some kind of bug with `#[expect(unused)]` when used with unit tests. When I keep it on a method that is only used in a unit test, it says that the method is actually in use so the lint isn't being upheld. But when I remove the `#[expect(unused)]` it says that the method is not in use.

tgross35:
That's not a bug - the type is indeed used when `cfg(test)` is set, and unused otherwise. So they need to be `cfg_attr(not(test), expect(unused))`.

ctest-next/src/ffi_items.rs
```diff
@@ -0,0 +1,291 @@
...
+                    visit_foreign_item_static(self, static_variable, &abi)
+                }
+                _ => (),
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
tgross35: Could you put the tests in a separate `tests.rs` file? And then move the `all_items.rs` contents to a `const ALL_ITEMS: &str = r#"..."#;` in `tests.rs` so it doesn't need to be a standalone file. You can just skip the `expand` step.

This is the last fix then please squash and LGTM!

tgross35:
Thank you!

PR #4487, opened 2025-06-14
make pidfd_info fields pub

the8472:
# Description

This marks the fields of `pidfd_info` pub.
This seems to have been an oversight, most of the other fields in module are also pub. I also mark it as non_exhaustive since the struct is extensible, new fields have been added recently.

# Sources

https://github.com/torvalds/linux/blob/19272b37aa4f83ca52bdf9c16d5d81bdd1354494/include/uapi/linux/pidfd.h#L109-L110

# Checklist

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -1366,21 +1366,22 @@ s! {
 
     // linux/pidfd.h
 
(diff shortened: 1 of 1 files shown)


the8472:
Considering that this is also hitting freebsd I think those warnings are unrelated.

PR #4488, opened 2025-06-14
Fix CI on FreeBSD 15

asomers:
A recent change has replaced an unused preprocessor symbol: P_UNUSED3

https://github.com/freebsd/freebsd-src/commit/33be1632047c05dbfcc139476e05f49c3a86d560

Another recent change has changed the size of a spare field: mc_spare

https://github.com/freebsd/freebsd-src/commit/eea3e4dd9703a252509d75814049aa8da5007ebb

# Description

# Sources

# Checklist


tgross35:
I still can't explain why ctest seems to work so poorly on FreeBSD, but it seems bimodal. I think we may as well get this in and try to poke things further from there.

PR #4492, opened 2025-06-16
Remove unessecary semicolons from definitions of `CMSG_NXTHDR`.

collinfunk:
I saw `clippy` complain about this on Solaris, and figured I should check the other definitions as well.


tgross35 (approved):
Thanks!

PR #4496, opened 2025-06-16
Add missing kenv function and constants on FreeBSD

GuillaumeGomez:


GuillaumeGomez:
The error doesn't seem related to this PR:

```
  error: function pointer comparisons do not produce meaningful results since their addresses are not guaranteed to be unique
    --> src/unix/linux_like/android/b64/mod.rs:19:9
     |
  19 |         pub sa_restorer: Option<extern "C" fn()>,
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: the address of the same function can vary between different codegen units
     = note: furthermore, different functions could have the same address after being merged together
     = note: for more information visit <https://doc.rust-lang.org/nightly/core/ptr/fn.fn_addr_eq.html>
```

tgross35:
Cc @asomers to review

@GuillaumeGomez could you change the target branch? That issue is fixed on main but it looks like you’re targeting the backport branch.

GuillaumeGomez:
Ah indeed, thanks! Sorry for the noise.

PR #4501, opened 2025-06-17
ctest: Add translation of Rust types.

mbyx:
# Description

The ability to translate various Rust types that are translatable into equivalent C types. Additionally some basic unit tests have been added.

# Sources

# Checklist


mbyx:
Tests seem to fail at a parse int error, so `rustc --version` is giving a different result there for some reason.

ctest-next/templates/test.rs
```diff
@@ -0,0 +1,74 @@
+use std::mem;
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+
+{% if rustc_version < RustcVersion::new(1, 30, 0) %}
```
tgross35: We don't have any need to support down this far (does this even build?) so I think you can strip the `RustcVersion` functionality out

ctest-next/src/translator.rs
```diff
@@ -0,0 +1,207 @@
...
+impl Translator {
+    /// Create a new translator.
+    pub(crate) fn new() -> Self {
+        Self::default()
+    }
+
+    /// Return whether a type is a Rust primitive type.
+    fn is_rust_primitive(&self, ty: &str) -> bool {
+        let rustc_types = [
+            "usize", "u8", "u16", "u32", "u64", "isize", "i8", "i16", "i32", "i64", "f32", "f64",
```
tgross35: i128 and u128 seem to be missing

ctest-next/src/translator.rs
```diff
@@ -0,0 +1,207 @@
...
+pub(crate) struct Translator {}
+
+impl Translator {
+    /// Create a new translator.
+    pub(crate) fn new() -> Self {
+        Self::default()
+    }
+
+    /// Return whether a type is a Rust primitive type.
+    fn is_rust_primitive(&self, ty: &str) -> bool {
```
tgross35: This doesn't need `&self` does it?

ctest-next/src/translator.rs
tgross35, line 1: Could you add a module-level doc comment?

ctest-next/templates/constants/test_constant.c
```diff
@@ -0,0 +1,8 @@
+{% let c_type = Translator::new().translate_type(constant.ty) %}
+{% let ident = constant.ident() %}
+
+static const {{ c_type }} __test_const_{{ ident }}_val = {{ ident }};
+
+const {{ c_type }}* __test_const_{{ ident }}(void) {
+    return &__test_const_{{ ident }}_val;
+}
```
tgross35: A handful of files seem to be missing the trailing newline:

image

ctest-next/src/tests.rs
```diff
@@ -28,64 +28,66 @@ extern "C" {
...
 
-    assert_eq!(
-        ffi_items
-            .foreign_functions()
-            .iter()
-            .map(|a| a.ident())
-            .collect::<Vec<_>>(),
-        ["malloc"]
-    );
+    assert_eq!(translator.translate_type(&ty), "uint8_t");
```
tgross35: This isn't correct right? `u8` is `uint8_t`, `Option<u8>` doesn't have an equivalent

ctest-next/src/translator.rs
```diff
@@ -0,0 +1,207 @@
...
+        if last.ident == "Option" {
+            if let syn::PathArguments::AngleBracketed(p) = &last.arguments {
+                if let syn::GenericArgument::Type(ty) = p.args.first().unwrap() {
+                    self.translate_type(ty)
+                } else {
+                    unimplemented!("Only simple generic types are supported!")
+                }
+            } else {
+                unreachable!("Option<T> cannot have parentheses.")
+            }
```
tgross35: The let-else guard pattern makes these kind of things flatter:

```suggestion
            let syn::PathArguments::AngleBracketed(p) = &last.arguments else {
                unreachable!("Option<T> cannot have parentheses.")
            };
            let syn::GenericArgument::Type(ty) = p.args.first().unwrap()  else {
                unimplemented!("Only simple generic types are supported!");
            };
            
            self.translate_type(ty)
```

ctest-next/src/translator.rs
```diff
@@ -0,0 +1,207 @@
...
+        ty.starts_with("c_") || rustc_types.contains(&ty)
+    }
+
+    /// Translate mutability from Rust to C.
+    fn translate_mut(&self, mutability: Option<syn::Token![mut]>) -> String {
+        mutability.map(|_| "const ").unwrap_or("").to_string()
+    }
+
+    /// Translate a Rust type into its equivalent C type.
+    pub(crate) fn translate_type(&self, ty: &syn::Type) -> String {
```
tgross35: This (along with the other functions here) should probably return a `Result` rather than panicking since there are a lot of possible failures we would like to test on. Maybe just use `thiserror`?

ctest-next/src/tests.rs
tgross35, line 1: Could you try to increase test coverage for translation here? Each function should have a few different asserts and also check that things we can't translate get rejected (which will be easier once fallible functions return a `Result<...>`).

ctest-next/src/tests.rs
```diff
@@ -28,64 +28,66 @@ extern "C" {
...
+        "char*(*)( uint8_t*, int16_t)"
     );
 }
+
+#[test]
+fn test_translation_type_array() {
+    let translator = Translator {};
+    let ty: syn::Type = syn::parse_str("[&u8; 2 + 2]").unwrap();
+
+    assert_eq!(translator.translate_type(&ty), "uint8_t*[2 + 2]");
```
tgross35: Arrays are very tricky to translate correctly. This should also check things like pointers to arrays, pointers to arrays with a size, function signatures with arrays, etc.

See https://github.com/mozilla/cbindgen/issues/171#issuecomment-1772086770 for a bit of context.

ctest-next/src/tests.rs
```diff
@@ -28,64 +28,66 @@ extern "C" {
...
-    );
+    assert_eq!(translator.translate_type(&ty), " int32_t* const*");
+}
+
+#[test]
+fn test_translation_type_reference() {
+    let translator = Translator {};
+    let ty: syn::Type = syn::parse_str("&u8").unwrap();
+
+    assert_eq!(translator.translate_type(&ty), "uint8_t*");
```
tgross35: Also check double references, combined pointers and references, mixed mut and const, etc.

ctest-next/templates/test.rs
```diff
@@ -0,0 +1,74 @@
...
+    if rust != c {
+        eprintln!("bad {}: rust: {} != c {}", attr, rust.pretty(),
+                    c.pretty());
+        FAILED.store(true, Ordering::SeqCst);
+    } else {
+        NTESTS.fetch_add(1, Ordering::SeqCst);
+    }
+}
+
+macro_rules! offset_of {
```
tgross35: `offset_of` is now a builtin macro so this shouldn't be needed

ctest-next/templates/constants/test_constant.c
tgross35, line 1: `test_constant.c` and `test_constant.rs` are tiny, could you just include them in `test.c` and `test.rs`?

ctest-next/templates/test.rs
tgross35, line 1: Could you please comment the template files about what the tests are doing?

ctest-next/build.rs
```diff
@@ -0,0 +1,31 @@
...
+    let host = env::var("HOST").unwrap();
+    let target = env::var("TARGET").unwrap();
+    let target_key = target.replace('-', "_");
+
+    println!("cargo:rustc-env=HOST_PLATFORM={host}");
+    println!("cargo:rustc-env=TARGET_PLATFORM={target}");
+
+    let linker = env::var(format!("CARGO_TARGET_{}_LINKER", target_key.to_uppercase()))
+        .or_else(|_| env::var("CC"))
+        .or_else(|_| env::var(format!("CC_{target_key}")))
```
tgross35: For linker, the env is `LD`. `CC` is just the compiler (don't you need that too though?)

ctest-next/src/generator.rs
```diff
@@ -1,28 +1,159 @@
...
+    }
+
+    /// Configures the target to compile C code for.
+    pub fn target(&mut self, target: &str) -> &mut Self {
+        self.target = Some(target.to_string());
+        self
+    }
+
+    /// Configures the host.
+    pub fn host(&mut self, host: &str) -> &mut Self {
```
tgross35: This isn't in the original `ctest`; why is it needed here? `cc` should be able to figure out the host.

ctest-next/tests/basic.rs
```diff
@@ -1,38 +1,84 @@
...
 fn test_entrypoint_simple() {
-    let mut generator = TestGenerator::new();
+    let crate_path = "tests/input/simple.rs";
 
-    generator
-        .generate("./tests/input/simple.rs", "simple_out.rs")
+    let (mut gen, out_dir) = default_generator(1).unwrap();
+    gen.header("simple.h")
+        .generate(crate_path, "simple_out")
         .unwrap();
```
tgross35: `.header` should be an included header and not an output, right? Where is `simple.h` coming from?

ctest-next/src/generator.rs
```diff
@@ -1,28 +1,159 @@
...
+    pub fn header(&mut self, header: &str) -> &mut Self {
+        self.headers.push(header.to_string());
+        self
+    }
+
+    /// Configures the target to compile C code for.
+    pub fn target(&mut self, target: &str) -> &mut Self {
+        self.target = Some(target.to_string());
+        self
+    }
```
tgross35: Could you update these with the rest of the help from https://docs.rs/ctest/latest/ctest/struct.TestGenerator.html?

ctest-next/src/generator.rs
```diff
@@ -1,28 +1,159 @@
...
-use crate::{expand, ffi_items::FfiItems, Result};
+use crate::{
+    expand,
+    ffi_items::FfiItems,
+    template::{CTestTemplate, RustTestTemplate},
+    Result,
+};
 
 /// A builder used to generate a test suite.
 #[non_exhaustive]
```
tgross35: This can be dropped now that there are nonpublic fields

ctest-next/build.rs
```diff
@@ -0,0 +1,31 @@
...
+        "CARGO_TARGET_{}_RUSTFLAGS",
+        target_key.to_uppercase()
+    ))
+    .unwrap_or_default();
+
+    println!("cargo:rustc-env=LINKER={linker}");
+    println!("cargo:rustc-env=RUNNER={runner}");
+    println!("cargo:rustc-env=FLAGS={flags}");
+
+    println!("cargo:rerun-if-changed-env=TARGET");
```
tgross35: This needs to be rerun if `CC`, `CARGO_TARGET...` and other env changes, correct?

ctest-next/build.rs
```diff
@@ -0,0 +1,31 @@
...
+        env::var(format!("CARGO_TARGET_{}_RUNNER", target_key.to_uppercase())).unwrap_or_default();
+
+    // As we invoke rustc directly this does not get passed to it, although RUSTFLAGS does.
+    let flags = env::var(format!(
+        "CARGO_TARGET_{}_RUSTFLAGS",
+        target_key.to_uppercase()
+    ))
+    .unwrap_or_default();
+
+    println!("cargo:rustc-env=LINKER={linker}");
```
tgross35:
```suggestion
    // Forward build environment variables so they can be used in tests
    println!("cargo:rustc-env=LINKER={linker}");
```

ctest-next/src/tests.rs
tgross35: We need to ensure that we also reject fat pointer types like `&str` and `&[T]`

ctest-next/templates/constants/test_constant.rs
```diff
@@ -0,0 +1,38 @@
...
+{% else %}
+    #[allow(non_snake_case)]
+    fn const_{{ ident }}() {
+        extern "C" {
+            #[allow(non_snake_case)]
+            fn __test_const_{{ ident }}() -> *const {{ ty }};
+        }
+        let val = {{ ident }};
+        unsafe {
+            let ptr1 = &val as *const _ as *const u8;
```
tgross35: Prefer `.cast::<...>()` / `.cast_mut()` / `.cast_const()` over `as` nowadays. So I think `ptr::from_ref(&val).cast::<u8>()`.

ctest-next/templates/test.rs
```diff
@@ -0,0 +1,74 @@
...
+macro_rules! p {
+    ($($i:ident)*) => ($(
+        impl Pretty for $i {
+            fn pretty(&self) -> String {
+                format!("{} ({:#x})", self, self)
+            }
+        }
+    )*)
+}
+p! { i8 i16 i32 i64 u8 u16 u32 u64 usize isize }
```
tgross35: Could we just use `Debug`?

ctest-next/templates/test.rs
```diff
@@ -0,0 +1,74 @@
...
+
+fn same<T: Eq + Pretty>(rust: T, c: T, attr: &str) {
+    if rust != c {
+        eprintln!("bad {}: rust: {} != c {}", attr, rust.pretty(),
+                    c.pretty());
+        FAILED.store(true, Ordering::SeqCst);
+    } else {
+        NTESTS.fetch_add(1, Ordering::SeqCst);
+    }
```
tgross35: `Relaxed` should be fine here right?

ctest-next/templates/test.rs
```diff
@@ -0,0 +1,74 @@
...
+        let offset_pointer = std::ptr::addr_of!((*base_pointer).$field);
+        (offset_pointer as u64) - (base_pointer as u64)
+    })
+}
+
+{% for constant in ffi_items.constants() %}
+    {% include "constants/test_constant.rs" %}
+{% endfor %}
+
+#[inline(never)]
```
tgross35: Why is this `inline(never)`?

ctest-next/tests/basic.rs
tgross35: To clarify: we should be building a `simple.c` file (using the template) and asserting it has the same contents as `tests/input/simple.c`. Or, if `CTEST_BLESS=1` is present, update the file instead. That way we see what the generated output looks like.

ctest-next/templates/test.rs
```diff
@@ -0,0 +1,74 @@
...
+        impl Pretty for $i {
+            fn pretty(&self) -> String {
+                format!("{} ({:#x})", self, self)
+            }
+        }
+    )*)
+}
+p! { i8 i16 i32 i64 u8 u16 u32 u64 usize isize }
+
+fn same<T: Eq + Pretty>(rust: T, c: T, attr: &str) {
```
tgross35: This should be something like `ensure_same` or `check_same` just to hint that it updates the test counters. Also we can get by with PartialEq rather than Eq right?

ctest-next/templates/constants/test_constant.rs
```diff
@@ -0,0 +1,38 @@
...
+            let ptr2 = __test_const_{{ ident }}() as *const u8;
+            for i in 0..mem::size_of::<{{ ty }}>() {
+                let i = i as isize;
+                same(*ptr1.offset(i), *ptr2.offset(i),
+                        &format!("{{ ident }} value at byte {}", i));
```
tgross35: Why go byte by byte instead of creating a slice?

ctest-next/src/runner.rs
```diff
@@ -0,0 +1,83 @@
...
+use std::env;
+use std::fs::{canonicalize, File};
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+/// Compiles a Rust source file and links it against a static library.
+///
+/// Returns the path to the generated binary.
+pub fn compile_test<P: AsRef<Path>>(
```
tgross35: The user of this crate will never call `compile_test` or `run_test` right? This is just for the internal implementation and tests.

If so, giving them a `__` name and marking them `#[doc(hidden)]` would be good.

ctest-next/src/generator.rs
```diff
@@ -1,28 +1,159 @@
...
+        let target = self
+            .target
+            .clone()
+            .unwrap_or(env::var("TARGET_PLATFORM").unwrap());
+        let host = self
+            .host
+            .clone()
+            .unwrap_or(env::var("HOST_PLATFORM").unwrap());
+
+        let mut cfg = cc::Build::new();
```
tgross35: Could this function move to `runner.rs` so the steps to build Rust and steps to build C are next to each other?

ctest-next/src/tests.rs
mbyx: This seems to be what ctest2 did, so I included it, although I don't think it's used in either `ctest-test` or `libc`. 
Screenshot 2025-06-26 101358

ctest-next/build.rs
mbyx: Doesn't the compiler also act as the linker if given the correct arguments? I didn't think of it that much but passing whatever is set for those three variables as the linker worked (and in the CI it's the compiler not the linker).

ctest-next/src/generator.rs
mbyx: Before ctest only worked inside of `build.rs` which set the `HOST` env var for us, but `ctest-next` forwards those from the `build.rs` instead so it has to be manually set for cc inside `generate`. I thought that it would be better to also have the option to manually set it.

ctest-next/tests/basic.rs
mbyx: But isn't that different from the old `ctest`? Here it already has a manually written header file for `t1.c`, it uses `t1.rs` to generate `t1gen.c` and `t1gen.rs`, `t1gen.c` includes the header file `t1.h` and uses it to define test functions we call in `t1gen.rs` via ffi. Then finally it would build `bin/t1.rs` that includes the original `t1.rs` as well as the generated one, to run the tests.
So the generated C file just holds tests to be ran and uses the header file to do it. 
Screenshot 2025-06-26 101358

ctest-next/src/generator.rs
mbyx: Do you mean the documentation? I chose not to add the examples since they would be different and will probably change until `ctest-next` is a bit more finished. The other methods in the generator I was planning to add in a different PR.

ctest-next/templates/constants/test_constant.rs
mbyx: The template is taken directly from the original, and that part specifically I don't really understand. If it has to be changed to use slices could you please tell me how?

ctest-next/src/runner.rs
mbyx: They don't necessarily have to use them but they can if the right environment variables are set. It removes the need to have binary files in `src/bin/` since that is generated and compiled automatically, and also makes it easier to test.

However it's definitely quite buggy at the moment with minor things causing it to fail, so I think it's best if we keep it internal for now and then when it's more robust it can be made public.

ctest-next/src/generator.rs
mbyx: It's heavily dependent on `TestGenerator` for configuration, converting it to a function would increase the number of parameters. Do you mean moving the impl for that method to runner.rs?

ctest-next/src/tests.rs
mbyx: I don't understand why we need to reject `&str`, `ctest` seems to support it and `ctest-test` also has a test variable that is `&str` that does get translated as a define.
Screenshot 2025-06-26 101358

ctest-next/src/translator.rs
mbyx: They don't have an equivalent in C (that is supported by all compilers) so they can't be translated. Should I make them give an error instead of being ignored?

ctest-next/src/tests.rs
tgross35: Wah! Yeah that's not good. I'm glad you made this noticeable in a test.

Any use in `libc` would probably flag the `improper_ctypes_definitions` lint, but ctest should reject it anyway (also anything else that triggers that lint)

ctest-next/src/tests.rs
tgross35: As a bit of a rhetorical question: what makes up a `&str`? For part of that., what does `size_of::<&str>()` print, and what does `sizeof(char*)` print in C?

ctest-next/src/tests.rs
tgross35: Side note, please copy+paste code or permalink it (when you highlight lines on the GH UI, it gives a permalink option under the three dots) rather than screenshotting, both here and on Zulip. Keeps things searchable and copy+pasteable.

ctest-next/src/generator.rs
tgross35: Sorry, yeah I just meant the docs for these two functions, "The generate C test..." and "Note that for Cargo builds...". Actually though, I think you _could_ add tests here already! They are marked `no_run` in `ctest` so the output doesn't actually matter, and I don't think the interface here should change.

ctest-next/src/generator.rs
tgross35: Could it just be a function that takes `&mut TestGenerator` rather than `&self`?

ctest-next/src/translator.rs
tgross35: It should be possible to turn them into (`unsigned`) `__int128` and just let the C compiler decide whether or not they are supported. They are used in some libc interfaces, but only on platforms where `__int128` exists.

ctest-next/build.rs
tgross35: Often they are but not necessarily, that's the difference between the `CC` and `LD` variables. The `CARGO_*_LINKER` env doesn't actually need to build anything (it's only used for `rustc`'s final link) so somebody could pass `ld` rather than `gcc`. At least, that _should_ work but the flags do need to line up.

So you should just reexport all variables starting with `CC` for the `cc` crate (probably just iterate the env and export everything starting with `CC`) and then use the `CARGO_*_LINKER` var to the `rustc` invocation with `-Clinker=...`.

ctest-next/tests/basic.rs
tgross35: Ah, you are correct. In that case we should have both `t1.h` which is handwritten, and `t1.c` which is the autogenerated file that gets updated if `CTEST_BLESS` is set.

ctest-next/build.rs
mbyx: In every Dockerfile, one of those three environment variables is set for the linker and it's always referencing the compiler, which is why it is like that.
image

ctest-next/templates/test.rs
tgross35: Ah, I see - it's using the hex impl for integers. Honestly I would just drop `Pretty` and split `same` into `same` (existing) and a new `same_hex`  that takes `T: LowerHex`.

ctest-next/templates/constants/test_constant.rs
tgross35: You can construct a slice with something like

```rust
slice::from_raw_parts(ptr::from_ref(&val).cast::<u8>(), size_of::<{{ ty }}>());
```

And then call `same_hex` (if you do https://github.com/rust-lang/libc/pull/4501/files/4241857e47e53fdd102d326f66bb7abb9ecca452#r2168478216, otherwise implement `Pretty`).

tgross35: Reading this a bit closer, there is actually a chance we are reading uninitialized data here which is spooky UB. So this should get a comment like:

```
// HACK: This may read uninitialized data! We do this because there isn't a good way to
// recursively iterate all fields.
```

We should probably try to fix this later, but I don't know how.

ctest-next/build.rs
tgross35: Right, in our cases we use the compiler for both. But we don't want it to blow up if this isn't the case of course, so we shouldn't assume that they will always be the same.

That snip actually has a good point though, the `AR_*` variables also need to be forwarded.

mbyx: Would they need to be? The ENV vars are set before it runs and they are not unset by `cc`. `TARGET` and `HOST` are only set for the duration of build.rs so they have to be forwarded. The linker and runner are forwarded there for convenience.

From it's docs:
"In addition to the above optional environment variables, cc-rs has some functions with hard requirements on some variables supplied by [cargo’s build-script driver](https://doc.rust-lang.org/cargo/reference/build-scripts.html#inputs-to-the-build-script) that it has the TARGET, OUT_DIR, OPT_LEVEL, and HOST variables."

So those 4 are the only ones that really have to be forwarded, linker and runner are just for convenience. In one of the previous CI runs for the powerpc64 architecture, I remember the CC_target variable was set in the debug output of `cc` despite `ctest` not setting it because it was already set by `ENV`.

tgross35: Fair enough, you're right that we don't actually need to manually do all of these. Anything `CARGO_*` does need to be extracted because we need to pass it to `rustc` (for linking) or the runner (for qemu) since we're not going via Cargo. But the rest we can just let `cc` figure out.

ctest-next/src/tests.rs
mbyx: Oh I understand, &str also stores the size so they aren't equivalent. However the tests never convert an &str into a char*, instead they convert char* to a CStr and then an &str. The translation is still needed to generate the function that returns the pointer to the char* however.

tgross35: Where exactly would this come into play; something like `char *c = "foo";` matching `const c: &str = "foo"`? You could check them for equality but still wouldn't be quite right, the 1:1 equivalent would be something like `const A: *const std::ffi::c_char = c"abc".as_ptr()`.

mbyx: From the `ctest-test` crate, `const T1S: &str = "Hello"` matching `#define T1S "Hello"`. But honestly what you reccomended is better, I think it was done previously because it was quicker to do, since I don't really see any reason to support it.

So fat pointers would be rejected, and their alternatives closer to the C type would be used in tests.

tgross35: C string literals are pretty new so that's probably part of it, it used to be more tedious. We should probably be rejecting anything that causes `improper_ctypes_definitions` ([includes `&str`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=11bbd95242f00b5dc91c1ef09ecd1176)) for now at least. (Not like actually checking it, but just using that as a guideline.

Note that i128 and u128 used to raise that lint but don't anymore, since pretty recently https://github.com/rust-lang/rust/pull/137306

ctest-next/tests/basic.rs
tgross35: ^ I forgot we also build a Rust file; that should get blessed as well. I guess if the test file is `basic.rs`, name the output file `basic.out.rs`. Maybe also `basic.out.c` rather than `basic.c` for some consistency?

tgross35:
Have you been able to run any cross build+test locally by the way? Thinking about things a bit more, I'm not sure things will work with the current setup (building in integration tests). The biggest problem is that on cross compiled targets it will be trying to run host tools from within the runner: e.g. when you're running an integration test in `qemu-system-arm`, it's going to try to launch `rustc` and `cc` which don't exist, or somehow try to use the host's which are for the wrong target. That probably explains the loongarch failure.

Unfortunately I think this might mean we have to use the original `ctest`'s method of using a separate test crate, running `ctest` in `build.rs`. You're welcome to try to get things working, but I expect the cross-arch targets might not be possible without quite a bit of hacking.

For this PR: we need to have the "bless" tests that show what the generated `.c` and `.rs` files look like (https://github.com/rust-lang/libc/pull/4501#discussion_r2168432601) but you don't need to worry about actually compiling the tests here. (do check that the generated files build+run without errors locally, though).

tgross35:
> Running the test doesn't invoke `rustc` or `cc` though

Doesn't it? The `#[test]` functions call `.generate(..)`, which definitely seems to call `cc`.

mbyx:
I'm not very familiar with how Rust syntax should be translated to C, so the unit tests were made according to my understanding, if the translation is wrong I'll change the parser. Similarly I made a bunch of changes to `templates/test.rs` that I have no idea if they're correct (they do work for the test case I added in simple.rs).

ctest-next/Cargo.toml
```diff
@@ -8,5 +8,12 @@ repository = "https://github.com/rust-lang/libc"
...
 [dependencies]
+askama = "0.14.0"
 cc = "1.2.25"
+quote = "1.0.40"
 syn = { version = "2.0.101", features = ["full", "visit", "extra-traits"] }
+thiserror = "2.0.12"
+
+[dev-dependencies]
+tempfile = "3.20.0"
+
```
tgross35:
```suggestion
```

Spurious newline

ctest-next/askama.toml
```diff
@@ -0,0 +1,3 @@
+[[escaper]]
+path = "askama::filters::Text"
+extensions = ["rs", "c", "cpp"]
```
tgross35, line 3: I'm not too familiar with the config, what does this actually do?

ctest-next/build.rs
```diff
@@ -0,0 +1,55 @@
...
+// Similarly FLAGS and RUNNER are also not an issue, if CARGO_TARGET_{}_RUSTFLAGS are present they're forwarded.
+// And RUSTFLAGS works by default anyway. Similarly the test binary doesn't require any external applications so just
+// the RUNNER is enough to run it.
+//
+// However since rustc and cc are the host versions, they will only work if we specify the correct variables for them.
+// Because we only use them to compile, not run things.
+// For CC we MUST specify CC or CC_target otherwise it will fail. (Other flags like AR etc. work without forwarding because it is run in the host.)
+// For rustc we MUST specify the correct linker. Usually this is the same as CC or CC_target.
+//
+// In the CI, the CARGO_TARGET_{} variables are always set.
```
tgross35: Small nit, could comments always be wrapped to 100? Most people editing Rust have their editors set up for that.

ctest-next/src/runner.rs
```diff
@@ -0,0 +1,160 @@
+use crate::{Result, TestGenerator};
+use std::env;
+use std::fs::{canonicalize, File};
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+/// Generate all tests for the given crate and output the Rust side to a file.
+pub fn generate_test(
```
tgross35, line 10: We should leave this `#[doc(hidden)]` as well

ctest-next/src/generator.rs
tgross35: It seems like this isn't used for tests or anything currently, so let's just remove it for now. Easy to add back once/if we need it.

ctest-next/src/generator.rs
```diff
@@ -1,28 +1,107 @@
...
+            .clone()
+            .unwrap_or_else(|| env::var("OUT_DIR").unwrap().into());
+        let output_file_path = output_directory.join(output_file_path);
+
+        // Generate the Rust side of the tests.
+        File::create(output_file_path.with_extension("rs"))?
+            .write_all(RustTestTemplate::new(&ffi_items)?.render()?.as_bytes())?;
+
+        // Generate the C side of the tests.
+        // FIXME: Cpp not supported yet.
```
tgross35:
```suggestion
        // FIXME(ctest): Cpp not supported yet.
```

We try to keep all FIXMEs in this repo tagged for searching, since we have a lot

ctest-next/src/runner.rs
```diff
@@ -0,0 +1,160 @@
...
+    let target = generator.target.clone().unwrap_or_else(|| {
+        env::var("TARGET").unwrap_or_else(|_| env::var("TARGET_PLATFORM").unwrap())
+    });
+    let host = generator
+        .host
+        .clone()
+        .unwrap_or_else(|| env::var("HOST").unwrap_or_else(|_| env::var("HOST_PLATFORM").unwrap()));
+
+    let mut cfg = cc::Build::new();
+    // FIXME: Cpp not supported.
```
tgross35: FIXME(ctest)

ctest-next/src/runner.rs
```diff
@@ -0,0 +1,160 @@
...
+#[doc(hidden)]
+pub fn __run_test<P: AsRef<Path>>(test_binary: P) -> Result<String> {
+    let runner = env::var("RUNNER").unwrap_or_default();
+    let output = if runner.is_empty() {
+        Command::new(test_binary.as_ref()).output()?
+    } else {
+        let mut args = runner.split_whitespace();
+        let mut cmd = Command::new(args.next().unwrap());
+        cmd.args(args).arg(test_binary.as_ref()).output()?
+    };
```
tgross35:
```suggestion
    let mut cmd;
    if runner.is_empty() {
        cmd = Command::new(test_binary.as_ref())
    } else {
        let mut args = runner.split_whitespace();
        cmd = Command::new(args.next().unwrap());
        cmd.args(args)
    };
   
    cmd.arg(test_binary.as_ref());
    eprintln!("command: {cmd:?}");
    let .. = cmd.output()?;
```

ctest-next/src/runner.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    if !output.status.success() {
+        return Err(std::str::from_utf8(&output.stderr)?.into());
+    }
+
+    // The template prints to stderr regardless.
+    Ok(std::str::from_utf8(&output.stderr)?.to_string())
```
tgross35: This comes from a different area of code of course, but why does it print to `stderr` rather than `stdout`?

ctest-next/build.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    let linker = env::var(format!("CARGO_TARGET_{target_key}_LINKER")).unwrap_or_default();
+    let runner = env::var(format!("CARGO_TARGET_{target_key}_RUNNER")).unwrap_or_default();
+    // As we invoke rustc directly this does not get passed to it, although RUSTFLAGS does.
+    let flags = env::var(format!("CARGO_TARGET_{target_key}_RUSTFLAGS")).unwrap_or_default();
+
+    // Forward build environment variables so they can be used in tests.
+    println!("cargo:rustc-env=LINKER={linker}");
+    println!("cargo:rustc-env=RUNNER={runner}");
+    println!("cargo:rustc-env=FLAGS={flags}");
```
tgross35: These should probably only reemit if they are set in the first place. So flip to:

```rust
let link_var = format!("CARGO_TARGET_{target_key}_LINKER");
println!("cargo:rerun-if-changed-env={link_var}");
if let Some(linker) = env::var(link_var) {
    println!("cargo:rustc-env=LINKER={linker}");
};
```

(note that it's good practice to keep rerun-if-changed next to wherever the vars get used)

ctest-next/templates/test.rs
```diff
@@ -0,0 +1,106 @@
...
+
+/// Check that the value returned from the Rust and C side in a certain test is equivalent.
+///
+/// Internally it will remember which checks failed and how many tests have been run. This
+/// method is the same as `check_same` but prints errors in bytes in hex.
+fn check_same_hex<T: PartialEq + LowerHex + Debug>(rust: T, c: T, attr: &str) {
+    if rust != c {
+        let rust = format!("{:?} ({:#x})", rust, rust);
+        let c = format!("{:?} ({:#x})", c, c);
+        eprintln!("bad {}: rust: {} != c {}", attr, rust, c);
```
tgross35: Clippy warns about noninlined format strings now, might as well just fold the strings in too

```suggestion
        eprintln!("bad {attr}: rust: {rust:?} ({rust:#x}) != c {c:?} ({c:#x})");
```

ctest-next/src/translator.rs
```diff
@@ -0,0 +1,319 @@
...
+use std::ops::Deref;
+
+use quote::ToTokens;
+use thiserror::Error;
+
+/// Errors that can occur during the translation of a type.
+#[derive(Debug, Error, PartialEq, Eq)]
+pub enum TranslationError {
+    /// The provided type is unknown or unrecognized.
+    #[error("{0}: unknown type.")]
```
tgross35: Since everything takes a string, that can be factored out. I think we may as well take a span here too so we can point at the origin of the problem:

```rust
struct TranslationError {
    kind: TranslationErrorKind,
    source: String,
    // if this isn't always available, I guess make it `Option`?
    source_span: Span,
}

impl TranslationError {
    pub(crate) fn new(kind: TranslationErrorKind, source: &str, source_span: Span) -> Self {
        // ...
    }
}

impl fmt::Display for TranslationError {
    fn fmt(f, ...) {
        write!(
            f, "{}: `{}`\nat {}:{}"
            self.kind,
            self.source,
            // You will need to add the crate `proc_macro2` with the `span-locations` and
            // `procmacro2_semver_exempt` features (fine since we're not using the original
            // `proc_macro` crate`), which is what `syn`'s `Span` types comes from
            self.span.file,
            self.span.start.line,
            self.span.start.column,
        )
    }
}

enum TranslationErrorKind {
     // Same as the existing `TranslationError`, just drop `String`
}
```

ctest-next/src/translator.rs
```diff
@@ -0,0 +1,319 @@
...
+                if inner_type.contains("(*)") {
+                    Ok(inner_type
+                        .replacen("(*)", &format!("(*{modifier})"), 1)
+                        .trim()
+                        .to_string())
+                } else {
+                    Ok(format!("{inner_type} {modifier}*").trim().to_string())
+                }
+            }
+            _ => Err(TranslationError::UnknownType(
```
tgross35: Based on how this error gets emitted, `UnsupportedType` seems like a more accurate name than `UnknownType` (since we might know what it is but just can't support it)

ctest-next/src/translator.rs
```diff
@@ -0,0 +1,319 @@
...
+            }
+            _ => {
+                let inner_type = self.translate_type(inner)?;
+                Ok(format!("{inner_type} {modifier}*"))
+            }
+        }
+    }
+
+    /// Translate a simple Rust expression to C.
+    #[expect(clippy::only_used_in_recursion)]
```
tgross35: Huh, what does it actually complain about here?

Also, when do we need to translate expressions?

ctest-next/templates/test.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    // Test that the string constant is the same in both Rust and C.
+    // While fat pointers can't be translated, we instead of * const c_char.
+    {% if ty == "* const c_char" %}
+        #[inline(never)]
+        #[allow(non_snake_case)]
+        fn const_{{ ident }}() {
+            extern "C" {
+                #[allow(non_snake_case)]
+                #[deny(improper_ctypes_definitions)]
```
tgross35: Just make these crate-level since we'll want them everywhere.

ctest-next/templates/test.rs
```diff
@@ -0,0 +1,106 @@
...
+
+
+{% for constant in ffi_items.constants() %}
+    {% let ty = constant.ty.to_token_stream().to_string() %}
+    {% let ident = constant.ident() %}
+
+    // Test that the string constant is the same in both Rust and C.
+    // While fat pointers can't be translated, we instead of * const c_char.
+    {% if ty == "* const c_char" %}
+        #[inline(never)]
```
tgross35: Why are these `inline(never)`?

ctest-next/templates/test.rs
```diff
@@ -0,0 +1,106 @@
...
+    if rust != c {
+        let rust = format!("{:?} ({:#x})", rust, rust);
+        let c = format!("{:?} ({:#x})", c, c);
+        eprintln!("bad {}: rust: {} != c {}", attr, rust, c);
+        FAILED.store(true, Ordering::Relaxed);
+    } else {
+        NTESTS.fetch_add(1, Ordering::Relaxed);
+    }
+}
+
```
tgross35, line 31:
```suggestion
```

Extra \n

ctest-next/src/tests.rs
```diff
@@ -28,64 +28,150 @@ extern "C" {
...
 
     assert_eq!(
-        ffi_items
-            .foreign_functions()
-            .iter()
-            .map(|a| a.ident())
-            .collect::<Vec<_>>(),
-        ["malloc"]
+        translator.translate_type(&ty),
+        Ok("uint8_t (*const *) [5]".to_string())
```
tgross35: This one isn't a valid C type.

It would be good to add some of these types as constants or struct fields in `simple.rs` (or a new test) so we make sure the output builds and tests properly.

ctest-next/src/tests.rs
```diff
@@ -28,64 +28,150 @@ extern "C" {
...
+    assert_eq!(collect_idents!(ffi_items.foreign_functions()), ["malloc"]);
+    assert_eq!(collect_idents!(ffi_items.foreign_statics()), ["baz"]);
+    assert_eq!(collect_idents!(ffi_items.structs()), ["Array"]);
+    assert_eq!(collect_idents!(ffi_items.unions()), ["Word"]);
+}
+
+#[test]
+fn test_translation_type_ptr() {
+    let translator = Translator {};
+    let ty: syn::Type = syn::parse_str("*const *mut i32").unwrap();
```
tgross35: Suggestion: add something like this since it's repeated in every function

```rust
fn ty(s: &str) -> Result<String, ...> {
    let translator = Translator {};
    let ty: syn::Type = syn::parse_str("*const *mut i32").unwrap();
    translator.translate_type(&ty)
}
```

Also you probably only need one `array` function, one `pointer`, etc, and then those can have >1 assert. More compact and easier to add test cases that way.

ctest-next/src/translator.rs
```diff
@@ -0,0 +1,319 @@
...
+                    syn::BinOp::BitXor(_) => Ok(format!("{left} ^ {right}")),
+                    _ => Err(TranslationError::UnknownOperator(
+                        b.op.to_token_stream().to_string(),
+                    )),
+                }
+            }
+            expr => Err(TranslationError::UnknownExpression(
+                expr.to_token_stream().to_string(),
+            )),
+        }
```
tgross35: Could we just leave expressions 1:1, i.e. only do `to_string()`? Most should map pretty directly, you shouldn't have to handle binops manually.

If anything, give it a regex pass and delete `(\B::)?\w+::` to get rid of paths.

ctest-next/templates/test.rs
```diff
@@ -0,0 +1,106 @@
...
+
+{% for constant in ffi_items.constants() %}
+    {% let ty = constant.ty.to_token_stream().to_string() %}
+    {% let ident = constant.ident() %}
+
+    // Test that the string constant is the same in both Rust and C.
+    // While fat pointers can't be translated, we instead of * const c_char.
```
tgross35, line 51: Move these comments inside the `{% if %}`, since right now both comments look like they're getting instantiated each time.

ctest-next/src/generator.rs
```diff
@@ -1,28 +1,107 @@
...
+    /// Configures the output directory of the generated Rust and C code.
+    pub fn out_dir<P: AsRef<Path>>(&mut self, p: P) -> &mut Self {
+        self.out_dir = Some(p.as_ref().to_owned());
+        self
+    }
+
+    /// Generate the Rust and C testing files.
+    ///
+    /// Returns the path to t generated file.
+    pub fn generate_files(
```
tgross35, line 82: For each generated file, you should do a regex to replace `\n{3,}` with `\n\n` to get rid of all the extra whitespace that the templates seem to insert.

ctest-next/templates/test.c
```diff
@@ -0,0 +1,21 @@
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+
+{% for header in headers %}
+    #include <{{ header }}>
+{% endfor %}
```
```suggestion
{% for header in headers %}
#include <{{ header }}>
{% endfor %}
```

I think the templates probably shouldn't indent within `{%...%}` directives. Slightly harder to read the template, but the output looks much more sane.

Ideally we'd pass the output through `rustfmt` / `clang-tidy`, but we can't assume those are available.

ctest-next/src/generator.rs
tgross35: Hm, it looks like Askama has `#[template(whitespace = "suppress")]` or `#[template(whitespace = "minimize")]`. Might be worth playing with before the regex thing.

ctest-next/templates/test.c
```diff
@@ -0,0 +1,21 @@
+#include <stdbool.h>
```
tgross35, line 4: Could you add this to both the Rust and C files so the generated outputs get a tag?

```c
/* This file was autogenerated by ctest; do not modify directly */
{# ↑ Doesn't apply here, this is the template! #}
```

ctest-next/tests/input/macro.h
```diff
@@ -0,0 +1,13 @@
...
+{
+    uint8_t x;
+    uint8_t y;
+};
+
+struct VecU16
+{
+    uint16_t x;
+    uint16_t y;
+};
```
tgross35: Please double check all files for a trailing \n, there are a few still (sometimes messes with tooling otherwise)

ctest-next/askama.toml
mbyx: `askama` was built mainly for HTML so it has ways to sanitize input, for html it sanitizes anything in <>, but for our purposes we want to disable sanitization (treat it as plain text) for .rs, .c and .cpp files.

ctest-next/src/runner.rs
mbyx: No idea why the original template did that, I can change it.

ctest-next/src/tests.rs
mbyx: Because it needs a name `uint8_t (*const * arr) [5]` right? I'm not exactly sure what the best way to deal with those is for now, I believe that ctest2 only supports arrays when testing extern statics for some reason, where it then uses that type and converts it to the proper one, and that isn't implemented yet. (The C side doesn't do anything but define the constants etc. only the Rust side declares them as static.)

ctest-next/src/runner.rs
tgross35: I think we might want to update this in the future, so stdout is captured but stderr just gets output directly. We probably can just have a dummy `RunTestError` rather than a `String`.

But it's fine for now.

ctest-next/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustflags = ["--cfg", "procmacro2_semver_exempt"]
```
tgross35: Oh, this is kind of crummy isn't it. `.cargo/config.toml` is kind of unreliable since it doesn't set config for anything that uses this as a dependency, I was thinking this was a Cargo feature.

So for now, could you just comment out the `.file()` `.start()` etc parts of the message with a `FIXME(ctest):...`? These just stabilized so we should be able to use them somewhat soon https://github.com/dtolnay/proc-macro2/issues/503.

ctest-next/src/tests.rs
tgross35: I guess we can just figure this out once we have better testing with C. For now, just comment out the test with a FIXME

ctest-next/tests/basic.rs
```diff
@@ -1,39 +1,121 @@
...
+/// If the contents do not match and LIBCBLESS is set, overwrite the
+/// test file with the content of the generated file.
+fn bless_equal(new_file: impl AsRef<Path>, old_file: impl AsRef<Path>) -> Result<bool> {
+    let new_content = fs::read_to_string(&new_file)?
+        .replace("\r\n", "\n")
+        .replace("\r", "");
+    let old_content = fs::read_to_string(&old_file)?
+        .replace("\r\n", "\n")
+        .replace("\r", "");
+
```
tgross35: You should be able to remove the first replace for each file, since the second replace does the same thing.

ctest-next/tests/basic.rs
```diff
@@ -1,39 +1,121 @@
...
+    let equal = new_content == old_content;
+    if env::var("LIBCBLESS").is_ok() && !equal {
```
tgross35:
```suggestion
    if env::var("LIBC_BLESS").is_ok() && !equal {
```

ctest-next/tests/basic.rs
```diff
@@ -1,39 +1,121 @@
...
+            .to_owned(),
+        temp_dir,
+    ))
+}
+
+/// Returns whether the contents of two files match.
+///
+/// If the contents do not match and LIBCBLESS is set, overwrite the
+/// test file with the content of the generated file.
+fn bless_equal(new_file: impl AsRef<Path>, old_file: impl AsRef<Path>) -> Result<bool> {
```
tgross35: No need to have a `Result` here I think, you can just do the assertions in this function. Use https://docs.rs/pretty_assertions/latest/pretty_assertions/ so there is a nice diff.

ctest-next/templates/test.rs
```diff
@@ -0,0 +1,111 @@
...
+    #![allow(non_snake_case)]
+    #![deny(improper_ctypes_definitions)]
+    use super::*;
+
+    use std::mem;
+    use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+    use std::fmt::{Debug, LowerHex};
+    use std::ptr;
+    use std::slice;
+    use std::ffi::CStr;
```
```suggestion
    use std::ffi::CStr;
    use std::fmt::{Debug, LowerHex};
    use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
    use std::{mem, ptr, slice};

    use super::*;
```

Making this match rustfmt

ctest-next/templates/test.rs
```diff
@@ -0,0 +1,111 @@
...
+use generated_tests::*;
+
+fn main() {
+    println!("RUNNING ALL TESTS");
+    run_all();
+    if FAILED.load( std::sync::atomic::Ordering::Relaxed) {
+        panic!("some tests failed");
+    } else {
+        println!("PASSED {} tests", NTESTS.load( std::sync::atomic::Ordering::Relaxed));
+    }
```
```suggestion
    if FAILED.load(std::sync::atomic::Ordering::Relaxed) {
        panic!("some tests failed");
    } else {
        println!(
            "PASSED {} tests",
            NTESTS.load(std::sync::atomic::Ordering::Relaxed)
        );
    }
```

Same thing, just making rustfmt happy with the output files

tgross35 (commented):
Few remaining small comments, then I think we can get this merged!

mbyx:
I'm not exactly sure why it's happening. The errors seem to be in `generate_files` and `__compile_test` but for `generate_files` both of the possible errors that could be generated there should give a message (not present for environment variables, and some kind of OS error for writing to a file/creation).
For `__compile_test` the only thing that could be failing is accessing the file stem, since that's an `Option` converted to a `Result`.

tgross35:
It might be worth adding a `ThisError` type for the `generator` module and returning that, to narrow down where exactly it is coming from. I don't know what the cause could be either.

There is a a nonzero chance that it's due to platform flakiness, in which case it is okay to update the CI config to skip tests on this platform.

@rustbot author

mbyx:
Given that they didn't fail this time it's probably a spurious `OsError` of some kind, that for some reason doesn't show up properly.

@rustbot ready

PR #4503, opened 2025-06-26
FreeBSD amd64: add mc_tlsbase member to mcontext_t

kostikbel:
# Description

The mc_tlsbase member was recently added to struct mcontext on amd64,
both on CURRENT and on stable/14.
This PR adds it to the Rust libc structure as well.
 
# Sources

https://github.com/freebsd/freebsd-src/blob/716b6667b65b50c15900be9f88cc67f2326872b4/sys/x86/include/ucontext.h#L163

# Checklist


tgross35 (approved):
Thank you!

PR #4505, opened 2025-06-26
Test semver check

tgross35:


tgross35:
No results because we can't test 32 bit with semver-checks :/

PR #4507, opened 2025-06-27
Use unique errno values for AIX

xingxue-ibm:
# Description
AIX defines some `errno` values as aliases of other `errno` values under the `_ALL_SOURCE` macro, but the POSIX standard requires each `errno` to be unique. This also causes some crates, such as `nix`, to fail. This PR updates the affected `errno` values in `libc` to address the issue.
 

# Sources

# Checklist


collinfunk:
Can confirm this is the case on an AIX 7.3 machine I have access to. `EWOULDBLOCK == EAGAIN` if `_XOPEN_SOURCE_EXTENDED == 1` and `ENOTEMPTY == EEXIST` if `defined _ALL_SOURCE`. I'm assuming that `_ALL_SOURCE` will define `_XOPEN_SOURCE_EXTENDED` to `1`, much like other feature macros.

Therefore, this change looks good to me.

tgross35 (approved):
Thanks @xingxue-ibm for the PR and @collinfunk for reviewing!

I'm assuming you want this in 0.2 so

PR #4508, opened 2025-06-28
unix: make `struct timespec` non-exhaustive.

xbjfk:
# Description

As discussed in rust-lang/libc#4463 time64 changes on both GNU and Musl libcs add padding fields to timespec. This can cause compilation failures when using time64 when regular 64-bit compiles fine. Add `non_exhaustive` to force people to use e.g. `..Default::default()` when initializing to prevent sudden errors appearing when using time64.

# Sources

Not applicable.

# Checklist

Note: this is a breaking change.


xbjfk:
Closing, as discussed most structs will be marked `non_exhaustive`

PR #4509, opened 2025-07-02
musl: mips64: Fix type of nlink_t

Gelbpunkt:
# Description

musl defines nlink_t to be an unsigned 32-bit integer on mips64, therefore changing the size of stat64 from 216 bytes to 208.

The current definition in the libc crate does not match this and therefore defines stat64 wrong on mips64 musl, which results in bogus readings of fields following st_nlink.

See https://git.musl-libc.org/cgit/musl/tree/arch/mips64/bits/alltypes.h.in#n22 for the musl definition of nlink_t.

See also https://godbolt.org/z/6Ed3fvv76 for a godbolt link that confirms the sizes.

Without this change, the Rust compiler is completely broken on mips64-unknown-linux-muslabi64: it fails to mmap builtin libraries because it thinks their size is 0! It isn't, the field offset is just wrongly defined in libc.

# Sources

https://git.musl-libc.org/cgit/musl/tree/arch/mips64/bits/alltypes.h.in#n22

https://godbolt.org/z/6Ed3fvv76

# Checklist


src/unix/linux_like/linux/musl/b64/mips64.rs
```diff
@@ -4,7 +4,7 @@ use crate::prelude::*;
 pub type wchar_t = i32;
 pub type __u64 = c_ulong;
 pub type __s64 = c_long;
-pub type nlink_t = u64;
+pub type nlink_t = u32;
```
tgross35: Mind using `c_uint` instead to match the source `unsigned`?

tgross35 (changes requested):

src/unix/linux_like/linux/musl/b64/mips64.rs
Gelbpunkt: Done :)

Gelbpunkt:
Since this fixes rustc:

PR #4510, opened 2025-07-02
musl: Decommonize definition of fanotify_event_metadata

Gelbpunkt:
# Description

musl has its own definition of the fanotify_event_metadata struct and doesn't use the one from the Linux kernel.

The difference here is that musl's mask field has the type unsigned long long, while the kernel uses __u64.

This currently causes libc-test to fail to compile on musl targets.

# Sources

Linux: https://github.com/torvalds/linux/blob/master/include/uapi/linux/fanotify.h#L143
musl: https://git.musl-libc.org/cgit/musl/tree/include/sys/fanotify.h#n15

# Checklist


src/unix/linux_like/linux/musl/mod.rs
```diff
@@ -122,6 +122,17 @@ impl siginfo_t {
...
 s! {
+    #[repr(align(8))]
+    pub struct fanotify_event_metadata {
+        pub event_len: __u32,
+        pub vers: __u8,
+        pub reserved: __u8,
+        pub metadata_len: __u16,
+        pub mask: c_ulonglong,
+        pub fd: c_int,
+        pub pid: c_int,
```
tgross35: This should probably also use `c_uint`, `c_uchar` etc to match the musl header, rather than the kernel `__u*` types

Gelbpunkt: Thanks, fixed

PR #4511, opened 2025-07-02
linux: gnu/musl: MAP_32BIT is only defined on x86

Gelbpunkt:
# Description

The Linux kernel only defines MAP_32BIT in the asm/mman.h header on x86. Remove the erraneous definitions for any other architectures on Linux targets.

This makes libc-test somewhat compile on powerpc64le musl (except semver, I'll follow up on that as well).

# Sources

See https://github.com/torvalds/linux/blob/master/arch/x86/include/uapi/asm/mman.h#L5

# Checklist

^ didn't test the other two targets, but this is definitely correct. semver tests only tested for this on x86 already


src/unix/linux_like/linux/musl/b64/powerpc64.rs
```diff
@@ -72,6 +72,7 @@ s! {
 }
 
 pub const MADV_SOFT_OFFLINE: c_int = 101;
+#[deprecated = "Linux does not define MAP_32BIT on any architectures other than x86 and x86_64, this constant will be removed in the future"]
```
tgross35: Small nit, adding a version and keeping line length

```rust
#[deprecated(since = "0.2.175", note = "Linux does not define MAP_32BIT on any architectures \
    other than x86 and x86_64, this constant will be removed in the future"]
```

Gelbpunkt: Done

tgross35 (approved):
Thanks!

PR #4513, opened 2025-07-02
musl: Fix definition of NCCS on powerpc(64)

Gelbpunkt:
# Description

These overwrite the value with their own.

# Sources

https://git.musl-libc.org/cgit/musl/tree/arch/powerpc/bits/termios.h#n2
https://git.musl-libc.org/cgit/musl/tree/arch/powerpc64/bits/termios.h#n2

# Checklist


tgross35 (approved):
Easy backport since users may inadvertently be doing an OOB write

PR #4514, opened 2025-07-02
ctest: Add skips and mappings.

mbyx:
# Description

This PR adds more methods to the frontend of `TestGenerator`, including the ability to skip items and map their identifiers and types to another if needed. This will be useful when porting `ctest-test` and for writing cleaner test templates for the other items in the future.

# Sources

# Checklist


ctest-next/src/lib.rs
```diff
@@ -31,3 +31,93 @@ pub type Error = Box<dyn std::error::Error>;
...
+    /// The C++ programming language.
+    CXX,
+}
+
+impl Language {
+    /// Return the extension corresponding to the chosen language.
+    pub fn extension(&self) -> &str {
+        match self {
+            Self::C => "c",
+            Self::CXX => "cpp",
```
tgross35: Cpp can also commonly be `cc` or `cxx`, this might need to return an array (unless it's only used for output files, not input?)

ctest-next/src/lib.rs
```diff
@@ -31,3 +31,93 @@ pub type Error = Box<dyn std::error::Error>;
...
+    /// The C programming language.
+    #[default]
+    C,
+    /// The C++ programming language.
+    CXX,
+}
+
+impl Language {
+    /// Return the extension corresponding to the chosen language.
+    pub fn extension(&self) -> &str {
```
tgross35: Can this be `pub(crate)`?

ctest-next/tests/basic.rs
```diff
@@ -37,7 +34,7 @@ fn bless_equal(new_file: impl AsRef<Path>, old_file: impl AsRef<Path>) {
     let new_content = fs::read_to_string(&new_file).unwrap().replace("\r", "");
     let old_content = fs::read_to_string(&old_file).unwrap().replace("\r", "");
 
-    let equal = new_content != old_content;
+    let equal = new_content == old_content;
     if env::var("LIBC_BLESS").is_ok() && !equal {
         fs::write(old_file, &new_content).unwrap();
     } else {
```
tgross35: You probably don't actually need to check `equal` before writing, no harm in overwriting the file if they're equal. So you can check `LIBC_BLESS` at the beginning and write the file if so, then read the file and assert it's equal.

ctest-next/tests/input/hierarchy.out.c
```diff
@@ -10,6 +10,6 @@ static bool __test_const_ON_val = ON;
 
 // Define a function that returns a pointer to the value of the constant to test.
 // This will later be called on the Rust side via FFI.
-bool* __test_const_ON(void) {
+ bool* __test_const_ON(void) {
```
tgross35: Looks like this might be spurious?

ctest-next/src/lib.rs
```diff
@@ -31,3 +31,93 @@ pub type Error = Box<dyn std::error::Error>;
...
+    }
+
+    /// Return the linkage corresponding to the chosen language.
+    pub fn linkage(&self) -> &str {
+        match self {
+            Self::C => "",
+            Self::CXX => "extern \"C\"",
+        }
+    }
+}
```
tgross35: Maybe name this `cpp_linkage`? I assumed it was Rust's `extern "C"` at first

ctest-next/src/generator.rs
```diff
@@ -76,9 +99,183 @@ impl TestGenerator {
...
+    /// Is argument of function an array?
+    ///
+    /// The closure denotes whether particular argument of a function is an array.
+    pub fn array_arg<F>(
+        &mut self,
+        f: impl Fn(crate::Fn, Parameter) -> bool + 'static,
+    ) -> &mut Self {
+        self.array_arg = Some(Box::new(f));
+        self
+    }
```
tgross35: For these docs, could you expand upon what we actually do with this information? I.e., if we know a function argument is an array, how does that change our output.

ctest-next/src/generator.rs
```diff
@@ -76,9 +99,183 @@ impl TestGenerator {
...
+            }
+        }));
+        self
+    }
+
+    /// Sets the programming language.
+    pub fn language(&mut self, language: Language) -> &mut Self {
+        self.language = language;
+        self
+    }
```
tgross35: Similar here, what effect does changing the programming language have?

ctest-next/src/generator.rs
```diff
@@ -76,9 +99,183 @@ impl TestGenerator {
...
+            }
+        }));
+        self
+    }
+
+    /// Configures how a Rust type is translated to a C type.
+    pub fn map_type(
+        &mut self,
+        f: impl Fn(&str, bool, bool) -> Option<String> + 'static,
+    ) -> &mut Self {
```
tgross35: Maybe `rename_fn` and `rename_type`? Also, what are the `bool`s in `map_type`?

tgross35: Ah, looks like `is_struct` and `is_union` from `ctest`. Maybe an `#[non_exhaustive] enum TyKind { Struct, Union }` instead?

ctest-next/src/generator.rs
```diff
@@ -112,20 +312,69 @@ impl TestGenerator {
...
         File::create(&c_output_path)
             .map_err(GenerationError::OsError)?
             .write_all(
-                CTestTemplate::new(headers, &ffi_items)
+                CTestTemplate::new(&ffi_items, self)
                     .render()
-                    .map_err(|e| GenerationError::TemplateRender("C".to_string(), e.to_string()))?
+                    .map_err(|e| {
+                        GenerationError::TemplateRender(
+                            self.language.extension().to_string(),
```
tgross35: Use a separate method like `display_name` for proper names like `C++` and `C`.

ctest-next/src/generator.rs
```diff
@@ -112,20 +312,69 @@ impl TestGenerator {
...
+    /// Does not skip specific tests or specific fields.
+    fn filter_ffi_items(&self, ffi_items: &mut FfiItems) {
+        let verbose = self.verbose_skip;
+
+        macro_rules! filter {
+            ($field:ident, $variant:ident, $label:literal) => {{
+                let (retained, skipped): (Vec<_>, Vec<_>) = ffi_items
+                    .$field
+                    .drain(..)
+                    .partition(|item| !self.skips.iter().any(|f| f(&MapInput::$variant(item))));
```
tgross35: I think `.drain(..)` could be `.into_iterator()` here. But [`extract_if`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extract_if) provides a nicer way yet to do this.

ctest-next/src/generator.rs
```diff
@@ -76,9 +99,183 @@ impl TestGenerator {
...
+    /// The closure denotes whether particular argument of a function is an array.
+    pub fn array_arg<F>(
+        &mut self,
+        f: impl Fn(crate::Fn, Parameter) -> bool + 'static,
+    ) -> &mut Self {
+        self.array_arg = Some(Box::new(f));
+        self
+    }
+
+    /// Configures whether the tests for a struct are emitted.
```
tgross35, line 276: To demo usage, could you please add a doc example to all these added methods? They probably have to be `no_run` like ctest is

tgross35:
I've been thinking a bit - I think that for the next PR, you should get tests running against the real `libc`, in parallel with the existing `ctest`. Since you have the skips, it should basically be possible to opt in to a subset with something like:

```rust
cfg.skip_static(c| 
    ![
        "some_static",
        "other_static",
    ].contains(c.name())
)
```

The reason I am suggesting this is because there is probably a lot that won't be obvious until this starts processing real code and we see the kind of failures that arise.

ctest-next/src/lib.rs
mbyx: It's only used for output files, we only take header files as input.

ctest-next/tests/input/hierarchy.out.c
mbyx: It's because of the support for CPP, now there can optionally be a linkage of extern "C" present. The space can be removed if we move the space to the linkage function of `Language`, but then it isn't clear in the template file why there isn't a space between {{ linkage }} and other items when you have to do it for other items.

tgross35: > then it isn't clear in the template file why there isn't a space between {{ linkage }} and other items when you have to do it for other items.

I think that's a fine tradeoff in exchange for better test output. But couldn't this use the `{{ .. -}}` option to trim whitespace after?

tgross35: I'm not sure how much value there is in being able to specify C++; what are the usecases here actually?

ctest-next/src/lib.rs
mbyx: It can be but it's used to make the integration tests more convenient (otherwise we'd have to specify the extension manually for every test even though we specified it for the generator).

ctest-next/tests/input/hierarchy.out.c
mbyx: `{{ .. -}}` wouldn't work, it would result in `extern "C"type func(..)` which won't compile.

As for use cases for C++, that's just ported from the original ctest, it's only tested in `ctest-test` but not used in libc from what I understand.

ctest-next/src/generator.rs
mbyx: `extract_if` was unfortunately stabilized in `1.87.0`, since our MSRV is `1.77.0` we can't use it.

ctest-next/tests/input/hierarchy.out.c
tgross35: If the only thing `Lang::CXX` does is to generate a C++ test file and use a C++ compiler, I think you may as well just drop it. We won't ever be generating constructs that only work in C++ but not C so I'm not really not sure what the point in the original `ctest` was. Also doesn't look like it has any use https://github.com/search?q=lang%3Arust+%2F%5C.language%5C%28Lang%3A%3ACXX%5C%29%2F&type=code

ctest-next/src/generator.rs
tgross35: Feel free to bump the msrv, it's a useful enough API to count as motivation imo. I'm sure it will come in handy more places.

ctest-next/Cargo.toml
```diff
@@ -18,3 +18,4 @@ thiserror = "2.0.12"
 [dev-dependencies]
 pretty_assertions = "1.4.1"
 tempfile = "3.20.0"
+
```
tgross35: Spurious newline?

ctest-next/src/generator.rs
```diff
@@ -65,20 +105,371 @@ impl TestGenerator {
...
+        self.volatile_item = Some(Box::new(f));
+        self
+    }
+
+    /// Is argument of function an array?
+    ///
+    /// The closure denotes whether particular argument of a function is an array.
+    /// This is used to figure out which pointer argument is actually an array. For
+    /// example, `uint8_t*` could be a pointer to an integer or pointer to the start
+    /// of an array.
```
tgross35: Wording is a bit redundant, maybe simplify to:

> This closure should return true if a pointer argument to a function should be generated with `T foo[]` syntax rather than `T *foo`.

ctest-next/src/generator.rs
```diff
@@ -65,20 +105,371 @@ impl TestGenerator {
...
+    ///         VolatileItemKind::StructField(s, f)
+    ///             if s.ident() == "foo_struct" && f.ident() == "foo_field"
+    ///              => true,
+    ///         _ => false,
+    /// }});
+    /// ```
+    pub fn volatile_item(&mut self, f: impl Fn(VolatileItemKind) -> bool + 'static) -> &mut Self {
+        self.volatile_item = Some(Box::new(f));
+        self
+    }
```
tgross35: Would it be possible to split this up some? To `volatile_field`, `volatile_static`, `volatile_function_arg` etc? I don't think there is much benefit to building an enum since it's just going to be `match`ed.

ctest-next/src/generator.rs
```diff
@@ -65,20 +105,371 @@ impl TestGenerator {
...
+    ///         Some(format!("{}_t", ty))
+    ///     } else {
+    ///         None
+    ///     }
+    /// });
+    /// ```
+    pub fn rename_type(
+        &mut self,
+        f: impl Fn(&str, TyKind) -> Option<String> + 'static,
+    ) -> &mut Self {
```
tgross35: Similarly, could this be split to `rename_struct` and `rename_union`? So we shouldn't need a `TyKind`.

ctest-next/tests/basic.rs
```diff
@@ -87,16 +83,32 @@ fn test_entrypoint_hierarchy() {
...
+fn test_map_simple() {
+    let include_path = PathBuf::from("tests/input");
+    let crate_path = include_path.join("simple.rs");
+    let library_path = "simple.out.map.a";
+
+    let (mut gen, out_dir) = default_generator(1, "simple.h").unwrap();
+    gen.rename_constant(|c| (c.ident() == "B").then(|| "C_B".to_string()));
+
+    check_entrypoint(&mut gen, out_dir, crate_path, library_path, include_path);
 }
```
tgross35: Could you change `.skip` and `.map` to `.with-skips` and `.with-renames`? Just trying to make the outputs a bit more obvious.

tgross35 (commented):
Few small things then I think this should be good to go

ctest-next/src/generator.rs
mbyx: I don't think it can. `MapInput` current is used to map both identifiers and types, so mapping `Foo` into `Foo_t` is different from mapping `Foo_t` to `struct Foo_t`. We use `Type` when renaming types and `Struct` or `Union` for renaming structs.

tgross35: I don't think I follow this. The output here gets consumed via

```rust
            MapInput::Type(ty, TyKind::Struct) => format!("struct {ty}"),
            MapInput::Type(ty, TyKind::Union) => format!("union {ty}"),
```

Could `rename_struct` and `rename_union` make new variants `MapInput::StructTy` and `MapInput::UnionTy` instead of both using `Type`?

ctest-next/src/lib.rs
```diff
@@ -31,3 +31,77 @@ pub type Error = Box<dyn std::error::Error>;
...
+    Fn(&'a crate::Fn),
+    #[expect(unused)]
+    Field(&'a Struct, &'a Field),
+    Alias(&'a Type),
+    Const(&'a Const),
+    Static(&'a Static),
+    Type(&'a str, TyKind),
+}
+
+// The From impls make it easier to write code in the test templates.
```
tgross35:
```suggestion
/* The From impls make it easier to write code in the test templates. */
```

Nit, there's a casual convention to use `/* ... */` for section heading comments

ctest-next/src/generator.rs
mbyx: That can work.

ctest-next/tests/basic.rs
tgross35: This still needs an adjustment; delete the `else` block and move `if env::var("LIBC_BLESS")` to the beginning of the function, before the calls to `fs::read_to_string`. Otherwise I think the `.unwrap()` would cause files to fail if they don't exist yet, even if `LIBC_BLESS` is passed

ctest-next/src/generator.rs
```diff
@@ -65,20 +105,489 @@ impl TestGenerator {
...
+    /// cfg.verbose_skip(true);
+    /// ```
+    pub fn verbose_skip(&mut self, skip: bool) -> &mut Self {
+        self.verbose_skip = skip;
+        self
+    }
+
+    /// Is field volatile?
+    ///
+    /// This is used to make sure that the generated test code also has the volatile keyword.
```
tgross35:
```suggestion
    /// Indicate that a struct field should be marked `volatile`.
```

Complete sentence for the summary, then I don't think we need the rest

ctest-next/src/generator.rs
```diff
@@ -65,20 +105,489 @@ impl TestGenerator {
...
+            } else {
+                false
+            }
+        }));
+        self
+    }
+
+    /// Is static variable volatile?
+    ///
+    /// This is used to make sure that the generated test code also has the volatile keyword.
```
tgross35:
```suggestion
    /// Indicate that a static should be marked `volatile`.
```

ctest-next/src/generator.rs
```diff
@@ -65,20 +105,489 @@ impl TestGenerator {
...
+            } else {
+                false
+            }
+        }));
+        self
+    }
+
+    /// Is function argument volatile?
+    ///
+    /// This is used to make sure that the generated test code also has the volatile keyword.
```
tgross35:
```suggestion
    /// Indicate that a function argument should be marked `volatile`.
```

ctest-next/src/generator.rs
```diff
@@ -65,20 +105,489 @@ impl TestGenerator {
...
+            } else {
+                false
+            }
+        }));
+        self
+    }
+
+    /// Is function return type volatile?
+    ///
+    /// This is used to make sure that the generated test code also has the volatile keyword.
```
tgross35:
```suggestion
    /// Indicate that a function's return type should be marked `volatile`.
```

ctest-next/src/generator.rs
```diff
@@ -65,20 +105,489 @@ impl TestGenerator {
...
+            if let VolatileItemKind::FnReturnType(func) = item {
+                f(func)
+            } else {
+                false
+            }
+        }));
+        self
+    }
+
+    /// Is argument of function an array?
```
tgross35:
```suggestion
    /// Indicate that a function pointer argument is an array.
```

ctest-next/src/lib.rs
```diff
@@ -31,3 +31,67 @@ pub type Error = Box<dyn std::error::Error>;
...
+    Struct(&'a Struct),
+    Fn(&'a crate::Fn),
+    #[expect(unused)]
+    Field(&'a Struct, &'a Field),
+    Alias(&'a Type),
+    Const(&'a Const),
+    Static(&'a Static),
+    Type(&'a str),
+    StructType(&'a str),
+    UnionType(&'a str),
```
tgross35: Could you add a doc to `Struct` and `StructType` showing the difference there?

tgross35 (commented):
Few misc comments, mostly just doc grammar fixes. Looks like there is a test failure

tgross35:
You'll probably need to rebase for the test failure

tgross35:
You just need to re-bless the tests no? The input files had to change at https://github.com/rust-lang/libc/pull/4525, the new files added here don't reflect that.

PR #4515, opened 2025-07-02
s390x-linux-musl: remove bogus constants

neuschaefer:
# Description

This change removes two constants from s390x.rs that are not present in musl-libc. This moves the `libc` testsuite closer to running successfully on s390x-linux-musl.

# Sources

# Checklist


tgross35:
Could you deprecate these rather than removing them? Obviously nobody should be using them, but we should give a bit of a migration path for anyone who isn't configuring these out on s390x.

You can add a `FIXME(s390x)` skip in the tests until we can actually remove them.

tgross35:
> Sure, I'll deprecate them instead. What should I use as the version number for the `since` parameter?

This will get in the next release, which will be 0.2.175. Thank you!

PR #4518, opened 2025-07-03
musl: powerpc(64): Decommonize termios definitions

Gelbpunkt:
# Description

PowerPC targets use their own, separate definitions of termios that change the order of the c_line and c_cc fields.

# Sources

https://git.musl-libc.org/cgit/musl/tree/arch/powerpc/bits/termios.h#n3
https://git.musl-libc.org/cgit/musl/tree/arch/powerpc64/bits/termios.h#n3

# Checklist


tgross35 (approved):

tgross35:
Android is just super flaky unfortunately, looks like it made it to the queue now

PR #4519, opened 2025-07-03
musl: powerpc64: Decommonize definition of shmid_ds

Gelbpunkt:
# Description

powerpc64 is the only 64-bit target that doesn't use the generic definition for these in musl.

# Sources

https://git.musl-libc.org/cgit/musl/tree/arch/powerpc64/bits/shm.h#n3

# Checklist

### DIFF
--- a/src/unix/linux_like/linux/musl/b64/mod.rs
+++ b/src/unix/linux_like/linux/musl/b64/mod.rs
@@ -17,6 +17,8 @@ s! {
         __val: [c_ulong; 16],
     }
 
+    // PowerPC implementation is special, see the subfolder.
+    #[cfg(not(target_arch = "powerpc64"))]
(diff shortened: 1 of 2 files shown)


Gelbpunkt:
With this, libc-test now fully passes on `powerpc64le-unknown-linux-musl`. I'll add a PR for testing it in CI momentarily.

Gelbpunkt:
@tgross35 seems like there are labels like `O-arm` and so on, but none for powerpc, could that perhaps be added?

tgross35:
> @tgross35 seems like there are labels like `O-arm` and so on, but none for powerpc, could that perhaps be added?

Sure, adding the autolabel in https://github.com/rust-lang/libc/pull/4523

PR #4521, opened 2025-07-03
ci/docs: Build and check powerpc64le-unknown-linux-musl

Gelbpunkt:
# Description

Now that all of the issues with powerpc64 + musl are resolved, let's make sure it won't regress in the future and also build the documentation for it.

This is a tier 2 target for a few months now and requires musl 1.2.3 as a baseline.

The big-endian target isn't tier 2 yet, I'm still working towards that, so we can't test it at the moment.

This obviously depends on all the remaining open pull requests with the fixes.

# Sources

https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2-with-host-tools

# Checklist

### DIFF
--- a/ci/install-musl.sh
+++ b/ci/install-musl.sh
@@ -68,6 +68,13 @@ case ${1} in
           ./configure --prefix="/musl-${musl_arch}" --enable-wrapper=yes
         make install -j4
         ;;
+    powerpc64*)
+        musl_arch=powerpc64
+        kernel_arch=powerpc
+        CC="${1}-linux-gnu-gcc" CFLAGS="-mlong-double-64" \
+          ./configure --prefix="/musl-${musl_arch}" --enable-wrapper=yes
+        make install -j4
+        ;;
     *)
         echo "Unknown target arch: \"${1}\""
         exit 1
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -188,6 +188,7 @@ jobs:
           - loongarch64-unknown-linux-musl
           - powerpc64-unknown-linux-gnu
           - powerpc64le-unknown-linux-gnu
+          - powerpc64le-unknown-linux-musl
           - riscv64gc-unknown-linux-gnu
           - s390x-unknown-linux-gnu
           - wasm32-unknown-emscripten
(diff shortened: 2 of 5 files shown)


ci/install-musl.sh
```diff
@@ -68,6 +68,13 @@ case ${1} in
           ./configure --prefix="/musl-${musl_arch}" --enable-wrapper=yes
         make install -j4
         ;;
+    powerpc64*)
+	musl_arch=powerpc64
+	kernel_arch=powerpc
+	CC=${1}-linux-gnu-gcc CFLAGS="-mlong-double-64" \
+	  ./configure --prefix="/musl-${musl_arch}" --enable-wrapper=yes
+	make install -j4
+	;;
```
tgross35: Quote `"${1}-linux-gnu-gcc"` just so we don't have to think about splitting, could you also match the indentation of the other arms?

Also, shouldn't long double either be ibm double-double or binary128?

Gelbpunkt: musl requires a 64-bit long double for powerpc64, see https://musl.openwall.narkive.com/ap0RkjNz/long-double-on-powerpc64
The -gnu GCC defaults to 128 and fails the configure check

tgross35: Interesting, I would have assumed they would switch to f128 at some point since AIUI there are plans to migrate PPC there eventually (not that I've heard anything about this since Fedora)

> The -gnu GCC defaults to 128 and fails the configure check

Just clarifying, ieee or ibm?

Gelbpunkt: `The default long double format is to use IBM extended double. Until all of the libraries are converted to use IEEE 128-bit floating point, it is not recommended to use --with-long-double-format=ieee.`
(From https://gcc.gnu.org/install/configure.html)

Gelbpunkt:
I rebased and addressed the comments, should be ready to go now.

Gelbpunkt:
Hm, verify-build fails on 1.63 since there is no std prebuilt on that version yet AFAICT.

tgross35:
I guess just add it to `rust_nightly_linux_targets` instead for now, with a note that we can move it when we change the MSRV

PR #4524, opened 2025-07-03
Add function getpeereid for AIX

xingxue-ibm:
# Description
AIX provides `getpeereid` in `libc` even though it is not mandated by POSIX. Since the `tokio` crate has a dependency on this function, this PR makes it available in the libc crate.

# Sources

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -5668,6 +5668,9 @@ fn test_aix(target: &str) {
             // https://github.com/gnzlbg/ctest/issues/68.
             "lio_listio" => true,
 
+            // The function is only available under macro _KERNEL in 'proto_uipc.h'.
+            "getpeereid" => true,
+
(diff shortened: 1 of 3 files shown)


tgross35 (approved):

collinfunk:
Can also confirm this is the case without logging into AIX. The `getpeerid` function declaration is hidden in a way that has caused me problems a few times previously. :)

collinfunk:
@tgross35 Shouldn't it be merged to libc-0.2, though?

xingxue-ibm:
Thank you so much!

collinfunk:
> Everything goes through main, just add the label if you'd like me to pick something up for 0.2 (which should be pretty much everything)

Yep, I figured I could add it myself. But wasn't sure if it was correct to do so.

Thank you for the clarification!

PR #4525, opened 2025-07-03
Cast to `c_char` rather than `i8`

xingxue-ibm:
# Description
The AIX CI failed with the following error because `c_char` is `u8` on AIX and casting to `i8` causes type mismatch. This PR changes to use `cast::<c_char>())` to cast `ptr`.
```
thread 'test_entrypoint_simple' panicked at ctest-next/tests/basic.rs:75:78:
called `Result::unwrap()` on an `Err` value: "error[E0308]: mismatched types\n  --> /tmp/.tmpL9FroM/simple.out.rs:53:38\n
        |\n
     53 |             let val = CStr::from_ptr(ptr.cast::<i8>());\n
        |                       -------------- ^^^^^^^^^^^^^^^^ expected `*const u8`, found `*const i8`\n
        |\n
```

# Sources

# Checklist

### DIFF
--- a/ctest-next/templates/test.rs
+++ b/ctest-next/templates/test.rs
@@ -56,7 +56,7 @@ mod generated_tests {
         let val = {{ ident }};
         unsafe {
             let ptr = *__test_const_{{ ident }}();
-            let val = unsafe {CStr::from_ptr(ptr)};
+            let val = CStr::from_ptr(ptr as _);
             let val = val.to_str().expect("const {{ ident }} not utf8");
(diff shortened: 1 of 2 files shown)


tgross35 (changes requested):
"c_char can be i8 or u8, so just cast to i8." huh, that's not right and I'm surprised more tests aren't failing. Could you use the more explicit `.cast::<c_char>()` syntax though?

xingxue-ibm:
Thank you!

PR #4527, opened 2025-07-04
musl: mips64: Use special MIPS definition of statfs

Gelbpunkt:
# Description

64-bit MIPS has the same special definition as 32-bit MIPS in musl.

# Sources

https://git.musl-libc.org/cgit/musl/tree/arch/mips64/bits/statfs.h

# Checklist

### DIFF
--- a/src/unix/linux_like/linux/musl/b64/mips64.rs
+++ b/src/unix/linux_like/linux/musl/b64/mips64.rs
@@ -76,6 +76,36 @@ s! {
         __unused1: c_ulong,
         __unused2: c_ulong,
     }
+
+    pub struct statfs {
+        pub f_type: c_ulong,
+        pub f_bsize: c_ulong,
+        pub f_frsize: c_ulong,
+        pub f_blocks: crate::fsblkcnt_t,
+        pub f_bfree: crate::fsblkcnt_t,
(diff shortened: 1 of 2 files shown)


src/unix/linux_like/linux/musl/b64/mips64.rs
```diff
@@ -76,6 +76,36 @@ s! {
...
+        pub f_blocks: crate::fsblkcnt_t,
+        pub f_bfree: crate::fsblkcnt_t,
+        pub f_files: crate::fsfilcnt_t,
+        pub f_ffree: crate::fsfilcnt_t,
+        pub f_bavail: crate::fsblkcnt_t,
+        pub f_fsid: crate::fsid_t,
+        pub f_namelen: c_ulong,
+        pub f_flags: c_ulong,
+        pub f_spare: [c_ulong; 5],
+    }
```
tgross35: This looks the same as `statfs`, could it just be an alias? Or use the `*64` types here to match the non-mips definition

Gelbpunkt: Indeed, it's the same. Good catch, I'll use the `*64` types :)

tgross35 (approved):
Thank you!

PR #4530, opened 2025-07-04
musl: mips64: Swap order of si_errno and si_code in siginfo_t

Gelbpunkt:
# Description

All MIPS targets, 32-bit and 64-bit, swap these around.

# Sources

https://git.musl-libc.org/cgit/musl/tree/arch/mips64/bits/signal.h#n103
https://git.musl-libc.org/cgit/musl/tree/include/signal.h#n100

# Checklist

### DIFF
--- a/src/unix/linux_like/linux/musl/mod.rs
+++ b/src/unix/linux_like/linux/musl/mod.rs
@@ -148,10 +148,10 @@ s! {
     // FIXME(union): C implementation uses unions
     pub struct siginfo_t {
         pub si_signo: c_int,
-        #[cfg(not(target_arch = "mips"))]
+        #[cfg(not(any(target_arch = "mips", target_arch = "mips64")))]
         pub si_errno: c_int,
         pub si_code: c_int,
-        #[cfg(target_arch = "mips")]
+        #[cfg(any(target_arch = "mips", target_arch = "mips64"))]
         pub si_errno: c_int,
         #[doc(hidden)]
         #[deprecated(


Gelbpunkt:
This fixes libc-test on `mips64-unknown-linux-muslabi64`. Since that's tier 3, I guess we can't test it in CI, bit of a shame since I'd prefer libc to not break on this target so that rustc remains bootstrappable.

tgross35:
> Since that's tier 3, I guess we can't test it in CI, bit of a shame since I'd prefer libc to not break on this target so that rustc remains bootstrappable.

I'm sure you were involved in the discussion but just in case: mips was demoted because we were having a number of problems, mostly with the backend ([more details here](https://github.com/rust-lang/compiler-team/issues/648)), that were causing CI flakiness across all repos. It's very possible to get it back up to T2, but in addition to interested maintainers on the Rust side, we definitely need to make sure there are people actively supporting LLVM for the target.

Gelbpunkt:
> Just makes it easier to test whether mips things build in rust-lang/rust.

The issue I was trying to get at is that currently, the mips64 musl target rustc builds fine, but it wasn't working whatsoever because of libc. No try-job for dist-* would have caught that unless it would have tried using the compiled artifacts. If we were to run libc-test against this target, we would at least rule that out as a reason for a broken compiler. Especially with how long things take to propagate through the ecosystem with new libc releases (rustc still vendors like 10? different versions), ensuring libc-test passes is even more important IMO. It will probably take a while until the compiler is ready to use on mips64-musl again, just because of the slow adoption.

tgross35:
I'm not strongly opposed to adding a job here if that's a problem and you are interested, it just has to be set up as non-blocking and hopefully not need too much new config. I think before doing that, however, it would be good to list yourself as a target maintainer for whichever you are interested in supporting (just a PR to update https://github.com/rust-lang/rust/tree/25cf7d13c960a3ac47d1424ca354077efb6946ff/src/doc/rustc/src/platform-support) to indicate this isn't for a totally unsupported target.

PR #4531, opened 2025-07-06
add prctl constants on android

dead10ck:
# Description

Add all constants from bionic's `prctl.h` header

# Sources

* https://android.googlesource.com/platform/bionic/+/7ac54f5c391bf3aeb9fec84a158939421fd8f505/libc/kernel/uapi/linux/prctl.h
* https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/kernel/uapi/linux/prctl.h;drc=7ac54f5c391bf3aeb9fec84a158939421fd8f505

# Checklist


dead10ck:
I am not sure if those FreeBSD failures have anything to do with my change. The errors appear to be about a missing symbol that only appears in the FreeBSD path.

src/unix/linux_like/android/mod.rs
```diff
@@ -3441,6 +3606,9 @@ safe_f! {
     }
 }
 
+#[cfg_attr(termux, link(name = "android-posix-semaphore"))]
+extern "C" {}
```
tgross35: What is this added for?

libc-test/build.rs
```diff
@@ -2288,7 +2291,10 @@ fn test_android(target: &str) {
             // FIXME(android): bad function pointers:
             "isalnum" | "isalpha" | "iscntrl" | "isdigit" | "isgraph" | "islower" | "isprint"
             | "ispunct" | "isspace" | "isupper" | "isxdigit" | "isblank" | "tolower"
-            | "toupper" => true,
+            | "toupper" | "getpwuid_r" | "tmpfile" | "getpwnam" | "getpwuid" | "tcdrain"
+            | "cfgetispeed" | "cfgetospeed" | "cfsetispeed" | "cfsetospeed" | "tcgetattr"
+            | "tcsetattr" | "tcflow" | "tcflush" | "tcgetsid" | "tcsendbreak" | "cfmakeraw"
+            | "cfsetspeed" => true,
```
tgross35: These don't seem relevant to the change here

src/unix/linux_like/android/mod.rs
```diff
@@ -2900,29 +2900,194 @@ pub const PF_VSOCK: c_int = AF_VSOCK;
...
+//   __u64 brk;
+//   __u64 start_stack;
+//   __u64 arg_start;
+//   __u64 arg_end;
+//   __u64 env_start;
+//   __u64 env_end;
+//   __u64 * auxv;
+//   __u32 auxv_size;
+//   __u32 exe_fd;
+// };
```
tgross35: Could you remove this if it is unneeded?

src/unix/linux_like/android/mod.rs
dead10ck: On Android's bionic, there are some libc functions which are not supported, like `sem_open`, `sem_close`, and `sem_unlink`, which causes link failures when compiling anything with libc as a dependency.

On the Android terminal app Termux, there is a substitute port available as a separate library, which must be linked in explicitly.

libc-test/build.rs
dead10ck: It's not strictly related, but tests for all these functions fail currently.

I split it out into another commit to make that clearer

tgross35: What circumstances make these fail? We have Android tests in CI and as far as I know this hasn't been a problem

src/unix/linux_like/android/mod.rs
tgross35: I think this is outside of the scope of `libc`. We are trying to reduce the number of libraries that get implicitly linked, and a third party library that isn't `libc` doesn't seem to make the cutoff.

Are the `sem_*` functions you mentioned not available via bionic in some cases?

dead10ck: Fair enough. At the end of the day, it's easy enough to work around in downstream projects by linking in this third party lib via rustflags, etc. I've removed this.

The `sem_*` functions are not available at all. Or rather, they are "there", but are simply stubbed out, so anything that tries to use them just fails.

https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/bionic/semaphore.cpp;drc=2e65afecba6877eaffa6bda825ae156e6c682fe5;l=124

In the case of termux, for example, this would render python script that used `multiprocessing` broken.

https://github.com/termux/termux-packages/pull/8990

bionic simply doesn't implement all of libc, so there are a few shims that are needed to supplement what's missing.

libc-test/build.rs
dead10ck: I'm not sure, to be honest; I didn't spend the time to dig into the failures. It could be something specific to termux. 

I went ahead and removed these changes.

dead10ck:
Anything else I can do?

tgross35 (approved):
LGTM as long as CI passes

PR #4532, opened 2025-07-07
Add `core::mem::{size_of, align_of}` to the prelude

tgross35:
Since 1.80 these, along with their `_val` versions, are in the default `core` prelude so they don't need to be imported. 1.80 exceeds our MSRV so we can't use it from there, but we can add it to our prelude for now.


tgross35:
Double checked that all currently supported rustc targets build locally

PR #4539, opened 2025-07-08
Source reorganization of musl `socket.h`

tgross35:


src/new/linux_uapi/mod.rs
```diff
@@ -0,0 +1,4 @@
+//! This directory maps to `include/uapi` in the Linux source tree.
```
neuschaefer, line 1: Nice, such a comment is  good to have

src/new/musl/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod arch;
```
neuschaefer, line 9: This could use a comment that mentions:
- this is musl-libc
- how it relates to the musl-libc source tree

tgross35: It has some much more detailed comments now

PR #4543, opened 2025-07-08
ctest: add tests for aliases, structs, unions, as well as a test crate.

mbyx:
# Description
This PR adds support for testing aliases, structs, and unions. It additionally ports `ctest-test` to use `ctest-next`.

# Sources

# Checklist


ctest-next-test/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
```
tgross35: Is there anything different between `ctest-next-test` and the existing `ctest-test`? If it's nothing significant, I think it would be better to reuse `ctest-test` and just add the invocations to test `ctest-next` in its `build.rs`, rather than duplicating everything.

ctest-next/src/ast/structure.rs
```diff
@@ -15,4 +13,9 @@ impl Struct {
     pub fn ident(&self) -> &str {
         &self.ident
     }
+
+    /// Return the public fields of the struct.
+    pub fn public_fields(&self) -> Vec<&Field> {
```
tgross35:
```suggestion
    pub fn public_fields(&self) -> impl Iterator<Item = &Field> {
```

That way you don't need to `.collect()`

ctest-next/src/ast/union.rs
```diff
@@ -15,4 +14,9 @@ impl Union {
     pub fn ident(&self) -> &str {
         &self.ident
     }
+
+    /// Return the public fields of the union.
+    pub(crate) fn public_fields(&self) -> Vec<&Field> {
```
tgross35: Use `impl Iterator<Item = &Field>` here too

ctest-next/src/ffi_items.rs
```diff
@@ -123,10 +122,27 @@ fn visit_foreign_item_fn(table: &mut FfiItems, i: &syn::ForeignItemFn, abi: &Abi
         syn::ReturnType::Type(_, ty) => Some(ty.deref().clone()),
     };
 
+    let link_name = i
+        .attrs
+        .iter()
+        .find(|attr| attr.path().is_ident("link_name"))
```
tgross35: We should probably error here if there are >1 `link_name` attrs

ctest-next/src/lib.rs
```diff
@@ -54,16 +57,17 @@ pub enum VolatileItemKind {
...
-    Field(&'a Struct, &'a Field),
+    Field(Either<&'a Struct, &'a Union>, &'a Field),
     Alias(&'a Type),
     Const(&'a Const),
     Static(&'a Static),
     Type(&'a str),
     /// This variant is used for renaming the struct type.
     StructType(&'a str),
     UnionType(&'a str),
+    FieldType(Either<&'a Struct, &'a Union>, &'a Field),
```
tgross35: Could these be split to `StructField` `StructFieldType` `UnionField` and `UnionFieldType` rather than using `Either`? Seems slightly more straightforward since you have to `match` somewhere anyway.

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+
+        type U = {{ ident }};
+        extern "C" {
+            fn __test_roundtrip_{{ ident }}(
+                size: i32, x: U, e: *mut c_int, pad: *const u8
+            ) -> U;
+        }
+        let pad = roundtrip_padding_{{ ident }}();
+        unsafe {
+            use std::mem::{MaybeUninit, size_of};
```
tgross35: (Applies multiple places) put these imports at the top of the file, since they'll be used everywhere.

Also you don't need to import `size_of`, it's in the prelude since 1.80

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+            check_same(mem::size_of::<{{ ident }}>() as u64,
+                __test_size_{{ ident }}(), "{{ ident }} size");
+            check_same(mem::align_of::<{{ ident }}>() as u64,
+                __test_align_{{ ident }}(), "{{ ident }} align");
+        }
+    }
+
+    {%- if translator.has_sign(ffi_items, alias.ty) +%}
+
+    /// Test that the aliased type has the same sign (signed or unsigned) in both Rust and C.
```
tgross35: "sign"->"signedness" is the word for describing whether or not a type is signed.

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+    /// This check can be performed because `!(0 as _)` yields either -1 or the maximum value
+    /// depending on whether a signed or unsigned type is used. This is simply checked on both
+    /// Rust and C sides to see if they are equal.
+    pub fn sign_{{ ident }}() {
+        extern "C" {
+            fn __test_signed_{{ ident }}() -> u32;
+        }
+        unsafe {
+            check_same(((!(0 as {{ ident }})) < (0 as {{ ident }})) as u32,
+                    __test_signed_{{ ident }}(), "{{ ident }} signed");
```
tgross35: Could you add some bindings to eliminate some of the nesting here? E.g. `let all_ones = !(0 as {{ ident }}); let all_zeros = 0 as {{ ident }}`

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+
+    /// Generates a padding map for a specific type.
+    ///
+    /// Essentially, it returns a list of bytes, whose length is equal to the size of the type in
+    /// bytes. Each element corresponds to a byte and has two values. `1` if the byte is padding,
+    /// and `0` if the byte is not padding.
+    ///
+    /// For type aliases, the padding map is all zeroes.
+    fn roundtrip_padding_{{ ident }}() -> Vec<u8> {
+        vec![0; mem::size_of::<{{ ident }}>()]
```
tgross35: May as well use the prelude

```suggestion
        vec![0; size_of::<{{ ident }}>()]
```

ctest-next/templates/test.c
```diff
@@ -24,3 +24,273 @@ static {{ c_type }} __test_const_{{ ident }}_val = {{ c_ident }};
...
+// Return the alignment of a type.
+uint64_t __test_align_{{ ident }}(void) {
+    typedef struct {
+        unsigned char c;
+        {{ c_type }} v;
+    } type;
+    type t;
+    size_t t_addr = (size_t)(unsigned char*)(&t);
+    size_t v_addr = (size_t)(unsigned char*)(&t.v);
+    return t_addr >= v_addr? t_addr - v_addr : v_addr - t_addr;
```
tgross35: This can just use `_Alignof`

ctest-next-test/Cargo.toml
mbyx: There's a bunch of removals, deleting the cpp side of the tests and modifying `all.rs`, other than that the actual test files have minor changes.

ctest-next/templates/test.c
```diff
@@ -24,3 +24,273 @@ static {{ c_type }} __test_const_{{ ident }}_val = {{ c_ident }};
 }
 {%- endfor +%}
 
+{%- for alias in ffi_items.aliases() +%}
+{%- let ident = alias.ident() +%}
+{%- let c_type = self.c_type(*alias).unwrap() +%}
+
+// Return the size of a type.
+uint64_t __test_size_{{ ident }}(void) { return sizeof({{ c_type }}); }
```
tgross35: I think these names could be adjusted a bi since they aren't really testing anything in these functions. Maybe `ctest_get_{{ident}}_size`, `ctest_get_{{ident}}_align`.

The `__` prefix shouldn't be needed if `ctest_` is used instead (I assume dunder was to avoid conflicts?)

ctest-next/src/translator.rs
```diff
@@ -306,12 +308,56 @@ impl Translator {
...
+                syn::Type::Never(_) => "void".to_string(),
+                syn::Type::Tuple(tuple) if tuple.elems.is_empty() => "void".to_string(),
+                _ => self.translate_type(ty.deref())?,
+            },
+        };
+        Ok((return_type, args, signature.variadic.is_some()))
+    }
+
+    /// Determine whether a C type is a signed type.
+    pub(crate) fn has_sign(&self, ffi_items: &FfiItems, ty: &syn::Type) -> bool {
```
tgross35: Could this be named `is_signed`?

ctest-next-test/Cargo.toml
tgross35: Feel free to remove the c++ tests and make whatever changes are needed, the current `ctest` won't be getting any updates so we don't really have anything to keep testing there.

I'd prefer modifying the existing `ctest-test` anyway to see what the actual diff is.

tgross35 (commented):
(first set, more in progress)

ctest-next/templates/test.c
```diff
@@ -24,3 +24,273 @@ static {{ c_type }} __test_const_{{ ident }}_val = {{ c_ident }};
...
+    size_t t_addr = (size_t)(unsigned char*)(&t);
+    size_t v_addr = (size_t)(unsigned char*)(&t.v);
+    return t_addr >= v_addr? t_addr - v_addr : v_addr - t_addr;
+}
+
+{%- if translator.has_sign(ffi_items, alias.ty) +%}
+
+// Return `1` if the type is signed, otherwise return `0`. 
+uint32_t __test_signed_{{ ident }}(void) {
+    return ((({{ c_type }}) -1) < 0);
```
```suggestion
    return (({{ c_type }}) -1) < 0;
```

Slightly unnest parens. Also rename this one `ctest_{{ ident }}_is_signed`

ctest-next/templates/test.c
```diff
@@ -24,3 +24,273 @@ static {{ c_type }} __test_const_{{ ident }}_val = {{ c_ident }};
...
+#ifdef _MSC_VER
+// Disable signed/unsigned conversion warnings on MSVC.
+// These trigger even if the conversion is explicit.
+#  pragma warning(disable:4365)
+#endif
+
+// Tests whether the type alias `x` when passed to C and back to Rust remains unchanged.
+// It checks if the size is the same as well as if the padding bytes are all in the correct place.
+{{ c_type }} __test_roundtrip_{{ ident }}(
+        int32_t rust_size, {{ c_type }} value, int* error, unsigned char* pad
```
```suggestion
    int32_t rust_size, {{ c_type }} value, int* error, unsigned char* pad
```

Indentation nit

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+            for i in 0..sz {
+                let c: u8 = (i % 256) as u8;
+                let c = if c == 0 { 42 } else { c };
+                let d: u8 = 255_u8 - (i % 256) as u8;
+                let d = if d == 0 { 42 } else { d };
+                x_ptr.add(i).write_volatile(c);
+                y_ptr.add(i).write_volatile(d);
+            }
+            let r: U = __test_roundtrip_{{ ident }}(sz as i32, x.assume_init(), &mut error, pad.as_ptr());
+            if error == 1 {
```
tgross35: Doesn't really matter here but `error != 0` is usually good practice (robust against added error codes)

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+            let y_ptr = y.as_mut_ptr().cast::<u8>();
+            let sz = size_of::<U>();
+            for i in 0..sz {
+                let c: u8 = (i % 256) as u8;
+                let c = if c == 0 { 42 } else { c };
+                let d: u8 = 255_u8 - (i % 256) as u8;
+                let d = if d == 0 { 42 } else { d };
+                x_ptr.add(i).write_volatile(c);
+                y_ptr.add(i).write_volatile(d);
+            }
```
tgross35: Could you comment what is going on here?

Also prefer `MaybeUninit::<U>::zeroed()` (applies more places)

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+            let r: U = __test_roundtrip_{{ ident }}(sz as i32, x.assume_init(), &mut error, pad.as_ptr());
```
tgross35: The `assume_init` isn't sound, it's unlikely but the compiler can do some wonky stuff here. You should replace the `U` in the function signature with `MaybeUninit<U>` and then don't `assume_init`.

I think this applies a few places.

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+            if error == 1 {
+                FAILED.store(true, Ordering::Relaxed);
+                return;
+            }
+            for (i, elem) in pad.iter().enumerate().take(size_of::<U>()) {
+                if *elem == 1 { continue; }
+                let rust = (*y_ptr.add(i)) as usize;
+                let c = (&r as *const _ as *const u8)
+                            .add(i).read_volatile() as usize;
```
tgross35: We have some better ways to handle pointers nowadays: `&raw const r` rather than `as *const _`, `.cast::<Ty>()` rather than `as *const Ty`, and `.addr()` rather than `as usize`

ctest-next/templates/test.c
tgross35: Slightly changing my mind here; could you use `ctest_size_of__{{ ident }}` and `ctest_align_of__{{ ident }}` instead?

Using `__` to separate "fields" is somewhat common. E.g. for `struct val_tree`, `ctest_size_of__val_tree` is a bit easier to tell that you're getting the `ctest_size_of` for `val_tree`, rather than `ctest_size_of_val_tree` which could be read as `ctest_size_of_val` for a struct named `tree`.

ctest-next/templates/test.c
```diff
@@ -24,3 +24,273 @@ static {{ c_type }} __test_const_{{ ident }}_val = {{ c_ident }};
...
+{%- let c_field_name = self.c_ident(MapInput::Field(Either::Left(structure), field)) +%}
+
+uint64_t __test_offset_{{ ident }}_{{ rust_field_name }}(void) {
+    return offsetof({{ c_type }}, {{ c_field_name }});
+}
+
+uint64_t __test_fsize_{{ ident }}_{{ rust_field_name }}(void) {
+    {{ c_type }}* foo = NULL;
+    return sizeof(foo->{{ c_field_name }});
+}
```
tgross35, line 65: Similar to the above, maybe name these `ctest_offset_of__{{ ident }}__{{ rust_field_name }}` and `ctest_field_size__{{ ident }}__{{ rust_field_name }}`

ctest-next/src/template.rs
```diff
@@ -45,35 +50,143 @@ impl<'a> CTestTemplate<'a> {
...
-        self.generator.map(item)
+
+        Ok(self.generator.map(item))
+    }
+
+    /// Modify a C function `signature` that returns a ptr `ty` to be correctly translated.
+    ///
+    /// Arrays and Function types in C have different rules for placement, such as array lengths
+    /// being placed after the parameter list.
+    pub(crate) fn c_signature(
```
tgross35, line 106: I'm not sure I follow this. Could it just turn a `ty` into a `String` single arg? I don't think there is any reason it needs to start with the whole function input.

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+            let uninit_ty = uninit_ty.as_ptr();
+            let ty_ptr = std::ptr::addr_of!((*uninit_ty).{{ field.ident() }});
+            let val = ty_ptr.read_unaligned();
+            check_same(offset_of!({{ ident }}, {{ field.ident() }}),
+                    __test_offset_{{ ident }}_{{ field.ident() }}() as usize,
+                    "field offset {{ field.ident() }} of {{ ident }}");
+            check_same(mem::size_of_val(&val) as u64,
+                    __test_fsize_{{ ident }}_{{ field.ident() }}(),
+                    "field size {{ field.ident() }} of {{ ident }}");
+        }
```
tgross35: * Use `zeroed` rather than `uninit`, basically everywhere (just tends to be slightly more predictable when things go wrong)
* `addr_of!` can now be `&raw const`
* `mem::size_of` -> `size_of`

ctest-next/src/ffi_items.rs
mbyx: The `Visit` trait doesn't return anything, so it would have to be a panic.

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+        unsafe {
+            let mut uninit_ty = std::mem::MaybeUninit::<{{ ident }}>::uninit();
+            let uninit_ty = uninit_ty.as_mut_ptr();
+            let ty_ptr_mut = std::ptr::addr_of_mut!(*uninit_ty);
+            let field_ptr = std::ptr::addr_of!((*uninit_ty).{{ field.ident() }});
+            check_same(field_ptr as *mut _,
+                    __test_field_type_{{ ident }}_{{ field.ident() }}(ty_ptr_mut),
+                    "field type {{ field.ident() }} of {{ ident }}");
+            #[allow(unknown_lints, forgetting_copy_types)]
+            mem::forget(uninit_ty);
```
tgross35: Why is this needed?

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+        }
+        {%- if !self::should_skip_field_type(generator, Either::Left(structure), field) +%}
+
+        extern "C" {
+            fn __test_field_type_{{ ident }}_{{ field.ident() }}(a: *mut {{ ident }}) -> *mut u8;
+        }
+        unsafe {
+            let mut uninit_ty = std::mem::MaybeUninit::<{{ ident }}>::uninit();
+            let uninit_ty = uninit_ty.as_mut_ptr();
+            let ty_ptr_mut = std::ptr::addr_of_mut!(*uninit_ty);
```
tgross35: This shouldn't need to be `mut`, C can just take a const pointer

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+        }
+        unsafe {
+            check_same(mem::size_of::<{{ ident }}>() as u64,
+                __test_size_{{ ident }}(), "{{ ident }} size");
+            check_same(mem::align_of::<{{ ident }}>() as u64,
+                __test_align_{{ ident }}(), "{{ ident }} align");
+        }
+    }
+
+    /// No idea what this does.
```
tgross35: Lol

The first bit is pretty easy: it just checks that the offsets of a field within a struct are the same, and that the field sizes are the same.

The second part makes an instance of the type and returns a pointer to the field, then checks it's the same in Rust. I'm not really sure what this is intended to test since it should be the same as `offset_of!`, but I guess it doesn't hurt to have.

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+        {%- endfor +%}
+        // This vector contains `1` if the byte is padding
+        // and `0` if the byte is not padding.
+        let mut pad = Vec::<u8>::new();
+        // Initialize all bytes as:
+        //  - padding if we have fields, this means that only
+        //  the fields will be checked
+        //  - no-padding if we have a type alias: if this
+        //  causes problems the type alias should be skipped
+        pad.resize(mem::size_of::<{{ ident }}>(), 1);
```
tgross35: Could the type be `Vec<bool>`?

Also, does this even need to be a vector? I think `[true; size_of::<{{ ident }}>]` should work.

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+        }
+        pad
+    }
+
+    /// Tests whether the type alias `x` when passed to C and back to Rust remains unchanged.
+    ///
+    /// It checks if the size is the same as well as if the padding bytes are all in the
+    /// correct place.
+    pub fn roundtrip_{{ ident }}() {
+        use std::ffi::c_int;
```
tgross35: Just make this a module-level import

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+            let sz = size_of::<U>();
+            for i in 0..sz {
+                let c: u8 = (i % 256) as u8;
+                let c = if c == 0 { 42 } else { c };
+                let d: u8 = 255_u8 - (i % 256) as u8;
+                let d = if d == 0 { 42 } else { d };
+                x_ptr.add(i).write_volatile(c);
+                y_ptr.add(i).write_volatile(d);
+            }
+            let r: U = __test_roundtrip_{{ ident }}(sz as i32, x.assume_init(), &mut error, pad.as_ptr());
```
tgross35: Same as earlier, this shouldn't `assume_init`

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+
+        type U = {{ ident }};
+        extern "C" {
+            fn __test_roundtrip_{{ ident }}(
+                size: i32, x: U, e: *mut c_int, pad: *const u8
+            ) -> U;
+        }
+        let pad = roundtrip_padding_{{ ident }}();
+        unsafe {
+            use std::mem::{MaybeUninit, size_of};
```
tgross35: Use a module-level import for `MaybeUninit`

ctest-next/templates/test.rs
mbyx: The imports that are within the test functions are there because they aren't used everywhere, moving them to the top level would cause unused import errors.

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+            }
+        }
+        pad
+    }
+
+    /// Tests whether the type alias `x` when passed to C and back to Rust remains unchanged.
+    ///
+    /// It checks if the size is the same as well as if the padding bytes are all in the
+    /// correct place.
+    pub fn roundtrip_{{ ident }}() {
```
tgross35: Could you add some comments within the function about what is happening?

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+            for i in 0..sz {
+                let c: u8 = (i % 256) as u8;
+                let c = if c == 0 { 42 } else { c };
+                let d: u8 = 255_u8 - (i % 256) as u8;
+                let d = if d == 0 { 42 } else { d };
+                x_ptr.add(i).write_volatile(c);
+                y_ptr.add(i).write_volatile(d);
+            }
+            let r: U = __test_roundtrip_{{ ident }}(sz as i32, x.assume_init(), &mut error, pad.as_ptr());
+            if error == 1 {
```
```suggestion
            if error != 0 {
```

ctest-next/templates/test.rs
```diff
@@ -82,7 +83,430 @@ mod generated_tests {
...
+                FAILED.store(true, Ordering::Relaxed);
+                return;
+            }
+            for (i, elem) in pad.iter().enumerate().take(size_of::<U>()) {
```
tgross35: Isn't `pad` already `size_of::<U>()`? Maybe this is something worth asserting, then you can remove the `.take()`

ctest-next/src/ffi_items.rs
tgross35: I think that's fine, shouldn't happen anyway

ctest-next/templates/test.rs
tgross35: That's fine, just `#[allow(unused_imports)]`

ctest-next/src/template.rs
mbyx: Where the original `signature` is placed depends on what is returned by that function, so we need to pass it as well.

ctest-next/templates/test.rs
mbyx: For `addr_of!` could you give an example?

mbyx:
Why is the style check failing for `libc-test`? I haven't touched that.

tgross35:
Sometimes we get new lints that that suggest changes to existing code

ctest-next/src/ffi_items.rs
```diff
@@ -123,10 +122,32 @@ fn visit_foreign_item_fn(table: &mut FfiItems, i: &syn::ForeignItemFn, abi: &Abi
...
+                }
+            }
+            None
+        }
+        _ => None,
+    });
+
+    if link_name_iter.next().is_some() {
+        panic!("multiple #[link_name = ...] attributes found");
+    }
```
tgross35:
```suggestion
    if let Some(attr) = link_name_iter.next() {
        panic!("multiple `#[link_name = ...]` attributes found: {attr:?}");
    }
```

Easy debugging to help figure out where

ctest-next/Cargo.lock
```diff
@@ -0,0 +1,485 @@
+# This file is automatically @generated by Cargo.
```
tgross35: This file shouldn't be needed since `ctest-next` is in the root workspace (and thus the root `Cargo.lock`)

ctest-next/templates/test.rs
tgross35: `addr_of!(foo.bar)` can basically turn into `&raw const foo.bar`. There is a bit more at the docs for `addr_of!` and for pointers https://doc.rust-lang.org/std/ptr/macro.addr_of.html https://doc.rust-lang.org/std/primitive.pointer.html#3-create-it-using-raw

ctest-next/src/template.rs
tgross35: Do you happen to have a small example?

ctest-next/templates/test.rs
```diff
@@ -5,12 +5,13 @@
 /// are not allowed at the top-level, so we hack around this by keeping it
 /// inside of a module.
 mod generated_tests {
-    #![allow(non_snake_case)]
+    #![allow(non_snake_case, unused_imports)]
```
tgross35, line 8: Put `#[allow(unused_imports)]` on the specific imports that are expected to possibly be unused, to avoid accidentally importing something within a function that won't be used.

ctest-next/templates/test.rs
```diff
@@ -77,15 +77,245 @@ mod generated_tests {
...
+    /// Rust and C sides to see if they are equal.
+    pub fn sign_{{ ident }}() {
+        extern "C" {
+            fn ctest_{{ ident }}_is_signed() -> u32;
+        }
+        let all_ones = !(0 as {{ ident }});
+        let all_zeros = 0 as {{ ident }};
+        unsafe {
+            check_same((all_ones < all_zeros) as u32,
+                    ctest_{{ ident }}_is_signed(), "{{ ident }} signed");
```
tgross35, line 112: Formatting nit

```suggestion
            check_same(
                (all_ones < all_zeros) as u32,
                ctest_{{ ident }}_is_signed(),
                "{{ ident }} signed",
            );
```

ctest-next/templates/test.rs
```diff
@@ -77,15 +77,245 @@ mod generated_tests {
...
+            let uninit_ty = MaybeUninit::<{{ ident }}>::zeroed();
+            let uninit_ty = uninit_ty.as_ptr();
+            let ty_ptr = &raw const ((*uninit_ty).{{ field.ident() }});
+            let val = ty_ptr.read_unaligned();
+            check_same(offset_of!({{ ident }}, {{ field.ident() }}),
+                    ctest_offset_of__{{ ident }}__{{ field.ident() }}() as usize,
+                    "field offset {{ field.ident() }} of {{ ident }}");
+            check_same(size_of_val(&val) as u64,
+                    ctest_field_size__{{ ident }}__{{ field.ident() }}(),
+                    "field size {{ field.ident() }} of {{ ident }}");
```
tgross35, line 159:
```suggestion
            check_same(
                offset_of!(ident, {{ field.ident() }}) as u64,
                ctest_offset_of__{{ ident }}__{{ field.ident() }}(),
                "field offset {{ field.ident() }} of {{ ident }}",
            );
            check_same(
                size_of_val(&val) as u64,
                ctest_field_size__{{ ident }}__{{ field.ident() }}(),
                "field size {{ field.ident() }} of {{ ident }}",
            );
```

Changing the first `offset_of` to cast to `u64` rather than turning the C call into `usize`, and formatting

ctest-next/templates/test.rs
tgross35: Bind the C call as `let c_is_signed = unsafe { ctest_{{ ident }}_is_signed() };`, that way the rest of this doesn't need to be `unsafe`

ctest-next/templates/test.rs
tgross35: The `read_unaligned` also isn't sound because it's possible that the type doesn't allow all zeros.

I don't want to delay this PR further but this does need to get changed in a followup, so please add a `// FIXME(soundness): cast to a byte array before reading`

ctest-next/templates/test.c
```diff
@@ -24,3 +24,97 @@ static {{ c_type }} __test_const_{{ ident }}_val = {{ c_ident }};
...
+    return offsetof({{ c_type }}, {{ c_field_name }});
+}
+
+uint64_t ctest_field_size__{{ ident }}__{{ rust_field_name }}(void) {
+    {{ c_type }}* foo = NULL;
+    return sizeof(foo->{{ c_field_name }});
+}
+
+{%- if !self::should_skip_struct_field_type(generator, structure, field) +%}
+{%- let signature = self.c_signature(field.ty, &format!("__test_field_type_{ident}_{rust_field_name}({c_type}* b)")).unwrap() +%}
```
tgross35, line 68: I think this function could be `ctest_get_pointer_to__{ident}__{rust_field_name}`

ctest-next/templates/test.rs
```diff
@@ -77,15 +77,245 @@ mod generated_tests {
...
+        extern "C" {
+            fn __test_field_type_{{ ident }}_{{ field.ident() }}(a: *const {{ ident }}) -> *mut u8;
+        }
+        unsafe {
+            let uninit_ty = MaybeUninit::<{{ ident }}>::zeroed();
+            let ty_ptr = uninit_ty.as_ptr();
+            let field_ptr = &raw const ((*ty_ptr).{{ field.ident() }});
+            check_same(field_ptr as *mut _,
+                    __test_field_type_{{ ident }}_{{ field.ident() }}(ty_ptr),
+                    "field type {{ field.ident() }} of {{ ident }}");
```
tgross35, line 172: There are some things I don't like about this `__test_field_type_` test but I don't want to block the PR on it. Could you just comment this test (both here and in C) with a FIXME? This applies to the later check of unions as well, and should also be able to comment out `c_signature` which I think should look a bit different.

ctest-next/templates/test.rs
```diff
@@ -77,15 +77,245 @@ mod generated_tests {
...
+        let mut v = Vec::<(usize, usize)>::new();
+        let bar = MaybeUninit::<{{ ident }}>::zeroed();
+        let bar = bar.as_ptr();
+        {%- else +%}
+        let v = Vec::<(usize, usize)>::new();
+        {%- endif +%}
+        {%- for field in fields +%}
+        unsafe {
+            let ty_ptr = &raw const ((*bar).{{ field.ident() }});
+            let val = ty_ptr.read_unaligned();
```
tgross35, line 201: Add a `// FIXME(soundness): ...`

ctest-test/src/t1.h
```diff
@@ -95,31 +100,22 @@ extern int32_t T1_arr5[1][2][3];
 
 extern int32_t T1_arr42[1][2][3];
 
-extern const int16_t* T1_sref;
-
-extern const int32_t* T1_mut_opt_ref;
-extern int32_t* T1_mut_opt_mut_ref;
-extern const int32_t* T1_const_opt_const_ref;
-
-extern void (*const T1_opt_fn1)(void);
```
tgross35, line 104: Why are these removed?

If something isn't yet working, comment it with a FIXME

ctest-test/src/t1.rs
```diff
@@ -88,7 +89,8 @@ extern "C" {
 }
 
 pub fn foo() {
-    assert_eq!(1, 1);
+    let x = 1;
+    assert_eq!(x, 1);
```
tgross35, line 93: Why is this needed?

ctest-test/src/t1.h
```diff
@@ -95,31 +100,22 @@ extern int32_t T1_arr5[1][2][3];
...
-/* FIXME(#4365): duplicate symbol errors when enabled
-// uint32_t (*(*T1_opt_fn2)(uint8_t))(uint16_t);
-// uint32_t (*(*T1_opt_fn3)(uint8_t(*)(uint8_t), uint16_t(*)(uint16_t)))(uint16_t);
```
tgross35, line 107: If this is no longer accurate, just uncomment the code so we can verify the test passes

ctest-test/src/t1.rs
```diff
@@ -195,6 +179,7 @@ struct timeval {
     tv_usec: c_int,
 }
 
+#[expect(unused)]
 #[repr(C)]
 struct log_record_t {
```
tgross35, line 184: Make them `pub` instead of `expect(unused)`

ctest-test/tests/all.rs
```diff
@@ -26,69 +28,20 @@ fn output(cmd: &mut Command) -> (String, ExitStatus) {
...
-
-    for error in errors {
-        println!("didn't find error: {error}");
-        bad = true;
-    }
-    if bad {
-        println!("output was:\n\n{o}");
-        panic!();
-    }
-}
```
tgross35, line 85: What happened to this block?

ctest-next/templates/common/test_roundtrip.c
```diff
@@ -0,0 +1,48 @@
...
+            = (unsigned char)(255) - (unsigned char)(i % 256);
+        d = d == 0? 42: d;
+        p[i] = d;
+    }
+    return value;
+}
+
+#ifdef _MSC_VER
+#  pragma warning(default:4365)
+#endif
```
tgross35, line 48: Missing newline at end of file

Same for the other files in this directory

ctest-next/src/template.rs
```diff
@@ -45,35 +49,170 @@ impl<'a> CTestTemplate<'a> {
...
     }
 }
+
+/* Helper functions to make the template code readable. */
+
+/// Determine whether a Rust alias/struct/union should have a round trip test.
+///
+/// By default all alias/struct/unions are roundtripped. Aliases or fields with arrays should
+/// not be part of the roundtrip.
+pub(crate) fn should_roundtrip(gen: &TestGenerator, ident: &str) -> bool {
```
tgross35, line 186: Could you define what "round trip" means here?

ctest-test/src/t1.rs
mbyx: It triggered a clippy lint about identical args used in `assert_eq`, it's the only lint left that triggers so I thought it would be better to fix it. Although the CI currently skips checking ctest-test I think.

ctest-test/src/t1.h
mbyx: This was before I learnt that `Option` was FFI safe in some cases, so I removed them. I'll add them back.

ctest-test/build.rs
```diff
@@ -1,107 +1,62 @@
-use std::process::Command;
```
tgross35, line 1: Could `ctest_next` be added to the tests without removing `ctest`? Or do the updates to tests make something fail?

ctest-test/tests/all.rs
mbyx: It was duplicated across the C and C++ test, nothing has been changed here.

ctest-test/build.rs
mbyx: `t2` would fail because `ctest` would fail on more errors than `ctest-next` since statics and functions haven't been implemented yet.

tgross35: Can't you use the skips?

mbyx: Unfortunately it's not even able to parse `pub const T1S: *const c_char = c"foo".as_ptr();` so a skip wouldn't work.

ctest-test/src/t1.h
mbyx: Functions and statics aren't tested right now, so for now uncommenting only confirms that they parse correctly. `ctest` would still fail if they're uncommented.

tgross35:
This PR is probably getting a bit too big, I think I need to look at some of these tests closer. Nothing related to what you are adding, it just seems like the existing `ctest` does some things that are questionable.

Would you be willing to split up this PR? It would be easier if we could land tests one at a time without delaying everything here because specific tests need more updates. Specifically, I'd suggest splitting up into something like one PR for each of the following:

1. (new) Move logic out of templates. I did some experimentation before suggesting this, you can basically take https://github.com/tgross35/rust-libc/commit/0fcec9d3955a874fc557f0299da1ce356399d26f with a few tweaks to get it to build (drop the new size+align test to keep an absolute minimum amount of work here.

    Basically the goal is to avoid anything other than `for` loops in the templates - no `let` bindings or function calls - which will make it easier to work on a single test at a time. This is easier to read and work on anyway.
1. (from this PR) add changes in `ctest-next/src/generator.rs` that allow skipping tests.
1. (from this PR) add `ctest-next` to `ctest-test`, in parallel with `ctest`. Needs (2) so you can skip everything that isn't yet implemented
1. (from this PR) add the size+align tests to the template to all types. Needs (1)
1. (from this PR) add field offset tests to the template for unions and structs. Needs (1)
1. (from this PR) add roundtrip tests to the template. Needs (1).
1. (from this PR) ... other tests to the template
1. (new, I know you're working on it) update `libc-test` to run both `ctest` and `ctest-next`. This needs (2) so you can skip things that aren't implemented

This way I don't think any work is blocked after 1 and 2 are merged, the rest of the tests can be improved before they get in.

ctest-test/build.rs
tgross35: Ah, I guess `skip_function` and `skip_static` haven't been added yet. If you add them then you should be able to `cfg.skip_function(|_f| true); cfg.skip_static(|_s| true)` in the tests until it works, without blocking on everything else.

The same is true for `libc-test`

tgross35:
@mbyx I think this one can be closed too, right? Looks like a lot of conflicts at this point so I'm not sure what would remain.

PR #4544, opened 2025-07-09
Revert of mut api breaking userspace users on AIX

entd:
# Description

Reverted C api changes introduced by commit c192a5cae31dd84ad17dc3c2dd0a3deaacae452a

IBM AIX does not define pointer constness traditionally in their headers. Some commit changed const pointers to mutable ones, those breaking many userspace libraries, as those in rust libc are usually defined on other posix platforms as const.

Some older commit (mainly c192a5cae31dd84ad17dc3c2dd0a3deaacae452a) changed the api.
I reverted fstat, swapon apis, to expose path variable pointer as const again.

# Sources

https://www.ibm.com/docs/en/aix/7.2.0?topic=s-stat-fstat-lstat-statx-fstatx-statxat-fstatat-fullstat-ffullstat-stat64-fstat64-lstat64-stat64x-fstat64x-lstat64x-stat64xat-subroutine)

# Checklist

### DIFF
--- a/src/unix/aix/mod.rs
+++ b/src/unix/aix/mod.rs
@@ -3246,7 +3246,7 @@ extern "C" {
     #[link_name = "nsendmsg"]
     pub fn sendmsg(sockfd: c_int, msg: *const msghdr, flags: c_int) -> ssize_t;
     pub fn setcontext(ucp: *const ucontext_t) -> c_int;
-    pub fn setdomainname(name: *mut c_char, len: c_int) -> c_int;
+    pub fn setdomainname(name: *const c_char, len: c_int) -> c_int;
     pub fn setgroups(ngroups: c_int, ptr: *mut crate::gid_t) -> c_int;
     pub fn setgrent();
     pub fn setmntent(filename: *const c_char, ty: *const c_char) -> *mut crate::FILE;
@@ -3254,7 +3254,7 @@ extern "C" {
     pub fn setpwent();
     pub fn setrlimit(resource: c_int, rlim: *const crate::rlimit) -> c_int;
     pub fn setrlimit64(resource: c_int, rlim: *const rlimit64) -> c_int;
-    pub fn settimeofday(tv: *mut crate::timeval, tz: *mut crate::timezone) -> c_int;
+    pub fn settimeofday(tv: *const crate::timeval, tz: *mut crate::timezone) -> c_int;
     pub fn setitimer(
         which: c_int,
         new_value: *const crate::itimerval,
@@ -3306,8 +3306,8 @@ extern "C" {
     pub fn strptime(s: *const c_char, format: *const c_char, tm: *mut crate::tm) -> *mut c_char;
     pub fn strsep(string: *mut *mut c_char, delim: *const c_char) -> *mut c_char;
     pub fn swapcontext(uocp: *mut ucontext_t, ucp: *const ucontext_t) -> c_int;
-    pub fn swapoff(puath: *mut c_char) -> c_int;
-    pub fn swapon(path: *mut c_char) -> c_int;
+    pub fn swapoff(puath: *const c_char) -> c_int;
+    pub fn swapon(path: *const c_char) -> c_int;
     pub fn sync();
     pub fn telldir(dirp: *mut crate::DIR) -> c_long;
     pub fn timer_create(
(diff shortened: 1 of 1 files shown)


entd:
I am just third party user not affiliated to the IBM. Maybe IBM maintainers her could tell better. I think it might be leftover from pre ANSI C, not that those headers are require non const pointers in C.

For me biggest problem, that many rust user libs expect consts here. Like simple networking libraries using c:statfs path fails because AIX backend in this library nothis does not wrap `const*` and expose `mut*`. 
As for forward compatibility I think mut* can be safely coerced to const* and those there should not be huge issues. 
The real decider should does AIX internally handle those path pointers as mutable or it is just leftover and mess from the pre-ANSI days.
If actually those are indeed need to be mutable we probably should wrap the function with string copy to have consistent api for external libraries.

Maybe @xingxue-ibm OP of the original commit can guide us here?

tgross35:
To clarify: the best case scenario here is that AIX maintainers (1) confirm the arguments can safely be pointers to read-only data, and (2) are willing to update the documentation / headers to reflect that. `const char * Path` is used for `stat` on the doc page linked in the top post so I would assume `statx` could/should be similar.

I am willing to merge this PR with only the maintainer confirmation, the docs update would just be nice to make things more clear for the future (and of course to help other consumers of those docs).

> If actually those are indeed need to be mutable we probably should wrap the function with string copy to have consistent api for external libraries. I feel this change should be in this library, not offloaded to the huge userspace which now would have write absurd amount AIX specific code just to accomodate one mut* not maching most of POSIX world.

This would not happen, `libc` provides bindings but does not abstract over platform differences. There are other libraries such as `nix` that provide higher-level interfaces.

entd:
There is worse type mismatches than consts, example ioctls are broken as constants now are long, but ioctl symbol expects int...:

`error[E0308]: mismatched types
    --> /home/ibrasiskis/.cargo/registry/src/index.crates.io-7f555b6b8ccf4919/jobserver-0.1.33/src/unix.rs:356:57
     |
356  |         cvt(unsafe { libc::ioctl(self.read.as_raw_fd(), libc::FIONREAD, len.as_mut_ptr()) })?;
     |                      -----------                        ^^^^^^^^^^^^^^ expected `i32`, found `i64`
     |                      |
     |                      arguments to this function are incorrect
     |
note: function defined here
    --> /home/ibrasiskis/.cargo/registry/src/index.crates.io-7f555b6b8ccf4919/libc-0.2.174/src/unix/aix/mod.rs:2994:12
     |
2994 |     pub fn ioctl(fildes: c_int, request: c_int, ...) -> c_int;
     |            ^^^^^
help: you can convert an `i64` to an `i32` and panic if the converted value doesn't fit
     |
356  |         cvt(unsafe { libc::ioctl(self.read.as_raw_fd(), libc::FIONREAD.try_into().unwrap(), len.as_mut_ptr()) })?;
     |                                                                       ++++++++++++++++++++

For more information about this error, try `rustc --explain E0308`.`

xingxue-ibm:
I’m very sorry for the delay (I was out of town). I’ve asked the AIX team whether the affected arguments truly need to be mutable, and I’ll keep you posted.

Regarding the type mismatch, that was my mistake when handling sign extension in constants created from macros in AIX headers. I’ll submit a PR with the fix shortly. My apologies for the error.

xingxue-ibm:
The AIX team confirmed the affected arguments are not actually modified. I've created PR [Add 'const' to some signatures for AIX](https://github.com/rust-lang/libc/pull/4563) to make the change.

The constants definitions for `ioctl()` issue is a bit tricky. The signature of `ioctl()` on AIX is
```
int ioctl(int FileDescriptor, int Command, void *Argument)
```
and `Command` uses constants from macro such as `_IOW`
```
#define _IOW(x,y,t)     (IOC_IN|((sizeof(t)&IOCPARM_MASK)<<16)|(x<<8)|y)
```
In 32-bit mode, the resulting value is 32-bit because `sizeof()` returns a 32-bit type. In 64-bit mode, the value becomes 64-bit, as `sizeof()` returns an `unsigned long long`. AIX relies on the C compiler to cast this value to int for the Command argument. However, defining these constants as `c_int` in Rust causes a mismatch with the C side during CI checks.

Aside from skipping these constants in CI (there are quite a few), are there better approaches? Any guidance would be greatly appreciated!

Solaris defines the macro as the following which makes the resulting value `int` regardless of the bit mode:
```
#define _IOW(x, y, t)   (IOC_IN | (((int)sizeof(t) & IOCPARM_MASK) << 16) | ((x) << 8) | (y))
```

xingxue-ibm:
Posted [Fix the types of constants used for the 'request' argument of ioctl() on AIX](https://github.com/rust-lang/libc/pull/4582) for the `FIONREAD` issue mentioned above.

PR #4546, opened 2025-07-13
Add some constants for Linux nl80211

a-gavin:
# Description

Add some of the many nl80211 constants required for WiFi configuration using the Netlink kernel interface.

Includes the following enums:
- nl80211_commands
- nl80211_attrs
- nl80211_iftype
- nl80211_mesh_setup_params
- nl80211_channel_type
- nl80211_chan_width
- nl80211_bss
- nl80211_auth_type
- nl80211_scan_flags

Includes '#define'ed commands or attrs defined for legacy compatibility but does not include those defined for programs to '#ifdef'

# Sources

To the best of my knowledge, these constants are accurate to the linux-wireless tree at tag 'wireless-next-2025-06-25' available [here](https://web.git.kernel.org/pub/scm/linux/kernel/git/wireless/wireless-next.git/). Almost all of these are available in Linus' upstream tree (and those that aren't will be as of 6.16 when it is released).

# Checklist

- [Y] Relevant tests in `libc-test/semver` have been updated
- [Y] No placeholder or unstable values like `*LAST` or `*MAX` are
- [Y] Tested locally (`cd libc-test && cargo test --target mytarget`);


a-gavin:
I understand now that this project pulls in linting from nightly, so once the PR to fix the style check failure lands ([here](https://github.com/rust-lang/libc/pull/4548)), I'll rebase my changes to re-run the CI suite.

a-gavin:
Leaving this PR open while I discuss with the maintainer of `neli`. Understood on this projects goal of not adding more "niche" values, especially when something like `linux-sys-raw` can add support for this so easily (and using automation).

PR #4551, opened 2025-07-16
ctest: extract bindings from template into source

mbyx:
# Description
Extracts out everything except for bindings from the test template, so that each test is parameterized by a single struct, based on this [commit](https://github.com/tgross35/rust-libc/commit/0fcec9d3955a874fc557f0299da1ce356399d26f).

# Sources

# Checklist


mbyx:
I think it would be a good idea to layout some naming rules from here on out, for example what to name any new tests, whether their C equivalents should have the same name or a slightly modified name, etc.

ctest-next/templates/test.rs
```diff
@@ -42,51 +44,47 @@ mod generated_tests {
...
     // Test that the string constant is the same in both Rust and C.
     // While fat pointers can't be translated, we instead use * const c_char.
-    pub fn const_{{ ident }}() {
+    pub fn {{ const_cstr.test_ident }}() {
         extern "C" {
-            fn __test_const_{{ ident }}() -> *const *const u8;
+            fn __{{ const_cstr.test_ident }}() -> *const *const u8;
         }
-        let val = {{ ident }};
+        let val = {{ const_cstr.rust_ident }};
```
tgross35, line 55: Leave this one as `ctest_const_cstr__{{ const_cstr.ident }}`

ctest-next/src/template.rs
```diff
@@ -2,78 +2,185 @@ use askama::Template;
...
+                if is_const_c_char_ptr {
+                    let item = TestCstr {
+                        test_ident: cstr_test_ident(constant.ident()),
+                        rust_ident: constant.ident().into(),
+                        c_ident: helper.c_ident(constant).into(),
+                        c_type: helper.c_type(constant)?.into(),
+                    };
+                    const_cstr_tests.push(item)
+                }
+            } else {
```
tgross35, line 90: This doesn't add any test in the case where it is a pointer but not a `CStr`. Either make the top-level `if` also check for the pointer type so the nested `if` isn't needed:

```rust
matches!(
    &constant.ty,
    syn::Type::Ptr(syn::Type::Path(path))
        if path.path.segments.last().unwrap().ident == "c_char"
        && ptr.mutability.is_none()
);
```

Or bump the MSRV by 1 again and use let chains to do the same thing but slightly cleaner.

ctest-next/src/template.rs
```diff
@@ -2,78 +2,185 @@ use askama::Template;
...
     }
+}
+
+/// Information required to test a constant CStr.
+#[derive(Clone, Debug)]
+pub(crate) struct TestCstr {
+    pub(crate) test_ident: BoxStr,
+    pub(crate) rust_ident: BoxStr,
+    pub(crate) c_ident: BoxStr,
+    pub(crate) c_type: BoxStr,
```
tgross35, line 120:
```suggestion
    pub test_ident: BoxStr,
    pub rust_ident: BoxStr,
    pub c_ident: BoxStr,
    pub c_type: BoxStr,
```

Optional nit: if a struct isn't public, `pub` (without crate) is fine to get the same visibility as the struct itself

ctest-next/templates/test.rs
```diff
@@ -42,51 +44,47 @@ mod generated_tests {
...
-    {%- else +%}
+    {%- for constant in ctx.const_tests +%}
 
     // Test that the value of the constant is the same in both Rust and C.
     // This performs a byte by byte comparision of the constant value.
-    pub fn const_{{ ident }}() {
+    pub fn {{ constant.test_ident }}() {
         extern "C" {
-            fn __test_const_{{ ident }}() -> *const {{ ty }};
+            fn __{{ constant.test_ident }}() -> *const {{ constant.rust_type }};
```
tgross35, line 73: Same, `ctest_const__{{ const_cstr.ident }}` is fine

tgross35:
My feedback isn't too significant, I'm going to merge it then add a followup commit

tgross35:
Following this up with a few fixes in https://github.com/rust-lang/libc/pull/4554

PR #4552, opened 2025-07-16
FreeBSD ktls fixes

neuschaefer:
# Description

This PR fixes two minor incompatibilities that I found by running `libc-test` on FreeBSD 15-CURRENT on ppc64 (big endian).

# Sources

https://github.com/freebsd/freebsd-src/commit/c9e9a0fe5b0f88561f55fb2f6f5354fbbd96dd5d

# Checklist

- [x] ~~Relevant tests in `libc-test/semver` have been updated~~ (no symbols changed)
- [x] No placeholder or unstable values like `*LAST` or `*MAX` are
  included (see [#3131](https://github.com/rust-lang/libc/issues/3131))
- [x] Tested locally (`cd libc-test && cargo test --target mytarget`);
  especially relevant for platforms that may not be checked in CI

PR #4553, opened 2025-07-16
freebsd15: Add ki_uerrmsg to struct kinfo_proc

neuschaefer:
# Description

During the development of FreeBSD 15, a field was added to `kinfo_proc`. With this PR and https://github.com/rust-lang/libc/pull/4552, the `libc-test` suite passes on FreeBSD 15 on ppc64.

# Sources

Upstream commit: https://github.com/freebsd/freebsd-src/commit/7212b37345936899e979c63d0b054e114576faa0

# Checklist

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/freebsd15/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/freebsd15/mod.rs
@@ -50,6 +50,7 @@ s! {
         _priv: [c_ulong; 8],
     }
 
+    #[non_exhaustive]
     pub struct kinfo_proc {
         /// Size of this structure.
         pub ki_structsize: c_int,


asomers:
This change only affects FreeBSD 15.  That release is still under development, and therefore no backwards-compatibility concerns apply.  So yes I think we should merge it to libc-0.2.

asomers (approved):

tgross35:
Thanks for confirming, that's perfect. I think we should still add `#[non_exhaustive]` to allow changing padding fields into named fields in the future.

neuschaefer:
> could you just make the struct `#[non_exhaustive]`? We're starting to do this for everything new, may as well do it with the change here as well

commit added.

PR #4555, opened 2025-07-17
ctest: add support for skipping more items and prepare for ctest-test port

mbyx:
# Description
Adds better support for skipping unions, fields, and tests for field types. Additionally makes needed changes to other parts of `ctest-next` so that the port PRs for `ctest-test` and `libc-test` don't have too many changes to `ctest-next`.

# Sources

# Checklist


ctest-next/src/generator.rs
```diff
@@ -453,7 +551,82 @@ impl TestGenerator {
...
+    ///
+    /// ```no_run
+    /// use ctest_next::TestGenerator;
+    ///
+    /// let mut cfg = TestGenerator::new();
+    /// cfg.skip_struct_field_type(|s, field| {
+    ///     s.ident() == "foo_t" || (s.ident() == "bar_t" && field.ident() == "bar")
+    /// });
+    /// ```
+    pub fn skip_struct_field_type(
```
tgross35, line 575: Could you add something like this to the docs?

> This is useful when, for some reason, a field type is intentionally not the same in C and Rust.

Same thing for union

ctest-next/src/generator.rs
```diff
@@ -581,6 +801,29 @@ impl TestGenerator {
...
+    /// # Examples
+    /// ```no_run
+    /// use ctest_next::TestGenerator;
+    ///
+    /// let mut cfg = TestGenerator::new();
+    /// cfg.skip_roundtrip(|s| {
+    ///     s.starts_with("foo_")
+    /// });
+    /// ```
+    pub fn skip_roundtrip(&mut self, f: impl Fn(&str) -> bool + 'static) -> &mut Self {
```
tgross35, line 830: Could you add something like this:

```rust
// FIXME(ctest): should arrays be handled differently?
```

Since we know when a type is an array, we should be able to handle it differently somewhere.

ctest-next/src/macro_expansion.rs
```diff
@@ -6,16 +6,29 @@ use std::process::Command;
...
+
+    // `libc` uses non standard cfg flags as well, which have to be manually expanded.
+    for (k, v) in cfg {
+        match v {
+            None => {
+                cmd.arg("--cfg").arg(k);
+            }
+            Some(val) => {
+                cmd.arg("--cfg").arg(format!("{k}=\"{val}\""));
+            }
```
tgross35: Small formatting tweak

```suggestion
            None => cmd.arg("--cfg").arg(k),
            Some(val) => cmd.arg("--cfg").arg(format!("{k}=\"{val}\"")),
```

ctest-next/src/runner.rs
```diff
@@ -60,6 +68,14 @@ pub fn generate_test(
         cfg.include(p);
     }
 
+    for flag in &generator.flags {
+        cfg.flag(flag);
+    }
+
+    for (a, b) in &generator.defines {
+        cfg.define(a, b.as_ref().map(|s| &s[..]));
+    }
```
tgross35: Maybe `(a, b)` -> `(k, v)` to hint what they are?

ctest-next/src/template.rs
```diff
@@ -168,17 +168,25 @@ impl<'a> TranslateHelper<'a> {
 
     /// Returns the equivalent C/Cpp type of the Rust item.
     pub(crate) fn c_type(&self, item: impl Into<MapInput<'a>>) -> Result<String, TranslationError> {
-        let item: MapInput = item.into();
+        let item: MapInput<'a> = item.into();
```
tgross35: Not that it matters, but the lifetime annotation isn't required is it?

ctest-next/src/translator.rs
```diff
@@ -31,6 +31,13 @@ impl TranslationError {
     }
 }
 
+impl From<TranslationError> for askama::Error {
+    fn from(err: TranslationError) -> Self {
+        // FIXME(ctest): Loses information due to Span not being Send+Sync.
+        askama::Error::Custom(err.kind.into())
```
tgross35: If this is a problem, replace `span: Span` with `location: BoxStr`, and `format!` the span in `TranslationError::new` so the span doesn't need to be in the error type.

tgross35: Or, since `proc_macro` doesn't support formatting for the `Span` yet, move the comments and make the default something like `BoxStr::from("<unknown>")` for now.

tgross35 (commented):
Few small requests but mostly looks pretty good to me

ctest-next/src/translator.rs
```diff
@@ -26,24 +27,29 @@ impl TranslationError {
         Self {
             kind,
             source: source.to_string(),
-            span,
+            span: format!(
+                "line {}, column {}, in {}",
```
tgross35: It's kind of standard to do `foo.rs:10:20`, some terminals will hyperlink that for you

```rust
"{fname}:{line}:{col}",
// FIXME(ctest): ...
fname = "<unknown file>"
line = span.start().line,
col = span.start().column,
```

tgross35 (approved):
Awesome, thanks!

PR #4545, opened 2025-07-10
linux: Add EXEC_RESTRICT_FILE and EXEC_DENY_INTERACTIVE securebits

Gelbpunkt:
# Description

These were added in 6.14.

# Sources

https://github.com/torvalds/linux/commit/a0623b2a1d595341971c189b90a6b06f42cd209d

# Checklist

- [x] Relevant tests in `libc-test/semver` have been updated
- [x] No placeholder or unstable values like `*LAST` or `*MAX` are
  included (see [#3131](https://github.com/rust-lang/libc/issues/3131))
- [x] Tested locally (`cd libc-test && cargo test --target mytarget`);
  especially relevant for platforms that may not be checked in CI

PR #4557, opened 2025-07-18
Add SECURE_* and SECBIT_* macros added in Linux 6.14.

collinfunk:
The following macros were added in Linux 6.14 [1].

[1] https://github.com/torvalds/linux/commit/a0623b2a1d595341971c189b90a6b06f42cd209d


tgross35:
It looks like this might have raced with another patch that adds these https://github.com/rust-lang/libc/pull/4545. Sorry about that, but thank you for the PR!

collinfunk:
> It looks like this might have raced with another patch that adds these #4545. Sorry about that, but thank you for the PR!

Ah, I looked for a pull request for `SECURE_`, but they labeled theirs as `EXEC_` which is also correct. No worries, thanks!

PR #4558, opened 2025-07-18
ctest: test ctest-next in ctest-test

mbyx:
# Description
- Adds support for testing `ctest-next` using the same test crate as `ctest`.
- Additionally fixes some bugs in the template (unnecessary unsafe, and missing `c_char` import).
- Removes tests for `&str`, they can be added back in with separate constants that are skipped by `ctest-next` if needed.
- Adds `link_name` support for functions and statics, which goes unused right now but will be used when tests for functions and statics are implemented.
- Removes the dependency on `libc`, as `ctest-next` cannot resolve it (can be fixed later, although it isn't really necessary anywhere)

# Sources

# Checklist

### DIFF
--- a/ctest-next/src/ffi_items.rs
+++ b/ctest-next/src/ffi_items.rs
@@ -3,7 +3,7 @@ use std::ops::Deref;
 use syn::punctuated::Punctuated;
 use syn::visit::Visit;
 
-use crate::{Abi, Const, Field, Fn, Parameter, Static, Struct, Type, Union};
+use crate::{Abi, BoxStr, Const, Field, Fn, Parameter, Static, Struct, Type, Union};
 
 /// Represents a collected set of top-level Rust items relevant to FFI generation or analysis.
 ///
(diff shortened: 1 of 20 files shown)


mbyx:
Unfortunately, I have no idea what the errors mean.

mbyx:
Fixes the previous error, but a new one pops up, looks like it's a mix of forgetting to ignore 2024 edition crates for targets that don't support it, and the fact that the old `ctest` cannot parse new syntax.

If we want to keep both we'll probably have to keep them separate as they were before. The edition was changed to 2024 since `ctest-next` is a dependency.

ctest-next/src/ffi_items.rs
```diff
@@ -94,6 +94,30 @@ fn collect_fields(fields: &Punctuated<syn::Field, syn::Token![,]>) -> Vec<Field>
...
+            None
+        }
+        _ => None,
+    });
+
+    if let Some(attr) = link_name_iter.next() {
+        panic!("multiple `#[link_name = ...]` attributes found: {attr:?}");
+    }
+
+    link_name
```
tgross35: This can be simplified a bit to return early if there is no `link_name`:

```rust
let link_name = link_name_iter.next()?;
if let Some(attr) = link_name_iter.next() {
    panic!("multiple `#[link_name = ...]` attributes found: {attr:?}");
}

if let syn::Meta::NameValue(nv) = &attr.meta
    && let syn::Expr::Lit(expr_lit) = &nv.value
    && let syn::Lit::Str(lit_str) = &expr_lit.lit {
    return Some(lit_str.value().into_boxed_str());
}

panic!("unrecognized `link_name` syntax: {attr:?}");
```

ctest-next/templates/test.rs
```diff
@@ -9,9 +9,10 @@
 mod generated_tests {
     #![allow(non_snake_case)]
     #![deny(improper_ctypes_definitions)]
-    use std::ffi::CStr;
+    use std::ffi::{CStr, c_char};
     use std::fmt::{Debug, LowerHex};
     use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+    #[expect(unused_imports)]
     use std::{mem, ptr, slice};
```
tgross35: I think this should be `allow` if the imports are unused only for specific expansions, or delete `mem`/`ptr`/`slice` if one of those is always unused.

ctest-test/src/t2.h
```diff
@@ -17,7 +17,8 @@ typedef struct {
   int64_t b;
 } T2Union;
 
-static void T2a(void) {}
+// FIXME(ctest): Cannot be uncommented until tests for functions are implemented in ctest-next.
+// static void T2a(void) {}
```
tgross35, line 21: Can the build.rs instead get a `.skip_fn(...)` with a FIXME?

mbyx: Unfortunately no, the goal of the function ptr check that hasn't been implemented yet is to make sure that the `T2a` in the header and the `T2a` in the C file point to different items, since `T2a` is a complete function, not just a declaration, it counts as unused regardless of being skipped or not.

PR #4560, opened 2025-07-23
ctest: Add roundtrip test.

mbyx:
# Description
Adds a roundtrip test to `ctest-next`.

# Sources

# Checklist


ctest-next/src/template.rs
```diff
@@ -193,6 +270,15 @@ impl TestTemplate {
  * - `c_ty`: The C type of the constant, qualified with `struct` or `union` if needed.
  */
 
+#[derive(Clone, Debug)]
+pub(crate) struct TestRoundtrip {
```
tgross35: Can this go at the end, to keep these structs in roughly the same order as the test template?

Same for `roundtrip_test_ident`

ctest-next/src/template.rs
```diff
@@ -180,6 +182,81 @@ impl TestTemplate {
...
+                        .clone()
+                        .into_iter()
+                        .filter(|f| f.public)
+                        .collect(),
+                    c_ty: helper.c_type(union_)?.into(),
+                    is_alias: false,
+                };
+                self.roundtrip_tests.push(item.clone());
+                self.test_idents.push(item.test_name);
+            }
```
tgross35: All three items do about the same thing here. Maybe extract this to a new method:

```rust
fn add_roundtrip_test(&mut self, ident: &str, fields: &[Field], c_ty: &str, is_alias: bool) {
    // ...
}
```

ctest-next/src/template.rs
```diff
@@ -180,6 +182,81 @@ impl TestTemplate {
...
+            if !should_skip_roundtrip_test {
+                let item = TestRoundtrip {
+                    test_name: roundtrip_test_ident(union_.ident()),
+                    id: union_.ident().into(),
+                    fields: union_
+                        .fields
+                        .clone()
+                        .into_iter()
+                        .filter(|f| f.public)
+                        .collect(),
```
tgross35: You are duplicating the entire `fields` vec here, then iterating on it and recollecting to a new vec. Instead, you should do `.filter(...).cloned().collect()` which will only construct one new vec and not do any cloning for non-public items.

ctest-next/templates/test.c
```diff
@@ -53,3 +53,53 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+    int size = (int)sizeof({{ item.c_ty }});
+    if (size != rust_size) {
+        fprintf(
+            stderr,
+            "size of {{ item.c_ty }} is %d in C and %d in Rust\n",
+            (int)size, (int)rust_size
+        );
+        *error = 1;
+        return value;
+    }
```
tgross35: If possible, more should be done on the Rust side than the C side. Can the Rust function that calls `ctest_roundtrip__...` first call `ctest_size_of_...` and assert it matches?

ctest-next/templates/test.c
```diff
@@ -53,3 +53,53 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+#  pragma warning(disable:4365)
+#endif
+{%- for item in ctx.roundtrip_tests +%}
+
+// Tests whether the struct/union/alias `x` when passed to C and back to Rust remains unchanged.
+// It checks if the size is the same as well as if the padding bytes are all in the correct place.
+{{ item.c_ty }} ctest_roundtrip__{{ item.id }}(
+    int32_t rust_size, {{ item.c_ty }} value, int* error, unsigned char* pad
+) {
+    volatile unsigned char* p = (volatile unsigned char*)&value;
```
tgross35: Add a note about why this is volatile

ctest-next/templates/test.c
```diff
@@ -53,3 +53,53 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+#ifdef _MSC_VER
+// Disable signed/unsigned conversion warnings on MSVC.
+// These trigger even if the conversion is explicit.
+#  pragma warning(disable:4365)
+#endif
+{%- for item in ctx.roundtrip_tests +%}
+
+// Tests whether the struct/union/alias `x` when passed to C and back to Rust remains unchanged.
+// It checks if the size is the same as well as if the padding bytes are all in the correct place.
+{{ item.c_ty }} ctest_roundtrip__{{ item.id }}(
```
tgross35: Mention specifically that this is when the type is passed and returned by value

ctest-next/templates/test.c
```diff
@@ -53,3 +53,53 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+        if (p[i] != c) {
+            *error = 1;
+            fprintf(
+                stderr,
+                "rust[%d] = %d != %d (C): Rust \"{{ item.id }}\" -> C\n",
+                i, (int)p[i], (int)c
+            );
+        }
+        unsigned char d
+            = (unsigned char)(255) - (unsigned char)(i % 256);
```
tgross35: It should be possible to use `uint8_t` rather than `unsigned char` (above as well)

ctest-next/templates/test.c
```diff
@@ -53,3 +53,53 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+            (int)size, (int)rust_size
+        );
+        *error = 1;
+        return value;
+    }
+    int i = 0;
+    for (i = 0; i < size; ++i) {
+        if (pad[i]) { continue; }
+        unsigned char c = (unsigned char)(i % 256);
+        c = c == 0? 42 : c;
```
tgross35: Maybe rename `c` to `expected`

ctest-next/templates/test.c
```diff
@@ -53,3 +53,53 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+            stderr,
+            "size of {{ item.c_ty }} is %d in C and %d in Rust\n",
+            (int)size, (int)rust_size
+        );
+        *error = 1;
+        return value;
+    }
+    int i = 0;
+    for (i = 0; i < size; ++i) {
+        if (pad[i]) { continue; }
```
tgross35: Could you add a note about why padding is skipped here?

ctest-next/templates/test.c
```diff
@@ -53,3 +53,53 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+            fprintf(
+                stderr,
+                "rust[%d] = %d != %d (C): Rust \"{{ item.id }}\" -> C\n",
+                i, (int)p[i], (int)c
+            );
+        }
+        unsigned char d
+            = (unsigned char)(255) - (unsigned char)(i % 256);
+        d = d == 0? 42: d;
+        p[i] = d;
```
tgross35: Could you add a comment about what this is doing?

ctest-next/templates/test.c
```diff
@@ -53,3 +53,53 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+// Disable signed/unsigned conversion warnings on MSVC.
+// These trigger even if the conversion is explicit.
+#  pragma warning(disable:4365)
+#endif
+{%- for item in ctx.roundtrip_tests +%}
+
+// Tests whether the struct/union/alias `x` when passed to C and back to Rust remains unchanged.
+// It checks if the size is the same as well as if the padding bytes are all in the correct place.
+{{ item.c_ty }} ctest_roundtrip__{{ item.id }}(
+    int32_t rust_size, {{ item.c_ty }} value, int* error, unsigned char* pad
```
tgross35: Take a `size_t` for `rust_size` so no casting is needed.

Also, the size of `pad` is know ahead of time right? I think the last arg could be `uint8_t pad[sizeof({{ item.c_ty }}]`.

ctest-next/templates/test.c
```diff
@@ -53,3 +53,53 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+        if (pad[i]) { continue; }
+        unsigned char c = (unsigned char)(i % 256);
+        c = c == 0? 42 : c;
+        if (p[i] != c) {
+            *error = 1;
+            fprintf(
+                stderr,
+                "rust[%d] = %d != %d (C): Rust \"{{ item.id }}\" -> C\n",
+                i, (int)p[i], (int)c
+            );
```
tgross35: Rather than printing from C, the signature could another arg `struct roundtrip_result errors[sizeof({{ item.c_ty }}]` with

```c
struct roundtrip_result {
    bool is_error;
    uint8_t c_val;
    uint8_t rust_val;
};
```

And then if there is an error, just set the fields. Then the printing can be done on the Rust side rather than C

ctest-next/templates/test.rs
```diff
@@ -137,6 +141,95 @@ mod generated_tests {
...
+        let bar = MaybeUninit::<{{ item.id }}>::zeroed();
+        let bar = bar.as_ptr();
+        {%- for field in item.fields +%}
+        unsafe {
+            let ty_ptr = &raw const ((*bar).{{ field.ident() }});
+            let val = ty_ptr.read_unaligned();
+            let size = size_of_val(&val);
+            let off = offset_of!({{ item.id }}, {{ field.ident() }});
+            v.push((off, size));
+        }
```
tgross35: Only up to `let val` needs to be in `unsafe`, the size/offset parts are fine

ctest-next/templates/test.rs
```diff
@@ -137,6 +141,95 @@ mod generated_tests {
...
+    /// Generates a padding map for a specific type.
+    ///
+    /// Essentially, it returns a list of bytes, whose length is equal to the size of the type in
+    /// bytes. Each element corresponds to a byte and has two values. `true` if the byte is padding,
+    /// and `false` if the byte is not padding.
+    ///
+    /// For type aliases, the padding map is all zeroes.
+    fn roundtrip_padding__{{ item.id }}() -> Vec<bool> {
+        {%- if item.fields.len() == 0 +%}
+        vec![!{{ item.is_alias }}; size_of::<{{ item.id }}>()]
```
tgross35: What does this do? It looks like it constructs a vector but doesn't assign it to anything

ctest-next/templates/test.rs
```diff
@@ -137,6 +141,95 @@ mod generated_tests {
...
+            v.push((off, size));
+        }
+        {%- endfor +%}
+        // This vector contains `true` if the byte is padding
+        // and `false` if the byte is not padding.
+        // Initialize all bytes as:
+        //  - padding if we have fields, this means that only
+        //  the fields will be checked
+        //  - no-padding if we have a type alias: if this
+        //  causes problems the type alias should be skipped
```
```suggestion
        // This vector contains `true` if the byte is padding and `false` if the byte is not
        // padding. Initialize all bytes as:
        //  - padding if we have fields, this means that only the fields will be checked
        //  - no-padding if we have a type alias: if this causes problems the type alias should
        //    be skipped
```

Nit, might as well use all 100 chars

ctest-next/templates/test.rs
```diff
@@ -137,6 +141,95 @@ mod generated_tests {
...
+    /// correct place.
+    pub fn {{ item.test_name }}() {
+        type U = {{ item.id }};
+        extern "C" {
+            fn ctest_roundtrip__{{ item.id }}(
+                size: i32, x: MaybeUninit<U>, e: *mut c_int, pad: *const bool
+            ) -> U;
+        }
+        let pad = roundtrip_padding__{{ item.id }}();
+        unsafe {
```
tgross35: Can you split up this unsafe block like the others?

ctest-next/templates/test.rs
```diff
@@ -137,6 +141,95 @@ mod generated_tests {
...
+            let y_ptr = y.as_mut_ptr().cast::<u8>();
+            let sz = size_of::<U>();
+            for i in 0..sz {
+                let c: u8 = (i % 256) as u8;
+                let c = if c == 0 { 42 } else { c };
+                let d: u8 = 255_u8 - (i % 256) as u8;
+                let d = if d == 0 { 42 } else { d };
+                x_ptr.add(i).write_volatile(c);
+                y_ptr.add(i).write_volatile(d);
+            }
```
tgross35: Could you add a comment? I think it's basically filling up the bytes with a deterministic pattern to detect when those unintentionally change, one pattern will be expected on the C side and the other will be expected on the Rust side

ctest-next/templates/test.c
tgross35: Scratch that, I think this can be even easier. Just take a `uint8_t [sizeof({{ item.c_ty }}]` and copy each byte, then validate it on the Rust side.

Don't `memcpy` in order to keep `volatile`.

ctest-next/templates/test.rs
mbyx: For aliases we assume that there are no padding bytes, for structs and unions, if there are no fields, then everything is padding, if there are fields, then we have to go through each field and figure out the padding.

tgross35: It makes sense, but the vector isn't doing anything (I think it might be missing an assignment).

On that note, could you add a type alias test to `simple`?

mbyx: It is being returned conditionally if the item has no fields.

ctest-next/src/template.rs
```diff
@@ -180,6 +182,55 @@ impl TestTemplate {
...
+        }
+        for union_ in helper.ffi_items.unions() {
+            let c_ty = helper.c_type(union_)?;
+            self._add_roundtrip_test(helper, union_.ident(), &union_.fields, &c_ty, false);
+        }
+
+        Ok(())
+    }
+
+    fn _add_roundtrip_test(
```
tgross35: Remove the leading `_`, that's not a Rust convention. Any chance you have a background in Python? :)

ctest-next/templates/test.c
```diff
@@ -53,3 +53,47 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+        // Instead we just make sure the the placement of the padding bytes remains the same.
+        if (pad[i]) { continue; }
+        uint8_t expected = (uint8_t)(i % 256);
+        expected = expected == 0? 42 : expected;
+        expected_pad[i] = expected;
+        // After we check that the pattern remained unchanged from Rust to C, we invert the pattern
+        // and send it back to Rust to make sure that it remains unchanged from C to Rust.
+        uint8_t d
+            = (uint8_t)(255) - (uint8_t)(i % 256);
+        d = d == 0? 42: d;
```
tgross35: Small formatting nits

```suggestion
        expected = expected == 0 ? 42 : expected;
        expected_pad[i] = expected;
        // After we check that the pattern remained unchanged from Rust to C, we invert the pattern
        // and send it back to Rust to make sure that it remains unchanged from C to Rust.
        uint8_t d = (uint8_t)(255) - (uint8_t)(i % 256);
        d = d == 0 ? 42: d;
```

ctest-next/templates/test.rs
tgross35: Ah! I see.

I think the control flow would be a bit more clear if done at runtime rather than using a template `if`, so it includes the `return` keyword.

```rust
if {%- if item.fields.is_empty() +%} {
    // Assume no padding bytes for aliases.
    // FIXME(ctest): What if it's an alias to a struct/union?
    return vec![!{{ item.is_alias }}; size_of::<{{ item.id }}>()];
}
```

ctest-next/templates/test.rs
```diff
@@ -138,6 +141,122 @@ mod generated_tests {
...
+        {%- for field in item.fields +%}
+
+        let ty_ptr = unsafe { &raw const ((*bar).{{ field.ident() }}) };
+        let val = unsafe { ty_ptr.read_unaligned() };
+
+        let size = size_of_val(&val);
+        let off = offset_of!({{ item.id }}, {{ field.ident() }});
+        v.push((off, size));
+        {%- endfor +%}
+         // This vector contains `true` if the byte is padding and `false` if the byte is not
```
tgross35: Nit, extra space

```suggestion
        // This vector contains `true` if the byte is padding and `false` if the byte is not
```

ctest-next/templates/test.c
```diff
@@ -53,3 +53,47 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+
+// Tests whether the struct/union/alias `x` when passed by value to C and back to Rust
+// remains unchanged.
+// It checks if the size is the same as well as if the padding bytes are all in the correct place.
+{{ item.c_ty }} ctest_roundtrip__{{ item.id }}(
+    {{ item.c_ty }} value,
+    int* error,
+    uint8_t pad[sizeof({{ item.c_ty }})],
+    uint8_t expected_pad[sizeof({{ item.c_ty }})]
+) {
```
tgross35: * `error` isn't ever used so it can be removed
* Maybe `pad` -> `is_padding_byte`  to be more clear. This could also be `const uint8_t`.

ctest-next/templates/test.c
```diff
@@ -53,3 +53,47 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+    // Otherwise the Rust side would not be able to see it.
+    volatile uint8_t* p = (volatile uint8_t*)&value;
+    int i = 0;
+    for (i = 0; i < size; ++i) {
+        // We skip padding bytes in both Rust and C because writing to it is undefined.
+        // Instead we just make sure the the placement of the padding bytes remains the same.
+        if (pad[i]) { continue; }
+        uint8_t expected = (uint8_t)(i % 256);
+        expected = expected == 0? 42 : expected;
+        expected_pad[i] = expected;
```
tgross35: I'm a bit confused about what is going on here. Doesn't this just fill `expected_pad` with a predictable value?

tgross35: I think it should set `expected_pad[i]` to `p[i]`. `expected_pad` also doesn't have anything to do with padding right? Maybe something like `value_bytes`

ctest-next/templates/test.rs
```diff
@@ -138,6 +141,122 @@ mod generated_tests {
...
+        extern "C" {
+            fn ctest_size_of__{{ item.id }}() -> u64;
+            fn ctest_roundtrip__{{ item.id }}(
+                x: MaybeUninit<U>, e: *mut c_int, pad: *const bool, expected: *mut u8
+            ) -> U;
+        }
+        let pad = roundtrip_padding__{{ item.id }}();
+        let mut error: c_int = 0;
+        let mut y = MaybeUninit::<U>::zeroed();
+        let mut x = MaybeUninit::<U>::zeroed();
```
tgross35: To match the comment from the C side, `pad` could be called `is_padding_byte`.

Could `x` and `y` get descriptive names? Maybe `input` and `expected`?

ctest-next/templates/test.rs
```diff
@@ -138,6 +141,122 @@ mod generated_tests {
...
+            return;
+        }
+
+        let mut expected = vec![0; size_of::<{{ item.id }}>()];
+        let r: U = unsafe {
+            ctest_roundtrip__{{ item.id }}(x, &mut error, pad.as_ptr(), expected.as_mut_ptr())
+        };
+
+        for (i, elem) in pad.iter().enumerate().take(size) {
+            if *elem { continue; }
```
tgross35: Rename `elem` to `is_padding_byte` (same for the next loop)

ctest-next/templates/test.rs
```diff
@@ -138,6 +141,122 @@ mod generated_tests {
...
+        let c_size = unsafe { ctest_size_of__{{ item.id }}() } as usize;
+        if size != c_size {
+            FAILED.store(true, Ordering::Relaxed);
+            eprintln!(
+                "size of {{ item.c_ty }} is {c_size} in C and {size} in Rust\n",
+            );
+            return;
+        }
+
+        let mut expected = vec![0; size_of::<{{ item.id }}>()];
```
tgross35: I don't think this is an expected value, since it tells us what the bytes look like on C's side. Maybe `c_value_bytes`?

ctest-next/templates/test.rs
```diff
@@ -138,6 +141,122 @@ mod generated_tests {
...
+        let r: U = unsafe {
+            ctest_roundtrip__{{ item.id }}(x, &mut error, pad.as_ptr(), expected.as_mut_ptr())
+        };
+
+        for (i, elem) in pad.iter().enumerate().take(size) {
```
```suggestion
        // Check that the value bytes as read from C match the byte we sent from Rust.
        for (i, elem) in pad.iter().enumerate().take(size) {
```

ctest-next/templates/test.rs
tgross35: Also, I don't think `y` / `expected` needs to be a value of type `U` actually, since we are neither passing nor returning it. It can just be a `[0u8; size]`, then it can be accessed by indexing rather than as a pointer.

ctest-next/templates/test.rs
```diff
@@ -138,6 +141,122 @@ mod generated_tests {
...
+            if *elem { continue; }
+            let rust = unsafe { *x_ptr.add(i) };
+            let c = expected[i];
+            if rust != c {
+                eprintln!("rust[{}] = {} != {} (C): Rust \"{{ item.id }}\" -> C", i, rust, c);
+                FAILED.store(true, Ordering::Relaxed);
+            }
+        }
+
+        for (i, elem) in pad.iter().enumerate().take(size) {
```
```suggestion
        // Check that value returned from C contains the bytes we expect.
        for (i, elem) in pad.iter().enumerate().take(size) {
```

ctest-next/templates/test.rs
```diff
@@ -138,6 +141,122 @@ mod generated_tests {
...
+            }
+        }
+        pad
+        {%- endif +%}
+    }
+
+    /// Tests whether the type alias `x` when passed to C and back to Rust remains unchanged.
+    ///
+    /// It checks if the size is the same as well as if the padding bytes are all in the
+    /// correct place.
```
```suggestion
    /// It checks if the size is the same as well as if the padding bytes are all in the
    /// correct place. For this test to be sound, `T` must be valid for any bitpattern.
```

tgross35 (commented):
A handful more comments here, but most are small and I think this should be the last round.

ctest-next/templates/test.rs
```diff
@@ -138,6 +141,128 @@ mod generated_tests {
...
+            let rust = unsafe { *input_ptr.add(i) };
+            let c = c_value_bytes[i];
+            if rust != c {
+                eprintln!("rust[{}] = {} != {} (C): Rust \"{{ item.id }}\" -> C", i, rust, c);
+                FAILED.store(true, Ordering::Relaxed);
+            }
+        }
+
+        // Check that value returned from C contains the bytes we expect.
+        for (i, is_padding_byte) in is_padding_byte.iter().enumerate().take(SIZE) {
```
tgross35: One more nit, these don't need to `.take()` right? Since `is_padding_byte` is already `SIZE`

tgross35: If it's a sanity thing for verifying the read is inbounds, you can add an `assert!(i <= SIZE, "sanity check failed")`

ctest-next/templates/test.rs
```diff
@@ -189,6 +190,129 @@ mod generated_tests {
...
+            fn ctest_size_of__{{ item.id }}() -> u64;
+            fn ctest_roundtrip__{{ item.id }}(
+                input: MaybeUninit<U>, is_padding_byte: *const bool, expected: *mut u8
+            ) -> U;
+        }
+
+        const SIZE: usize = size_of::<U>();
+
+        let is_padding_byte = roundtrip_padding__{{ item.id }}();
+        let mut expected = [0u8; SIZE];
```
tgross35: The `expected` function signature can be updated to be called `value_bytes` like the C signature, then the `expected` array can be called `c_value_bytes`.

Also, this value and `c_value_bytes` can both be the same type - either a stack array or a vector. Allocating is probably a bit better in case of huge types.

mbyx: I don't understand what you mean by your second point, that they can be the same type. Which two?

tgross35: `expected` is a ` [0u8; SIZE]`, `c_value_bytes` is a `vec![0u8; SIZE];`. They need to store the same amount of data, so might as well make them both a `Vec`

mbyx: There's two different variables called `c_value_bytes` and its difficult to separate them so that it doesn't matter, so `expected` needs a different name.

tgross35: I'm sorry, I misread that bit. That name was fine, it didn't need to change. `expected` or e.g. `expected_return` would be fine.

tgross35: @mbyx ^ mind changing the name back? After that this is good to go

mbyx: Done!

PR #4561, opened 2025-07-23
ctest: add tests for field size, offset, and field ptr

mbyx:
# Description
Adds tests for checking if the size and offset of a field is the same in Rust and C, as well as the pointer to that field.

# Sources

# Checklist

### DIFF
--- a/ctest-next/src/template.rs
+++ b/ctest-next/src/template.rs
@@ -1,9 +1,12 @@
+use std::ops::Deref;
+
 use askama::Template;
 use quote::ToTokens;
+use syn::spanned::Spanned;
 
 use crate::ffi_items::FfiItems;
-use crate::translator::Translator;
-use crate::{BoxStr, MapInput, Result, TestGenerator, TranslationError};
+use crate::translator::{TranslationErrorKind, Translator, translate_abi, translate_expr};
+use crate::{BoxStr, Field, MapInput, Result, TestGenerator, TranslationError, VolatileItemKind};
 
 /// Represents the Rust side of the generated testing suite.
 #[derive(Template, Clone)]
@@ -46,6 +49,8 @@ impl CTestTemplate {
 /// Stores all information necessary for generation of tests for all items.
 #[derive(Clone, Debug, Default)]
 pub(crate) struct TestTemplate {
+    pub field_ptr_tests: Vec<TestFieldPtr>,
+    pub field_size_offset_tests: Vec<TestFieldSizeOffset>,
     pub signededness_tests: Vec<TestSignededness>,
     pub size_align_tests: Vec<TestSizeAlign>,
     pub const_cstr_tests: Vec<TestCStr>,
@@ -69,6 +74,8 @@ impl TestTemplate {
         template.populate_const_and_cstr_tests(&helper)?;
         template.populate_size_align_tests(&helper)?;
         template.populate_signededness_tests(&helper)?;
+        template.populate_field_size_offset_tests(&helper)?;
+        template.populate_field_ptr_tests(&helper)?;
 
         Ok(template)
     }
@@ -180,6 +187,144 @@ impl TestTemplate {
 
         Ok(())
     }
+
+    /// Populates field size and offset tests for structs/unions.
+    ///
+    /// It also keeps track of the names of each test.
+    fn populate_field_size_offset_tests(
+        &mut self,
+        helper: &TranslateHelper,
+    ) -> Result<(), TranslationError> {
+        let should_skip = |map_input| helper.generator.skips.iter().any(|f| f(&map_input));
+
+        let struct_fields = helper
+            .ffi_items
+            .structs()
+            .iter()
+            .flat_map(|struct_| struct_.fields.iter().map(move |field| (struct_, field)))
+            .filter(|(struct_, field)| {
+                !should_skip(MapInput::StructField(struct_, field)) && field.public
+            })
+            .map(|(struct_, field)| {
+                (
+                    struct_.ident(),
+                    field,
+                    helper.c_type(struct_),
+                    helper.c_ident(MapInput::StructField(struct_, field)),
+                )
+            });
+        let union_fields = helper
+            .ffi_items
+            .unions()
+            .iter()
+            .flat_map(|union_| union_.fields.iter().map(move |field| (union_, field)))
+            .filter(|(union_, field)| {
+                !should_skip(MapInput::UnionField(union_, field)) && field.public
+            })
+            .map(|(union_, field)| {
+                (
+                    union_.ident(),
+                    field,
+                    helper.c_type(union_),
+                    helper.c_ident(MapInput::UnionField(union_, field)),
+                )
+            });
+
+        for (id, field, c_ty, c_field) in struct_fields.chain(union_fields) {
+            let item = TestFieldSizeOffset {
+                test_name: field_size_offset_test_ident(id, field.ident()),
+                id: id.into(),
+                c_ty: c_ty?.into(),
+                field: field.clone(),
+                c_field: c_field.into_boxed_str(),
+            };
+            self.field_size_offset_tests.push(item.clone());
+            self.test_idents.push(item.test_name);
+        }
+
+        Ok(())
+    }
+
+    /// Populates field tests for structs/unions.
+    ///
+    /// It also keeps track of the names of each test.
+    fn populate_field_ptr_tests(
+        &mut self,
+        helper: &TranslateHelper,
+    ) -> Result<(), TranslationError> {
+        let should_skip = |map_input| helper.generator.skips.iter().any(|f| f(&map_input));
+
+        let struct_fields = helper
+            .ffi_items
+            .structs()
+            .iter()
+            .flat_map(|s| s.fields.iter().map(move |f| (s, f)))
+            .filter(|(s, f)| {
+                !should_skip(MapInput::StructField(s, f))
+                    && !should_skip(MapInput::StructFieldType(s, f))
+                    && f.public
+            })
+            .map(|(s, f)| {
+                (
+                    s.ident(),
+                    f,
+                    helper.c_type(s),
+                    helper.c_ident(MapInput::StructField(s, f)),
+                    if !helper.generator.volatile_items.is_empty()
+                        && helper
+                            .generator
+                            .volatile_items
+                            .iter()
+                            .any(|vf| vf(VolatileItemKind::StructField(s.clone(), f.clone())))
+                    {
+                        "volatile "
+                    } else {
+                        ""
+                    },
+                )
+            });
+        let union_fields = helper
+            .ffi_items
+            .unions()
+            .iter()
+            .flat_map(|u| u.fields.iter().map(move |f| (u, f)))
+            .filter(|(u, f)| {
+                !should_skip(MapInput::UnionField(u, f))
+                    && !should_skip(MapInput::UnionFieldType(u, f))
+                    && f.public
+            })
+            .map(|(u, f)| {
+                (
+                    u.ident(),
+                    f,
+                    helper.c_type(u),
+                    helper.c_ident(MapInput::UnionField(u, f)),
+                    "",
+                )
+            });
+
+        for (id, field, c_ty, c_field, volatile_keyword) in struct_fields.chain(union_fields) {
+            let field_return_type = helper
+                .make_cdecl(
+                    &format!("ctest_field_ty__{}__{}", id, field.ident()),
+                    &field.ty,
+                )?
+                .into_boxed_str();
+            let item = TestFieldPtr {
+                test_name: field_ptr_test_ident(id, field.ident()),
+                id: id.into(),
+                c_ty: c_ty?.into(),
+                field: field.clone(),
+                c_field: c_field.into_boxed_str(),
+                volatile_keyword: volatile_keyword.into(),
+                field_return_type,
+            };
+            self.field_ptr_tests.push(item.clone());
+            self.test_idents.push(item.test_name);
+        }
+
+        Ok(())
+    }
 }
 
 /* Many test structures have the following fields:
@@ -228,6 +373,26 @@ pub(crate) struct TestConst {
     pub c_ty: BoxStr,
 }
 
+#[derive(Clone, Debug)]
+pub(crate) struct TestFieldPtr {
+    pub test_name: BoxStr,
+    pub id: BoxStr,
+    pub field: Field,
+    pub c_field: BoxStr,
+    pub c_ty: BoxStr,
+    pub volatile_keyword: BoxStr,
+    pub field_return_type: BoxStr,
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct TestFieldSizeOffset {
+    pub test_name: BoxStr,
+    pub id: BoxStr,
+    pub field: Field,
+    pub c_field: BoxStr,
+    pub c_ty: BoxStr,
+}
+
 fn signededness_test_ident(ident: &str) -> BoxStr {
     format!("ctest_signededness_{ident}").into()
 }
@@ -244,17 +409,35 @@ fn const_test_ident(ident: &str) -> BoxStr {
     format!("ctest_const_{ident}").into()
 }
 
+fn field_ptr_test_ident(ident: &str, field_ident: &str) -> BoxStr {
+    format!("ctest_field_ptr_{ident}_{field_ident}").into()
+}
+
+fn field_size_offset_test_ident(ident: &str, field_ident: &str) -> BoxStr {
+    format!("ctest_field_size_offset_{ident}_{field_ident}").into()
+}
+
 /// Wrap methods that depend on both ffi items and the generator.
-struct TranslateHelper<'a> {
+pub(crate) struct TranslateHelper<'a> {
     ffi_items: &'a FfiItems,
     generator: &'a TestGenerator,
     translator: Translator,
 }
 
 impl<'a> TranslateHelper<'a> {
+    /// Create a new translation helper.
+    #[cfg_attr(not(test), expect(unused))]
+    pub(crate) fn new(ffi_items: &'a FfiItems, generator: &'a TestGenerator) -> Self {
+        Self {
+            ffi_items,
+            generator,
+            translator: Translator::new(),
+        }
+    }
+
     /// Returns the equivalent C/Cpp identifier of the Rust item.
     pub(crate) fn c_ident(&self, item: impl Into<MapInput<'a>>) -> String {
-        self.generator.map(item)
+        self.generator.rty_to_cty(item)
     }
 
     /// Returns the equivalent C/Cpp type of the Rust item.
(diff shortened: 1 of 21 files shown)


ctest-next/src/template.rs
```diff
@@ -193,6 +374,25 @@ impl TestTemplate {
...
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct TestFieldSizeOffset {
+    pub test_name: BoxStr,
+    pub id: BoxStr,
+    pub field: Field,
+    pub c_field: BoxStr,
+    pub c_ty: BoxStr,
+}
```
tgross35: Nit, I think these should go after `TestConst` since they're used later in the templates. Same for the new `field_*` functions.

ctest-next/templates/test.c
```diff
@@ -53,3 +53,28 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+{%- for item in ctx.field_ptr_tests +%}
+
+// Return a pointer to a struct/union field.
+// This field can have a normal data type, or it could be a function pointer or an array, which
+// have different syntax, so it's signature must be modified.
+// The name of the function remains `ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}`
+{{ item.volatile_keyword }}{{ item.c_signature }} {
+    return &b->{{ item.c_field }};
+}
+{%- endfor +%}
```
tgross35: Missing `\n`

ctest-next/templates/test.c
```diff
@@ -53,3 +53,28 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+
+{%- for item in ctx.field_ptr_tests +%}
+
+// Return a pointer to a struct/union field.
+// This field can have a normal data type, or it could be a function pointer or an array, which
+// have different syntax, so it's signature must be modified.
+// The name of the function remains `ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}`
+{{ item.volatile_keyword }}{{ item.c_signature }} {
```
tgross35: Can the args and return type be split to separate fields so this is directly visible rather than a comment? So

```c
{{ item.volatile_keyword }} {{ item.c_ret_ty }} ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}({{ item.c_args }})
```

ctest-next/templates/test.c
```diff
@@ -53,3 +53,28 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+
+// Return the offset of a struct/union field.
+uint64_t ctest_offset_of__{{ item.id }}__{{ item.field.ident() }}(void) {
+    return offsetof({{ item.c_ty }}, {{ item.c_field }});
+}
+
+// Return the size of a struct/union field.
+uint64_t ctest_size_of__{{ item.id }}__{{ item.field.ident() }}(void) {
+    {{ item.c_ty }}* foo = NULL;
+    return sizeof(foo->{{ item.c_field }});
```
tgross35: This is technically okay because `sizeof` is evaluated at compile time, but it still looks like a NULL deref. Does this work?

```c
sizeof((({{ item.c_ty }}){}).{{ item.c_field }})
```

ctest-next/templates/test.rs
```diff
@@ -137,6 +139,57 @@ mod generated_tests {
...
+                    ctest_offset_of__{{ item.id }}__{{ item.field.ident() }}() as usize,
+                    "field offset {{ item.field.ident() }} of {{ item.id }}");
+            check_same(size_of_val(&val) as u64,
+                    ctest_size_of__{{ item.id }}__{{ item.field.ident() }}(),
+                    "field size {{ item.field.ident() }} of {{ item.id }}");
+
+            let uninit_ty = MaybeUninit::<{{ item.id }}>::zeroed();
+            let ty_ptr = uninit_ty.as_ptr();
+            let field_ptr = &raw const ((*ty_ptr).{{ item.field.ident() }});
+            check_same(field_ptr as *mut _,
```
tgross35: `.cast::<ty>()` rather than `as *`

ctest-next/templates/test.rs
```diff
@@ -137,6 +139,57 @@ mod generated_tests {
...
+    pub fn {{ item.test_name }}() {
+        extern "C" {
+            fn ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}(a: *const {{ item.id }}) -> *mut u8;
+        }
+
+        unsafe {
+            let uninit_ty = MaybeUninit::<{{ item.id }}>::zeroed();
+            let ty_ptr = uninit_ty.as_ptr();
+            let field_ptr = &raw const ((*ty_ptr).{{ item.field.ident() }});
+            check_same(field_ptr as *mut _,
```
tgross35: `.cast`

ctest-next/templates/test.rs
```diff
@@ -137,6 +139,57 @@ mod generated_tests {
...
+        }
+
+        unsafe {
+            let uninit_ty = MaybeUninit::<{{ item.id }}>::zeroed();
+            let uninit_ty = uninit_ty.as_ptr();
+            let ty_ptr = &raw const (*uninit_ty).{{ item.field.ident() }};
+            let val = ty_ptr.read_unaligned();
+            check_same(offset_of!({{ item.id }}, {{ item.field.ident() }}),
+                    ctest_offset_of__{{ item.id }}__{{ item.field.ident() }}() as usize,
+                    "field offset {{ item.field.ident() }} of {{ item.id }}");
```
tgross35: Cast the `offset_of!` result to `u64` rather than the other one to `usize` (always fitting into the larger value), same as the `size_of_val` check

ctest-next/templates/test.rs
```diff
@@ -137,6 +139,57 @@ mod generated_tests {
...
+                    ctest_size_of__{{ item.id }}__{{ item.field.ident() }}(),
+                    "field size {{ item.field.ident() }} of {{ item.id }}");
+
+            let uninit_ty = MaybeUninit::<{{ item.id }}>::zeroed();
+            let ty_ptr = uninit_ty.as_ptr();
+            let field_ptr = &raw const ((*ty_ptr).{{ item.field.ident() }});
+            check_same(field_ptr as *mut _,
+                    ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}(ty_ptr),
+                    "field type {{ item.field.ident() }} of {{ item.id }}");
+        }
```
tgross35: This block can be split a bit so we can see what `unsafe` is handling. Something like:

1. `let field_ptr = &raw const ((*ty_ptr).{{ item.field.ident() }})` should get its own `unsafe` block that says it doesn't actually read `ty_ptr` (AIUI requiring `unsafe` for this op is a limitation of the `unsafe` check, can't cause UB)
2. Getting the size and offset from C should each get their own `unsafe` block assigned to a variable
3. The `read_unaligned` function should get its own block
4. The rest should be able to exist outside of `unsafe`

ctest-next/templates/test.rs
```diff
@@ -137,6 +139,57 @@ mod generated_tests {
...
+
+{%- for item in ctx.field_ptr_tests +%}
+
+    /// Tests if the pointer to the field is the same in Rust and C.
+    pub fn {{ item.test_name }}() {
+        extern "C" {
+            fn ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}(a: *const {{ item.id }}) -> *mut u8;
+        }
+
+        unsafe {
```
tgross35: Similar to the above, everything except `let field_ptr = ...` and `let c_field_ptr = ctest_field_ptr__...(...)` doesn't need to be in `unsafe`

ctest-next/templates/test.c
tgross35: Oh, I see the problem, function pointer syntax is horrid.

I'd suggest eliminating this with a typedef when you're dealing with function pointers:

```c
typedef ret_ty (*some_name)(argty1, argty2);
some_name ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}(...) { /* ... */ }
```

`some_name` can be something like `ctest_field_ty__{{ item.id }}__{{ item.field.ident() }}`

ctest-next/templates/test.rs
```diff
@@ -13,6 +13,8 @@ mod generated_tests {
     use std::fmt::{Debug, LowerHex};
     use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
     use std::{mem, ptr, slice};
+    #[expect(unused_imports)]
+    use std::mem::{MaybeUninit, offset_of};
```
tgross35: Shouldn't this be `allow` rather than `expect`? Because when there are field tests, it's not actually unused.

ctest-next/src/template.rs
```diff
@@ -180,6 +187,172 @@ impl TestTemplate {
...
+                    c_ty: helper.c_type(union_)?.into(),
+                    field: field.clone(),
+                    c_field: helper
+                        .c_ident(MapInput::UnionField(union_, field))
+                        .into_boxed_str(),
+                };
+                self.field_size_offset_tests.push(item.clone());
+                self.test_idents.push(item.test_name);
+            }
+        }
```
tgross35: Since structs and unions use the same field types, I think this could be collapsed to a single iterator. Something like:

```rust
let should_skip = |map_input| helper
    .generator
    .skips
    .iter()
    .any(|f| f(map_input));

let struct_fields = helper.ffi_items.structs().iter()
    .flat_map(|st| st.fields().map(|field| (st, field))
    .filter(|(st, field)| should_skip(&MapInput::StructField(st, field))
    .map(|(st, field)| field);
let union_fields = helper.ffi_items.unions().iter()
    .flat_map(|un| un.fields().map(|field| (un, field))
    .filter(|(un, field)| should_skip(&MapInput::UnionField(un, field))
    .map(|(un, field)| field);

for field in struct_fields.chain(union_fields) {
    // ...
}
```

Something similar for `populate_field_ptr_tests`

ctest-next/templates/test.rs
```diff
@@ -137,6 +139,51 @@ mod generated_tests {
...
+        let uninit_ty = uninit_ty.as_ptr();
+
+        let ty_ptr = unsafe { &raw const (*uninit_ty).{{ item.field.ident() }}   };
+        let val = unsafe { ty_ptr.read_unaligned() };
+
+        let ctest_field_offset = unsafe { ctest_offset_of__{{ item.id }}__{{ item.field.ident() }}() };
+        check_same(offset_of!({{ item.id }}, {{ item.field.ident() }}) as u64, ctest_field_offset,
+            "field offset {{ item.field.ident() }} of {{ item.id }}");
+
+        let ctest_field_size = unsafe { ctest_size_of__{{ item.id }}__{{ item.field.ident() }}() };
```
tgross35: Safety comments

```suggestion
        // SAFETY: we assume the field access doesn't wrap
        let ty_ptr = unsafe { &raw const (*uninit_ty).{{ item.field.ident() }}   };
        
        // SAFETY: we assume that all zeros is a valid bitpattern for `ty_ptr`, otherwise the
        // test should be skipped.
        let val = unsafe { ty_ptr.read_unaligned() };

        // SAFETY: FFI call with no preconditions
        let ctest_field_offset = unsafe { ctest_offset_of__{{ item.id }}__{{ item.field.ident() }}() };
        check_same(offset_of!({{ item.id }}, {{ item.field.ident() }}) as u64, ctest_field_offset,
            "field offset {{ item.field.ident() }} of {{ item.id }}");

        // SAFETY: FFI call with no preconditions
        let ctest_field_size = unsafe { ctest_size_of__{{ item.id }}__{{ item.field.ident() }}() };
```

mbyx:
It's fine to review the field ptr checks now, splitting it would be harder.

ctest-next/src/template.rs
```diff
@@ -180,6 +187,138 @@ impl TestTemplate {
...
+                    helper.c_type(u),
+                    helper.c_ident(MapInput::UnionField(u, f)),
+                    "",
+                )
+            });
+
+        for (id, field, c_ty, c_field, volatile_keyword) in struct_fields.chain(union_fields) {
+            let c_signature = helper
+                .translate_signature_returning_ptr(
+                    &format!("ctest_field_ty__{}__{}", id, field.ident(),),
```
tgross35:
```suggestion
                    &format!("ctest_field_ty__{}__{}", id, field.ident()),
```

ctest-next/src/template.rs
```diff
@@ -291,4 +458,139 @@ impl<'a> TranslateHelper<'a> {
...
+
+    /// Modify function signatures to properly return pointer types in C.
+    ///
+    /// In C, function pointers and arrays have a different syntax to return them,
+    /// and this translation is done by this method.
+    pub(crate) fn translate_signature_returning_ptr(
+        &self,
+        signature: &str,
+        return_type: &syn::Type,
+    ) -> Result<String, TranslationError> {
```
tgross35: This can be a general cdecl generator right, not specific to signatures anymore right? Maybe `translate_signature_returning_ptr` -> `make_cdecl`, `signature` -> `name`, and `return_type` -> `ty`.

ctest-next/src/template.rs
```diff
@@ -291,4 +458,139 @@ impl<'a> TranslateHelper<'a> {
...
+                            _ => {
+                                return Err(TranslationError::new(
+                                    TranslationErrorKind::NotFfiCompatible,
+                                    &p.to_token_stream().to_string(),
+                                    inner_ty.span(),
+                                ));
+                            }
+                        }
+                    }
+                }
```
tgross35: There is some similar logic for handling `Option` elsewhere isn't there? Could this reuse the valid type checks?

ctest-next/src/template.rs
```diff
@@ -291,4 +458,139 @@ impl<'a> TranslateHelper<'a> {
...
+                    let len_inner = translate_expr(&inner.len);
+                    return Ok(format!(
+                        "{inner_type} (*{signature})[{len_outer}][{len_inner}]",
+                    ));
+                } else {
+                    let elem_type = self.c_type_raw(elem)?;
+                    return Ok(format!("{elem_type} (*{signature})[{len_outer}]"));
+                }
+            }
+            _ => (),
```
tgross35: Should unknown types return an error?

ctest-next/src/template.rs
```diff
@@ -291,4 +458,139 @@ impl<'a> TranslateHelper<'a> {
...
+        };
+
+        Ok(self.generator.map(item))
+    }
+
+    /// Partially translate a Rust bare function type into its equivalent C type.
+    ///
+    /// It returns the translated return type, translated argument types, and whether
+    /// it is variadic as a tuple.
+    pub(crate) fn translate_signature_partial(
```
tgross35: No need to be public right?

ctest-next/src/template.rs
```diff
@@ -291,4 +458,139 @@ impl<'a> TranslateHelper<'a> {
...
+            },
+        };
+        Ok((return_type, args, signature.variadic.is_some()))
+    }
+
+    /// Modify function signatures to properly return pointer types in C.
+    ///
+    /// In C, function pointers and arrays have a different syntax to return them,
+    /// and this translation is done by this method.
+    pub(crate) fn translate_signature_returning_ptr(
```
tgross35: What this is doing is pretty complex, could you add some unit tests demonstrating its behavior?

ctest-next/src/template.rs
```diff
@@ -291,4 +458,139 @@ impl<'a> TranslateHelper<'a> {
 
         Ok(self.generator.map(item))
     }
+
+    /// Recursively get the properly mapped type for some syn::Type.
```
tgross35: Slight disambiguation

```suggestion
    /// Get the properly mapped type for some `syn::Type`, recursing for pointer types as needed.
```

ctest-next/src/template.rs
```diff
@@ -291,4 +458,139 @@ impl<'a> TranslateHelper<'a> {
...
+    ///
+    /// This method is meant to only be used to translate simpler types, such as primitives or
+    /// pointers/references to primitives.
```
tgross35: Could you mention that it adds the `struct` / `union` keyword as needed?

ctest-next/src/template.rs
```diff
@@ -291,4 +458,139 @@ impl<'a> TranslateHelper<'a> {
...
+        let unmapped_c_type = self.translator.translate_type(ty)?;
+        let item = if self.ffi_items.contains_struct(&type_name) {
+            MapInput::StructType(&unmapped_c_type)
+        } else if self.ffi_items.contains_union(&type_name) {
+            MapInput::UnionType(&unmapped_c_type)
+        } else {
+            MapInput::Type(&unmapped_c_type)
+        };
+
+        Ok(self.generator.map(item))
```
tgross35: If you get the chance, `map` could use a more descriptive name. Maybe `rty_to_cty`?

ctest-next/src/template.rs
```diff
@@ -291,4 +458,139 @@ impl<'a> TranslateHelper<'a> {
 
         Ok(self.generator.map(item))
     }
+
+    /// Recursively get the properly mapped type for some syn::Type.
+    ///
+    /// This method is meant to only be used to translate simpler types, such as primitives or
+    /// pointers/references to primitives.
+    fn c_type_raw(&self, ty: &syn::Type) -> Result<String, TranslationError> {
```
tgross35: Maybe `basic_c_type`? I don't think `raw` quite fits here

ctest-next/src/template.rs
```diff
@@ -291,4 +458,139 @@ impl<'a> TranslateHelper<'a> {
 
         Ok(self.generator.map(item))
     }
+
```
tgross35, line 477: All of the below functions are related to translating more complex Rust types to C types, not directly related to filling out the template. Could they be in the `translator` module, possibly as methods on `Translator`?

ctest-next/templates/test.c
```diff
@@ -53,3 +53,27 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+    return sizeof((({{ item.c_ty }}){}).{{ item.c_field }});
+}
+{%- endfor +%}
+
+{%- for item in ctx.field_ptr_tests +%}
+
+// Return a pointer to a struct/union field.
+// This field can have a normal data type, or it could be a function pointer or an array, which
+// have different syntax, so it's signature must be modified.
+typedef {{ item.volatile_keyword }}{{ item.c_signature }};
```
tgross35: Some update to the wording and `c_signature` is needed since this isn't a signature anymore

ctest-next/templates/test.c
```diff
@@ -53,3 +53,27 @@ uint32_t ctest_signededness_of__{{ alias.id }}(void) {
...
+ctest_field_ty__{{ item.id }}__{{ item.field.ident() }} ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}({{ item.c_ty }} *b) {
```
```suggestion
ctest_field_ty__{{ item.id }}__{{ item.field.ident() }}
ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}({{ item.c_ty }} *b) {
```

Long line

ctest-next/templates/test.rs
```diff
@@ -137,6 +139,56 @@ mod generated_tests {
...
+
+    /// Tests if the pointer to the field is the same in Rust and C.
+    pub fn {{ item.test_name }}() {
+        extern "C" {
+            fn ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}(a: *const {{ item.id }}) -> *mut u8;
+        }
+
+        let uninit_ty = MaybeUninit::<{{ item.id }}>::zeroed();
+        let ty_ptr = uninit_ty.as_ptr();
+        // SAFETY: We don't read `field_ptr`, only compare the pointer itself.
```
tgross35, line 183:
```suggestion
        // SAFETY: We don't read `field_ptr`, only compare the pointer itself. The assumption is made that
        // this does not wrap the address space.
```

Learned there is another precondition here

ctest-next/tests/input/simple.h
tgross35, line 1: It looks like the template tests don't show these tests at all. Could you some test cases so we can see the result?

tgross35 (changes requested):

ctest-next/src/template.rs
mbyx: It was duplicated in the previous PR, but not anymore.

ctest-next/src/template.rs
mbyx: `basic_c_type` and `make_cdecl` require `ffi_items` and `generator`, so I think it would be better to keep them here. `translate_signature_partial` depends on `basic_c_type` as well.

tgross35 (approved):

PR #4563, opened 2025-07-23
Add 'const' to some signatures for AIX

xingxue-ibm:
# Description

The AIX signatures for some non-POSIX functions differ from those on platforms like Linux: some arguments are not marked with the `const` qualifier, even though they are not modified. To be consistent with other platforms, `const` is added to the Rust declarations.

# Sources

# Checklist


tgross35 (approved):
Thank you!

entd:
Thanks for this work guys.

Issue #4419, opened 2025-04-15
Remove uses of `enum`


tgross35:
Rust enums become UB if they hold an unexpected value, while C enums are effectively a grouped list of named constants. This causes easy UB if e.g. C adds a value or returns a value not represented by the enum (as is allowed in C), and this value gets passed to Rust.

The easiest thing to do is map enums to constants, which we already do in most cases. We could also consider an `enum!` macro that figures out the values.

Some more context is at https://mdaverde.com/posts/rust-bindgen-enum/ and https://github.com/rust-lang/rust-bindgen/issues/2646.

madsmtm:
See also https://github.com/rust-lang/rfcs/pull/3803.

PR #4575, opened 2025-07-24
libc: remove uses of enum as per #4419

mbyx:
# Description
Removes all uses of `enum`s, replacing them with `c_enums`. Since c_enums creates numeric types and constants, `cfg_attr`s were also removed for deriving `Copy` and `Clone`.

# Sources

# Checklist


src/macros.rs
```diff
@@ -223,10 +223,24 @@ macro_rules! e {
 ///
 /// See <https://github.com/rust-lang/libc/issues/4419> for more.
 macro_rules! c_enum {
-    (
+    ($(
+        $(#[repr($repr:ty)])?
+        enum $ty_name:ident {
```
tgross35: Could this require the `pub` keyword since we make the types `pub`? Similar to the other macros

src/unix/linux_like/linux/mod.rs
```diff
@@ -798,7 +798,6 @@ s! {
     }
 
     // linux/openat2.h
-    #[non_exhaustive]
```
tgross35: Why are these changes here?

PR #4577, opened 2025-07-24
Restore `struct winsize` for AIX

xingxue-ibm:
# Description
This fixes the problem that `struct winsize` was guarded out for AIX incorrectly.

# Sources

# Checklist


collinfunk:
`s/Retore/Restore`

On AIX and glibc `struct winsize` are defined in `sys/ioctl.h`. Most other platforms define it in `termios.h`. Maybe that is why missed. This change looks good to me though.

PR #4578, opened 2025-07-24
Restore 'struct ld_info' and related types for AIX

xingxue-ibm:
# Description
`struct ld_info` and related types are used by the AIX system call `loadquery()`. This patch reintroduces them with adjustments to pass CI.

# Sources

# Checklist

### DIFF
--- a/src/unix/aix/powerpc64.rs
+++ b/src/unix/aix/powerpc64.rs
@@ -1,6 +1,12 @@
 use crate::off_t;
 use crate::prelude::*;
 
+// Define lock_data_instrumented as an empty enum
+missing! {
+    #[cfg_attr(feature = "extra_traits", derive(Debug))]
+    pub enum lock_data_instrumented {}
+}
+
 s! {
     pub struct sigset_t {
         pub ss_set: [c_ulong; 4],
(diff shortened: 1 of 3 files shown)


src/unix/aix/powerpc64.rs
```diff
@@ -327,6 +394,127 @@ cfg_if! {
                 self.__si_flags.hash(state);
             }
         }
+
+        impl PartialEq for _kernel_simple_lock {
```
tgross35: We aren't adding `PartialEq` traits anymore, since they are planned to be remove in 1.0 (trivial unsoundness). These can just be dropped.

Some more info at https://github.com/rust-lang/libc/issues/3880

src/unix/aix/powerpc64.rs
```diff
@@ -256,6 +259,70 @@ s_no_extra_traits! {
         pub __pad: [c_int; 3],
     }
 
+    pub union _kernel_simple_lock {
+        pub _slock: c_long,
+        // Should be pointer to 'lock_data_instrumented'
+        pub _slockp: *mut c_void,
```
tgross35: Is `lock_data_instrumented` an opaque struct? If so, you can define it as an empty enum, e.g. https://github.com/rust-lang/libc/blob/5c2db83f665a7fdfeb4535cfc163f2bc3da61eb5/src/unix/linux_like/mod.rs#L11-L14

src/unix/aix/powerpc64.rs
```diff
@@ -256,6 +259,70 @@ s_no_extra_traits! {
...
+            a: c_long,
+            b: crate::caddr_t,
+            c: c_long,
+            d: c_long,
+        ) -> c_int,
+        pub fo_select:
+            extern "C" fn(file: *mut file, a: c_int, b: *mut c_ushort, c: extern "C" fn()) -> c_int,
+        pub fo_close: extern "C" fn(file: *mut file) -> c_int,
+        pub fo_fstat: extern "C" fn(file: *mut file, sstat: *mut crate::stat) -> c_int,
+    }
```
tgross35: Are the function pointers here ever null? They may need to be `Option<fn(...)>`

Also, are the headers/docs for any of these structs public?

tgross35 (changes requested):

src/unix/aix/powerpc64.rs
xingxue-ibm: I couldn't find the `sys/file.h` header file online. It’s only mentioned in the AIX 7.3 documentation under t [List of Major Control Block Header Files](https://www.ibm.com/docs/en/aix/7.3.0?topic=files-list-major-control-block-header). The header is installed by default on AIX systems.

It seems that `struct fileops` defines the file operations for an open file or pipe. I'm not completely sure whether its function pointers can ever be NULL, but I suppose using `Option<fn(...)>` in Rust is a safe choice. Thanks for the guidance!

`struct fileops` and `struct file` are guarded under macro `_KERNEL`.
```
typedef struct fileops
{
        int     (*fo_rw)(struct file *, enum uio_rw, struct uio *,
                         ext_t, file_secattr_t *);
        int     (*fo_ioctl)(struct file *, long, caddr_t, ext_t, long);
        int     (*fo_select)(struct file *, int, ushort, ushort *, void (*)());
        int     (*fo_close)(struct file *);
        int     (*fo_fstat)(struct file *, struct stat *);
} fileops_t;
...
/* One file structure is allocated for each open/creat/pipe call. Main use is
 * to hold the read/write pointer associated with each open file.
 *
 * Portions of this structure are overlaid when placed on the freelist.
 * Binary compatibility concerns require that fields *not* be rearranged for
 * sanity.
 */
__align(256) struct file {
        long    f_flag;         /* see fcntl.h for definitions               */
        int     f_count;        /* reference count                           */
        short   f_options;      /* file flags not passed through vnode layer */
        short   f_type;         /* descriptor type                           */

        /* The following fields are overloaded when the file struct is freed,
         * except for the locks which must remain static. See struct freefile.
         */

        struct vnode *f_data;           /* pointer to vnode struct           */
#define f_vnode       f_data

        offset_t        f_offset;       /* read/write character pointer      */
        off_t           f_dir_off;      /* BSD style dir offsets             */
        struct ucred    *f_cred;        /* process cred at open              */

        Simple_lock     f_lock;         /* file structure fields lock        */
        Simple_lock     f_offset_lock;  /* file structure offset field lock  */

        caddr_t         f_vinfo;        /* any info vfs needs                */
        fileops_t       *f_ops;         /* operations vector                 */

        caddr_t         f_parentp;      /* the file's parent vnode or fid    */
        caddr_t         f_fnamep;       /* fname pointer, may be embedded    */
        char            f_fdata[F_FDATASZ];/* embedded file data, pad to 256 */
};
```
There is a [description of `struct ld_info`](https://www.ibm.com/docs/en/aix/7.3.0?topic=files-ldrh-file).

src/unix/aix/powerpc64.rs
xingxue-ibm: Done, thank you!

src/unix/aix/powerpc64.rs
xingxue-ibm: Removed them, thanks!

Are these `PartialEq`s needed when cherry-picking this patch to the stable branch?

tgross35: Oh sorry, I missed part of my sentence; we aren't doing `PartialEq` or the other traits (`Eq`, `Hash`) so they can all be removed. (Does this actually build in its current state with `Eq` but not `PartialEq`?)

> Are these `PartialEq`s needed when cherry-picking this patch to the stable branch?

They aren't, basically everything new should follow the 1.0 rules so this is good.

xingxue-ibm: No worries, and thanks for the clarification! I’ve removed `Eq` and `Hash` as well. The code builds with `Eq` and `Hash` without `PartialEq`, but having `Debug` fails the build.

PR #4579, opened 2025-07-25
libc: make structs non_exhaustive by default.

mbyx:
# Description
By default all structs that use the `s` and `s_no_extra_traits` macros are now `non_exhaustive`. Structs can opt out of this if their first attribute is `#[@not_non_exhaustive]`.

The `@` was necessary so that Rust could differentiate between a normal attribute (which cannot have an `@`) and this. Another possible syntax that we could use instead is `#[@not(non_exhaustive)]`.

# Sources

# Checklist

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -101,21 +101,25 @@ macro_rules! prelude {
 /// Implement `Clone` and `Copy` for a struct, as well as `Debug`, `Eq`, `Hash`, and
 /// `PartialEq` if the `extra_traits` feature is enabled.
 ///
+/// By default, it will mark a struct as `non_exhaustive`. To opt out, use the attribute
+/// `#[@not_non_exhaustive]` as the first attribute of the struct.
+///
 /// Use [`s_no_extra_traits`] for structs where the `extra_traits` feature does not
 /// make sense, and for unions.
 macro_rules! s {
     ($(
+        $(#[@$not_non_exhaustive:ident])*
         $(#[$attr:meta])*
         pub $t:ident $i:ident { $($field:tt)* }
     )*) => ($(
-        s!(it: $(#[$attr])* pub $t $i { $($field)* });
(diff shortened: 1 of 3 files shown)


src/macros.rs
```diff
@@ -93,20 +93,38 @@ macro_rules! prelude {
...
 /// `PartialEq` if the `extra_traits` feature is enabled.
 ///
+/// By default, it will mark a struct as `non_exhaustive`. To opt out, use the attribute
+/// `#[@not_non_exhaustive]` as the first attribute of the struct.
+///
 /// Use [`s_no_extra_traits`] for structs where the `extra_traits` feature does not
 /// make sense, and for unions.
 macro_rules! s {
     ($(
+        $(#[@$not_non_exhaustive:ident])*
```
tgross35, line 111: This should just be able to match a literal, right?

```suggestion
        $(#[not_non_exhaustive])?
```

mbyx: Nope, it fails with `"attempted to repeat an expression containing no syntax variables"`

mbyx:
The errors are strange, is it because `ctest` does not support some syntax? It seems to fail for `c_enum` as well but that hasn't been touched.

tgross35:
> The errors are strange, is it because `ctest` does not support some syntax? It seems to fail for `c_enum` as well but that hasn't been touched.

Yeah, that's the kind of random problem we run into with the current ctest. Completely unrelated changes make random failures show up.

The rewrite will be quite nice :)

mbyx:
For platforms where ctest has been upgraded, like apple, problems similar to #4651 appear. Also the current implementation of the macro doesn't seem to work very well either.

Link to the issue in Rust: https://github.com/rust-lang/rust/issues/132699

tgross35:
Unfortunately anything along these lines is blocked by https://github.com/rust-lang/rust/issues/132699

@rustbot blocked

PR #4581, opened 2025-07-25
linux/musl: Make time_t and suseconds_t 64-bit on all musl platforms

kraj:
time_t and suseconds_t are defined as 64-bit quanity on musl systems regardless of 32bit or 64bit CPU word length

grep "_Int64" in musl tree shows this

arch/aarch64/bits/alltypes.h.in:#define _Int64 long arch/arm/bits/alltypes.h.in:#define _Int64 long long arch/i386/bits/alltypes.h.in:#define _Int64 long long arch/loongarch64/bits/alltypes.h.in:#define _Int64 long arch/m68k/bits/alltypes.h.in:#define _Int64 long long arch/microblaze/bits/alltypes.h.in:#define _Int64 long long arch/mips/bits/alltypes.h.in:#define _Int64 long long arch/mips64/bits/alltypes.h.in:#define _Int64 long arch/mipsn32/bits/alltypes.h.in:#define _Int64 long long arch/or1k/bits/alltypes.h.in:#define _Int64 long long arch/powerpc/bits/alltypes.h.in:#define _Int64 long long arch/powerpc64/bits/alltypes.h.in:#define _Int64 long arch/riscv32/bits/alltypes.h.in:#define _Int64 long long arch/riscv64/bits/alltypes.h.in:#define _Int64 long arch/s390x/bits/alltypes.h.in:#define _Int64 long
arch/sh/bits/alltypes.h.in:#define _Int64 long long arch/x32/bits/alltypes.h.in:#define _Int64 long long arch/x86_64/bits/alltypes.h.in:#define _Int64 long include/alltypes.h.in:TYPEDEF _Int64 time_t;
include/alltypes.h.in:TYPEDEF _Int64 suseconds_t;

# Description

# Sources

# Checklist

### DIFF
--- a/src/unix/linux_like/linux/musl/mod.rs
+++ b/src/unix/linux_like/linux/musl/mod.rs
@@ -12,8 +12,8 @@ pub type clock_t = c_long;
                 See #1848 for more info."
     )
 )]
-pub type time_t = c_long;
-pub type suseconds_t = c_long;
+pub type time_t = i64;
+pub type suseconds_t = i64;
 pub type ino_t = u64;
 pub type off_t = i64;
 pub type blkcnt_t = i64;


tgross35:
Unfortunately this isn't just a trivial incorrect type; we also need to use the new link names when the 1.2 definitions are used. There is a work in progress at https://github.com/rust-lang/libc/pull/4463.

kraj:
> Unfortunately this isn't just a trivial incorrect type; we also need to use the new link names when the 1.2 definitions are used. There is a work in progress at #4463.

yeah I did realize it after some testing. We need to keep the underlying types matches with musl, so perhaps `c_long` and `c_longlong` based upon architecture would be right thing to do.

PR #4582, opened 2025-07-25
Fix the types of constants used for the 'request' argument of ioctl() on AIX

xingxue-ibm:
# Description
The `request` argument of the `ioctl()` function has type `c_int`, but the associated constants were mistakenly defined as `c_long` due to an oversight on my part.

# Sources

# Checklist


libc-test/build.rs
```diff
@@ -5593,6 +5593,21 @@ fn test_aix(target: &str) {
         // values because non-unique values are being used which will
         // fail the test when _ALL_SOURCE is defined.
         "EWOULDBLOCK" | "ENOTEMPTY" => true,
+
+        // These constants are intended for use as the 'int request' argument to 'ioctl()'.
```
tgross35: Could you make this comment `FIXME(ctest)`? Because if it's an issue that only shows up during tests but not in C, presumably we could do some better validation on our end.

xingxue-ibm: > Could you make this comment `FIXME(ctest)`? Because if it's an issue that only shows up during tests but not in C, presumably we could do some better validation on our end.

Good point! Thanks!

collinfunk:
I assume this is because BSD and Linux use `long` for the `request` argument `ioctl`, despite POSIX requiring `int`. AIX follows POSIX and uses `int` so this is correct.

tgross35 (approved):

PR #4589, opened 2025-07-28
Map madvise() to posix_madvise() for AIX

xingxue-ibm:
# Description
The AIX signature of `madvise()` differs from the POSIX specification, which expects `void *` as the type of the `addr` argument, whereas AIX uses `caddr_t` (i.e., `char *`). This patch updates the `madvise()` signature and maps it to `posix_madvise()` to be consistent with other platforms.

# Sources

# Checklist


xingxue-ibm:
> We shouldn't do this kind of remapping here, higher-level crates can smooth this out instead. In general, `libc` should match the platform bindings as closely as possible.

I see, that makes sense. Thanks for the guidance as always!

PR #4591, opened 2025-07-29
ci: Switch s390x to a native job

tgross35:


tgross35:
I'm -1 on these kind of changes nowadays. It's hard enough to manage kernel and libc versions when using docker on the usual GHA runners, having separate runners doesn't get us enough to justify the added version differences.

PR #4594, opened 2025-07-29
ctest: add foreign fn tests

mbyx:
# Description
Adds support for testing foreign functions. A lot of hard to track bugs had to be fixed to do this:
- Translator was changed to use exactly the same parsing format as the original ctest. It was easier to do everything this way, but I believe it does cause it to become more specific in the types it works with.
- Previously things like `*const T1Bar` was being translated as `struct *const T1Bar`, that has been fixed, although in a hacky way.

# Sources

# Checklist


ctest-next/src/ast/function.rs
```diff
@@ -6,14 +6,13 @@ use crate::{Abi, BoxStr, Parameter};
...
     pub(crate) public: bool,
-    #[expect(unused)]
     pub(crate) abi: Abi,
     pub(crate) ident: BoxStr,
     pub(crate) link_name: Option<BoxStr>,
-    #[expect(unused)]
     pub(crate) parameters: Vec<Parameter>,
     #[expect(unused)]
     pub(crate) return_type: Option<syn::Type>,
+    pub(crate) bare_fn_signature: BoxStr,
```
tgross35: Could this just store a `syn::Signature`? Then add a `-> String` method that can be called as needed

ctest-next/src/generator.rs
```diff
@@ -855,6 +856,21 @@ impl TestGenerator {
...
+    ///
+    /// The closure is given a Rust FFI function and returns whether
+    /// the test will be generated.
+    ///
+    /// By default generated tests will ensure that the function pointer in C
+    /// corresponds to the same function pointer in Rust. This can often
+    /// uncover subtle symbol naming issues where a header file is referenced
+    /// through the C identifier `foo` but the underlying symbol is mapped to
+    /// something like `__foo_compat`.
+    pub fn skip_fn_ptrcheck(&mut self, f: impl Fn(&str) -> bool + 'static) -> &mut Self {
```
tgross35, line 879: Could you add an example?

ctest-next/src/template.rs
```diff
@@ -482,36 +583,97 @@ impl<'a> TranslateHelper<'a> {
     fn basic_c_type(&self, ty: &syn::Type) -> Result<String, TranslationError> {
         let type_name = match ty {
             syn::Type::Path(p) => p.path.segments.last().unwrap().ident.to_string(),
-            syn::Type::Ptr(p) => self.basic_c_type(&p.elem)?,
-            syn::Type::Reference(r) => self.basic_c_type(&r.elem)?,
+            // Using recursion here causes breakage.
```
tgross35: What breaks here?

mbyx: Recursing caused `*const T1Bar` to be translated as `struct const struct T1Bar *`, #4617 removes `basic_c_type` so it's not a problem.

ctest-next/src/template.rs
```diff
@@ -549,4 +613,29 @@ impl<'a> TranslateHelper<'a> {
...
+                TranslationErrorKind::UnsupportedType,
+                signature,
+                Span::call_site(),
+            )
+        })?;
+        if let syn::Type::BareFn(f) = item.ty.deref() {
+            return Ok(f.clone());
+        }
+        panic!("`parse_signature_to_bare_fn` returned a non function type.")
+    }
```
tgross35: What are you trying to guard against here with the panic?

String manipulation is never very robust, and you have `syn` types for everything already. Why not construct a `TypeBareFn` from `Signature`? This also allows you to reject specific syntax that you don't want to support.

ctest-next/src/translator.rs
```diff
@@ -197,17 +198,79 @@ impl<'a> Translator<'a> {
...
+                });
+
+                // The volatile keyword is only applied to the inner most named type.
+                let mut current = &mut cty;
+                while let cdecl::CTy::Ptr { ty, .. } = current {
+                    current = ty;
+                }
+                if let cdecl::CTy::Named { qual, .. } = current {
+                    qual.volatile = is_volatile;
+                }
```
tgross35: Since this is used a couple times, could a method be added to cdecl?

```rust
impl CTy {
    fn walk_mut(&mut self, mut f: impl FnMut(&mut Self)) {
        let mut head = self;
        loop {
            f(head);
            match head {
                Self::Named { .. } => break,
                Self::Ptr { ty, .. } => head = ty,
                Self::Array { ty, .. } => head = ty,
                Self::Fn { ret, .. } => head = ret,
            }
        }
    }
}
```

ctest-next/src/translator.rs
```diff
@@ -260,8 +323,50 @@ impl<'a> Translator<'a> {
 
     /// Translate a Rust pointer into its equivalent C pointer.
     fn translate_ptr(&self, ptr: &syn::TypePtr) -> Result<cdecl::CTy, TranslationError> {
-        let inner_type = self.translate_type(ptr.elem.deref())?;
-        Ok(ptr_with_inner(inner_type, ptr.mutability))
+        let modifier = translate_mut(ptr.mutability);
+
```
tgross35: I think this can be improved, I have a suggestion but need to write it up tomorrow

ctest-next/src/translator.rs
```diff
@@ -197,17 +198,79 @@ impl<'a> Translator<'a> {
...
+            .foreign_functions
+            .iter()
+            .find(|f| f.ident == name.into());
+
         let mut parameters = function
             .inputs
             .iter()
-            .map(|arg| self.translate_type(&arg.ty))
+            .enumerate()
+            .map(|(i, arg)| {
```
tgross35: This `map` closure is huge. Can you just make it a standalone function?

ctest-next/templates/test.rs
```diff
@@ -314,6 +314,20 @@ mod generated_tests {
...
+{%- for item in ctx.foreign_fn_tests +%}
+
+    pub fn {{ item.test_name }}() {
+        extern "C" {
+            fn ctest_foreign_fn__{{ item.id }}() -> *const *mut u32;
+        }
+        unsafe {
+            check_same({{ item.id }} as usize,
+                    ctest_foreign_fn__{{ item.id }}() as usize,
+                    "{{ item.id }} function pointer");
```
tgross35: Use the newer `.addr()` rather than `as usize`, and pull the call to `ctest_foreign_fn__` out (since it's the only `unsafe` bit here)

Also could you add a comment about what this test does here?

ctest-next/templates/test.rs
```diff
@@ -314,6 +314,20 @@ mod generated_tests {
         }
     }
 {%- endfor +%}
+
+{%- for item in ctx.foreign_fn_tests +%}
+
+    pub fn {{ item.test_name }}() {
+        extern "C" {
+            fn ctest_foreign_fn__{{ item.id }}() -> *const *mut u32;
```
tgross35: What is `*const *mut u32`?

ctest-next/templates/test.c
```diff
@@ -118,3 +118,11 @@ ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}({{ item.c_ty }} *b) {
 #ifdef _MSC_VER
 #  pragma warning(default:4365)
 #endif
+
+{%- for item in ctx.foreign_fn_tests +%}
+
+typedef {{ item.return_type }};
+ctest_foreign_fn_type__{{ item.id }} ctest_foreign_fn__{{ item.id }}(void) {
+    return {{ item.c_val }};
+}
```
tgross35, line 134: IIUC, this returns a typed pointer that gets implicitly cast to `*const *mut u32`. Couldn't this be simplified to `(void *)` cast the function pointer on the C side, and return `void *` rather than needing to construct the typedef?

mbyx: I know data types can be converted to `void *`, but is converting arrays and function pointers to `void *` platform agnostic?

tgross35: It's tricky; technically the answer is no, function pointers and data pointers do not need to have the same representation. But the current implementation is also currently broken on platforms where they aren't compatible, because the Rust signature does not match.

It is okay to convert function pointers to other function pointer types in C. So I think maybe a solution is to do this:

```c
/* put this at the top of the file */
typedef void (*ctest_void_func)(void);

ctest_void_func ctest_whatever() {
    return (void_func)other_function;
}
```

which just makes the types easier to work with (you never need to write the type of `other_function`). Then the Rust definition will be compatible:

```rust
unsafe extern "C" {
    safe fn ctest_whatever() -> Option<unsafe extern "C" fn() -> ()>;
}
```

And you can use [`fn_addr_eq`](https://doc.rust-lang.org/std/ptr/fn.fn_addr_eq.html) to do the check.

edit: the current implementation is probably actually fine because afaik Rust doesn't support any platforms where they aren't compatible, and the mismatch is at a machine code level rather than a language-level (my original suggestion of casting to `void *` would be language-level). Changing to use `fn_addr_eq` would still be safest, though.

ctest-next/src/template.rs
mbyx: All the string manipulation is doing is removing the name of the function from the signature (since that isn't allowed in bare fn types). I can't do that directly from `syn::Signature` since it's not allowed to have an empty `Ident`.

The panic should actually be an unreachable though.

tgross35: `AST -> string -> modify -> parse to AST` is just a definite code smell to avoid: remember that things like the below, while ugly, is completely valid Rust syntax that this crate may have to deal with in some places :)

```rust
fn foo<const N: (usize)>() {}
fn bar<T: (Copy)>() {}
static FOO: (fn()) = foo::<{ (10) }>;
```

Doing things at an AST level can be more complicated. But that's a good thing: it makes you think about what needs to get passed, what gets modified, and what gets rejected.

> I can't do that directly from `syn::Signature` since it's not allowed to have an empty `Ident`.

Think about it a bit differently: both your input and your output are valid Rust code, so syn must have a representation for it. You just need to figure out what it is and how to translate between them.

In this case, your input is [`Signature`](https://docs.rs/syn/latest/syn/struct.Signature.html) and the output is [`TypeBareFn`](https://docs.rs/syn/latest/syn/struct.TypeBareFn.html). Most things have a 1:1 mapping that you just copy over (you'll need to iterate the args). For anything that doesn't copy over, you should consider if it should be rejected (constness, async, generics, etc).

tgross35: Another option here is to use `quote`, something like this:

```rust
let mut args = Vec::new();
for arg in sig.inputs {
    match arg {
        FnArg::Typed(pat) => args.push(pat),
        FnArg::Receiver(_) => panic!("`self` args are not supported: got `{arg}`),
    }
}
let ty = quote! { fn(#(#args),*) };
```

That's a bit easier than building a `TypeBareFn` from scratch, but still avoids string hazards.

ctest-next/templates/test.c
tgross35: ^ Can this signature be updated to return a pointer to a void function? That avoids needing to write the C type at all

ctest-next/src/template.rs
tgross35: Actually, with https://github.com/rust-lang/libc/pull/4594#discussion_r2266012088 this all goes away right?

ctest-next/templates/test.c
```diff
@@ -118,3 +118,20 @@ ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}({{ item.c_ty }} *b) {
...
+
+typedef void (*ctest_void_func)(void);
+ctest_void_func ctest_foreign_fn__{{ item.id }}(void) {
+    return (ctest_void_func){{ item.c_val }};
+}
+{%- endfor +%}
+
+#ifdef _MSC_VER
+#  pragma warning(default:4191)
+#endif
```
tgross35: Looks like the trailing \n got dropped here

ctest-next/templates/test.c
```diff
@@ -118,3 +118,20 @@ ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}({{ item.c_ty }} *b) {
...
 #endif
+
+#ifdef _MSC_VER
+// Disable function pointer type conversion warnings on MSVC.
+// The conversion may fail only if we call that function, however we only check its address.
+#  pragma warning(disable:4191)
+#endif
+{%- for item in ctx.foreign_fn_tests +%}
+
+typedef void (*ctest_void_func)(void);
```
tgross35: This needs to move to the top of the file, otherwise it will get defined multiple times. This is much cleaner though, thank you!

ctest-next/templates/test.rs
```diff
@@ -314,6 +314,21 @@ mod generated_tests {
...
+
+    /// Check if the Rust and C side function pointers point to the same underlying function.
+    pub fn {{ item.test_name }}() {
+        extern "C" {
+            fn ctest_foreign_fn__{{ item.id }}() -> unsafe extern "C" fn();
+        }
+        let actual = unsafe { ctest_foreign_fn__{{ item.id }}() };
+        let expected: unsafe extern "C" fn() = unsafe {
+            mem::transmute({{ item.id }} as *const ())
+        };
```
tgross35: No `transmute` needed, you can compare function pointers of different types:

```rust
let a: unsafe extern "C" fn(usize) -> *mut c_void = libc::malloc;
let b: unsafe extern "C" fn(*mut c_void) = libc::free;
assert!(!std::ptr::fn_addr_eq(a, b));
```

the `let` binding might be needed so it gives you the right thing.

ctest-next/src/translator.rs
```diff
@@ -260,8 +284,50 @@ impl<'a> Translator<'a> {
...
+                                ty.deref_mut()
+                        {
+                            qual.constness = Constness::Mut;
+                        }
+                    }
+                }
+
+                Ok(ptr_ty)
+            }
+        }
```
tgross35: Is any of this needed anymore, or the change to `apply_volatile_and_array_arg`?

This does need to get fixed, but if it's not needed for this PR than I would rather drop the changes here and merge the rest. I think the fix should look more like what I suggested here https://github.com/rust-lang/libc/pull/4617#discussion_r2256988486 (we can't assume that an array always turns into a pointer)

ctest-next/templates/test.rs
mbyx: That doesn't work, it fails with `FnPtr` not implemented for `*const ()`. It also doesn't work with `{{ item.id }}` directly.

tgross35: Where is the `*const` coming from? Could you paste the code and error?

tgross35: This does seem to work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=378ac3c0e7aa61a0a99911f305b230f2

mbyx: That's because `abort` is a function that has no parameters. If it does then it fails with error (taken from ctest-test):
```rs
expected fn pointer `unsafe extern "C" fn() -> ()`
found fn item `unsafe extern "C" fn(*mut c_void, *mut c_void)
```

snippet causing that:
```rs
let actual = unsafe { ctest_foreign_fn__{{ item.id }}() };
let expected: unsafe extern "C" fn() = {{ item.id }};
check_same(ptr::fn_addr_eq(expected, actual), true, "{{ item.id }} function pointer");
```

Dropping the type coercion of `unsafe extern "C" fn()` completely results in (taken from ctest-next)
```rs
error[E0277]: the trait bound `unsafe extern "C" fn(usize) -> *mut c_void` {foo::malloc}: FnPtr is not satisfied
```

tgross35: Ah, darn, so we do need the type to do the coercion. Alright, let's just do the safe thing; skip the transmute (to avoid some `unsafe`), just `as` cast both function pointers to a `u64` and compare those. Rust allows for this https://doc.rust-lang.org/std/primitive.fn.html#casting-to-and-from-integers (I'm not exactly clear whether or not C does).

That may not work on cheri, but it will be a while before we need to cross that bridge.

mbyx: Done.

Also what is cheri?

tgross35: An experimental arch that has 128-bit pointers :D that's not the address space though, the top half is a tag about what the pointer is allowed to do. (Pretty interesting to look into if you are interested in that sort of thing)

PR #4597, opened 2025-07-29
AIX: Fix the types of 'struct stat'/'struct stat64' fields 'st_*tim'

xingxue-ibm:
# Description
On AIX, when `_ALL_SOURCE` is defined, the `st_*tim` fields in `struct stat`/`struct stat64` use `struct st_timespec`, whereas under POSIX, they use `struct timespec`. This patch updates the type to align with the POSIX declaration.

# Sources

# Checklist


xingxue-ibm:
> Would it make sense to add this as a define at https://github.com/xingxue-ibm/libc/blob/a861a0b0cb16dd0cfcb642c3d294fba754cf6df2/libc-test/build.rs#L5470?

Yeah, I tried that before. Defining `_XOPEN_SOURCE` excludes quite a few non-POSIX functions.

PR #4600, opened 2025-07-30
libc-test: port windows to use ctest-next

mbyx:
# Description
This PR adds support for testing the windows platform using `ctest-next` as well as the original `ctest`. A minor bug was also fixed along the way, where we have some public primitive types that aren't available in C (this can be fixed with skips instead as well).

# Sources

# Checklist


mbyx:
Not sure what the `arm-linux-androideabi` is about. Also since `ctest-next` is more specific about renaming and skipping types it is possible I might put say a struct type in a rename union type block accidentally, and I wouldn't be able to figure it out.

libc-test/build.rs
```diff
@@ -957,6 +959,124 @@ fn test_windows(target: &str) {
     cfg.generate(src_hotfix_dir().join("lib.rs"), "main.rs");
 }
 
+fn test_windows_next(target: &str) {
```
tgross35: Please add a comment about what this is

Also, are there any notable changes from the existing Windows function?

libc-test/test/main_next.rs
```diff
@@ -0,0 +1,6 @@
+#![allow(deprecated)]
```
tgross35: Does this work if you apply it to the `use libc::*` rather than to everything?

libc-test/build.rs
```diff
@@ -164,6 +171,13 @@ fn main() {
     let re = regex::bytes::Regex::new(r"(?-u:\b)crate::").unwrap();
     copy_dir_hotfix(Path::new("../src"), &hotfix_dir, &re, b"::");
 
+    // FIXME(ctest): Only needed until ctest-next supports all tests.
+    std::fs::write(
+        format!("{}/main_next.rs", std::env::var("OUT_DIR").unwrap()),
+        "\nfn main() { println!(\"test result: ok\"); }\n",
+    )
+    .unwrap_or_default();
```
tgross35: I assume this is providing a default for targets that don't yet have `ctest-next`, which gets overwritten by those that do. Could you make the comment saying that?

Also, why `unwrap_or_default()`?

libc-test/test/main_next.rs
tgross35, line 1: Could you name this file `ctest_next`? I know the current one is called `main` but that's a confusing name

libc-test/build.rs
mbyx: Other than different names and arguments to the test generator, the only notable difference is that `ctest-next` doesn't need `src-hotfix`, and funnily enough it breaks in `src-hotfix` anyway, so it uses the crate root directly.

libc-test/test/ctest_next.rs
tgross35, line 1: I didn't mean to delete all of the original ctest *right now* :), only that you can swap a target from ctest to ctest-next rather than running both for that specific target

libc-test/build.rs
```diff
@@ -840,49 +832,46 @@ fn test_windows(target: &str) {
...
         match ty {
             // Just pass all these through, no need for a "struct" prefix
-            "FILE" | "DIR" | "Dl_info" => ty.to_string(),
-
+            "FILE" | "DIR" | "Dl_info" => ty.to_string().into(),
+            t if t.ends_with("_t") => t.to_string().into(),
+            // Windows uppercase structs don't have `struct` in fr.into()ont:
+            t if ty.chars().next().unwrap().is_uppercase() => t.to_string().into(),
+            "stat" => "struct __stat64".to_string().into(),
+            "utimbuf" => "struct __utimbuf64".to_string().into(),
```
tgross35, line 843: What's the difference with original ctest here?

libc-test/build.rs
```diff
@@ -840,49 +832,46 @@ fn test_windows(target: &str) {
...
+            _ => None,
+        }
+    });
+    cfg.rename_type(move |ty| {
```
tgross35, line 847: In ctest we should rename this function `rename_alias` or `rename_type_alias`, I was thinking it handled all types for a bit

libc-test/build.rs
mbyx: `ctest` uses a single method for renaming types, whereas we have multiple, so that gets split up.

libc-test/test/ctest_next.rs
mbyx: So I should keep `test_windows` but only run `test_windows_next`?

libc-test/build.rs
mbyx: It handles all non struct and union types, we don't actually have a `rename_alias` method (although we can add it), since `ctest` doesn't use it. But `rename_alias` would be a different method (`rename_type` uses `MapInput::Type`, `rename_alias` would use `MapInput::Alias`).

tgross35: Is `rename_type` just a catch-all for renaming anything? If so, tbh I think we should remove it: it's better to be more specific about what you expect to rename

libc-test/test/ctest_next.rs
tgross35: Yes please :) though you can still call it `test_windows` since eventually there will be no `next`

I just don't want to lose testing for everything else, there will be PRs merging in the meantime

mbyx: I'm not sure I understand. The way I changed it, `test_windows` used `ctest-next` and generated `ctest_output.rs`. The other tests are unchanged to use `ctest`. So the only thing being lost is the ctest version of the windows test.

tgross35: Oh, I think I get it. Sorry, didn't realize that it uses the same output file name.

PR #4601, opened 2025-07-30
ctest: add foreign static test

mbyx:
# Description
Adds support for testing extern statics.

Also adds support for parsing any dimensional array in `make_cdecl`, and makes it easier to extract ffi safe types from `Option`.

# Sources

# Checklist

### DIFF
--- a/ctest-next/src/template.rs
+++ b/ctest-next/src/template.rs
@@ -50,6 +50,7 @@ impl CTestTemplate {
 /// Stores all information necessary for generation of tests for all items.
 #[derive(Clone, Debug, Default)]
 pub(crate) struct TestTemplate {
+    pub foreign_static_tests: Vec<TestForeignStatic>,
     pub field_ptr_tests: Vec<TestFieldPtr>,
     pub field_size_offset_tests: Vec<TestFieldSizeOffset>,
     pub roundtrip_tests: Vec<TestRoundtrip>,
@@ -77,6 +78,7 @@ impl TestTemplate {
         template.populate_field_ptr_tests(&helper)?;
         template.populate_roundtrip_tests(&helper)?;
         template.populate_foreign_fn_tests(&helper)?;
+        template.populate_foreign_static_tests(&helper)?;
 
         Ok(template)
     }
@@ -414,6 +416,28 @@ impl TestTemplate {
 
         Ok(())
     }
+
+    /// Populates tests for foreign statics, keeping track of the names of each test.
+    fn populate_foreign_static_tests(
+        &mut self,
+        helper: &TranslateHelper,
+    ) -> Result<(), TranslationError> {
+        for static_ in helper.ffi_items.foreign_statics() {
+            let rust_ty = static_.ty.to_token_stream().to_string().into_boxed_str();
+
+            let item = TestForeignStatic {
+                test_name: static_test_ident(static_.ident()),
+                id: static_.ident().into(),
+                c_val: helper.c_ident(static_).into_boxed_str(),
+                rust_ty,
+            };
+
+            self.foreign_static_tests.push(item.clone());
+            self.test_idents.push(item.test_name);
+        }
+
+        Ok(())
+    }
 }
 
 /* Many test structures have the following fields:
@@ -498,6 +522,14 @@ pub(crate) struct TestForeignFn {
     pub id: BoxStr,
 }
 
+#[derive(Clone, Debug)]
+pub(crate) struct TestForeignStatic {
+    pub test_name: BoxStr,
+    pub id: BoxStr,
+    pub c_val: BoxStr,
+    pub rust_ty: BoxStr,
+}
+
 fn signededness_test_ident(ident: &str) -> BoxStr {
     format!("ctest_signededness_{ident}").into()
 }
@@ -528,6 +560,10 @@ fn roundtrip_test_ident(ident: &str) -> BoxStr {
 
 fn foreign_fn_test_ident(ident: &str) -> BoxStr {
     format!("ctest_foreign_fn_{ident}").into()
+}
+
+fn static_test_ident(ident: &str) -> BoxStr {
+    format!("ctest_static_{ident}").into()
 }
 
 /// Wrap methods that depend on both ffi items and the generator.
--- a/ctest-next/templates/test.c
+++ b/ctest-next/templates/test.c
@@ -137,3 +137,12 @@ ctest_void_func ctest_foreign_fn__{{ item.id }}(void) {
 #ifdef _MSC_VER
 #  pragma warning(default:4191)
 #endif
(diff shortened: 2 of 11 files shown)


ctest-next/templates/test.rs
```diff
@@ -314,6 +314,37 @@ mod generated_tests {
         }
     }
 {%- endfor +%}
+
+{%- for static_ in ctx.foreign_static_tests +%}
+
+    // Tests if the pointer to the static variable matches in both Rust and C.
+    pub fn {{ static_.test_name }}() {
+    {%- if static_.rust_ty.contains("extern") +%}
```
tgross35: What exactly is different about this block?

mbyx: In one we return the `fn ptr` directly and in the other we return a reference to the static.

tgross35: Why/how is that identified by `rust_ty.contains("extern")`?

I'm also still not sure, how is this different from the function part added in https://github.com/rust-lang/libc/pull/4594?

ctest-next/templates/test.c
```diff
@@ -118,3 +118,19 @@ ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}({{ item.c_ty }} *b) {
...
+{%- if static_.rust_ty.contains("extern") +%}
+typedef {{ static_.return_type }};
+ctest_static_ty__{{ static_.id }} ctest_static__{{ static_.id }}(void) {
+    return {{ static_.c_val }};
+}
+{%- else +%}
+typedef {{ static_.volatile_keyword }}{{ static_.c_mutable }}{{ static_.return_type }};
+ctest_static_ty__{{ static_.id }} ctest_static__{{ static_.id }}(void) {
+    return &{{ static_.c_val }};
+}
```
tgross35: Since these return pointers, can they cast them on the C side and return `void *`? Rather than needing the more complex definition

ctest-next/templates/test.rs
```diff
@@ -314,6 +314,37 @@ mod generated_tests {
...
+    {%- else +%}
+        extern "C" {
+            fn ctest_static__{{ static_.id }}() -> *{{ static_.mutable }} {{ static_.rust_ty }};
+        }
+        unsafe {
+            // We must use addr_of! here because of https://github.com/rust-lang/rust/issues/114447
+            check_same(std::ptr::addr_of!({{ static_.id }}) as usize,
+                    ctest_static__{{ static_.id }}() as usize,
+                    "{{ static_.id }} static");
+        }
```
tgross35, line 337: Few small things:
 
* `addrr_of!(x)` should be `&raw const x` in more modern Rust. I don't think this needs a comment, it's standard practice nowadays
* Prefer `.addr()` rather than `as usize`. 
* Only the call to `ctest_static` is unsafe, the rest doesn't need to be in the block

ctest-next/templates/test.rs
mbyx: It isn't exactly different, in that that block and the function part are both to return function pointers, just one is used for actual foreign functions and this for extern statics.

tgross35: Oh I see, I was reading this wrong. 

General theme comment from a couple PRs - we have an AST for the Rust side and an IR for C, doing things with those should almost always be preferred over string manipulation. It's much easier to be robust and avoid edge cases that way, e.g. this would trip over a `struct external`. (You have the `static_.ty` in `template.rs`).

ctest-next/src/template.rs
```diff
@@ -380,6 +383,66 @@ impl TestTemplate {
...
+
+            let item = TestForeignStatic {
+                test_name: static_test_ident(static_.ident()),
+                id: static_.ident().into(),
+                c_val: helper.c_ident(static_).into_boxed_str(),
+                rust_ty,
+                is_bare_fn,
+                mutable: if static_.mutable { "mut" } else { "const" }.into(),
+                volatile_keyword: { if is_volatile { "volatile " } else { "" }.into() },
+                c_mutable: if c_ty.contains("const") || static_.mutable {
```
tgross35: Nit: the `*mutable` names sound a bit like booleans. Maybe `ptr_mutability_kw` and `c_const_kw`?

ctest-next/src/lib.rs
```diff
@@ -107,3 +108,21 @@ impl<'a> From<&'a Union> for MapInput<'a> {
         MapInput::Union(u)
     }
 }
+
+/* A simple helper to find any nested bare fn types, such as `Option<unsafe extern "C" fn()>` */
+
+struct BareFnFinder {
```
tgross35: This can be a doc comment on the struct

ctest-next/src/template.rs
```diff
@@ -457,6 +520,19 @@ pub(crate) struct TestRoundtrip {
...
 
+#[derive(Clone, Debug)]
+pub(crate) struct TestForeignStatic {
+    pub test_name: BoxStr,
+    pub id: BoxStr,
+    pub c_val: BoxStr,
+    pub rust_ty: BoxStr,
+    pub is_bare_fn: bool,
+    pub mutable: BoxStr,
+    pub c_mutable: BoxStr,
```
tgross35: Nit: these can be `&'static str`

ctest-next/templates/test.c
tgross35: @mbyx this one still needs to be changed. It's also checking for an `extern` substring still.

The change should look something like https://github.com/rust-lang/libc/pull/4594#discussion_r2264939261 for the function pointer part.

ctest-next/src/template.rs
tgross35: With https://github.com/rust-lang/libc/pull/4601#discussion_r2264698834 these actually shouldn't be needed.

ctest-next/templates/test.rs
```diff
@@ -314,6 +314,34 @@ mod generated_tests {
...
+    pub fn {{ static_.test_name }}() {
+    {%- if static_.is_bare_fn +%}
+        extern "C" {
+            fn ctest_static__{{ static_.id }}() -> {{ static_.rust_ty }};
+        }
+        let actual = unsafe {
+            std::mem::transmute::<_, *const ()>(*(&raw const {{ static_.id }})).addr()
+        };
+        let expected = unsafe {
+            std::mem::transmute::<_, *const ()>(ctest_static__{{ static_.id }}()).addr()
```
tgross35: Add `self` to the import at https://github.com/mbyx/libc/blob/494e5e5627035b847984501faaa8d34e4c40a9d4/ctest-next/templates/test.rs#L19

tgross35: Please add comments, this is tricky. Also, make sure this branch is covered in one of the tests.

Issue #1637, opened 2020-01-05
Using the correct type of sighandler_t windows


vorner:
I'm opening this here as continuation of discussion on #1626. This also might be related to #1359.

Currently, `sighandler_t` is an alias to `usize` on Windows targets. While this works, it is strictly speaking wrong and the right type should be something like `Option<extern "C" fn(c_int) -> ()>`.

There are, however, concerns about the switch:
* It is an API change that might break someone's code. At least a crater run would be needed.
* What is the correct type? Due to the fact this doesn't always contain a valid function pointer, but also some marker constants (eg. `SIG_IGN`, which has value of 1), it is unclear if this is correct on the Rust side. It seems the consensus is function pointers need to be non-null (solved by the Option) and properly aligned, but the alignment of function pointer on all platforms Windows currently run is 1, so for that reason this should be fine.

vorner:
Specifically in answer to this https://github.com/rust-lang/libc/pull/1626#issuecomment-570232985

> How does this work in C?
>
> IIRC, C requires function pointers to be properly aligned as well, at least when you call them, right ?

I guess it doesn't call these weird constants, eg on the windows kernel side it might look like something like:

```C
switch (handler) {
  case SIG_DFL: // = 0
    do_default();
    break;
  case SIG_IGN: // = 1
    break;
  ...
  default:
    handler();
    break;
}
```

Therefore, in case it is one of the constants, the value is never used as a pointer, only compared. Which you probably can do in C, especially if you are the platform owner and control the compiler.

I was worried if, on the Rust side, it could be UB just to *create* the invalid function pointer ‒ at least, it is insta-UB to create an enum with invalid discriminant or reference to invalid memory, you don't even have to use it.

nico-abram:
Maybe a `MaybeUninit<Option<extern "C" fn(c_int) -> ()>>` could be used to make sure the alignment requirement doesn't trigger UB?

tgross35:
Here's one msys reference for newlib https://github.com/msys2/msys2-runtime/blob/7484c5594806de5751a264600f648d246851fa15/newlib/libc/include/signal.h#L15C9-L15C22 which makes `sighandler_t` a `_sig_func_ptr`. I assume `Option<extern "C" fn(_: c_int) -> ()>` would be the correct type here, but I can't find a source (please post a header link if you have one).

PR welcome to change this in 1.0.

retep998:
[...]

PR #4602, opened 2025-07-31
libc: use more accurate type of sighandler_t on windows

mbyx:
# Description
Changes the type of `sighandler_t` on windows from `usize` to `*const c_void` as per #1637 to more accurately represent the fact that it is a pointer to a function (since the value of the address is also sometimes matched without dereferencing it).

A better type would be `Option<extern fn ...>` however because of the dual use of `sighandler_t` as a constant, it would be undefined behavior to construct that type from something like `5`.

# Sources
https://github.com/msys2/msys2-runtime/blob/23a25d49e3b8c18f12ce588869d4cb24de9f2454/newlib/libc/machine/cris/sys/signal.h#L23

# Checklist

### DIFF
--- a/src/windows/mod.rs
+++ b/src/windows/mod.rs
@@ -10,7 +10,7 @@ pub type ptrdiff_t = isize;
 pub type intptr_t = isize;
 pub type uintptr_t = usize;
 pub type ssize_t = isize;
-pub type sighandler_t = usize;
+pub type sighandler_t = *const c_void;
 
 pub type wchar_t = u16;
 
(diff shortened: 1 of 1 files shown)


src/windows/mod.rs
```diff
@@ -10,7 +10,7 @@ pub type ptrdiff_t = isize;
 pub type intptr_t = isize;
 pub type uintptr_t = usize;
 pub type ssize_t = isize;
-pub type sighandler_t = usize;
+pub type sighandler_t = *const c_void;
```
tgross35, line 13: One concern: if Rust ever does get raw function pointers, it would be a breaking change to start using them. So I'm thinking maybe we should create an opaque type that would let us change to a raw function pointer at some point, ideally something compatible with https://github.com/rust-lang/libs-team/issues/589#issuecomment-2981331798.

Will need to think about this a bit

PR #4605, opened 2025-08-01
NetBSD: add `ptsname_r()` support

0-wiz-0:
# Description

This adds NetBSD support for `ptsname_r`. The function was added in NetBSD 7.

# Sources

https://github.com/NetBSD/src/blob/0b011669be2b100a2339b1818db4629981dcdbf0/include/stdlib.h#L326

# Checklist

  - haven't done this since I don't know what 'mytarget' is and 'netbsd' didn't work.
 

I based this on the FreeBSD version, I'm not sure what the sort order in the file is supposed to be though.


0-wiz-0:
Ok, this included quite too many diffs.

PR #4607, opened 2025-08-01
Restore some non-POSIX functions for AIX

xingxue-ibm:
# Description
These non-POSIX functions were previously removed because they are guarded by the `_KERNEL` macro in AIX headers. After discussion, we thought they might be useful, so this patch restores them with fixes.

# Sources

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -5738,6 +5738,14 @@ fn test_aix(target: &str) {
             "setdomainname" | "settimeofday" | "statfs" | "statfs64" | "statx" | "swapoff"
             | "swapon" | "utmpname" | "setgroups" => true,
 
+            // These non-POSIX functions are guarded by the _KERNEL macro in the AIX headers.
+            "recvmmsg" | "sendmmsg" | "sctp_opt_info" | "sctp_peeloff" | "sethostid"
+            | "sethostname" | "splice" => true,
+
+            // 'mount' is available in the system's libc.a and has a man page, but it is
+            // not declared in the AIX headers."
+            "mount" => true,
+
             _ => false,
         }
     });
--- a/src/unix/aix/mod.rs
+++ b/src/unix/aix/mod.rs
@@ -3027,6 +3027,7 @@ extern "C" {
     pub fn mincore(addr: caddr_t, len: size_t, vec: *mut c_char) -> c_int;
     pub fn mkfifoat(dirfd: c_int, pathname: *const c_char, mode: mode_t) -> c_int;
     pub fn mknodat(dirfd: c_int, pathname: *const c_char, mode: mode_t, dev: dev_t) -> c_int;
+    pub fn mount(device: *const c_char, path: *const c_char, flags: c_int) -> c_int;
     pub fn mprotect(addr: *mut c_void, len: size_t, prot: c_int) -> c_int;
     pub fn mq_close(mqd: crate::mqd_t) -> c_int;
     pub fn mq_getattr(mqd: crate::mqd_t, attr: *mut crate::mq_attr) -> c_int;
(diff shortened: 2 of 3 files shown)


src/unix/aix/mod.rs
```diff
@@ -3226,6 +3234,14 @@ extern "C" {
         policy: c_int,
         param: *const crate::sched_param,
     ) -> c_int;
+    pub fn sctp_opt_info(
+        sd: c_int,
+        id: c_uint,
+        opt: c_int,
+        arg_size: *mut c_void,
+        size: *mut size_t,
+    ) -> c_int;
```
tgross35: Per https://www.ibm.com/docs/en/aix/7.2.0?topic=s-sctp-opt-info-subroutine, `opt` is `sctp_assoc_t`. I assume this is defined to `unsigned`, but is it worth a typedef?

src/unix/aix/mod.rs
```diff
@@ -3239,13 +3255,16 @@ extern "C" {
     pub fn semget(key: crate::key_t, nsems: c_int, semflag: c_int) -> c_int;
     pub fn semop(semid: c_int, sops: *mut sembuf, nsops: size_t) -> c_int;
     pub fn send_file(socket: *mut c_int, iobuf: *mut sf_parms, flags: c_uint) -> ssize_t;
+    pub fn sendmmsg(sockfd: c_int, msgvec: *mut mmsghdr, vlen: c_uint, flags: c_int) -> c_int;
```
tgross35, line 3259: It looks like this only has three fields at https://www.ibm.com/docs/en/aix/7.3.0?topic=s-sendmmsg-subroutine, is that inaccurate?

libc-test/build.rs
```diff
@@ -5738,6 +5738,14 @@ fn test_aix(target: &str) {
             "setdomainname" | "settimeofday" | "statfs" | "statfs64" | "statx" | "swapoff"
             | "swapon" | "utmpname" | "setgroups" => true,
 
+            // These non-POSIX functions are guarded by the _KERNEL macro in the AIX headers.
+            "recvmmsg" | "sendmmsg" | "sctp_opt_info" | "sctp_peeloff" | "sethostid"
+            | "sethostname" | "splice" => true,
+
+            // 'mount' is available in the system's libc.a and has a man page, but it is
+            // not declared in the AIX headers."
+            "mount" => true,
```
tgross35: Just checking, are these somewhat common to use on AIX even though they aren't defined? `mount` on other platforms usually has a `void *data` arg of fs-specific flags, I'm wondering if there's a chance the AIX version changes to take one as well.

`sendmmsg`/`recvmmsg` are pretty common, I'm surprised nobody has moved your `#ifdef __KERNEL` a few lines lower yet 🙂

src/unix/aix/mod.rs
xingxue-ibm: Thanks for checking. Yeah, there seems to be some AIX header weirdness. The AIX man page specifies the type `sctp_assoc_t` for the argument `id`, but the prototype in the header files has type `uint32_t` for it.
Man page:
```
       int sctp_opt_info(sd, id, opt, *arg_size, *size);
       int sd;
       sctp_assoc_t id;
       int opt;
       void *arg_size;
       size_t *size;
```
`net/proto_uipc.h`:
```
int             sctp_opt_info PROTO((int, uint32_t, int, void *, size_t *));
```
I agree it makes it consistent with the man page if we define its type as `sctp_assoc_t` on the Rust side. The typedef of `sctp_assoc_t` is defined in header `netinet/sctp.h` so I am pulling the header in. `sctp_peeloff` and `sctp_opt_info` are also declared in `netinet/sctp.h` without the `_KERNEL` guard so I am removing them from the function skip list in `build.rs`.
```
typedef uint32_t sctp_assoc_t;
int sctp_peeloff(int s, uint32_t * assoc_id);
int sctp_opt_info(int s, uint32_t id, int opt, void *arg, size_t *size);
int sctp_getladdrs(int sd, sctp_assoc_t id, struct sockaddr **addrs);
int sctp_getpaddrs(int sd, sctp_assoc_t id, struct sockaddr **addrs);
```

src/unix/aix/mod.rs
xingxue-ibm: The online doc does not seem to be accurate. The prototype in AIX header `net/proto_uipc.h` is as follows. I will open an issue against the AIX doc.
```
int             sendmmsg PROTO((int, struct mmsghdr *, unsigned int, int));
```

libc-test/build.rs
xingxue-ibm: Thanks again for checking. I checked the documentation for the latest AIX release (v7.3), and `mount` still takes three arguments. By definition, `_KERNEL` is meant to guard code used for writing kernel extensions, but it seems some of these functions are also used in OSS. Yes, `sendmmsg` and `recvmmsg` should be moved out of the block. I will open an issue against the AIX headers for these functions as well as `mount`.

xingxue-ibm:
Thank you for checking and pointing out the discrepancies in the man pages.

Issue #1453, opened 2019-08-02
Wrap padding fields in `Padding(MaybeUninit<T>)`


gnzlbg:
libc uses integers and aggregates of integers for explicit padding byte fields. When these fields are returned from C, they might be uninitialized. We currently do not have any wording in the spec saying that an uninitialized integer is not instant undefined behavior, and there are proposals that define them to both be ok, and to actually be instant undefined behavior. 

These padding fields should use `MaybeUninit<INT>` instead.

The right way to actually fix this issue is to actually remove these padding fields. They only exist because `repr(align(N))`, `repr(packed(N))`, `union`s, etc. did not exist when these types were added. For toolchains that actually support `MaybeUninit`, we can actually do better and don't use any kind of padding fields at all. For toolchains that do not support `repr(align(N))`, `repr(packed(N))`, `union`s, etc. `MaybeUninit` does not exist and cannot be actually implemented in Rust, so there is nothing we can do.

For C types using FAM, there is no way to avoid undefined behavior in Rust today, so we have to use explicit padding fields.

Removing these fields or changing their type are major breaking changes when these fields are public. Most of these fields are private, but some of them aren't. 

Rust currently does not exploit this potential undefined behavior for anything, and for most targets, it cannot actually exploit it since the C library is dynamically linked or linked without cross-lang-lto. If we start using cross-lang-lto, e.g., when statically-linking musl, this could become an issue in the future.

gnzlbg:
cc @emilio - I think `rust-bindgen` probably has the same issues when it comes to automatically-generated explicit padding fields. The constraints of fixing that there are probably the same as the constraints of fixing that here: where MaybeUninit is available, one can probably do better and not emit padding fields in the first place, and where MaybeUninit is not available, there is no way to avoid the UB anyways. For FAMs, I don't think rust-bindgen can do anything either.

emilio:
Yes, that's right. I've opened https://github.com/rust-lang/rust-bindgen/issues/1606 to track this.

DemiMarie-parity:
I don’t believe that this is likely to cause problems in practice: C code in system libraries is not visible to the Rust compiler, so `rustc` has no way to know that these fields *are* uninitialized.  bindgen has a bigger problem because it may be used with LTO.

glandium:
> C code in system libraries is not visible to the Rust compiler, so rustc has no way to know that these fields are uninitialized.

Actually, that's not true with linker-plugin-lto.

[...]

PR #4609, opened 2025-08-02
libc: wrap padding fields with Padding newtype.

mbyx:
# Description
Changes the types of fields with name `__pad`, `_pad`, `__padding`, `_padding`, `__rsvd`, `__unused` to use a new type wrapper around `MaybeUninit` as per #1453 

# Sources

# Checklist


src/types.rs
```diff
@@ -0,0 +1,12 @@
+use core::mem::MaybeUninit;
+
+/// A transparent wrapper over `MaybeUninit<T>` to represent uninitialized padding.
```
tgross35:
```suggestion
/// A transparent wrapper over `MaybeUninit<T>` to represent uninitialized padding
/// while providing `Default`.
```

To clarify why we don't just use `Default`

tgross35:
I think most of the errors should be resolved with 141aea1c5c1d13c6193056862a04000d860ece51. If you would be able to rebase again, I think this might be able to merge with a few smaller deltas.

@rustbot author

mbyx:
Rebased. Seems to error because `Padding` doesn't implement `Hash`, `Eq`, and `PartialEq`, because the `s` macro tries to implement it for the struct. Would implementing those for `Padding` be well defined?

Also it seems that the style check fails at the first formatting error? My editor seems to have a different style for imports than `libc` so I may have caused breakage in more files than just the one that the style check shows.

src/types.rs
tgross35, line 1: Could you split the changes to this file to a separate commit? It's distinct from the rest of the mechanical changes in this PR.

src/types.rs
```diff
@@ -30,6 +32,21 @@ impl<T: Copy> fmt::Debug for Padding<T> {
...
+impl<T: Copy> Hash for Padding<T> {
+    fn hash<H: hash::Hasher>(&self, _state: &mut H) {}
+}
+
+#[cfg(feature = "extra_traits")]
+impl<T: Copy> PartialEq for Padding<T> {
+    fn eq(&self, _other: &Self) -> bool {
+        true
+    }
+}
```
tgross35, line 47: These are user-facing, so add a doc comment to each of these impls saying their behavior (i.e. that they give the results they do to act like the fields don't exist)

src/unix/nto/neutrino.rs
```diff
@@ -1245,13 +1245,13 @@ extern "C" {
         __id: crate::clockid_t,
         _new: *const crate::_clockperiod,
         __old: *mut crate::_clockperiod,
-        __reserved: c_int,
+        __reserved: Padding<c_int>,
```
Hoverbear, line 1248: I believe the changes to this function may be unintended. This results in a build failure on nto:

```
error: type `types::Padding<i32>` is more private than the item `neutrino::ClockPeriod`
    --> src/unix/nto/neutrino.rs:1244:5
     |
1244 | /     pub fn ClockPeriod(
1245 | |         __id: crate::clockid_t,
1246 | |         _new: *const crate::_clockperiod,
1247 | |         __old: *mut crate::_clockperiod,
1248 | |         __reserved: Padding<c_int>,
1249 | |     ) -> c_int;
     | |_______________^ function `neutrino::ClockPeriod` is reachable at visibility `pub`
     |
note: but type `types::Padding<i32>` is only usable at visibility `pub(crate)`
    --> src/types.rs:17:1
     |
  17 | pub(crate) struct Padding<T: Copy>(MaybeUninit<T>);
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     = note: `-D private-interfaces` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(private_interfaces)]`
```

tgross35: Crosslinking since I stumbled back across this: this was resolved by b7ee7452756c184378e3acacba7103daf3bb3b94

PR #4610, opened 2025-08-02
Libc test apple port

mbyx:
# Description
Ports the apple platform to also be tested using `ctest-next`.

# Sources

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -78,7 +78,9 @@ fn ctest_cfg() -> ctest::TestGenerator {
 }
 
 fn ctest_next_cfg() -> ctest_next::TestGenerator {
-    ctest_next::TestGenerator::new()
+    let mut cfg = ctest_next::TestGenerator::new();
+    cfg.skip_private(true);
+    cfg
 }
 
 fn do_semver() {
@@ -217,7 +219,8 @@ fn test_apple(target: &str) {
     let x86_64 = target.contains("x86_64");
     let i686 = target.contains("i686");
 
-    let mut cfg = ctest_cfg();
+    let mut cfg = ctest_next_cfg();
+
     cfg.flag("-Wno-deprecated-declarations");
     cfg.define("__APPLE_USE_RFC_3542", None);
 
@@ -330,11 +333,12 @@ fn test_apple(target: &str) {
         [x86_64]: "crt_externs.h",
     }
 
-    cfg.skip_struct(move |ty| {
-        if ty.starts_with("__c_anonymous_") {
-            return true;
-        }
-        match ty {
+    // Skip anonymous unions/structs.
+    cfg.skip_union(|u| u.ident().starts_with("__c_anonymous_"));
+    cfg.skip_struct(|s| s.ident().starts_with("__c_anonymous_"));
+
+    cfg.skip_struct(|s| {
+        match s.ident() {
             // FIXME(macos): The size is changed in recent macOSes.
             "malloc_zone_t" => true,
             // it is a moving target, changing through versions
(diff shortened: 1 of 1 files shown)


libc-test/build.rs
```diff
@@ -330,11 +331,14 @@ fn test_apple(target: &str) {
...
     }
 
-    cfg.skip_struct(move |ty| {
-        if ty.starts_with("__c_anonymous_") {
-            return true;
-        }
-        match ty {
+    // Skip anonymous unions/structs.
+    cfg.skip_union(|u| u.ident().starts_with("__c_anonymous_"));
+    cfg.skip_struct(|s| s.ident().starts_with("__c_anonymous_"));
```
tgross35, line 338: At some point it would probably be good to figure out a way to test these. Maybe something that could be used like:

```rust
let re = Regex::new(r"^__c_anonymous_(?P<sname>\w+)_(?P<fname>\w+)$");
cfg.anonymous_struct(|u| {
    let name = s.ident();
    if name.starts_with("__c_anonymous_") {
        let caps = re.captures(name).unwrap();
        return Some((caps["sname"], caps["fname"]);
    }
    None
)};
```

to identify the struct and field they belong to, then building a test should be pretty easy.

But that's something to do much later :)

libc-test/build.rs
```diff
@@ -430,39 +433,28 @@ fn test_apple(target: &str) {
...
-        }
-    });
+    cfg.volatile_struct_field(|s, f| s.ident() == "aiocb" && f.ident() == "aio_buf");
 
-    cfg.type_name(move |ty, is_struct, is_union| {
-        match ty {
-            // Just pass all these through, no need for a "struct" prefix
-            "FILE" | "DIR" | "Dl_info" => ty.to_string(),
-
-            // OSX calls this something else
```
tgross35, line 446: Keep this note with the sighandler_t rename (really not all that useful but at least shows it's intentional)

libc-test/build.rs
```diff
@@ -430,39 +433,28 @@ fn test_apple(target: &str) {
...
+        match f.ident() {
+            n if n.ends_with("_nsec") && s.ident().starts_with("stat") => {
+                n.replace("e_nsec", "espec.tv_nsec").into()
             }
             // FIXME(macos): sigaction actually contains a union with two variants:
             // a sa_sigaction with type: (*)(int, struct __siginfo *, void *)
             // a sa_handler with type sig_t
-            "sa_sigaction" if struct_ == "sigaction" => "sa_handler".to_string(),
-            s => s.to_string(),
+            "sa_sigaction" if s.ident() == "sigaction" => "sa_handler".to_string().into(),
```
tgross35: Nit: use `Some(...)` rather than `.into()`, more clear what the conversion is

libc-test/build.rs
```diff
@@ -416,8 +417,10 @@ fn test_apple(target: &str) {
         }
     });
 
-    cfg.skip_field(move |struct_, field| {
-        match (struct_, field) {
+    cfg.skip_struct_field(move |struct_, field| {
+        match (struct_.ident(), field.ident()) {
+            // FIXME(union): actually a union
+            ("sigevent", "sigev_value") => true,
```
tgross35: I don't see this in the before code, is it needed? There were some recent changes to sigevent, make sure to rebase if you haven't in the past couple of days

libc-test/build.rs
```diff
@@ -330,11 +331,14 @@ fn test_apple(target: &str) {
...
-        }
-        match ty {
+    // Skip anonymous unions/structs.
+    cfg.skip_union(|u| u.ident().starts_with("__c_anonymous_"));
+    cfg.skip_struct(|s| s.ident().starts_with("__c_anonymous_"));
+
+    cfg.skip_struct(|s| {
+        match s.ident() {
+            // FIXME(union): actually a union
+            "sigval" => true,
```
tgross35: This one looks new too

libc-test/build.rs
```diff
@@ -217,7 +217,8 @@ fn test_apple(target: &str) {
     let x86_64 = target.contains("x86_64");
     let i686 = target.contains("i686");
 
-    let mut cfg = ctest_cfg();
+    let mut cfg = ctest_next_cfg();
+    cfg.skip_private(true);
```
tgross35: Can this go in `ctest_next_cfg` since it will always be true?

It would be nice if we could figure it out some way to test private items eventually since there are some that come from C definitions but we don't expose. Not sure how to do this since they aren't public, maybe ctest could paste them into the output file from the AST.

libc-test/build.rs
tgross35: Actually it would be easier than that, `cfg.is_field_anonymous` would be enough since `ctest` already knows the type. Opened https://github.com/rust-lang/libc/issues/4650.

libc-test/build.rs
mbyx: I'll have to look at the linux port a bit more carefully then, since it was in a separate function for a long time rebasing didn't catch that. This problem won't affect other ports, since only linux, windows, and apple were based on a much older commit.

PR #4613, opened 2025-08-03
ctest-next: miscellaneous filtering bug fixes

mbyx:
# Description
Prepares for porting `libc-test` platforms to use `ctest-next` by fixing numerous bugs. rust-lang/libc#4600 and rust-lang/libc#4610 both depend on this PR, so this should be merged first.

Fixes:
- Excludes libc-test from MSRV checks since it will use `ctest-next` in the future, which is a edition 2024 crate.
- Modify `translate_expr` to support translating `[ty; 3usize]` properly.
- Adds a new test that for now always passes, but will be gradually used as the test for `ctest-next` backed platforms.
- Moves filtering `ffi_items` to the `TranslateHelper` since `template.rs` does the test filtering for consistency.
- Fixes a bug where unions were not being skipped properly.
- Fixes a bug where structs that were skipped for testing were not recognized as structs when testing a different item with that same struct as a field that shouldn't be skipped.
- A bug where sometimes pointers to structs would be translated as `struct const StructName*`, and sometimes the struct label would not be applied at all.

# Sources

# Checklist

### DIFF
--- a/ctest-next/src/template.rs
+++ b/ctest-next/src/template.rs
@@ -3,6 +3,7 @@ use proc_macro2::Span;
 use quote::ToTokens;
 use syn::spanned::Spanned;
 
+use crate::cdecl::Constness;
 use crate::ffi_items::FfiItems;
 use crate::translator::{TranslationErrorKind, Translator};
 use crate::{
@@ -86,7 +87,7 @@ impl TestTemplate {
         &mut self,
         helper: &TranslateHelper,
     ) -> Result<(), TranslationError> {
-        for constant in helper.ffi_items.constants() {
+        for constant in helper.filtered_ffi_items.constants() {
             if let syn::Type::Ptr(ptr) = &constant.ty
                 && let syn::Type::Path(path) = &*ptr.elem
                 && path.path.segments.last().unwrap().ident == "c_char"
@@ -124,7 +125,7 @@ impl TestTemplate {
         &mut self,
         helper: &TranslateHelper,
     ) -> Result<(), TranslationError> {
-        for alias in helper.ffi_items.aliases() {
+        for alias in helper.filtered_ffi_items.aliases() {
             let item = TestSizeAlign {
                 test_name: size_align_test_ident(alias.ident()),
                 id: alias.ident().into(),
@@ -134,7 +135,7 @@ impl TestTemplate {
             self.size_align_tests.push(item.clone());
             self.test_idents.push(item.test_name);
         }
-        for struct_ in helper.ffi_items.structs() {
+        for struct_ in helper.filtered_ffi_items.structs() {
             let item = TestSizeAlign {
                 test_name: size_align_test_ident(struct_.ident()),
                 id: struct_.ident().into(),
@@ -144,7 +145,7 @@ impl TestTemplate {
             self.size_align_tests.push(item.clone());
             self.test_idents.push(item.test_name);
         }
-        for union_ in helper.ffi_items.unions() {
+        for union_ in helper.filtered_ffi_items.unions() {
             let item = TestSizeAlign {
                 test_name: size_align_test_ident(union_.ident()),
                 id: union_.ident().into(),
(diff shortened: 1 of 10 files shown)


ctest-next/src/template.rs
```diff
@@ -483,20 +480,61 @@ fn roundtrip_test_ident(ident: &str) -> BoxStr {
 
 /// Wrap methods that depend on both ffi items and the generator.
 pub(crate) struct TranslateHelper<'a> {
+    filtered_ffi_items: FfiItems,
     ffi_items: &'a FfiItems,
```
tgross35, line 537: Does this need to keep the unfiltered `ffi_items` at all anymore? It seems cleaner to only keep the filtered items here, if possible.

libc-test/test/main_next.rs
```diff
@@ -0,0 +1,5 @@
+#![allow(unused_imports, deprecated)]
+
+use libc::*;
```
tgross35: Can `unused_imports` apply only to this `libc::*` import? So we don't apply it to the contents of `main_next.rs`.

ctest-next/src/template.rs
mbyx: It's to fix one of the bugs, if say a struct `T1Bar` is skipped for testing it won't be in `filtered_ffi_items`, but if some other test for a different struct returns a `*const T1Bar`, then it would fail to realize that `T1Bar` is a struct, since originally we were using the `filtered_ffi_items` in `c_type`.

ctest-next/src/template.rs
```diff
@@ -495,11 +497,52 @@ pub(crate) struct TranslateHelper<'a> {
...
+                            .iter()
+                            .any(|f| f(&MapInput::$variant(item)))
+                            || (self.generator.skip_private && !item.public)
+                    })
+                    .collect();
+                if verbose {
+                    skipped
+                        .iter()
+                        .for_each(|item| eprintln!("Skipping {} \"{}\"", $label, item.ident()));
+                }
```
tgross35: No need to collect to an intermediate `Vec`:

```rust
let skipped = self
    .filtered_ffi_items
    // ...
    .extract_if(/* ... */);
for item in skipped {
    if verbose {
        eprintln!("...");
    }
}
```

Also, put the `self.generator.skip_private && !item.public` check before checking `self.generator.skips`, so it doesn't need to test all the skips if there is an easy return.

ctest-next/src/template.rs
```diff
@@ -520,9 +563,12 @@ impl<'a> TranslateHelper<'a> {
...
             // For structs/unions/aliases, their type is the same as their identifier.
-            MapInput::Alias(a) => (a.ident(), cdecl::named(a.ident(), cdecl::Constness::Mut)),
-            MapInput::Struct(s) => (s.ident(), cdecl::named(s.ident(), cdecl::Constness::Mut)),
-            MapInput::Union(u) => (u.ident(), cdecl::named(u.ident(), cdecl::Constness::Mut)),
+            // FIXME(ctest): For some specific primitives such as c_uint, they don't exist on the
+            // C side and have to be manually translated. If they are removed to use `std::ffi`,
+            // then this becomes unneeded (although it won't break).
+            MapInput::Alias(a) => (a.ident(), cdecl::named(a.ident(), Constness::Mut)),
+            MapInput::Struct(s) => (s.ident(), cdecl::named(s.ident(), Constness::Mut)),
+            MapInput::Union(u) => (u.ident(), cdecl::named(u.ident(), Constness::Mut)),
```
tgross35: What is the FIXME here? This part doesn't have to do with primitives.

ctest-next/src/translator.rs
```diff
@@ -363,6 +363,16 @@ pub(crate) fn ptr_with_inner(
...
+        // This is done to deal with things like 3usize.
+        syn::Expr::Index(i) => {
+            let base = translate_expr(&i.expr);
+            let index = translate_expr(&i.index);
+            format!("{base}[{index}]")
+        }
+        syn::Expr::Lit(l) => match &l.lit {
+            syn::Lit::Int(i) => i.base10_digits().to_string(),
+            _ => l.to_token_stream().to_string(),
+        },
```
tgross35: Is that comment above the right block? For the integer parts, you can turn e.g. `3usize` into `((size_t)(3))` to keep the type around.

Could you make sure this is tested in ctest-next/src/tests.rs?

PR #4615, opened 2025-08-04
Split qnx target based on target env

AkhilTThomas:
- Created nto7x and nto8x modules.
- Merged neutrino.rs into parent module.
- nto7x/mod.rs is identical to the original mod.rs with only difference being the crate import adjustments.
- nto8x/mod.rs is same as nto7x/mod.rs (placeholder till the actual 8.0 support PR)

# Description

This PR is a preparation to implement the QNX 8.0 support by splitting the implementation logic into separate modules.
It's similar to the approach used by `freebsd` to support multiple versions which have a different "C" function / structure / constants.

 - No functional changes are added to the targets `nto70` , `nto71` and `nto71_iosock`.
 - The target `nto80` is currently identical to the `nto7071` implementation to keep the files diff-able.

# Sources

# Checklist


AkhilTThomas:
Pinging @jonathanpallant @flba-eb Could you take a look at this?

jonathanpallant (commented):
LGTM

src/unix/nto/nto7x/mod.rs
```diff
@@ -0,0 +1,3401 @@
...
+
+cfg_if! {
+    if #[cfg(target_arch = "x86_64")] {
+        pub use crate::unix::nto::x86_64::*;
+    } else if #[cfg(target_arch = "aarch64")] {
+        use crate::unix::nto::aarch64::*;
+    } else {
+        panic!("Unsupported arch");
+    }
+}
```
tgross35, line 11: Why is the first branch `pub use` but the second is `use`?

Also, typically `libc` just exports nothing if the architecture isn't supported, rather than having a compiler error. If this is kept, it should be `compiler_error!` rather than `panic!` , and it would be good to add "for QNX Neutrino" to the message (since presumably we support the architecture otherwise).

src/unix/nto/nto8x/mod.rs
```diff
@@ -0,0 +1,3401 @@
...
+
+cfg_if! {
+    if #[cfg(target_arch = "x86_64")] {
+        use crate::unix::nto::x86_64::*;
+    } else if #[cfg(target_arch = "aarch64")] {
+        use crate::unix::nto::aarch64::*;
+    } else {
+        panic!("Unsupported arch");
+    }
+}
```
tgross35, line 11: Same questions as for nto7x

tgross35 (changes requested):
> * Created nto7x and nto8x modules.

This part seems reasonable

> * Merged neutrino.rs into parent module.

Does `neutrino.rs` map to a `neutrino.h`? If so, I would prefer to keep this separate since we are migrating other targets to something similar, i.e. mapping one `.h` file to one `.rs` file rather than keeping everything together. If there are version-specific aspects, these parts can go in `{nto7x,nto8x}/neutrino.h` 

> * nto7x/mod.rs is identical to the original mod.rs with only difference being the crate import adjustments.
> * nto8x/mod.rs is same as nto7x/mod.rs (placeholder till the actual 8.0 support PR)

I don't understand the reasoning behind this part of the change: typically we keep code that is common across all versions in a parent module, and then only parts that are different move into version-specific modules. For FreeBSD the version-specific files are only a few hundred lines. Having an identical or nearly-identical 3k line file here isn't very nice.

AkhilTThomas:
@tgross35 Thank you for the review! and sorry about the delayed response.

> Does `neutrino.rs` map to a `neutrino.h`? If so, I would prefer to keep this separate since we are migrating other targets to something similar, i.e. mapping one `.h` file to one `.rs` file rather than keeping everything together. If there are version-specific aspects, these parts can go in `{nto7x,nto8x}/neutrino.h`

Yes, `neutrino.rs` maps to `neutrino.h` , i merged it to have only mod.rs and arch specific files. If the idea moving forward is to have one to one mapping with the original header i can move it back.

> > * nto7x/mod.rs is identical to the original mod.rs with only difference being the crate import adjustments.
> > * nto8x/mod.rs is same as nto7x/mod.rs (placeholder till the actual 8.0 support PR)
> 
> I don't understand the reasoning behind this part of the change: typically we keep code that is common across all versions in a parent module, and then only parts that are different move into version-specific modules. For FreeBSD the version-specific files are only a few hundred lines. Having an identical or nearly-identical 3k line file here isn't very nice.

This was done to keep the diff minimal with the existing `nto7071` modules and to make review easier.

src/unix/nto/nto7x/mod.rs
AkhilTThomas: This was an over sight, i will fix it.

tgross35:
This has been superseded by https://github.com/rust-lang/libc/pull/4984

Issue #4611, opened 2025-08-02
libresolv


lorenzleutgeb:
`resolv.h` exposes facilities to resolve DNS records. See:

 - [Google Bionic](https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/include/resolv.h)
 - [GNU C Library](https://sourceware.org/git/?p=glibc.git;a=blob;f=include/resolv.h;h=4dbbac3800b7ef302c8c93eb1cb3855c5a1c3d83;hb=d2097651cc57834dbfcaa102ddfacae0d86cfb66) (I admit that I am slightly confused why it is not mentioned in the [GNU C Library Manual](https://sourceware.org/glibc/manual/2.42/).)
 - [musl](https://git.musl-libc.org/cgit/musl/tree/include/resolv.h)
 - [FreeBSD](https://cgit.freebsd.org/src/tree/lib/libc/resolv)

I would like to understand whether there is a particular reason why the symbols in `resolv.h` are not covered by the `libc` crate. If there is no particular reason, I would like to start working on inclusion. My original motivation is to resolve DNS records (of type SRV, TXT) and realized that the Rust standard library does not support this. My next natural step was to check the `libc` crate.

Target Triples: `*-unknown-linux-gnu`, `*-unknown-android`, `*-unknown-freebsd` (`*` to be narrowed down by testing).

References:
 - Zig seems to support it: https://github.com/ziglang/zig/issues/12628
 - There's a [`resolv`](https://crates.io/crates/resolv) crate, which wraps [`libresolv-sys`](https://crates.io/crates/libresolv-sys). It might benefit from this API.

PR #4618, opened 2025-08-05
Add resolv

lorenzleutgeb:
# Description

See rust-lang/libc#4611 

# Sources

# Checklist

### DIFF
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -1262,6 +1262,11 @@ pub const PIPE_BUF: usize = 4096;
 
 pub const SI_LOAD_SHIFT: c_uint = 16;
 
+// arpa/nameser.h
+pub const T_A: c_int = 1;
+pub const T_CNAME: c_int = 5;
+pub const C_IN: c_int = 1;
+
 // si_code values
 pub const SI_USER: c_int = 0;
 pub const SI_KERNEL: c_int = 0x80;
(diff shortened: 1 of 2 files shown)


devnexen:
several things at quick glance :
- res_init is already defined for supported platforms (see src/unix/mod.rs)

src/unix/linux_like/mod.rs
```diff
@@ -2062,6 +2067,45 @@ extern "C" {
 
     pub fn getdomainname(name: *mut c_char, len: size_t) -> c_int;
     pub fn setdomainname(name: *const c_char, len: size_t) -> c_int;
+
+    // <resolv.h>
+    #[allow(dead_code)]
```
tgross35: This shouldn't be needed

src/unix/linux_like/mod.rs
```diff
@@ -2062,6 +2067,45 @@ extern "C" {
...
+        op: c_int,
+        dname: *const c_char,
+        class: c_int,
+        kind: c_int,
+        data: *mut c_char,
+        newrr: *const c_void,
+        datalen: c_int,
+        buf: *mut c_char,
+        buflen: c_int,
+    ) -> c_int;
```
tgross35, line 2143: This doesn't seem to match the definitions linked at https://github.com/rust-lang/libc/issues/4611: those have `data` as `const` and followed by an integer

tgross35 (changes requested):
What exactly is the linking story here, by the way? Are the symbols always distributed with the libc binary?

src/unix/linux_like/mod.rs
devnexen: also few arguments misplacements see above my comment and on linux it should be linked to libresolv.

tgross35:
@lorenzleutgeb ping on this, it's been inactive for a little while

lorenzleutgeb:
@tgross35 thanks for the ping. I applied the fixes in https://github.com/rust-lang/libc/pull/4618#issuecomment-3162655659. To be honest I do not understand which change "on linux it should be linked to libresolv" (https://github.com/rust-lang/libc/pull/4618/changes#r2265716911) implies. I am willing to learn!

PR #4619, opened 2025-08-06
Export UDP_SEGMENT and UDP_GRO consts on Android

hulthe:
# Description

Export constants for the UDP-specific socket options `UDP_SEGMENT` and `UDP_GRO` on Android.
These are already exposed on Linux proper [here](https://github.com/rust-lang/libc/blob/7f3d69697632b79e8a4e1eecd3705e5ab048f66c/src/unix/linux_like/linux/mod.rs#L2866).

# Sources

[linux/include/uapi/linux/udp.h](https://github.com/torvalds/linux/blob/cca7a0aae8958c9b1cd14116cb8b2f22ace2205e/include/uapi/linux/udp.h#L30)

# Checklist


maurer:
The two you've added are fine. If you want to add more, please restrict yourself to those in https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/kernel/uapi/linux/udp.h - mostly that means no `UDP_ENCAP_OVPNINUDP`.

(This is `libc`, so it's `bionic` bindings, not Linux kernel bindings. When a constant is supported in Android it'll make its way into the bionic uapi generated headers.)

PR #4620, opened 2025-08-07
Add `sigqueue`

yshui:
# Description

Add prototype for `sigqueue`

# Sources

# Checklist


src/unix/mod.rs
```diff
@@ -43,6 +43,27 @@ missing! {
 }
 pub type locale_t = *mut c_void;
 
+s_no_extra_traits! {
+    pub union sigval {
+        pub sival_ptr: *mut c_void,
+        pub sival_int: c_int,
+    }
```
tgross35: If this change is done here, it can't be released with 0.2.x. Could you put this part in a separate PR from adding `sigqueue`?

yshui:
kinda in a weird situation:

android [has it](https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/include/signal.h#173).
illumos seems to [have it](https://github.com/illumos/illumos-gate/blob/5e96687e18c05c76765c7d6920cd0b6505f2d587/usr/src/uts/common/syscall/sigqueue.c#L115). (can't find the header, though)
cygwin [has it](https://github.com/mirror/newlib-cygwin/blob/1b7c72fdcc4bde7520407d2d3364146f04fb8312/winsup/cygwin/include/cygwin/signal.h#L415).
redox [has it, IIUC](https://gitlab.redox-os.org/redox-os/relibc/-/blob/master/src/platform/redox/signal.rs#L73)
qnx neutrino [has it](https://www.qnx.com/developers/docs/7.1/#com.qnx.doc.neutrino.lib_ref/topic/s/sigqueue.html)
aix [has it](https://www.ibm.com/docs/hu/aix/7.3.0?topic=s-sigqueue-subroutine)
haiku [has it](https://github.com/haiku/haiku/blob/4df505d911132b51a6d8cce6062799f6cf537f6c/src/system/libroot/posix/signal/sigqueue.cpp#L19)

i think openbsd and hurd are the ones that don't have it.

yshui:
and sigqueue is POSIX.1-2008, not a os-specific extensions

libc-test/semver/redox.txt
```diff
@@ -234,6 +234,7 @@ qsort
 reallocarray
 setgrent
 setpwent
+sigqueue
 setrlimit
 setservent
 sigtimedwait
```
tgross35: sigqueue has to go after setservent here

After this, lgtm

yshui: :man_facepalming:

PR #4632, opened 2025-08-11
libc: add padding struct

mbyx:
# Description
Adds a struct for padding types.

# Sources

# Checklist

### DIFF
--- a/libc-test/test/check_style.rs
+++ b/libc-test/test/check_style.rs
@@ -20,7 +20,7 @@ use style::{Result, StyleChecker};
 const SKIP_PREFIXES: &[&str] = &[
     // Don't run the style checker on the reorganized portion of the crate while we figure
     // out what style we want.
-    "new/",
+    "new/", "types.rs",
 ];
 
 #[test]
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,0 +1,18 @@
+//! Platform-agnostic support types.
+
+use core::mem::MaybeUninit;
+
+use crate::prelude::*;
+
+/// A transparent wrapper over `MaybeUninit<T>` to represent uninitialized padding
+/// while providing `Default`.
+#[allow(unused)]
+#[repr(transparent)]
+#[derive(Clone, Copy)]
+pub(crate) struct Padding<T: Copy>(MaybeUninit<T>);
+
+impl<T: Copy> Default for Padding<T> {
+    fn default() -> Self {
+        Self(MaybeUninit::zeroed())
+    }
+}
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -64,6 +64,8 @@ macro_rules! cfg_if {
 /// Create an internal crate prelude with `core` reexports and common types.
 macro_rules! prelude {
     () => {
+        mod types;
+
         /// Frequently-used types that are available on all platforms
         ///
         /// We need to reexport the core types so this works with `rust-dep-of-std`.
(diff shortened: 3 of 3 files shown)


src/types.rs
```diff
@@ -0,0 +1,13 @@
+use core::mem::MaybeUninit;
+
+/// A transparent wrapper over `MaybeUninit<T>` to represent uninitialized padding
+/// while providing `Default`.
+#[repr(transparent)]
+#[derive(Clone, Copy)]
+pub struct Padding<T: Copy>(MaybeUninit<T>);
```
tgross35: I don't think we need to export this, `pub(crate)` should be fine

src/types.rs
```diff
@@ -0,0 +1,13 @@
+use core::mem::MaybeUninit;
```
tgross35, line 3:
```suggestion
//! Platform-agnostic support types.

use core::mem::MaybeUninit;
```

src/lib.rs
```diff
@@ -43,6 +43,9 @@ pub use core::ffi::c_void;
 #[allow(unused_imports)] // needed while the module is empty on some platforms
 pub use new::*;
 
+mod types;
+pub use types::*;
```
tgross35: There shouldn't be anything to export

mbyx:
It seems like ctest still fails with just these changes.

src/types.rs
```diff
@@ -0,0 +1,16 @@
+//! Platform-agnostic support types.
+
+use core::mem::MaybeUninit;
```
tgross35, line 3: This will need to explicitly import `core::default::Default`, possibly some other tweaking around that (we don't get the `core` prelude when built as a dep of `std`)

src/types.rs
```diff
@@ -0,0 +1,20 @@
+//! Platform-agnostic support types.
+
+use core::clone::Clone;
+use core::default::Default;
+use core::marker::Copy;
+use core::mem::MaybeUninit;
+use core::prelude::rust_2024::derive;
```
tgross35: If you need to import all of these, just add `Default` to the prelude macro and add `use crate::prelude::*` here.

This adds a bit of trickiness: there is no `crate::prelude` in some cases https://github.com/rust-lang/libc/blob/c0071cc9f9385da6352e40824cfd77334e85f72d/src/lib.rs#L143-L145, so the import would fail in this module. To avoid that problem, move `mod types` into the prelude as well.

libc-test/test/check_style.rs
```diff
@@ -20,7 +20,7 @@ use style::{Result, StyleChecker};
 const SKIP_PREFIXES: &[&str] = &[
     // Don't run the style checker on the reorganized portion of the crate while we figure
     // out what style we want.
-    "new/",
+    "new/", "types.rs",
```
tgross35, line 23:
```suggestion
    "new/",
    "types.rs",
```

(rustfmt doesn't run on this directory, I need to fix that eventually)

mbyx: I ran `rustfmt` locally and the original is what it outputs, should I still change it?

tgross35: Huh, that is just surprising output. Don't worry about it then :)

src/types.rs
```diff
@@ -0,0 +1,19 @@
+//! Platform-agnostic support types.
+
+use core::mem::MaybeUninit;
+use core::prelude::rust_2024::derive;
```
tgross35: _Sigh_ I was thinking this was already in the prelude, sorry I keep steering you wrong here.

Add this to `prelude!` instead, that's consistent. Also it will need to be `::core::prelude::v1::derive` like https://github.com/rust-lang/libc/blob/c0071cc9f9385da6352e40824cfd77334e85f72d/src/macros.rs#L117C15-L117C43, the 2024 prelude isn't available until way after our MSRV.

PR #4635, opened 2025-08-11
triagebot.toml: Fix src/new/linux_uapi path

neuschaefer:
Commit 3356f1217 ("Begin source reorganization with `linux/can.h`") introduced a new directory structure for reorganizing the source code. It is called `src/new/` everywhere but in triagebot.toml, likely because `src/reorg/` is an older name used during development.

cc @tgross35


neuschaefer:
@tgross35 is there a design document for the new structure? At a cursory glance I didn't find one

PR #4640, opened 2025-08-11
Fix CI on FreeBSD 13/14

neuschaefer:
# Description

This is an attempt to get the CI green again on FreeBSD.

FreeBSD 13 i686 unfortunately still fails a test:

```
failures:
    test_skip_simple
```

I haven't investigated the cause, but it's 32-bit specific.

# Sources

P_IDLEPROC:
- introduced in FreeBSD 15: commit 33be1632047c ("racct: Fix accounting of CPU time for the system idle process")

mcontext::mc_tlsbase:
- introduced in FreeBSD 15: commit eea3e4dd9703 ("amd64: add mc_tlsbase member to mcontext")

Links are in the commit messages

# Checklist

- [x] ~~Relevant tests in `libc-test/semver` have been updated~~ none needed, I think

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/x86_64/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/x86_64/mod.rs
@@ -98,6 +98,7 @@ s_no_extra_traits! {
     }
 
     #[repr(align(16))]
+    #[non_exhaustive]
     pub struct mcontext_t {
         pub mc_onstack: register_t,
         pub mc_rdi: register_t,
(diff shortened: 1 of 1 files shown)


src/unix/bsd/freebsdlike/freebsd/x86_64/mod.rs
tgross35, line 1: While you're adjusting the fields here, would you mind marking the struct `#[non_exhaustive]`?

neuschaefer: ah yes, good point

neuschaefer: done

PR #4644, opened 2025-08-12
ctest: fix bug where functions and statics aren't skipped

mbyx:
# Description
Fixed `extern fn`s and `extern static`s not being skipped properly. Also added a `rename_alias` method that should be preferred to be used over `rename_type`.

# Sources

# Checklist


mbyx:
I don't think `rename_type` can be deleted. `rename_alias` would only work for `MapInput` items that are passed as `MapInput::Alias`, so it only affects the alias tests. But if some other test uses for example `*const type_alias` (or even just `field: type_alias`), then that is regarded as a `MapInput::Type`, so `rename_alias` wouldn't work.

Definitely something to be improved, since that part of the codebase is definitely the most buggy.

PR #4647, opened 2025-08-13
libc-test: Switch Linux to ctest-next

mbyx:
# Description
Ports the linux platform to also be tested using `ctest-next`.

~Blocked on rust-lang/libc#4674.~

For now it skips tests for `pid_type` and `proc_cm_event` C enums (technically they weren't tested before either) because C seems to hide them in userspace.

# Sources

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -3548,7 +3548,7 @@ fn test_vxworks(target: &str) {
     ctest::generate_test(&mut cfg, "../src/lib.rs", "ctest_output.rs").unwrap();
 }
 
-fn config_gnu_bits(target: &str, cfg: &mut ctest_old::TestGenerator) {
+fn config_gnu_bits(target: &str, cfg: &mut ctest::TestGenerator) {
     let pointer_width = env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap_or_default();
     if target.contains("gnu")
         && target.contains("linux")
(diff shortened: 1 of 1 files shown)


mbyx:
So the temporary fix to the enum problem mostly works, but it does still seem like `pid_type` and its constants are not being found. I'd like to wait a bit longer on this port, and actually figure out why the original ctest does not have to deal with these enums.

mbyx:
I tested in a completely new crate with the old ctest cloned beside it, it seems that it can't detect an item wrapped inside of the `c_enum` macro, effectively hiding it from ctest.

Then I started doing some print debugging in the old ctest, and from what I can understand, given a crate that only has the `c_enum` macro and an invocation of that macro, after expanding the crate with
```rs
let krate = ecx.monotonic_expander().expand_crate(krate);
```
it seems to just leave the macro definition behind, deleting the invocation in its internal representation.

Then I switched to a crate with only the `s` and `__item` macros and an invocation of `s`, but mysteriously that worked without issue. Same with the `e` and `__item` macros.

At first I thought it was because of a lack of using `__item` in `c_enum`, so I modified `c_enum` to use it, but even that didn't work.

So I think that all this time ctest has just quietly been ignoring `c_enum` invocations. (Which would be easy to check, just run `libc-test` on linux and look at the generated files, I can't because the original ctest fails on WSL for me).

libc-test/build.rs
```diff
@@ -3642,15 +3633,16 @@ fn test_linux(target: &str) {
...
+    let mut cfg = ctest_next_cfg();
     if musl_v1_2_3 {
         cfg.cfg("musl_v1_2_3", None);
     }
-    cfg.define("_GNU_SOURCE", None);
-    // This macro re-defines fscanf,scanf,sscanf to link to the symbols that are
-    // deprecated since glibc >= 2.29. This allows Rust binaries to link against
-    // glibc versions older than 2.29.
-    cfg.define("__GLIBC_USE_DEPRECATED_SCANF", None);
+    cfg.skip_private(true)
```
tgross35: This is done in the global config now, right?

mbyx:
If `pid_type` can't be tested, then its best to block this on #4658, so that skipping `pid_type` is just a one liner.

mbyx:
Just one error left. It seems to be related to 64 and 32 bit time. The test fails on a 32 bit platform with 64 bit time, so it wants to use `__suseconds_t` but can't find it. @tgross35, would this be related to the multiple time related PRs or is it just a `cfg` issue?

PR #4648, opened 2025-08-13
libc: port freebsd to use ctest-next

mbyx:
# Description

# Sources

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -78,7 +78,9 @@ fn ctest_cfg() -> ctest::TestGenerator {
 }
 
 fn ctest_next_cfg() -> ctest_next::TestGenerator {
-    ctest_next::TestGenerator::new()
+    let mut cfg = ctest_next::TestGenerator::new();
+    cfg.skip_private(true);
+    cfg
 }
 
 fn do_semver() {
@@ -2286,7 +2288,7 @@ fn test_android(target: &str) {
 
 fn test_freebsd(target: &str) {
     assert!(target.contains("freebsd"));
-    let mut cfg = ctest_cfg();
+    let mut cfg = ctest_next_cfg();
 
     let freebsd_ver = which_freebsd();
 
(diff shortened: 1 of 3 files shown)


mbyx:
All checks passed... except that the freebsd checks didn't even run. @tgross35 is there a way to only run those checks and not the others just to make sure there isn't any minor bug?

libc-test/build.rs
```diff
@@ -2443,24 +2443,19 @@ fn test_freebsd(target: &str) {
             | "devstat_support_flags"
             | "devstat_type_flags"
             | "devstat_match_flags"
-            | "devstat_priority" => ty.to_string(),
+            | "devstat_priority" => ty.to_string().into(),
 
             // FIXME(freebsd): https://github.com/rust-lang/libc/issues/1273
-            "sighandler_t" => "sig_t".to_string(),
+            "sighandler_t" => "sig_t".to_string().into(),
```
tgross35: Wrap in `Some` rather than using `.into()` for clarity

libc-test/build.rs
```diff
@@ -2864,18 +2861,12 @@ fn test_freebsd(target: &str) {
...
-        match i {
-            // aio_buf is a volatile void* but since we cannot express that in
-            // Rust types, we have to explicitly tell the checker about it here:
-            StructField(ref n, ref f) if n == "aiocb" && f == "aio_buf" => true,
-            _ => false,
-        }
-    });
+    // aio_buf is a volatile void* but since we cannot express that in
+    // Rust types, we have to explicitly tell the checker about it here:
+    cfg.volatile_struct_field(|s, f| s.ident() == "aiocb" && f.ident() == "aio_buf");
```
tgross35, line 2872: Small nit: it's not exactly canonical Rust, but keeping it `match` does make it easier to add/remove items. No strong need to update this one, though.

(Sorry if I said this in another PR already, can't remember if I did)

mbyx:
Almost there. It seems only `freebsd-13 i686` is failing, which I believe is the 32bit version? The others are 64 bit and aren't failing, and the error seems to be about missing a `reg32` struct.

mbyx:
It feels strange to finally see a successful freebsd run.

Could the difference in naming be a bug that slipped by when freebsd mostly failed? It doesn't look like `reg32` or the like is ever skipped.

neuschaefer:
FWIW, this breaks libc-test on freebsd15/ppc64le, which was previously working. `ctest` segfaults without producing useful output.

PR #4653, opened 2025-08-14
libc: port openbsd to ctest-next

semarie:


semarie:
freebsd failures seems unrelated.

the PR restore CI on OpenBSD (when ran natively)

PR #4654, opened 2025-08-14
linux: Use the input_event_X macros to access input_event members

snogge:
# Description

Use the `input_event_sec` and `input_event_usec` members (macros in C) on all platforms, no matter whether `time_t` is 32 or 64 bits. 
This means the same members are available in the struct no matter whether it is a 32 or 64 bit platform or the linux_time_bits64 config is set.

This is a breaking change.

# Sources

https://github.com/torvalds/linux/blob/0cc53520e68bea7fb80fdc6bdf8d226d1b6a98d9/include/uapi/linux/input.h#L28

# Checklist

This is a breaking change and should not be backported to the 0.2 branch.

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -332,21 +332,18 @@ s! {
     }
 
     pub struct input_event {
-        // FIXME(1.0): Change to the commented variant, see https://github.com/rust-lang/libc/pull/4148#discussion_r1857511742
         #[cfg(any(target_pointer_width = "64", not(linux_time_bits64)))]
-        pub time: crate::timeval,
-        // #[cfg(any(target_pointer_width = "64", not(linux_time_bits64)))]
-        // pub input_event_sec: time_t,
-        // #[cfg(any(target_pointer_width = "64", not(linux_time_bits64)))]
-        // pub input_event_usec: suseconds_t,
-        // #[cfg(target_arch = "sparc64")]
-        // _pad1: c_int,
+        pub input_event_sec: crate::time_t,
         #[cfg(all(target_pointer_width = "32", linux_time_bits64))]
         pub input_event_sec: c_ulong,
 
+        #[cfg(any(target_pointer_width = "64", not(linux_time_bits64)))]
+        pub input_event_usec: crate::suseconds_t,
         #[cfg(all(target_pointer_width = "32", linux_time_bits64))]
         pub input_event_usec: c_ulong,
 
+        #[cfg(target_arch = "sparc64")]
+        _pad1: c_int,
         pub type_: __u16,
         pub code: __u16,
         pub value: __s32,


src/unix/linux_like/linux/mod.rs
```diff
@@ -332,21 +332,18 @@ s! {
...
-        pub time: crate::timeval,
-        // #[cfg(any(target_pointer_width = "64", not(linux_time_bits64)))]
-        // pub input_event_sec: time_t,
-        // #[cfg(any(target_pointer_width = "64", not(linux_time_bits64)))]
-        // pub input_event_usec: suseconds_t,
-        // #[cfg(target_arch = "sparc64")]
-        // _pad1: c_int,
+        pub input_event_sec: crate::time_t,
         #[cfg(all(target_pointer_width = "32", linux_time_bits64))]
         pub input_event_sec: c_ulong,
```
tgross35, line 234: Could you add a comment that these are actually combined  as `timeval`?

src/unix/linux_like/linux/mod.rs
```diff
@@ -332,21 +332,18 @@ s! {
...
 
+        #[cfg(any(target_pointer_width = "64", not(linux_time_bits64)))]
+        pub input_event_usec: crate::suseconds_t,
         #[cfg(all(target_pointer_width = "32", linux_time_bits64))]
         pub input_event_usec: c_ulong,
 
+        #[cfg(target_arch = "sparc64")]
+        _pad1: c_int,
```
tgross35: I would like to backport this part of the change. It is technically breaking but our definition is incorrect without it, and sparc64 isn't all that popular of a target. Would you be able to split this to a separate commit?

Padding fields should also get wrapped in `Padding<...>` now

tgross35: Hm - actually, I think this change may not be correct on its own. Looks like the `__usec`/`input_event_usec` field would also need to be changed from a long to an int https://github.com/torvalds/linux/blob/3f9f0252130e7dd60d41be0802bf58f6471c691d/include/uapi/linux/input.h#L35C6-L42.

We should probably do that change still to make sure users don't expect the upper half of that field to work, just annoying that sparc has a special-cased `suseconds_t` in the first place.

(context: https://github.com/torvalds/linux/commit/2e746942ebacf1565caa72cf980745e5ce297c48)

tgross35: Nevermind, I see now that the field is of type `suseconds_t`. Sorry for the confusion; please disregard everything in this thread except for the `Padding<...>` bit.

tgross35:
The musl time64 changes caused a conflict here so this will need a rebase (sorry about that)

> @tgross35 The change causes CI errors for `i686-linux-unknown-musl`. I was not sure what the right thing to do was - I have not been keeping up to date on changes - so I added another time64 exception for musl.

I _think_ it should be fixed by https://github.com/rust-lang/libc/pull/4463 which resolved everything else in that block (hence the conflict), musl on relevant platforms should be setting the linux time64 cfg now.

src/unix/linux_like/linux/mod.rs
snogge: done

tgross35 (approved):

PR #4655, opened 2025-08-14
Libc promote ctest next

mbyx:
# Description
Promotes `ctest-next` to be the new `ctest`. Platforms not yet ported use the older version of `ctest` from crates.io as `ctest-old`. Testing for `ctest-old` has been removed from `ctest-test`.

# Sources

# Checklist


.github/workflows/ci.yaml
```diff
@@ -320,7 +320,7 @@ jobs:
     - name: Install Rust
       run: rustup update "$MSRV" --no-self-update && rustup default "$MSRV"
     - name: Remove edition 2024 crates
-      run: perl -i -ne 'print unless /"ctest-(next|test)",/ || /"libc-test",/' Cargo.toml
+      run: perl -i -ne 'print unless /"ctest(?:-test)?",/ || /"libc-test",/' Cargo.toml
```
tgross35: Delete this bit now since the ctest-next msrv is the ctest msrv.

While you're here, could you rename this job from "Check MSRV" to "Check ctest MSRV"?

mbyx: The version of cargo the CI uses doesn't seem to support 2024 edition. EDIT: It seems that the command used for MSRV picks up the old ctest and not the new one since they have the same name.

.github/workflows/ci.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
     name: Verify build
     strategy:
       matrix:
-        toolchain: [stable, nightly, 1.63.0]
+        toolchain: [stable, nightly, 1.88.0]
```
tgross35: No don't change this; this is the MSRV for all of `libc`

.github/workflows/ci.yaml
```diff
@@ -88,8 +88,6 @@ jobs:
           if [ "${{ matrix.toolchain }}" = "1.63.0" ]; then
               # Remove `-Dwarnings` at the MSRV since lints may be different
               export RUSTFLAGS=""
-              # Remove `ctest-next` which uses the 2024 edition
-              perl -i -ne 'print unless /"ctest-(next|test)",/ || /"libc-test",/' Cargo.toml
```
tgross35, line 92: This needs to remain but be updated to `"ctest(-test)?"

.github/workflows/ci.yaml
```diff
@@ -306,21 +304,19 @@ jobs:
...
+    name: Check ctest MSRV
     runs-on: ubuntu-24.04
     timeout-minutes: 10
     env:
       RUSTFLAGS: # No need to check warnings on old MSRV, unset `-Dwarnings`
     steps:
     - uses: actions/checkout@master
     - run: |
-        msrv="$(cargo metadata --format-version 1 | jq -r --arg CRATE_NAME ctest  '.packages | map(select(.name == $CRATE_NAME)) | first | .rust_version')"
+        msrv="$(cargo metadata --format-version 1 | jq -r --arg CRATE_NAME ctest  '.packages | map(select(.name == $CRATE_NAME)) | last | .rust_version')"
```
tgross35: The ordering of the packages isn't guaranteed, so this may not always work. You can grab only the local crate by changing the `map(select(...))` query to add `and (.id | startswith("path+file"))`

If you could rewrap this a bit too that would be good, the line is pretty long already. (Bash should let you split lines anywhere within the `$(...)`)

ctest/Cargo.toml
```diff
@@ -1,24 +1,20 @@
...
-description = "Automated tests of FFI bindings."
-exclude = ["CHANGELOG.md"]
-edition = "2021"
+version = "0.5.0-beta.0"
+edition = "2024"
+rust-version = "1.88"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/rust-lang/libc"
-rust-version = "1.63.0"
+publish = false
```
tgross35, line 8: Ah autopublish failed, could you put up a PR to set this `true`? I'll get it later today if you're not active right now

PR #4657, opened 2025-08-15
procctl for freebsd 15 upgrade.

devnexen:
adding missing PROC_SPROTECT helpers while at it.

ref: https://github.com/freebsd/freebsd-src/commit/dabf006a638fdc44cdcf69731de8ac83959db731#diff-769c07b1f90eef1351a20f40593b254f85925859273219036320baa1d086774cR68


neuschaefer:
Could you please rebase on the latest commit of the main branch? This will resolve some unrelated FreeBSD CI failures and thus make the CI results clearer.


> adding missing PROC_SPROTECT helpers while at it.
> 
> ref: https://github.com/freebsd/freebsd-src/blob/c992ac6213276f54d868f317cc5092f8aed4ff54/sys/sys/procctl.h#L68

c992ac6213276f54d868f317cc5092f8aed4ff54 seems to be a rather arbitrary recent commit. I'd suggest linking dabf006a638fdc44cdcf69731de8ac83959db731 instead, which is the commit that introduced the new constants.
Perhaps with this link: https://github.com/freebsd/freebsd-src/commit/dabf006a638fdc44cdcf69731de8ac83959db731#diff-769c07b1f90eef1351a20f40593b254f85925859273219036320baa1d086774cR68

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -3012,6 +3012,11 @@ pub const PROC_NO_NEW_PRIVS_STATUS: c_int = 20;
 pub const PROC_WXMAP_CTL: c_int = 21;
 pub const PROC_WXMAP_STATUS: c_int = 22;
 pub const PROC_PROCCTL_MD_MIN: c_int = 0x10000000;
```
neuschaefer: I think it's best to keep `PROC_PROCCTL_MD_MIN` at the end, it doesn't really have a place between `PROC_WXMAP_STATUS` and `PROC_LOGSIGEXIT_CTL` in the original headers

src/unix/bsd/freebsdlike/freebsd/mod.rs
```diff
@@ -4788,6 +4793,14 @@ safe_f! {
     pub {const} fn PR_SCTP_VALID_POLICY(x: c_int) -> bool {
         PR_SCTP_POLICY(x) <= SCTP_PR_SCTP_MAX
     }
+
+    pub {const} fn PPROT_OP(o: c_int) -> c_int {
```
neuschaefer: I _personally_ would prefer the PPROT changes in a separate commit (same PR is ok), so they can get a more focused commit message (perhaps with a reference link as well)

neuschaefer (changes requested):
Hi, thanks for your contribution! It looks good (to my eyes at least) for the most part

src/unix/bsd/freebsdlike/freebsd/mod.rs
devnexen: but ultimately I will have to squash it so..

src/unix/bsd/freebsdlike/freebsd/mod.rs
devnexen: sure

src/unix/bsd/freebsdlike/freebsd/mod.rs
tgross35: Just fyi, >1 commit per PR is fine if they're meaningful (the merge strategy here is rebase). We just don't want the progress commits like "fix abc" "update" etc.

libc-test/build.rs
```diff
@@ -2743,6 +2743,17 @@ fn test_freebsd(target: &str) {
...
+            // Added in FreeBSD 15
+            "PROC_LOGSIGEXIT_CTL"
+            | "PROC_LOGSIGEXIT_STATUS"
+            | "PROC_LOGSIGEXIT_CTL_NOFORCE"
+            | "PROC_LOGSIGEXIT_CTL_FORCE_ENABLE"
+            | "PROC_LOGSIGEXIT_CTL_FORCE_DISABLE"
+                if Some(15) > freebsd_ver =>
+            {
+                true
+            }
```
tgross35, line 2782: Why are these added to FreeBSD < 15 in the first place? I think they could just live in `src/unix/bsd/freebsdlike/freebsd/freebsd15`

devnexen: I think what you refer to (ie freebsd < version >) is the place where data types change across releases.

tgross35: Not sure I understand - `freebsd15/mod.rs` also includes constants.

devnexen: cc @asomers

asomers: If a function or constant _changes_ between releases, then we need to put it in the appropriately versioned file.  That's so users don't, for example, call the FreeBSD-11-compat version of `fstatat` with the FreeBSD 12+ `struct stat`.  That would be bad.  But if a function, struct, or constant newly appears in some new FreeBSD release, then it's fine to put it into the main freebsd/mod.rs file.  If a user tries to use it while running a too-old version of FreeBSD, they'll simply get an `ENOSYS` error, which isn't too bad.

libc-test/semver/freebsd.txt
```diff
@@ -1047,6 +1047,12 @@ POSIX_SPAWN_SETSCHEDULER
 POSIX_SPAWN_SETSIGDEF
 POSIX_SPAWN_SETSIGMASK
 PPPDISC
+PPROT_CLEAR
```
asomers, line 1051: Why is the list of symbols you're adding here different than what you're adding in freebsd/mod.rs?

devnexen: I also added forgotten ones.

tgross35:
@asomers any thoughts about this?

I think it should be fine but needs an update because of the new macro syntax for `const`.

asomers:
Apart from the macro syntax issue, it LGTM.

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/4657/1a9bdc4d16edc21dfa66e655d1edf8e4eac16ba8..270249666bd9cd5aa953c8d479ad967e1635954e)*

PR #4658, opened 2025-08-15
ctest: add suport for C enum

mbyx:
# Description
Adds the ability to specify whether a type alias and a set of constants that use that type alias maps to a C enum, as well as the ability to skip them.

# Sources

# Checklist


mbyx:
The signededness of a C enum is implementation defined, so should that test be skipped for all C enums?

ctest-next/src/generator.rs
```diff
@@ -206,6 +209,20 @@ impl TestGenerator {
...
+    /// Indicate that a type alias is actually a C enum.
+    ///
+    /// # Examples
+    /// ```no_run
+    /// use ctest_next::TestGenerator;
+    ///
+    /// let mut cfg = TestGenerator::new();
+    /// cfg.c_enum(|e| e == "pid_type");
+    /// ```
+    pub fn c_enum(&mut self, f: impl Fn(&str) -> bool + 'static) -> &mut Self {
```
tgross35: Could this take `&Type` rather than `&str`? To match `rename_alias`.

Maybe the name `alias_is_c_enum` would be a bit more descriptive, there are a handful of ways to represent C enums https://mdaverde.com/posts/rust-bindgen-enum/

mbyx:
If I use an unsigned type in `simple.h` for the enum, it fails on the signededness test on windows. If I use a signed type it fails on linux. This isn't really a problem in the actual ports because the windows port doesn't use enums, but these integration tests are run on all platforms.

ctest-next/src/generator.rs
mbyx: It can't, because it's also used to filter out constants with the same type. The only way to do it would be to create a `Type` struct in that method, but then the `public` and `ty` fields would be incorrect.

src/macros.rs
```diff
@@ -276,7 +279,9 @@ macro_rules! c_enum {
 
     // Use a specific type if provided, otherwise default to `c_uint`
     (@ty $repr:ty) => { $repr };
+    (@ty_win $repr:ty) => { $repr };
     (@ty) => { $crate::c_uint };
+    (@ty_win) => { $crate::c_int };
```
tgross35: Instead of using different macro arms, add a `pub(crate) type CEnumRepr` to `src/types.rs` and use it here.

This will also simplify the tests, you can do e.g. `0 as CEnumRepr` rather than splitting on Windows

PR #4670, opened 2025-08-21
libc: Switch tier 3 targets to the new ctest

mbyx:
# Description
Switches some tier 3 platforms to the new ctest.

Progress:
- [X] dragonfly
- [X] cygwin
- [X] redox
- [x] vxworks
- [x] nto-qnx
- [x] aix

# Sources

# Checklist


mbyx:
Haiku seems to require CPP support, so this is blocked on that.

PR #4671, opened 2025-08-21
adding dlvsym support for linux/GNU, freebsd, dragonflybsd and netbsd.

devnexen:
ref: https://linux.die.net/man/3/dlvsym
ref: https://man.freebsd.org/cgi/man.cgi?query=dlvsym
ref: https://github.com/DragonFlyBSD/DragonFlyBSD/blob/088552723935447397400336f5ddb7aa5f5de660/include/dlfcn.h#L126
ref: https://man.netbsd.org/dlfcn.3

### DIFF
--- a/libc-test/semver/dragonfly.txt
+++ b/libc-test/semver/dragonfly.txt
@@ -1317,6 +1317,7 @@ dirfd
 dirname
 dl_iterate_phdr
 dl_phdr_info
+dlvsym
 drand48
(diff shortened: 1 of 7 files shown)


Cargo.lock
tgross35, line 1: Looks like this may have gotten picked up unintentionally, would you be able to drop it? Otherwise LGTM

Issue #4598, opened 2025-07-29
ctest: Allow defining/undefining values for specific headers


tgross35:
Inspired by https://github.com/rust-lang/libc/pull/4597, it may be nice if we had the ability to define a value only for specific headers.

PR #4673, opened 2025-08-22
ctest: add defines for specific headers and language selection

mbyx:
# Description
Allows defining values only for certain headers, as well as choosing which language to use when generating tests.

Closes rust-lang/libc#4598

# Sources

# Checklist

### DIFF
--- a/ctest/src/generator.rs
+++ b/ctest/src/generator.rs
@@ -13,8 +13,8 @@ use crate::ffi_items::FfiItems;
 use crate::template::{CTestTemplate, RustTestTemplate};
 use crate::translator::translate_primitive_type;
 use crate::{
-    Const, Field, MapInput, Parameter, Result, Static, Struct, TranslationError, Type, Union,
-    VolatileItemKind, expand, get_build_target,
+    BoxStr, Const, Field, Language, MapInput, Parameter, Result, Static, Struct, TranslationError,
+    Type, Union, VolatileItemKind, expand, get_build_target,
 };
 
 /// A function that takes a mappable input and returns its mapping as `Some`, otherwise
@@ -35,14 +35,16 @@ type CEnum = Box<dyn Fn(&str) -> bool>;
 #[derive(Default)]
 #[expect(missing_debug_implementations)]
 pub struct TestGenerator {
-    pub(crate) headers: Vec<String>,
+    pub(crate) headers: Vec<(BoxStr, Vec<BoxStr>)>,
     pub(crate) target: Option<String>,
     pub(crate) includes: Vec<PathBuf>,
     out_dir: Option<PathBuf>,
     pub(crate) flags: Vec<String>,
-    pub(crate) defines: Vec<(String, Option<String>)>,
+    pub(crate) global_defines: Vec<(String, Option<String>)>,
     cfg: Vec<(String, Option<String>)>,
     mapped_names: Vec<MappedName>,
+    /// The programming language to generate tests in.
+    pub(crate) language: Language,
     pub(crate) skips: Vec<Skip>,
     pub(crate) verbose_skip: bool,
     pub(crate) volatile_items: Vec<VolatileItem>,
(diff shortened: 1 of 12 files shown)


mbyx:
FreeBSD is failing in a different way than usual. It might be because `HashMap` doesn't maintain order of the headers.

ctest/src/generator.rs
```diff
@@ -104,6 +108,46 @@ impl TestGenerator {
...
+    /// # Examples
+    ///
+    /// ```no_run
+    /// use ctest::TestGenerator;
+    ///
+    /// let mut cfg = TestGenerator::new();
+    /// cfg.header_with_defines("foo.h", vec![])
+    ///    .header_with_defines("bar.h", vec!["DEBUG", "DEPRECATED"]);
+    /// ```
+    pub fn header_with_defines(&mut self, header: &str, defines: Vec<&str>) -> &mut Self {
```
tgross35: For `defines`, take an `impl Iterator<Item = &str>` so it accepts arrays and slices as well.

ctest/src/generator.rs
```diff
@@ -104,6 +108,46 @@ impl TestGenerator {
...
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// use ctest::{TestGenerator, Language};
+    ///
+    /// let mut cfg = TestGenerator::new();
+    /// cfg.language(Language::CXX);
+    /// ```
+    pub fn language(&mut self, language: Language) -> &mut Self {
```
tgross35: In the docs, could you clarify where this difference shows up? Is it just for what compiler is invoked on the generated files?

ctest/src/lib.rs
```diff
@@ -74,6 +74,35 @@ pub(crate) enum MapInput<'a> {
...
+        }
+    }
+
+    /// Return the linkage to use for each language.
+    pub fn linkage(&self) -> &str {
+        match self {
+            Self::C => "",
+            Self::CXX => r#"extern "C" "#,
+        }
+    }
```
tgross35: Can these be `pub(crate)`?

ctest/templates/test.c
```diff
@@ -21,7 +33,7 @@ static char *ctest_const_{{ const_cstr.id }}_val_static = {{ const_cstr.c_val }}
 
 // Define a function that returns a pointer to the value of the constant to test.
 // This will later be called on the Rust side via FFI.
-char *ctest_const_cstr__{{ const_cstr.id }}(void) {
+{{ self.linkage }}char *ctest_const_cstr__{{ const_cstr.id }}(void) {
```
tgross35: Could `self.linkage` instead be a `#define CTEST_EXTERN` within the existing `#ifdef __cplusplus`? That way a `linkage` variable doesn't need to propagate through everything.

ctest/src/template.rs
```diff
@@ -36,6 +38,8 @@ impl RustTestTemplate {
 pub(crate) struct CTestTemplate {
     pub template: TestTemplate,
     pub headers: Vec<String>,
+    pub header_defines: HashMap<String, Vec<String>>,
```
tgross35: Instead of keeping this as separate fields, could `headers` instead become `Vec<(BoxStr, Vec<BoxStr>)>`? Or put those two fields into a new struct.

This should also allows the same header to be used more than once with a different set of `#define`s. Probably not very useful for our purposes, but it does come up in C land.

ctest/tests/input/macro.h
```diff
@@ -11,3 +12,16 @@ struct VecU16
...
+    int8_t x;
+    int8_t y;
+};
+
+struct VecI16
+{
+    int16_t x;
+    int16_t y;
+};
+#endif
```
tgross35: Trailing `\n`.

Which editor do you use? Many have the option to ensure a single trailing newline, e.g. for vscode it's `files.insertFinalNewline` and for Helix it is `insert-final-newline`.

ctest/tests/input/macro.h
```diff
@@ -1,5 +1,6 @@
 #include <stdint.h>
 
+#ifdef UNSIGNED
```
tgross35: Since we only use this file once I don't think we need "real" content here. Instead it could be:

```c
#ifndef SUPPRESS_ERROR
#error Expected SUPPRESS_ERROR to be defined (testing per-file defines)
#endif
```

And then in a separate header, swap the `ifdef` for a `ifndef` to validate that the define is only getting set for a single file like we expect:

```c
#ifdef SUPPRESS_ERROR
#error Expected SUPPRESS_ERROR to not be defined (testing per-file defines)
#endif
```

PR #4674, opened 2025-08-22
libc: fix reg32 on 32 bit freebsd

mbyx:
# Description
Attempts to fix errors in freebsd i686 by making sure that the correct build target is used when expanding macros.

# Sources
https://cgit.freebsd.org/src/tree/sys/x86/include/reg.h?h=releng/13.1

# Checklist

### DIFF
--- a/ctest/src/macro_expansion.rs
+++ b/ctest/src/macro_expansion.rs
@@ -6,7 +6,11 @@ use std::process::Command;
 use crate::{EDITION, Result};
 
 /// Use rustc to expand all macros and pretty print the crate into a single file.
-pub fn expand<P: AsRef<Path>>(crate_path: P, cfg: &[(String, Option<String>)]) -> Result<String> {
+pub fn expand<P: AsRef<Path>>(
+    crate_path: P,
+    cfg: &[(String, Option<String>)],
+    target: String,
+) -> Result<String> {
     let rustc = env::var("RUSTC").unwrap_or_else(|_| String::from("rustc"));
 
     let mut cmd = Command::new(rustc);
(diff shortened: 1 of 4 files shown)


mbyx:
There's only one `reg32` struct in `libc`, and it should only be included if the target is 64 bit, but it fails on i686 which is a 32 bit platform. Looking at the `.cirrus.yml` file, I found the evidence: `# Test i686 FreeBSD in 32-bit emulation on a 64-bit host.`

What's happening is that libc is being expanded as if it was 64 bit, causing this problem.

I'm not sure but we might be able to fix this if we pass in the correct target when expanding macros?

ctest/src/macro_expansion.rs
tgross35, line 1: The changes to this file LGTM, but the commit title is "libc: fix reg32 on 32 bit freebsd" (8d34aff51d5287798b125eb52ce8eaa94fbf1dfe), this doesn't have anything to do with that. Could you split changes to `ctest/` to a separate commit?

tgross35: Sorry, small correction - there isn't anything to split, just update the commit title to say what the change actually does (pass the target to macro expansion), and mention in the description that this fixes reg32 on 32-bit FreeBSD.

ctest/src/macro_expansion.rs
```diff
@@ -16,6 +21,16 @@ pub fn expand<P: AsRef<Path>>(crate_path: P, cfg: &[(String, Option<String>)]) -
         .arg(EDITION) // By default, -Zunpretty=expanded uses 2015 edition.
         .arg(canonicalize(crate_path)?);
 
+    let target = target
+        .or_else(|| env::var("TARGET").ok())
+        .or_else(|| env::var("TARGET_PLATFORM").ok())
+        .ok_or(GenerationError::EnvVarNotFound(
+            "TARGET, TARGET_PLATFORM".to_string(),
+        ))?;
```
tgross35: This same code exists at https://github.com/rust-lang/libc/blob/586ccc1e98f4d28d3c3a80b0fcb2235228f0b6c0/ctest/src/runner.rs#L24-L31. Could it be split to a separate function and reused?

PR #4675, opened 2025-08-24
Switch tier 2 targets to the new ctest

mbyx:
# Description
~Blocked on rust-lang/libc#4674.~

Progress:
- [X] wasi
- [X] solaris/illumos
- [X] NetBSD
- [X] emscripten
- [X] android

The `ioctl` function is now skipped on android because the new method of checking function addresses does not work with overloaded functions.

# Sources

# Checklist


mbyx:
The error here is caused by constants like `J1939_MAX_UNICAST_ADDR`, which are only present in the recently added `new` module. This should only be added if the target is linux and not android. So I'm assuming it's the same issue as before, macro expansion needs to be specified the correct target manually. So most likely, blocked on #4674

PR #4676, opened 2025-08-25
Adding required defines and functions for vxworks

gautam899:
Hi, this adds required defines and functions for VxWorks. I have also added changes in libc-test required for vxworks

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -1841,11 +1841,28 @@ cfg_if! {
     if #[cfg(any(target_os = "aix", target_os = "nto"))] {
         extern "C" {
             pub fn cfmakeraw(termios: *mut crate::termios) -> c_int;
-            pub fn cfsetspeed(termios: *mut crate::termios, speed: crate::speed_t) -> c_int;
         }
     } else if #[cfg(not(any(target_os = "solaris", target_os = "illumos",)))] {
         extern "C" {
             pub fn cfmakeraw(termios: *mut crate::termios);
+        }
+    }
+}
+
+cfg_if! {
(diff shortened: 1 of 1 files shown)


biabbas (approved):

libc-test/build.rs
```diff
@@ -3539,7 +3551,7 @@ fn test_vxworks(target: &str) {
     // FIXME(vxworks)
     cfg.skip_fn(move |func| match func.ident() {
         // sighandler_t
-        "signal"
+        | "signal"
         // not used in static linking by default
         | "dlerror" => true,
         _ => false,
```
tgross35, line 3524: Possible unintentional change

hax0kartik:
> Thank you for confirming that. `_realpath` and `sa_u_t ` were removed then?
> 
> These seem like breaking changes in the C world as well, did they come with a new vxworks release? We currently document a minimum of 7, but that will need to be increased if `std` relies on anything here https://doc.rust-lang.org/beta/rustc/platform-support/vxworks.html. Or split the targets like for QNX. (Just trying to wrap my head around what exactly the changes here mean for users.)
> 
> `flock` makes sense, there is some history at #4043. That struct never actually made it to a release because of [#4043 (comment)](https://github.com/rust-lang/libc/pull/4043#issuecomment-2493348078), so the fix here seems correct, and what should have been added in the first place.

Yes, with vxworks 7 25.09 release, _realpath was made private, and the changes related to timespec were done. The reason behind removing sa_u_t is pretty much the same as #4249. 

We are also working on upstreaming the standard library patches which will require these changes. I think we should update that document to reflect the minimum supported release id as well.

PR #4677, opened 2025-08-28
add `qsort_r` for unix, and `qsort` and `qsort_s` for windows

folkertdev:
# Description

Add `qsort_r` for apple, freebsd and netbsd targets, and `qsort` and `qsort_s` for windows targets.

# Sources

- https://linux.die.net/man/3/qsort_r
- https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/qsort_r.3.html 
- https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/qsort?view=msvc-170
- https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/qsort-s?view=msvc-170
- https://man.freebsd.org/cgi/man.cgi?query=qsort_r&sektion=3&format=html
- https://man.netbsd.org/qsort.3

# Checklist

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/freebsd11/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/freebsd11/mod.rs
@@ -424,6 +424,18 @@ extern "C" {
     // in FreeBSD 12
     pub fn dirname(path: *const c_char) -> *mut c_char;
     pub fn basename(path: *const c_char) -> *mut c_char;
+
+    // Argument order of the function pointer changed in FreeBSD 14. From 14 onwards the signature
+    // matches the POSIX specification by having the third argument be a mutable pointer, on
+    // earlier versions the first argument is the mutable pointer.
+    #[link_name = "qsort_r@FBSD_1.0"]
+    pub fn qsort_r(
+        base: *mut c_void,
+        num: size_t,
+        size: size_t,
+        arg: *mut c_void,
+        compar: Option<unsafe extern "C" fn(*mut c_void, *const c_void, *const c_void) -> c_int>,
+    );
 }
 
 cfg_if! {
--- a/src/unix/bsd/freebsdlike/freebsd/freebsd12/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/freebsd12/mod.rs
@@ -468,6 +468,15 @@ extern "C" {
 
     pub fn dirname(path: *mut c_char) -> *mut c_char;
     pub fn basename(path: *mut c_char) -> *mut c_char;
+
+    #[link_name = "qsort_r@FBSD_1.0"]
+    pub fn qsort_r(
+        base: *mut c_void,
+        num: size_t,
+        size: size_t,
+        arg: *mut c_void,
+        compar: Option<unsafe extern "C" fn(*mut c_void, *const c_void, *const c_void) -> c_int>,
+    );
 }
 
 cfg_if! {
--- a/src/unix/bsd/freebsdlike/freebsd/freebsd13/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/freebsd13/mod.rs
@@ -507,6 +507,15 @@ extern "C" {
 
     pub fn dirname(path: *mut c_char) -> *mut c_char;
     pub fn basename(path: *mut c_char) -> *mut c_char;
+
+    #[link_name = "qsort_r@FBSD_1.0"]
+    pub fn qsort_r(
+        base: *mut c_void,
+        num: size_t,
+        size: size_t,
+        arg: *mut c_void,
+        compar: Option<unsafe extern "C" fn(*mut c_void, *const c_void, *const c_void) -> c_int>,
+    );
 }
 
 #[link(name = "kvm")]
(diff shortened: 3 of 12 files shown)


semarie:
`qsort_r` doesn't exist in OpenBSD, but the NetBSD definition is placed
in common file. Please move the definition in N#tBSD sp#cific file.

Thanks

folkertdev:
Sure, I moved the definition to `src/unix/bsd/netbsdlike/netbsd/mod.rs`, and I believe the semver entry was already correct.

collinfunk:
There is also a seperate and incompatible API that exists on older versions of FreeBSD before it was changed in 14.0 [1] [2]. Not sure if that needs to be handled here.

[1] https://www.gnu.org/software/gnulib/manual/html_node/qsort_005fr.html
[2] https://man.freebsd.org/cgi/man.cgi?query=qsort_r

src/unix/bsd/freebsdlike/freebsd/freebsd14/mod.rs
```diff
@@ -509,6 +509,14 @@ extern "C" {
...
     pub fn dirname(path: *mut c_char) -> *mut c_char;
     pub fn basename(path: *mut c_char) -> *mut c_char;
+
+    pub fn qsort_r(
+        base: *mut c_void,
+        num: size_t,
+        size: size_t,
+        compar: Option<unsafe extern "C" fn(*const c_void, *const c_void, *mut c_void) -> c_int>,
+        arg: *mut c_void,
+    );
```
tgross35, line 519: Isn't `thunk` before `compar` in FreeBSD 14 and 15?

folkertdev: it's not, the signature is

```
 void
       qsort_r(void *base,	       size_t nmemb,		  size_t size,
	   int (*compar)(const void *, const void *, void *), void *thunk);
```

- https://man.freebsd.org/cgi/man.cgi?query=qsort_r&apropos=0&sektion=0&manpath=FreeBSD+15.0-CURRENT&arch=default&format=html
- https://man.freebsd.org/cgi/man.cgi?query=qsort_r&apropos=0&sektion=0&manpath=FreeBSD+14.0-RELEASE&arch=default&format=html

(you have to re-submit the form for the correct docs version to load)

tgross35: Argh sorry I commented the exact opposite of what I meant: on <= 13, it looks like it is before `compar`

```text
       void
       qsort_r(void *base,     size_t nmemb,	 size_t	size,	  void *thunk,
	   int (*compar)(void *, const void *, const void *));
```

image

tgross35: Cc @asomers for the FreeBSD review in any case

folkertdev: ah, right, fixed.

src/unix/bsd/freebsdlike/freebsd/freebsd11/mod.rs
```diff
@@ -424,6 +424,17 @@ extern "C" {
     // in FreeBSD 12
     pub fn dirname(path: *const c_char) -> *mut c_char;
     pub fn basename(path: *const c_char) -> *mut c_char;
+
+    // Argument order of the function pointer changed in FreeBSD 14. From 14 onwards the signature
+    // matches the POSIX specification by having the third argument be a mutable pointer, on
+    // earlier versions the first argument is the mutable pointer.
+    pub fn qsort_r(
```
asomers, line 432:
```suggestion
    #[link_name = "qsort_r@FBSD_1.0"]
    pub fn qsort_r(
```

asomers: I think you need to specify the link_name here.  Otherwise, you will link to the wrong symbol if you build CI in FreeBSD 11 compatible mode (or 12, or 13) but run if on FreeBSD 14 or 15.  For example, if you try to run `sh ci/run.sh x86_64-unknown-freebsd` on FreeBSD 14 or 15 (note the lack of `LIBC_CI=1`).

src/unix/bsd/freebsdlike/freebsd/freebsd12/mod.rs
```diff
@@ -468,6 +468,14 @@ extern "C" {
 
     pub fn dirname(path: *mut c_char) -> *mut c_char;
     pub fn basename(path: *mut c_char) -> *mut c_char;
+
+    pub fn qsort_r(
```
asomers, line 473:
```suggestion
    #[link_name = "qsort_r@FBSD_1.0"]
    pub fn qsort_r(
```

src/unix/bsd/freebsdlike/freebsd/freebsd13/mod.rs
```diff
@@ -507,6 +507,14 @@ extern "C" {
 
     pub fn dirname(path: *mut c_char) -> *mut c_char;
     pub fn basename(path: *mut c_char) -> *mut c_char;
+
+    pub fn qsort_r(
```
asomers, line 512:
```suggestion
    #[link_name = "qsort_r@FBSD_1.0"]
    pub fn qsort_r(
```

asomers (changes requested):

asomers (approved):
I think this PR is ok now.  HOWEVER, I would like to wait to merge it until after #4685, #4690, and #4691 are resolved.  The we can rebase this PR, and see if CI passes on FreeBSD.

PR #4681, opened 2025-09-01
libc: Switch Haiku to `ctest` and remove `ctest_old` dependency

mbyx:
# Description
Ports Haiku to use the new `ctest`, so that everything now depends on this new `ctest`.

# Sources

# Checklist


mbyx:
Done

PR #4682, opened 2025-09-01
libc: uncomment items that were commented due to old ctest

mbyx:
# Description

# Sources

# Checklist

### DIFF
--- a/src/primitives.rs
+++ b/src/primitives.rs
@@ -99,35 +99,6 @@ cfg_if! {
             target_os = "watchos"
         ))
     ))] {
-        // This introduces partial support for FFI with __int128 and
-        // equivalent types on platforms where Rust's definition is validated
-        // to match the standard C ABI of that platform.
-        //
-        // Rust does not guarantee u128/i128 are sound for FFI, and its
-        // definitions are in fact known to be incompatible. [0]
-        //
-        // However these problems aren't fundamental, and are just platform
-        // inconsistencies. Specifically at the time of this writing:
-        //
-        // * For x64 SysV ABIs (everything but Windows), the types are underaligned.
-        // * For all Windows ABIs, Microsoft doesn't actually officially define __int128,
-        //   and as a result different implementations don't actually agree on its ABI.
-        //
-        // But on the other major aarch64 platforms (android, linux, ios, macos) we have
-        // validated that rustc has the right ABI for these types. This is important because
-        // aarch64 uses these types in some fundamental OS types like user_fpsimd_struct,
-        // which represents saved simd registers.
-        //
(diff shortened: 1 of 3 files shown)


src/primitives.rs
```diff
@@ -143,34 +145,32 @@ cfg_if! {
...
+        static_assert_eq!(align_of::<__int128>(), _ALIGN_128);
+
+        static_assert_eq!(size_of::<__uint128>(), _SIZE_128);
+        static_assert_eq!(align_of::<__uint128>(), _ALIGN_128);
+
+        static_assert_eq!(size_of::<__int128_t>(), _SIZE_128);
+        static_assert_eq!(align_of::<__int128_t>(), _ALIGN_128);
+
+        static_assert_eq!(size_of::<__uint128_t>(), _SIZE_128);
+        static_assert_eq!(align_of::<__uint128_t>(), _ALIGN_128);
```
tgross35: This whole bit can actually be deleted, also the entire "This introduces partial support for FFI with __int128..." section above. The mentioned ABI issue has since been resolved, and since https://github.com/rust-lang/rust/pull/137306 we say `i128` will be compatible with `__i128` on platforms where it is available.

Should resolve the test failure.

mbyx: The other error is related to `debug_assert`, I'm not sure where to put the import, since putting it at the top will make it unused on some platforms.

tgross35: I guess just do `core::debug_assert!`, same for any other needed imports here. The style thing should be fixed, but I don't think it's really worth worrying about

src/primitives.rs
```diff
@@ -88,101 +88,7 @@ pub type uint32_t = u32;
...
-        // The list of supported architectures and OSes is intentionally very restricted,
-        // as careful work needs to be done to verify that a particular platform
-        // has a conformant ABI.
-        //
-        // [0]: https://github.com/rust-lang/rust/issues/54341
-
-        /// C `__int128` (a GCC extension that's part of many ABIs)
-        pub type __int128 = i128;
-        /// C `unsigned __int128` (a GCC extension that's part of many ABIs)
-        pub type __uint128 = u128;
```
tgross35: These should probably be kept under the above `#[cfg(all(...))]` gate

src/primitives.rs
tgross35, line 1: Sorry for the confusion; to clarify a bit, for https://github.com/rust-lang/libc/pull/4682#discussion_r2324382950 I only meant that the quoted bit (long explanation + checks) can be deleted. `__u?int128(_t)?` need to stay around on the same platforms, so the `cfg_if` block needs to stay.

This should let you drop the build.rs changes

PR #4683, opened 2025-09-01
Add contents of `linux/can/bcm.h`

coolreader18:
# Description

There's already `can::{raw, j1939}`, and this adds `bcm`.

# Sources

https://github.com/torvalds/linux/blob/b320789d6883cc00ac78ce83bccbfe7ed58afcf0/include/uapi/linux/can/bcm.h

# Checklist


src/new/linux_uapi/linux/can/bcm.rs
```diff
@@ -0,0 +1,47 @@
...
+pub const TX_READ: crate::__u32 = 3;
+pub const TX_SEND: crate::__u32 = 4;
+pub const RX_SETUP: crate::__u32 = 5;
+pub const RX_DELETE: crate::__u32 = 6;
+pub const RX_READ: crate::__u32 = 7;
+pub const TX_STATUS: crate::__u32 = 8;
+pub const TX_EXPIRED: crate::__u32 = 9;
+pub const RX_STATUS: crate::__u32 = 10;
+pub const RX_TIMEOUT: crate::__u32 = 11;
+pub const RX_CHANGED: crate::__u32 = 12;
```
tgross35: Use the `c_enum!` macro since the source is an enum

coolreader18: The original enum has no name, though

coolreader18: Added anonymous enum support to `c_enum!`, though I totally get if you'd rather that be split to another PR.

tgross35:
Hey, sorry for the delay here. I wanted to require making it explicit that the enum should be anonymous, so I just went ahead and merged that in 02c4d661d222482ea18dc3baa149e0bed7210f3e. Would you be able to update using `#anon`?

@rustbot author

src/new/linux_uapi/linux/can/bcm.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub struct bcm_msg_head {
+        pub opcode: crate::__u32,
+        pub flags: crate::__u32,
+        pub count: crate::__u32,
+        pub ival1: bcm_timeval,
+        pub ival2: bcm_timeval,
+        pub can_id: canid_t,
+        pub nframes: crate::__u32,
+        pub frames: [can_frame; 0],
+    }
```
tgross35, line 20: `u32` can be used directly rather than `crate::__u32`. Eventually we'll be getting rid of the `__u*` types.

libc-test/build.rs
```diff
@@ -4822,7 +4823,9 @@ fn test_linux(target: &str) {
         // the `xsk_tx_metadata_union` field is an anonymous union
         (struct_ == "xsk_tx_metadata" && field == "xsk_tx_metadata_union") ||
         // After musl 1.2.0, the type becomes `int` instead of `long`.
-        (old_musl && struct_ == "utmpx" && field == "ut_session")
+        (old_musl && struct_ == "utmpx" && field == "ut_session") ||
+        // `frames` is a VLA
```
tgross35: Nit: VLAs are runtime-sized arrays on the stack. The empty array at the end of a struct is a "flexible array member"

coolreader18: Ah, gotcha. That's what I thought, but it was called as such above, so maybe I'll edit that too.

tgross35:
Thank you!

Issue #4651, opened 2025-08-14
Warning: <type> is not FFI-safe, after recent changes involving `#[non_exhaustive]`


neuschaefer:
After the recent commit (9ee2d2a0f6180e092ada3f9a5960a31b290fa755) in which I added `#[non_exhaustive]` to `kproc_info` on FreeBSD 15, I get warnings like the following when I run libc-test:

```
warning: `extern` block uses type `libc::kinfo_proc`, which is not FFI-safe
      --> /home/jn/libc/target/debug/build/libc-test-9d66e31a05563bd6/out/ctest_output.rs:132002:62
       |
132002 |             fn ctest_field_ptr__kinfo_proc__ki_structsize(a: *const kinfo_proc) -> *mut u8;
       |                                                              ^^^^^^^^^^^^^^^^^ not FFI-safe
       |
       = note: this struct is non-exhaustive
       = note: `#[warn(improper_ctypes)]` on by default

```

This seems to point to a larger problem with using `#[non_exhaustive]` in the libc crate.

cc @tgross35

neuschaefer:
There is one instance in the tests where a type U=kinfo_proc is passed by value, it's the roundtrip test, `ctest_roundtrip_kinfo_proc` (Rust) / `ctest_roundtrip__kinfo_proc` (C) in this case. The C function takes a `MaybeUninit<U>` as its first argument, and returns a `U`.

If this is considered all right, I'd suggest `#[allow]`-ing `improper_ctypes` for the ctest_output module.

PR #4685, opened 2025-09-05
libc: remove TIOCMGDTRWAIT and TIOCMSDTRWAIT for freebsd15

mbyx:
# Description

FreeBSD 15 removed `TIOCMGDTRWAIT` and `TIOCMSDTRWAIT`, so the corresponding constants are changed to only appear in FreeBSD 13 and 14 builds.

# Sources
https://github.com/freebsd/freebsd-src/commit/4b4cf0d52c43e33c32983df60a8ea16b3d9e2f20
https://cgit.freebsd.org/src/commit/?id=4b4cf0d52c43e33c32983df60a8ea16b3d9e2f20

# Checklist


mbyx:
This fixes the error in freebsd15 related to these constants, so that we now get the same issue as #4651. Since fixing it is being actively worked on in https://github.com/rust-lang/rust/pull/134697, eventually freebsd15 will fix itself.

asomers (approved):

PR #4688, opened 2025-09-07
Raise libc's FreeBSD ABI to 12

asomers:
# Description

Raise the ABI used by libc on FreeBSD to FreeBSD 12, rather than 11.  That includes significant changes, such as 64-bit inodes, and support for the riscv64 GENERIC kernel.  This is a backport of #2406 .

# Checklist


build.rs
```diff
@@ -69,11 +69,11 @@ fn main() {
         println!("cargo:warning=setting FreeBSD version to {vers}");
         vers
     } else if libc_ci {
-        which_freebsd().unwrap_or(11)
+        which_freebsd().unwrap_or(12)
     } else if rustc_dep_of_std {
         12
     } else {
-        11
+        12
```
joshtriplett, line 74: This makes the last two branches of the if identical, which is causing it to fail checks in CI.

cperciva:
Thanks everyone for getting this landed!

Issue #4692, opened 2025-09-13
glibc 2.42 changed the definitions of baud rate constants (B50, B75, ..., B9600, B57600, ...)


de-vri-es:
In glibc 2.41 and before, baud rate constants like B50, B75, ..., B9600, B57600, B115200, ... were enum-like values. B50 was 1, B75 was 2, etc.

In glibc 2.42 they have been redefined. B50 is now just 50, and B75 is now just 75. Additionally, the `tcset{i,o,}speed` and `tcget{i,o,}speed` functions are adjusted to deal with this internally (the kernel struct going to the ioctl still needs the legacy values, or a custom baud rate with `BOTHER`).

The old baud rate constants are still available as __B50, __B75, ...

https://inbox.sourceware.org/libc-announce/5906001.DvuYhMxLoT@pinacolada/T/#u

> * On Linux, the <termios.h> interface now supports arbitrary baud rates;   speed_t is redefined to simply be the baud rate specified as an unsigned int, which matches the kernel interface.

So.. is there a clean way to deal with this in the libc crate? The only portable thing for an application to do now is to manually do the `TC{G,S}ETS2` ioctl and completely ignore the baud rate constants.

hpax:
See also discussion at:  https://github.com/nix-rust/nix/issues/2672

de-vri-es:
Maybe explicitly linking to an old version of the `tc{g,s}et{i,o,}speed` functions that support the old baud rate constants will work for now?

I guess the only question is: which version? :]

hpax:
That would be a very poor solution, and will make users do what they have done in C code, poorly, for a long time now: invoke ioctl() directly, usually incorrectly. 

See my notes in the other linked thread.

The legacy version number varies depending on architecture; also, on MIPS and SPARC the tcsetattr() and tcgetattr() symbols are also versioned, as the format of struct termios has changed. 

There are also the new cfget/setbaud() interfaces. If I'm not entirely mistaken, you might be able to do a weak undefined symbol to pick them up if available. 

Finally, here is a trick to probe at initialization time: create an empty struct termios, and call something like cfsetospeed(&dummy_termios, 0x40000000) – of of succeeds, you are using the new glibc.

You *do* know that the legacy baud rates constants, as well as struct termios format, are architecture-dependent, right?!

de-vri-es:
No problem, thanks for the useful insights and information :)

hpax:
[...]

PR #4697, opened 2025-09-16
Fix baud rate constants for glibc 2.42 and up.

de-vri-es:
In glibc 2.42, baud rate constants changed. Their value is now simply the baud rate as integer. Additionally, the `tcset{i,o,}speed` function accepts any arbitrary baud rate, and the `tcget{i,o,}speed` can return any arbitrary baud rate.

On MIPS and SPARC, the termios struct from glibc now also has the `c_ispeed` and `c_ospeed` fields. This is not a backwards compatible change, and I'm not sure how this should be addressed. Adding public or private fields will break code that create these structs manually. But we need to make sure the size is correct for glibc 2.42.

Making the existence of these fields depend on the glibc version is a nightmare for docs.rs and portable code. But adding these fields on mips and sparc without glibc 2.42 means the fields will not be initialized by libc calls, so anyone using `MaybeUninit::assume_init()` could be triggering undefined behaviour.

Also note: there is no explicit linking against glibc 2.42 symbols. I was thinking we could force having `GLIBC_2.42` as undefined symbol, but if you call any of the libc functions that interpret or return the baud rates, you will already be linking against versioned symbols implicitly and the binary will fail to load at runtime on systems with an older glibc. I think this is probably good enough.

Should fix rust-lang/libc#4692.

# Sources
Baud rate constants: https://github.com/bminor/glibc/blob/13d67746cbe1273afaf6b9de9d6065ab76ee7697/bits/termios-baud.h#L24-L70
termios struct: https://github.com/bminor/glibc/blob/13d67746cbe1273afaf6b9de9d6065ab76ee7697/sysdeps/unix/sysv/linux/bits/termios-struct.h#L32-L42


hpax:
How come this is present in the "new" definition?

        cfg_if! {
            if #[cfg(not(any(target_arch = "sparc", target_arch = "sparc64")))] {

src/unix/linux_like/linux/gnu/mod.rs
```diff
@@ -1045,6 +1053,48 @@ pub const REG_EEND: c_int = 14;
...
+        pub const B1500000: crate::speed_t = 1500000;
+        pub const B2000000: crate::speed_t = 2000000;
+        cfg_if! {
+            if #[cfg(not(any(target_arch = "sparc", target_arch = "sparc64")))] {
+                pub const B2500000: crate::speed_t = 2500000;
+                pub const B3000000: crate::speed_t = 3000000;
+                pub const B3500000: crate::speed_t = 3500000;
+                pub const B4000000: crate::speed_t = 4000000;
+            }
+        }
```
de-vri-es, line 1094: @hpax: You mean this? This is just a pragmatic choice to avoid having constants in the `libc` crate whose presence depends on the glibc version at build time. They're not available pre-2.42. Starting from 2.42, the baud rate constants are basically pointless (which is nice), so I don't think anyone is hurt by not having them.

asomers (changes requested):
This PR runs contrary to existing libc practice.  The existing practice is to provide bindings that work with the oldest supported version of each OS.  That's why libc doesn't run bindgen at build time.  Instead, it relies on pregenerated bindings.  For glibc, the bindings must be compatible with glibc 2.17, because that's Rust's minimum requirement, and Rust depends on libc.
Furthermore, attempting to detect the installed glibc version at build time makes cross-compiling much harder.
So I think that instead of attempting to detect the glibc version at runtime, we should only use the old baud rate symbols.  And we should set the ELF symbol version (i.e. Rust's `#[link_name]` attribute) for any glibc function that takes those symbols.  That way Rust programs using libc will continue to work with any glibc version.

de-vri-es:
Any symbol really. It will automatically use the latest version of any symbol you use. And if that one isn't present in the old glibc being used at runtime, the dynamic linker will fail to load the binary.

For example: when I compile a program on Arch Linux, I can not run it on Ubuntu, because the glibc of ubuntu is too old. The other way around works fine.

de-vri-es:
> As @asomers mentioned, there is no guarantee that the host glibc matches the version that is getting linked, so this isn't going to be robust. I don't think we have a good way to detect that version, so our best bet (for now) is going to have to be based on `link_name`.

This isn't totally true. If your program calls cfgetispeed (for example), it will end up linking `cfgetispeed@GLIBC_2.42`. This is just what the linker does already.

If you try to run the program with an older glibc it will fail to resolve the symbol at runtime.

This already happens now with all symbols from glibc.

So I think technically the solution in the PR is fine, and mimics how glibc is supposed to be used.

de-vri-es:
> I see you got it with your edit - but yeah, doing something like e.g. building on Arch and deploying on Debian means you'll hit the same problems, just a bit more sneaky. I'm going to ask around and see if anyone can think of a good way for us to do more proper symbol versioning (more to talk about at #4700).

Yeah, my point was that this already happens now, so I didn't think it is a *new* problem. But when cross compiling for a totally different architecture it is certainly a problem.

> In the meantime, is there any downside of specifying a specific older symbol version? Aside from, of course, not having the latest API.

It should be fine to link the older versions. That's what binaries linked against older glibc versions do anyway. As long as we get all the right ones :]

tgross35:
The problem is what is mentioned at https://github.com/rust-lang/libc/pull/4697#pullrequestreview-3241669574. The build script is checking the host glibc version: if the program will run on the same machine then that's fine, but if you're distributing a binary or cross compiling then this becomes meaningless.

oech3:
>The build script is checking the host glibc

So we only need solution for this with this PR. Right? (We can already distribute binary linked with old glibc to Arch).

PR #4695, opened 2025-09-15
Move deprecated attribute from `extern "C"` block to functions

joshtriplett:
Fixes `useless_deprecated` error that's breaking CI.

PR #4698, opened 2025-09-18
Update Ubuntu to 25.10

tgross35:
Cherry picked from https://github.com/rust-lang/libc/pull/4695

### DIFF
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -2682,7 +2682,7 @@ pub const ST_NOATIME: c_ulong = 1024;
 pub const ST_NODIRATIME: c_ulong = 2048;
 
 pub const RTLD_NEXT: *mut c_void = -1i64 as *mut c_void;
-pub const RTLD_DEFAULT: *mut c_void = 0i64 as *mut c_void;
+pub const RTLD_DEFAULT: *mut c_void = ptr::null_mut();
 pub const RTLD_NODELETE: c_int = 0x1000;
 pub const RTLD_NOW: c_int = 0x2;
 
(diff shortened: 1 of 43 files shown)


src/unix/linux_like/linux/mod.rs
```diff
@@ -2129,7 +2129,8 @@ pub const _PC_ALLOC_SIZE_MIN: c_int = 18;
 pub const _PC_SYMLINK_MAX: c_int = 19;
 pub const _PC_2_SYMLINKS: c_int = 20;
 
-pub const MS_NOUSER: c_ulong = 0xffffffff80000000;
+// FIXME: should this be an int? The suffix is `U` not `UL`.
+pub const MS_NOUSER: c_ulong = 1 << 31;
```
collinfunk, line 2133: I wrote the glibc patch. Sorry, I didn't expect breakage here.

Here is my rational from the commit message [1]:
```
Using gcc -Wshift-overflow=2 -Wsystem-headers to compile a file
including <sys/mount.h> will cause a warning since 1 << 31 is undefined
behavior on platforms where int is 32-bits.
```
Technically unsigned integers aren't allowed in C enum types, it is a GCC extension that is widely supported. We decided it was safe to use since the extension is already required for `EPOLLONESHOT` and `EPOLLET` [2].

Regarding the `FIXME`, I am not super familiar with Rust, but the `mount` function uses `unsigned long` for the `flags` argument. Until C23 enums could not have a type other than `int`, otherwise `UL` probably would have made more sense. Hopefully that helps you a bit in your decision.

[1] https://inbox.sourceware.org/libc-alpha/20250217055647.493638-1-collin.funk1@gmail.com/
[2] https://inbox.sourceware.org/libc-alpha/875xjwvrhj.fsf@oldenburg.str.redhat.com/

tgross35: Your patch makes complete sense, I assume the intent was always to be `0x80000000` rather than the sign-extended `0xffffffff80000000`. I just left that fixme as something to double check after the experimentation here; surprised you found this PR out of the blue, but thank you for saving me the work :)

> I wrote the glibc patch. Sorry, I didn't expect breakage here.

Not at all a problem, there's nothing user-visible—just a test failure checking the equality of constants. Expected whenever the libraries we test against get updated.

PR #4720, opened 2025-09-23
Test a few tier 3 targets

Gelbpunkt:
# Description

Please don't look at this yet, I'm messing with CI.

# Sources

# Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -277,6 +277,66 @@ jobs:
           path: ${{ env.ARCHIVE_PATH }}
           retention-days: 5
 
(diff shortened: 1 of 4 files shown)


tgross35:
@Gelbpunkt out of curiosity, any reason not to promote these targets to T2 so we can test them proper? You're effectively acting as the target maintainer already :) (@neuschaefer seems involved in a lot of the same targets, they be interested in co-maintaining)

tgross35:
Fyi some of the ctest issues you're hitting here were probably resolved by 111d3286d24dda0f9067687dcf413ce733982f04

PR #4725, opened 2025-09-26
Move keyctl to new/linux_uapi

neuschaefer:
# Description

These constants are entirely defined in the Linux UAPI.

cc @tgross35 @Gelbpunkt 

# Sources

* https://elixir.bootlin.com/linux/v6.16.9/source/include/uapi/linux/keyctl.h

# Checklist

- [ ] ~~Relevant tests in `libc-test/semver` have been updated~~ no change

### DIFF
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -889,42 +889,6 @@ pub const GENL_ID_PMCRAID: c_int = crate::NLMSG_MIN_TYPE + 2;
 
 pub const ELFOSABI_ARM_AEABI: u8 = 64;
 
-// linux/keyctl.h
-pub const KEYCTL_DH_COMPUTE: u32 = 23;
-pub const KEYCTL_PKEY_QUERY: u32 = 24;
-pub const KEYCTL_PKEY_ENCRYPT: u32 = 25;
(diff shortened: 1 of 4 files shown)


src/unix/linux_like/linux/gnu/mod.rs
```diff
@@ -889,42 +889,6 @@ pub const GENL_ID_PMCRAID: c_int = crate::NLMSG_MIN_TYPE + 2;
...
-pub const KEYCTL_SUPPORTS_DECRYPT: u32 = 0x02;
-pub const KEYCTL_SUPPORTS_SIGN: u32 = 0x04;
-pub const KEYCTL_SUPPORTS_VERIFY: u32 = 0x08;
-cfg_if! {
-    if #[cfg(not(any(
-        target_arch = "mips",
-        target_arch = "mips32r6",
-        target_arch = "mips64",
-        target_arch = "mips64r6"
-    )))] {
```
neuschaefer, line 911: I'm not sure why glibc and MIPS were special-cased here. In theory, that should not be necessary for constants defined in the Linux UAPI

Gelbpunkt: AFAICT this was wrong and it's good that it's being removed here

PR #4726, opened 2025-09-26
linux-musl-s390x: Add __psw_t/fprefset_t/*context_t

neuschaefer:
# Description

Define signal-related structs according to musl-libc. These structs are already mentioned in libc-test/semver/linux-s390x.txt, so it makes sense to define the for musl, not just for glibc.

# Sources

https://elixir.bootlin.com/musl/v1.2.5/source/arch/s390x/bits/signal.h

# Checklist

- [x] ~~Relevant tests in `libc-test/semver` have been updated~~ no change

### DIFF
--- a/src/unix/linux_like/linux/musl/b64/s390x.rs
+++ b/src/unix/linux_like/linux/musl/b64/s390x.rs
@@ -84,6 +84,31 @@ s! {
         pub f_flags: c_uint,
         pub f_spare: [c_uint; 4],
     }
+
+    pub struct __psw_t {
+        pub mask: c_ulong,
+        pub addr: c_ulong,
+    }
+
+    pub struct fpregset_t {
+        pub fpc: c_uint,
+        pub fprs: [fpreg_t; 16],
+    }
+
+    pub struct mcontext_t {
+        pub psw: __psw_t,
+        pub gregs: [c_ulong; 16],
+        pub aregs: [c_uint; 16],
+        pub fpregs: fpregset_t,
+    }
+
+    pub struct ucontext_t {
+        pub uc_flags: c_ulong,
+        pub uc_link: *mut ucontext_t,
+        pub uc_stack: crate::stack_t,
+        pub uc_mcontext: mcontext_t,
+        pub uc_sigmask: crate::sigset_t,
+    }
 }
 
 s_no_extra_traits! {


src/unix/linux_like/linux/musl/b64/s390x.rs
```diff
@@ -84,6 +84,32 @@ s! {
...
     }
+
+    pub struct __psw_t {
+        pub mask: u64,
+        pub addr: u64,
+    }
+
+    pub struct fpregset_t {
+        pub fpc: u32,
+        __pad: u32,
```
Gelbpunkt: I don't see this padding in musl sources?

src/unix/linux_like/linux/musl/b64/s390x.rs
```diff
@@ -84,6 +84,32 @@ s! {
         pub f_flags: c_uint,
         pub f_spare: [c_uint; 4],
     }
+
+    pub struct __psw_t {
+        pub mask: u64,
+        pub addr: u64,
```
Gelbpunkt: Use `c_ulong` here

src/unix/linux_like/linux/musl/b64/s390x.rs
```diff
@@ -84,6 +84,32 @@ s! {
...
+    }
+
+    pub struct fpregset_t {
+        pub fpc: u32,
```
Gelbpunkt: `c_uint`

src/unix/linux_like/linux/musl/b64/s390x.rs
```diff
@@ -84,6 +84,32 @@ s! {
...
+        __pad: u32,
+        pub fprs: [fpreg_t; 16],
+    }
+
+    pub struct mcontext_t {
+        pub psw: __psw_t,
+        pub gregs: [u64; 16],
```
Gelbpunkt: `c_ulong`

src/unix/linux_like/linux/musl/b64/s390x.rs
```diff
@@ -84,6 +84,32 @@ s! {
...
+        pub aregs: [u32; 16],
```
Gelbpunkt: `c_uint`

src/unix/linux_like/linux/musl/b64/s390x.rs
neuschaefer: It isn't explicit, but the compiler ends up adding it anyway. Many other structs in `libc` are defined with explicit padding. I don't know the official policy (either way it should ideally be enforced in CI)

neuschaefer: removed for now.

Gelbpunkt: As far as I can tell, padding fields are only used when they are also defined in the libc headers

PR #4728, opened 2025-09-28
redox: more sysconf constants

willnode:
# Description

This PR is a part of effort to introduce uutils `nproc` which requires constants `_SC_NPROCESSORS_CONF` and `_SC_NPROCESSORS_ONLN`. I see that there's other sysconf constants that's not updated here so I add them too.

# Sources

+ [Relibc constants](https://gitlab.redox-os.org/redox-os/relibc/-/blob/master/src/header/unistd/sysconf/redox.rs#L14)
+ [merged `SC_NPROCESSORS_*` implementation](https://gitlab.redox-os.org/redox-os/relibc/-/merge_requests/709)

# Checklist


jackpot51 (approved):

libc-test/semver/redox.txt
```diff
@@ -251,8 +251,27 @@ _PC_SOCK_MAXBUF
 _PC_SYMLINK_MAX
 _PC_SYNC_IO
 _POSIX_VDISABLE
+_SC_ARG_MAX
```
JohnTitor, line 254: Thank you for keeping it sync!

PR #4734, opened 2025-10-01
add HWCAP consts

brad0:
# Description

add HWCAP consts

# Sources

https://github.com/bminor/glibc/blob/3ab9b88e2ac91062b6d493fe32bd101a55006c6a/elf/elf.h
https://github.com/freebsd/freebsd-src/blob/85007872d1227006adf2ce119fe30de856cbe12d/sys/sys/elf_common.h


brad0:
At first I thought this was failing in Rust code but then looking at the build logs it is in C code. The AT_HWCAP3/4 consts wouldn't be in the auxv.h header with older glibc I'm guessing is on that Linux system and FreeBSD 13 / 14 wouldn't have this either.

PR #4737, opened 2025-10-03
Fix the import path for size_of and align_of

r-chiba:
# Description

This change fixes to import `size_of` and `align_of` from the core crate.

We have seen an error failing to resolve this import directive.

# Sources

# Checklist

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -78,6 +78,8 @@ macro_rules! prelude {
             #[allow(unused_imports)]
             pub(crate) use core::marker::{Copy, Send, Sync};
             #[allow(unused_imports)]
+            pub(crate) use core::mem::{align_of, align_of_val, size_of, size_of_val};
+            #[allow(unused_imports)]
             pub(crate) use core::option::Option;
             #[allow(unused_imports)]
             pub(crate) use core::prelude::v1::derive;
@@ -86,8 +88,6 @@ macro_rules! prelude {
 
             #[allow(unused_imports)]
             pub(crate) use fmt::Debug;
-            #[allow(unused_imports)]
-            pub(crate) use mem::{align_of, align_of_val, size_of, size_of_val};
 
             #[allow(unused_imports)]
             pub(crate) use crate::types::{CEnumRepr, Padding};


r-chiba:
I think the rustc version is 1.88.0 but I couldn't find the exact command used for the build.

The context is that the error happened to one of the components of (the internal version of) Android's AOSP project when I updated the libc crate version used for Android build (internal version of https://cs.android.com/android/platform/superproject/main/+/main:external/rust/android-crates-io/) from 0.2.172 to 0.2.175.

I'm so sorry but I'm not a developer of the component failed to build and I have no idea how the component is built. And unfortunately I don't have a minimal reproducible case of the failure.

tgross35:
Sorry for the delay, just got a chance to circle back here. I think the problem is indeed that you are somehow building with the old edition. If I invoke rustc directly with `rustc src/lib.rs --crate-type=lib` to use the default 2015 edition, I get:

```text
error[E0432]: unresolved import `fmt`                                                                                                                                                                                  --> src/macros.rs:88:28                                                                                                                                                                                               |                                                                                                                                                                                                                 88 |             pub(crate) use fmt::Debug;                                                                                                                                                                             |                            ^^^ help: a similar path exists: `self::fmt`                                                                                                                                            |                                                                                                                                                                                                                   ::: src/lib.rs:96:9
   |
96 |         prelude!();
   |         ---------- in this macro invocation
   |
   = note: this error originates in the macro `prelude` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0432]: unresolved import `mem`
  --> src/macros.rs:90:28
   |
90 |             pub(crate) use mem::{align_of, align_of_val, size_of, size_of_val};
   |                            ^^^ help: a similar path exists: `self::mem`
   |
  ::: src/lib.rs:96:9
   |
96 |         prelude!();
   |         ---------- in this macro invocation
   |
   = note: this error originates in the macro `prelude` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0659]: `Debug` is ambiguous
  --> src/unix/mod.rs:41:14
   |
41 |     #[derive(Debug)]
   |              ^^^^^ ambiguous name
   |
   = note: ambiguous because of a conflict between a name from a glob import and an outer scope during import or macro resolution
note: `Debug` could refer to the unresolved item imported here
  --> src/unix/mod.rs:6:5
   |
 6 | use crate::prelude::*;
   |     ^^^^^^^^^^^^^^^^^
```

Which seems to be what you have plus a few fmt-related errors (I assume you may be testing before a6e75638ce4e). Adding the ` --edition=2021` flag to the `rustc` invocation fixes things.

The edition is configured to 2021 [here](https://github.com/rust-lang/libc/blob/2bf1f66ea4a4c0061bed58efe6fc6d5891a8cb0a/Cargo.toml#L10), which gets set automatically if you build with Cargo. So my guess is that the build system you're using is just missing this step, and it happened to work until now.

As this is Android-related, @maurer may know some more

PR #4738, opened 2025-10-06
Define _CS_PATH on the BSDs

krobelus:
According to 
https://man.netbsd.org/confstr.3

_CS_PATH is obsoleted by sysctl (which has a USER_CS_PATH equivalent),
but Linux doesn't have that.

So the simplest thing for applications is to use _CS_PATH which is
part of POSIX. Define, matching the <unistd.h> header.

We could maybe share this definition in src/unix/bsd/netbsdlike/mod.rs,
but I saw that existing definitions are not shared either, so I'm
not sure.

	$ grep src/unix/bsd/netbsdlike -e _PC_LINK_MAX
	src/unix/bsd/netbsdlike/netbsd/mod.rs:1599:11:pub const _PC_LINK_MAX: c_int = 1;
	src/unix/bsd/netbsdlike/openbsd/mod.rs:1193:11:pub const _PC_LINK_MAX: c_int = 1;

Originally reported in https://github.com/fish-shell/fish-shell/issues/11892


semarie (changes requested):
- LGTM for OpenBSD part
- but you added `_CS_PATH` in netbsd.txt and I don't see corresponding mod.rs edited

JohnTitor (approved):
Thank you!

PR #4747, opened 2025-10-14
Add more wasi definitions for libstd

alexcrichton:
This commit fills out definitions in libc for rust-lang/rust#147572 notably filling out some fs-related functions as well as many pthread-related functions. The pthread-related functions were not available originally with wasi-libc but nowadays are stubs for single-threaded behavior. The goal is to make wasi targets more "unix like" in libstd and have less WASI-specific code.

All sources for these definitions are taken from https://github.com/WebAssembly/wasi-libc/

### DIFF
--- a/src/wasi/mod.rs
+++ b/src/wasi/mod.rs
@@ -175,9 +175,8 @@ s! {
         __fds: [c_int; FD_SETSIZE as usize],
     }
 
-    #[repr(align(4))]
     pub struct pthread_attr_t {
-        size: [u8; 36],
+        size: [c_long; 9],
     }
 
     pub struct pthread_mutexattr_t {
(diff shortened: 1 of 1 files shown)


src/wasi/mod.rs
```diff
@@ -170,6 +174,47 @@ s! {
         __nfds: usize,
         __fds: [c_int; FD_SETSIZE as usize],
     }
+
+    #[repr(align(4))]
+    pub struct pthread_attr_t {
+        size: [u8; 36],
+    }
```
tgross35: Is 64-bit wasi something we'll need to support? If so, maybe `[c_ulong; if size_of::<c_long>() == 8 { 7 } else { 9 }]` to account for that, and it's a bit easier to match up to the source anyway (assuming this is the right definition https://github.com/WebAssembly/wasi-libc/blob/6b45da5b05bc0edda355a6de46101d4b21631985/libc-top-half/musl/include/alltypes.h.in#L108). Same kind of thing for the others from that block.

alexcrichton: Eventually, yes. For now though I don't think it's worth it. Currently there is no 64-bit WASI target (https://github.com/WebAssembly/component-model/issues/22) so there's nothing to test against. I'm happy, however, to come around later and fix this once 64-bit targets do exist though!

alexcrichton: I like the idea of using a more representative type internally, though, so I've switched to that.

PR #4750, opened 2025-10-17
Declare field 'tv_nsec' of structure 'timespec' as 'i32' for AIX

xingxue-ibm:
On AIX, <sys/stat.h> declares field `tv_nsec` of `struct timespec` as `long`, but the underlying system calls return a 4-byte value in both 32-bit and 64-bit modes. This PR changes to declare `tv_nsec` as `i32` in the `libc` crate to avoid including the other undefined 4 bytes in the 64-bit mode.

# Description

# Sources

# Checklist

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -68,7 +68,7 @@ s! {
 
     // linux x32 compatibility
     // See https://sourceware.org/bugzilla/show_bug.cgi?id=16437
-    #[cfg(not(target_env = "gnu"))]
+    #[cfg(all(not(target_env = "gnu"), not(target_os = "aix")))]
     pub struct timespec {
         pub tv_sec: time_t,
         #[cfg(all(target_arch = "x86_64", target_pointer_width = "32"))]
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -5587,6 +5587,12 @@ fn test_aix(target: &str) {
             // The field 'data' is actually a unnamed union in the AIX header.
             "pollfd_ext" if field.ident() == "data" => true,
 
+            // On AIX, <stat.h> declares 'tv_nsec' as 'long', but the
+            // underlying system calls return a 32-bit value in both 32-bit
+            // and 64-bit modes. In the 'libc' crate it is declared as 'i32'
+            // to match the system call. Skip this field.
+            "timespec" if field.ident() == "tv_nsec" => true,
+
             _ => false,
         }
     });
--- a/src/unix/aix/mod.rs
+++ b/src/unix/aix/mod.rs
@@ -264,6 +264,15 @@ s! {
         pub tv_nsec: c_int,
     }
 
+    // On AIX, <stat.h> declares 'tv_nsec' as 'long', but the underlying
+    // system calls return a 4-byte value in both 32-bit and 64-bit modes.
+    // It is declared as 'c_int' here to avoid using the other undefined 4
+    // bytes in the 64-bit mode.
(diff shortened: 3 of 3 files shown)


src/unix/mod.rs
```diff
@@ -68,12 +68,28 @@ s! {
 
     // linux x32 compatibility
     // See https://sourceware.org/bugzilla/show_bug.cgi?id=16437
+    // On AIX, <stat.h> declares 'tv_nsec' as 'long', but the underlying
+    // system calls return a 4-byte value in both 32-bit and 64-bit modes.
+    // It is declared as 'i32' to avoid using the other undefined 4 bytes
+    // in the 64-bit mode.
     #[cfg(not(target_env = "gnu"))]
     pub struct timespec {
```
tgross35: To make things a bit cleaner, maybe exclude `target_os = "aix"` here? Then put the struct in `src/unix/aix/mod.rs`

tgross35 (changes requested):

tgross35:
> On AIX, <sys/stat.h> declares field `tv_nsec` of `struct timespec` as `long`, but the underlying system calls return a 4-byte value in both 32-bit and 64-bit modes. This PR changes to declare `tv_nsec` as `i32` in the `libc` crate to avoid including the other undefined 4 bytes in the 64-bit mode.

I think this is reasonable to change on our side, but isn't it a problem for C users as well if the header doesn't match the syscalls? Technically posix specifies `tv_nsec` as `long` too https://pubs.opengroup.org/onlinepubs/009695399/basedefs/time.h.html so I'm surprised the syscalls differ (X32 is a weird exception here)

xingxue-ibm:
> > On AIX, <sys/stat.h> declares field `tv_nsec` of `struct timespec` as `long`, but the underlying system calls return a 4-byte value in both 32-bit and 64-bit modes. This PR changes to declare `tv_nsec` as `i32` in the `libc` crate to avoid including the other undefined 4 bytes in the 64-bit mode.
> 
> I think this is reasonable to change on our side, but isn't it a problem for C users as well if the header doesn't match the syscalls? Technically posix specifies `tv_nsec` as `long` too https://pubs.opengroup.org/onlinepubs/009695399/basedefs/time.h.html so I'm surprised the syscalls differ (X32 is a weird exception here)

The problem can be reproduced with a C test case as well. AIX should have mapped `stat` and its variants to a new set of ABIs that return `tv_nsec` as a `long` when compiling for POSIX code, but it appears that this was not done. However, AIX provides something such as the `statx` system call that accepts two additional arguments, including `Command`. The caller can specify the `STX_TIMESPEC` flag to use `struct timespec` instead of `st_timespec`. In 64-bit mode, this flag causes `statx` to treat `tv_nsec` as an 8-byte value. However, the actual value returned in `tv_nsec` is still only 4 bytes, with the upper 4 bytes cleared, no additional precision is provided.

Exposing `statx` and the other `stat` variants with the `x` suffix, along with the full set of flags for the `Command` flag definitions, and implementing Rust’s `stat` functions on top of them would be quite complex. Moreover, doing so would not provide any real benefit, since no extra precision is gained. On the other hand, the `libc` crate already contains code to handle a similar situation on x86, where `tv_nsec` is always 8 bytes in both 32-bit and 64-bit modes. After discussion internally, we decided it would be cleaner to the same for AIX.

src/unix/mod.rs
xingxue-ibm: Changed as suggested, thanks!

PR #4751, opened 2025-10-19
Add LINK_STATE_* definitions to NetBSD and OpenBSD.

pusateri:
# Description

Add missing interface LINK_STATE_* definitions from sys/net/if.h

# Sources

https://github.com/NetBSD/src/blob/54613ac4db31045d0ca58a7e2da7af005c6b5cd2/sys/net/if.h#L207
https://github.com/openbsd/src/blob/bac4ed8757bbe68d4cac117487958588ae27e7c2/sys/net/if.h#L135

# Checklist

- [ x] Relevant tests in `libc-test/semver` have been updated
- [x ] No placeholder or unstable values like `*LAST` or `*MAX` are
- [ x] Tested locally (`cd libc-test && cargo test --target mytarget`);

### DIFF
--- a/src/new/netbsd/net/if_.rs
+++ b/src/new/netbsd/net/if_.rs
@@ -29,6 +29,10 @@ s! {
     }
 }
 
+pub const LINK_STATE_UNKNOWN: c_int = 0; // link invalid/unknown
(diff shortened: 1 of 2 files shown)


tgross35 (approved):
LGTM but please squash

PR #4755, opened 2025-10-21
test: ignore some consts on windows-gnu

JohnTitor:
# Description

# Sources

# Checklist


tgross35:
(2c28eacc031e4 was the backport here)

PR #4769, opened 2025-10-24
Add missing definitions for QNX8 for several BPF-related structures and constants

rafaeling:
# Description

These structures are required for working with BPF filters and network interfaces on the QNX8 platform. Their absence currently prevents building or using BPF-related functionality with libc on QNX8 targets.

QNX Network Stack Versioning:

- legacy `io-pkt` stack:
> target_env = "nto70": QNX 7.0
> target_env` = "nto71": Standard QNX 7.1 (default legacy stack)

- modern `io-sock` stack:
> target_env = "nto71_iosock": QNX 7.1 with the optional new stack
> target_env = "nto80": QNX 8.0

# Sources
Updated `src/new/*` accordingly

# Checklist

- [?] Relevant tests in `libc-test/semver` have been updated
       **-> No tests for QNX?**
- [*] No placeholder or unstable values like `*LAST` or `*MAX` are
- [?] Tested locally (`cd libc-test && cargo test --target mytarget`);  especially relevant for platforms that may not be checked in CI.
       **-> even main branch does not work.
       Error: could not execute process `/home/user/libc/target/x86_64-pc-nto-qnx800/debug/deps/libc-67e46525c5118905` (never executed)**

### DIFF
--- a/src/new/nto/net/bpf.rs
+++ b/src/new/nto/net/bpf.rs
@@ -1,0 +1,83 @@
+use crate::bpf_insn;
+
+pub const BPF_LD: u16 = 0x00;
+pub const BPF_LDX: u16 = 0x01;
+pub const BPF_ST: u16 = 0x02;
+pub const BPF_STX: u16 = 0x03;
+pub const BPF_ALU: u16 = 0x04;
+pub const BPF_JMP: u16 = 0x05;
+pub const BPF_RET: u16 = 0x06;
+pub const BPF_MISC: u16 = 0x07;
+pub const BPF_W: u16 = 0x00;
+pub const BPF_H: u16 = 0x08;
+pub const BPF_B: u16 = 0x10;
+pub const BPF_IMM: u16 = 0x00;
+pub const BPF_ABS: u16 = 0x20;
+pub const BPF_IND: u16 = 0x40;
+pub const BPF_MEM: u16 = 0x60;
+pub const BPF_LEN: u16 = 0x80;
+pub const BPF_MSH: u16 = 0xa0;
+pub const BPF_ADD: u16 = 0x00;
+pub const BPF_SUB: u16 = 0x10;
+pub const BPF_MUL: u16 = 0x20;
+pub const BPF_DIV: u16 = 0x30;
+pub const BPF_OR: u16 = 0x40;
+pub const BPF_AND: u16 = 0x50;
+pub const BPF_LSH: u16 = 0x60;
+pub const BPF_RSH: u16 = 0x70;
+pub const BPF_NEG: u16 = 0x80;
+pub const BPF_MOD: u16 = 0x90;
+pub const BPF_XOR: u16 = 0xa0;
+pub const BPF_JA: u16 = 0x00;
+pub const BPF_JEQ: u16 = 0x10;
+pub const BPF_JGT: u16 = 0x20;
+pub const BPF_JGE: u16 = 0x30;
+pub const BPF_JSET: u16 = 0x40;
+pub const BPF_K: u16 = 0x00;
+pub const BPF_X: u16 = 0x08;
+pub const BPF_A: u16 = 0x10;
+pub const BPF_TAX: u16 = 0x00;
+pub const BPF_TXA: u16 = 0x80;
+
+f! {
+    pub fn BPF_CLASS(code: u32) -> u32 {
+        code & 0x07
+    }
+
+    pub fn BPF_SIZE(code: u32) -> u32 {
+        code & 0x18
+    }
+
+    pub fn BPF_MODE(code: u32) -> u32 {
+        code & 0xe0
+    }
+
+    pub fn BPF_OP(code: u32) -> u32 {
+        code & 0xf0
+    }
+
+    pub fn BPF_SRC(code: u32) -> u32 {
+        code & 0x08
+    }
+
+    pub fn BPF_RVAL(code: u32) -> u32 {
+        code & 0x18
+    }
+
+    pub fn BPF_MISCOP(code: u32) -> u32 {
+        code & 0xf8
+    }
+
+    pub fn BPF_STMT(code: u16, k: u32) -> bpf_insn {
+        bpf_insn {
+            code,
+            jt: 0,
+            jf: 0,
+            k,
+        }
+    }
+
+    pub fn BPF_JUMP(code: u16, k: u32, jt: u8, jf: u8) -> bpf_insn {
+        bpf_insn { code, jt, jf, k }
+    }
+}
--- a/src/new/nto/net/if_.rs
+++ b/src/new/nto/net/if_.rs
@@ -1,0 +1,32 @@
+use crate::prelude::*;
+
+s_no_extra_traits! {
+    pub union __c_anonymous_ifr_ifru {
+        pub ifru_addr: crate::sockaddr,
+        pub ifru_dstaddr: crate::sockaddr,
+        pub ifru_broadaddr: crate::sockaddr,
+        pub ifru_buffer: ifreq_buffer,
+        pub ifru_flags: [c_short; 2],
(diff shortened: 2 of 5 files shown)


src/unix/nto/mod.rs
```diff
@@ -794,6 +798,35 @@ s_no_extra_traits! {
...
+    pub struct ifreq {
+        /// if name, e.g. "en0"
+        pub ifr_name: [c_char; crate::IFNAMSIZ],
+        pub ifr_ifru: __c_anonymous_ifr_ifru,
+    }
+
+    pub union __c_anonymous_ifc_ifcu {
+        pub ifcu_buf: *mut c_char,
+        pub ifcu_req: *mut ifreq,
+    }
```
tgross35: Just noticed something: is this union used at all?

tgross35 (changes requested):
Would you actually mind moving this to the `src/new` module? We're slowly starting to migrate things there since it's easier to match up with what a user can `#include`. I believe this would be `src/new/nto/net/if.h` and `src/new/nto/net/bpf.h`, unless nto has a different structure. You can either move the existing BPF items over or not, your call.

Then this update to `src/new/nto/mod.rs`:

```rust
pub(crate) mod net {
    pub(crate) mod route;
}
```

And at https://github.com/rust-lang/libc/blob/564e0efe817e8d2e147c21fe2227c2d78eff44ff/src/new/mod.rs#L171-L186, add a nto arm and re-export everything from the two modules.

After that and the `__c_anonymous_ifc_ifcu` removal I'll merge this given we haven't heard back from the maintainers but this is straightforward.

src/unix/nto/mod.rs
rafaeling: Actually not, thanks!! I think I used as reference the linux_like definition, will remove it,

src/unix/nto/mod.rs
```diff
@@ -877,7 +877,7 @@ pub const MS_SYNC: c_int = 2;
 pub const SCM_RIGHTS: c_int = 0x01;
 pub const SCM_TIMESTAMP: c_int = 0x02;
 cfg_if! {
-    if #[cfg(not(target_env = "nto71_iosock"))] {
+    if #[cfg(not(any(target_env = "nto71_iosock", target_env = "nto80")))] {
```
tgross35: Are these not available on 8.0 either?

If this change is related to making tests pass on 8.0 rather than as part of the BPF, would you be able to put it in a separate commit?

src/new/nto/net/bpf.rs
```diff
@@ -0,0 +1,82 @@
...
+        code,
+        jt: 0,
+        jf: 0,
+        k,
+    }
+}
+
+pub fn BPF_JUMP(code: u16, k: u32, jt: u8, jf: u8) -> bpf_insn {
+    bpf_insn { code, jt, jf, k }
+}
```
tgross35: This file and `if_.rs` seem to be missing the final `\n`

tgross35: These functions are `extern "C"` in Linux (via the `f!` macro), they should probably be the same here.

src/unix/nto/mod.rs
rafaeling: No, these are not available on 8.0, so this is part of the BPF topic.

tgross35: I still don't follow the connection - why are sysctl things like `HW_IOSTATS` related to BPF?

src/new/nto/net/if_.rs
```diff
@@ -0,0 +1,33 @@
+use crate::prelude::*;
+
+s! {
+    pub struct ifreq {
+        /// if name, e.g. "en0"
+        pub ifr_name: [c_char; crate::IFNAMSIZ],
+        pub ifr_ifru: __c_anonymous_ifr_ifru,
+    }
```
tgross35: This probably has to go in `s_no_extra_traits` since the union doesn't have them (you can check that things work with `--feature extra_traits`)

rafaeling: I moved the structs into `s_no_extra_traits` (as linux does) but building with `--features extra_traits`still fails because the union would need `Eq`, `Hash`, and `PartialEq`.

On the BSD targets (Apple/FreeBSD/OpenBSD) these unions do implement `Eq`, `Hash`, and `PartialEq` and `struct ifreq` is into `s!` macro, should I mirror that behavior here as well? (I tried it and also fails with same error, not sure how to proceed)

src/unix/nto/mod.rs
rafaeling: Sorry for the confusion.

I was just referring specifically to the `BIOC*` constants needed for BPF. I initially assumed that all of the constants in that section were BPF-related, but that turns out not to be the case. After checking the actual BIOC* values used for BPF on QNX 8, I added the new flag because the build was incorrectly picking up the constants from the else branch (intended for QNX 7.1/7.0 I think).

There are many other constants defined there that are unrelated to BPF. My assumption is that `nto71_iosock` and `nto80` share the same constant layout, but I dont work for QNX, so I had to assume a few things.

src/new/nto/net/if_.rs
tgross35: What exactly is failing? If `__c_anonymous_ifr_ifru` and `ifreq` are both in `s_no_extra_traits`, that macro shouldn't try to derive them.

Don't handwrite trait impls, we do have them in some places but we're trying to move away from that because of the easy accidental unsoundness with unions.

rafaeling: Dont know why, but retried again and works now, maybe did something wrong before.
Check last commit

src/unix/nto/mod.rs
tgross35: That makes more sense to me, but I'd like the maintainers to double check since it's a bit surprising.

@flba-eb @gh-tr @jonathanpallant @japaric could you review and/or test the changes to these constants on QNX 8?

src/new/nto/net/if_.rs
tgross35: LGTM! I'm going to need to wait for somebody to look at https://github.com/rust-lang/libc/pull/4769#discussion_r2541770571 though.

This will need a squash before merge.

src/unix/nto/mod.rs
flba-eb: This looks okay, but I suggest to inverse the logic:
```suggestion
    if #[cfg(any(target_env = "nto70", target_env = "nto71"))] {
```

1. Positive logic is easier to understand
2. It is more likely that the list of QNX versions using the older network stack will not need to be modified in the future, but the versions with the new stack will probably need to be adjusted with each new QNX version.

tgross35: Are these constants all only for the old network stack then? A comment would be nice then if you could add one @rafaeling.

rafaeling: @tgross35 updated and comment added, once you review it, I will make the squash

src/unix/nto/mod.rs
```diff
@@ -876,8 +876,18 @@ pub const MS_SYNC: c_int = 2;
...
+// The `if` block targets the legacy `io-pkt` stack.
+// - target_env = "nto70": QNX 7.0
+// - target_env = "nto71": Standard QNX 7.1 (default legacy stack)
+//
+// The `else` block targets the modern `io-sock` stack.
+// - target_env = "nto71_iosock": QNX 7.1 with the optional new stack
+// - target_env = "nto80": QNX 8.0
 cfg_if! {
-    if #[cfg(not(target_env = "nto71_iosock"))] {
+    if #[cfg(any(target_env = "nto70", target_env = "nto71"))]
```
tgross35: This is missing the opening `{`

rafaeling: Thanks!! fixed and squashed

PR #4771, opened 2025-10-25
reorg: Fill in the `src/new` structure

tgross35:
Ensure each library we provide bindings for has a home module in order
to make it easier to fill things in. Where possible, I included a link
to the source tree.

There is also a new module `common` that contains definitions intended
to be shared across multiple libraries.


tgross35:
Todo: triagebot

PR #4779, opened 2025-10-26
VxWorks: Add missing defines/functions needed by rust stdlib

hax0kartik:
# Description

This adds a few defines/functions which are needed to build rust stdlib. (https://github.com/rust-lang/rust/issues/148125)

# Sources

# Checklist


tgross35 (approved):
Thanks!

PR #4781, opened 2025-10-26
Adding defines and functions required for vxworks

gautam899:
# Description

Adding functions and defines required for vxworks. This supersedes rust-lang/libc#4676 . 
In this PR we have guarded the struct stat, and sigaction in such a way that it works for older (< 25.09) as well as newer versions of vxworks.

## Authors
- @gautam899 
- @mirimmad 

# Sources

# Checklist

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -281,8 +281,10 @@ fn emcc_version_code() -> Option<u64> {
     Some(major * 10000 + minor * 100 + patch)
 }
 
+/// Retrieve the VxWorks release version from the environment variable set by the VxWorks build
+/// environment, in `(minor, patch)` form. Currently the only major version supported by Rust
+/// is 7.
 fn vxworks_version_code() -> Option<(u32, u32)> {
-    // Retrieve the VxWorks release version from the environment variable set by the VxWorks build environment
     let version = env::var("WIND_RELEASE_ID").ok()?;
 
     let mut pieces = version.trim().split(['.']);


build.rs
```diff
@@ -274,6 +281,17 @@ fn emcc_version_code() -> Option<u64> {
...
+fn vxworks_version_code() -> Option<u64> {
+    let version = env::var("WIND_RELEASE_ID").ok()?;
+
+    let mut pieces = version.trim().split(['.']);
+
+    let major: u64 = pieces.next().and_then(|x| x.parse().ok()).unwrap_or(0);
+    let minor: u64 = pieces.next().and_then(|x| x.parse().ok()).unwrap_or(0);
+
+    Some(major * 100 + minor)
+}
```
tgross35: To make things a bit easier, just return `Option<(u32, u32)>` rather than squishing. The comparison can use `v < (25, 9)`

build.rs
```diff
@@ -85,6 +86,12 @@ fn main() {
         _ => (),
     }
 
+    match vxworks_version_code() {
+        Some(v) if (v < 2509) => set_cfg("vxworks_version_old"),
+        // VxWorks version >= 25.09
```
tgross35: It would be good to be specific about the version in the cfg, e.g. `vxworks_24` or `vxworks_lt_25_09`

build.rs
```diff
@@ -274,6 +281,17 @@ fn emcc_version_code() -> Option<u64> {
     Some(major * 10000 + minor * 100 + patch)
 }
 
+fn vxworks_version_code() -> Option<u64> {
+    let version = env::var("WIND_RELEASE_ID").ok()?;
```
tgross35, line 288: Could you add a note about where this env comes from? Is it usually set by vxworks builds, or is this invented just for the purposes here?

src/vxworks/mod.rs
tgross35, line 1: Are all of the additions here available on both vxworks versions?

libc-test/build.rs
```diff
@@ -3333,13 +3333,32 @@ fn test_neutrino(target: &str) {
     ctest::generate_test(&mut cfg, "../src/lib.rs", "ctest_output.rs").unwrap();
 }
 
+fn which_vxworks() -> Option<u64> {
```
tgross35: Ditto re: `Option<(u32, u32)>`

libc-test/build.rs
```diff
@@ -3393,27 +3420,33 @@ fn test_vxworks(target: &str) {
...
         "SIG_DFL" | "SIG_ERR" | "SIG_IGN"
         // This is not defined in vxWorks
-        | "RTLD_DEFAULT"   => true,
+        | "RTLD_DEFAULT" | "S_ISVTX" | "SIGIO" | "SIGWINCH" | "SIGLOST" | "PRIO_PROCESS"  => true,
         _ => false,
     });
     // FIXME(vxworks)
     cfg.skip_alias(move |ty| match ty.ident() {
         "stat64" | "sighandler_t" | "off64_t" => true,
+        "__uint128" => true,
```
tgross35: This shouldn't be needed anymore, it's now set at https://github.com/gautam899/libc/blob/3f231f486f4a60ca9152e111fd8653596d6d4aeb/libc-test/build.rs#L83

libc-test/build.rs
tgross35, line 1: For all of the added skips, could you add a comment about what's wrong? So we know what it takes to resolve them.

build.rs
tgross35: Did versioning change to be based on years? Everything I can find from a quick look seems to say vxworks 7 is the latest.

If there is >1 version we support, it may be worth mentioning that in https://doc.rust-lang.org/nightly/rustc/platform-support/vxworks.html (rather than just a minimum)

src/vxworks/mod.rs
gautam899: Yes. The changes are tested for both the VxWorks versions.

build.rs
gautam899: In the VxWorks setup, the environment variable WIND_RELEASE_ID is always set as the version of the release.

libc-test/build.rs
gautam899: Made the required changes!

libc-test/build.rs
gautam899: Noted! This is now removed.

build.rs
gautam899: Rust is not supported for versions less that vxworks-7

tgross35 (approved):
LGTM! Could you please squash?

build.rs
tgross35: > Rust is not supported for versions less that vxworks-7

Right: but what's the relationship between 7 and 25.09? Is 25.09 something like a release code for "version 7", or is it the 25.9 the version released after version 7?

(no action needed to the code here, I'm just clarifying)

hax0kartik: Yup, its something like the version code. Or you could think as "7" as the major release and 25.09 as the minor and patch respectively.

tgross35: That makes sense, thanks. When I’m back at a computer I’ll push a commit explaining that (unless one of you beat me to it) because I’m bound to forget it, then merge this

src/vxworks/mod.rs
```diff
@@ -1730,6 +2135,9 @@ extern "C" {
     // unistd.h
     pub fn getppid() -> pid_t;
 
+    // unistd.h
+    pub fn setpgid(pid: pid_t, pgid: pid_t) -> pid_t;
```
tgross35, line 2139: Does vxworks actually deviate from posix here on the return type? https://pubs.opengroup.org/onlinepubs/009604599/functions/setpgid.html

mirimmad: pid_t is an alias for c_int.

tgross35: There is no ABI-level difference, but do the headers actually call for a pid_t return? Seems unusual.

mirimmad: I can confirm that setpgid (in VxWorks) does infact return "pid_t". It seems vxworks here deviates from posix.

Issue #3890, opened 2024-08-29
Add NetBSD to CI


tgross35:
`x86_64-unknown-netbsd` is a tier two target, and it looks like cirrus CI might now have support for it https://github.com/cirruslabs/cirrus-ci-docs/issues/970. We should try to get it in our CI.

Cc @semarie if you are interested in working on this since I know you review the OpenBSD support. I don't think we could add OpenBSD since it is tier 3, but NetBSD would at least get something in the same family tested.

PR #4782, opened 2025-10-27
ci: Add testing for NetBSD

tgross35:
The x86-64 NetBSD target is tier 2, so we can start testing it.

Closes: https://github.com/rust-lang/libc/issues/3890

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -1375,6 +1375,15 @@ fn test_netbsd(target: &str) {
         }
     });
 
+    cfg.skip_fn_ptrcheck(move |func| {
+        match func {
+            // New symbol version present in NetBSD10, but we keep the old versions for NetBSD9
+            // compatibility.
+            "getmntinfo" | "statvfs" | "fstatvfs" | "getvfsstat" | "sigaction" => true,
+            _ => false,
+        }
+    });
+
     cfg.skip_struct_field_type(move |struct_, field| {
         // This is a weird union, don't check the type.
         (struct_.ident() == "ifaddrs" && field.ident() == "ifa_ifu") ||
--- a/src/new/netbsd/sys/timex.rs
+++ b/src/new/netbsd/sys/timex.rs
@@ -89,6 +89,7 @@ s! {
 }
 
 extern "C" {
+    #[link_name = "__ntp_gettime50"]
     pub fn ntp_gettime(buf: *mut ntptimeval) -> c_int;
     pub fn ntp_adjtime(buf: *mut timex) -> c_int;
 }
--- a/src/new/netbsd/utmp_.rs
+++ b/src/new/netbsd/utmp_.rs
@@ -27,6 +27,7 @@ s! {
 extern "C" {
     pub fn utmpname(file: *const c_char) -> c_int;
     pub fn setutent();
+    #[link_name = "__getutent50"]
     pub fn getutent() -> *mut utmp;
     pub fn endutent();
 }
--- a/src/new/netbsd/utmpx_.rs
+++ b/src/new/netbsd/utmpx_.rs
@@ -67,16 +67,26 @@ extern "C" {
     pub fn setutxent();
     pub fn endutxent();
 
+    #[link_name = "__getutxent50"]
     pub fn getutxent() -> *mut utmpx;
+    #[link_name = "__getutxid50"]
     pub fn getutxid(ut: *const utmpx) -> *mut utmpx;
+    #[link_name = "__getutxline50"]
     pub fn getutxline(ut: *const utmpx) -> *mut utmpx;
+    #[link_name = "__pututxline50"]
     pub fn pututxline(ut: *const utmpx) -> *mut utmpx;
 
+    #[link_name = "__updwtmpx50"]
     pub fn updwtmpx(file: *const c_char, ut: *const utmpx) -> c_int;
+    #[link_name = "__getlastlogx50"]
     pub fn getlastlogx(fname: *const c_char, uid: crate::uid_t, ll: *mut lastlogx)
         -> *mut lastlogx;
+
+    #[link_name = "__updlastlogx50"]
     pub fn updlastlogx(fname: *const c_char, uid: crate::uid_t, ll: *mut lastlogx) -> c_int;
+    #[link_name = "__getutmp50"]
     pub fn getutmp(ux: *const utmpx, u: *mut crate::utmp);
+    #[link_name = "__getutmpx50"]
     pub fn getutmpx(u: *const crate::utmp, ux: *mut utmpx);
     pub fn utmpxname(file: *const c_char) -> c_int;
 }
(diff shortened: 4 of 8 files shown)


tgross35:
Also, is https://github.com/NetBSD/src/tree/aedc44b2ed927468d4f7bb37cddc443d9077f1de/include the correct location to find all of NetBSD's libc headers?

tgross35:
> I've asked in our mailing lists for help/pointers.

That's rather kind, thank you for reaching out. Do you have a link to the archive? I'm not sure which list this would be.

I'm probably going to re-read through my patches here and apply the series tomorrowish since there is followup work here. If you get the chance to review before then (or retroactively), it would be welcome. Note that there is a small handful of breaking changes but IMO nothing worth worrying about (either removed upstream or is already currently broken in some way).

@semarie there are a few things that touched OpenBSD here. I verified things still build, but you may wish to double check.

0323pin:
@tgross35 that would be [tech-pkg](https://mail-index.netbsd.org/tech-pkg/tindex.html).

https://mail-index.netbsd.org/tech-pkg/2025/10/29/msg031689.html

devnexen:
> > I don't understand the output, but this seem all to be versioned syscalls, the exported name in NetBSD libc is not the API name, e.g. ntp_gettime is currently called __ntp_gettime50, but there are also two older versions called __ntp_gettime30 and ntp_gettime. In C / C++ we use the preprocessor to make code access the right symbol.
> 
> Thanks for taking a look, that's pretty consistent with other platforms. I just totally missed the renaming in source. (The test output isn't meaningful, just a comparison of function pointers to detect these kind of mismatches.)
> 
> I think everything should be resolved as of the latest push. Not sure why `pkg_add curl` is now suddenly failing in the job, though.

Weird, it could happen if the ```PKG_PATH``` env var is not set. Note that curl on netbsd `may` not support ssl (maybe it had changed since though).

MartinHusemann:
> Weird, it could happen if the `PKG_PATH` env var is not set. Note that curl on netbsd `may` not support ssl (maybe it had changed since though, I recall before I needed to set no ssl certificate check in curl options).

curl should support ssl if the base OS is new enough (NetBSD 10 should do). You could check if /usr/sbin/certctl exists.
If the system has been setup manually the certificates may need to be rehashed, run "certctl rehash" as root to do this (this is only required once after installation, or when the root certs have been updated, usually via OS updates)

riastradh:
FYI, public-facing notes on the library symbol versioning scheme: https://wiki.netbsd.org/symbol_versions/

(The versioningsyscalls(9) man page is mainly about system calls and the user<->kernel interface, rather than library symbols and the userland library/program<->library interface.)

Two of the symbols you're using here now, `__statvfs90` and `__fstatvfs90`, only exist in NetBSD>=10.0, so attempts to use them will fail on NetBSD 9.x, which is not yet EOL, or earlier versions.  (Yes, this is a confusing naming scheme, for hysterical raisins, sorry.)

Not sure exactly how you're using the header files (I see them mentioned in the patch, but I also see structure definitions spelled out in Rust in the patch), but you may want to compile things against the header files of the earliest still-supported version of NetBSD, unless you can conditionalize the users on the NetBSD version too.

Using the older symbols will always work on newer versions of NetBSD, but only with the structures and constants from the older version---e.g., if you call the _symbol_ `fstatvfs` (from <10), but you use 10's `struct statvfs` _layout_ from 10's <sys/statvfs.h> (which declares the C function with `__RENAME(__fstatvfs90)` so users will use the _symbol_ `__fstatvfs90`), the results will be nonsense.  (And using the newer symbols won't work at all on older versions of NetBSD; the run-time loader won't find them.)

> I think everything should be resolved as of the latest push. Not sure why `pkg_add curl` is now suddenly failing in the job, though.

Do you set PKG_PATH or /etc/pkg_install.conf to point at a binary package repository?  This doesn't happen out of the box (many users build their own custom packages, particularly on platforms other than x86 and Arm; NetBSD is not as tightly coupled with pkgsrc as some other systems are with their packages), so if you want to use the TNF-built binary packages, you may want to do:

```
export PKG_PATH="https://cdn.NetBSD.org/pub/pkgsrc/packages/NetBSD/$(uname -p)/$(uname -r|cut -f '1 2' -d.|cut -f 1 -d_)/All"
```

(This---or something similar, with pkgin(1)---is done by the interactive installer sysinst(1), but you're presumably not going through that path.)

tgross35:
Thanks for all the added context here.

> Two of the symbols you're using here now, `__statvfs90` and `__fstatvfs90`, only exist in NetBSD>=10.0, so attempts to use them will fail on NetBSD 9.x, which is not yet EOL, or earlier versions. (Yes, this is a confusing naming scheme, for hysterical raisins, sorry.)

Ah, thanks for noticing! Only these two are a problem right? I'm not sure what release the `sigaction` change was in https://github.com/NetBSD/src/commit/02edde8aa38228d705b8bde8fc62ffc368389357. I'll drop the `link_name` bit for any of the problematic ones here.

> Not sure exactly how you're using the header files (I see them mentioned in the patch, but I also see structure definitions spelled out in Rust in the patch), but you may want to compile things against the header files of the earliest still-supported version of NetBSD, unless you can conditionalize the users on the NetBSD version too.

The Rust definitions in this repo is more or less meant to be checked-in versions of the headers available on the system, so we check things against them. The tests that are run generate a .rs file (using this library) and a .c file (`#include`ing as applicable) that check things like struct size, field alignment, function pointer values, etc. and compare them. It's a bit crude but fairly effective at detecting mismatches.

Usually we test against the latest stable OS since it means we can verify any newer API. It looks like vmactions supports 9.4 though, so I'll run that once things are working again just to double check that the failures are compatible. (If anyone has it ready locally, it's just `cargo t -p libc-test` to run these tests.)

> Using the older symbols will always work on newer versions of NetBSD, but only with the structures and constants from the older version---e.g., if you call the _symbol_ `fstatvfs` (from <10), but you use 10's `struct statvfs` _layout_ from 10's <sys/statvfs.h> (which declares the C function with `__RENAME(__fstatvfs90)` so users will use the _symbol_ `__fstatvfs90`), the results will be nonsense. (And using the newer symbols won't work at all on older versions of NetBSD; the run-time loader won't find them.)

I think all of the changes for versions are forward-compatible right? The `struct statvfs` change is just a field addition. There are some changes here that wouldn't be, but those are just things that were broken on our end.

> Ugh, this may also be a stupid problem on our end---folks are investigating.

PKG_PATH wasn't being set before when things were running (I assume maybe vmactions does that?) so this sounds possible, thanks.

0-wiz-0:
> Also, is https://github.com/NetBSD/src/tree/aedc44b2ed927468d4f7bb37cddc443d9077f1de/include the correct location to find all of NetBSD's libc headers?

I think that's most of them, yes. Most architecture-specific ones are in the architecture or machine dependent directory instead, e.g. https://github.com/NetBSD/src/tree/trunk/sys/arch/amd64/include

Reading from the PR what you want (a full set of headers for a NetBSD release)  I suggest using the release tarballs instead, for example https://cdn.netbsd.org/pub/NetBSD/NetBSD-10.1/amd64/binary/sets/ for the NetBSD 10.1/amd64=x86_64 files.

Hope this helps, and thanks a lot for your work!

PR #4790, opened 2025-10-27
[AIX] Make errno EWOULDBLOCK an alias of EAGAIN

xingxue-ibm:
# Description
Although the POSIX standard requires `errno` values to be distinct, it allows `EWOULDBLOCK` to have the same value as `EAGAIN` (https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/errno.h.html).
```
[EWOULDBLOCK]
    Operation would block (may be the same value as [EAGAIN]).
```

# Sources

# Checklist


xingxue-ibm:
Thank you, @tgross35!

PR #4796, opened 2025-10-28
ctest: Emit the `enum` keyword for enum field tests

tgross35:
Currently the return type for a function is missing the `enum` keyword. Resolve this by using the same logic for struct fields as for translating other paths.


tgross35:
Fyi @mbyx

PR #4798, opened 2025-10-28
ctest: Replace check_same_hex with check_same_bytes

tgross35:
This allows us to print the entire slice on failure rather than only a
single mismatched byte.

A possible future improvement here would be to check if the type is an
integer in our template and call `check_same` instead if so, which would
let us print the decimal value as well.


tgross35:
For the first commit, verified that the volatile typedef does emit the warning in https://github.com/rust-lang/libc/pull/4800

PR #4799, opened 2025-10-28
ctest: Run `all.rs` on all targets

tgross35:


mbyx:
I believe the failure in finding errors in `t2` and failing in `t1` are because the test results in all.rs are not platform agnostic, similar to that sign issue we had a while back in Linux vs. Windows.

Since the tier 1 targets all passed, it could also be due to some strange cross compilation issue as well.

PR #4823, opened 2025-11-06
linux/aarch64: make mcontext_t.__reserved pub

jstarks:
# Description

The `mcontext_t.__reserved` field is documented to contain extra context, such as FP state and the ESR register.

# Sources

* [Linux uapi](https://github.com/torvalds/linux/blob/dc77806cf3b4788d328fddf245e86c5b529f31a2/arch/arm64/include/uapi/asm/sigcontext.h#L36)
* [Glibc](https://github.com/bminor/glibc/blob/975c8c4e22f73fb60996f6bcc2cf1a6f6af70928/sysdeps/unix/sysv/linux/aarch64/sys/ucontext.h#L63)
* [Musl](https://github.com/bminor/musl/blob/1b76ff0767d01df72f692806ee5adee13c67ef88/arch/aarch64/bits/signal.h#L22)
* [Android](https://android.googlesource.com/platform/bionic.git/+/refs/heads/main/libc/kernel/uapi/asm-arm64/asm/sigcontext.h#17)

# Checklist

### DIFF
--- a/src/unix/linux_like/android/b64/aarch64/mod.rs
+++ b/src/unix/linux_like/android/b64/aarch64/mod.rs
@@ -74,7 +74,8 @@ s! {
         pub sp: c_ulonglong,
         pub pc: c_ulonglong,
         pub pstate: c_ulonglong,
-        __reserved: [u64; 512],
+        _padding: u64,
+        pub __reserved: [u8; 4096],
     }
 
     pub struct user_fpsimd_struct {
--- a/src/unix/linux_like/linux/gnu/b64/aarch64/mod.rs
+++ b/src/unix/linux_like/linux/gnu/b64/aarch64/mod.rs
@@ -219,7 +219,8 @@ s! {
         pub sp: c_ulonglong,
         pub pc: c_ulonglong,
         pub pstate: c_ulonglong,
-        __reserved: [u64; 512],
+        _padding: u64,
+        pub __reserved: [u8; 4096],
     }
 
     pub struct user_fpsimd_struct {
--- a/src/unix/linux_like/linux/musl/b64/aarch64/mod.rs
+++ b/src/unix/linux_like/linux/musl/b64/aarch64/mod.rs
@@ -103,7 +103,8 @@ s! {
         pub sp: c_ulong,
         pub pc: c_ulong,
         pub pstate: c_ulong,
-        __reserved: [u64; 512],
+        _padding: u64,
+        pub __reserved: [u8; 4096],
     }
 
     #[repr(align(8))]


tgross35 (changes requested):
(updating status)

jstarks:
> Is there official documentation outside of the uapi comment saying what this field contains? The manpage I found says ([link](https://man7.org/linux/man-pages/man3/getcontext.3.html)):
> 
> > The mcontext_t type is machine-dependent and opaque.
> 
> Given that and the fact that the field is actually named reserved, I am hesitant to expose this. If you actually need these fields and want to opt into whatever sort of instability they have, it would be better to define your own struct and transmute.

The uapi header is the only documentation I am aware of. But uapi documents interfaces that are stable. I certainly agree that the `__reserved` name is a red flag--nonetheless, it's a guaranteed stable part of the Linux ABI and by extension the glibc/musl API. I think they're just stuck with the poor choice of name, which hasn't changed since it was introduced in 2012.

If you search around (<https://github.com/search?q=__reserved+uc_mcontext&type=code>), you can find lots of C/C++ code accessing this field as documented in the uapi definitions. It should be possible to port such code to Rust without manually defining types.

I suppose to really complete that promise, `libc` should also define `_aarch64_ctx` and some of the magic numbers and other types from uapi. I'd be happy to do that as well, if you think it's a good idea. musl and Android define these types in their headers but glibc does not--I would guess glibc users probably include the uapi header directly to get these definitions, but I'm not sure.

maurer:
tl;dr: `__reserved` isn't defined the way you want, nothing is defined on `mcontext_t` (only `sigcontext_t`)

Honestly, I think that making `mcontext_t` *fully* opaque might be the "right way" be in line with spec unless individual `libc`s make claims in their documentation about the usability of `mcontext_t` fields. We'd have a separate `sigcontext_t` struct which exposes just the usable fields, and the fact that in for most libcs, we'd have a backend type of `#[repr(transparent)] pub struct mcontext_t(sigcontext_t)` would make the backing of `mcontext_t` implementation dependent, like it's supposed to be. All that said, I think breaking people to be pedantically right here is not worthwhile.

> The uapi header is the only documentation I am aware of. But uapi documents interfaces that are stable.

The UAPI header also says "(Note: records do not necessarily occur in the order shown here.)", which means that you cannot actually read from that `__reserved` field in a stable way. The interface is stable, sure, but the contents of that field are not - that's why they're marked reserved. As you pointed out at the end of your last comment, you'd need several other structs to actually interpret this correctly.

> it's a guaranteed stable part of the Linux ABI and by extension the glibc/musl API

This is also not true - being part of the Linux ABI doesn't automatically make it part of the libc API. A libc would be within its rights to replace all `mcontext_t`s with the same contents xor'd by `0x42` and just xor it back before it sends it to the kernel, since it's opaque. For a `ucontext_t`, only the fields  `uc_link`, `uc_sigmask`, and `uc_stack` are guaranteed to be something you can meaningfully access.

P.S.

This would be something where a user who was super sure what kernel layout they were using (and so don't care that the layout of the `reserved` section is subject to change) would be happy to have `zerocopy::AsBytes` on this struct, as it'd let them write a safe transmute to their pinky-swear version of the struct.

tgross35:
Thank you for the PR, but I am going to close this. We currently have a number of issues related to fields that aren't intended to be public but are in Rust, so I don't think adding another is a good idea. Implicitly making the struct exhaustive also makes me hesitant (unfortunately we can't yet use `#[non_exhaustive]`).

My recommended solution here is to redefine the types in the libraries you use them, and `transmute` as needed. This doesn't strike me as unreasonable because the data that might be read from these fields is already platform- and arch-specific.

> (even happier if we were allowed to put `zerocopy` derivations on them, but I know we can't have those in `libc` yet)

Backlinking, I think this would be acceptable but it needs a formal proposal and has questions to answer. [#t-libs > libc optional dependency on zerocopy](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/libc.20optional.20dependency.20on.20zerocopy/with/474423715)

jstarks:
> There isn’t really any ambiguity that a field named __reserved isn’t meant to be public.

I am definitely sympathetic to your goal of not exposing fields defined in order to pad out a structure to a certain size or alignment. I have already argued that this field is not such a field, despite the terrible name--the glibc header even has a comment explaining the use of the field, including the fact that it must match the Linux uapi definitions. I'm not sure what else would convince you, short of pulling in the glibc and musl ARM64 maintainers, which I don't intend to do.

> My note about it being platform-specific was that if you already need to define structures for whatever is represented by __reserved’ on any given platform, is it really that much additional work to define ‘mcontext_t` as well?

I agree that I don't actually need this change to do what I want. I've already worked around this in my crate. I was hoping to add this definition to help others in my position in the future. Alas.

Thanks for taking a look.

PR #4830, opened 2025-11-16
Define eventfd on NetBSD

krobelus:
Like FreeBSD, NetBSD supports eventfd, see
https://man.netbsd.org/eventfd.2.
OpenBSD does not AFAICT.

(for some reason my local rustfmt wants to reformat imports)

### DIFF
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -7,6 +7,7 @@ use crate::{
 };
 
 pub type blksize_t = i32;
+pub type eventfd_t = u64;
 pub type fsblkcnt_t = u64;
 pub type fsfilcnt_t = u64;
 pub type idtype_t = c_int;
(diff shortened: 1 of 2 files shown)


src/unix/bsd/netbsdlike/netbsd/mod.rs
```diff
@@ -2197,6 +2203,10 @@ extern "C" {
     pub fn getmntinfo(mntbufp: *mut *mut crate::statvfs, flags: c_int) -> c_int;
     pub fn getvfsstat(buf: *mut crate::statvfs, bufsize: size_t, flags: c_int) -> c_int;
 
+    pub fn eventfd(init: c_uint, flags: c_int) -> c_int;
+    pub fn eventfd_read(fd: c_int, value: *mut eventfd_t) -> c_int;
+    pub fn eventfd_write(fd: c_int, value: eventfd_t) -> c_int;
```
JohnTitor: Any reason that these fn args names don't match with man page?

tgross35:
Please include permalinks to the headers too, manpages don't show the values. In this case: https://github.com/NetBSD/src/blob/d04b0c735abc997743bb3faa74464524cbe7becd/sys/sys/eventfd.h.

To fix the CI failure you'll need to add the relevant header to `libc-test/build.rs`.

Why is the arg name changed in all the other files? That should be dropped, or at least moved to a separate commit if there is a reason.

krobelus:
split commits and updated tests and commit messages.
I'm not sure if this works for 100% of all NetBSD 10 systems out there,
but this was introduced in 2021 so it might be fine.

PR #4837, opened 2025-11-20
Gate `__wasilibc_register_preopened_fd` on WASI

alexcrichton:
This API is only available in WASIp1 and as the libc build for WASIp2 is changing the symbol was recently deleted. This functionality makes sense for WASIp1, but for WASIp2 it's a fundamentally different set of syscalls so this is no longer provided by libc.

# Description

# Sources

# Checklist

### DIFF
--- a/ci/docker/wasm32-wasip1/Dockerfile
+++ b/ci/docker/wasm32-wasip1/Dockerfile
@@ -10,7 +10,7 @@ RUN /wasi.sh
 # library.
 ENV CARGO_TARGET_WASM32_WASIP1_RUNNER=wasmtime \
     CARGO_TARGET_WASM32_WASIP1_LINKER=/opt/wasi-sdk/bin/clang \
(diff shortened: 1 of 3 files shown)


tgross35:
Looks like there is another error:

```text
              2: failed to resolve import `env::__wasi_clock_time_get
```

Do the versions at https://github.com/rust-lang/libc/blob/c4dcfd23f00ec3fe6b7a2a67416084c174222754/ci/wasi.sh#L14-L15 also need to be bumped?

alexcrichton:
Indeed! I'm realizing that such a bump shouldn't be necessary, however, so I've bundled a few changes here:

* To prevent a mismatch between the libc.a that rustc nightly ships, and what's tested here in CI, `-Ctarget-feature=-crt-static` is replaced with `-Clink-self-contained=n`. That should insulate against future problems like this where the wasi-sdk in CI here diverges from the wasi-sdk in rustc.
* I've gone ahead and updated to wasi-sdk-29 as well
* The function in question remains gated as it's still removed in wasi-sdk-29

Issue #4740, opened 2025-10-08
Cirrus CI: nightly freebsd-13 x86_64 is quite flaky


JohnTitor:
That CI env is currently quite fragile and fails at almost time.

The failure looks like ([link](https://cirrus-ci.com/task/6378181797609472)):
```
failures:

---- test_entrypoint_macro stdout ----

thread 'test_entrypoint_macro' (103040) panicked at ctest/tests/basic.rs:60:71:
called `Result::unwrap()` on an `Err` value: MacroExpansion("tests/input/macro.rs", "macro expansion failed with signal: 11 (SIGSEGV) (core dumped): ")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It happens on random tests.
It'd be great if we could avoid it by tweaking CI config or something else.

Related: https://github.com/rust-lang/libc/pull/4641

tgross35:
I don't think there is anything actionable on our end here since the intermediate segfaults seem to be a kernel bug https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=285873. Since FreeBSD 13 is (aiui) approaching EOL, it's probably not worth too much effort trying to work around it - I've just been rerunning that job when it fails on FreeBSD-related PRs.

JohnTitor:
Closing as the job is removed due to EoL.

PR #4838, opened 2025-11-20
ci: Disable tests on FreeBSD 13

tgross35:
These jobs almost always fail with a segfault so unfortunately aren't very useful. There isn't anything we can do to fix them on the libc side, given the issue comes from a kernel bug. The failing jobs have been a source of contributor confusion, so remove them here.

Link: https://github.com/rust-lang/libc/issues/4740
Link: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=285873

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -7,15 +7,16 @@ task:
     # https://github.com/rust-lang/rust/issues/132185
     RUST_BACKTRACE: "0"
   matrix:
-    - name: nightly freebsd-13 i686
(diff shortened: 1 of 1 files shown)


tgross35:
That was what I thought https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=285873 was suggesting, given it only showed up on certain minor versions. Reading through again though, I see you mention it's likely coming from a library.

In any case, I still don't think there is anything we can do on this end.

Issue #4829, opened 2025-11-15
ctest: Generate extern "C" with unsafe


JohnTitor:
ctest currently emits `extern "C"` but edition 2024 rejects unless it has unsafe declaration. It'd be great if we could emit edition 2024 style externs (as an option).

tgross35:
Cc @mbyx if you're interested in adding this

PR #4843, opened 2025-11-21
ctest: allow different code generation according to edition

mbyx:
# Description
Closes rust-lang/libc#4829 

Adds the ability to choose which Rust edition to generate code against.

# Checklist


tgross35 (approved):
Thank you!

PR #4845, opened 2025-11-21
qurt: Add support for Qualcomm QuRT

androm3da:
The QuRT RTOS -- used primarily on the Hexagon architecture -- has support for large parts of POSIX.

# Description

Add support for Hexagon QuRT OS.

# Sources

The corresponding sources from the Hexagon SDK from https://github.com/snapdragon-toolchain/hexagon-sdk or https://softwarecenter.qualcomm.com/catalog/item/Hexagon_SDK

# Checklist

- [x] N/A - Relevant tests in `libc-test/semver` have been updated

### DIFF
--- a/src/new/qurt/mod.rs
+++ b/src/new/qurt/mod.rs
@@ -1,0 +1,163 @@
+//! QuRT (Qualcomm Real-Time OS) bindings
+//!
+//! QuRT is Qualcomm's real-time operating system for Hexagon DSP architectures.
+//! Headers available via the
+//! Hexagon SDK: https://softwarecenter.qualcomm.com/catalog/item/Hexagon_SDK
+
+use crate::prelude::*;
+
+// Basic C types for QRT
+pub type intptr_t = isize;
+pub type uintptr_t = usize;
+pub type ptrdiff_t = isize;
+pub type size_t = uintptr_t;
+pub type ssize_t = intptr_t;
+
+// Process and system types
+pub type pid_t = c_int;
+pub type uid_t = c_uint;
+pub type gid_t = c_uint;
+
+// Time types
+pub type time_t = c_longlong;
+pub type suseconds_t = c_long;
+pub type useconds_t = c_ulong;
+pub type clockid_t = c_int;
+pub type timer_t = *mut c_void;
+
+// File system types
+pub type dev_t = c_ulong;
+pub type ino_t = c_ulong;
+pub type mode_t = c_uint;
+pub type nlink_t = c_uint;
+pub type off_t = c_long;
+pub type blkcnt_t = c_long;
+pub type blksize_t = c_long;
+
+// Thread types based on QuRT pthread implementation
+pub type pthread_t = c_uint;
+pub type pthread_key_t = c_int;
+pub type pthread_once_t = c_int;
+pub type pthread_mutex_t = c_uint;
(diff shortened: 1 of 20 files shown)


src/new/qurt/mod.rs
```diff
@@ -0,0 +1,15 @@
...
+pub(crate) mod fcntl;
+pub(crate) mod limits;
+pub(crate) mod pthread;
+pub(crate) mod semaphore;
+pub(crate) mod signal;
+pub(crate) mod stdio;
+pub(crate) mod stdlib;
+pub(crate) mod sys;
+pub(crate) mod time;
+pub(crate) mod unistd;
```
tgross35, line 163: I assume this needs to be filled in right? Or deleted.

I'd appreciate if you could at least make unistd.rs and pthread.rs in `src/new` roughly matching their `.h` files, since I've ported most other platforms at this point. Everything else is up to you; you can either add the modules like you have here (eventually everything will hopefully be like this) or just put it all in `src/qurt` like most other platforms (and I'll move them over later).

src/new/qurt/mod.rs
```diff
@@ -0,0 +1,15 @@
+//! QuRT (Qualcomm Real-Time OS) bindings
+//!
+//! QuRT is Qualcomm's real-time operating system for Hexagon DSP architectures.
```
tgross35, line 3: Mind adding a link to the SDK sources?

androm3da: Here, in the PR comments or in the comments of `src/new/qurt/mod.rs`?

tgross35: The doc comment - I've been doing this for `src/new` to give some breadcrumbs for anyone who wants to check / update things.

androm3da: Ok, added.

src/qurt/mod.rs
tgross35, line 1: It looks like a lot of the defs in this file are duplicated between `src/qurt` and `src/new/qurt`, don't they need to be dropped?

Other than this, LGTM.

tgross35:
Also if you could squash any commits that have back-and-forth (e.g. adding `src/qurt` then later removing it) that would be helpful, makes backporting cleaner.

src/qurt/mod.rs
androm3da: Sorry: I have removed these duplicates now.

src/qurt/mod.rs
```diff
@@ -0,0 +1,380 @@
...
+    pub fn sqrt(x: c_double) -> c_double;
+    pub fn sqrtf(x: c_float) -> c_float;
+    pub fn ceil(x: c_double) -> c_double;
+    pub fn ceilf(x: c_float) -> c_float;
+    pub fn fabs(x: c_double) -> c_double;
+    pub fn fabsf(x: c_float) -> c_float;
+    pub fn floor(x: c_double) -> c_double;
+    pub fn floorf(x: c_float) -> c_float;
+    pub fn fmod(x: c_double, y: c_double) -> c_double;
+    pub fn fmodf(x: c_float, y: c_float) -> c_float;
```
tgross35: Noticed one other thing: we don't provide bindings for the math functions in libc, std binds them as necessary for the platform.

androm3da: Ok, fixed.

PR #4846, opened 2025-11-21
linux, android: Add definition for IUCLC

3v1n0:
# Description

Add `IUCLC` definition as per glibc and musl value, as it's required by nix (termios) and uutils for `stty`

This is defined in termios-c_iflag.h (or termbits.h) and it's shared across all the linux archs (a part powerpc and alpha), even though it's not POSIX

# Sources

- Linux: https://github.com/torvalds/linux/blob/2eba5e05d9bcf4cdea995ed51b0f07ba0275794a/include/uapi/asm-generic/termbits.h#L61
- Glibc: https://github.com/bminor/glibc/blob/ff92750112e782300ff455c13d9bb0ae1603aa65/sysdeps/unix/sysv/linux/bits/termios-c_iflag.h#L33
- Musl: https://github.com/bminor/musl/blob/1b76ff0767d01df72f692806ee5adee13c67ef88/arch/generic/bits/termios.h#L39
- Android: https://cs.android.com/android/platform/superproject/main/+/main:external/musl/arch/generic/bits/termios.h?q=IUCLC&ss=android%2Fplatform%2Fsuperproject%2Fmain

# Checklist


maurer (approved):

tgross35 (approved):
LGTM, thanks for including links!

PR #4850, opened 2025-11-29
m68k: add syscalls 451-469 for linux

Takashiidobe:
# Description

This PR adds some of the newer m68k linux syscalls.

Unsure why I have merge conflicts since the diff looks clean on the files changed tab.

# Sources

[The new syscalls in linux/m68k](https://github.com/torvalds/linux/blob/19eef1d98eeda3745df35839190b7d4a4adea656/arch/m68k/kernel/syscalls/syscall.tbl#L453-L471)

# Checklist


tgross35:
> Unsure why I have merge conflicts since the diff looks clean on the files changed tab.

It looks like this is based on 600e1d195458e0474c9d8c9ffcc574a36f364afb, which was from about a year ago. This file doesn't get touched much, but there are a handful of changes since then.

PR #4858, opened 2025-12-02
linux, fuchsia: Mark mq_attr padding area as such

3v1n0:
# Description

As per commit c100954964 (see https://github.com/rust-lang/libc/pull/4806#issuecomment-3603204960) the hard-coded partial equality implementations
have been dropped in favor for auto-generated ones, but they did not
work for mq_attr, since the padding area was not typed correctly.

Fix this, using the Padding type.

# Checklist


tgross35 (approved):
I guess this was missed in the patch where we wrapped (almost) all padding fields in `Padding`, thank you for adding it!

decathorpe:
Please also backport this to the v0.2 branch:
https://github.com/rust-lang/libc/issues/4892

PR #4862, opened 2025-12-03
cleanup: Mark private padding fields as Padding

3v1n0:
# Description

We have more private fields that are padding areas that are currently
not marked as Padding, which is causing the generated Hash
and PartialEq implementations to take care of them wrongly.

Fix them

This is a pre-requisite and API-compatible version of 

# Checklist

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -478,7 +478,7 @@ s! {
         #[cfg(target_endian = "little")]
         pub f_fsid: c_ulong,
         #[cfg(all(target_pointer_width = "32", not(target_arch = "x86_64")))]
-        __f_unused: c_int,
+        __f_unused: Padding<c_int>,
         #[cfg(target_endian = "big")]
         pub f_fsid: c_ulong,
         pub f_flag: c_ulong,
--- a/src/unix/bsd/freebsdlike/dragonfly/mod.rs
+++ b/src/unix/bsd/freebsdlike/dragonfly/mod.rs
@@ -215,7 +215,7 @@ s! {
         pub cr_uid: crate::uid_t,
         pub cr_ngroups: c_short,
         pub cr_groups: [crate::gid_t; 16],
-        __cr_unused1: *mut c_void,
+        __cr_unused1: Padding<*mut c_void>,
     }
 
     pub struct stack_t {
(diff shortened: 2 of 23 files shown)


src/types.rs
```diff
@@ -22,6 +22,18 @@ impl<T: Copy> Default for Padding<T> {
     }
 }
 
+impl<T: Copy> Padding<T> {
+    /// Const constructor for uninitialized padding in const contexts.
+    // FIXME: Change this into zeroed() and use MaybeUninit::zeroed()
+    // when we depend on rustc 1.75.0.
```
3v1n0, line 28: @tgross35 I added this so that we were following what `default()` does, but... Wouldn't actually be more correct if we instead do also:

```diff
diff --git a/src/types.rs b/src/types.rs
index 555d08727..2e218cd80 100644
--- a/src/types.rs
+++ b/src/types.rs
@@ -18,7 +18,7 @@ pub(crate) struct Padding<T: Copy>(MaybeUninit<T>);
 
 impl<T: Copy> Default for Padding<T> {
     fn default() -> Self {
-        Self(MaybeUninit::zeroed())
+        Self(MaybeUninit::uninit())
     }
 }
```

As padding memory isn't supposed to be touched, so not even needs to be initialized.

It's also true that if some part of the code goes out of memory boundaries we may get in troubles, but... Memory sanitizers could even have a better job with that.

src/types.rs
```diff
@@ -22,6 +22,18 @@ impl<T: Copy> Default for Padding<T> {
...
+    #[allow(unused)]
+    pub const fn uninit() -> Self {
```
tgross35: (commit 1) Could you make this `pub(crate)`? Currently it doesn't matter since `Padding` is crate-private, but if we ever make it public I'd rather do the methods separately.

src/types.rs
tgross35: I dis consider that - it would certainly work but not be any more/less correct. I still think `zeroed` makes sense because that will give you an all-zeros bitpattern for any struct containing padding, at least from C's perspective where these are actual fields (e.g. in the case of `__reserved` fields that may hold actual data), which is arguably "safer".

The only downside is a trivial bit of added cost in calling `default()`. I'm not too concerned about this, however because most users will continue wrapping their structs in `MaybeUninit` and writing fields as-needed.

(If it would help, feel free to put something like this in a comment)

3v1n0: Ok, fair enough... I in fact did not propose that change since I knew it was kinda of gray area

tgross35 (approved):
Looks awesome, thank you!

PR #4867, opened 2025-12-03
Change type of `AF_INET` and `AF_INET6` to `sa_family_t`

### LINKED ISSUE #468: AF_INET and AF_INET6 should be of type sa_family_t

GuillaumeGomez:
Fixes https://github.com/rust-lang/libc/issues/468.
Part of rust-lang/libc#3248.


GuillaumeGomez:
Thanks tgross, CI passed. :)

dybucc:
not sure if it's worth mentioning this as it seems sa_family_t is more meant to be capable of holding the macro expansion for AF_* macros rather than have the correct size and alignment requirements, but under illumos, depending on whether the type is used in unix domain sockets or in other sockaddr interfaces, sa_family_t will be one of c_uchar or u16 [^5]. the former is 1 byte while the latter is 2 bytes wide, as per [^1] [^2] [^3] [^4].

[^1]: https://github.com/illumos/illumos-gate/blob/a4c0fc073b79c010ae612eb38914998e3bb81078/usr/src/boot/sys/sys/limits.h#L43
[^2]: https://github.com/illumos/illumos-gate/blob/a4c0fc073b79c010ae612eb38914998e3bb81078/usr/src/boot/sys/x86/include/_limits.h#L50
[^3]: https://github.com/illumos/illumos-gate/blob/a4c0fc073b79c010ae612eb38914998e3bb81078/usr/src/head/iso/limits_iso.h#L56
[^4]: https://github.com/illumos/illumos-gate/blob/a4c0fc073b79c010ae612eb38914998e3bb81078/usr/src/uts/common/sys/types.h#L601
[^5]: https://github.com/search?q=repo%3Aillumos%2Fillumos-gate+path%3A**%2F*.h+%2F%28%3F-i%29%5Cbtypedef%5Cb.%2B%5Cbsa_family_t%5Cb%2F&type=code

dybucc:
much in the same vein as [^1], our current definition for sa_family_t on the GNU Hurd may be wrong. if we follow upstream glibc, there does not seem to be any sysdeps/ override for the mach kernel so the default definition in [^2] is that of a c_ushort and not a c_uchar as we have in [^3].

[^1]: https://github.com/rust-lang/libc/pull/4867#issuecomment-4955426473
[^2]: https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/sockaddr.h;h=842be8613300b4048b35a46c964952c1e2f90f6c;hb=HEAD#l28
[^3]: https://github.com/rust-lang/libc/blob/6094feef0ba041cdfa35c81a14c2b319eb197771/src/unix/hurd/mod.rs#L174

dybucc:
Just rebased to latest main at [^1]. Not sure if the author still intends on
finishing this up.

Pinging @tgross35 because I've seen you comment on a different stale PR.

[^1]:
    https://github.com/dybucc/libc/commit/81b9e72fd307649521791602b0c54a3ef9401ea0

PR #4811, opened 2025-11-03
Add missing constants from `unistd.h` for newlib targets

FenrirWolf:
# Description

A [new API](https://github.com/rust-lang/rust/pull/146937) was recently added to `std`, and as a result a number of newlib-based targets are [failing to build](https://does-it-build.noratrieb.dev/nightly?nightly=2025-10-06) due to a missing definition for `_SC_HOST_NAME_MAX` from the `unistd.h` header. This PR adds that missing constant along with a number of other `sysconf` symbols that never made it into the crate for these targets.

# Sources

https://github.com/bminor/newlib/blob/master/newlib/libc/include/sys/unistd.h#L354-L512

# Checklist

- [x] Relevant tests in `libc-test/semver` have been updated
- [x] No placeholder or unstable values like `*LAST` or `*MAX` are
  included (see [#3131](https://github.com/rust-lang/libc/issues/3131))
- [x] Tested locally (`cd libc-test && cargo test --target mytarget`);
  especially relevant for platforms that may not be checked in CI

@rustbot label +stable-nominated

PR #4875, opened 2025-12-04
newlib: Fix ambiguous glob exports and other warnings for Vita and 3DS

pheki:
# Description

This is actually a followup of rust-lang/libc#4811. It turns out that it accidentally introduced an ambiguous glob export for `_SC_PAGESIZE` and `_SC_GETPW_R_SIZE_MAX`. After updating newlib, rustc is fails to compile for these targets as it can't resolve `_SC_PAGESIZE` here: https://github.com/rust-lang/rust/blob/5325015e29874334003d33d77b295d57f593f06a/library/std/src/sys/pal/unix/os.rs#L530-L532

While checking, I realized there were also other warnings, for many newlib structures that ended up being unused, which I fixed:

### `hostent` definitions:

1. src/unix/mod.rs: The only actually resolved by rustc, and the most similar to [newlib since at least 2017](https://github.com/bminor/newlib/blob/f49f54b00d80ec12d6875def9252afba37f51825/newlib/libc/sys/rtems/include/netdb.h#L106-L113)

2. src/unix/newlib/mod.rs: This one has been there since the start of the newlib module, which matched the 3DS upstream definition at the time. Since then, rust-lang/libc#3863 arrived removing the 3DS (horizon) from that definition as the upstream header had been updated.

3. src/unix/newlib/horizon/mod.rs: This one is the new correct definition for horizon, which was deliberately added in rust-lang/libc#3863 but is not being used as it's shadowed by `1.`. I updated `1.` so this definition will now be used.

### `pollfd`

The definition in src/unix/mod.rs is the one being used and it matches with upstream newlib and the vita. 

The definition in src/unix/newlib/mod.rs was unused but it matches with the 3ds headers, so I moved it into horizon and opted it out from the above.

### `unix::newlib::generic::{sigset_t, stat, dirent}` 

All unused as they were shadowed by `vita`'s own definitions. Same goes for `horizon` except for `dirent`.

# Sources

- newlib hostent: https://github.com/bminor/newlib/blob/f49f54b00d80ec12d6875def9252afba37f51825/newlib/libc/sys/rtems/include/netdb.h#L106-L113
- vita hostent: https://github.com/vitasdk/newlib/blob/11c38129967fe2f8893d24dd7742b6ed6bd54d5f/newlib/libc/sys/vita/include/netdb.h#L40-L46
- horizon hostent: https://github.com/devkitPro/libctru/blob/00760077a8418284554980290e20f85231c69c2b/libctru/include/netdb.h#L11-L17
- newlib pollfd: https://github.com/bminor/newlib/blob/f49f54b00d80ec12d6875def9252afba37f51825/newlib/libc/sys/rtems/include/sys/poll.h#L51-L55
- vita pollfd: https://github.com/vitasdk/newlib/blob/11c38129967fe2f8893d24dd7742b6ed6bd54d5f/newlib/libc/sys/vita/sys/poll.h#L49-L53
- horizon pollfd: https://github.com/devkitPro/libctru/blob/00760077a8418284554980290e20f85231c69c2b/libctru/include/poll.h#L12-L17

# Checklist

Checked locally using `cargo +nightly check --target {target} -Z build-std=core` for:

- `armv7-sony-vita-newlibeabihf`: No warnings
- `armv6k-nintendo-3ds`: No warnings
- `armv7-rtems-eabihf`: Only an unused import warning on `src/new/mod.rs`: `pub(crate) use rtems::*;`

### DIFF
--- a/src/new/vita/mod.rs
+++ b/src/new/vita/mod.rs
@@ -1,4 +1,2 @@
 //! VITASDK system library.
 // FIXME(vita): link to headers or manpages needed.
-
-pub(crate) mod unistd;
--- a/src/new/vita/unistd.rs
+++ b/src/new/vita/unistd.rs
@@ -1,7 +1,0 @@
-//! Header: `unistd.h`
-
-pub use crate::new::common::posix::unistd::{
-    STDERR_FILENO,
-    STDIN_FILENO,
-    STDOUT_FILENO,
-};
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -142,7 +142,7 @@ s! {
         pub ipv6mr_interface: c_uint,
     }
 
-    #[cfg(not(target_os = "cygwin"))]
+    #[cfg(all(not(target_os = "cygwin"), not(target_os = "horizon")))]
     pub struct hostent {
         pub h_name: *mut c_char,
         pub h_aliases: *mut *mut c_char,
@@ -156,6 +156,7 @@ s! {
         pub iov_len: size_t,
     }
 
+    #[cfg(not(target_os = "horizon"))]
     pub struct pollfd {
         pub fd: c_int,
         pub events: c_short,
--- a/src/unix/newlib/generic.rs
+++ b/src/unix/newlib/generic.rs
@@ -1,16 +1,15 @@
 //! Common types used by most newlib platforms
 
-use crate::off_t;
+#[allow(unused_imports)] // needed for platforms that don't use the prelude here
 use crate::prelude::*;
 
 s! {
+    #[cfg(all(not(target_os = "vita"), not(target_os = "horizon")))]
     pub struct sigset_t {
-        #[cfg(target_os = "horizon")]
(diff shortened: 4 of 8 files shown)


tgross35 (approved):
Thank you for the fix!

tgross35:
I have an unrelated question if you are familiar with newlib: how does it handle 64-bit time_t? E.g. musl did a complete changeover at once, glibc has the test macros.

tgross35:
Oh awesome, that makes things much easier for us.

pheki:
Thanks!

As a note and for future reference, vita's newlib [uses that option](https://github.com/vitasdk/buildscripts/blob/9523a297e4d91e61413a0334928fd15a3ca36310/CMakeLists.txt#L543) (now located [here](https://github.com/vitasdk/buildscripts/blob/49c7878bd96ce2a2dcf7dac947d32a5254f307e9/cmake/recipes/BuildSdkComponents.cmake#L44)), so it has 32-bit time_t.

PR #4876, opened 2025-12-05
Make l4re pthread struct fields private

farao:
As discussed in the previous L4Re PR, the pthread struct fields shall be private.


tgross35:
Hey, I'm sorry this took so long but it is in yesterday's release (0.2.179)

PR #4877, opened 2025-12-08
Remove `[no-mentions]` handler in our triagebot config

Urgau:
This PR removes the `[no-mentions]` handler in our triagebot config as [GitHub is removing notifications for @-mentions in commit messages on December 8th 2025.](https://github.blog/changelog/2025-11-07-removing-notifications-for-mentions-in-commit-messages/)

cf. https://github.com/rust-lang/triagebot/issues/2225


tgross35 (approved):
Thanks!

PR #4880, opened 2025-12-10
Remove nonexistent clocks on WASI

alexcrichton:
These were added way back when in rust-lang/libc#2499 and while they may have existed in a historical version of wasi-libc they most certainly do not exist in the current wasi-libc. Any usage of them is already-breaking, so this deletes these items.


alexcrichton:
Sounds good to me! And yeah I'm comfortable backporting this myself, and feel free to redirect any confusion to me if you happen to see it

PR #4884, opened 2025-12-16
musl: s390x: Fix the values of `SIGSTKSZ` `MINSIGSTKSZ`

Gelbpunkt:
# Description

This fixes the remaining issues with the s390x musl definitions. libc-test passes fine after these changes.

# Sources

https://github.com/bminor/musl/blob/master/arch/s390x/bits/signal.h#L5-L6

# Checklist


tgross35 (approved):
Thanks!

PR #4886, opened 2025-12-19
netbsd/riscv64.rs: make changes so that this builds again.

he32:
# Description

These are changes which appear to be required to get this file to build (and cross-build) again.
Admittedly, some of the additions are based on error messages emitted by the rustc compiler.

# Sources

# Checklist

### DIFF
--- a/src/unix/bsd/netbsdlike/netbsd/riscv64.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/riscv64.rs
@@ -2,17 +2,9 @@ use crate::prelude::*;
 use crate::PT_FIRSTMACH;
 
 pub type __greg_t = u64;
-pub type __cpu_simple_lock_nv_t = c_int;
-pub type __gregset = [__greg_t; _NGREG];
-pub type __fregset = [__fpreg; _NFREG];
-
-s! {
-    pub struct mcontext_t {
-        pub __gregs: __gregset,
-        pub __fregs: __fregset,
-        __spare: [crate::__greg_t; 7],
-    }
-}
+pub type __cpu_simple_lock_nv_t = c_uint;
+pub type __gregset_t = [__greg_t; _NGREG];
+pub type __fregset_t = [__fpreg; _NFREG];
 
 s_no_extra_traits! {
     pub union __fpreg {
@@ -21,23 +13,36 @@ s_no_extra_traits! {
     }
 }
 
+s! {
+    pub struct mcontext_t {
+        pub __gregs: __gregset_t,
+        pub __fregs: __fregset_t,
+        __spare: [crate::__greg_t; 7],
+    }
+}
+
 cfg_if! {
     if #[cfg(feature = "extra_traits")] {
         impl PartialEq for __fpreg {
-            fn eq(&self, other: &Self) -> bool {
-                unsafe { self.u_u64 == other.u_u64 }
+            fn eq(&self, other: &__fpreg) -> bool {
+                unsafe { self.u_u64 == other.u_u64 || self.u_d == other.u_d }
             }
         }
         impl Eq for __fpreg {}
         impl hash::Hash for __fpreg {
             fn hash<H: hash::Hasher>(&self, state: &mut H) {
-                unsafe { self.u_u64.hash(state) };
+                unsafe {
+                    self.u_u64.hash(state);
+                }
             }
         }
     }
 }
 
-pub(crate) const _ALIGNBYTES: usize = size_of::<c_long>() - 1;
+// gcc for riscv64 defines `BIGGEST_ALIGNMENT`, but it's mesured in bits.
+pub(crate) const __BIGGEST_ALIGNMENT_IN_BITS__: usize = 128;
+// `_ALIGNBYTES` is measured in, well, bytes.
+pub(crate) const _ALIGNBYTES: usize = (__BIGGEST_ALIGNMENT_IN_BITS__ / 8) - 1;
 
 pub const PT_GETREGS: c_int = PT_FIRSTMACH + 0;
 pub const PT_SETREGS: c_int = PT_FIRSTMACH + 1;


he32:
> Not sure how those two failing CI jobs can be related to my proposed changes...

Evidently caused by what I had *not* done (merged upstream 'main' branch into mine).

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
```diff
@@ -2,42 +2,52 @@ use crate::prelude::*;
...
-        pub __gregs: __gregset,
-        pub __fregs: __fregset,
-        __spare: [crate::__greg_t; 7],
-    }
+#[derive(Debug)]
+pub struct mcontext_t {
+    pub __gregs: __gregset_t,
+    pub __fregs: __fregset_t,
+    __spare: [crate::__greg_t; 7],
 }
```
tgross35: Could you add back the `s!` wrapper?

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
```diff
@@ -2,42 +2,52 @@ use crate::prelude::*;
...
-    pub struct mcontext_t {
-        pub __gregs: __gregset,
-        pub __fregs: __fregset,
-        __spare: [crate::__greg_t; 7],
-    }
+#[derive(Debug)]
+pub struct mcontext_t {
+    pub __gregs: __gregset_t,
+    pub __fregs: __fregset_t,
+    __spare: [crate::__greg_t; 7],
```
tgross35: While you're here, you may as well wrap the `__spare` field in `Padding` (we're doing this for private unused fields)

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
```diff
@@ -2,42 +2,52 @@ use crate::prelude::*;
...
-            }
+    }
+}
+impl core::fmt::Debug for __fpreg {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        unsafe {
+            f.debug_struct("__fpreg")
+                .field("u_u64", &self.u_u64)
+                .field("u_d", &self.u_d)
+                .finish()
```
tgross35: Could you add back the `s_no_extra_traits` and `cfg` wrappers for `__fpreg`? You can just make it match https://github.com/rust-lang/libc/blob/0a5a8b65f98441cb71e7aaefd81672080d819443/src/unix/linux_like/linux/gnu/b64/s390x.rs#L218-L241.

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
```diff
@@ -2,42 +2,52 @@ use crate::prelude::*;
...
         }
     }
 }
 
-pub(crate) const _ALIGNBYTES: usize = size_of::<c_long>() - 1;
+pub(crate) const _ALIGNBYTES: usize = 0xf;
```
tgross35: Do you have a source for this? Per https://github.com/NetBSD/src/blob/29216014b5d25ec3b457bbca2bb2bc6a80d7a1fd/sys/arch/riscv/include/cdefs.h#L6 it looks like the existing definition is correct.

(If you could add source links to the PR description for all changes here, that would be great)

tgross35 (changes requested):
Sorry for the breakage, thank you for the PR!

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
he32: Well, one source is:

```
riscv64: {48} find /usr/include -type f | xargs egrep __ALIGNBYTES
/usr/include/sys/param.h:#define STACK_ALIGNBYTES	__ALIGNBYTES
/usr/include/sys/param.h:#define ALIGNBYTES	__ALIGNBYTES
/usr/include/sys/socket.h:#define __CMSG_ALIGN(n)	(((n) + __ALIGNBYTES) & ~__ALIGNBYTES)
/usr/include/riscv/cdefs.h:#define	__ALIGNBYTES	((size_t)(__BIGGEST_ALIGNMENT__ - 1U))
riscv64: {49} find /usr/include -type f | xargs egrep __BIGGEST_ALIGNMENT
/usr/include/riscv/cdefs.h:#define	__ALIGNBYTES	((size_t)(__BIGGEST_ALIGNMENT__ - 1U))
/usr/include/riscv/param.h:#define	ALIGNBYTES32		__BIGGEST_ALIGNMENT__
riscv64: {50} gcc -E -dD - < /dev/null | grep ALIGNMENT
#define __BIGGEST_ALIGNMENT__ 16
riscv64: {51} uname -a
NetBSD riscv64 10.99.12 NetBSD 10.99.12 (GENERIC64) #0: Sat Sep  7 14:28:19 UTC 2024  mkrepro@mkrepro.NetBSD.org:/usr/src/sys/arch/riscv/compile/GENERIC64 riscv
riscv64: {52}
```

I also have a test program I compiled & ran:

```
riscv64: {52} cat t.c
#include <sys/socket.h>
#include <pthread.h>
#include <stdio.h>

int
main(int argc, char **argv)
{
	int i;

	printf("sizeof cmsghdr: %u\n", sizeof(struct cmsghdr));
	printf("__ALIGNBYTES: %d\n", __ALIGNBYTES);

	printf("sizeof __pthread_spin_t: %u\n", sizeof(__pthread_spin_t));

	for (i=0; i<16; i++) {
		printf("cmsg_space(%d): %u", i, CMSG_SPACE(i));
		printf(" CMSG_ALGN(%d): %u\n", i, __CMSG_ALIGN(i));
	}
	return 0;
}
riscv64: {53} cc t.c
riscv64: {54} ./a.out
sizeof cmsghdr: 12
__ALIGNBYTES: 15
sizeof __pthread_spin_t: 4
cmsg_space(0): 16 CMSG_ALGN(0): 0
cmsg_space(1): 32 CMSG_ALGN(1): 16
cmsg_space(2): 32 CMSG_ALGN(2): 16
cmsg_space(3): 32 CMSG_ALGN(3): 16
cmsg_space(4): 32 CMSG_ALGN(4): 16
cmsg_space(5): 32 CMSG_ALGN(5): 16
cmsg_space(6): 32 CMSG_ALGN(6): 16
cmsg_space(7): 32 CMSG_ALGN(7): 16
cmsg_space(8): 32 CMSG_ALGN(8): 16
cmsg_space(9): 32 CMSG_ALGN(9): 16
cmsg_space(10): 32 CMSG_ALGN(10): 16
cmsg_space(11): 32 CMSG_ALGN(11): 16
cmsg_space(12): 32 CMSG_ALGN(12): 16
cmsg_space(13): 32 CMSG_ALGN(13): 16
cmsg_space(14): 32 CMSG_ALGN(14): 16
cmsg_space(15): 32 CMSG_ALGN(15): 16
riscv64: {55}
```

That said, I cannot offhand find where inside the gcc sources this predefined value comes from.  It does evidently not come from this place:

```
castor: {14} pwd
/usr/src/external/gpl3/gcc/dist/gcc/config
castor: {15} find riscv/ -type f | xargs grep BIGGEST_ALIGN
riscv/riscv.h:#define BIGGEST_ALIGNMENT 128
castor: {16}
```

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
he32: Will do, in a commit shortly to be done.

he32: Actually, my attempt at using `Padding` appears to have failed with
```
error[E0412]: cannot find type `Padding` in this scope
  --> /usr/pkgsrc/wip/rust192/work/rustc-1.92.0-src/vendor/libc-0.2.175/src/unix/bsd/netbsdlike/netbsd/riscv64.rs:20:18
   |
20 |         __spare: Padding<[crate::__greg_t; 7]>,
   |                  ^^^^^^^ not found in this scope
```
when this is cross-built for `riscv64` as part of the rust compiler.

Insights into how to correctly fix this gratefully accepted.

he32: > While you're here, you may as well wrap the __spare field in Padding (we're doing this for private unused fields)

Can you please point to an example, so that I can cargo-cult it? :)

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
```diff
@@ -21,23 +13,49 @@ s_no_extra_traits! {
...
+        impl core::fmt::Debug for __fpreg {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                unsafe {
+                    f.debug_struct("__fpreg")
+                        .field("u_u64", &self.u_u64)
+                        .field("u_d", &self.u_d)
+                        .finish()
+                }
             }
         }
```
tgross35: You should verify this builds with `--features extra_traits`. `s_no_extra_traits` adds a `Clone`, `Copy`, and `Debug` impl so I suspect it will fail.

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
```diff
@@ -21,23 +13,49 @@ s_no_extra_traits! {
...
-            fn eq(&self, other: &Self) -> bool {
-                unsafe { self.u_u64 == other.u_u64 }
+        impl core::cmp::PartialEq for __fpreg {
+            fn eq(&self, other: &__fpreg) -> bool {
+                unsafe { self.u_u64 == other.u_u64 || self.u_d == other.u_d }
+            }
+        }
+        impl core::cmp::Eq for __fpreg {}
+        impl core::marker::Copy for __fpreg {}
+        impl core::clone::Clone for __fpreg {
```
tgross35: No need to add the `core::cmp` prefix, it's in libc's prelude

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
```diff
@@ -21,23 +13,49 @@ s_no_extra_traits! {
...
 cfg_if! {
     if #[cfg(feature = "extra_traits")] {
-        impl PartialEq for __fpreg {
-            fn eq(&self, other: &Self) -> bool {
-                unsafe { self.u_u64 == other.u_u64 }
+        impl core::cmp::PartialEq for __fpreg {
+            fn eq(&self, other: &__fpreg) -> bool {
+                unsafe { self.u_u64 == other.u_u64 || self.u_d == other.u_d }
+            }
+        }
```
tgross35, line 30: Leave this body as-is, not having a false positive for ints is more important than indicating the -0==0 case for floats. Also matches other similar impls.

Or replace it with `unimplemented!` like the source I linked, these traits exist to fit in API signatures but they're going away in 1.0.

src/unix/bsd/netbsdlike/netbsd/riscv64.rs
tgross35: I could still use some context here

tgross35 (changes requested):

he32:
> I could still use some context here

Oh?  I thought I quoted from the gcc sources to adequatly address this.  But to repeat in another way:

Gcc's riscv.h file (https://github.com/NetBSD/src/blob/trunk/external/gpl3/gcc/dist/gcc/config/riscv/riscv.h line 213)

```
/* There is no point aligning anything to a rounder boundary than this.  */
#define BIGGEST_ALIGNMENT 128
```

However, this is measured in bits, and gets converted to the predefined `__BIGGEST_ALIGNMENT__` macro via
(https://github.com/NetBSD/src/blob/trunk/external/gpl3/gcc/dist/gcc/cppbuiltin.cc line 170):

```
  cpp_define_formatted (pfile, "__BIGGEST_ALIGNMENT__=%d",
			BIGGEST_ALIGNMENT / BITS_PER_UNIT);
```

which ends up as 16.  This results in an `__ALIGNBYTES` value of 15 or 0xf via

```
#define __ALIGNBYTES    ((size_t)(__BIGGEST_ALIGNMENT__ - 1U))
```

(from `/usr/include/machine/cdefs.h`, visible via https://github.com/NetBSD/src/blob/trunk/sys/arch/riscv/include/cdefs.h)

he32:
> If you could instead change to
```
pub(crate) const __BIGGEST_ALIGNMENT__: usize = 128;
pub(crate) const _ALIGNBYTES: usize = __BIGGEST_ALIGNMENT__ - 1;
```
> I think that would be best since it mirrors the source.

Actually, no, that is wrong.  Internally in gcc, `BIGGEST_ALIGNMENT` is measured in bits.  You need to divide by "bits-per-byte" before subtracting one to get the correct value for `__ALIGNBYTES`.  Instead perhaps something like

```
// gcc for riscv64 defines `BIGGEST_ALIGNMENT`, but it's mesured in bits.
pub(crate) const __BIGGEST_ALIGNMENT_IN_BITS__: usize = 128;
// `__ALIGNBYTES` is measured in, well, bytes.
pub(crate) const __ALIGNBYTES: usize = (__BIGGEST_ALIGNMENT_IN_BITS__ / 8) - 1;
```

I assume I don't need to comment the literal 8...

he32:
> Forgot to mention the division but that's what I meant. Thanks for the updates, LGTM but could you please squash?

OK.  I tried, but not only am I a rust newbie, I'm evidently also a git newbie, so what I thought would be one commit ended up as two for some reason or other.  Do you want me to make another stab at squashing them back to one?

tgross35:
> > Forgot to mention the division but that's what I meant. Thanks for the updates, LGTM but could you please squash?
> 
> OK. I tried, but not only am I a rust newbie, I'm evidently also a git newbie, so what I thought would be one commit ended up as two for some reason or other. Do you want me to make another stab at squashing them back to one?

No worries, I took care of it :) it looks like you squashed the __pthread_spin_t commit with one of this PR's commits then squashed the rest , but the pthread bit was already applied as d4dd6d8d9696c42afde356413d52ad87962bf95a.

For reference I just re-split the commits into pthread and the rest (`git reset $(git merge-base HEAD main)`, then 2x `git add -p` followed by `git commit` to select hunks of the diff to add to a new commit) and then rebased. Probably also would have figured itself out with only a rebase, I just didn't realize what had happened at first.

Thanks for the patch!

PR #4890, opened 2025-12-26
getitimer and setitimer for linux

youknowone:
# Description

fix rust-lang/libc#1347

# Sources
- https://github.com/bminor/glibc/blob/ae612c45efb5e34713859a5facf92368307efb6e/time/sys/time.h#L147-L157
- https://github.com/bminor/musl/blob/79bdacff83a6bd5b70ff5ae5eb8b6de82c2f7c30/include/sys/time.h#L22-L23

# Checklist

  - No update. already included


tgross35 (approved):
Thanks!

PR #4902, opened 2025-12-29
NetBSD if_msghdr fix for ilp32 hosts

he32:
# Description

This fixes the following errors from `cd libc-test; cargo test`:

```
    bad `if_msghdr` size: rust: 148 != c 152
    bad `if_msghdr` align: rust: 4 != c 8
and
    size of `struct if_msghdr` is 152 in C and 148 in Rust
```

as verified natively on NetBSD/i386 10.0 with rust 1.92.0.

# Sources

* https://nxr.netbsd.org/xref/src/sys/net/if.h#790 for the original C definition of `struct if_msghdr`.

# Checklist


tgross35 (approved):
Thank you!

PR #4903, opened 2025-12-30
Fix CMSG_NXTHDR soundness and boundary checks

gibbz00:
### Fixes:

- Properly reflect that musl and some descendants do not allow zero-sized payloads being in `CMSG_NXTHDR` [musl_ref].

- Unsoundness in linux and l4re. Makes sure that the header of next is within max, returning null if not. (Similar to how glibc does it.) It is the fix that should have been made rust-lang/libc#1235, but instead it added pretty severe soundness issues. (Mentioning it for context, not to point fingers.) No checks were previously being done to assert that next as usize + size_of::<cmsghdr>() < max. Wrapping offset calculations could then lead to buffer over-reads in the following (*next).cmsg_len.

### CMSG_NXTHDR Testing:

- Cleanup correctness. In turn removing some test values being ignored when targeting AIX.
- Re-enabling Sparc64 testing (fixing rust-lang/libc#1239), which for some reason consistently hit the UB conditions.
- Actually testing expected behavior when hitting the max controllen boundary.

[musl_ref]: https://www.openwall.com/lists/musl/2025/12/29/1

### DIFF
--- a/src/unix/linux_like/linux_l4re_shared.rs
+++ b/src/unix/linux_like/linux_l4re_shared.rs
@@ -1506,7 +1506,14 @@ f! {
             return core::ptr::null_mut();
         }
 
-        let max_addr = (*mhdr).msg_control as usize + (*mhdr).msg_controllen as usize;
+        let mut max_addr = (*mhdr).msg_control as usize + (*mhdr).msg_controllen as usize;
+
+        if cfg!(any(target_env = "musl", target_env = "ohos")) {
+            // musl and some of its descendants do `>= max_addr`
+            // comparisons in the if statement below.
+            // https://www.openwall.com/lists/musl/2025/12/27/1
+            max_addr -= 1;
+        }
 
         if next_cmsg as usize + size_of::<crate::cmsghdr>() > max_addr {
             core::ptr::null_mut::<crate::cmsghdr>()
--- a/src/unix/linux_like/emscripten/mod.rs
+++ b/src/unix/linux_like/emscripten/mod.rs
@@ -1257,7 +1257,7 @@ f! {
         }
         let next = (cmsg as usize + super::CMSG_ALIGN((*cmsg).cmsg_len as usize)) as *mut cmsghdr;
         let max = (*mhdr).msg_control as usize + (*mhdr).msg_controllen as usize;
-        if (next.offset(1)) as usize > max {
+        if (next.offset(1)) as usize >= max {
             core::ptr::null_mut::<cmsghdr>()
         } else {
             next as *mut cmsghdr


src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -1493,15 +1493,14 @@ f! {
...
+
+        let next_addr = cmsg as usize + super::CMSG_ALIGN((*cmsg).cmsg_len as usize);
+        let max_addr = (*mhdr).msg_control as usize + (*mhdr).msg_controllen as usize;
+
+        if next_addr + size_of::<crate::cmsghdr>() > max_addr {
             core::ptr::null_mut::<crate::cmsghdr>()
         } else {
-            next
+            next_addr as _
         }
```
tgross35, line 1522: First commit: could you use `.add` (or `.wrapping_add` as appropriate) rather than casting to `usize` and back? That should be easier to read and also preserves provenance.

libc-test/tests/cmsg.rs
tgross35, line 1: Commit 2 (test assertions): LGTM

tgross35, line 1: Commit 4 (sparc64): LGTM

libc-test/tests/cmsg.rs
```diff
@@ -74,14 +74,10 @@ mod t {
...
-            for cmsg_len in 0..64 {
-                // Address must be a multiple of 0x4 for testing on AIX.
-                if cfg!(target_os = "aix") && cmsg_len % std::mem::size_of::<cmsghdr>() != 0 {
-                    continue;
-                }
+            for cmsg_payload_len in 0..64 {
                 unsafe {
                     pcmsghdr.cast::<u8>().write_bytes(0, CAPACITY);
-                    (*pcmsghdr).cmsg_len = cmsg_len as _;
+                    (*pcmsghdr).cmsg_len = libc::CMSG_LEN(cmsg_payload_len as _) as _;
```
tgross35: Fyi @xingxue-ibm you may want to double check that this test runs on AIX, but I assume the fix here will resolve the issue in the comment.

libc-test/tests/cmsg.rs
tgross35, line 1: Commit 3 (test CMSG_LEN): LGTM

tgross35, line 1: Commit 6 (assertion context): LGTM

libc-test/tests/cmsg.rs
```diff
@@ -64,18 +64,33 @@ mod t {
...
+                        current_cmsghdr_ptr = libc_next;
+                    }
+                }
+
                 unsafe {
                     pcmsghdr.cast::<u8>().write_bytes(0, CAPACITY);
-                    (*pcmsghdr).cmsg_len = libc::CMSG_LEN(cmsg_payload_len as _) as _;
-                    let libc_next = libc::CMSG_NXTHDR(&mhdr, pcmsghdr);
-                    let next = cmsg_nxthdr(&mhdr, pcmsghdr);
-                    assert_eq!(libc_next, next);
```
tgross35: Commit 5 (boundary behavior test): The code LGTM, but would you mind adding some comments to source about what this is intended to check?

gibbz00: Not at all 😊 better now?

src/unix/linux_like/linux_l4re_shared.rs
gibbz00: Good point about provenance, fixed :) Stayed with `add()` as wrapping add wouldn't work with `if next_addr > max_addr { return ptr::null(); }`, and I don't think that it is sound to dereference a non-continuous allocation?

libc-test/tests/cmsg.rs
tgross35: Yes, thanks!

src/unix/linux_like/linux_l4re_shared.rs
gibbz00:
```
error: current MSRV (Minimum Supported Rust Version) is `1.63.0` but this item is stable since `1.75.0`
   --> src/new/common/posix/sys/socket.rs:148:35
    |
148 |     let next_cmsg = unsafe { cmsg.byte_add(CMSG_ALIGN(cmsghdr.cmsg_len as usize)) };
```
🥲

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -1498,7 +1498,14 @@ f! {
...
             .cast::<crate::cmsghdr>();
-        let max_addr = (*mhdr).msg_control as usize + (*mhdr).msg_controllen as usize;
+
+        #[allow(unused_mut)]
+        let mut max_addr = (*mhdr).msg_control as usize + (*mhdr).msg_controllen as usize;
+
+        #[cfg(any(target_env = "musl", target_env = "ohos"))]
+        {
+            max_addr -= 1;
+        }
```
tgross35: Prefer `cfg!(...)` over `#[cfg(...)]` where possible, since that means we get type checking regardless of platform. It will also remove the need for `allow(unused_mut)`.

Could you add a comment about `>=` vs. `>`?

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -1493,15 +1493,17 @@ f! {
...
             return core::ptr::null_mut::<crate::cmsghdr>();
         }
-        let next =
-            (cmsg as usize + super::CMSG_ALIGN((*cmsg).cmsg_len as usize)) as *mut crate::cmsghdr;
-        let max = (*mhdr).msg_control as usize + (*mhdr).msg_controllen as usize;
-        if (next.wrapping_offset(1)) as usize > max
-            || next as usize + super::CMSG_ALIGN((*next).cmsg_len as usize) > max
-        {
+
+        // IMPROVEMENT(1.75): use `.byte_add()`
```
tgross35: Nit: we use `FIXME(msrv)` for such things

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -1493,15 +1493,17 @@ f! {
...
+        let next_cmsg = (cmsg.cast::<u8>())
+            .add(super::CMSG_ALIGN((*cmsg).cmsg_len as usize))
+            .cast::<crate::cmsghdr>();
+        let max_addr = (*mhdr).msg_control as usize + (*mhdr).msg_controllen as usize;
+
+        if next_cmsg as usize + size_of::<crate::cmsghdr>() > max_addr {
             core::ptr::null_mut::<crate::cmsghdr>()
         } else {
-            next
+            next_cmsg as _
```
tgross35: Commit 1 nit: could you fill in `as *mut crate::cmsghdr` here? Somewhat more robust to avoid open-ended casts where possible.

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -1493,15 +1493,24 @@ f! {
...
             return core::ptr::null_mut::<crate::cmsghdr>();
         }
-        let next =
-            (cmsg as usize + super::CMSG_ALIGN((*cmsg).cmsg_len as usize)) as *mut crate::cmsghdr;
-        let max = (*mhdr).msg_control as usize + (*mhdr).msg_controllen as usize;
-        if (next.wrapping_offset(1)) as usize > max
-            || next as usize + super::CMSG_ALIGN((*next).cmsg_len as usize) > max
+
+        // IMPROVEMENT(1.75): use `.byte_add()`
+        let next_cmsg = (cmsg.cast::<u8>())
```
tgross35: Nit: extra parens shouldn't be needed

src/unix/linux_like/linux_l4re_shared.rs
tgross35: Ah yeah, unfortunately we have some version restrictions. Usually they don't come up too often but unfortunately you ran into one such case.

I think I would prefer using `wrapping_add` here because that makes this function precondition-free. You are right that it may not make a difference when you use the result, but it's the distinction between having UB when you call this function with invalid inputs vs. having UB only if you use the output.

(Not that this is ever actually going to matter in practice, but it's unlikely to get optimizations from `add` anyway.)

src/unix/linux_like/linux_l4re_shared.rs
gibbz00: I tried `cfg_if!` at first but realized that it didn't work.  I've now done what you said, but I had to add https://github.com/rust-lang/libc/pull/4903/changes/987a331097679053c0117b670feee4acdf50eb58. Changing `cfg_if!` feels a bit out of scope for this PR. Or what do you think?

tgross35: I didn't mention `cfg_if!` :) `if cfg!(any(target_env = "musl", target_env = "ohos")) { max_addr -= 1 }`

(I am still happy to have that `cfg_if` update though)

src/macros.rs
```diff
@@ -6,60 +6,58 @@
 ///
 /// This allows you to conveniently provide a long list #[cfg]'d blocks of code
 /// without having to rewrite each clause multiple times.
+///
+/// Implementation is effectively copy-pasted from the [`cfg_if`](cfg_if_crate),
+/// explanatory comments. Further explanatory source code comments on how the
+/// macro works can be found theret
```
tgross35: `explanatory comments` is just on its own, and `s/theret/there.`.

You can just copy+paste the whole thing, no need to delete the comments.

src/lib.rs
```diff
@@ -32,6 +32,8 @@
 
 #[macro_use]
 mod macros;
+pub(crate) use macros::*;
```
tgross35: `pub(crate) use macros::cfg_if;` so we don't accidentally pick up something unintended.

src/unix/linux_like/linux_l4re_shared.rs
tgross35: Heh, literally [`FIXME(msrv)`](https://github.com/search?q=repo%3Arust-lang%2Flibc+%2FFIXME%5C%28msrv%2F&type=code) (`FIXME(<version>)` is used to indicate a libc version rather than rustc, though it's only 1.0 right now)

src/unix/linux_like/linux_l4re_shared.rs
gibbz00: I thought for some reason that `.add()` would panic due to arithmetic overflow but see now that's not the case. So yes, I do understand and agree that `add()` should not be used, (but also that the function should be precondition-free).

However, I don't think `.wrapping_add()` is enough. Isn't point of `next_addr > max_addr` to check if `cmsg_len` can be trusted to not be out of bounds? Returning null if not?

Say for example that:

- The ancillary buffer is within the byte addresses 200 and 400.
- cmsg_addr = 300
- cmsg_len = usize::MAX - 350

Then: 
- next_addr = 300.wrapping_add(usize::MAX - 350) <=> next_addr = 50 

Which in turn makes 50 > 400 false (next_addr > max_addr), so no null pointer is returned :/

As such, would it not be better to also do  

```rust
// non-msrv compatible `wrapping_byte_add used` for brevity
let next_cmsg = cmsg.wrapping_byte_add(cmsg_len);
// in case the offset wraps
if (next_cmsg as usize) < (cmsg as usize) {
    return std::ptr::null_mut();
}
```

This should satisfy all points:

1) No UB with `add()`.
2) Wrapping addition still returns null.
3) No need for extra pre-condition checks.

src/unix/linux_like/linux_l4re_shared.rs
gibbz00: My bad, brain farted 😅

src/unix/linux_like/linux_l4re_shared.rs
gibbz00: Yeah sorry, confused them for some reason, fixed.

src/macros.rs
gibbz00: Problem is that the comments reference some issues, PR, and unit tests specific to that crate. I removed the commit for now, given that just `cfg!` works well. Probably better for the change to be a separate PR anyway.

src/lib.rs
gibbz00: (Dropped commit.)

src/unix/linux_like/linux_l4re_shared.rs
gibbz00: Adding a test that fails without the wrap assertion is pretty straightforward:

```rust
#[test]
fn wrapping_next() {
    let mut cmsghdr = unsafe { std::mem::zeroed::<cmsghdr>() };
    cmsghdr.cmsg_len = usize::MAX - 100;

    let mut mhdr = unsafe { std::mem::zeroed::<msghdr>() };
    mhdr.msg_control = (&raw mut cmsghdr).cast();
    mhdr.msg_controllen = 200;

    let result = unsafe { CMSG_NXTHDR(&raw const mhdr, &raw const cmsghdr) };
    assert!(result.is_null());
}
```

tgross35 (approved):
The new `wrapping_add` bit LGTM, thanks for all the updates!

PR #4904, opened 2025-12-30
netbsd/mod.rs: fix so that self-tests pass on both ilp32 and lp64.

he32:
# Description

This is mostly related to siginfo_t.
An error in si_status() implementation was also fixed.

Tests for now verified to pass on `NetBSD/i386 10.0` with `rust 1.92.0`.

# Sources

* https://nxr.netbsd.org/xref/src/sys/sys/siginfo.h#50 for `ksiginfo_t`, and `siginfo_t` is defined in terms of it.
* https://nxr.netbsd.org/xref/src/sys/sys/siginfo.h#161 for `si_status()` macro.

# Checklist


tgross35 (approved):
Thank you!

PR #4908, opened 2026-01-03
Consolidate CMSG_* implementations

gibbz00:
~~Builds upon rust-lang/libc#4903~~Merged

The original intent was to simplify the reasoning for how they differ across platforms. It should as a consequence also make it easier to extend target support.

### Changes:

- CMSG_* functions which can be const are all marked const -- consistently [^*].
- Removes unsafe from `CMSG_ALIGN`, `CMSG_SPACE`, `CMSG_LEN`.

### Fixes:

- Custom CMSG_FIRSTHDR implementation for VxWorks had missed to check that `mhdr.msg_controllen >= size_of::<cmsghdr>()`, as per POSIX 1003.1-2024.

### Documents:

- Usage and safety requirements.
- Adds missing references to upstream headers.

### Breaking API change Suggestions

- Perhaps CMSG_FIRSTHDR can take mhdr by reference? First thing being done is otherwise a raw pointer deref under practically no safety guarantees. Doing so would also remove the need to mark it unsafe.

- Same reasoning can be applied to mhdr in CMSG_NXTHDR, difference being that the function must remain unsafe for dereferencing cmsg.

[^*]: Except `CMSG_DATA` on solarish. It does `align(cmsg_addr + size_of(cmsg))` rather than `cmsg_addr + align(size_of(cmsg>))`. The others can get away with not casting the cmsg pointer to usize by instead casting cmsg to a byte pointer and doing the addition using `byte_ptr.offset()`.

### DIFF
--- a/src/new/apple/libc/mod.rs
+++ b/src/new/apple/libc/mod.rs
@@ -1,0 +1,6 @@
+//! Entrypoint for Apple headers, usually found as part of the Xcode SDK.
+//!
+//! <https://github.com/apple-oss-distributions/Libc/tree/main/include>
+
+pub(crate) mod signal;
+pub(crate) mod unistd;
--- a/src/new/common/posix/sys/socket.rs
+++ b/src/new/common/posix/sys/socket.rs
@@ -1,0 +1,300 @@
+//! Header: `sys/socket.h`
+//!
+//! <https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/sys_socket.h.html>
+
+// TEMP: until allow(unsafe_op_in_unsafe_fn) is removed in global level.
+#![deny(unsafe_op_in_unsafe_fn)]
+
+use core::mem::size_of;
+
+use crate::prelude::*;
+use crate::{
+    cmsghdr,
+    msghdr,
+};
+
+/// Rounds `len` *up* to the next alignment boundary.
+pub const fn CMSG_ALIGN(len: size_t) -> size_t {
+    align_impl(len, size_of::<__ALIGN_BOUNDARY>())
+}
+
+pub(crate) const fn align_impl(len: usize, align: usize) -> usize {
+    let mask = align - 1;
+    (len + mask) & !mask
+}
+
+// FIXME(#3240): consider changing signatures to `CMSG_{SPACE,LEN}(length: size_t) -> size_`
+
+/// Total length of a non-padded control message for a payload of size `length`.
+///
+/// This function is almost exclusively used setting [`cmsghdr::cmsg_len`].
+/// It should *not* be used for determining the actual ancillary data buffer
+/// size. [`CMSG_SPACE`] should instead be for that.
+#[cfg(not(any(target_os = "illumos", target_os = "solaris")))]
+pub const fn CMSG_LEN(length: c_uint) -> c_uint {
+    let length = length as size_t;
+
+    // See `CMSG_SPACE` impl about this sometimes being a no-op.
+    let len = CMSG_ALIGN(size_of::<cmsghdr>()) + length;
+
+    len as c_uint
+}
+
+/// Total length of a padded control message for a payload of size `length`.
+///
+/// This can be used to allocate space dynamically for the ancillary data.
+/// It should should *not* be used to initialize [cmsghdr::cmsg_len], given
+/// that the returned value includes padding bytes. Use instead [`CMSG_LEN`]
+/// for that.
+pub const fn CMSG_SPACE(length: c_uint) -> c_uint {
+    let length = length as size_t;
+
+    // NB: left hand side `align` is no-op when `size_of::<cmsghdr>() %
+    // size_of::<__ALIGN_BOUNDARY>() == 0`. Such is the case on Linux, and
+    // probably why some implementations there don't bother with the lhs
+    // align.
+    let space = CMSG_ALIGN(size_of::<cmsghdr>()) + CMSG_ALIGN(length);
+
+    space as c_uint
+}
+
+/// Returns a pointer to the payload data array associated with for the provided header.
+///
+/// # Safety
+///
+/// Safety in calling this function are based on the safety conditions needed for calling `ptr.offset()`.
+///
+/// * `cmsg` must point to some allocation and the resulting offset may not wrap around an address space.
+//
+// TODO: From https://man7.org/linux/man-pages/man3/cmsg.3.html, include it for the rest?
+//
+// The pointer returned cannot be assumed to be suitably aligned for
+// accessing arbitrary payload data types. Applications should not cast it
+// to a pointer type matching the payload, but should instead use `memcpy`
+// to copy data to or from a suitably declared object.
+#[cfg(not(any(target_os = "illumos", target_os = "solaris")))]
+pub const unsafe fn CMSG_DATA(cmsg: *const cmsghdr) -> *mut c_uchar {
+    let ptr = cmsg as *mut crate::c_uchar;
+    // See `CMSG_SPACE` impl about this sometimes being a no-op.
+    let byte_offset = CMSG_ALIGN(size_of::<cmsghdr>()) as isize;
+
+    // SAFETY:
+    // - Offset fits in isize (at most size_of::<u128>() = 16)
+    // - Caller must uphold the safety contract of non-wrapping offset
+    unsafe { ptr.offset(byte_offset) }
+}
+
+/// Return a pointer to the first [`cmsghdr`] in [`msghdr`].
+///
+/// A null pointer is returned if [`msghdr::msg_controllen`] is less than
+/// `size_of::<cmsghdr>()`. This can mean that there was no ancillary data
+/// in the first place (msg_controllen is zero), or that the ancillary data
+/// has been truncated.
+pub const unsafe fn CMSG_FIRSTHDR(mhdr: *const msghdr) -> *mut cmsghdr {
+    // IMPROVEMENT: take mhdr by ref and make CMSG_FIRSTHDR safe?
+    let msghdr = unsafe { *mhdr };
+
+    // Linux's syscall interface uses `size_t` for `msg_controllen` even if
+    // POSIX dictates it being `socklen_t` (>=u32). The cast accommodates
+    // for both, even if redundant for Linux.
+    #[allow(clippy::unnecessary_cast)]
+    let msg_controllen = msghdr.msg_controllen as size_t;
+
+    if msg_controllen < size_of::<cmsghdr>() {
+        core::ptr::null_mut()
+    } else {
+        msghdr.msg_control.cast()
+    }
+}
+
+/// # SAFETY
+///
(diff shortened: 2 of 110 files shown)


src/new/apple/libc/mod.rs
tgross35, line 1: Is this one actually getting picked up? The module is defined here https://github.com/rust-lang/libc/blob/ab195ebff42d2e7ed890795869882c0626fbdc4e/src/new/apple/mod.rs#L9-L12

src/posix.rs
tgross35, line 1: Looks like this may have gotten added by accident

src/new/apple/libc/sys/socket.rs
```diff
@@ -0,0 +1,20 @@
+//! Header: `sys/socket.h`
+//!
+//! <https://github.com/freebsd/freebsd-src/blob/main/sys/sys/socket.h>
```
tgross35: Probably not in the Apple module? 🙂

src/new/common/posix/sys/socket.rs
```diff
@@ -0,0 +1,299 @@
...
+pub const fn CMSG_ALIGN(len: size_t) -> size_t {
+    align_impl(len, size_of::<__ALIGN_BOUNDARY>())
+}
+
+pub(crate) const fn align_impl(len: usize, align: usize) -> usize {
+    let mask = align - 1;
+    (len + mask) & !mask
+}
+
+// TODO(#3240): consider changing signatures to `CMSG_{SPACE,LEN}(length: size_t) -> size_`
```
tgross35: This is meant to be committed right? The rust-lang org's convention is that TODO is for things you plan to fix before merge, FIXME is for things that will live in the codebase, so this should change if so.

src/new/common/posix/sys/socket.rs
```diff
@@ -0,0 +1,299 @@
...
+            } else if #[cfg(target_arch = "powerpc64")] {
+                pub(crate) type __ALIGN_BOUNDARY = c_long;
+            } else if #[cfg(target_arch = "riscv64")] {
+                pub(crate) type __ALIGN_BOUNDARY = c_long;
+            } else if #[cfg(target_arch = "sparc64")] {
+                pub(crate) type __ALIGN_BOUNDARY = u128;
+            }
+        }
+    }
+}
```
tgross35, line 300: We're trying to move away from this kind of nested config in the new structure. Could you move them to the respective `socket` modules instead?

Also, this is something used so the shared impl is possible rather than actually being in source, right? A comment to that effect would be helpful.

tgross35: Where do these definitions come from by the way? Looking at e.g. NetBSD I see https://github.com/NetBSD/src/blob/e46385ea0f041fae85cfb2a10b8b4910124827b8/sys/sys/socket.h#L544-L548 with `__ALIGNBYTES` defined in a few different places https://github.com/search?q=repo%3ANetBSD%2Fsrc+%2F%23define+__ALIGNBYTES%2F&type=code, but nothing using types like `double`.

src/new/common/bsd.rs
```diff
@@ -1 +1,17 @@
 //! Interfaces common across the BSD family.
+
+pub(crate) mod sys {
+    pub(crate) mod socket {
```
tgross35: Could you make `sys` a new directory and `socket` a new module in it? I've been trying to avoid >1 layer of nesting in the `new` modules, and `sys` tends to have enough contents that it's worth its own directory.

(Applies to a few places)

src/new/common/posix/sys/socket.rs
```diff
@@ -0,0 +1,299 @@
...
+) -> *mut cmsghdr {
+    // TODO: why not take mhdr by ref?
+    let msghdr = unsafe { *mhdr };
+
+    // SAFETY:
+    // - Caller ensured pointer is not null.
+    // - Pointer retrieved from either CMSG_FIRSTHDR or CMSG_NXTDR,
+    //   thus ensuring that is lies between mhdr_addr and mhdr_addr +
+    //   msg_controllen.
+    let cmsghdr = unsafe { *cmsg };
```
tgross35: Did these come from the original impls?

I would slightly prefer reading the fields only as-needed, so as to avoid creating UB if the struct is partially initialized. IMO it's a terrible idea to ever use non-zeroed libc types but unfortunately people do it, so we should avoid giving them more unsoundness then they'd get from the same macro calls in C.

src/new/common/posix/sys/socket.rs
```diff
@@ -0,0 +1,299 @@
...
+///   originally retrieved from a call to `CMSG_FIRSTHDR`.
+#[cfg(not(any(
+    target_env = "musl",
+    target_env = "ohos",
+    target_os = "emscripten",
+    target_os = "fuchsia"
+)))]
+pub unsafe fn CMSG_NXTHDR(mhdr: *const msghdr, cmsg: *const cmsghdr) -> *mut cmsghdr {
+    unsafe { next_impl(mhdr, cmsg, true) }
+}
```
tgross35, line 126: The BSDs define a `CMSG_NXTHDR` too right? Doesn't this wind up duplicated?

src/new/mod.rs
```diff
@@ -173,40 +173,74 @@ cfg_if! {
...
+        cfg_if! {
+            if #[cfg(any(target_env = "musl", target_env = "ohos"))] {
+                // ./musl
+                pub use sys::socket::*;
+            } else if #[cfg(target_env = "gnu")] {
+                // ./glibc/sysdeps/unix/linux
+                pub use net::route::*;
+                pub use sys::socket::*;
+            }
+        }
```
tgross35: What's the reason for moving this to be nested under linux? We do have glibc targets that don't set `target_os = "linux"` (e.g. vxworks, hurd) and the same is possible for musl.

It's fine if these modules export nothing on those platforms if they don't exist, e.g. in the new glibc module there's only `sysdeps/unix/linux`.

src/new/common/posix/sys/socket.rs
```diff
@@ -0,0 +1,299 @@
...
+
+pub(crate) unsafe fn next_impl(
+    mhdr: *const msghdr,
+    cmsg: *const cmsghdr,
+    allow_zero_sized_payload: bool,
+) -> *mut cmsghdr {
```
tgross35, line 132: Anything here that couldn't be made `const` at our MSRV? I think you may as well constify `CMSG_NXTHDR` if so, I could imagine it's occasionally useful.

src/new/apple/libc/mod.rs
gibbz00: There's a `pub(crate) use libc::*;` a bit further down :)

src/posix.rs
gibbz00: Correct, thanks for the catch!

src/new/apple/libc/mod.rs
gibbz00: Hah, see what you mean. I seem to have made it into a `mod.rs` whilst keeping the `mod libc {}` somehow. I probably did when adding a `libc/sys/socket.rs`, realizing later on that it should be under `xnu/`, (hence the out-of-place freebsd copy-pasta). 

Anyhow, I've now removed the `libc/sys/socket.rs`, making the `mod.rs` creation irrelevant to this PR... Do you want med to keep it anyway?

src/new/common/posix/sys/socket.rs
gibbz00: Ah, thanks for letting me know. Corrected it to `FIXME`.

src/new/apple/libc/sys/socket.rs
gibbz00: Removed, thanks.

src/new/common/bsd.rs
gibbz00: Done ️✌️ Anything still missing?

src/new/common/posix/sys/socket.rs
gibbz00: Wrote it in such a way that the BSDs reuse the `CMSG_NXTHDR`, something like;

```rust
// common/bsd
pub unsafe fn CMSG_NXTHDR(mhdr: *const msghdr, cmsg: *const cmsghdr) -> *mut cmsghdr {
    if cmsg.is_null() {
        CMSG_FIRSTHDR(mhdr)
    } else {
        CMSG_NXTHDR(mhdr, cmsg)
    }
}
```

It's actually the more POSIX compliant way:

> If the first argument is a pointer to a msghdr structure and the second argument is a null pointer, this macro shall be equivalent to CMSG_FIRSTHDR(mhdr).

https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/sys_socket.h.html

src/new/common/posix/sys/socket.rs
gibbz00: It's the usize pointer casts, tried to look into it as well.

> Transmuting pointers to integers in a const context is undefined behavior, unless the pointer was originally created from an integer. 

https://doc.rust-lang.org/std/mem/fn.transmute.html#transmutation-between-pointers-and-integers

src/new/common/posix/sys/socket.rs
gibbz00: Agreed. Changed it to:

```rust
let cmsg_len = unsafe { *cmsg }.cmsg_len;

// ..

let mut max_addr = {
    // IMPROVEMENT: why not take mhdr by ref?
    let msghdr = unsafe { *mhdr };
    msghdr.msg_control as usize + msghdr.msg_controllen as usize
};
```

src/new/common/posix/sys/socket.rs
gibbz00: Yeah, I know :/ They were in their respective modules at first, (not the `new/` ones), but then I started to get some lint warnings. Something about "declaration in module after use", can't remember which one exactly.

I left them here, expecting some discussion on what best to do. Should have been a bit clearer on that part.

They're technically all used solely for CMSG_ALIGN. Went ahead and removed `pub(crate)` to make that clearer.

> Where do these definitions come from by the way? 

Yeah, I also noticed a lot of inconsistencies with upstream, notably in what type is used for the `size_of`. Most are pretty harmless, Ex. Linux-like CMSG_ALIGN implementations normally use `size_t` or `long`, but this crate will often use `usize`.

Decided on not "correct" these. Wasn't keen on creating any unintended breaking changes. The one's I did change though were those that where previously defined as the result of `size_of::<something>()`, opting for the corresponding `uN` type alias. Solarish for example, had its alignment constant changed from `const _CMSG_HDR_ALIGNMENT: usize = 8;` to `type __ALIGN_BOUNDARY = u64;`.

gibbz00: So what do you think? Make an exception here to using nested config in new and add a source comment on why? I may otherwise be a bit confusing that `common/posix/sys/socket.rs` is importing a single use `pub(crate) __ALIGN_BOUNDARY` defined in a `<target>/sys/socket.rs`.

Issue #4913, opened 2026-01-05
0.1.178->0.1.179 regression: 32-bit musl `struct stat` fields


tgross35:
It looks like 793fa971aabfbadb1ef9564cdc9fcdc31f3ef636 removed the `st_atime + st_atime_nsec` and similar fields, which should still exist under `#[cfg(not(musl_v1_2_3))]`. Result:

https://docs.rs/libc/0.2.179/i686-unknown-linux-musl/libc/struct.stat.html
vs.
https://docs.rs/libc/0.2.178/i686-unknown-linux-musl/libc/struct.stat.html.

Cc @xbjfk

xbjfk:
I'm sorry about this - this was a poor oversight on my part - I've created #4914 to fix this

tgross35:
I'm actually not quite sure why tests didn't pick this up. There are some overrides https://github.com/rust-lang/libc/blob/17adf2d8003fd5021358905abdfeed68d504070f/libc-test/build.rs#L3974-L3980 but it should be getting checked.

PR #4914, opened 2026-01-05
musl: Fix incorrect definitions of struct stat on some architectures

xbjfk:
Fixes: rust-lang/libc#4913

# Description

This fixes the regression introduced in 793fa971aabfbadb1ef9564cdc9fcdc31f3ef636 and ensures the old struct stat definitions remain.

# Checklist


tgross35 (approved):
Thank you!

PR #4916, opened 2026-01-06
AIX: Only test addresses that are multiples of the size of cmsghdr

xingxue-ibm:
# Description
AIX does not apply any alignment or padding to ancillary data, and `CMSG_ALIGN()` is a no-op. Therefore, we only test addresses that are multiples of the size of `cmsghdr` on AIX. The original condition for AIX was removed in https://github.com/rust-lang/libc/commit/f391df35f33a15ec71efc81d200d77fb30ef3bbe, which led to the test failure.

# Sources

# Checklist

### DIFF
--- a/libc-test/tests/cmsg.rs
+++ b/libc-test/tests/cmsg.rs
@@ -73,6 +73,14 @@ mod t {
             mhdr.msg_controllen = (160 - trunc) as _;
 
             for cmsg_payload_len in 0..64 {
+                // AIX does not apply any alignment or padding to ancillary
+                // data and CMSG_ALIGN() is a noop. So only test addresses
+                // that are multiples of the size of cmsghdr here.
+                if cfg!(target_os = "aix") && cmsg_payload_len % std::mem::size_of::<cmsghdr>() != 0
+                {
+                    continue;
+                }
+
                 let mut current_cmsghdr_ptr = pcmsghdr;
                 assert!(!current_cmsghdr_ptr.is_null());
                 let mut count = 0;


gibbz00:
@xingxue-ibm 

I'm sorry, but I don't understand how this makes sense. You say that AIX does not apply any padding, yet the payload is required to be a multiple of `size_of::<cmsghdr>`? Sounds to me like AIX's CMSG_ALIGN then be:

```
fn CMSG_ALIGN(len: usize) -> usize { (len + size_of::<cmsghdr>() - 1) & !(size_of::<cmsghdr>() - 1) }
```

How do you otherwise deal with payloads that aren't multiples of exactly `cmsghdr` bytes? Say when sending only one file descriptor over a unix socket?

gibbz00:
... Or in other words, to me it sounds like the AIX C implementation does apply padding, but the Rust implementation doesn't, hence the test failure. If so, then the solution shouldn't be to ignore the test values where the implementations differ, but rather fix the Rust implementation. This may be easier with setting the `__ALIGN_BOUNDARY` from #4908, but I doubt it will land in before 1.0. 

What kind of test failures were you getting?

gibbz00:
Thanks for the quick response.

If you don't mind me asking, how does AIX's CMSG_SPACE C definition look like?

gibbz00:
Oh wow, interesting. So it's not that AIX supports unaligned access, but more so that it doesn't try to account for it at all.

> Misaligned pointer dereference: address must be a multiple of 0x4 but is 0x1108c63e5

If I'm reading that correctly, isn't the minimum alignment 4 bytes? Not cmsghdr's 16?

Did the tests not work with `cmsg_payload_len % std::mem::size_of::<u32>() != 0`?

That would explain a lot about how it doesn't affect AIX in practice. Can't think of any cmsg payload that isn't divisible to 4.

PR #4917, opened 2026-01-07
ci: Add rtems to basic tier 3 checks

thesummer:
# Description

This PR adds rtems to the basic tier 3 target checks.

# Sources

N/A

# Checklist

- [x] Relevant tests in `libc-test/semver` have been updated --> N/A
  included (see [#3131](https://github.com/rust-lang/libc/issues/3131)) --> N/A
  especially relevant for platforms that may not be checked in CI --> does not compile, but same for other tier 3 targets

### DIFF
--- a/ci/verify-build.py
+++ b/ci/verify-build.py
@@ -155,6 +155,7 @@ TARGETS = [
     Target("aarch64-unknown-openbsd", dist=False),
     Target("aarch64-wrs-vxworks", dist=False),
     Target("armebv7r-none-eabihf", dist=False),
+    Target("armv7-rtems-eabihf", dist=False),
     Target("armv7-wrs-vxworks-eabihf", dist=False),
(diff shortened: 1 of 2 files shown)


ci/verify-build.py
```diff
@@ -156,6 +156,7 @@ class TargetResult:
     Target("aarch64-wrs-vxworks", dist=False),
     Target("armebv7r-none-eabihf", dist=False),
     Target("armv7-wrs-vxworks-eabihf", dist=False),
+    Target("armv7-rtems-eabihf", dist=False),
     Target("armv7r-none-eabihf", dist=False),
     Target("armv7s-apple-ios", dist=False),
```
tgross35, line 161: Nit: could you keep this sorted? LGTM otherwise

thesummer: My bad. Should be fixed now.

PR #4918, opened 2026-01-07
Fixing the improper type of sighandler_t  on unix

### LINKED ISSUE #1359: sighandler_t has an incorrect type on most unix targets

ItshMoh:
Fixes: [#1359](https://github.com/rust-lang/libc/issues/1359) 

Continued from rust-lang/libc#2107

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -260,12 +260,12 @@ cfg_if! {
     if #[cfg(feature = "extra_traits")] {
         impl PartialEq for __c_anonymous_sigaction_handler {
             fn eq(&self, other: &__c_anonymous_sigaction_handler) -> bool {
-                unsafe{ self.default == other.default };
+                unsafe{ self.default == other.default }
             }
         }
         impl Eq for __c_anonymous_sigaction_handler{}
         impl fmt::Debug for __c_anonymous_sigaction_handler {
-            fn fmt(&self, f: &mut fmt::Formatter) -> ::fmt::Result
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result
             {
                 f.debug_struct("sigaction_t")
                    .field("value", unsafe{ &self.default })
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -44,7 +44,7 @@ extern_ty! {
 #[cfg(not(target_os = "nuttx"))]
 pub type locale_t = *mut c_void;
 
-s_no_extra_traits!{
+s_no_extra_traits! {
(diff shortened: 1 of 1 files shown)


src/unix/mod.rs
```diff
@@ -44,6 +44,18 @@ extern_ty! {
 #[cfg(not(target_os = "nuttx"))]
 pub type locale_t = *mut c_void;
 
+s_no_extra_traits! {
+    pub union __c_anonymous_sigaction_handler{
+        pub sa_handler: Option<extern "C" fn(c_int) -> ()>,
+        pub sa_sigaction: Option<extern "C" fn(
```
tgross35, line 50: Could you make these functions `unsafe`?

src/unix/mod.rs
```diff
@@ -242,12 +254,28 @@ cfg_if! {
...
+            }
+        }
+        impl Eq for __c_anonymous_sigaction_handler{}
+        impl hash::Hash for __c_anonymous_sigaction_handler {
+            fn hash<H: hash::Hasher>(&self, state: &mut H) {
+                unsafe{ self.default.hash(state) };
+            }
+        }
+    }
+}
```
tgross35, line 273: Please delete these, we are avoiding adding more iffy trait impls (even though this one is reasonably sound)

src/unix/mod.rs
```diff
@@ -44,6 +44,18 @@ extern_ty! {
 #[cfg(not(target_os = "nuttx"))]
 pub type locale_t = *mut c_void;
 
+s_no_extra_traits! {
+    pub union __c_anonymous_sigaction_handler{
```
tgross35, line 48: This doesn't need to be a `__c_anonymous` type, it can just be `union sighandler_t` with no alias since we're exporting it this way anyway.

tgross35 (changes requested):

src/unix/mod.rs
tgross35, line 48: Could you also add a comment about what the variants are representing? The two function pointers come from source, the integer variant is needed since sentinel values that aren't function pointers are used.

(maybe there's a better name for `default`? I think `raw` may make more sense)

tgross35:
I'm sort of thinking that maybe we should have a type indicating raw function pointers and have those in a two-variant union, rather than a three-variant union with Rust function pointers. Sketched some of this up at https://github.com/rust-lang/libc/issues/4998, curious what you think @JohnTitor

PR #4925, opened 2026-01-12
Add ENOTCAPABLE for macOS

Thomasdezeeuw:
# Description

Adds the `ENOTCAPABLE` constant. Also updates ELAST to 107 since a new error was added.

# Sources

* https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/errno.h#L267-L267

# Checklist


Thomasdezeeuw:
I was also wondering if `ELAST` should be deleted? But since it exists already it might be a breaking change?

tgross35 (dismissed):
Needs updates given the CI failure.

Which macos version is xnu-12377.1.9? If it is MacOS 26, GHA should support that now, and you could send a PR bumping it. Otherwise you can add a skip in `libc-test/build.rs`, but the CI bump would be much preferred.

Thomasdezeeuw:
With the update of macOS the CI now starts to fail:
```
 bad `HOST_VM_INFO64_COUNT` value at byte 0: rust: 38 (0x26) != c 40 (0x28)
    rust bytes: 26 00 00 00

    c bytes:    28 00 00 00

bad `vm_statistics64_data_t` size: rust: 152 != c 160
bad `vm_statistics64` size: rust: 152 != c 160
size of `vm_statistics64_data_t` is 160 in C and 152 in Rust
size of `struct vm_statistics64` is 160 in C and 152 in Rust
```

So we need #4926 as well, or ignore it somehow. I didn't see the `HOST_VM_INFO64_COUNT` failure locally oddly enough.

JohnTitor:
> I was also wondering if ELAST should be deleted? But since it exists already it might be a breaking change?

Yeah, it should be removed in 1.0.

So we only can add `ENOTCAPABLE` on 0.2 as any other things would be a breaking change, a solution here is moving things other than `ENOTCAPABLE` addiction (and test skip) to #4926, ship this on 0.2, then introduce breaking changes on 1.0 with #4926.

tgross35:
Could you do the CI bump in a separate PR? To keep things that are required from that separate from new API.

tgross35:
Yeah that's expected, bumping CI typically requires a few commits to update changed values like that. Getting that through before doing any brand new API just makes things simpler.

Thomasdezeeuw:
I've dropped the CI commit (with a force push).

JohnTitor (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/4925/7833bf83a5ca781368c047229f0116775d7f777f..d867eeac7892f1a396a0c6a419470555800f477e)*

PR #4931, opened 2026-01-20
AIX: Revert timespec.tv_nsec to c_long and use st_timespec for stat

xingxue-ibm:
# Description
The previous commit https://github.com/rust-lang/libc/pull/4750 changed the `tv_nsec` field of the `timespec` structure to `i32` to match the output types of system calls `fstat` and `fstat64`. However, it turns out that other system calls, such as `nanosleep`, expect the `tv_nsec` field to be of type `c_long`. This patch restores the type of `timespec.tv_nsec` to `c_long` and uses `struct st_timespec` in `struct stat` and `struct stat64` for the `fstat` and `fstat64` system calls, aligning these definitions with those in the AIX <sys/stat.h> header.

In <sys/stat.h>:
```
/* internal timespec to preserve binary compatibility */
typedef struct st_timespec {
        time_t  tv_sec;         /* seconds */
        int     tv_nsec;        /* and nanoseconds */
} st_timespec_t;
```

# Sources

# Checklist


JohnTitor (approved):
Thanks!

PR #4934, opened 2026-01-22
tests linux: do not skip tests depending on kernel <= 6.8

Marcondiro:
# Description

The Linux images used in CI use kernel 6.11, and headers are mostly >=6.8 at the time of writing, the only exceptions begin 32 bit architectures which are still on old headers due to the 32 -> 64 bit time_t enforcement in newer images.

Remove the tests requiring a kernel <= 6.8 from the skipped ones for 64 bit images not using musl.

Bump x86 32 images since x86 won't get the 64 bits time_t.

# Sources

.

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2056,9 +2056,6 @@ fn test_android(target: &str) {
             // Needs a newer Android SDK for the definition
             "P_PIDFD" => true,
 
-            // Requires Linux kernel 5.6
-            "VMADDR_CID_LOCAL" => true,
-
             // FIXME(android): conflicts with standard C headers and is tested in
             // `linux_termios.rs` below:
             "BOTHER" => true,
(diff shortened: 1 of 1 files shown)


libc-test/build.rs
```diff
@@ -4106,14 +4092,17 @@ fn test_linux(target: &str) {
...
+            // FIXME(musl): A field was added in linux 6.8, not yet in musl
+            // FIXME(linux32): A field was added in linux 6.8
             // https://github.com/torvalds/linux/commit/341ac980eab90ac1f6c22ee9f9da83ed9604d899
             // The previous version of the struct was removed in 6.11 due to a bug.
             // https://github.com/torvalds/linux/commit/32654bbd6313b4cfc82297e6634fa9725c3c900f
-            "xdp_umem_reg" => true,
+            "xdp_umem_reg" if musl || pointer_width == 32 => true,
 
-            // FIXME(linux): Requires >= 6.8 kernel headers.
+            // FIXME(linux): A new field was added to `xsk_tx_metadata_request` in linux 6.15.
```
tgross35: Would you mind making this `FIXME(1.0,linux)` so we know it's a breaking change we want to do?

ci/docker/i686-unknown-linux-musl/Dockerfile
```diff
@@ -1,9 +1,6 @@
-FROM ubuntu:23.10
+FROM ubuntu:25.04
 
-# FIXME(time): we are using an EOL release because 24.04 changes to 64-bit time
-RUN sed -i -E 's/(archive|security)\.ubuntu\.com/old-releases.ubuntu.com/g' \
-        /etc/apt/sources.list && \
-    dpkg --add-architecture i386 && \
+RUN dpkg --add-architecture i386 && \
```
tgross35, line 3: I'd love to do this for all the 32-bit linux targets, any reason why it's only i686 here?

Marcondiro: I tried to do the same for armhf but it was failing due tothe transition to 64 bit time.
Looking around I found a Debian page about the transition towards 64 bit time where they say that x86 ABI wont change while arm got the ABI update. So, the `# FIXME` here is indeed correct for 32 bit architectures excluding x86.

tgross35: Oh, so x86-32 didn't actually do the 64-bit time_t update yet? If so then this makes sense.

Marcondiro: Yep, correct. As far as I understood, Debian x86 32 will never get the 64 bit time_t.
Since they no longer provide x86 ISOs, the x86 32 stuff is there mainly for retrocompatibility so they preferred to keep it as is.

tgross35: Oh that's right, I did hear about that for Debian. In that case I think this is probably close to ready if you want to mark it ready for review.

libc-test/build.rs
Marcondiro: sure, done

PR #4882, opened 2025-12-12
links old version of cf{g,s}et{i,o}speed for glibc

usamoi:
# Description

~Fixes~ https://github.com/rust-lang/libc/issues/4692.

# Sources

https://github.com/bminor/glibc

<details>

<summary>cfsetspeed</summary>

```
./sysdeps/mach/hurd/i386/libc.abilist:GLIBC_2.2.6 cfsetspeed
./sysdeps/mach/hurd/x86_64/libc.abilist:GLIBC_2.38 cfsetspeed
./sysdeps/unix/sysv/linux/aarch64/libc.abilist:GLIBC_2.17 cfsetspeed
./sysdeps/unix/sysv/linux/aarch64/libc.abilist:GLIBC_2.42 cfsetspeed
./sysdeps/unix/sysv/linux/alpha/libc.abilist:GLIBC_2.0 cfsetspeed
./sysdeps/unix/sysv/linux/alpha/libc.abilist:GLIBC_2.42 cfsetspeed
./sysdeps/unix/sysv/linux/arc/libc.abilist:GLIBC_2.32 cfsetspeed
./sysdeps/unix/sysv/linux/arc/libc.abilist:GLIBC_2.42 cfsetspeed
./sysdeps/unix/sysv/linux/arm/be/libc.abilist:GLIBC_2.4 cfsetspeed
./sysdeps/unix/sysv/linux/arm/be/libc.abilist:GLIBC_2.42 cfsetspeed
./sysdeps/unix/sysv/linux/arm/le/libc.abilist:GLIBC_2.4 cfsetspeed
./sysdeps/unix/sysv/linux/arm/le/libc.abilist:GLIBC_2.42 cfsetspeed
./sysdeps/unix/sysv/linux/csky/libc.abilist:GLIBC_2.29 cfsetspeed
./sysdeps/unix/sysv/linux/csky/libc.abilist:GLIBC_2.42 cfsetspeed
./sysdeps/unix/sysv/linux/hppa/libc.abilist:GLIBC_2.2 cfsetspeed
./sysdeps/unix/sysv/linux/hppa/libc.abilist:GLIBC_2.42 cfsetspeed
./sysdeps/unix/sysv/linux/i386/libc.abilist:GLIBC_2.0 cfsetspeed
./sysdeps/unix/sysv/linux/i386/libc.abilist:GLIBC_2.42 cfsetspeed
[...]

PR #4938, opened 2026-01-24
glibc: Update symbol versions for `tc{get,set}attr`

usamoi:
# Description

Followup of https://github.com/rust-lang/libc/pull/4882.

Thanks to @de-vri-es for the reminder, I noticed that due to changes in the `termios` structure, mips(64) and sparc(64) gained new versions of the `tcgetattr` and `tcsetattr` symbols in GLIBC 2.42. This PR also binds versions for the `tcgetattr` and `tcsetattr` symbols on mips(64) and sparc(64). This PR fixes tests on mips(64) and sparc(64).

In addition, I noticed that Rust also supports several tier-3 targets with `target_arch` set to `mips32r6` and `mips64r6`. This PR applies the previous fix to these two architectures as well.

# Sources

https://github.com/bminor/glibc

# Checklist

- [x] Tested locally (on sparc64 and mips64, with qemu and crosstool-ng)

### DIFF
--- a/libc-test/tests/linux_gnu_baud.rs
+++ b/libc-test/tests/linux_gnu_baud.rs
@@ -13,21 +13,24 @@ fn baud() {
     assert!(ret >= 0);
     let terminal_fd = unsafe { open(buffer.as_ptr(), O_RDWR | O_NOCTTY) };
     assert!(terminal_fd >= 0);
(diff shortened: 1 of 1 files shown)


usamoi:
> any reason this is marked as a draft?

> I need some time to set up the toolchain, so this PR will be marked as draft before testing

PR #4942, opened 2026-01-27
musl: introduce musl_stat_timespec cfg for struct stat

heiher:
# Description

Some musl targets use out-of-line timespec fields in struct stat. This property is independent of the `musl_v1_2_3` switch.

Add a new `musl_stat_timespec` cfg and use it to control the presence of inline time fields in musl struct stat definitions, avoiding incorrect API assumptions based on musl version alone.

# Sources

- https://github.com/kraj/musl/tree/v1.2.5

# Checklist

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -28,6 +28,7 @@ const ALLOWED_CFGS: &[&str] = &[
     // Corresponds to `__USE_TIME_BITS64` in UAPI
     "linux_time_bits64",
     "musl_v1_2_3",
+    "musl_stat_timespec",
     // Corresponds to `_REDIR_TIME64` in musl
     "musl32_time64",
     "vxworks_lt_25_09",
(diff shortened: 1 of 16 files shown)


xbjfk:
I would personally push for using `timespec` to:
 - Closer match the libc source 
 - Be POSIX correct and match the correct C usage / docs / man pages.
 - Have more compatibility with other architectures (hermit, AIX and hurd are already)
 - Make struct definitions easier to read (especially with mixed-endian architecture)
 - We can remove a lot of test skips.

I personally think the best path forward would be to revert 55fa65b3e026594c6418eb2bc8e98c1feb7667da and reintroduce the name change across all platforms in the 1.0 branch only for the reasons listed above (virtually every UNIX platform supported by libc uses timespec).

IIRC this was changed with time64 changes the because having conditional definitions for the padding was a bit unwieldy.

Will look into making a PR for this.

PR #4946, opened 2026-01-27
Add siginfo_t::si_status

Thomasdezeeuw:
# Description

Adds `siginfo_t::si_status`.

# Sources

Definition of `siginfo_t`: https://github.com/openbsd/src/blob/7a3d6f69f0a9b1766a54f549b550113ea4d460f1/sys/sys/siginfo.h#L133-L174

Accessing `si_status`: https://github.com/openbsd/src/blob/7a3d6f69f0a9b1766a54f549b550113ea4d460f1/sys/sys/siginfo.h#L179

# Checklist

### DIFF
--- a/src/unix/bsd/netbsdlike/openbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/openbsd/mod.rs
@@ -770,10 +770,9 @@ impl siginfo_t {
         #[repr(C)]
         struct siginfo_proc {
             _si_signo: c_int,
-            _si_errno: c_int,
             _si_code: c_int,
-            #[cfg(target_pointer_width = "64")]
-            __pad1: Padding<c_int>,
+            _si_errno: c_int,
+            _pad: Padding<[c_int; SI_PAD]>,
             _pid: crate::pid_t,
             _uid: crate::uid_t,
             _utime: crate::clock_t,


src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -765,6 +765,23 @@ impl siginfo_t {
         }
         (*(self as *const siginfo_t).cast::<siginfo_timer>()).value
     }
+
+    pub unsafe fn si_status(&self) -> c_int {
+        #[repr(C)]
+        struct siginfo_proc {
+            _si_signo: c_int,
+            _si_errno: c_int,
+            _si_code: c_int,
```
semarie, line 773: the order is wrong: first `si_code`, next `si_errno`. no a big deal (as we only use the size of the field) but it will be better to be right in case of copy/pasting

src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -765,6 +765,23 @@ impl siginfo_t {
...
+            #[cfg(target_pointer_width = "64")]
```
semarie: no need of `cfg()`

src/unix/bsd/netbsdlike/openbsd/mod.rs
```diff
@@ -765,6 +765,23 @@ impl siginfo_t {
...
+            __pad1: Padding<c_int>,
```
semarie: wrong. use ` _pad: Padding<[c_int; SI_PAD]>,`

semarie (approved):
LGTM thanks

PR #4948, opened 2026-01-27
Fix const for field `tm_zone` in `struct tm`

semarie:
# Description

The PR makes the CI to ignore `tm_zone` field in `struct tm`. The const has been recently added in OpenBSD (see sources).

The another way is to properly define `struct tm` in OpenBSD, and move the currently global definition to each unix platform. It might be preferable. Feel free to ask for this way :smiley: 

# Sources

- https://github.com/openbsd/src/commit/4a796cf042d7cd180a2f093282195306a54bbdc4
- https://github.com/openbsd/src/blob/7a3d6f69f0a9b1766a54f549b550113ea4d460f1/include/time.h#L98

# Checklist

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -262,6 +262,20 @@ s! {
         __unused8: Padding<*mut c_void>,
     }
 
+    pub struct tm {
+        pub tm_sec: c_int,
+        pub tm_min: c_int,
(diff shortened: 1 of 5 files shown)


semarie:
the change (on OpenBSD side) is for upcoming OpenBSD-7.9

JohnTitor (approved):
Cool, thank you!

PR #4954, opened 2026-01-30
RFC 2292 struct and constant definitions

Skyb0rg007:
# Description

This PR adds the structs and constants as defined by RFC 2292 - Advanced Sockets API for IPv6.
These constants are available for use in get/setsockopt on supported Unix platforms.
The operations defined on the icmp6_filter struct are defined as C macros, so the implementations are not included here.

# Sources

https://github.com/bminor/glibc/blob/04e750e75b73957cf1c791535a3f4319534a52fc/inet/netinet/icmp6.h
https://github.com/kraj/musl/blob/kraj/master/include/netinet/icmp6.h
https://github.com/freebsd/freebsd-src/blob/9ae367d11de8abbdf53884836c9ba30908c5c8db/sys/netinet/icmp6.h

Also in the RFC: https://datatracker.ietf.org/doc/html/rfc2292#section-3.2

# Checklist

Unfortunately I could not get the libc-tests working locally, so this PR is marked as draft.


Skyb0rg007:
Alright, I need help. On *BSD platforms (FreeBSD, macOS, etc), `struct in6_addr` is defined as the union of `uint32_t[4]`, `uint16_t[8]`, and `uint8_t[16]`, so Rust encodes it as `[u8; 16]` with `repr(align(4))`.

But the `nd_(router|neighbor)_(solicit|advert)` structures are marked as `__packed` (so their C alignment is 1), even though they include a `in6_addr`. Rust won't allow using `repr(packed)` on a struct containing a `repr(align(...))` attribute, so I'm stuck.

Linux doesn't mark these structs as `__packed`, so there's no issue there.

Skyb0rg007:
For now, I simply guard the offending struct definitions with `#[cfg(...)]`. I still think the `src/unix/mod.rs` file is the right place (instead of one of the platform-specific subdirectories) because these structures and definitions are specified by network protocols. I think the three "complete" solutions would be:

1. Add `struct ip6_addr_packed` without an alignment attribute. Use that for these fields.
2. Use `[u8; 16]` instead of `ip6_addr` for these platforms.
3. Break alignment compatibility with these structures on the offending platforms. From the FreeBSD documentation, the use of `__packed` is used in network protocol code to avoid padding bytes, not to allow non-aligned data.

Otherwise I think this PR is ready to review.

src/unix/mod.rs
```diff
@@ -219,6 +219,231 @@ s! {
...
+    #[cfg_attr(any(target_os = "linux", target_os = "emscripten"), repr(align(4)))]
+    pub struct ip6_rthdr0 {
+        pub ip6r0_nxt: u8,
+        pub ip6r0_len: u8,
+        pub ip6r0_type: u8,
+        pub ip6r0_segleft: u8,
+        #[cfg(not(any(target_os = "linux", target_os = "emscripten")))]
+        pub ip6r0_reserved: u32,
+        #[cfg(any(target_os = "linux", target_os = "emscripten"))]
+        pub ip6r0_reserved: u8,
```
tgross35, line 259: For any reserved/padding fields, please make them private and wrap them in `Padding`

src/unix/mod.rs
```diff
@@ -219,6 +219,231 @@ s! {
...
+        ip6f_reserved: u8,
+        ip6f_offlg: u16,
+        ip6f_ident: u32,
+    }
+
+    pub struct icmp6_filter {
+        #[cfg(not(target_os = "solaris"))]
+        pub icmp6_filt: [u32; 8],
+        #[cfg(target_os = "solaris")]
+        pub __icmp6_filt: [u32; 8],
```
tgross35: This struct should be opaque, can you make the fields private?

src/unix/mod.rs
tgross35, line 1: I think enabling these in the `unix` module is too high level; do we know anything about AIX, fuchsia, redox, etc?

Instead, could you put these into the `linux_like` and `bsd` modules? That will mean some duplication, but that's probably fine because it will simplify some of the repetitive padding config.

src/unix/mod.rs
```diff
@@ -219,6 +219,231 @@ s! {
...
+        any(
+            target_os = "android",
+            target_os = "freebsd",
+            target_os = "openbsd",
+            target_os = "dragonfly",
+            target_os = "macos"
+        ),
+        repr(packed)
+    )]
+    #[cfg_attr(any(target_os = "linux", target_os = "emscripten"), repr(align(4)))]
```
tgross35, line 250: Could you add links to the PR description to the headers for these other platforms that have the specific alignment requirements? (Looks like only glibc/musl/freebsd are there)

tgross35 (changes requested):
Sorry for the delay, thank you for the changes here. The contents themselves look good, but I think they should be moved out of the root unix module.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/4954/34527ca4c0b43f10062a810789953a538a76ed7d..a96955a6aa721a1517665e359e7f1b4753f271f9)*

src/unix/mod.rs
tgross35: What would be ideal here is to instead put these in `src/new/common/bsd/{icmp6.rs,ip6.rs}` and  `src/new/common/linux_like/{icmp6.rs,ip6.rs}`, to match the header structure. The reexport them in `src/new/{bionic_libc,musl,glibc,openbsd,freebsd,...}/{icmp6.rs,ip6.rs}`, similar to what is done with unistd, and update `src/new/mod.rs`.

I won't require that because this has already been waiting a while, but eventually everything will move to that new structure so anything to help that is appreciated.

src/unix/mod.rs
Skyb0rg007: Because these fields are part of a networking protocol, I don't think it makes sense to make these private. These fields were certainly just padding when the protocol was written, but in theory a new version of the IPv6 routing header could make use of these fields.

src/unix/mod.rs
Skyb0rg007: The macros `ICMP6_FILTER_SET(PASS|BLOCK)(ALL)?` need to be defined for this type. Is the best way to handle this to make the fields `pub (crate)` and implement those macros as toplevel Rust functions?

src/unix/mod.rs
Skyb0rg007: Do you mean link to this PR, or directly link to the glibc/freebsd headers in the comment

src/unix/mod.rs
Skyb0rg007: I'm not sure about AIX, but Fuchsia does support [icmp6_filter](https://fuchsia.dev/reference/cpp/struct/icmp6-filter) and RedoxOS (relibc) does not, at least for now.

I can try refactoring the code into those two files, but a lot of the differences with respect to alignment doesn't have a clean linux/bsd divide with android's implementation being more similar to that of the BSDs.
We might want to come up with a solution to the problem I remarked on [here](https://github.com/rust-lang/libc/pull/4954#issuecomment-3843534209) before I add definitions of these structs, because any solution will possibly change the public API for these structs anyways.

Also, I'm fine with a more major structural change to properly model the addition.
I have some more additions from RFC 3542 (the follow-on to 2292) which I plan to make a PR for after this is accepted. It just happens that the feature I was personally missing from libc was specified in the first Advanced IPv6 Sockets API RFC.

src/unix/mod.rs
tgross35: Just in the PR description, and at least for the tier 2 targets (Android, emscripten, MacOS). I'm just looking for something to compare the aligned/packed/etc config against.

src/unix/mod.rs
tgross35: I think it's okay to omit Fuchsia and others that aren't covered in CI for now. If a need comes up then usually one of their target maintainers does the addition, which also means they can run tests.

Could you split to BSD and Linux-like, then add a `cfg(not(target_os = "android"))` to the Linux version? I think that would still be cleaner.

> Also, I'm fine with a more major structural change to properly model the addition. I have some more additions from RFC 3542 (the follow-on to 2292) which I plan to make a PR for after this is accepted. It just happens that the feature I was personally missing from libc was specified in the first Advanced IPv6 Sockets API RFC.

If you have more followups here then using the new structure might be nice, it's certainly more organized than the status quo. Up to you.

src/unix/mod.rs
tgross35: Is there any use reading/writing them currently? My concern here is that if they do wind up being used, the fields may go through changes in C that we can't really mirror on the Rust side. E.g. they could rename the fields, possibly with a shim like `#define ip6r0_new_nice_field_name ip6r0_reserved`, or do something like split the `u32` field into two `u16`s which is fine in C but breaking in Rust.

Open to thoughts but I'd rather err on the side of caution and make them private unless they're useful now.

src/unix/mod.rs
tgross35: Seems like you may have started work here, but that seems reasonable to me.

Skyb0rg007:
I made a PR for the ICMP6_FILTER macros; I'll rebase on top of that once its accepted.

PR #4959, opened 2026-02-05
Macro cleanups

nnethercote:
I was building rustc with `-Zmacro-stats`, which drew my attention to the `s!` macro, and I saw some improvements could be made.


tgross35:
Most of this looks good but the deprecated change seems to be causing the CI failure. Not sure if there is a good solution there.

tgross35 (approved):
Thank you!

PR #4960, opened 2026-02-09
ci: Ensure TARGET_REF is set during scheduled jobs

tgross35:
This was causing a failure in prep-semver-baseline.sh for the daily cache-building job since there is no base for jobs that run on a branch.


JohnTitor (approved):
Makes sense 👍

PR #4966, opened 2026-02-10
linux: update ptrace_syscall_info struct

Marcondiro:
Hi,

# Description

linux: update `ptrace_syscall_info struct`:
- Add new `flags` field and reserveds.

The updated version has a different size on 32 bits arches (84 -> 88 bytes) while the size stays the same on 64 bits.

# Sources

[ptrace.h](https://github.com/torvalds/linux/blob/72c395024dac5e215136cbff793455f065603b06/include/uapi/linux/ptrace.h#L83-L106)
[kernel lore](https://lore.kernel.org/lkml/20250303112044.GF24170@strace.io/)

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -4150,6 +4150,10 @@ fn test_linux(target: &str) {
             // FIXME(musl): New fields in newer versions
             "utmpx" if !old_musl => true,
 
+            // FIXME(linux): Requires >= 6.16 kernel headers.
+            // On 64 bits the size did not change, skip only for 32 bits.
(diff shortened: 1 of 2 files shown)


tgross35:
Could you split the `rval` rename from the rest? I'd actually be okay to backport the flags and `reserved` portion since we don't guarantee anything about padding/reserved fields. Just not the renaming since that's an actual break.

@rustbot author

Marcondiro:
Sure, I've just opened #5016 for the renaming, thanks

@rustbot ready

Issue #4975, opened 2026-02-14
Implement `Default`


tgross35:
Discussed at [#t-libs > Questions and polls about &#96;libc&#96; 1.0 @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/Questions.20and.20polls.20about.20.60libc.60.201.2E0/near/511749874), we would like to have a `Default` impl for pretty much everything. This should be part of `s!`/`s_no_extra_traits!` but won't work out of the box.

telcharr:
Mind if I take this? I messed around with a prototype `impl_default!` that generates the impl field by field. `Default::default()` for normal fields, `#[custom_default(...)]` for arrays larger than 32, and `mem::zeroed()` for unions. I also got `s!` dispatching off a `#[skip_derive_default]` marker. Tested it on a plain struct, the big arrays, a union, and a struct containing a union, all passing.

Two questions. Is zeroing unions ok when there's no `custom_default`? Seems right for the plain C unions but wanted to check. Also, do you want Default on everything by default eventually, or keep it opt-in per struct? Can't really flip it on globally without breaking every big-array struct until they're annotated, so I've got it opt-in for now.

Only limitation is that `custom_default` has to be the first attribute on a field since the macro matches literally. Doesn't seem like it'd be a problem, but worth mentioning.

tgross35:
> Mind if I take this? I messed around with a prototype `impl_default!` that generates the impl field by field. `Default::default()` for normal fields, `#[custom_default(...)]` for arrays larger than 32, and `mem::zeroed()` for unions. I also got `s!` dispatching off a `#[skip_derive_default]` marker. Tested it on a plain struct, the big arrays, a union, and a struct containing a union, all passing.

Help would be very welcome! Especially if you have it working already, that's amazing.

> Two questions. Is zeroing unions ok when there's no `custom_default`? Seems right for the plain C unions but wanted to check. 

In general using `mem::zeroed` should be okay for all `libc` types, and is probably actually preferable for unions so we don't need to remember what the largest field is. But since this operation still requires asserting that zeros are a valid bitpattern, I'd like to have `unsafe` be needed somewhere in the macro.

Could something like this work in the macro?

```rust
#[unsafe(default_via_zeroed)]
union bar {
    a: i32,
    b: f32,
}
```

I'd recommend doing that later though, and just starting with `#[custom_default(unsafe { mem::zeroed<some_union>() })]` the few places it's needed.

[...]

PR #4976, opened 2026-02-14
Implement `Default` for `timeval` and `timespec`

tgross35:
With the experimental time64 support, some of these structs gain a padding field on 32-bit platforms. We are planning to give everything a `Default` impl anyway ([1]) so start with these to ease the transition.

[1]: https://github.com/rust-lang/libc/issues/4975


JohnTitor (approved):
👍

PR #4977, opened 2026-02-14
build: Replace `RUST_LIBC_UNSTABLE` env with `libc_unstable*` cfg

tgross35:
The current environment variable setup makes it easy to set a
configuration for both the host (build scripts) and target, even when
that is not the intent. Change to a rustc cfg, which is easier to set
for only the target.

This was briefly discussed on Zulip around [1].

[1]: https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/Questions.20and.20polls.20about.20.60libc.60.201.2E0/near/511723332

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -195,30 +195,30 @@ jobs:
           - target: aarch64-linux-android
           - target: aarch64-unknown-linux-musl
           - target: aarch64-unknown-linux-musl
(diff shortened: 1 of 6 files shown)


tgross35:
Woo I think this works finally. FreeBSD is the only remaining `RUST_LIBC_UNSTABLE` flag, which is taken care of in https://github.com/rust-lang/libc/pull/5201.

tgross35:
I'm going to go ahead since this seems functional, and there are a few things that want to stack on top of it

PR #4983, opened 2026-02-18
ci: Test powerpc64-unknown-linux-musl in tier 2

Gelbpunkt:
# Description

Since 1.95.0, this is now a tier 2 target.

# Sources

# Checklist


tgross35:
Does linking with musl-gcc happen to work? I think it's in the `musl-tools` package and might be a bit easier.

Any idea if debian is going to move to elfv2 at some point? It seems like support for elfv1 is pretty poor at this point.

ci/docker/powerpc64-unknown-linux-musl/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+    libc6-dev \
+    make \
+    patch \
+    qemu-user \
+    rsync \
+    xz-utils
+
+ARG MUSL_VERSION
+COPY install-musl.sh /
+COPY compiler_builtins.rlib /
```
tgross35: I think it would be better to just download in the dockerfile rather than copying from the sysroot, to keep things more self-contained. It should be possible to grab  https://static.rust-lang.org/dist/2026-06-26/rust-std-nightly-powerpc64-unknown-linux-musl.tar.xz, looks like 26MB

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/4983/0d9bc6e6c21f85254cb0b339eec42527675ef620..503d11809cb50c5549ed5beb9a8ee36101fb34e7)*

ci/install-musl.sh
```diff
@@ -83,7 +83,7 @@ case ${1} in
     powerpc64*)
         musl_arch=powerpc64
         kernel_arch=powerpc
-        CC="${1}-linux-gnu-gcc" CFLAGS="-mlong-double-64" \
+        CC="${1}-linux-gnu-gcc" CFLAGS="-mlong-double-64 -mabi=elfv2" LIBCC="/compiler_builtins.rlib" \
             ./configure --prefix="/musl-${musl_arch}" --enable-wrapper=yes
         make install -j4
```
tgross35, line 99: This special config is only actually needed for BE right? Might be worth splitting this block, especially if downloading it only on the one target.

Please add a comment too

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/4983/0d9bc6e6c21f85254cb0b339eec42527675ef620..503d11809cb50c5549ed5beb9a8ee36101fb34e7)*

ci/docker/powerpc64-unknown-linux-musl/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN /install-musl.sh powerpc64 "$MUSL_VERSION"
+
+# Rust's unwind requires a libgcc_s.so.1
+# We instead get the unwinding symbols from the self-contained libunwind.a
+# since that's included in the toolchain
+RUN touch /musl-powerpc64/lib/libgcc_s.so.1
```
tgross35, line 27: Could you note that rustc links it unconditionally? Hence why we need the dummy.

Somebody on Zulip might know more than me about if there's a better solution here

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/4983/0d9bc6e6c21f85254cb0b339eec42527675ef620..503d11809cb50c5549ed5beb9a8ee36101fb34e7)*

tgross35 (changes requested):

tgross35 (approved):
Looks great, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/4983/d59583409474a46bd91c535dcf903fd041a3509c..dd8218d8df1d8838aa9329fc9c72112cf067ce32)*

PR #4984, opened 2026-02-18
Re-organized the QNX support.

jonathanpallant:
# Description

As most of the differences are related to the change in network stack half-way through the life of QNX Neutrino 7.1, the definitions are now split into io-pkt (the older stack) and io-sock (the newer stack).

I didn't add any architecture definitions for x86 (as would be required for i686-pc-nto-qnx700) because I don't have access to QNX Neutrino 7.0 to verify what should go here. But there's a clear place to add it later.

I do have QNX Neutrino 7.1 and QNX OS 8.0 and I have verified each of the types in the io-pkt and io-sock modules against the relevant C header files. I have not verified the remaining types or functions - I assume they are correct.

This is an alternative to https://github.com/rust-lang/libc/pull/4615, which seems to have stalled.

# Sources

For io-sock, [QNX says](https://www.qnx.com/developers/docs/8.0/com.qnx.doc.neutrino.io_sock/topic/socket_api.html):

> The io-sock networking stack supports all the routines that application programmers would expect, including (but not limited to) the functions listed below. The FreeBSD 13.5 manual pages at https://www.freebsd.org/cgi/man.cgi are the best guide to their use:

For io-pkt, you can refer to https://www.qnx.com/developers/docs/7.1/.

# Checklist

   - There aren't any for QNX
- [ ] No placeholder or unstable values like `*LAST` or `*MAX` are included (see [#3131](https://github.com/rust-lang/libc/issues/3131))
   - I left all the existing ones
   - I tried running this locally and I got: *not implemented: Foreign functions are unlikely to have any other pattern.* I don't know if we normally run the tests for this target.

### DIFF
--- a/src/unix/nto/io_pkt/mod.rs
+++ b/src/unix/nto/io_pkt/mod.rs
@@ -1,0 +1,160 @@
+//! Definitions specific to QNX versions using io-pkt networking
+//!
+//! This module applies to:
+//!
+//! * `aarch64-unknown-nto-qnx700`
+//! * `i686-pc-nto-qnx700`
+//! * `aarch64-unknown-nto-qnx710`
+//! * `x86_64-pc-nto-qnx710`
+
+use crate::prelude::*;
+
+s! {
+    #[repr(packed)]
+    pub struct in_addr {
+        pub s_addr: crate::in_addr_t,
+    }
+
+    pub struct sockaddr_in {
+        pub sin_len: u8,
+        pub sin_family: crate::sa_family_t,
+        pub sin_port: crate::in_port_t,
+        pub sin_addr: crate::in_addr,
+        pub sin_zero: [i8; 8],
+    }
+
+    pub struct in_pktinfo {
+        pub ipi_addr: crate::in_addr,
+        pub ipi_ifindex: c_uint,
+    }
+
+    #[repr(packed)]
+    pub struct arphdr {
+        pub ar_hrd: u16,
+        pub ar_pro: u16,
+        pub ar_hln: u8,
+        pub ar_pln: u8,
+        pub ar_op: u16,
+    }
+
+    pub struct unpcbid {
+        pub unp_pid: crate::pid_t,
+        pub unp_euid: crate::uid_t,
+        pub unp_egid: crate::gid_t,
+    }
+
+    pub struct mmsghdr {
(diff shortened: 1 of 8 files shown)


src/unix/nto/io_pkt/mod.rs
tgross35, line 1: Do `io_pkt` and `io_sock` match up with specific headers or directories? If so, they can go in `src/new/nto` which is where we're migrating files that closer match up with source. (Ideally also add a tag like `//! Header: `uapi/linux/can.h` to make it easy to match up). You can re-order things to be in the same order as the source headers too if it's easier to follow.

I assume `neutrino.rs` could probably also move here

jonathanpallant: There are two trees, one for io-pkt and one for io-sock. They contain files with the same names, and the file contents are similar but slightly different.

I guess there's a trade-off between having the Rust code faithfully match the C headers, but with a lot of duplication. Or, as I have done here, have one Rust file that covers all the common elements, and then have only their differences pulled out to their own files.

tgross35 (approved):
I'm sorry this has been sitting for such a long amount of time, we've been stretched on review power for a little while. This will need a rebase but I have no concerns here.

An optional change, if you're interested, is to move `arch/*`, `io_pkt`, `io_sock`, and `neutrino.rs` to `src/new/nto` since I believe those modules are well-organized at this point. Otherwise I can do it later.

@rustbot author for the rebase

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/4984/a41bc082ed4e42d8817caadfc82fabed7f6fd9bc..f3a43cfa20f54cac3e1161552a61013146e3d4c8)*

jonathanpallant:
I'll leave the moving around of files for another time, because with https://github.com/rust-lang/rust/pull/158449 we might want to do something different.

Looking at the rebase now.

jonathanpallant:
I'm having a bit of trouble running `ci/style.py` locally because my Linux is quite old (Kernel 5.15) and some CAN_xxx constants are missing?

But I think I worked out it wanted the imports at the bottom of the file, not the top.

tgross35:
I need to fix the style tests so they don't run `ctest`, that's kind of silly (artifact of ctest mostly happening in build scripts), but LGTM. Thanks!

PR #4988, opened 2026-02-21
Linux: add `name_to_handle_at` and `open_by_handle_at`

samu698:
# Description

Add file handle operations for Linux.

In particular I added:
- `name_to_handle_at()`
- `open_by_handle_at()`
- `file_handle`
- `MAX_HANDLE_SZ`
- `AT_HANDLE_MNT_ID_UNIQUE`
- `AT_HANDLE_FID`
- `AT_HANDLE_CONNECTABLE`

# Sources

- [`file_handle` struct (glibc)](https://github.com/bminor/glibc/blob/04e750e75b73957cf1c791535a3f4319534a52fc/sysdeps/unix/sysv/linux/bits/fcntl-linux.h#L364-L373)
- [Flags (glibc)](https://github.com/bminor/glibc/blob/04e750e75b73957cf1c791535a3f4319534a52fc/sysdeps/unix/sysv/linux/bits/fcntl-linux.h#L380-L387)
- [Functions (glibc)](https://github.com/bminor/glibc/blob/04e750e75b73957cf1c791535a3f4319534a52fc/sysdeps/unix/sysv/linux/bits/fcntl-linux.h#L455-L464)

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -3671,6 +3671,7 @@ fn test_linux(target: &str) {
     }
 
     let arm = target.contains("arm");
+    let eabihf = target.contains("eabihf");
     let aarch64 = target.contains("aarch64");
     let i686 = target.contains("i686");
     let ppc = target.contains("powerpc");
@@ -4487,6 +4488,16 @@ fn test_linux(target: &str) {
             // FIXME(linux):  Requires >= 6.16 kernel headers.
             "PTRACE_SET_SYSCALL_INFO" => true,
 
+            // FIXME(linux): Requires >= 6.13 kernel headers.
+            "AT_HANDLE_CONNECTABLE" => true,
+
+            // FIXME(linux): Requires >= 6.12 kernel headers.
+            "AT_HANDLE_MNT_ID_UNIQUE" => true,
+
+            // FIXME(musl): This value is not yet in musl.
+            // eabihf targets are tested using an older version of glibc
+            "AT_HANDLE_FID" if musl || eabihf => true,
+
             _ => false,
         }
     });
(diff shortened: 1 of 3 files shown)


samu698:
This PR is missing the flags `AT_HANDLE_FID`, `AT_HANDLE_MNT_ID_UNIQUE`, `AT_HANDLE_CONNECTABLE` because the versions of glibc used in CI are outdated and don't contain these flags, also musl never added these flags.

Also CI fails with musl because the struct file_handle is defined as a flexible array and the size check fails.

So I have a few questions about this PR
- Should I put the flag definitions in the linux/gnu directory (even if they are valid to use on linux systems)
- Should I exclude the file_handle struct from the musl tests

libc-test/build.rs
```diff
@@ -4487,6 +4491,16 @@ fn test_linux(target: &str) {
             // FIXME(linux):  Requires >= 6.16 kernel headers.
             "PTRACE_SET_SYSCALL_INFO" => true,
 
+            // FIXME(linux): Requires >= 6.13 kernel headers.
+            "AT_HANDLE_CONNECTABLE" => true,
+
+            // FIXME(linux): Requires >= 6.12 kernel headers.
+            "AT_HANDLE_MNT_ID_UNIQUE" => true,
```
samu698, line 4495: These flags are a recent addition to Linux and they are present only in the more recent versions of glibc

libc-test/build.rs
```diff
@@ -4487,6 +4491,16 @@ fn test_linux(target: &str) {
...
+
+            // FIXME(musl): This value is not yet in musl.
+            // eabihf targets are tested using an older version of glibc
+            "AT_HANDLE_FID" if musl || eabihf => true,
```
samu698, line 4499: eabihf targets use an older version of glibc in CI, where this flag wasn't added

libc-test/build.rs
```diff
@@ -4142,6 +4143,10 @@ fn test_linux(target: &str) {
             // FIXME(musl): New fields in newer versions
             "utmpx" if !old_musl => true,
 
+            // FIXME(musl): This struct is contains a flexible array field that causes the size
+            // check to fail
+            "file_handle" if musl => true,
```
tgross35: Put this in `skip_struct_field` instead so we don't miss the whole thing (can be grouped under the existing `// FAM` comment)

PR #4989, opened 2026-02-21
Correct the definition of `pthread_t` on Darwin

### LINKED ISSUE #2903: Incorrect `pthread_t` definition on Darwin (should be a raw pointer, not an `uintptr_t`)

chengr4:
# Description

This PR fixes rust-lang/libc#2903. This PR picks up the work started by @highjeans in rust-lang/libc#4336 (close rust-lang/libc#4336) which had stalled. I just followed up the comments from @tgross35 in the PR and tried to finish it.

> Thanks for leaving a great start point to study this crate.

Closes: https://github.com/rust-lang/libc/pull/4336

# Sources

- https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/EXTERNAL_HEADERS/sys/_pthread/_pthread_types.h#L103
- https://github.com/apple-oss-distributions/libpthread/blob/main/include/sys/_pthread/_pthread_t.h#L31

# Checklist

### DIFF
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -5,7 +5,22 @@ pub type useconds_t = u32;
 pub type blkcnt_t = i64;
 pub type socklen_t = u32;
 pub type sa_family_t = u8;
-pub type pthread_t = crate::uintptr_t;
+
+cfg_if! {
+    if #[cfg(target_vendor = "apple")] {
+        extern_ty! {
+            pub type _opaque_pthread_t;
+        }
+
+        type __darwin_pthread_t = *mut _opaque_pthread_t;
+
+        pub type pthread_t = __darwin_pthread_t;
+    } else {
+        // FIXME(1.0): verify other BSDs? Glancing around, other BSDs also look like a pointer
+        pub type pthread_t = crate::uintptr_t;
+    }
+}
+
 pub type nfds_t = c_uint;
 #[cfg(target_os = "dragonfly")]
 pub type regoff_t = c_int;
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -5,7 +5,22 @@ pub type useconds_t = u32;
 pub type blkcnt_t = i64;
 pub type socklen_t = u32;
 pub type sa_family_t = u8;
-pub type pthread_t = crate::uintptr_t;
+
+cfg_if! {
+    if #[cfg(target_vendor = "apple")] {
+        extern_ty! {
+            pub type _opaque_pthread_t;
+        }
+
+        type __darwin_pthread_t = *mut _opaque_pthread_t;
+
+        pub type pthread_t = __darwin_pthread_t;
+    } else {
+        // FIXME(1.0): verify other BSDs? Glancing around, other BSDs also look like a pointer
+        pub type pthread_t = crate::uintptr_t;
+    }
+}
+
 pub type nfds_t = c_uint;
 pub type regoff_t = off_t;
 
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -9,6 +9,9 @@ pub type sa_family_t = u8;
 cfg_if! {
     if #[cfg(target_vendor = "apple")] {
         extern_ty! {
+            // Internal definition that must be public because it appears in `pthread_t`, but is
+            // not meant to be user-facing.
+            #[doc(hidden)]
             pub type _opaque_pthread_t;
         }
 
--- a/libc-test/build/main.rs
+++ b/libc-test/build/main.rs
@@ -262,7 +262,7 @@ fn test_apple(t: &Target) {
     cfg.skip_struct(move |s| {
         match s.ident() {
             // Extern types
-            "DIR" | "FILE" | "fpos_t" | "timezone" => true,
+            "DIR" | "FILE" | "fpos_t" | "timezone" | "_opaque_pthread_t" => true,
 
             // FIXME(macos): The size is changed in recent macOSes.
             "malloc_zone_t" => true,
--- a/libc-test/semver/apple.txt
+++ b/libc-test/semver/apple.txt
@@ -1798,6 +1798,7 @@ __PTHREAD_RWLOCKATTR_SIZE__
 __PTHREAD_RWLOCK_SIZE__
 __darwin_mcontext64
 __error
+_opaque_pthread_t
 abs
 acct
 aio_cancel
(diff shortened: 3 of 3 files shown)


chengr4:
@highjeans Thank for answering; I totally forgot 🫠. Again, Thanks for leaving the PR for me to study this crate.

@tgross35 

Actually, I failed `cargo test --test ctest` at my local.

```bash
bad `ELAST` value at byte 0: rust: 106 (0x6a) != c 107 (0x6b)
    rust bytes: 6a 00 00 00
    c bytes:    6b 00 00 00
bad `HOST_VM_INFO64_COUNT` value at byte 0: rust: 38 (0x26) != c 40 (0x28)
    rust bytes: 26 00 00 00
    c bytes:    28 00 00 00
bad `vm_statistics64_data_t` size: rust: 152 != c 160
bad `vm_statistics64` size: rust: 152 != c 160
size of `vm_statistics64_data_t` is 160 in C and 152 in Rust

size of `struct vm_statistics64` is 160 in C and 152 in Rust
```

Can you please guide me how to evaluate and fix this bug?

JohnTitor:
@chengr4 because your system has different version(s) than CI, I think. It's fine as long as CI passes (unless it's unspported).
Could you squash commits into one? Otherwise LGTM on my end.

chengr4:
ok will do it after 6/12

chengr4:
@JohnTitor if commit mesage is not clear enough, help me fix it. Thanks

src/unix/bsd/mod.rs
```diff
@@ -5,7 +5,21 @@ pub type useconds_t = u32;
 pub type blkcnt_t = i64;
 pub type socklen_t = u32;
 pub type sa_family_t = u8;
-pub type pthread_t = crate::uintptr_t;
+
+cfg_if! {
+    if #[cfg(target_vendor = "apple")] {
+        extern_ty! {
+            pub enum _opaque_pthread_t {}
+        }
```
tgross35, line 16: This macro's syntax changed recently, you'll need to rebase then this can become `pub type _opaque_pthread_t;`.

However, could you try making it non-pub? Since we don't care about the internals. You may need a `#[allow(private_interfaces)]` on `pthread_t`.

src/unix/bsd/mod.rs
```diff
@@ -5,7 +5,21 @@ pub type useconds_t = u32;
...
+
+        type __darwin_pthread_t = *mut _opaque_pthread_t;
+        pub type pthread_t = __darwin_pthread_t;
+    } else {
+        // FIXME(1.0): verify other BSDs? Glancing around, other BSDs also look like a pointer
+        pub type pthread_t = crate::uintptr_t;
```
tgross35, line 23: Feel free to change these here since you've already done the work

tgross35 (changes requested):
Regarding the commit: whoever submits the final patch should be listed as the author, which in this case is you. You can change the author field with `git commit --amend --reset-author`.

However, of course you also need to give credit to anyone who contributed to the patch, which you can add with:

```
Co-authored-by: highjeans <foo@bar.com>
```

at the end of the commit (get their email from the commit log before you reset it)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/4989/38c6231f667b698224c5dc4ffa87b7c48e0e2b47..3f7fcfb59d2313d59453ee6cf00b1f91daa458dd)*

src/unix/bsd/mod.rs
chengr4: I can take it over after the apples look good to you.

src/unix/bsd/mod.rs
chengr4: I tried making `_opaque_pthread_t` non-pub, but it caused numerous compiler errors and warnings regarding `private_interfaces`

E.g.

```bash
error: type `_opaque_pthread_t` is private
  --> src/new/apple/libpthread/pthread_/introspection.rs:27:17
   |
27 |         thread: pthread_t,
   |                 ^^^^^^^^^ private type

error: type `_opaque_pthread_t` is private
  --> src/new/apple/libpthread/pthread_/introspection.rs:33:17
   |
33 |         thread: pthread_t,
   |                 ^^^^^^^^^ private type

error: type `_opaque_pthread_t` is private
 --> src/new/apple/libpthread/pthread_/pthread_spis.rs:9:17
  |
9 |         thread: *mut crate::pthread_t,
  |                 ^^^^^^^^^^^^^^^^^^^^^ private type
```

tgross35 (approved):
I had a couple remaining changes but just added a commit, sorry this has lingered. Thank you for the fix!

Marked declined because this is user-facing.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/4989/bd53ab4040b5791ae22417ac77fb0b61736a2313..2bbac8f2825d98a6c4951b016253e8f287c460d1)*

PR #5002, opened 2026-03-04
FreeBSD: Added `xfile` structure and file descriptor types

antoncxx:
# Description

Added binding for `xfile` structure from `sys/file.h` and descriptor types.

# Sources

#### FreeBSD 15
- [xfile](https://github.com/freebsd/freebsd-src/blob/b17655c95c4cb313355de1ffd74cb8a4dcc1a053/sys/sys/file.h#L243)
- [descriptors](https://github.com/freebsd/freebsd-src/blob/b17655c95c4cb313355de1ffd74cb8a4dcc1a053/sys/sys/file.h#L59)

#### FreeBSD 14
- [xfile](https://github.com/freebsd/freebsd-src/blob/e0e1e9e083097383bcebf0b6cf65af46653e2f73/sys/sys/file.h#L234)
- [descriptors](https://github.com/freebsd/freebsd-src/blob/e0e1e9e083097383bcebf0b6cf65af46653e2f73/sys/sys/file.h#L58)

#### FreeBSD 13
- [xfile](https://github.com/freebsd/freebsd-src/blob/55eff489d9cdae5620bf60bb79d996a059404ed2/sys/sys/file.h#L224)
- [descriptors](https://github.com/freebsd/freebsd-src/blob/55eff489d9cdae5620bf60bb79d996a059404ed2/sys/sys/file.h#L58)

#### ~~FreeBSD 12~~
- ~~[xfile](https://github.com/freebsd/freebsd-src/blob/9aed7107bbfbe9f919b547d68e9cd228a1a1852c/sys/sys/file.h#L217)~~
- ~~[descriptors](https://github.com/freebsd/freebsd-src/blob/9aed7107bbfbe9f919b547d68e9cd228a1a1852c/sys/sys/file.h#L58)~~

#### ~~FreeBSD 11~~
- ~~[xfile](https://github.com/freebsd/freebsd-src/blob/5ed7eb0d97ba4436218e810f61bd059acba984c2/sys/sys/file.h#L209)~~
- ~~[descriptors](https://github.com/freebsd/freebsd-src/blob/5ed7eb0d97ba4436218e810f61bd059acba984c2/sys/sys/file.h#L56)~~

#### ~~FreeBSD 10~~
- ~~[xfile](https://github.com/freebsd/freebsd-src/blob/28cce1d5b5f5636459f6fa4b99d32139af8ecd2b/sys/sys/file.h#L198)~~
- ~~[descriptors](https://github.com/freebsd/freebsd-src/blob/28cce1d5b5f5636459f6fa4b99d32139af8ecd2b/sys/sys/file.h#L56)~~

# Checklist

### DIFF
--- a/src/new/freebsd/sys/file.rs
+++ b/src/new/freebsd/sys/file.rs
@@ -1,0 +1,46 @@
+//! Header: `sys/file.h`
+//!
+//! https://github.com/freebsd/freebsd-src/blob/main/sys/sys/file.h
+
+use crate::prelude::*;
+
+pub const DTYPE_NONE: c_int = 0;
+pub const DTYPE_VNODE: c_int = 1;
+pub const DTYPE_SOCKET: c_int = 2;
+pub const DTYPE_PIPE: c_int = 3;
+pub const DTYPE_FIFO: c_int = 4;
+pub const DTYPE_KQUEUE: c_int = 5;
+pub const DTYPE_CRYPTO: c_int = 6;
+pub const DTYPE_MQUEUE: c_int = 7;
+pub const DTYPE_SHM: c_int = 8;
+pub const DTYPE_SEM: c_int = 9;
+pub const DTYPE_PTS: c_int = 10;
+pub const DTYPE_DEV: c_int = 11;
+pub const DTYPE_PROCDESC: c_int = 12;
+pub const DTYPE_EVENTFD: c_int = 13;
+pub const DTYPE_TIMERFD: c_int = 14;
+pub const DTYPE_INOTIFY: c_int = 15;
+pub const DTYPE_JAILDESC: c_int = 16;
+
+s! {
+    #[cfg(not(any(freebsd10, freebsd11)))]
+    pub struct xfile {
+        pub xf_size: crate::ksize_t,
+        pub xf_pid: crate::pid_t,
+        pub xf_uid: crate::uid_t,
+        pub xf_fd: c_int,
+        _xf_int_pad1: Padding<c_int>,
+        pub xf_file: crate::kvaddr_t,
+        pub xf_type: c_short,
+        _xf_short_pad1: Padding<c_short>,
+        pub xf_count: c_int,
+        pub xf_msgcount: c_int,
+        _xf_int_pad2: Padding<c_int>,
+        pub xf_offset: crate::off_t,
+        pub xf_data: crate::kvaddr_t,
+        pub xf_vnode: crate::kvaddr_t,
+        pub xf_flag: c_uint,
+        _xf_int_pad3: Padding<c_int>,
+        _xf_int64_pad: Padding<[i64; 6]>,
+    }
+}
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2747,6 +2747,9 @@ fn test_freebsd(target: &str) {
             // Added in FreeBSD 15
             "AT_HWCAP3" | "AT_HWCAP4" if Some(15) > freebsd_ver => true,
 
+            // Added in FreeBSD 15
+            "DTYPE_INOTIFY" | "DTYPE_JAILDESC" if Some(15) > freebsd_ver => true,
+
             _ => false,
         }
     });
--- a/libc-test/semver/freebsd.txt
+++ b/libc-test/semver/freebsd.txt
@@ -2425,4 +2425,5 @@ vmtotal
 wait4
 waitid
 xallocx
+xfile
 xucred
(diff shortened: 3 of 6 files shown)


src/new/freebsd/sys/file.rs
```diff
@@ -0,0 +1,82 @@
...
+pub const DTYPE_TIMERFD: c_int = 14;
+#[cfg(any(freebsd11, freebsd12, freebsd13))]
+pub const DTYPE_LINUXTFD: c_int = 14;
+#[cfg(freebsd15)]
+pub const DTYPE_INOTIFY: c_int = 15;
+#[cfg(freebsd15)]
+pub const DTYPE_JAILDESC: c_int = 16;
+
+s! {
+    pub struct xfile {
```
tgross35, line 27: Could this whole struct just be `#[cfg(not(any(freebsd10, freebsd11)))]`? I don't think we really need the EOL support so that simplifies config.

src/new/freebsd/sys/file.rs
```diff
@@ -0,0 +1,82 @@
...
+#[cfg(any(freebsd10, freebsd11, freebsd12))]
+pub const DTYPE_LINUXEFD: c_int = 13;
+#[cfg(any(freebsd14, freebsd15))]
+pub const DTYPE_TIMERFD: c_int = 14;
+#[cfg(any(freebsd11, freebsd12, freebsd13))]
+pub const DTYPE_LINUXTFD: c_int = 14;
+#[cfg(freebsd15)]
+pub const DTYPE_INOTIFY: c_int = 15;
+#[cfg(freebsd15)]
+pub const DTYPE_JAILDESC: c_int = 16;
```
tgross35: Similarly, I'd prefer to just drop anything that only exists on EOL versions (<13).

src/new/freebsd/sys/file.rs
antoncxx: Absolutely. I will update the PR tomorrow.

src/new/freebsd/sys/file.rs
antoncxx: Done.

src/new/freebsd/sys/file.rs
antoncxx: Done.

src/new/freebsd/sys/file.rs
```diff
@@ -0,0 +1,51 @@
...
+pub const DTYPE_PTS: c_int = 10;
+pub const DTYPE_DEV: c_int = 11;
+pub const DTYPE_PROCDESC: c_int = 12;
+pub const DTYPE_EVENTFD: c_int = 13;
+#[cfg(freebsd13)]
+pub const DTYPE_LINUXTFD: c_int = 14;
+#[cfg(any(freebsd14, freebsd15))]
+pub const DTYPE_TIMERFD: c_int = 14;
+#[cfg(freebsd15)]
+pub const DTYPE_INOTIFY: c_int = 15;
```
asomers: Even though `DTYPE_INOTIFY` didn't exist on FreeBSD 13 and 14, libc doesn't need to worry about that.  It's safe to define it here for all versions.  It's preferable, even, because that way Rust programs can use it without any special compiler settings for libc.  Otherwise it would be off-limits, since libc always uses the FreeBSD 13 ABI.  The only symbols that you need to be careful with are symbols that _changed_ between FreeBSD versions, not symbols that got added, like `struct kevent`.
```suggestion
pub const DTYPE_INOTIFY: c_int = 15;
```

src/new/freebsd/sys/file.rs
```diff
@@ -0,0 +1,51 @@
...
+pub const DTYPE_SHM: c_int = 8;
+pub const DTYPE_SEM: c_int = 9;
+pub const DTYPE_PTS: c_int = 10;
+pub const DTYPE_DEV: c_int = 11;
+pub const DTYPE_PROCDESC: c_int = 12;
+pub const DTYPE_EVENTFD: c_int = 13;
+#[cfg(freebsd13)]
+pub const DTYPE_LINUXTFD: c_int = 14;
+#[cfg(any(freebsd14, freebsd15))]
+pub const DTYPE_TIMERFD: c_int = 14;
```
asomers: This symbol's name changed in FreeBSD 14, even though it already existed.  So you might consider adding the old name as an alias.  However, since there is literally zero Rust code already using the old name, it's probably better not to add a symbol that will always be deprecated.
```suggestion
pub const DTYPE_TIMERFD: c_int = 14;
#[deprecated(since = "1.0.0", note = "Use DTYPE_TIMERFD instead")]
pub const DTYPE_LINUXTFD: c_int = DTYPE_TIMERFD;
```

src/new/freebsd/sys/file.rs
```diff
@@ -0,0 +1,51 @@
...
+#[cfg(freebsd15)]
+pub const DTYPE_INOTIFY: c_int = 15;
+#[cfg(freebsd15)]
+pub const DTYPE_JAILDESC: c_int = 16;
```
asomers: Just like `DTYPE_INOTIFY`, you can define this for all FreeBSD versions.
```suggestion
pub const DTYPE_JAILDESC: c_int = 16;
```

sun793208-jpg (approved):

asomers (approved):

Issue #4998, opened 2026-03-02
How should we handle raw function pointers?


tgross35:
There are a handful of cases where we have something that is canonically a function pointer, but also accepts integer values (`sigaction`-related items are one example). Just using `usize`/`size_t` works but it loses that type information. We have `Option<extern "C" fn(...) -> ...>` in some cases, but that makes stumbling into UB too easy if the API expects integer sentinels.

What we really need is raw function pointers, but that would be a long way away. See [#t-lang > Function pointers that work like references](https://rust-lang.zulipchat.com/#narrow/channel/213817-t-lang/topic/Function.20pointers.20that.20work.20like.20references/with/531648743).

I'm thinking maybe we should add a wrapper type to help out here:

```rust
use core::mem::MaybeUninit;

/// Also acts as `Sealed`. Mostly serves as a lint within libc that we don't forget
/// the `unsafe extern "C"`.
/// 
/// # Safety
/// Only implement on `unsafe extern "C"` function pointers. `Option<Self>` must
/// be the same size as `Self`.
unsafe trait FfiSafeFn: Copy {}

unsafe impl<T1, U> FfiSafeFn for unsafe extern "C" fn(T1) -> U {}
unsafe impl<T1, T2, U> FfiSafeFn for unsafe extern "C" fn(T1, T2) -> U {}
unsafe impl<T1, T2, T3, U> FfiSafeFn for unsafe extern "C" fn(T1, T2, T3) -> U {}

/// Has the same repr as `F`
#[allow(private_bounds)]
#[repr(transparent)]
#[derive(Clone, Copy, Debug)]
pub struct RawFnPointer<F: FfiSafeFn>(MaybeUninit<Option<F>>);

#[allow(private_bounds)]
impl<F: FfiSafeFn> RawFnPointer<F> {
    // Internal helper.
    const fn from_int(val: usize) -> Self { /* ... */ }
}

/* Maybe? To make constructing easier.
 *
 * Free functions to avoid breakage if `RawFnPointer` can become an alias
 * in the future. */

/// # Safety
///
/// Must point to a valid function.
#[allow(private_bounds)]
pub unsafe fn raw_to_fn_pointer<F: FfiSafeFn>(f: RawFnPointer<F>) -> F {
    unsafe { f.0.assume_init().unwrap() }
}

#[allow(private_bounds)]
pub fn raw_from_fn_pointer<F: FfiSafeFn>(f: F) -> RawFnPointer<F> {
[...]

PR #5008, opened 2026-03-09
Fix: Replace sighandler_t with sig_t for Apple and BSDs

### LINKED ISSUE #1273: Incorrect type name sighandler_t should be sig_t on apple targets

chengr4:
# Description

This PR tries to fix rust-lang/libc#1273.

> I am not able to evaluate whether `#[deprecated]` should be added.

# Sources

> Apple I checked at my local (Aarch MacOS 26.3.1)

- (Freebsd) https://github.com/freebsd/freebsd-src/blob/016570c4463d5908953355ee1cf9a385ad9601b4/sys/sys/signal.h
- (Dragonfly) https://github.com/DragonFlyBSD/DragonFlyBSD/blob/51591aff67978c86de99839e31d8c8fcc54bc316/sys/sys/signal.h
- https://www.gnu.org/software/gnulib/manual/html_node/signal_002eh.html
- The type `sighandler_t` (a GNU extension) is not defined on most non-glibc platforms: macOS 11.1
, FreeBSD 14.0, NetBSD 10.0, OpenBSD 7.5, AIX 5.1, HP-UX 11, Solaris 11.4, Cygwin, mingw, MSVC 14.

# Checklist

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -327,7 +327,7 @@ s! {
 
     pub struct sigaction {
         // FIXME(union): this field is actually a union
-        pub sa_sigaction: crate::sighandler_t,
+        pub sa_sigaction: crate::sig_t,
         pub sa_mask: sigset_t,
         pub sa_flags: c_int,
     }
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -326,9 +326,11 @@ s! {
         _pad: Padding<[usize; 9]>,
     }
 
+    // FIXME(1.0): This should not implement `PartialEq`
+    #[allow(unpredictable_function_pointer_comparisons)]
     pub struct sigaction {
         // FIXME(union): this field is actually a union
-        pub sa_sigaction: crate::sighandler_t,
+        pub sa_sigaction: crate::sig_t,
         pub sa_mask: sigset_t,
         pub sa_flags: c_int,
     }
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -11,6 +11,7 @@ pub type nfds_t = c_uint;
 pub type regoff_t = c_int;
 #[cfg(not(target_os = "dragonfly"))]
 pub type regoff_t = off_t;
+pub type sig_t = Option<unsafe extern "C" fn(crate::c_int)>;
 
 s! {
     pub struct sockaddr {
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -18,6 +18,18 @@ pub type pid_t = i32;
 pub type in_addr_t = u32;
 pub type in_port_t = u16;
 pub type sighandler_t = size_t;
+#[cfg(any(
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "tvos",
+    target_os = "watchos",
+    target_os = "visionos",
+    target_os = "freebsd",
+    target_os = "dragonfly",
+    target_os = "openbsd",
+    target_os = "netbsd"
+))]
+pub type sig_t = sighandler_t;
 pub type cc_t = c_uchar;
 
 cfg_if! {
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -1336,7 +1336,23 @@ extern "C" {
     #[cfg_attr(gnu_file_offset_bits64, link_name = "ftruncate64")]
     pub fn ftruncate(fd: c_int, length: off_t) -> c_int;
 
+    #[cfg(not(any(
+        target_vendor = "apple",
+        target_os = "freebsd",
+        target_os = "dragonfly",
+        target_os = "netbsd",
+        target_os = "openbsd"
+    )))]
     pub fn signal(signum: c_int, handler: sighandler_t) -> sighandler_t;
+
+    #[cfg(any(
+        target_vendor = "apple",
+        target_os = "freebsd",
+        target_os = "dragonfly",
+        target_os = "netbsd",
+        target_os = "openbsd"
+    ))]
+    pub fn signal(signum: c_int, handler: sig_t) -> sig_t;
 
     #[cfg_attr(target_os = "netbsd", link_name = "__getrusage50")]
     #[cfg_attr(gnu_time_bits64, link_name = "__getrusage64")]
--- a/libc-test/semver/apple.txt
+++ b/libc-test/semver/apple.txt
@@ -2195,6 +2195,7 @@ shmctl
 shmdt
 shmget
 shmid_ds
+sig_t
 sigaltstack
 sigevent
 siginfo_t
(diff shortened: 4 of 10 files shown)


dybucc:
Not sure if the author intends on finishing this up. If not, I've solved the
merge conflicts and (possibly) tweaked some the patchset at [^1].

Note these changes should be enough to close issue #1273, but it doesn't solve
the underlying problem with `sigaction` (which is prevalent across other
targets.) There's some places where it's used inappropriately.

[^1]: <https://github.com/dybucc/libc/commit/9f9458a9d3446d88d2b5dcbb8a837dbcaf34681e>

chengr4:
This PR was waiting for someone to review it for me. I can take another look after 8/9. If you have already resolved the issue and it is ready to merge, feel free to go ahead!

dybucc:
Nay, just wanted to know if you were still on it. I did make some changes,
though, as I believe the actual definition will change once we settle on the
right bindings for `siginfo_t` across all targets (see rust-lang/libc#1359.)

But this PR can probably merged as the required changes once we get `siginfo_t`
right are minimal. Though I don't think we should deprecate it. `siginfo_t`
actually exists (with sort of mangled name) in FreeBSD [^1] [^2].

All other relevant targets (BSDs and Apple) can just rely on `sig_t`, as that's
their definition, while hiding `sighandler_t` in the docs for anything but
FreeBSD (and every target this PR doesn't change.)

It would be nice to have `siginfo_t` be an alias to the function and `sig_t` be
a pointer to this alias. But Rust doesn't allow function pointer decay through
function aliases when `typedef`ining a function without prepending `*`.

[^1]: <https://github.com/freebsd/freebsd-src/blob/a9305a6f50be89f22c7a469aeee4c3bea2086ace/sys/sys/signal.h#L160>
[^2]: <https://github.com/freebsd/freebsd-src/blob/a9305a6f50be89f22c7a469aeee4c3bea2086ace/sys/sys/signal.h#L408>

src/unix/mod.rs
```diff
@@ -18,6 +18,18 @@ pub type pid_t = i32;
...
+    target_os = "ios",
+    target_os = "tvos",
+    target_os = "watchos",
+    target_os = "visionos",
+    target_os = "freebsd",
+    target_os = "dragonfly",
+    target_os = "openbsd",
+    target_os = "netbsd"
+))]
+pub type sig_t = sighandler_t;
```
tgross35: This can go in `src/unix/bsd/mod.rs` so we don't need the cfg

tgross35 (changes requested):
Sorry I've missed this, one request then LGTM. Please squash as well.

I think we should deprecate `sighandler_t` on platforms where it doesn't exist, adding `__sighandler_t` if that's relevant. But not something to do here.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5008/84304260e1179f2d5323777265c11a288d6d962b..d9d5db81817ea8a2e1266f23c2bb156f97ca72ee)*

src/unix/bsd/mod.rs
```diff
@@ -11,6 +11,22 @@ pub type nfds_t = c_uint;
...
+// and <https://github.com/freebsd/freebsd-src/blob/77d6c45afdca8a524a88edfb3097d4d9dc90b583/sys/sys/signal.h#L408>
+// for FreeBSD details.
+//
+// Because `sighandler_t` is defined in FreeBSD as a function alias that
+// implicitly decays to the function pointer and not a "straight" alias to a
+// function pointer, we can probably keep the same definitions in other BSDs.
+//
+// They have no `sighandler_t` but do have the same `sig_t`. Of course, this is
+// going to need tweaking because raw pointers to functions aren't a thing just
+// yet in Rust.
```
madsmtm: Isn't `Option<unsafe extern "C" fn()>` basically a raw pointer to a function?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5008/df65648ed34ef002c16c19e11f8cc62dbab82324..5526ab89b9574687bf0cdd3332edd109e89de463)*

tgross35: Kind of but in a very footgunny way, IMO. https://github.com/rust-lang/libc/issues/4998 and the thread https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/type.20representing.20raw.20function.20pointers/with/606804857

src/unix/bsd/mod.rs
```diff
@@ -11,6 +11,22 @@ pub type nfds_t = c_uint;
...
+pub type sig_t = *const crate::sighandler_t;
```
tgross35: Actually, reading things closer since Mads pointed it out, I think this should just be a `type sig_t = Option<unsafe extern "C" fn(c_int)>` and then `sighandler_t` needs to be deleted. Opsem says that's sound even for non-dereferenceable values, which was one of the questions in the past. The other question is whether or not it's a footgun, which I still think it is, but that applies to all function pointers in this crate.

Sorry for adding another change, but I think that gets us closer to where we want to be. You can also drop this comment since it will only have the problem that all function pointers have, and I don't think there's anything really worth noting about the FreeBSD definition (but do feel free to put any of it into the commit message).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5008/df65648ed34ef002c16c19e11f8cc62dbab82324..930ef42e65ecc65484319545dde477be318336cd)*

tgross35: Also it looks like you need to recheck the semver files based on the CI failure.

tgross35 (changes requested):

src/unix/bsd/mod.rs
chengr4: Thanks for letting me be the puppet so that I can learn a bit from such a high density of information.   

> Also it looks like you need to recheck the semver files based on the CI failure.

My fault. just thought it was error on Github side.

src/unix/bsd/apple/mod.rs
```diff
@@ -326,9 +326,11 @@ s! {
         _pad: Padding<[usize; 9]>,
     }
 
+    // FIXME(1.0): This should not implement `PartialEq`
+    #[allow(unpredictable_function_pointer_comparisons)]
```
chengr4: Warning: This is entirely an AI-generated solution. I did review it, but it's honestly beyond my current capability.

tgross35 (approved):
Thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5008/764b10dae4f54ad6ea81c65b2bffc75b7ed4ba63..af5ebe9a87bb3c1b1b5739eab953c6d37bc7450d)*

tgross35:
stable-declined since this is breaking

tgross35:
Note that `sighandler_t` also needs to be removed at some point, https://github.com/rust-lang/libc/issues/1273#issuecomment-5406372892

PR #5010, opened 2026-03-11
feat: add support for 32 bit `time_t` in Windows

dybucc:
# Description

This PR adds support symmetric to that offered in some Unix platforms for using 32-bit `time_t` in Windows. This PR also removes some of the `skip()` functions used with `ctest` in Windows, as it renames some symbols to properly reflect whether they are 32-bit or 64-bit compliant, and adds deprecation notices to functions that are not supported by Microsoft anymore (as per their security guidelines.)

There's a few functions _with_ secure variants (suffixed with `_s`) that are altogether missing and yet do appear in the Windows CRT docs, but that's work for another PR.

In the CI scripts, there seemed to be a slight inconsistency between what the environment variable `ARCH`, and the environment variable `ARCH_BITS` were meant to be checked for, so that also got tweaked.

An issue for the deprecation notices hasn't yet been opened, but if the PR seems viable, it can be readily opened before merging.

# Sources

* Example of function documented in the Windows CRT with secure variants, and the corresponding functionality when `_USE_32BIT_TIME_T` is defined.
  https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/ctime-s-ctime32-s-ctime64-s-wctime-s-wctime32-s-wctime64-s?view=msvc-170
* Microsoft security guidelines, mentioning deprecation of functions for which there's secure, `_s`-suffixed variants.
  https://learn.microsoft.com/en-us/cpp/c-runtime-library/security-features-in-the-crt?view=msvc-170

# Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -128,6 +128,11 @@ jobs:
           #  env: { ARCH_BITS: 32, ARCH: i686 }
           - target: i686-pc-windows-msvc
             os: windows-2025
+            env: { ARCH_BITS: 32, ARCH: i686 }
+          - target: i686-pc-windows-msvc
+            os: windows-2025
+            env: { ARCH_BITS: 32, ARCH: i686, RUST_LIBC_UNSTABLE_WINDOWS_TIME32: 1 }
+            artifact-tag: windows_time32
           - target: i686-unknown-linux-gnu
           - target: x86_64-pc-windows-gnu
             os: windows-2025
--- a/build.rs
+++ b/build.rs
@@ -31,6 +31,8 @@ const ALLOWED_CFGS: &[&str] = &[
     // Corresponds to `_REDIR_TIME64` in musl
     "musl32_time64",
     "vxworks_lt_25_09",
+    // Corresponds to `_USE_32BIT_TIME_T` in Windows CRT
+    "windows_use_time32",
 ];
 
 // Extra values to allow for check-cfg.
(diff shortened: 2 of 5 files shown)


dybucc:
> What is the point of this? Per @ChrisDenton, `_USE_32BIT_TIME_T` has been deprecated a long time ago [#gsoc > Idea: modernize the libc crate @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/421156-gsoc/topic/Idea.3A.20modernize.20the.20libc.20crate/near/576842593).
> 
> > All other changes resulted from running `./ci/style.py`, so they're not really part of the PR.
> 
> The vast majority of this PR is reformatting, I suspect your local rustfmt isn't picking up .rustfmt.toml for some reason. If there really was that amount of reformatting to do, it should be a separate PR anyway rather than mixed in with other API changes.

Just ammended the commit without the formatting changes. Adding support for `_USE_32BIT_TIME_T` was to be a tad bit more compliant with the Microsoft CRT docs, and how they seem to have routines specific to both `time64_t`, `time32_t`, and `time_t`, the latter depending on whether the above feature test macro is defined.

The reason why the PR was opened was because some uses of the `ctest`'s `skip_*` methods in `test_windows` under `libc-test` got removed without affecting test success. This may or may not have been due to the changes in this PR.

Should the PR be altogether closed, or should the changes to add support for `_USE_32BIT_TIME_T` be the only ones removed?

tgross35:
Are you referring to this? https://github.com/rust-lang/libc/blob/82b748093aa9e9de75722ee68814d0c80c710650/libc-test/build.rs#L784-L785

If so, the fix here is to deprecate the alias on Windows so we can remove it in 1.0. We don't have any reason to try to support the legacy behavior.

PR #5015, opened 2026-03-14
Re-export `core::ffi` integer types rather than redefining

tgross35:
Since the MSRV was raised to 1.65, we can use these aliases directly rather than copying them. In a future version, we can consider deprecation.


ChrisDenton:
I think you can remove this test too, since I think it's just testing that libc types match std types: https://github.com/rust-lang/libc/blob/main/libc-test/tests/primitive_types.rs

PR #5021, opened 2026-03-16
Fix the soundness bug in the representation of extern types

tgross35:
Since the very first import dafaca90f095 ("Initial import of liblibc"), `libc` has used uninhabited enums to represent C's incomplete/opaque types. While this is, as far as I know, technically okay when working behind raw pointers, it means that using reference types like `&FILE` can lead to easy UB.

Resolve this by changing the representation to a `!Sync + !Send + !Unpin` ZST, as recommended by the nomicon [1]. The loss of auto traits technically makes this user-visible, but it is unlikely that anybody who is doing sound things was relying on these. 

I also used this as an opportunity to add a forward-compatibility note about intended use that should allow us to switch to real extern types once those are available.

[1]: https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs

Cc @RalfJung

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -246,17 +246,36 @@ macro_rules! s_no_extra_traits {
 macro_rules! extern_ty {
     ($(
         $(#[$attr:meta])*
-        pub enum $i:ident {}
+        pub type $i:ident;
     )*) => ($(
         $(#[$attr])*
-        // FIXME(1.0): the type is uninhabited so these traits are unreachable and could be
-        // removed.
+        /// This is an extern type ("opaque" or "incomplete" type in C).
+        ///
+        /// <div class="warning">
+        /// This type's current representation allows inspecting some properties, such as via
+        /// <code>size_of</code>, and it is technically possible to construct the type within
+        /// <code>MaybeUninit</code>, However, this <strong>MUST NOT</strong> be relied upon
+        /// because a future version of <code>libc</code> may switch to a proper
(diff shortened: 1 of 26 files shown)


src/macros.rs
```diff
@@ -246,17 +246,37 @@ macro_rules! s_no_extra_traits {
...
+        /// <a href="https://rust-lang.github.io/rfcs/1861-extern-types.html">extern type</a>
+        /// representation when available.
+        /// </div>
+        // ^ unfortunately warning blocks currently don't render markdown so we need to
+        // use raw HTML.
+        //
+        // Representation based on the Nomicon:
+        // <https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs>.
+        //
         // FIXME(1.0): the type is uninhabited so these traits are unreachable and could be
```
RalfJung: The FIXME seems outdated now?

tgross35: You're right; we should still remove the traits but the reason needs updating (will fix)

JohnTitor (approved):
I like the idea! LGTM as long as fixing the test and updating FIXME.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5021/bc30895d168990a8993a9bd03aa9b630fa5b3412..6c88eff1c60c1ee343a09ebb63d1635922cb7c8e)*

PR #5022, opened 2026-03-17
Update FreeBSD 14 CI image

asomers:
And update the definition of struct mcontext_t.  An additional field has
been added, and the spare field reduced accordingly.
    
No backwards-compatibility concerns, because we still build with a
FreeBSD 12 ABI by default.

Source:
https://github.com/freebsd/freebsd-src/commit/fa10e2fb88224b1b30abb6c94a9cf4c99ce5103f

### DIFF
--- a/src/unix/bsd/freebsdlike/freebsd/x86_64/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/x86_64/mod.rs
@@ -119,12 +119,12 @@ s! {
         pub mc_xfpustate: register_t,
         pub mc_xfpustate_len: register_t,
         // freebsd < 15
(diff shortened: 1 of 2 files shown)


src/unix/bsd/freebsdlike/freebsd/x86_64/mod.rs
```diff
@@ -119,12 +119,12 @@ s! {
         pub mc_xfpustate: register_t,
         pub mc_xfpustate_len: register_t,
         // freebsd < 15
```
chengr4, line 121: This is not a high priority, but I believe the comment was forgotten.

PR #5023, opened 2026-03-18
wasi: add all _SC_* sysconf constants from wasi-libc

surban:
# Description

Add the complete set of `_SC_*` sysconf constants for WASI targets from wasi-libc's musl-based `unistd.h`.

Applies to both `wasm32-wasip1` and `wasm32-wasip1-threads`.

# Sources

* wasi-libc musl `unistd.h`: https://github.com/WebAssembly/wasi-libc/blob/2fc32bc81b9f07f8d9525edea59bfbaf760c06d6/libc-top-half/musl/include/unistd.h#L396

# Checklist

- [x] Relevant tests in semver have been updated


surban:

Done.

PR #5032, opened 2026-03-26
Deprecate windows `time64_t`

dybucc:
# Description

Functionality related to the Windows `time64_t` has been deprecated in favor of
a single, 64-bit wide `time_t`. This has also required some work into getting
rid of the conditional compilation uses of `time_t` on GNU target environments,
and tweaking the `max_align_t` type. But that was separated onto a different PR
in rust-lang/libc#5050.

The `FIXME` comment on the tier 1 platform support for Windows with GNU has also
been removed as no segfaults were observed.

There's not yet an issue open specifically for deprecation of non-64-bit wide
time and offset-related values, so the linked issue in the deprecation notice
for `time64_t` is still pending.

# Sources

- Mingw source code showing how is the bit-width of `time_t` values is handled
  under Windows GNU.
  <https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/crt/time.h#L54-L57>
  <https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/crt/time.h#L235-L250>

# Checklist

- [x] No placeholder or unstable values like `*LAST` or `*MAX` are included (see
  [#3131](https://github.com/rust-lang/libc/issues/3131))

### DIFF
--- a/src/windows/mod.rs
+++ b/src/windows/mod.rs
@@ -18,13 +18,7 @@ pub type clock_t = i32;
 
 pub type errno_t = c_int;
 
-cfg_if! {
-    if #[cfg(all(target_arch = "x86", target_env = "gnu"))] {
-        pub type time_t = i32;
-    } else {
-        pub type time_t = i64;
-    }
-}
+pub type time_t = i64;
 
 pub type off_t = i32;
 pub type dev_t = u32;
@@ -34,6 +28,12 @@ extern_ty! {
     pub type timezone;
 }
 
+#[deprecated(
+    since = "1.0.0",
+    note = "This time-related value, among others, is part of the shift \
+            towards using a single, 64-bit-sized `time_t`. See #PENDING for \
+            discussion."
+)]
 pub type time64_t = i64;
 
 pub type SOCKET = crate::uintptr_t;
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -775,7 +775,6 @@ fn test_cygwin(target: &str) {
 fn test_windows(target: &str) {
     assert!(target.contains("windows"));
     let gnu = target.contains("gnu");
-    let i686 = target.contains("i686");
 
     let mut cfg = ctest_cfg();
 
@@ -840,8 +839,6 @@ fn test_windows(target: &str) {
     cfg.skip_alias(move |alias| match alias.ident() {
         "SSIZE_T" if !gnu => true,
         "ssize_t" if !gnu => true,
-        // FIXME(windows): The size and alignment of this type are incorrect
-        "time_t" if gnu && i686 => true,
         _ => false,
     });
 
--- a/libc-test/tests/windows_time.rs
+++ b/libc-test/tests/windows_time.rs
@@ -1,0 +1,22 @@
+//! Ensures Windows `time`-related routines align with `libc`'s interface. By
+//! default, both MSVC and GNU (under `mingw`) expose 64-bit symbols. `libc`
+//! also does that, but there's been slight inconsistencies in the past.
+
+#![cfg(windows)]
+
+/// Ensures a 64-bit write is performed on values that should always be 64 bits.
+/// This may fail if
+/// (1) the routine links with its 32-bit variant. This only happens if
+///     `_USE_32BIT_TIME_T` is defined. In theory, this should not be
+///     possible when working with Rust's `libc`.
+/// (2) Or `time_t` is 32-bits, and a 64-bit write overwrites both array items.
+///     This should neither be possible unless the above macro is defined.
+///     Support for non-64-bit values in `libc` should thus be non-existent.
+#[test]
+fn test_64_bit_store() {
+    let mut time_values: [libc::time_t; 2] = [123, 456];
+    let ptr = time_values.as_mut_ptr();
+    unsafe { libc::time(ptr) };
+    assert!(time_values[0] != 123);
+    assert_eq!(time_values[1], 456);
+}


.github/workflows/ci.yaml
```diff
@@ -123,15 +123,13 @@ jobs:
             os: windows-11-arm
           - target: aarch64-unknown-linux-gnu
             os: ubuntu-24.04-arm
-          # FIXME: It currently causes segfaults.
-          #- target: i686-pc-windows-gnu
-          #  env: { ARCH_BITS: 32, ARCH: i686 }
+          - target: i686-pc-windows-gnu
+            os: windows-2025
```
tgross35: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5032/34527ca4c0b43f10062a810789953a538a76ed7d..84d1dfb7f9a922353acaedd1730df8c2fe0fd168)*

Would you mind splitting whatever is needed to fix this to a separate PR? That would be great to have independent of deprecation.

Think that includes this portion of the file, the shell script, and `src/windows/gnu/mod.rs`.

src/windows/mod.rs
```diff
@@ -18,13 +18,7 @@ pub type clock_t = i32;
...
 pub type errno_t = c_int;
 
-cfg_if! {
-    if #[cfg(all(target_arch = "x86", target_env = "gnu"))] {
-        pub type time_t = i32;
-    } else {
-        pub type time_t = i64;
-    }
-}
+pub type time_t = i64;
```
tgross35, line 21: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5032/34527ca4c0b43f10062a810789953a538a76ed7d..84d1dfb7f9a922353acaedd1730df8c2fe0fd168)*

Have you run any tests before and after this change for whether things work? I'm wondering about this group of functions where it seems like we're already linking the time64 version in some cases https://github.com/rust-lang/libc/blob/ff0e6164f209073c3856e3709c4dcae1e59c6a11/src/windows/mod.rs#L393-L413, and am thinking that we might have a bug with the current (before this PR) impl.

To test this you could use the `libc` from `main` and pass a pointer to a `[time_t; 2]` to the `time` function, both initialized to 0x1234abcd or whatever. If only one of them gets overwritten with the current time, there isn't a bug and things are working despite the link name. If one gets the current time and the other gets zeroed, then we have a bug.

This PR would fix it but we can't backport because of the breakage, so we'll need configure the `link_name` attributes behind `not(x86+gnu)`.

(I realize that's rather confusing, let me know if anything needs clarification)

dybucc: Regarding the first question on tests, I have run the existing `libc` test suite
both before and after the changes, though admittedly I did not run it prior to
making the changes to `src/windows/mod.rs`, _but_ without skipping `time_t` in
`build.rs`.

I'm not sure what do you mean when you say those routines may be faulty in the
current `main` worktree. From what I could gather of the Windows CRT docs (e.g.
[[1]], as all routines explain things similarly in this respect,) there
shouldn't be any issues, considering the non-suffixed variants are just wrappers
for the suffixed, 64-bit variants. So linking the 64-bit symbol with a type that
is also 64-bits wide now should be fine. Still, I understand there may be some
concerns with GNU on Windows, as in those cases, the expected `time_t` (prior to
this PR) was 32-bits wide, so that could be an issue.

Either way, I'll get back in a few hours once I actually run your test.

[1]: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/time-time32-time64?view=msvc-170

.github/workflows/ci.yaml
dybucc: Done. Those changes are now in #5050.

src/windows/mod.rs
dybucc: There may be a bug in the current `main` worktree. Running on Windows x86 GNU,
with a two-element array of 32-bit `time_t`, a 64-bit store is being performed
on the pointer's pointee that gets passed to one of these 64-bit-linked routines
(tested it out with `time()` in `libc`, which links by default on all
targets/environments to `_time64()`) as one of the elements is zeroed after the
call.

I would personally say the bug is not in linking these routines by default to
their 64-bit variants, but rather in the fact that we assume a 32-bit `time_t`
on Windows GNU x86. By default, Rust's `libc` behaves correctly under MSVC, as
`time_t` is always 64-bit wide unless `USE_32BIT_TIME_T` is defined, no matter
the target triple, and in `libc` we opted out of adding support for that feature
test macro. On the other hand, when running under `mingw` with GNU in x86, we
always define `time_t` to be 32-bit wide. This, according to [[1]] and [[2]], is
incorrect, as behavior in `mingw` is exactly like that in MSVC when it comes to
bit-width; Namely, define `USE_32BIT_TIME_T` if you want 32-bit `time_t` and
non-suffixed variants to default to their 32-bit variants, otherwise, it's
64-bit `time_t` all the way through.

[1]: https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/crt/time.h#L54-L57
[2]: https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/crt/time.h#L235-L250

Running the test with the changes in this PR solves the issues, simply because
there's no 32-bit `time_t` exposed on Rust's `libc` under Windows anymore.

dybucc:
Just updated the PR description and removed the stable nomination, as there's
clearly no backwards-compatible changes to removing a type of a certain
bit-width, which is what's happening in Windows GNU x86.

@rustbot label -stable-nominated

tgross35:
> Just updated the PR description and removed the stable nomination, as there's clearly no backwards-compatible changes to removing a type of a certain bit-width, which is what's happening in Windows GNU x86.

The current behavior is broken so we do need a fix on stable. Could you put up a PR gating the `link_name` attributes?

People also need a way to experiment on stable, so I think we need to make this backportable by putting the changes behind a cfg. We're moving away from environment variables, the new way of setting cfg via `RUSTFLAGS` should look something like this https://github.com/rust-lang/libc/pull/4977 (you can ignore the verify-build updates).

src/windows/mod.rs
tgross35: Would you be able to add that test to `libc-test/tests`?

dybucc: Done.

dybucc:
According to the default behavior in both MSVC and GNU on Windows, whether we
even have the `link_name` attribute linking to the 64-bit routines does not seem
relevant. I just tested again without changing the link symbol in `libc` with
the same test I submitted to the PR, and there were no changes. This is
expected, considering the default is to define the non-suffixed variants in
terms of their 64-bit variants.

If anything, I'd say the fix should be to have the routine be linked to the
32-bit variant. But I don't think this is a good way of going about stuff,
considering that is not meant to be something Rust's `libc` does, but rather
something that is defined through the `_USE_32BIT_TIME_T` feature test macro,
and then conditionally linked to the 32-bit variant if such macro is defined.

The fix on stable may be to just expose a 32-bit `time_t` under Windows GNU (so
no changes from `libc`'s current `main` in that respect,) and to have the
`link_name` instead redirect to the 32-bit variant. This may not be a great idea
because defining `_USE_32BIT_TIME_T` makes a few more things be conditionally
compiled. But it can be done if all the places in both MSVC and `mingw` that are
affected (and exposed through Rust's `libc`) get refactored accordingly. This
shouldn't require exposing a new `cfg`.

But I could be misunderstanding what you mean by "putting the changes behind a
`cfg`" and "gating the `link_name`."

dybucc:
Alternatively, and judging by the documentation on _some_ of the MSVC CRT docs
[[1]], it seems as if calling directly onto the suffixed variants is possible
without breaking something. This could make for a quick fix on stable as linking
to the 32-bit variants would be feasible only for 32-bit `time_t`, which itself
only happens under Windows when running in x86 and GNU. This would make it
easier as no other refactors for other non-32-bit types would be required (as
would otherwise be if we implemented all of the conditionally-compiled changes
that defining `_USE_32BIT_TIME_T` forces.)

Just opened #5059 with this fix.

[1]: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/difftime-difftime32-difftime64

dybucc:
Gating the changes behind a `cfg` is only feasible if the tests are still made to skip `time_t` and
records relying on it for some of its fields. This actually defeats the purpose of this PR, which
targets 1.0, so I've opened up #5062 with the same changes but with an additional `cfg` that allows
switching to a 64-bit `time_t` when running under Windows GNU x86.

dybucc:

tgross35:
Similar to the other deprecations,

@rustbot blocked

SnoozeThis:
(https://snoozeth.is/u5ssmUMA3fM) I will wait until Fri, 04 Sep 2026 04:30:56 UTC and then add label S-waiting-on-review and remove label S-blocked.

@rustbot claim.

SnoozeThis:
Resolved.

@rustbot label +S-waiting-on-review -S-blocked. @rustbot release-assignment.

PR #5041, opened 2026-04-02
linux: Qdd sockaddr_iucv

clayton615:
This adds the sockaddr_iucv define in netiucv/iucv. It is primarily used on the s390x platform when linux is running under z/VM, but is present on all the linux systems I've been able to test on.

# Description

Adds support for sockaddr_iucv

# Sources

https://github.com/torvalds/linux/blob/5619b098e2fbf3a23bf13d91897056a1fe238c6d/include/net/iucv/iucv.h

# Checklist

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -3781,6 +3781,7 @@ fn test_linux(target: &str) {
         "netinet/ip.h",
         "netinet/tcp.h",
         "netinet/udp.h",
+        (gnu, "netiucv/iucv.h"),
         (l4re, "netpacket/packet.h"),
         "poll.h",
         "pthread.h",
--- a/src/unix/linux_like/linux/gnu/b64/mod.rs
+++ b/src/unix/linux_like/linux/gnu/b64/mod.rs
@@ -98,6 +98,17 @@ s! {
         __glibc_reserved4: Padding<crate::__syscall_ulong_t>,
     }
 
+    // net/iucv/iucv.h
+
+    pub struct sockaddr_iucv {
+        pub siucv_family: crate::sa_family_t,
+        __siucv_port: Padding<u16>, // unused crate::in_port_t
+        __siucv_addr: Padding<u32>, // unused crate::in_addr_t
+        pub siucv_nodeid: [c_char; 8],
+        pub siucv_user_id: [c_char; 8],
+        pub siucv_name: [c_char; 8],
+    }
+
     pub struct timex {
         pub modes: c_uint,
         #[cfg(all(target_arch = "x86_64", target_pointer_width = "32"))]


clayton615:
Nevermind, I'm realizing I have no clue how to tell what archs have the header vs which don't. I know I want it on s390x to work, and at least x86 so that rust analyzer won't scream while I'm writing code around it.

clayton615:
Huzzah, all tests pass and I think I've done this all correctly. 
Apologies for the churn, if you're thinking "This guy has no clue what he's doing" you're right, this is my first contribution to any project. 
Let me know if I've done anything wrong or need to change something.

src/unix/linux_like/mod.rs
```diff
@@ -235,6 +235,24 @@ s! {
     }
 }
 
+cfg_if! {
+    if #[cfg(all(
+        any(target_arch = "x86_64", target_arch = "s390x"),
+        not(any(target_env = "musl", target_os = "android"))
+    ))] {
+        s! {
+            pub struct sockaddr_iucv {
```
tgross35: Put this in `linux_like/linux/gnu`; currently this is available on theoretical emscripten and l4re targets which I think probably doesn't make sense. I think you may also be able to drop the `target_arch` list too if you do that, and just leave it enabled everywhere.

(test config will need to be updated to match)

src/unix/linux_like/mod.rs
```diff
@@ -235,6 +235,24 @@ s! {
...
+                pub siucv_family: crate::sa_family_t,
+                pub siucv_port: crate::in_port_t,
+                pub siucv_addr: crate::in_addr_t,
+                pub siucv_nodeid: [c_char; 8],
+                pub siucv_user_id: [c_char; 8],
+                pub siucv_name: [c_char; 8],
```
tgross35: Per https://github.com/torvalds/linux/blob/7ca6d1cfec80ebe46cc063f3284c5896c344d9a1/include/net/iucv/af_iucv.h#L46-L50 all except for the first field are reserved. Does common access require using them or could we make them private in case things change?

clayton615: Port and Addr don't do anything as far as I know. Making them private should be fine. NodeID and UserID are used to specify your target z/VM ID to connect to, and Name is the application. However, my basic understanding of the C code is that NodeID is actually unused in the linux implementation, so I don't know if the right thing to do is make it private or leave it public in the hopes that some day it's fixed. 

In other words, at least UserID and Name should be public. I'll defer to experience on NodeID.

src/unix/linux_like/mod.rs
clayton615: Okay, I'll try it under linux/gnu and see if the tests pass. Thanks.

clayton615:
Can I just do `#[cfg(target_pointer_width = "64)]` under here? Or should I add it under ./b64/mod.rs? I'm trying to be tidy and follow conventions, but I'm not sure how best proceed.

src/unix/linux_like/mod.rs
tgross35: If you don't know of port and addr to do anything and they have a "reserved" comment, would you mind making them private? It's nice to be on the conservative side here because a private field also makes the struct effectively non-exhaustive so we can add fields in the future (just using `#[non_exhaustive]` would be great but it has a bug currently). And it's easy enough to undo that if we do need them on the future.

tgross35:
Also the CI failure should be fixed if you can rebase.

clayton615:
Whoops, was just trying to rebase, didn't mean to push changes yet. I need more practice with git. Since I'm hear, if I make the fields siucv_port and siucv_addr private, it causes the tests from cargo test to fail. Should I be trying to edit those as well to not do whatever it is that it's doing on them?

tgross35:
Mind posting the failure? I think things should work. But if making them private turns out to be more effort than it's worth, feel free to just leave them as-is 🙂

clayton615:
There's about 10 or so errors, all like the following:

```
   Compiling libc-test v0.1.0 (/home/clayton/Documents/Source/libc/libc-test)
error[E0616]: field `siucv_port` of struct `libc::sockaddr_iucv` is private
      --> /home/clayton/Documents/Source/libc/target/debug/build/libc-test-85dc5bd74a0d5163/out/ctest_output.rs:142987:55
       |
142987 |         let ty_ptr = unsafe { &raw const (*uninit_ty).siucv_port   };
       |                                                       ^^^^^^^^^^ private field

error[E0616]: field `siucv_port` of struct `libc::sockaddr_iucv` is private
      --> /home/clayton/Documents/Source/libc/target/debug/build/libc-test-85dc5bd74a0d5163/out/ctest_output.rs:142993:46
       |
142993 |         check_same(offset_of!(sockaddr_iucv, siucv_port) as u64, ctest_field_offset,
       |                                              ^^^^^^^^^^ private field

```
I'm going to be honest, I'm not really looking through the test code. Since it picked up my changes and started testing them, I was hoping that was enough. If I should be digging through them and making changes there too, let me know, but honestly since it's clearly running some tests against my new struct I was hoping that was "good enough".

clayton615:
Looks like all the tests are passing and I think I've rebased correctly. Is there any additional action necessary on my end to move this forward?

clayton615:
@tgross35 sorry to pester, but I need to know if any additional action is required from me. 
I realized I let it sit for a while but I'm still trying to drive this forward as I can.

JohnTitor:
@clayton615 You need to use `Padding` to make the fields private, see other structs for examples.
Also please actually rebase and squash commits into one instead of merging.

@rustbot author

clayton615:
I'm still pretty new at all of this, but I think I've done what you asked. It looks like it's squashed into one commit, and I rebased on Main. I've also used padding on the two unused fields. Can someone confirm this is correct? Thanks for all of your patience while I'm learning.

clayton615:
Why did the android test just cancel itself? I don't see any obvious error in the log... Is there a way for me to restart it without having to push another commit?

tgross35 (approved):
Android CI is unfortunately flaky, see https://github.com/rust-lang/libc/issues/4297. But this should be good, thank you for the updates!

I pushed your branch to clean up the commit message, the fixup messages that aren't relevant to the final patch don't need to be kept around. (If using interactive rebase, `f`/`fixup` rather than `s`/`squash` is probably what you actually want).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5041/064dc78220fc185dc1e679583dc28d3efa48ccd5..e5c608e5ae5e71669cc0a0151ab9bf48d13a68fe)*

clayton615:
Thanks everyone. Will this also trickle into 0.2? Most packages rely on that branch, so I'm hoping it can be added there. I did mark it for stable nominated, so just checking if it will happen or not.

tgross35:
It will, the `stable-nominated` label means it will get picked up in the next backport round

PR #5045, opened 2026-04-07
horizon: Change POLL constants from c_short to c_int

hyperc4m:
# Description

This change modifies constants to match the definition of the types in pollfd
https://github.com/rust-lang/libc/blob/b6ab1fbc3946975593d3b3ef3b733ed97ae64146/src/unix/newlib/horizon/mod.rs#L31

# Sources

This struct is meant to match the libctru pollfd which also uses ints for all the struct members.
https://github.com/devkitPro/libctru/blob/36fe1ada5b7ebe53ba4decda36d764a55f8fefb6/libctru/include/poll.h#L12

# Checklist

### DIFF
--- a/src/unix/newlib/horizon/mod.rs
+++ b/src/unix/newlib/horizon/mod.rs
@@ -150,16 +150,16 @@ pub const SOL_CONFIG: c_uint = 65534;
 pub const PTHREAD_STACK_MIN: size_t = 4096;
 pub const WNOHANG: c_int = 1;
 
-pub const POLLIN: c_short = 0x0001;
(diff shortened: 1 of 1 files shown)


tgross35 (approved):
In that case LGTM, thanks! Pushed a `horizon:` tag.

The CI issues were from a nightly break, should be a workaround merged now.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5045/e5880b8471f55c02a1cc047277c1bfd8801f44b3..5c29cebe6b648b68ed59596c65b093b692cf1067)*

PR #5046, opened 2026-04-07
uclibc: fix build + CI, add unstable time64 support for newer uclibc toolchains 

skrap:
# Description

Goals of this change:

- Fix uclibc build
- Fix uclibc CI
- support uclibc configurations which use a 64-bit `time_t` configuration, which is enabled on the latest bootlin toolchains.

## Testing instructions

For testing, I have done the following:

 - build a custom rust toolchain against these changes (cherry-picked onto the libc-0.2 branch)
 - verified `libc-test` passes against the last bootlin toolchain to use 32-bit `time_t` (bleeding-edge-2024.02-1)
 - verified `libc-test` passes against the most recent bootlin toolchain (bleeding-edge-2025.08-1)
   - This required setting `RUSTFLAGS="--cfg=libc_unstable_uclibc_time64"` as expected

# Sources

I am not changing APIs so I don't think this applies.

# Checklist

  - NOTE: I only included such values where semver required them.


ci/install-uclibc.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/bash
+#
+# Installs the appropriate uclibc toolchain into /toolchain
+
+set -eux
+
+time64="$1"
+
+if [ "${time64:-0}" != "0" ]; then
```
skrap, line 9: I added this script to use a toolchain with or without 64-bit `time_t` depending on `$1`.

libc-test/build.rs
```diff
@@ -3934,8 +3940,7 @@ fn test_linux(target: &str) {
             "linux/wait.h",
             "linux/wireless.h",
             "sys/fanotify.h",
-            // <sys/auxv.h> is not present on uclibc
-            (!uclibc, "sys/auxv.h"),
+            "sys/auxv.h",
```
skrap, line 3965: uclibc has this header, at least in the toolchain versions in the CI.

libc-test/build.rs
```diff
@@ -4627,8 +4789,9 @@ fn test_linux(target: &str) {
             // Needs glibc 2.33 or later.
             "mallinfo2" => true,
 
-            // Not defined in uclibc as of 1.0.34
+            // Not defined in uclibc as of 1.0.45
             "gettid" if uclibc => true,
+            "getauxval" if uclibc => true,
```
skrap, line 4852: libc doesn't define `getauxval`.

src/unix/linux_like/linux/uclibc/mod.rs
```diff
@@ -87,37 +87,6 @@ s! {
         pub nr: crate::__s32,
     }
 
-    #[cfg_attr(
```
skrap, line 90: Several definitions here were moved around in the linux/l4re shared restructuring.  These are now defined elsewhere; remove them here.

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -805,11 +805,9 @@ pub const EM_M32R: u16 = 88;
 pub const EM_MN10300: u16 = 89;
 pub const EM_MN10200: u16 = 90;
 pub const EM_PJ: u16 = 91;
-#[cfg(not(target_env = "uclibc"))]
```
skrap, line 801: Many of these guarded constants are actually defined in uclibc.  I removed guards where appropriate.

src/unix/linux_like/mod.rs
```diff
@@ -805,11 +805,8 @@ pub const IP_TRANSPARENT: c_int = 19;
 pub const IP_ORIGDSTADDR: c_int = 20;
 pub const IP_RECVORIGDSTADDR: c_int = IP_ORIGDSTADDR;
 pub const IP_MINTTL: c_int = 21;
-#[cfg(not(target_env = "uclibc"))]
```
skrap: Similar removal of unnecessary guards.

skrap:
CI seems to error on [a rustc issue](https://github.com/rust-lang/rust/issues/154878) when building; I think this is unrelated.  Awaiting a fix there, but feel free to review in the meantime!

.github/workflows/ci.yaml
```diff
@@ -225,6 +225,9 @@ jobs:
           - target: x86_64-unknown-linux-musl
             env: { RUST_LIBC_UNSTABLE_MUSL_V1_2_3: 1 }
             artifact-tag: new-musl
+          - target: armv7-unknown-linux-uclibceabihf
+          - target: armv7-unknown-linux-uclibceabihf
+            env: { TEST_UCLIBC_TIIME64: 1 }
```
tgross35: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/b6ab1fbc3946975593d3b3ef3b733ed97ae64146..d234f7de419c7289637350af539b0a75ebf1de81)*

This is fine for the purposes of this PR, but will need to get removed before merge; tier 2 vs. Tier 3 serves as the cutoff between what we do and don't test.

If you are interested in getting this target better tested and putting support behind that, please consider [getting it upgraded to tier 2](https://doc.rust-lang.org/rustc/target-tier-policy.html#tier-2-target-policy) :)

ci/docker/armv7-unknown-linux-uclibceabihf/Dockerfile
tgross35, line 1: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/b6ab1fbc3946975593d3b3ef3b733ed97ae64146..d234f7de419c7289637350af539b0a75ebf1de81)*

Note that despite that, having the script updates is fine.

src/unix/linux_like/linux_l4re_shared.rs
```diff
@@ -1196,26 +1188,22 @@ pub const CMSPAR: crate::tcflag_t = 0o10000000000;
...
+pub const MFD_HUGE_8MB: c_uint = 0x5c000000;
+pub const MFD_HUGE_16MB: c_uint = 0x60000000;
+pub const MFD_HUGE_32MB: c_uint = 0x64000000;
+pub const MFD_HUGE_256MB: c_uint = 0x70000000;
+pub const MFD_HUGE_512MB: c_uint = 0x74000000;
+pub const MFD_HUGE_1GB: c_uint = 0x78000000;
+pub const MFD_HUGE_2GB: c_uint = 0x7c000000;
+pub const MFD_HUGE_16GB: c_uint = 0x88000000;
+pub const MFD_HUGE_MASK: c_uint = 63;
+pub const MFD_HUGE_SHIFT: c_uint = 26;
```
tgross35: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/b6ab1fbc3946975593d3b3ef3b733ed97ae64146..d234f7de419c7289637350af539b0a75ebf1de81)*

@farao do the constants in this file that are changed here not exist on l4re? If so, they could just move to `linux` rather than `linux_l4re_shared`.

ci/run-docker.sh
```diff
@@ -44,6 +44,13 @@ run() {
         fi
     fi
 
+    if [[ "$run_target" = *"uclibc"* ]]; then
+        if [ -n "${TEST_UCLIBC_TIIME64:-}" ]; then 
+            build_args+=("--build-arg=TEST_UCLIBC_TIIME64=1")
+            export RUSTFLAGS="$RUSTFLAGS --cfg=libc_unstable_uclibc_time64"
+        fi
+    fi
+
```
tgross35: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/b6ab1fbc3946975593d3b3ef3b733ed97ae64146..d234f7de419c7289637350af539b0a75ebf1de81)*

Nit: can probably remove the `run_target = uclibc` guard, that env likely shouldn't be set by anything else and the cfg wouldn't have an effect anyway.

libc-test/build.rs
```diff
@@ -4359,18 +4376,159 @@ fn test_linux(target: &str) {
             // Skip as this signal codes and trap reasons need newer headers
             "TRAP_PERF" => true,
 
-            // kernel constants not available in uclibc 1.0.34
-            "EXTPROC"
+            // constants not available in uclibc 1.0.45
+            // but defined outside the uclibc library,
+            // e.g. file format constants or kernel-defined
+            // constants.
```
tgross35: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/b6ab1fbc3946975593d3b3ef3b733ed97ae64146..d234f7de419c7289637350af539b0a75ebf1de81)*

Nit: can be rewrapped to 88 or 100 chars to match the rest (kind of wish rustfmt just did this)

libc-test/build.rs
```diff
@@ -4359,18 +4376,159 @@ fn test_linux(target: &str) {
...
+            "ALG_SET_KEY_BY_KEY_SERIAL"
+            | "AT_MINSIGSTKSZ"
+            | "BUS_MCEERR_AO"
+            | "BUS_MCEERR_AR"
+            | "CAN_BUS_OFF_THRESHOLD"
+            | "CAN_CTRLMODE_TDC_AUTO"
+            | "CAN_CTRLMODE_TDC_MANUAL"
+            | "CAN_ERR_CNT"
```
tgross35: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/b6ab1fbc3946975593d3b3ef3b733ed97ae64146..d234f7de419c7289637350af539b0a75ebf1de81)*

It seems like many of these should be fine if included via `linux/... .h` paths; is there a reason that isn't recommended or doesn't work on uclibc?

If so, it would be good to mark them deprecated on the target and remove in a release or two.

.github/workflows/ci.yaml
skrap: Gotcha, will remove.  I don't know of anyone else in the world who uses this platform, so I don't think it's a good candidate for tier 2.  I do have organizational support for maintaining this port but I worry that if this platform were promoted to tier 2 that any breakage would become a blocker to rust progress, and I would be a single point of failure.
(There are certainly other uclibc users, but none are tier 2.  armv7 is obviously well-supported.)

libc-test/build.rs
skrap: I can do that.

I was unsure of the philosophy here with respect to kernel-defined constants.  They would either be present in `linux/*.h` headers or not depending on the kernel version, regardless of whatever libc library/version was being used on the system.  So what triggers a particular constant being promoted into the `libc` crate?

skrap: I included the constants via linux UAPI headers as much as possible, but in some cases doing so conflicted with the definitions from uclibc.  In this case, I left a comment.

In order to increase readability, I moved the uclibc-specific handling of exceptions to its own code branch and out of the catch-all branch, mirroring how musl does things.

ci/run-docker.sh
skrap: Done.

skrap:
I addressed the review feedback.  Thanks for having a look!

I had to fence out `POSIX_SPAWN_SETSID` under `target_env = uclibc`.  This is a GNU-only constant and is not implemented in uclibc.  It was mistakenly included.

src/unix/linux_like/linux_l4re_shared.rs
farao: yes that's true they do not exist on l4re so they should rather be moved to linux. I think I only put them in here to have all the MFD_* consts in one file.

skrap:
@tgross35 I am unsure of the desired workflow here - I see a "request re-review" button in the github UI.  Is this on your plate already, or do I push that button now that I have addressed your feedback?  (Not trying to rush, just trying to learn the intended workflow.)

ci/install-uclibc.sh
```diff
@@ -0,0 +1,20 @@
...
+    version='bleeding-edge-2025.08-1'
+else
+    version='bleeding-edge-2024.05-1'  # last version with 32-bit time_t
+fi
+
+mkdir /toolchain
+
+curl --retry 5 -L "https://toolchains.bootlin.com/downloads/releases/toolchains/armv7-eabihf/tarballs/armv7-eabihf--uclibc--${version}.tar.xz" | \
+tar xjf - -C /toolchain --strip-components=1
```
tgross35, line 18: No need to change this of course but fyi, `|` at the end of the line allows wrapping. So the `\` isn't needed

src/unix/linux_like/linux_l4re_shared.rs
tgross35: What about the constants like `EM_OPENRISC` or `PTHREAD_MUTEX_STALLED` above?

@skrap for the constants here, could you update them to `not(target_os = "l4re")` rather than removing the gate? I think that's easier than moving them all.

tgross35: Maybe also with a comment that this is done to keep similar constants in the same file

libc-test/build.rs
```diff
@@ -4302,6 +4325,170 @@ fn test_linux(target: &str) {
...
+                | "PR_MDWE_REFUSE_EXEC_GAIN" => return true,
+
+                // defined as a synonym for EM_ARC_COMPACT in gnu but not uclibc
+                "EM_ARC_A5" => return true,
+
+                /*
+                Here are a list of kernel UAPI constants which appear in linux/ headers,
+                but cannot be imported due to conflicts with the uclibc headers.
+                The conflicting linux/ header is noted.
+                 */
```
tgross35, line 4452: What is the expected flow for these when you're writing C if there are conflicts?

ci/install-uclibc.sh
```diff
@@ -0,0 +1,20 @@
...
+
+set -eux
+
+time64="$1"
+
+if [ "${time64:-0}" != "0" ]; then
+    version='bleeding-edge-2025.08-1'
+else
+    version='bleeding-edge-2024.05-1'  # last version with 32-bit time_t
+fi
```
tgross35, line 13: Does this more or less mean that you're stuck on an old uclibc version if you want to use time APIs? There aren't the duplicate symbols and header config like glibc has?

If so, I think we could flip the default at some point in the near future. Technically breaking but we can get away with it on T3 targets, and musl and glibc aren't quite as stuck.

tgross35 (changes requested):
Think the todo list here should be about the last one, also some questions.

> @tgross35 I am unsure of the desired workflow here - I see a "request re-review" button in the github UI. Is this on your plate already, or do I push that button now that I have addressed your feedback? (Not trying to rush, just trying to learn the intended workflow.)

You can request a review. Usually I just go based on labels but requesting a review should also flip to `S-waiting-on-review` IIRC (as does `@rustbot review`).

The cycle is a bit slow for libc for nontrivial PRs, typically I go through everything every ~2 weeks.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/ff0e6164f209073c3856e3709c4dcae1e59c6a11..8f8451e945bc8a2ecacd49c7905ef4272cd2d3c6)*

src/unix/linux_like/linux/uclibc/arm/mod.rs
```diff
@@ -2,15 +2,24 @@ use crate::off64_t;
...
+cfg_if! {
+    // Set cfg(libc_unstable_uclibc_time64) in rustflags if your uclibc has 64-bit time
+    if #[cfg(linux_time_bits64)] {
+        pub type time_t = c_longlong;
+        pub type suseconds_t = c_longlong;
+    } else {
+        pub type time_t = c_long;
+        pub type suseconds_t = c_long;
+    }
+}
```
tgross35, line 15: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/ff0e6164f209073c3856e3709c4dcae1e59c6a11..8f8451e945bc8a2ecacd49c7905ef4272cd2d3c6)*

Do any structs need to be updated here? I know on other targets the padding in the timespec struct sometimes needs to be adjusted. I assume there are no function link names to update if uclibc had a pretty clean cut 32->64 transition.

ci/install-uclibc.sh
skrap: Ah, TIL!

src/unix/linux_like/linux_l4re_shared.rs
skrap: Re `EM_OPENRISC`:

Some of these are a bit unexpected to have live in libc, to be honest.  The EM_* constants are defined not by libc or by the kernel, but by the [ELF format spec](https://gabi.xinuos.com/elf/a-emachine.html).  They're fixed no matter what OS or std library we're running.  So for this constant, I would prefer that `libc` present a unified interface to higher-level crates, rather than adopt the uclibc quirk of having `EM_OR1K` be a synonym for `EM_OPENRISC`.

Re `PTHREAD_MUTEX_STALLED`:
I'm not sure I understand the question here.  This constant isn't conditional based on the `target_env`.

libc-test/build.rs
skrap: My view on it is that these constants are passed through the underlying target_env and to the kernel, rather than invoking specific behavior within the target_env c std lib.  For constants where this is true, it means that they're safe and appropriate to use regardless of whether the c std lib re-exports them or not.  The case where they could lead to unintentional bad behavior is when the c std lib needed to change behavior based on the constant.  I am not individually researching and testing each of these, so it's possible that there's a bug here, but I would like to consider those issues as outside the scope of this PR.  (This is just moving around these constants and fixing the build.)

To answer your question, if I were writing a piece of C code and needed one of these constants, I would probably find a way to untangle the header mess or just copy the `#define` into my implementation.  Ugly, yes, but this is the state of system headers in C in my experience.  

There are also of other examples in `libc-test/build.rs` of this type of issue - e.g. see `test_linux_like_apis()` which appears to try to work around this issue for combinations of specific constants and headers which don't work.

ci/install-uclibc.sh
skrap: Ah, no - let me explain.  uClibc is intended for use on embedded devices where space (and thus code size) is at a premium.  A developer would typically build their own uclibc with only the features required for their use case.  uclibc facilitates this customization by having a configuration step (via a menuconfig, like the kernel menuconfig system, if you've ever used that) where the developer chooses the features they like, including whether to use a 32-bit or 64-bit `time_t`, and then builds their own uclibc for their project.

So we cannot say for sure within the `libc` crate which size of `time_t` will be present in the underlying uclibc library.

The bootlin folks have chosen to enable the 64-bit `time_t` feature on *their* build of uclibc as of the toolchain noted above.  Having a reference toolchain is really useful for ease of CI and developer bootstrapping purposes, but the uclibc configuration choices that the bootlin toolchain makes aren't necessarily the same (nor binary compatible) with any other developer's use of uclibc.

This is why I expect everyone using uclibc with rust ends up either using `-Zbuild-std` or building their own cross toolchain, in either case targeting their project-specific uclibc headers, and (as of this change) making use of the `--cfg=libc_unstable_uclibc_time64` rustflag if needed.

src/unix/linux_like/linux/uclibc/arm/mod.rs
skrap: I am relying on the libc-test infrastructure a bit here to catch errors, but I think everything should be OK.  I am happy to be the point person for any issues raised for 64-bit `time_t` in uclibc, if bugs are reported.

src/unix/linux_like/linux_l4re_shared.rs
skrap: > could you update them to not(target_os = "l4re")

I think your request is reasonable, but I would like to ask that the requested l4re change go into a separate PR, as removing those constants for the l4re build could easily introduce breakage, and I don't know anything about that platform.

skrap (commented):
@rustbot ready

Thanks for the review!  I have replied inline to the issues raised.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/ff0e6164f209073c3856e3709c4dcae1e59c6a11..8f8451e945bc8a2ecacd49c7905ef4272cd2d3c6)*

src/unix/linux_like/linux_l4re_shared.rs
farao: To answer the question a few comments above: yes, all constants that were disabled for uclibc are not present on l4re. I would of course prefer if they stayed disabled for l4re in one way or the other

skrap:
Hi @tgross35 , sorry to bother you.  I wondered if I could get some of your time on this request, as there are starting to be other areas of work piled up behind it.  Thanks for any time you can spare.

skrap:
I have fixed the conflicts generated by https://github.com/rust-lang/libc/pull/5141.  The CI results above are nonsense AFAICT.  They seem to be testing unrelated changes.

ci/docker/armv7-unknown-linux-uclibceabihf/Dockerfile
```diff
@@ -14,11 +14,9 @@ RUN sed -i -E 's/(archive|security)\.ubuntu\.com/old-releases.ubuntu.com/g' \
         qemu-user \
         xz-utils
 
-RUN mkdir /toolchain
-
-RUN curl --retry 5 -L https://toolchains.bootlin.com/downloads/releases/toolchains/armv7-eabihf/tarballs/armv7-eabihf--uclibc--bleeding-edge-2021.11-1.tar.bz2 | \
-    tar xjf - -C /toolchain --strip-components=1
-RUN /toolchain/relocate-sdk.sh
+ARG TEST_UCLIBC_TIIME64
```
JohnTitor: typo?

```suggestion
ARG TEST_UCLIBC_TIME64
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/342994381dbb4c8c753e41034aaeda01f6df13b9..d6f3bc87a08e83bc08784afb145223a328f332c8)*

ci/install-uclibc.sh
```diff
@@ -0,0 +1,20 @@
...
+# Installs the appropriate uclibc toolchain into /toolchain
+
+set -eux
+
+time64="$1"
+
+if [ "${time64:-0}" != "0" ]; then
+    version='bleeding-edge-2025.08-1'
+else
+    version='bleeding-edge-2024.05-1'  # last version with 32-bit time_t
```
JohnTitor: I've checked the tarball but seems it defines `__UCLIBC_USE_TIME64__`, could you re-confirm if it really uses 32-bit version?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/342994381dbb4c8c753e41034aaeda01f6df13b9..d6f3bc87a08e83bc08784afb145223a328f332c8)*

ci/install-uclibc.sh
JohnTitor, line 1: I've checked the permission but it doesn't seem executable:
```
> ls -l ci/install-uclibc.sh
.rw-r--r--
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/342994381dbb4c8c753e41034aaeda01f6df13b9..d6f3bc87a08e83bc08784afb145223a328f332c8)*

src/unix/linux_like/linux_l4re_shared.rs
JohnTitor: @skrap I would accept _breakage_ here as there shouldn't exist real user on l4re give the comments from Marius. Could you address the suggestion?

ci/docker/armv7-unknown-linux-uclibceabihf/Dockerfile
skrap: Ah, sorry, earlier in the review I was informed that CI for this target would not be enabled as the target is tier 3, so I disabled the CI integration and left it untested.  Thanks for the suggestion.

ci/install-uclibc.sh
skrap: Yes, you are right.   This ci code was untested.  The correct last version without time64 was bleeding-edge-2024.02-1.  (The PR description is correct.)

ci/install-uclibc.sh
skrap: Addressed.

src/unix/linux_like/linux_l4re_shared.rs
skrap: Addressed.

skrap (commented):
Addressed review feedback

@JohnTitor I had previously been told that tier 3 platforms would not run in CI, but if you are able to run the CI test manually somehow, that would be helpful to know if it works!  Otherwise if you'd prefer to have the untested CI code removed, I can do that.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/c4ea92b6ccf512cd1caff2c6cf3a401506badff7..46cea20e70324c176df05181422ef6546a6a10b0)*

ci/install-uclibc.sh
tgross35: My feeling is this unfortunately isn't very sustainable. It's not ideal to have all the old kernel version skips in `build.rs`, they're going to be perpetually out of date.

How hard is it to build uclibc? We're already building musl from source, doing the same here would probably be best.

No need to do that in this PR but I do think you should drop the `2024.02-1` option from this script, and any test skips that go along with it. Then switch to building from source in a followup.

libc-test/build.rs
```diff
@@ -3738,6 +3738,12 @@ fn test_linux(target: &str) {
             cfg.cfg("musl_redir_time64", None);
         }
     }
+    let uclibc_use_time64 = env::var("CARGO_CFG_LIBC_UNSTABLE_UCLIBC_TIME64")
+        .map(|val| val != "0")
+        .unwrap_or(false);
```
tgross35, line 3757: For libc 1.0 I'd like the default to be 64-bit `time_t` on all platforms that support it, so if it's going to be valuable to continue supporting 32-bit then I think might be better as `CARGO_CFG_LIBC_UNSTABLE_UCLIBC_TIME_BITS` eventually, set to either "32" or "64".

No need to do this change here, just an FYI for the future.

.github/workflows/ci.yaml
tgross35: Just noting that it's pretty easy to set up a downstream buildbot that runs tests - you can have an empty GitHub repo with a scheduled period CI job that clones this repo and tests. I think a few other T3 platforms do that.

libc-test/build.rs
tgross35: > I was unsure of the philosophy here with respect to kernel-defined constants. They would either be present in `linux/*.h` headers or not depending on the kernel version, regardless of whatever libc library/version was being used on the system. So what triggers a particular constant being promoted into the `libc` crate?

We assume a recent kernel, which for most platforms is usually whatever GitHub Actions gives us. API may be added whenever it's in a released kernel, might just need a test skip if we're testing with something older.

tgross35 (changes requested):
I have one last thing, I think we should drop the time32 test config because it's literally always going to be failing. But after that, diff LGTM. Please squash the "Addressing review feedback" commit into one of the previous commits.

Very sorry about the delay here, I had pretty much no time to review anything the past couple of months. I should be back now.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/1a9bdc4d16edc21dfa66e655d1edf8e4eac16ba8..cc02669889ed0670470ecdd3c730fe3bea887b77)*

ci/install-uclibc.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/bash
```
tgross35: Sorry, one other thing: just use `#!/bin/bash`, the path under `usr` isn't available on all platforms.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/1a9bdc4d16edc21dfa66e655d1edf8e4eac16ba8..cc02669889ed0670470ecdd3c730fe3bea887b77)*

ci/install-uclibc.sh
skrap: It's definitely feasible to build uclibc in CI.  I can take this as a task for a future improvement, especially if some uclibc target would ever become tier 2.  Right now I can't really figure out how to make the Docker script work, however, so I will leave it as vestigial.

(In case it matters, the issue I'm hitting is that the macro expansion code in libc-test is complaining that rustc fails with `can't find crate for core`, which presumably *can* be made to work since tier 2 platforms work, but I'm not sure how! If you have insights I'll make it work in a separate PR.)

I would really like to get this change in, so would you please accept the PR without me going back and re-auditing each kernel constant? It is really time consuming, and isn't a standard which is applied to any other target's script, as far as I can tell.  I would be happy to do so in a future PR.

Alternatively, I could just delete the CI infrastructure associated with this target, as it cannot work and is not run.

libc-test/build.rs
skrap: Thanks, noted.  I'm fine with this plan.

skrap:
Addressed much of the feedback, but did not alter the libc test yet. I will do so in a future PR.

@rustbot review

ci/install-uclibc.sh
tgross35: > (In case it matters, the issue I'm hitting is that the macro expansion code in libc-test is complaining that rustc fails with `can't find crate for core`, which presumably _can_ be made to work since tier 2 platforms work, but I'm not sure how! If you have insights I'll make it work in a separate PR.)

The T2 targets have dist artifacts so we can just use rustup, which happens here https://github.com/rust-lang/libc/blob/d73763fe397958f427b360fdc76da56df44f1a89/ci/install-rust.sh#L27-L28. You'll need to make sure Cargo is getting `-Zbuild-std=core`, can you see if setting this env works? https://github.com/rust-lang/libc/blob/d73763fe397958f427b360fdc76da56df44f1a89/ci/run.sh#L20

> I would really like to get this change in, so would you please accept the PR without me going back and re-auditing each kernel constant? It is really time consuming, and isn't a standard which is applied to any other target's script, as far as I can tell. I would be happy to do so in a future PR.
> 
> Alternatively, I could just delete the CI infrastructure associated with this target, as it cannot work and is not run.

No audit needed, I meant that you should delete the skips under "The canonical uClibc toolchain, bootlin bleeding-edge-2024.02-1, uses linux 5.15, so several constants are not available." since nothing else in the build script needs to be updated for a kernel that old.

But that's all fair enough, this has been waiting a while so I'll apply it. Just send the CI PR when you have it ready.

tgross35 (approved):
Thank you for the changes!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5046/d73763fe397958f427b360fdc76da56df44f1a89..6fcae129a275ad9a8d55b63b196989d9c4516bf4)*

ci/install-uclibc.sh
skrap: Thanks, but the macro expansion is not done via cargo build, but rather `rustc` directly, so the `-Zbuild-std` flag is not available.  See [ctest/src/macro_expansion.rs](https://github.com/rust-lang/libc/blob/d73763fe397958f427b360fdc76da56df44f1a89/ctest/src/macro_expansion.rs#L19-L24) for where this is done.  

I don't understand how this could work without a built std lib.  Do you know?

tgross35: Ah, I didn't realize that was what was failing. I'm not really sure what's best there. Maybe invoking via `cargo rustc` rather than `rustc` directly would work?

skrap: Tried that already, sadly.  :( no dice.

tgross35: It seems like that should work, what failed?

If you wouldn't mind creating a new issue for this, that would probably be ideal so it doesn't get lost here.

skrap: Filed https://github.com/rust-lang/libc/issues/5204
Thanks for the help on this!

PR #5050, opened 2026-04-10
feat: add back support for GNU Windows x86 in CI

dybucc:
# Description

After having tested out GNU on Windows x86, it seemed like the only issue was
related to a wrongly aligned/sized `max_align_t`. This has been fixed, and the
test suite seems to be running just fine in CI now for the above
platform/environment.

There were also some environment variables in CI that seemed to be set for the
purposes of changing the Mingw toolchain set up, but that weren't truly being
used as the job that ran the script that used those was being triggered earlier
in the CI pipeline. Because CI logs seem to reveal this has been the case for
some time, this has also been altogether removed.

# Sources

- Mingw source code outlining the structure layout of `max_align_t`
  https://github.com/mingw-w64/mingw-w64/blob/9b3dd0125792fe94d16cacdc596dbd42fca1b369/mingw-w64-headers/crt/stddef.h#L424-L427

# Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -125,15 +125,13 @@ jobs:
             os: windows-11-arm
           - target: aarch64-unknown-linux-gnu
             os: ubuntu-24.04-arm
-          # FIXME: It currently causes segfaults.
-          #- target: i686-pc-windows-gnu
-          #  env: { ARCH_BITS: 32, ARCH: i686 }
+          - target: i686-pc-windows-gnu
+            os: windows-2025
           - target: i686-pc-windows-msvc
             os: windows-2025
           - target: i686-unknown-linux-gnu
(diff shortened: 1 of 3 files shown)


dybucc:
Just run CI on my fork with this exact same branch and there were no issues. I
also don't think there should be any issues, considering the changes in this PR
only affect Windows x86, and CI seems to fail in Linux AArch64.

Maybe running CI again should solve it.

JohnTitor (approved):
Oh, cool! Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5050/ff0e6164f209073c3856e3709c4dcae1e59c6a11..82298802bfd427fe1fe65a7d1bdf48154dbda403)*

PR #5058, opened 2026-04-14
chore: migrate from Cirrus CI to GHA

### LINKED ISSUE #5052: CirrusCI will shut down on 2026-06-01
https://cirruslabs.org/#:~:text=Cirrus%20CI%20will%20shut%20down%20effective%20Monday%2C%20June%201%2C%202026.

JohnTitor:
# Description

Fix https://github.com/rust-lang/libc/issues/5052

This migrates FreeBSD CI from Cirrus CI to GitHub Actions to address Cirrus CI shutdown.
The new GHA matrix should have the same workflow as Cirrus as possible.

# Sources

# Checklist


JohnTitor:
NetBSD CDN is down but it's unrelated to this change, ready for review.

tgross35 (approved):
Looks great to me, but maybe hold off for a day in case @asomers gets the chance to take a look.

Also maybe update the commit/PR description to say it's migrating to GHA? From the quick read I assumed you were just deleting the job :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5058/f3417ae86f015b9d0d8789b8fa11e64f9fec0adf..11836d284b1c53f53a91ea67b73a167a5186a1e1)*

.github/workflows/ci.yaml
```diff
@@ -274,11 +274,45 @@ jobs:
...
     timeout-minutes: 25
     steps:
       - uses: actions/checkout@v6
+      - name: Test on FreeBSD
+        uses: vmactions/freebsd-vm@v1
+        if: contains(matrix.target, 'freebsd')
+        with:
+          release: ${{ matrix.release }}
+          usesh: true
+          mem: 4096
```
asomers: What's wrong with the default memory?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5058/f3417ae86f015b9d0d8789b8fa11e64f9fec0adf..11836d284b1c53f53a91ea67b73a167a5186a1e1)*

.github/workflows/ci.yaml
```diff
@@ -274,11 +274,45 @@ jobs:
...
+          copyback: false
+          prepare: |
+            set -eux
+            export HOME=/tmp
```
asomers: This workflow isn't using cargo cache.  It certainly isn't, since you set `copyback: false`.  So you can delete this line.
```suggestion
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5058/f3417ae86f015b9d0d8789b8fa11e64f9fec0adf..11836d284b1c53f53a91ea67b73a167a5186a1e1)*

.github/workflows/ci.yaml
```diff
@@ -274,11 +274,45 @@ jobs:
...
+      - name: Test on FreeBSD
+        uses: vmactions/freebsd-vm@v1
+        if: contains(matrix.target, 'freebsd')
+        with:
+          release: ${{ matrix.release }}
+          usesh: true
+          mem: 4096
+          copyback: false
+          prepare: |
+            set -eux
```
asomers: Why the "-u" ?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5058/f3417ae86f015b9d0d8789b8fa11e64f9fec0adf..11836d284b1c53f53a91ea67b73a167a5186a1e1)*

tgross35: I think it's good practice to always set this in scripts. Expected nullable variables are rare enough to have `:-` on a case-by-case basis.

.github/workflows/ci.yaml
JohnTitor: Copied from the NetBSD with no thinking, sorry!
Removed in https://github.com/rust-lang/libc/pull/5058/changes/8c3f4fe8725f078e1fd970ffc081d5d585fefb2e

.github/workflows/ci.yaml
JohnTitor: Removed: https://github.com/rust-lang/libc/pull/5058/changes/8c3f4fe8725f078e1fd970ffc081d5d585fefb2e

.github/workflows/ci.yaml
JohnTitor: I guess I copied it from somewhere while referencing a prior-art PR 🙇 
Addressed in https://github.com/rust-lang/libc/pull/5058/changes/8c3f4fe8725f078e1fd970ffc081d5d585fefb2e

JohnTitor:
> Also maybe update the commit/PR description to say it's migrating to GHA? From the quick read I assumed you were just deleting the job :)

Renamed, it was too cool :P

.github/workflows/ci.yaml
```diff
@@ -274,11 +274,42 @@ jobs:
...
+            set -x
+            pkg install -y libnghttp2 curl
+            curl https://sh.rustup.rs -sSf --output rustup.sh
+            sh rustup.sh -y --default-toolchain nightly --profile=minimal
+            . "$HOME/.cargo/env"
+            if [ "${{ matrix.target }}" = "i686-unknown-freebsd" ]; then
+                rustup target add i686-unknown-freebsd
+            fi
+          run: |
+            set -x
```
asomers: I think you still want `-e` here, or else the job will be considered passing even if the first run.sh invocation fails.
```suggestion
            set -xe
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5058/f3417ae86f015b9d0d8789b8fa11e64f9fec0adf..8c3f4fe8725f078e1fd970ffc081d5d585fefb2e)*

asomers (changes requested):

.github/workflows/ci.yaml
JohnTitor: Should I apply it to NetBSD scripts? I guess it benefits on it as well.

asomers: > Should I apply it to NetBSD scripts? I guess it benefits on it as well.

Yes, please do.  Or, just remove the `set -x`, because the default is equivalent to `-e`.

JohnTitor: Addressed, thanks!

asomers (approved):

.github/workflows/ci.yaml
```diff
@@ -274,11 +274,42 @@ jobs:
...
+          copyback: false
+          prepare: |
+            set -ex
+            pkg install -y libnghttp2 curl
+            curl https://sh.rustup.rs -sSf --output rustup.sh
+            sh rustup.sh -y --default-toolchain nightly --profile=minimal
+            . "$HOME/.cargo/env"
+            if [ "${{ matrix.target }}" = "i686-unknown-freebsd" ]; then
+                rustup target add i686-unknown-freebsd
+            fi
```
xtqqczze:
```suggestion
            rustup target add ${{ matrix.target }}
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5058/f3417ae86f015b9d0d8789b8fa11e64f9fec0adf..326bbfd11f9bbecc4b0daad5a37572579100af5f)*

.github/workflows/ci.yaml
tgross35: > Yes, please do. Or, just remove the `set -x`, because the default is equivalent to `-e`.

Why would this make a difference? `set` only changes the specified options and doesn't override others, unless GH is doing some inspection here.

Not that I think it's a bad thing to be explicit.

.github/workflows/ci.yaml
tgross35: On others the default host toolchain works. But also no harm since rustup does the same check, 🤷‍♂️

.github/workflows/ci.yaml
tgross35: Oh I was saying the other way around, `-u` should always be included. Not that it's super useful here since the only thing to catch is unset `$HOME` (which is actually a problem we've historically hit with some runners).

tgross35 (approved):
LGTM with or without @xtqqczze's suggestion, thank you for taking care of this!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5058/f3417ae86f015b9d0d8789b8fa11e64f9fec0adf..326bbfd11f9bbecc4b0daad5a37572579100af5f)*

.github/workflows/ci.yaml
xtqqczze: It's less code, and would make things future-proof in case targets like `aarch64-unknown-freebsd` are added to the job matrix later.

asomers: The funny thing about i686-unknown-freebsd is that we don't run that check in an i686 VM.  We run it in an x86_64 VM, using 32-bit compat.  So the `rustup target add` really is needed.

.github/workflows/ci.yaml
```diff
@@ -274,11 +274,40 @@ jobs:
...
+            target: x86_64-unknown-freebsd
+          - release: "15.0"
+            target: x86_64-unknown-freebsd
           - target: x86_64-pc-solaris
           - target: x86_64-unknown-netbsd
     timeout-minutes: 25
     steps:
       - uses: actions/checkout@v6
+      - name: Test on FreeBSD
+        uses: vmactions/solaris-vm@v1.3.2
```
asomers: Oops, looks like a copy/paste error.
```suggestion
        uses: vmactions/freebsd-vm@v1
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5058/f3417ae86f015b9d0d8789b8fa11e64f9fec0adf..4eaaff50b9e4a4ad7f6462430090d08194e16725)*

asomers (changes requested):

.github/workflows/ci.yaml
xtqqczze: Right, so `rustup target add` wouldn't be needed for `aarch64-unknown-freebsd`; `i686-unknown-freebsd` really is a special case.

.github/workflows/ci.yaml
JohnTitor: Ah, good catch, fixed!

asomers (approved):

.github/workflows/ci.yaml
xtqqczze: https://github.com/rust-lang/libc/pull/5064

PR #5059, opened 2026-04-14
windows(gnu): link to 32-bit time routines in x86 and add test

dybucc:
## Description

Until rust-lang/libc#5032, whenever `libc` was running under Windows x86 with GNU, the default
bit-width of `time_t` values was 32-bits. This doesn't quite align with the defaults in MSVC and
`mingw`, but that's been addressed in that PR.

This PR is a quick patch for stable, as the changes in rust-lang/libc#5032 are not
backwards-compatible. It separates the routines that under Windows use `time_t` values and changes
their link-time symbols to the 32-bit variants exposed in both MSVC and `mingw`, for the affected
target platform/environment.

It also adds a test, akin to the one in rust-lang/libc#5032, that should ensure the correct
bit-widths are in use for both the types and the linked routine symbols.

Note that if both rust-lang/libc#5062 and this PR are to be merged, it's preferable to merge first rust-lang/libc#5062, as this
PR will then incorporate support for the `cfg` introduced in that PR.

### Sources

None. If any, the same as rust-lang/libc#5032 as this stems from it.

### Checklist

- [x] No placeholder or unstable values like `*LAST` or `*MAX` are included (see
  [#3131](https://github.com/rust-lang/libc/issues/3131))

### DIFF
--- a/libc-test/tests/windows_time.rs
+++ b/libc-test/tests/windows_time.rs
@@ -1,0 +1,31 @@
+//! Ensures Windows `time`-related routines align with `libc`'s interface. By
+//! default, both MSVC and GNU (under `mingw`) expose 64-bit symbols, but in
+//! stable we need to cope with a 32-bit `time-t`, so the routines should link
+//! to their 32-bit variants.
+
+#![cfg(windows)]
+
+/// Ensures a 64-bit write is performed on values that should always be 64 bits,
+/// and that a corresponding 32-bit write is performed on values that should be
+/// 32 bits. This basically makes sure `time_t`'s bit-width aligns with those of
+/// functions that expect it as a parameter.
+#[test]
+fn test_bitwidth_store() {
+    if cfg!(all(target_arch = "x86", target_env = "gnu")) {
+        assert_eq!(size_of::<libc::time_t>(), 4);
+    } else {
+        assert_eq!(size_of::<libc::time_t>(), 8);
+    }
+
+    let mut time_values: [libc::time_t; 2] = [123, 456];
+    let ptr = time_values.as_mut_ptr();
+
+    unsafe { libc::time(ptr) };
+
+    // Succeeds if the first array value is written. Fails if both elements are
+    // overwritten. That would imply:
+    // - We are linking a `time()` that expects a 64-bit `time_t`.
+    // - We are exposing and using here a 32-bit `time_t`.
+    assert_ne!(time_values[0], 123);
+    assert_eq!(time_values[1], 456);
+}
--- a/src/windows/mod.rs
+++ b/src/windows/mod.rs
@@ -389,11 +389,33 @@ extern "C" {
     pub fn signal(signum: c_int, handler: sighandler_t) -> sighandler_t;
     pub fn raise(signum: c_int) -> c_int;
 
+    // By default, the following link to 64-bit variants where the expected
+    // `time_t` is also 64-bits.
+    //
+    // Under Windows x86 with GNU, `time_t` is still 32-bit wide on stable, so
+    // these routines have to link with their 32-bit variants.
+    cfg_if! {
+        if #[cfg(all(target_arch = "x86", target_env = "gnu"))] {
+            #[link_name = "_ctime32"]
+            pub fn ctime(sourceTime: *const time_t) -> *mut c_char;
+            #[link_name = "_difftime32"]
+            pub fn difftime(timeEnd: time_t, timeStart: time_t) -> c_double;
+            #[link_name = "_gmtime32_s"]
+            pub fn gmtime_s(destTime: *mut tm, srcTime: *const time_t) -> c_int;
+            #[link_name = "_localtime32_s"]
+            pub fn localtime_s(tmDest: *mut tm, sourceTime: *const time_t) -> crate::errno_t;
+            #[link_name = "_time32"]
+            pub fn time(destTime: *mut time_t) -> time_t;
+        } else {
+            pub fn ctime(sourceTime: *const time_t) -> *mut c_char;
+            pub fn difftime(timeEnd: time_t, timeStart: time_t) -> c_double;
+            pub fn gmtime_s(destTime: *mut tm, srcTime: *const time_t) -> c_int;
+            pub fn localtime_s(tmDest: *mut tm, sourceTime: *const time_t) -> crate::errno_t;
+            pub fn time(destTime: *mut time_t) -> time_t;
+        }
+    }
+
     pub fn clock() -> clock_t;
-    pub fn ctime(sourceTime: *const time_t) -> *mut c_char;
-    pub fn difftime(timeEnd: time_t, timeStart: time_t) -> c_double;
-    #[link_name = "_gmtime64_s"]
-    pub fn gmtime_s(destTime: *mut tm, srcTime: *const time_t) -> c_int;
     #[link_name = "_get_daylight"]
     pub fn get_daylight(hours: *mut c_int) -> errno_t;
     #[link_name = "_get_dstbias"]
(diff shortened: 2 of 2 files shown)


dybucc:

src/windows/mod.rs
```diff
@@ -389,11 +389,36 @@ extern "C" {
...
     pub fn raise(signum: c_int) -> c_int;
 
+    // By default, the following link to 64-bit variants where the expected
+    // `time_t` is also 64-bits.
+    //
+    // Under Windows x86 with GNU, `time_t` is still 32-bit wide on stable, so
+    // these routines have to link with their 32-bit variants.
+    //
+    // FIXME(#5062): once that PR gets merged, add `not(windows_gnu_time64)` to
+    // the below predicate, within the `all()` clause.
```
tgross35: This should be right around the corner, so let's just wait on it.

@rustbot blocked

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5059/8132431a05a94bfce91b56fd1bac74e224d9230a..288ec56247011947a1bee4f54858d1ac431ff897)*

dybucc: The PR should be ready now.

@rustbot ready

libc-test/tests/windows_time.rs
```diff
@@ -0,0 +1,26 @@
...
+/// Ensures a 64-bit write is performed on values that should always be 64 bits,
+/// and that a corresponding 32-bit write is performed on values that should be
+/// 32 bits. This basically makes sure `time_t`'s bit-width aligns with those of
+/// functions that expect it as a parameter.
+#[test]
+fn test_bitwidth_store() {
+    #[cfg(all(target_arch = "x86", target_env = "gnu", not(gnu_time_bits64)))]
+    assert_eq!(size_of::<libc::time_t>(), 4);
+    #[cfg(any(not(target_arch = "x86", target_env = "gnu"), gnu_time_bits64))]
+    assert_eq!(size_of::<libc::time_t>(), 8);
```
tgross35: Nit: prefer `if cfg!(...)` over `#[cfg(...)]` where possible, the advantage is that it validates all branches even when the cfg doesn't match.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5059/69c52f243cf5497e90a59875a4443171c782edfc..8b5714c6cb675af7eb12065fd2b36cb79ca12c03)*

libc-test/tests/windows_time.rs
```diff
@@ -0,0 +1,26 @@
...
+
+    let mut time_values: [libc::time_t; 2] = [123, 456];
+    let ptr = time_values.as_mut_ptr();
+
+    unsafe { libc::time(ptr) };
+
+    assert_ne!(time_values[0], 123);
+    assert_eq!(time_values[1], 456);
```
tgross35: Suggested comment based on https://github.com/rust-lang/libc/pull/5032#discussion_r3063027805:

> If only the first array value is written, libc's `time_t` is accurate for what the function `time` expects. If both array values wind up overwritten, we are probably linking a version of `time` that expects 64-bit `time_t` while our `time_t` is only 32 bits.

Also can you confirm that this fails with the current version?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5059/69c52f243cf5497e90a59875a4443171c782edfc..8b5714c6cb675af7eb12065fd2b36cb79ca12c03)*

src/windows/mod.rs
```diff
@@ -389,11 +389,33 @@ extern "C" {
     pub fn signal(signum: c_int, handler: sighandler_t) -> sighandler_t;
     pub fn raise(signum: c_int) -> c_int;
 
+    // By default, the following link to 64-bit variants where the expected
+    // `time_t` is also 64-bits.
+    //
+    // Under Windows x86 with GNU, `time_t` is still 32-bit wide on stable, so
+    // these routines have to link with their 32-bit variants.
+    cfg_if! {
+        if #[cfg(all(target_arch = "x86", target_env = "gnu", not(gnu_time_bits64)))] {
```
tgross35: Move anything `gnu_time_bits64` bit to https://github.com/rust-lang/libc/pull/5062, since it's not getting fully tested here and that PR doesn't currently do anything on its own.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5059/69c52f243cf5497e90a59875a4443171c782edfc..8b5714c6cb675af7eb12065fd2b36cb79ca12c03)*

tgross35 (changes requested):

libc-test/tests/windows_time.rs
dybucc: I can. CI doesn't catch this because we skip over `time_t` on Windows. I removed that in #5062.

src/windows/mod.rs
tgross35: I meant only the `not(gnu_time_bits64)` bit of the config, given the test failures and lack of it being tested. The link fixes are fine, that should be in this PR because it's a standalone fix. 

Though if it's easier for you at this point to roll the two PRs together, that's fine too

dybucc: Reverted and done. This PR should then be merged first. I'll solve the conflicts in #5062 then.


EDIT: No good. Reverted again. The linking issues remain solved in this PR. The other one will have to add the condition to the predicate once this one gets merged.

libc-test/tests/windows_time.rs
```diff
@@ -0,0 +1,32 @@
...
+    let mut time_values: [libc::time_t; 2] = [123, 456];
+    let ptr = time_values.as_mut_ptr();
+
+    unsafe { libc::time(ptr) };
+
+    // Succeeds if the first array value is written. Fails if both elements are
+    // overwritten. That would imply:
+    // - We are linking a `time()` that expects a 64-bit `time_t`.
+    // - We are exposing and using here a 32-bit `time_t`.
+    assert_matches!(time_values, [n, 456] if n != 123);
```
tgross35: Nit: the original with two asserts was easier to read

tgross35 (approved):
One nit and history cleanup then LGTM

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5059/076d48ffbddf9408319f72ef7361452f8b0206d1..a07e2c55af4acc42ae568b6411b65f720eb3b9b6)*

tgross35:
Question: why doesn't this cause the function pointer comparison tests to fail on win32-gnu?

dybucc:
> Question: why doesn't this cause the function pointer comparison tests to fail on win32-gnu?

Function pointer checks are skipped on Windows [[1]]. I've tried to look into this. I haven't found the reason why stuff is failing. The decorated names in MSVC seem to use `cdecl` but the unresolved errors to all exposed routines also happen in x86_64. I've experimented adding `cdecl` under x86. This does not seem to work. Function pointer checks still fail in x86_64 when they shouldn't. This includes non-`time_t` routines. There's no documentation on the linking of decorated names in MSVC either.

[1]: https://github.com/rust-lang/libc/blob/main/libc-test/build.rs#L873

tgross35:
Only on MSVC or on GNU as well? We could at least limit the skips if it's the former.

Calling convention shouldn't affect the function pointer checks regarding address mismatches. It would be nice if we could check calling convention in ctest somehow but I don't believe those mismatches show up anywhere.

You might get a hint if you make a Rust library where you call a `libc` symbol and a C library where you call the same symbol, then `llvm-nm` them if there's a mismatch in the undefined symbols list. But I don't think it's worth digging into too much.

tgross35:
Mind just opening a new issue with your findings so we don't lose this?

dybucc:
It happens on both MSVC and GNU. Just opened the issue.

src/windows/mod.rs
```diff
@@ -407,10 +429,6 @@ extern "C" {
         size_in_bytes: size_t,
         index: c_int,
     ) -> errno_t;
-    #[link_name = "_localtime64_s"]
-    pub fn localtime_s(tmDest: *mut tm, sourceTime: *const time_t) -> crate::errno_t;
-    #[link_name = "_time64"]
-    pub fn time(destTime: *mut time_t) -> time_t;
     #[link_name = "_tzset"]
```
tgross35, line 432: I missed that these removed link names, which broke some things in the latest release. @dybucc would you be able to prepare a PR adding them back? If not I'll get to it a bit later.

tgross35: Covered in https://github.com/rust-lang/libc/pull/5300

PR #5062, opened 2026-04-16
windows: expose `cfg` for 64-bit `time_t`

dybucc:
# Description

Under Windows GNU x86, `time_t` doesn't have the right size and alignment. This is a backwards-incompatible change, so for folks to experiment on stable, a `cfg` has been exposed to allow exposing a 64-bit `time_t` instead. Tests don't fail because we still skip `time_t` and records containing some value of such type whenever we're under this target platform/environment.

In rust-lang/libc#5059, we address the issue with routines involving `time_t` values by making them link to their 32-bit symbols on that platform/environment.

This should allow not breaking existing code (even though incorrect,) while still allowing new code to experiment with correct size/alignment for `time_t` values. The `cfg` tries to follow the new practices introduced in rust-lang/libc#4977.

An additional job was not added to the test matrix in CI because it was instead introduced as part of the tests in `ci/run.sh` that run only under 32-bit machine word and GNU-hosted targets.

## Sources

None, if any the same as rust-lang/libc#5032 as this stems from it.

## Checklist

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -24,7 +24,8 @@ const ALLOWED_CFGS: &[&str] = &[
     "freebsd15",
     // Corresponds to `_FILE_OFFSET_BITS=64` in glibc
     "gnu_file_offset_bits64",
-    // Corresponds to `_TIME_BITS=64` in glibc
+    // Corresponds to `_TIME_BITS=64` in glibc. Also used in x86 Windows with
+    // GNU to expose a 64-bit `time_t`.
     "gnu_time_bits64",
     "libc_deny_warnings",
     // Corresponds to `__USE_TIME_BITS64` in UAPI
@@ -144,7 +145,7 @@ fn main() {
     }
 
     if target_env == "gnu"
-        && target_os == "linux"
+        && matches!(target_os.as_str(), "linux" | "windows")
         && target_ptr_width == "32"
         && target_arch != "riscv32"
         && target_arch != "x86_64"
@@ -173,7 +174,7 @@ fn main() {
         let (timebits, filebits) = match (tb_env.as_deref(), fb_env.as_deref()) {
             (Ok(_), Ok(_)) => panic!(
                 "Do not set both libc_unstable_gnu_time_bits and \
-                libc_unstable_gnu_file_offset_bits"
+                 libc_unstable_gnu_file_offset_bits"
             ),
             (Err(_), Err(_)) => (defaultbits, defaultbits),
             (Ok(tb), Err(_)) if tb == "64" => (tb, tb),
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -784,6 +784,23 @@ fn test_windows(target: &str) {
     }
     cfg.define("_WIN32_WINNT", Some("0x8000"));
 
+    let win_gnu_x86_time64 = match env::var("CARGO_CFG_LIBC_UNSTABLE_GNU_TIME_BITS") {
+        Ok(v) if v == "64" => true,
+        Ok(v) if v == "32" => false,
+        Ok(_) => {
+            panic!("Invalid value for `libc_unstable_gnu_time_bits`. Must be 32, 64 or unset.");
+        }
+        Err(_) => false,
+    };
+
+    // Needed for the Windows `time_t` test.
+    println!("cargo::rustc-check-cfg=cfg(gnu_time_bits64)");
+
+    if i686 && gnu && win_gnu_x86_time64 {
+        cfg.cfg("gnu_time_bits64", None);
+        println!("cargo::rustc-cfg=gnu_time_bits64");
+    }
+
     headers!(
         cfg,
         "direct.h",
@@ -841,7 +858,7 @@ fn test_windows(target: &str) {
         "SSIZE_T" if !gnu => true,
         "ssize_t" if !gnu => true,
         // FIXME(windows): The size and alignment of this type are incorrect
-        "time_t" if gnu && i686 => true,
+        "time_t" if gnu && i686 && !win_gnu_x86_time64 => true,
         _ => false,
     });
 
--- a/src/windows/mod.rs
+++ b/src/windows/mod.rs
@@ -19,7 +19,13 @@ pub type clock_t = i32;
 pub type errno_t = c_int;
 
 cfg_if! {
-    if #[cfg(all(target_arch = "x86", target_env = "gnu"))] {
+    // FIXME(1.0): Windows GNU has 64-bit `time_t` by default. In x86 this is
+    // wrongly bound.
+    if #[cfg(all(
+        target_arch = "x86",
+        target_env = "gnu",
+        not(gnu_time_bits64)
+    ))] {
         pub type time_t = i32;
     } else {
         pub type time_t = i64;
@@ -395,7 +401,11 @@ extern "C" {
     // Under Windows x86 with GNU, `time_t` is still 32-bit wide on stable, so
     // these routines have to link with their 32-bit variants.
     cfg_if! {
-        if #[cfg(all(target_arch = "x86", target_env = "gnu"))] {
+        if #[cfg(all(
+            target_arch = "x86",
+            target_env = "gnu",
+            not(gnu_time_bits64),
+        ))] {
             #[link_name = "_ctime32"]
             pub fn ctime(sourceTime: *const time_t) -> *mut c_char;
             #[link_name = "_difftime32"]
--- a/libc-test/tests/windows_time.rs
+++ b/libc-test/tests/windows_time.rs
@@ -11,7 +11,11 @@
 /// functions that expect it as a parameter.
 #[test]
 fn test_bitwidth_store() {
-    if cfg!(all(target_arch = "x86", target_env = "gnu")) {
+    if cfg!(all(
+        target_arch = "x86",
+        target_env = "gnu",
+        not(gnu_time_bits64),
+    )) {
         assert_eq!(size_of::<libc::time_t>(), 4);
     } else {
         assert_eq!(size_of::<libc::time_t>(), 8);


dybucc:

build.rs
```diff
@@ -131,6 +134,14 @@ fn main() {
         }
     }
 
+    if target_arch == "x86"
+        && target_os == "windows"
+        && target_env == "gnu"
+        && env_flag("CARGO_CFG_LIBC_UNSTABLE_WINDOWS_GNU_TIME64")
+    {
+        set_cfg("windows_gnu_time64");
+    }
```
tgross35: Being this is only set on 32-bit platforms rather than on all platforms that have 64-bit time, I think a name like `windows_gnu32_time64` or `win32_gnu_time64` would be preferable.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5062/8132431a05a94bfce91b56fd1bac74e224d9230a..7050e1728075027181f4f4bc835503d256144ae6)*

tgross35: Actually, could we just reuse `gnu_time_bits64` for the Windows config? Since that applies here too.

We should probably look at `gnu_file_offset_bits64` at some point too, since `#define _FILE_OFFSET_BITS 64` seems to change `off_t` on win-gnu https://c.godbolt.org/z/fPrzjGzfx. Wrote this up at https://github.com/rust-lang/libc/issues/5181.

ci/run.sh
```diff
@@ -54,4 +54,5 @@ if [ "$env" = "gnu" ] && [ "$bits" = "32" ]; then
     RUST_LIBC_UNSTABLE_GNU_FILE_OFFSET_BITS=64 $cmd -- $test_flags
     # shellcheck disable=SC2086
     RUST_LIBC_UNSTABLE_GNU_TIME_BITS=64 $cmd -- $test_flags
+    RUSTFLAGS="$RUSTFLAGS --cfg=libc_unstable_windows_gnu_time64=\"1\"" $cmd -- $test_flags
 fi
```
tgross35, line 58: This block is running on all 32-bit targets, e.g. 32-bit Arm Linux. If there is a special flag for Windows then you'd want to add an `os` var similar to `env` and `bits`, and filter on that. (Of course different if we can just reuse the gnu flags)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5062/8132431a05a94bfce91b56fd1bac74e224d9230a..7050e1728075027181f4f4bc835503d256144ae6)*

src/windows/mod.rs
```diff
@@ -34,6 +34,12 @@ extern_ty! {
     pub enum timezone {}
 }
 
+#[deprecated(
+    since = "1.0.0",
+    note = "This time-related value, among others, is part of the shift \
+            towards using a single, 64-bit-sized `time_t`. See #PENDING for \
+            discussion."
+)]
 pub type time64_t = i64;
```
tgross35: Like the others, let's hold off on deprecating these at least until we stabilize the flags.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5062/8132431a05a94bfce91b56fd1bac74e224d9230a..7050e1728075027181f4f4bc835503d256144ae6)*

tgross35 (changes requested):

tgross35:
Nit on commit style: deprecation is kind of a minor thing. The more substantial change in the commit is to add the feature flag, so that's what the summary and PR description should highlight. Or split deprecation to a separate commit (within the same PR is fine of course).

Also `fix:` doesn't seem quite right since nothing changes out of the box with this PR. I have a mild preference for scope-tagged commits anyway, e.g. ``mingw: Introduce config for 64-bit `time_t` on 32-bit platforms``, since that makes the changelog way easier to generate.

dybucc:
Will edit once I have the changes to the patch ready. Thanks.

dybucc:
@tgross35 The PR has been updated.

There have been changes. The build script is one of them. CI is also included. They use the GNU `cfg`. This was using the pre-#4977 style. I updated it. #4977 should come first.

@rustbot blocked

build.rs
```diff
@@ -24,7 +24,7 @@ const ALLOWED_CFGS: &[&str] = &[
     "freebsd15",
     // Corresponds to `_FILE_OFFSET_BITS=64` in glibc
     "gnu_file_offset_bits64",
-    // Corresponds to `_TIME_BITS=64` in glibc
+    // Corresponds to `_TIME_BITS=64` in glibc. `_USE_32BIT_TIME_T` in Windows.
```
tgross35: It shouldn't, right? That should only be applicable on win-msvc, and on MSVC targets time_t is already always 64 bits. I believe this should only affect i686-win-gnu, which uses the same `_TIME_BITS=64`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5062/69c52f243cf5497e90a59875a4443171c782edfc..7b9905e1156bde742810e37e5e75f3b10743d11d)*

libc-test/build.rs
```diff
@@ -784,6 +784,15 @@ fn test_windows(target: &str) {
...
     cfg.define("_WIN32_WINNT", Some("0x8000"));
 
+    (gnu && i686).then(|| {
+        env::var("CARGO_CFG_LIBC_UNSTABLE_GNU_TIME_BITS")
+            .map_err(|_| ())
+            .and_then(|val| val.parse::<usize>().map_err(|_| ()))
+            .and_then(|v| (v == 64).then_some("gnu_time_bits64").ok_or(()))
+            .map(|op| _ = cfg.cfg(op, None))
+            .unwrap_or_default()
+    });
```
tgross35: There's no need to use `then` when there is no return value, `if` is more clear. But this can be simplified anyway:

```rust
if let Ok(tb) = env::var("CARGO_CFG_LIBC_UNSTABLE_GNU_TIME_BITS") {
    assert!(
        tb == "64" || tb == "32",
        "Invalid value for libc_unstable_gnu_time_bits. Must be 32, 64 or unset"
    );

    if gnu && i686 && tb == "64" {
        cfg.cfg("gnu_time_bits64", None);
    }
}
```

Do we need a `cfg.define` here? The other instances of `CARGO_CFG_LIBC_UNSTABLE_GNU_TIME_BITS` have one.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5062/69c52f243cf5497e90a59875a4443171c782edfc..7b9905e1156bde742810e37e5e75f3b10743d11d)*

build.rs
```diff
@@ -143,8 +143,10 @@ fn main() {
         set_cfg("linux_time_bits64");
     }
 
+    // FIXME(1.0): Windows GNU has 64-bit `time_t` by default. In x86 this is
+    // wrongly bound.
```
tgross35: This FIXME would be better placed by the `time_t` definition than in the build script.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5062/69c52f243cf5497e90a59875a4443171c782edfc..7b9905e1156bde742810e37e5e75f3b10743d11d)*

tgross35 (changes requested):

libc-test/build.rs
dybucc: We don't. Windows with GNU (under MinGW) has the same behavior as MSVC. It exposes a 64-bit `time_t` by default. The feature test macro to use a 32-bit `time_t` is also the same as in MSVC.

Recall this `cfg` is meant as a bugfix to faulty bindings in the `libc` crate. `time_t` is currently 32-bits in `main` when in Windows x86 with GNU. This doesn't match MinGW headers. We just so happen to be "reusing" `gnu_time_bits64`.

`_TIME_BITS` is incompatible with `_USE_32BIT_TIME_T` in MinGW. The former is sort of supported. They check it's not defined when checking for the latter [[1]].

[1]: https://github.com/mingw-w64/mingw-w64/blob/31bd54ab7d5fe03c67ed2bb1a57e531b9c7f8cc4/mingw-w64-headers/crt/_mingw.h.in#L393

build.rs
dybucc: That's not right. My comment is also wrong. Let me clarify.

In both MSVC and MinGW headers `time_t` is always 64-bit. In both of them one can use `_USE_32BIT_TIME_T` to get a 32-bits `time_t`. The `_TIME_BITS` feature test macro is sort of supported on MinGW. Though it's useless. It's only checked before checking that `_USE_32BIT_TIME_T` is defined [[1]].

What we expose is wrong. This PR is more of a bugfix on our side.

[1]: https://github.com/mingw-w64/mingw-w64/blob/31bd54ab7d5fe03c67ed2bb1a57e531b9c7f8cc4/mingw-w64-headers/crt/_mingw.h.in#L393

tgross35: What do you mean by useless? It appears to control what `time_t` is via `_USE_32BIT_TIME_T` https://github.com/mingw-w64/mingw-w64/blob/31bd54ab7d5fe03c67ed2bb1a57e531b9c7f8cc4/mingw-w64-headers/crt/time.h#L52-L59

dybucc: I'm referring to `_TIME_BITS`.

dybucc:
Seems done. We don't check for `time_t` on Windows x86 GNU if `gnu_time_bits64` is not set.

@rustbot ready

dybucc:
I had to add a `cfg` in the `libc-test` build script. The Windows test added on #5059 needs it with the changes in this PR.

I'm not sure if this is the way to go.

libc-test/build.rs
```diff
@@ -784,6 +785,31 @@ fn test_windows(target: &str) {
...
+                true
+            } else {
+                false
+            }
+        } else {
+            false
+        }
+    } else {
+        false
+    };
```
tgross35: Could replace the `false` chaining with:

```rust
let tb_env = env::var("CARGO_CFG_LIBC_UNSTABLE_GNU_TIME_BITS");
let redirect_time_t = match tb_env {
    Ok(tb) if tb == "64" => true,
    Ok(tb) if tb == "32" => false,
    Ok(x) => panic!("Invalid value`{x}` for ... "),
    Err(_) => false,
};

println!("rustc-check-cfg...");
if redirect_time_t {
    println!(...);
    cfg.cfg(...);
}
```

A name like `win32_time64` would be more descriptive than `redirect_time_t`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5062/ed3a6f91d9d64befee093e5725464f784892eed4..50299b1593466e972ceb21c819ea45829fbf760f)*

build.rs
```diff
@@ -24,7 +24,8 @@ const ALLOWED_CFGS: &[&str] = &[
     "freebsd15",
     // Corresponds to `_FILE_OFFSET_BITS=64` in glibc
     "gnu_file_offset_bits64",
-    // Corresponds to `_TIME_BITS=64` in glibc
+    // Corresponds to `_TIME_BITS=64` in glibc. Also used in Windows to fix
+    // faulty `time_t` bindings.
     "gnu_time_bits64",
```
tgross35: Try to be more precise with wording like faulty/wrong/broken. In this case there should be nothing broken in Rust since the link name fix merged, it's just a mismatch to what users get on C. (Applies to the commit message and a few other PRs too).

Also 64-bit time_t wasn't even always an option until last year https://github.com/mingw-w64/mingw-w64/commit/fe21b7ec19b2a765a8e20821543e309ac687117c and uclibc a few years before that - libc's bindings just happen to predate that. (Even glibc and musl support didn't fully exist until ~2020.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5062/ed3a6f91d9d64befee093e5725464f784892eed4..50299b1593466e972ceb21c819ea45829fbf760f)*

libc-test/build.rs
```diff
@@ -841,7 +867,7 @@ fn test_windows(target: &str) {
         "SSIZE_T" if !gnu => true,
         "ssize_t" if !gnu => true,
         // FIXME(windows): The size and alignment of this type are incorrect
-        "time_t" if gnu && i686 => true,
+        "time_t" if gnu && i686 && !redirect_time_t => true,
```
tgross35: I think you may be able to delete the `time_t` exclusion entirely with:

```rust
if i686 && !redirect_time_t {
    cfg.define("_TIME_BITS", Some("32"));
}
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5062/ed3a6f91d9d64befee093e5725464f784892eed4..50299b1593466e972ceb21c819ea45829fbf760f)*

dybucc: Doesn't seem like it. It may be that the MinGW pre-defined macros target the x86_64 host even if the binary targets x86. That would make the conditional in [^1] evaluate false.

[^1]: https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/crt/_mingw.h.in#L393

build.rs
dybucc: Noted.

libc-test/build.rs
dybucc: Done.

build.rs
tgross35: Right - `_TIME_BITS` is used to control `_USE_32BIT_TIME_T` https://github.com/mingw-w64/mingw-w64/blob/31bd54ab7d5fe03c67ed2bb1a57e531b9c7f8cc4/mingw-w64-headers/crt/_mingw.h.in#L393-L397 which controls what `time_t` is.

tgross35 (approved):
LGTM, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5062/ed3a6f91d9d64befee093e5725464f784892eed4..fc36273a7ccca009ec9a9a9be6182eca32972305)*

PR #5065, opened 2026-04-20
Add support for `aarch64-unknown-linux-pauthtest` target

jchlanda:
# Description

This PR adds support for `aarch64-unknown-linux-pauthtest`, a target that
enables Pointer Authentication Code (PAC) support in Rust on AArch64 ELF based
Linux systems using a `pauthtest` ABI (provided by LLVM) and pauthtest-enabled
sysroot with custom [musl](https://github.com/access-softek/musl), serving as a reference libc implementation.

Please consult a rust-lang PR for the details on the target: https://github.com/rust-lang/rust/pull/155722

# Sources

This target reuses the existing musl implementation paths in libc and does not
introduce new platform-specific APIs. `aarch64-unknown-linux-pauthtest` should
be treated as musl-compatible.

# Checklist

- [x] Relevant tests in `libc-test/semver` have been updated - N/A (no semver-affecting API changes)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -47,7 +47,15 @@ const CHECK_CFG_EXTRA: &[(&str, &[&str])] = &[
     ),
     (
         "target_env",
-        &["illumos", "wasi", "aix", "ohos", "nto71_iosock", "nto80"],
+        &[
+            "illumos",
+            "wasi",
+            "aix",
+            "ohos",
+            "pauthtest",
+            "nto71_iosock",
+            "nto80",
+        ],
     ),
     (
         "target_arch",
@@ -112,11 +120,15 @@ fn main() {
     let mut musl_v1_2_3 = env_flag("RUST_LIBC_UNSTABLE_MUSL_V1_2_3");
     println!("cargo:rerun-if-env-changed=RUST_LIBC_UNSTABLE_MUSL_V1_2_3");
 
-    // OpenHarmony uses a fork of the musl libc
-    let musl = target_env == "musl" || target_env == "ohos";
+    // OpenHarmony and pauthtest use a fork of the musl libc
+    let musl = target_env == "musl" || target_env == "ohos" || target_env == "pauthtest";
 
-    // loongarch64, hexagon, and ohos only exist with recent musl
-    if target_arch == "loongarch64" || target_arch == "hexagon" || target_env == "ohos" {
+    // loongarch64, hexagon, ohos and pauthtest only exist with recent musl
+    if target_arch == "loongarch64"
+        || target_arch == "hexagon"
+        || target_env == "ohos"
+        || target_env == "pauthtest"
+    {
         musl_v1_2_3 = true;
     }
 
--- a/build.rs
+++ b/build.rs
@@ -35,6 +35,7 @@ const ALLOWED_CFGS: &[&str] = &[
     // Corresponds to `_REDIR_TIME64` in musl: symbol redirects to __*_time64
     "musl_redir_time64",
     "vxworks_lt_25_09",
+    "libc_pauthtest",
 ];
 
 // Extra values to allow for check-cfg.
@@ -69,6 +70,14 @@ fn main() {
     let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
     let target_ptr_width = env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap_or_default();
     let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap_or_default();
+    let target_abi = env::var("CARGO_CFG_TARGET_ABI").unwrap_or_default();
+
+    // FIXME(msrv): Once the MSRV is 1.78, use `cfg(target_abi = "pauthtest")`
+    // directly instead of translating it to `libc_pauthtest`. `target_abi`
+    // cannot be used directly in cfg expressions on the current MSRV.
+    if target_abi == "pauthtest" {
+        println!("cargo:rustc-cfg=libc_pauthtest");
+    }
 
     // FIXME: this can be removed in 1-2 releases
     println!("cargo:rerun-if-env-changed=RUST_LIBC_UNSTABLE_FREEBSD_VERSION");
@@ -122,8 +131,12 @@ fn main() {
     // OpenHarmony uses a fork of the musl libc
     let musl = target_env == "musl" || target_env == "ohos";
 
-    // loongarch64, hexagon, and ohos only exist with recent musl
-    if target_arch == "loongarch64" || target_arch == "hexagon" || target_env == "ohos" {
+    // loongarch64, hexagon, ohos and pauthtest only exist with recent musl
+    if target_arch == "loongarch64"
+        || target_arch == "hexagon"
+        || target_env == "ohos"
+        || target_abi == "pauthtest"
+    {
         musl_v1_2_3 = true;
     }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,6 +94,7 @@ targets = [
     "aarch64-unknown-freebsd",
     "aarch64-unknown-hermit",
     "aarch64-unknown-illumos",
+    "aarch64-unknown-linux-pauthtest",
     "aarch64-unknown-netbsd",
     "aarch64-unknown-nto-qnx800",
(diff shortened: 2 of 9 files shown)


tgross35:
Based on the todo it seems like you may already be aware, but targets need to be added to rustc first as no-std. After that happens you can bring up libc, then std.

How much does the fork deviate from musl? If it’s identical but built differently, it may be worth considering setting ‘target_env=“musl”` but `target_abi=“pauthtest”` in rustc so things mostly “just work”.

jchlanda:
> Based on the todo it seems like you may already be aware, but targets need to be added to rustc first as no-std. After that happens you can bring up libc, then std.

I knew about [the need](https://rustc-dev-guide.rust-lang.org/building/new-target.html#patching-crates) for patches to `libc` and `cc`. Did not realise about the no-std first. Do you mean that the target should first have [`TargetMetadata.std`](https://github.com/jchlanda/rust/blob/e4537bab00589db0ed39abd33c3bccf4a25b0cc0/compiler/rustc_target/src/spec/mod.rs#L2178) set to `false`?


> How much does the fork deviate from musl? If it’s identical but built differently, it may be worth considering setting ‘target_env=“musl”`but`target_abi=“pauthtest”` in rustc so things mostly “just work”.

The API surface should be identical. However they are most definitely not interchangeable.
This will be more clear once the rust-lang PR lands, but in there I've used `aarch64-unknown-linux-pauthtest`, and (somewhat incorrectly) gated the code on the environment part of the triple.  We have a follow up task to resolve the target to a set of features, just like it's done in clang [where it solves platform/environment specifics on the driver](https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChains/Linux.cpp#L552) level and later on everything is based on [language flags](https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChains/Linux.cpp#L501) that are there regardless of the platform. And finally are resolved to a [concrete signing schema](https://github.com/llvm/llvm-project/blob/main/clang/lib/Frontend/CompilerInvocation.cpp#L1445).

Another deviation is that for the new target statically linked libc is not an option, so perhaps it would be good to keep the distinction clear also at the cfg checks?

tgross35:
Rust PR for reference https://github.com/rust-lang/rust/pulls/jchlanda

tgross35:
> > Based on the todo it seems like you may already be aware, but targets need to be added to rustc first as no-std. After that happens you can bring up libc, then std.
> 
> I new about [the need](https://rustc-dev-guide.rust-lang.org/building/new-target.html#patching-crates) for patches to `libc` and `cc`. Did not realise about the no-std first. Do you mean that the target should first have [`TargetMetadata.std`](https://github.com/jchlanda/rust/blob/e4537bab00589db0ed39abd33c3bccf4a25b0cc0/compiler/rustc_target/src/spec/mod.rs#L2178) set to `false`?

Yes, but I think that flag only controls `-Zbuild-std`. Beyond that, the initial r-l/r PR shouldn't touch `library/std` - after rustc is capable of building something for the target, it's a lot easier to bring std up incrementally. Also the initial merge of compiler support serves as an indicator to the ecosystem that the target is official, so dependencies of std know they can start accepting support patches (e.g. this crate).

> > How much does the fork deviate from musl? If it’s identical but built differently, it may be worth considering setting ‘target_env=“musl”`but`target_abi=“pauthtest”` in rustc so things mostly “just work”.
> 
> The API surface should be identical. However they are most definitely not interchangeable. This will be more clear once the rust-lang PR lands, but in there I've used `aarch64-unknown-linux-pauthtest`, and (somewhat incorrectly) gated the code on the environment part of the triple. We have a follow up task to resolve the target to a set of features, just like it's done in clang [where it solves platform/environment specifics on the driver](https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChains/Linux.cpp#L552) level and later on everything is based on [language flags](https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChains/Linux.cpp#L501) that are there regardless of the platform. And finally are resolved to a [concrete signing schema](https://github.com/llvm/llvm-project/blob/main/clang/lib/Frontend/CompilerInvocation.cpp#L1445).

I don't know the exact details about pauthtest but I still think what I described sounds like a reasonable enough fit. If it looks like musl for ~95% of the API then it's going to be easiest to call it `target_env = "musl"` since `target_env` mostly serves as the library config. In the more rare cases that you need to know the difference, checking `target_abi = "pauthtest"` will work. "Same env, different ABI" also sounds like the right intuition here from my (limited) understanding.

But that's just my thoughts, the others may feel differently. Probably worth a Zulip discussion at some point.

> Another deviation is that for the new target statically linked libc is not an option, so perhaps it would be good to keep the distinction clear also at the cfg checks?

You'd still be able to `cfg(target_abi = "pauthtest")` where needed. And targets that support both can usually toggle `+crt-static`/`-crt-static` so code that needs to know about linkage has to be able to support both to some extent.

On that note, are there other `libc`s that could work for pauthtest eventually aside from the musl fork?

jchlanda:
Moving to `target_env = "musl" target_abi = "pauthtest"` greatly simplified this PR.

Although the parent change ([introduction](https://github.com/rust-lang/rust/pull/155722) of aarch64-unknown-linux-pauthtest) has not landed yet, it seems this should not prevent this PR from being marked as ready for review.

jchlanda:
@davidtwco, @folkertdev, @tgross35, @madsmtm FWI this is a follow up to https://github.com/rust-lang/rust/pull/155722

jchlanda:
Hi @tgross35

The PR [introducing the pauthtest](https://github.com/rust-lang/rust/pull/155722) target has been merged into main. I'd be very grateful if we could move this one forward as well. Please let me know if you'd like anything else done for this PR.

Thank you!

tgross35 (changes requested):
You'll need to update https://github.com/rust-lang/libc/blob/03ff348e8a626069954584b4bfc8f6d09c94841c/build.rs#L41 but otherwise LGTM, this is much cleaner.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5065/03ff348e8a626069954584b4bfc8f6d09c94841c..6eac188b794baf0060c18d28c3b0051bf9baf5c1)*

build.rs
```diff
@@ -69,6 +70,13 @@ fn main() {
     let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
     let target_ptr_width = env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap_or_default();
     let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap_or_default();
+    let target_abi = env::var("CARGO_CFG_TARGET_ABI").unwrap_or_default();
+
+    // `target_abi` is an unstable feature and can not be used directly in `cfg` expressions, for
+    // `pauthtest` translate it to `libc_pauthtest`.
```
tgross35: Add `// FIXME(msrv)` and note this is 1.78:

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5065/03ff348e8a626069954584b4bfc8f6d09c94841c..ad05d09e9e9fe3c6d0b328d8054df7d6edb1864b)*

jchlanda: Done.

tgross35 (approved):
Thanks! Please squash then LGTM

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5065/03ff348e8a626069954584b4bfc8f6d09c94841c..afa243ba47d4d5506141ac2215171619341795ec)*

jchlanda:
Please not, the failure: https://github.com/rust-lang/libc/actions/runs/28580626120/job/84739740809?pr=5065#step:9:2195
doesn't seem to be related to this patch, `c_void_returns` lint triggers in `src/unix/aix/mod.rs`.

folkertdev:
Latest nightly added a lint warning on `-> c_void`, so yeah this is not your fault, just needs a fix here probably.

jchlanda:
> Latest nightly added a lint warning on `-> c_void`, so yeah this is not your fault, just needs a fix here probably.

I've addressed it in here: https://github.com/rust-lang/libc/pull/5240
Will wait for fix PR to go in and rebase this one afterwards.

PR #5068, opened 2026-04-20
Add compat module to provide openpty/forkpty on aix

dalvescb:
# Description

I respectfully would like to request allowing this `compat` module for AIX to be added. There is precedent for this as `solarish` also has a module to export the same missing functions `openpty` and `forkpty`  https://github.com/rust-lang/libc/blob/088a28455a81c6f9cf770cea42b94380d3cc9b05/src/unix/solarish/compat.rs

Without this module, development on AIX would be significantly hindered. `nix` has its own wrapper for `openpty`, but most crates use `libc` directly. Porting the functionality in another crate like `nix` would require having every other crate that uses `libc` directly to switch to the `nix` API, a big ask to simply support a niche OS like AIX. This patch would only affect AIX, and I would be happy to take full responsibility maintaining it 

# Sources

# Checklist

- [ X] Relevant tests in `libc-test/semver` have been updated
- [ X] No placeholder or unstable values like `*LAST` or `*MAX` are
- [X ] Tested locally (`cd libc-test && cargo test --target mytarget`);

### DIFF
--- a/src/unix/aix/compat.rs
+++ b/src/unix/aix/compat.rs
@@ -1,0 +1,198 @@
+// Common functions that are unfortunately missing on AIX
+
+use crate::unix::aix::*;
+use crate::{c_char, c_int};
+
+#[cfg(target_os = "aix")]
+unsafe fn bail(fdm: c_int, fds: c_int) -> c_int {
+    let e = *_Errno();
+    if fds >= 0 {
+        crate::close(fds);
+    }
+    if fdm >= 0 {
+        crate::close(fdm);
+    }
+    *_Errno() = e;
+    -1
+}
+
+#[cfg(target_os = "aix")]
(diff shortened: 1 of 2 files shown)


tgross35:
Thank you for the interest but we don’t want to add more polyfill in this crate. The Solarish versions are planned to be removed when possible https://github.com/rust-lang/libc/issues/4045.

Using one of the wrapper crates like nix or rustix is the correct thing to do if you need functionality that isn’t available on some platforms, so users should already be expecting this from other cases. I believe std also has a polyfill for this as well.

Alternatively, the AIX maintainers have been helpful, cc @xingxue-ibm in case this is a planned feature :)

JohnTitor:
Triage: Closing in favor of the above comment. Thank you for the PR anyway!

PR #5069, opened 2026-04-21
POSIX: change struct stat/stat64 times to timespec

### LINKED ISSUE #4965: What time fields should `struct stat` have?
Prior discussion at https://github.com/rust-lang/libc/pull/4942

@xbjfk said:

> I would personally push for using `timespec` to:
> 
>     * Closer match the libc source
>     * Be POSIX correct and match the correct C usage / docs / man pages.
>     * Have more compatibility with other architectures (hermit, AIX and hurd are already)
>     * Make struct definitions easier to read (especially with bi-endian architectures)
>     * We can remove a lot of test skips.
> 
> I personally think the best path forward would be to revert [55fa65b](https://github.com/rust-lang/libc/commit/55fa65b3e026594c6418eb2bc8e98c1feb7667da) so that people can try the new experimental musl with less breaking changes (and fixing loongarch) and reintroduce the name change without a feature gate for all platforms in the 1.0 branch only for the reasons listed above (virtually every UNIX platform supported by libc uses timespec).
> 
> IIRC this was changed with time64 changes the because having conditional definitions for the padding was a bit unwieldy.
> 
> Will look into making a PR for this.

xbjfk:
# Why?
The primary reason for this change is that this is the correct [POSIX specification](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/sys_stat.h.html), and using _nsec etc. may be considered deprecated - meaning this better follows the C struct definition more closely as  well as C man pages / docs. Additionally, this makes the struct easier to read, and removes the need for many test skips, and brings all established platforms in parity with Hurd etc. which already use the new declaration in this crate. More discussion in rust-lang/libc#4965

# Description

Since this was evidently a problem when doing musl changes and v1 being in alpha, it's time to rip the band aid off and do it.

Of course, there are a couple unsolved questions:
 - Windows doesn't have the _nsec fields: https://github.com/rust-lang/libc/blob/17adf2d8003fd5021358905abdfeed68d504070f/src/windows/mod.rs#L52-L54 There is also a note that should probably be addressed before 1.0: https://github.com/rust-lang/libc/blob/17adf2d8003fd5021358905abdfeed68d504070f/src/windows/mod.rs#L43
 - Newlib has padding where the nsecs should be: https://github.com/rust-lang/libc/blob/0dddfc8c344fb622b8cd78f7f24b4d1ace302e51/src/unix/newlib/generic.rs#L22-L27 But upstream uses timespec for some architectures: https://github.com/cygwin/cygwin/blob/a0ccec7e7a16b01037ece1947610f768ab7f081a/newlib/libc/include/sys/stat.h#L37-L59
 - I don't know what is going on with Apple - they use `xtimespec` instead of `xtim`: https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/stat.h#L202-L206 - but interestingly, Swift seems to rename this to the POSIX correct form, despite the underlying C headers not having it: https://github.com/apple/swift-system/blob/main/Sources/System/FileSystem/Stat.swift#L428. We need to pick one.

Also, solid and qurt do not have nsec at all. NTO and VxWorks have already migrated. 

# Sources

In each relevant commit.

# Checklist

As discussed in / Fixes: rust-lang/libc#4965


xbjfk:
Hmm, since freebsd11 no longer depends on c_long, the prelude is no longer necessary - but it feels wrong to remove.

tgross35 (commented):
All of this looks pretty good to me, but I'm going to hold off on merging for a bit. We have a handful of changes coming down the pipeline that are doing 64-bit time conversions, which this will conflict with, and it will be easier to apply this after all of that rather than before. (Not on its own, but taking stable backports into consideration).

Would you mind adding the reasoning (posix spec alignment) to the PR description? For any drive by readers.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5069/545663c3553890efe170b273ecd73b6f3dfb9f75..1ae80a9d09aa2f9f7cc6eb9615eadfeb989296de)*

tgross35:
For some time to get time64 changes in,
@SnoozeThis 2 months

SnoozeThis:
I don't understand what you want to wait for (unknown domain 2). See https://snoozethis.com/commands/ for examples.

tgross35:
@SnoozeThis wait 2 months

tgross35:
Come on alarm clock...
@rustbot blocked

SnoozeThis:
(https://snoozeth.is/WTvxKr50tBM) I will wait until Wed, 19 Aug 2026 07:45:28 UTC and then add label S-waiting-on-review and remove label S-blocked.

@rustbot claim.

xbjfk:
> Would you mind adding the reasoning (posix spec alignment) to the PR description? For any drive by readers.

Done - I made a nice and prominent "Why?" section at the top :)

maurer (approved):
Looks OK for Android.  https://cs.android.com/android/platform/superproject/+/android-latest-release:bionic/libc/include/sys/types.h;l=97-99?q=include%2Fsys%2Ftypes.h had me worried, but that maps it to long in the end, so we're fine.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5069/545663c3553890efe170b273ecd73b6f3dfb9f75..1ae80a9d09aa2f9f7cc6eb9615eadfeb989296de)*

SnoozeThis:
Resolved.

@rustbot label +S-waiting-on-review -S-blocked. @rustbot release-assignment.

tgross35 (changes requested):
Would you be up for merging these definitions in a commit prior to the timespec change? Similar to what fc8a345915b2594fbc1e5c8d1cf28648f3bbe852, 72f013506e8e58e4e4f8d3c506c2028037db6543, and others did, that would reduce the surface area quite a bit.

I've been meaning to do this before replying on this PR but still haven't yet, so figured I'd at least let you know that's what it's blocked on in case you have the time to pick it up.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5069/545663c3553890efe170b273ecd73b6f3dfb9f75..1ae80a9d09aa2f9f7cc6eb9615eadfeb989296de)*

xbjfk:
Sure, I'll give it a try. I've been looking at doing that for some of the musl stuff already.

Issue #5063, opened 2026-04-17
GitHub Actions: SHA-pin external actions?


JohnTitor:
I'd like to SHA-pin third-party actions to reduce security concerns, but I just want to make sure it's unpinned intentionally or not.

So, considering recent supply-chain attacks like tj-actions/reviewdog, it's important to sha-pin reference to third-party actions.
As priot art, other projects on rust-lang do it (with renovate enabled), like crates.io: https://github.com/rust-lang/crates.io/blob/main/.github/workflows/ci.yml

One minus point is that it becomes harder to maintain by hands. Though it can be resolved by using Renovate or something else. I'm a bit concerned it'll add some burdens for us, but it'd be acceptable I guess (we can rate-limit updates to reduce maintenance needs, for example).
But yeah, I think this is a discussion point hence I opened this issue.

@tgross35 any thoughts on this?

tgross35:
I think it's a good idea to run Zizmor like we've started adding to other repos, and address things that is shows. For compiler-builtins we went through this series to get scanning and renovate set up, something similar is probably reasonable here:

* https://github.com/rust-lang/compiler-builtins/commit/3d14fe4e9cfa28c7a6426eb803f141633871c2fd
* https://github.com/rust-lang/compiler-builtins/commit/e2aaeebb513093bb37bfb130026e884c0c31c785
* https://github.com/rust-lang/compiler-builtins/commit/81033cae94755d6c4382237a19622a181c6333c4

JohnTitor:
I'll try out that 👍

PR #5078, opened 2026-04-25
chore: SHA-pin actions

JohnTitor:
# Description

cc https://github.com/rust-lang/libc/issues/5063

This SHA-pins all the actions usage to introduce zizmor.

# Sources

# Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,4 +1,5 @@
 name: CI
+permissions: {}
 
 on:
   merge_group:
@@ -30,7 +31,9 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
       - name: Setup Rust toolchain
         run: ./ci/install-rust.sh && rustup component add rustfmt
       - name: Check style
@@ -44,9 +47,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
       - run: rustup update stable --no-self-update
-      - uses: Swatinem/rust-cache@v2
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
       # Here we use the latest stable Rust toolchain already installed by GitHub
       # Ideally we should run it for every target, but we cannot rely on unstable toolchains
       # due to Clippy not being consistent between them.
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -30,7 +31,9 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
xtqqczze, line 34:
```suggestion
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```

etc

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5078/545663c3553890efe170b273ecd73b6f3dfb9f75..c91f7dd552c5dfb36d67a0ed0916e2faed212e3e)*

xtqqczze (changes requested):

.github/workflows/ci.yaml
JohnTitor: Hm, why? It's less secure to depend on mutable tags. Also full semver is clearer to understand the contents of releases.

xtqqczze: It keeps the existing intent, which isn't to pin to v6.0.2 specifically.

JohnTitor: The goal here isn't to keep things same but harden security. That's part of it.

tgross35: We'll have dependabot updates working before landing this so we can keep the pins but still have automatic minor bumps. Zizamor would complain about a `v6` tag, anyway.

tgross35 (commented):
When this is closer to ready, could you split to one commit fixing things and one adding zizmor to CI? Thinking about ability to backport

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5078/545663c3553890efe170b273ecd73b6f3dfb9f75..c91f7dd552c5dfb36d67a0ed0916e2faed212e3e)*

.github/workflows/ci.yaml
xtqqczze: > Zizamor would complain about a `v6` tag, anyway.

Actually, zizmor doesn’t care about the comment. I was mainly concerned that using the full version tag there might affect Dependabot updates, but it seems Dependabot updates the comment correctly either way.

JohnTitor: Yeah, Dependabot checks Git tags and both are so. The difference is that `vN` is mutable and `vX.Y.Z` is generally not.

JohnTitor:
@tgross35 
Good point, addressed.

.github/workflows/ci.yaml
xtqqczze: Would be nice if zizmor validated commits match the tag in a comment.

tgross35 (approved):
LGTM, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5078/c12ee660bc874b73d24448e7d2f21259e91c0666..d886c9b02d8101cb779c44a8e80d9a77440bb37f)*

.github/workflows/ci.yaml
tgross35: I'm rather late here but it does, though only with `--pedantic` https://docs.zizmor.sh/audits/#ref-version-mismatch

xtqqczze: I just tried this with zizmor 1.25.2, including `--pedantic` and none of these triggered [ref-version-mismatch](https://docs.zizmor.sh/audits/#ref-version-mismatch) for me:

```yml
- uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v1
- uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v1.0.0
- uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v99.0.0
```

tgross35: Did you provide a GH token? It's kind of hard to find in the docs but anything that requires online checking also needs `--gh-token $(gh auth token)`.

Haven't tested this recently but IIRC that's what I ran into.

PR #5080, opened 2026-04-25
cargo: use workspace-level lints

xtqqczze:
Closes https://github.com/rust-lang/libc/pull/4406

### DIFF
--- a/ctest/src/ffi_items.rs
+++ b/ctest/src/ffi_items.rs
@@ -241,8 +241,7 @@ impl<'ast> Visit<'ast> for FfiItems {
             .abi
             .name
             .clone()
-            .map(|s| Abi::from(s.value().as_str()))
-            .unwrap_or_else(|| Abi::C);
+            .map_or(Abi::C, |s| Abi::from(s.value().as_str()));
 
         for item in &i.items {
             match item {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -175,14 +175,11 @@ members = [
     "libc-test",
 ]
 
-# FIXME(msrv): These should be renamed as `[workspace.lints.*]` once MSRV is above 1.64
-#       This way all crates can use it with `[lints] workspace=true` section
-
-[lints.rust]
+[workspace.lints.rust]
 # FIXME(cleanup): make ident usage consistent in each file
 unused_qualifications = "allow"
 
-[lints.clippy]
+[workspace.lints.clippy]
 # Enable pedantic lints - use this manually once in a while, but don't enable by default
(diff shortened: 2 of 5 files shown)


tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5080/545663c3553890efe170b273ecd73b6f3dfb9f75..c1c93ddb3bbdef211637dcd507f1ea91668b816a)*

ctest/src/ffi_items.rs
```diff
@@ -241,8 +241,7 @@ impl<'ast> Visit<'ast> for FfiItems {
             .abi
             .name
             .clone()
-            .map(|s| Abi::from(s.value().as_str()))
-            .unwrap_or_else(|| Abi::C);
+            .map_or_else(|| Abi::C, |s| Abi::from(s.value().as_str()));
```
tgross35: Actually sorry can you change this to `.map(|s| Abi::from(s.value().as_str())).unwrap_or(Abi::C)` ? Not sure why it's using `or_else` here, constant value.

Feel free to delete the `map_unwrap_or` lint from the table too, imo that lint is always questionable and I'm not sure why it's enabled now.

tgross35 (changes requested):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5080/545663c3553890efe170b273ecd73b6f3dfb9f75..c1c93ddb3bbdef211637dcd507f1ea91668b816a)*

ctest/src/ffi_items.rs
xtqqczze: I've pushed changes to remove the closure, what is the issue with the `map_unwrap_or` lint specifically?

tgross35: It has nonobvious ordering since the name is `map_or_else` but the parameters are `(or_else, map)`. Two closure function calls also tend to be harder to read. Iirc there's a clippy issue to not suggest `map_or_else` if it would use two closures.

Some more at [#t-lang > broken and un-fixable parts of Rust @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/213817-t-lang/topic/broken.20and.20un-fixable.20parts.20of.20Rust/near/591001454)

tgross35 (approved):

ctest/src/ffi_items.rs
xtqqczze: There is https://github.com/rust-lang/rust-clippy/issues/15829 for this exact case.

PR #5081, opened 2026-04-25
linux: add SO_DETACH_REUSEPORT_BPF

swananan:
# Description
Add `SO_DETACH_REUSEPORT_BPF` for Linux, following the same approach as rust-lang/libc#3917.

This exposes the constant on the Linux arch modules that define socket option values directly (`generic`, `mips`, `powerpc`, and `sparc`) and updates the Linux semver list.

`SO_DETACH_REUSEPORT_BPF` is already defined by musl starting in v1.2.0, but libc CI still runs old-musl jobs as well, so the libc-test constant check is still skipped for musl.

# Sources
- Linux UAPI: https://github.com/torvalds/linux/blob/897d54018cc9aa97fd1529ca08a53b429d05a566/include/uapi/asm-generic/socket.h#L117-L123
- musl v1.2.0: https://github.com/ifduyue/musl/blob/v1.2.0/include/sys/socket.h#L1572-L1576

# Checklist


swananan:
Looks like CI is flaky, will retry this tomorrow.

JohnTitor (approved):
Looks good, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5081/545663c3553890efe170b273ecd73b6f3dfb9f75..93b26c00bf9db09bd4fb633cd6ad655c701004be)*

PR #5090, opened 2026-04-29
fix : incorrect POLLOUT def on horizonOS/armv6k-nintendo-3ds

kore-signet:
# Description

Current definition for POLLOUT is incorrect - as described [here](https://www.3dbrew.org/wiki/SOCU:poll), it should be 0x08.

Unfortunately, it seems like the current definition for POLLERR is overlapping as 0x08. According to the linked docs, POLLERR is actually not a supported operation on this target (neither is POLLHUP, which is also currently defined). 

I'm not super clear on what the best procedure would be here - should POLLERR & POLLHUP be deleted as definitions? 

# Sources

- https://www.3dbrew.org/wiki/SOCU:poll

# Checklist

(libc-test does not currently support horizonOS, as far as I can tell?)

### DIFF
--- a/src/unix/newlib/horizon/mod.rs
+++ b/src/unix/newlib/horizon/mod.rs
@@ -134,12 +134,14 @@ pub const WNOHANG: c_int = 1;
 
 pub const POLLIN: c_int = 0x0001;
 pub const POLLPRI: c_int = 0x0002;
-pub const POLLOUT: c_int = 0x0004;
+pub const POLLOUT: c_int = 0x0008;
 pub const POLLRDNORM: c_int = 0x0040;
 pub const POLLWRNORM: c_int = POLLOUT;
 pub const POLLRDBAND: c_int = 0x0080;
 pub const POLLWRBAND: c_int = 0x0100;
+/// POLLERR behavior on 3DS+HorizonOS is unclear, and it may be unsupported.
 pub const POLLERR: c_int = 0x0008;
+/// POLLHUP behavior on 3DS+HorizonOS is unclear, and it may be unsupported.
 pub const POLLHUP: c_int = 0x0010;
 pub const POLLNVAL: c_int = 0x0020;
 


tgross35:
> I'm not super clear on what the best procedure would be here - should POLLERR & POLLHUP be deleted as definitions?

Are they defined when they try to use them in C? If so, we should match that. It's fine to remove if they aren't in the platform's libc since this is a T3 target.

The existing part of this PR lgtm.

Pinging 3DS target maintainers: @Meziu @AzureMarker @ian-h-chamberlain

kore-signet:
> Are they defined when they try to use them in C? If so, we should match that. It's fine to remove if they aren't in the platform's libc since this is a T3 target.

The closest thing HorizonOS has to a libc is libctru, which has the following def for POLLERR & POLLHUP:
```c
#define POLLHUP		0x04 // unknown ???
#define POLLERR		0x08 // probably
```

So yes, it is technically defined in C, but not very strongly, haha. I'll defer to the platform maintainers :)

tgross35:
It actually has those "unknown" and "probably" comments? That's entertaining :) (I assume these are the results of reverse engineering)

Since they do exist there I'd just leave them in our `libc`. Feel free to add a terse doc comment indicating they may not be well supported. This PR looks fine to me either way, feel free to unmark it as a draft when you're ready.

kore-signet:
It really does have the comments, haha! I'll add the comment marking it out and open the PR for a merge tmrw :)

Meziu:
Sorry for not responding to the pings earlier, these changes should be fine and matching libctru is the correct approach. Knowing that library I'd say those are most likely just placeholder costants.

kore-signet:
Got it! Added a doc comen, and now it's ready :)

kore-signet:
Accidentally reverted my own fix, lmao. Fixed it in the latest commit, should be all good :)

tgross35 (approved):
Please squash but lgtm

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5090/a8e295810877b50961f1217330108a8c37505fe0..f15b18af03a80aaba602bda145344dce2aaca032)*

tgross35:
Thanks!

kore-signet:
@tgross35 looks like it got pushed out of merge queue by a download error, could we get it re-added?

PR #5098, opened 2026-05-07
cygwin: fix cpuset_t typo in CPU_ZERO

phdye:
# Description

`CPU_ZERO` in `src/unix/cygwin/mod.rs:1712` declares its parameter as `&mut cpuset_t`, but no type by that name exists in the cygwin module. The struct is `cpu_set_t` (defined at line 91), and every sibling function (`CPU_SET`, `CPU_CLR`, `CPU_ISSET`, `CPU_COUNT`, `CPU_COUNT_S`, `CPU_EQUAL`) correctly uses `cpu_set_t`.

The typo prevents the cygwin module from compiling on `x86_64-pc-cygwin`:

```
error[E0425]: cannot find type `cpuset_t` in this scope
    --> src/unix/cygwin/mod.rs:1712:34
     |
1712 |     pub fn CPU_ZERO(cpuset: &mut cpuset_t) -> () {
     |                                  ^^^^^^^^
```

# Sources

Internal — typo in this crate. The `cpu_set_t` struct is the only definition; no upstream header reference needed.

# Checklist

- [ ] Relevant tests in `libc-test/semver` have been updated — N/A; not adding/removing symbols.
- [x] Tested locally — with this fix applied on `main`, `cargo test --target x86_64-pc-cygwin` builds and passes the full suite (2968 ctest cases, all other suites green).

### DIFF
--- a/src/unix/cygwin/mod.rs
+++ b/src/unix/cygwin/mod.rs
@@ -1709,7 +1709,7 @@ f! {
         s as c_int
     }
 
-    pub fn CPU_ZERO(cpuset: &mut cpuset_t) -> () {
+    pub fn CPU_ZERO(cpuset: &mut cpu_set_t) -> () {
         cpuset.bits.fill(0);
     }
 


phdye:
Companion PR for the same Cygwin port: #5099 (off-by-one in `utsname.sysname` length). The two are independent but `libc-test` on `x86_64-pc-cygwin` only runs end-to-end with both fixes applied — this PR is the prerequisite that lets the cygwin module compile, and #5099 then passes the suite.

JohnTitor (approved):

PR #5110, opened 2026-05-19
Feat(apple): Add the necessary for the `bsdshortinfo` API of libproc

PaulDance:
# Description

This adds the item necessary in order to get the "BSD short info" of processes by interacting with `proc_pidinfo`. Namely: `PROC_PIDT_SHORTBSDINFO` and `proc_bsdshortinfo`.

# Sources

 * [`PROC_PIDT_SHORTBSDINFO`](https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/proc_info.h#L760);
 * [`proc_bsdshortinfo`](https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/proc_info.h#L85);

# Checklist

  especially relevant for platforms that may not be checked in CI.
  The concerned target is in tier 1 and I couldn't easily make my cross-compilation setup work for this.

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -603,6 +603,35 @@ s! {
         pub pbi_start_tvusec: u64,
     }
 
+    pub struct proc_bsdshortinfo {
+        /// Process ID.
+        pub pbsi_pid: u32,
+        /// Process parent ID.
+        pub pbsi_ppid: u32,
+        /// Process perp ID.
+        pub pbsi_pgid: u32,
+        /// `p_stat` value: `SZOMB`, `SRUN`, etc.
+        pub pbsi_status: u32,
+        /// Up to 16 characters of process name.
+        pub pbsi_comm: [c_char; MAXCOMLEN],
+        /// 64bit, emulated, etc.
+        pub pbsi_flags: u32,
+        /// Current UID on process.
+        pub pbsi_uid: crate::uid_t,
+        /// Current GID on process.
+        pub pbsi_gid: crate::gid_t,
+        /// Current RUID on process.
+        pub pbsi_ruid: crate::uid_t,
+        /// Current RGID on process.
+        pub pbsi_rgid: crate::gid_t,
+        /// Current SVUID on process.
+        pub pbsi_svuid: crate::uid_t,
+        /// Current SVGID on process.
+        pub pbsi_svgid: crate::gid_t,
+        /// Reserved for future use.
+        pub pbsi_rfu: u32,
+    }
+
     pub struct proc_taskallinfo {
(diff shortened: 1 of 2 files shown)


src/unix/bsd/apple/mod.rs
```diff
@@ -603,6 +603,35 @@ s! {
         pub pbi_start_tvusec: u64,
     }
 
+    pub struct proc_bsdshortinfo {
+        /// Process ID.
+        pbsi_pid: u32,
```
JohnTitor: It'd be better to make filelds `pub` for use.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5110/c5c1a5a451e949f0ec3890039877fd2cbd7e4ddf..783bf9edfc242770b38ea647e4904b6f3ee28f0a)*

JohnTitor:
Could you also squash commits into one?

src/unix/bsd/apple/mod.rs
PaulDance: Oof, yikes. I adapted this from internal code and completely forgot about that :sweat_smile:

Thanks, done.

PaulDance:
Done!

JohnTitor:
Thanks!
Seems we've hit an Actions outage, I'll get back tomorrow.

JohnTitor (approved):

PaulDance:
Thanks!

src/unix/bsd/apple/mod.rs
```diff
@@ -603,6 +603,35 @@ s! {
...
+        /// Current RUID on process.
+        pub pbsi_ruid: crate::uid_t,
+        /// Current RGID on process.
+        pub pbsi_rgid: crate::gid_t,
+        /// Current SVUID on process.
+        pub pbsi_svuid: crate::uid_t,
+        /// Current SVGID on process.
+        pub pbsi_svgid: crate::gid_t,
+        /// Reserved for future use.
+        pub pbsi_rfu: u32,
```
tgross35, line 632: We've been making reserved fields nonpublic, which also has the desirable effect of making the struct non-exhaustive (since unfortunately `#[non_exhaustive]` can't be used, https://github.com/rust-lang/rust/issues/132699). Changing this in https://github.com/rust-lang/libc/pull/5174

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5110/59e4578dae5a78f116b8448da829d9fd249e64e1..ef196918cffa40d46d3273f48a624936b8ff15e6)*

PR #5118, opened 2026-05-27
bsd: Annotate `ELAST` constants as unstable

dybucc:
# Description

This set of constants had already been discussed to cause some issues in rust-lang/libc#3131. This patch deprecates such bug-prone and latest-error symbols such that the interface to the `libc` crate is not prone to SemVer-breakage as often as the values change upstream.

Note the NetBSD constant was left unmodified as it already had been annotated with a deprecation attribute.

If you yourself want to help out in the deprecation effort for this type of constants across the `libc` crate, maybe try out the [`libc-constant-deprecator`][deprecator] crate. It's been the tool used to perform the changes that have been submitted in this PR, and it very much welcomes bug reports.

[deprecator]: https://github.com/dybucc/libc-constant-deprecator

## Sources

- Upstream source code for the FreeBSD `ELAST` macro declaration, mentioning its use as a "limit" value for other `errno` values.

  > <https://github.com/freebsd/freebsd-src/blob/bd15d6ef126ee4c0eac931117f6bbbf6f9a3fc72/sys/sys/errno.h#L184>
- Upstream source code for the OpenBSD equivalent macro declaration.

  > <https://github.com/openbsd/src/blob/58ff4898d99200dea776fa757004330318255d45/sys/sys/errno.h#L175>
- Upstream source code for the XNU kernel equivalent macro declaration.

  > <https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/errno.h#L270>
- Upstream source code for the DragonFlyBSD equivalent macro declaration.

  > <https://github.com/DragonFlyBSD/DragonFlyBSD/blob/7b7fb99bddde8ded0bee8f43df4026ca02be0938/sys/sys/errno.h#L206>

## Checklist

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -2143,6 +2143,9 @@ pub const ENOTRECOVERABLE: c_int = 104;
 pub const EOWNERDEAD: c_int = 105;
 pub const EQFULL: c_int = 106;
 pub const ENOTCAPABLE: c_int = 107;
+
+/// This symbols is prone to change across releases upstream.
+/// See the [usage guidelines](crate::#usage-guidelines) for details and use.
 pub const ELAST: c_int = 107;
 
 pub const EAI_AGAIN: c_int = 2;
--- a/src/unix/bsd/freebsdlike/dragonfly/mod.rs
+++ b/src/unix/bsd/freebsdlike/dragonfly/mod.rs
@@ -548,7 +548,11 @@ pub const ENOMEDIUM: c_int = 93;
 pub const ENOTRECOVERABLE: c_int = 94;
 pub const EOWNERDEAD: c_int = 95;
 pub const EASYNC: c_int = 99;
+
+/// This symbols is prone to change across releases upstream.
+/// See the [usage guidelines](crate::#usage-guidelines) for details and use.
 pub const ELAST: c_int = 99;
+
 pub const RLIMIT_POSIXLOCKS: c_int = 11;
 #[deprecated(since = "0.2.64", note = "Not stable across OS versions")]
 pub const RLIM_NLIMITS: crate::rlim_t = 12;
--- a/src/unix/bsd/freebsdlike/freebsd/freebsd11/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/freebsd11/mod.rs
@@ -258,7 +258,10 @@ s! {
     }
 }
 
+/// This symbols is prone to change across releases upstream.
+/// See the [usage guidelines](crate::#usage-guidelines) for details and use.
 pub const ELAST: c_int = 96;
+
 pub const RAND_MAX: c_int = 0x7fff_fffd;
 pub const KI_NSPARE_PTR: usize = 6;
 pub const MINCORE_SUPER: c_int = 0x20;
--- a/src/unix/bsd/freebsdlike/freebsd/freebsd12/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/freebsd12/mod.rs
@@ -304,6 +304,9 @@ s! {
 }
 
 pub const RAND_MAX: c_int = 0x7fff_fffd;
+
+/// This symbols is prone to change across releases upstream.
+/// See the [usage guidelines](crate::#usage-guidelines) for details and use.
 pub const ELAST: c_int = 97;
 
 /// max length of devicename
--- a/src/unix/bsd/freebsdlike/freebsd/freebsd13/mod.rs
+++ b/src/unix/bsd/freebsdlike/freebsd/freebsd13/mod.rs
@@ -317,6 +317,9 @@ s! {
 }
 
 pub const RAND_MAX: c_int = 0x7fff_ffff;
+
+/// This symbols is prone to change across releases upstream.
+/// See the [usage guidelines](crate::#usage-guidelines) for details and use.
 pub const ELAST: c_int = 97;
 
 pub const KF_TYPE_EVENTFD: c_int = 13;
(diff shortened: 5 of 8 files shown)


JohnTitor (changes requested):
I think you should deprecate them for 0.2 and remove them for 1.0. Same for similar your PRs.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5118/2364caf83be0f55ead8bd34dfb82cce66a17c387..cd01c4cb490ec37613fd8f262f20321169ab8fad)*

dybucc:
> I think you should deprecate them for 0.2 and remove them for 1.0. Same for similar your PRs.
>
> [View changes since this review][changes]

[changes]: https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5118/2364caf83be0f55ead8bd34dfb82cce66a17c387..cd01c4cb490ec37613fd8f262f20321169ab8fad_

I don't quite understand what you mean. You mention deprecating for the 0.2 release, and removal for
the 1.0 release, but literally doing that means opening up separate PRs where I target these same
changes for the `libc-0.2` branch, and removal (alongside modification of the corresponding `*.txt`
SemVer files) for the `main` branch. Just so I don't get things wrong, is that what you're implying?

Also, I left some comments, especially in #5122, that I would like to get external feedback on, if
you don't mind.

@rustbot ready

JohnTitor:
Adding a deprecation attr for 1.0 doesn't make sense at all. You need to deprecate them for 0.2, see https://github.com/rust-lang/libc/pull/4343 as an example.

@rustbot author

dybucc:
> Adding a deprecation attr for 1.0 doesn't make sense at all. You need to deprecate them for 0.2,
> see #4343 as an example.
> 
> @rustbot author

Thanks. Done.

Should I also remove them from the SemVer plain text files?

@rustbot ready

JohnTitor:
> Should I also remove them from the SemVer plain text files?

No in this PR, let's do that in a PR actually removing them.

JohnTitor (approved):
LGTM, but this would make a lot of warnings in the world so cc @tgross35 to be seconded if we're ready to go ahead or not.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5118/dbd7ad9a0ef5205226d8541f2a1a7cbdd9da2989..8ec90b6893ea92ab085b27f02e9eadbd002bb0e5)*

dybucc:

tgross35:
I've been mulling this over a lot the past couple of weeks and have been tending to agree with Amanieu's comment in https://github.com/rust-lang/libc/issues/3131#issuecomment-2000239138. That is, I don't think deprecating to remove is the right path here: there are completely valid cases for these constants, e.g. allocating an array that can then be indexed by all possible values of an enum or looping through possible values of a flag. (I've actually needed both of these in the past couple of weeks.) 

Would it be possible to find+replace the deprecation notices with a doc comment indicating that these constant values are subject to change? It can just be a terse comment linking to usage guidelines, which I'm adding in https://github.com/rust-lang/libc/pull/5179 to help address this case.

@JohnTitor I'd like your thoughts here of course, given your concern in https://github.com/rust-lang/libc/pull/3040#issuecomment-1453389931. My thinking is that the failure modes here likely aren't any worse than any other constant like unsupported syscalls or flags - `ELAST` is just a pointer to some specific errno, you could use the errno regardless of whether or not `ELAST` tells you which one it is.

---

> If you yourself want to help out in the deprecation effort for this type of constants across the `libc` crate, maybe try out the [`libc-constant-deprecator`](https://github.com/dybucc/libc-constant-deprecator) crate. It's been the tool used to perform the changes that have been submitted in this PR, and it very much welcomes bug reports.

That is quite a cool tool! You may want to rename/reframe it as something like "API reviewer" because I assume it is useful beyond `libc` and could pretty easily do more than deprecate. E.g. add doc comments :) or add `#[non_exhaustive]` to structs that are already `non_exhaustive` via a field (which we will eventually want once the rustc bug is fixed).

dybucc:
Done. From reading what may become the new usage guidelines in #5179, I decided to instead add a new macro for declaring these types of constants that automatically adds the doc comment. I've opened another PR that solely defines the macro.

Should I close the analogous PRs that remove the symbols?

src/macros.rs
```diff
@@ -402,6 +402,21 @@ macro_rules! offset_of {
...
+/// Defines constants with an accompanying doc comment pointing out their
+/// instability, and links to the crate documentation for usage guidelines.
+macro_rules! ct {
+    ($($it:item)+) => {
+$(
+/// This constant, among others often used in C for the purposes of denoting the
+/// latest value or limit in a set of constants, is likely to change upstream.
+/// For correct usage, see the [crate-level documentation][docs].
+///
+/// [docs]: index.html#usage-recommendations
```
tgross35: If we had https://github.com/rust-lang/rust/issues/143549 then that would be one thing, but I'd like to avoid more functionlike macros then necessary because they tend to make things pretty messy. I think the comment can be made to fit on 1-2 lines so it's easier to copy around to all relevant values, something like:

```rust
/// This is an end marker constant, its value is likely to change across releases. See
/// [Best Practices](crate::#best-practices) for details.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5118/1a9bdc4d16edc21dfa66e655d1edf8e4eac16ba8..e70bebb0c6857705aebe2a8275ac3b5706abe3ca)*

tgross35:
> Should I close the analogous PRs that remove the symbols?

I think so. My concern is that removing them will break valid code where the only alternatives are going to be strictly worse.

src/macros.rs
dybucc: Done on this PR. Will be updating the rest soon.

tgross35 (approved):
LGTM, thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5118/cef4fa4f03fce3c3f6933e1463bfcd4439a20a2c..2415e63e17f1b383d344e7487b8ceac49570906a)*

PR #5120, opened 2026-05-28
Annotate unstable `*LAST` constants

dybucc:
# Description

This PR adds documentation. Some symbols are troublesome. They can change upstream. The Rust change is non-breaking. This requires annotating. It requires usage advice. This patch links to it.

## Notes

QNX is untouched. Sources were not found. Their SDK sign up is broken.

Solaris is mostly untouched. Sources were not found. Changes include generic symbols. These were sourced from Illumos.

## Sources

- FreeBSD manpage detailing how `UTXDB_LASTLOGIN` is *not* the type of constant we would want to deprecate.

  > <https://man.freebsd.org/cgi/man.cgi?query=getutxent&sektion=3>
- DragonFlyBSD sources showing how they define a `UTX_DB_LASTLOGX` symbol but not a `UTX_DB_LASTLOG` symbol as we do.

  > <https://github.com/DragonFlyBSD/DragonFlyBSD/blame/59b78e44f96799c5f46fb3d6d616c18d838a75d0/include/utmpx.h#L89>
- Android sources showing how they use the `NF_*_PRI_LAST` in much the same manner as the Linux sources (further below.)

  > <https://cs.android.com/search?q=NF_BR_PRI_LAST&ss=android%2Fplatform%2Fsuperproject>
  > <https://cs.android.com/search?q=NF_IP_PRI_LAST&sq=&ss=android%2Fplatform%2Fsuperproject>
  > <https://cs.android.com/search?q=NF_IP6_PRI_LAST&sq=&ss=android%2Fplatform%2Fsuperproject>
- Linux sources showing the use of `NF_*_PRI_LAST` constants as trailing enumeration constants defined in terms of large values.

  > <https://github.com/torvalds/linux/blob/eb3f4b7426cfd2b79d65b7d37155480b32259a11/include/uapi/linux/netfilter_bridge.h#L42>
  > <https://github.com/torvalds/linux/blob/eb3f4b7426cfd2b79d65b7d37155480b32259a11/include/uapi/linux/netfilter_ipv4.h#L44>
  > <https://github.com/torvalds/linux/blob/eb3f4b7426cfd2b79d65b7d37155480b32259a11/include/uapi/linux/netfilter_ipv6.h#L47>
- Illumos sources showing how `_PC_LAST` should be changed when adding new `_PC` constants.

[...]

PR #5121, opened 2026-05-28
unix: annotate unstable `*COUNT` constants

dybucc:
# Description

This PR adds documentation. It links to usage advice. Some symbols need this. They can change between upstream releases. They are not considered SemVer-breaking.

I have some doubts. They are concerned with the GNU Hurd. The `DLFO_STRUCT_HAS_EH_COUNT` constant was not found. It has not suffered any changes.

## Sources

- Haiku sources showing how the `B_MEDIA_STALE_CHANGE_COUNT` is **not** the type of constant we would want to deprecate. The code upstream itself does not seem to use it as a delimitter value, as can be seen in the second search result of the below reference.

  > <https://github.com/search?q=repo%3Ahaiku%2Fhaiku%20B_MEDIA_STALE_CHANGE_COUNT&type=code>
- DragonFly BSD sources showing how the `CPUCTL_CPUID_COUNT` constant is **not** one we would want to deprecate, even though it may seem like it could be deprecated (because it appears as a `*COUNT` symbol after a bunch of other symbols.) This constant is actually used for literal CPU counts in
  <https://github.com/DragonFlyBSD/DragonFlyBSD/blob/master/sys/dev/misc/cpuctl/cpuctl.c#L145>.

  > <https://github.com/DragonFlyBSD/DragonFlyBSD/blob/master/sys/sys/cpuctl.h#L62>
- OpenBSD sources showing how the `KERN_TTYCOUNT` is actually used as a counter for TTY devices and not for other constants (just another enum-like "variant".) This does not qualify it for deprecation.

  > <https://github.com/search?q=repo%3Aopenbsd%2Fsrc%20KERN_TTYCOUNT&type=code>
- OpenBSD sources showing how the `KERN_EVCOUNT` is used as an event counter (with use akin to that of the above list item.) This does not qualify it for deprecation.

  > <https://github.com/search?q=repo%3Aopenbsd%2Fsrc%20KERN_EVCOUNT&type=code>
- NetBSD sources showing how the `DCCP_OPT_NDP_COUNT` constant is used as just a regular constant (not meant for deprecation.)

  > <https://github.com/NetBSD/src/blob/d28519920c966c62fedd65ccc0dbd8b628401da5/sys/netinet/dccp.h#L177>
- Linux sources showing how the `TIOCGICOUNT` constant seems to be used not as a counter for other symbols but as (possibly runtime) interrumpt counter (not meant to be deprecated.)

  > <https://github.com/torvalds/linux/blob/eb3f4b7426cfd2b79d65b7d37155480b32259a11/include/uapi/asm-generic/ioctls.h#L99>
- FreeBSD sources showing how the `IFMIB_IFCOUNT` constant is promptly liable to upstream changes as it denotes the number of configured interfaces.

  > <https://github.com/freebsd/freebsd-src/blob/f9f46294d6af2a937afa74938bd4bb6826cbb921/sys/net/if_mib.h#L60>
- XNU kernel sources showing how the `RTV_HOPCOUNT` is not the type of constant we would want to deprecate. It seems to be more of an enum "variant" among other similar constants.

  > <https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/net/route.h#L228>
- Illumos sources showing that the `IPV6_MINHOPCOUNT` is not dependent of other source code chagnes (and thus does not require deprecation.)

  > <https://github.com/illumos/illumos-gate/blob/7124a0cccb6ebf2bcffdab36f9bdf8fe4c23d77f/usr/src/uts/common/netinet/in.h#L1309>
- L4RE sources showing how the `SIOCGIFCOUNT` constant is more of an enum "variant" among other values, and thus not a targe of these deprecation efforts.

  > <https://github.com/search?q=repo%3Akernkonzept%2Fl4re-core%20SIOCGIFCOUNT&type=code>
- Illumos sources showing how the `LGRP_RSRC_COUNT` constant is meant to count the number of resource types, which for the time being are the two constants below it.

  > <https://github.com/search?q=repo%3Aillumos%2Fillumos-gate%20LGRP_RSRC_COUNT&type=code>
- Linux sources mentioning that the constants declared prior to the `NFNL_SUBSYS_COUNT` symbol could have been in an enumeration where the last value is the current "limit."

  > <https://github.com/search?q=repo%3Atorvalds%2Flinux%20NFNL_SUBSYS_COUNT&type=code>
- Android sources replicating similar behavior as in Linux.

  > <https://cs.android.com/search?q=NFNL_SUBSYS_COUNT&sq=&ss=android%2Fplatform%2Fsuperproject>

## Checklist

### DIFF
--- a/src/unix/bsd/apple/mod.rs
+++ b/src/unix/bsd/apple/mod.rs
@@ -3736,7 +3736,11 @@ pub const MNT_NOWAIT: c_int = 2;
 
 // <mach/thread_policy.h>
 pub const THREAD_STANDARD_POLICY: c_int = 1;
+
+/// This symbol is prone to change across releases upstream.
+/// See the [usage guidelines](crate#usage-guidelines) for details.
 pub const THREAD_STANDARD_POLICY_COUNT: c_int = 0;
+
 pub const THREAD_EXTENDED_POLICY: c_int = 1;
 pub const THREAD_TIME_CONSTRAINT_POLICY: c_int = 2;
 pub const THREAD_PRECEDENCE_POLICY: c_int = 3;
@@ -3798,7 +3802,11 @@ pub const VM_PAGE_QUERY_PAGE_CS_NX: i32 = 0x400;
 
 // mach/task_info.h
 pub const TASK_THREAD_TIMES_INFO: u32 = 3;
+
+/// This symbol is prone to change across releases upstream.
+/// See the [usage guidelines](crate#usage-guidelines) for details.
 pub const HOST_CPU_LOAD_INFO_COUNT: u32 = 4;
+
 pub const MACH_TASK_BASIC_INFO: u32 = 20;
 
 pub const MACH_PORT_NULL: i32 = 0;
@@ -3861,7 +3869,10 @@ pub const COPYFILE_STATE_DST_BSIZE: c_int = 12;
 pub const COPYFILE_STATE_BSIZE: c_int = 13;
 
 // <sys/attr.h>
+/// This symbol is prone to change across releases upstream.
+/// See the [usage guidelines](crate#usage-guidelines) for details.
 pub const ATTR_BIT_MAP_COUNT: c_ushort = 5;
+
 pub const FSOPT_NOFOLLOW: u32 = 0x1;
 pub const FSOPT_NOFOLLOW_ANY: u32 = 0x800;
 pub const FSOPT_REPORT_FULLSIZE: u32 = 0x4;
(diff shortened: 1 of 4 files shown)


dybucc:
CI is failing for reasons unrelated to the changes introduced in this patch. A rerun should do it.

dybucc:

tgross35:
For updating doc comments like the others, @rustbot author

dybucc:
@tgross35 The PR is updated. The description has changed. It should be clearer now.

@rustbot ready

src/unix/bsd/apple/mod.rs
```diff
@@ -3736,7 +3736,11 @@ pub const MNT_NOWAIT: c_int = 2;
 
 // <mach/thread_policy.h>
 pub const THREAD_STANDARD_POLICY: c_int = 1;
+
+/// This symbol is prone to change across releases upstream.
+/// See the [usage guidelines](crate::#usage-guidelines) for details and use.
 pub const THREAD_STANDARD_POLICY_COUNT: c_int = 0;
+
```
tgross35, line 3743: Might be okay since it has a value of 0?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5121/00183a7e838b35441481bbb712046b0b124def19..07fba219bc66e760bbf5fcde7f9941c512a0ad6e)*

src/unix/bsd/apple/mod.rs
```diff
@@ -3798,7 +3802,11 @@ pub const VM_PAGE_QUERY_PAGE_CS_NX: i32 = 0x400;
 
 // mach/task_info.h
 pub const TASK_THREAD_TIMES_INFO: u32 = 3;
+
+/// This symbol is prone to change across releases upstream.
+/// See the [usage guidelines](crate::#usage-guidelines) for details and use.
```
tgross35: Same message nit as https://github.com/rust-lang/libc/pull/5120#discussion_r3465462086

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5121/00183a7e838b35441481bbb712046b0b124def19..07fba219bc66e760bbf5fcde7f9941c512a0ad6e)*

src/unix/bsd/apple/mod.rs
dybucc: It's meant to denote a number of arguments. Each thread policy is parameterized. The standard thread policy is the default. It requires no parameters. It's declared in [[1]].

It's value has been the same for over three decades. Though the choice of declaration has changed. Five years ago they hardcoded the `0` value [[2]]. It had been using a more programmatic expression before that. Maybe that's indicative of a commitment?

[1]: https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/osfmk/mach/thread_policy.h#L93
[2]: https://github.com/apple-oss-distributions/xnu/blame/5004dab003baab80eb0c5bebf3b5e086ccd299fb/osfmk/mach/thread_policy.h#L97-L98

src/unix/bsd/apple/mod.rs
dybucc: Done.

tgross35 (approved):
LGTM but has a WIP commit to squash, and needs CI fixed.

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5121/95959125cea361c1b3676b02adb1e4e1772cae9c..c6a4fcd34d61c0ae6f1e3849ef8c0f13ba36de92)*

dybucc:
Done.

@rustbot ready

PR #5127, opened 2026-05-31
fuchsia: clean up module

dybucc:
# Description

This is a clean-up patch. Changes are concerned with the `fuchsia` module.

There were lacking definitions. There were records with the wrong alignment. There were types that don't exist upstream. This has been fixed. But there's more stuff to fix.

Testing is pending. **This time testing is required. Two supported Fuchsia targets are tier 2. There's no CI job for them.**

## Sources

Paths are relative to the downloadable SDK. Look under the `obj` directory. Changes have been made relative to all `NEXT` releases.

---

- `wchar_t` is not different across architectures. No specific definition was found. Verifying the current definition was not possible.
- `nlink_t` is not different across architectures.
- `blksize_t` is not different across architectures.
- `ssizet_t` had a wrong definition. Other `_t`-terminated types had similar issues. Those will not be further repeated in this list.
- A number of types did not exist in the SDK. They exist in vendored libc implementations. These are not part of the SDK. They can be found in `cs.opensource.google`. This includes suffixed offset types. These are omitted from further mention.

> `sysroot/include/bits/alltypes.h`

---

- `stat` had a mismatched definition.

  > `sysroot/include/bits/stat.h`

- `glob_t` had a mismatched definition.

  > `sysroot/include/glob.h`

---

- `signal.h` types often had a mismatched definition. This header file appears later on in this source list. They refer to different paths. They contain different definitions.
- Affected types include:
  - `siginfo_t`.
  - `sigevent`.
  - `sigaction`.
- There was trouble with constants. `SIG_DFL` and others are function pointers. They are cast to function pointers from integers. They are sometimes `NULL`. This is not allowed in Rust. I have wrapped them in `Option`. They are all `None` now. They should not all be `None`. Transmutation from a pointer without provenance to a function pointer is required. This causes compile-time errors.

> `sysroot/include/signal.h`

---

- `stat64` doesn't exist.
- `ipc_perm` doesn't exist.
- `epoll_data` doesn't exist. Deprecation lints here keep popping up. I've attempted to silence them. My attempts have been futile.
- `epoll_event` doesn't exist.
- `ff_effect` doesn't exist.
- `termios2` doesn't exist.
- `sysinfo` doesn't exist.
- `mq_attr` doesn't exist.
- `sockaddr_nl` doesn't exist.
- `RLIM_SAVED_MAX` doesn't exist.
- `RLIM_SAVED_CUR` doesn't exist.
- `RLIM_INFINITY` doesn't exist.
- `RLIMIT_RTTIME` doesn't exist.
- `RLIMIT_NLIMITS` doesn't exist.

---

There were conditional checks against 32-bit targets. Rust does not support 32-bit Fuchsia targets. The Fuchsia project does not support 32-bit machine word targets.

---

There was an oddity with `pthread.h` types. They were defined with a single field. This field sometimes had size equivalent to the sum of upstream fields. This has been replaced with the real fields. This has also fixed some size and alignment mismatches.

This has required changes in macros. An attempt has been made to fit the POSIX definitions. This means expanding `static` initializers to `{ 0 }`. The new type definitions get all their fields zeroed. Some are made into null pointers. Should they be made into `Option::None`s? They are not function pointer fields.

> `sysroot/include/pthread.h`

---

Macros needed fixing. Their definition did not return a value. But a value was returned upstream. Some types needed small fixes.

> `sysroot/include/sched.h`

---

- `m_context` had a mismatched definition. Some modules didn't have a definition.
- `u_context` had a mismatched definition. Some modules didn't have a definition.

> `sysroot/include/bits/signal.h`

## Checklist


dybucc:

tgross35:
@rustbot blocked

Like other deprecations, holding off until we can actually give the users an alternative.

SnoozeThis:
(https://snoozeth.is/db3wDy9dFKA) I will wait until Wed, 19 Aug 2026 08:31:22 UTC and then add label S-waiting-on-review and remove label S-blocked.

@rustbot claim.

dybucc:
There's some problems in Fuchsia. Each vendored libc implementation has a different implementation. Sometimes they can be unified. Other times it's not possible. An example of the latter is `stat`. It resolves to different padding fields.

Should I also expose `cfg`s for each libc implementation?

dybucc:
@tgross35 The PR has been remade. It now cleans up the `fuchsia` module.

The description is hopefully clearer. The details of the changes are included there. There should be no need for additional `cfg`s. Fuchsia has sane defaults and no suffixed types whatsoever. This is only taking into consideration their downloadable SDK. Third party libc implementations they vendor have been ignored.

One deprecation causes CI to fail. Its uses are `allow`ed. That seems insufficient.

Style checks for reasons unknown. The indentation is right. The diagnosed code does not fit the code in my branch. That's likely related to macro expansion issues.

@rustbot ready review

src/fuchsia/aarch64.rs
```diff
@@ -15,20 +16,21 @@ s! {
...
-        pub st_atime: crate::time_t,
-        pub st_atime_nsec: c_long,
-        pub st_mtime: crate::time_t,
-        pub st_mtime_nsec: c_long,
-        pub st_ctime: crate::time_t,
-        pub st_ctime_nsec: c_long,
-        __unused: Padding<[c_uint; 2]>,
+        pub st_atim: crate::timespec,
+        pub st_mtim: crate::timespec,
+        pub st_ctim: crate::timespec,
```
tgross35: https://github.com/rust-lang/libc/pull/5069 will do all the `timespec` changes at once, let's wait for that here.

src/fuchsia/aarch64.rs
```diff
@@ -62,6 +68,25 @@ s! {
         __unused1: Padding<c_ulong>,
         __unused2: Padding<c_ulong>,
     }
+
+    // FIXME(i128): __reserved is meant to be an array of `long double`s. That
+    // requires native support for `i128`. This is currently missing.
```
tgross35: If you have a source for `long double` being `f128`, mind posting it at https://github.com/rust-lang/rust/pull/140417? I would have assumed `f64`.

Also this comment says `i128` but should be `f128`.

tgross35 (changes requested):
There is a lot of orthogonal changes going on here. Could you split it into separate commits to make this more clear? Each section of the PR description could be its own commit, but at a minimum please be sure to separate things like non-functional cleanup and deprecations from anything that's a behavior change. All of this in the same PR is completely fine of course.

(Splitting changes that can stand on their own is good practice in general, see https://www.kernel.org/doc/html/v7.0/process/submitting-patches.html#separate-your-changes for some of the canonical guidelines. Any descriptions/links can also live in the commit messages, it's fine to say "see commit messages for details" in the top post.)

> Paths are relative to the downloadable SDK. Look under the `obj` directory. Changes have been made relative to all `NEXT` release

In case you haven't come across it, web sources are at https://cs.opensource.google/fuchsia/fuchsia.

> Testing is pending. **This time testing is required. Two supported Fuchsia targets are tier 2. There's no CI job for them.**

It's not required - T2 targets are only "guaranteed to build", which the "verify build" jobs do. We do test a handful of T2 targets, but that's out of convenience rather than requirement.

That said, of course testing everything is ideal. I'll ping the Fuschia maintainers after history gets split.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5127/63001363035fc2ed974f82689416c85e3725ebab..48d1e3d7dfad35067b0ab87765da4c693752b319)*

tgross35:
> There's some problems in Fuchsia. Each vendored libc implementation has a different implementation. Sometimes they can be unified. Other times it's not possible. An example of the latter is `stat`. It resolves to different padding fields.
> 
> Should I also expose `cfg`s for each libc implementation?

Each fuchsia target should only support a single libc. I don't see that listed in https://doc.rust-lang.org/rustc/platform-support/fuchsia.html but perhaps it should be? You could probably open a r-l/r PR updating that if there is a straightforward answer, or an issue if there isn't.

dybucc:
> There is a lot of orthogonal changes going on here. Could you split it into separate commits to make this more clear? Each section of the PR description could be its own commit, but at a minimum please be sure to separate things like non-functional cleanup and deprecations from anything that's a behavior change. All of this in the same PR is completely fine of course.
> 
> (Splitting changes that can stand on their own is good practice in general, see https://www.kernel.org/doc/html/v7.0/process/submitting-patches.html#separate-your-changes for some of the canonical guidelines. Any descriptions/links can also live in the commit messages, it's fine to say "see commit messages for details" in the top post.)
> 
> > Paths are relative to the downloadable SDK. Look under the `obj` directory. Changes have been made relative to all `NEXT` release
> 
> In case you haven't come across it, web sources are at https://cs.opensource.google/fuchsia/fuchsia.
> 
> > Testing is pending. **This time testing is required. Two supported Fuchsia targets are tier 2. There's no CI job for them.**
> 
> It's not required - T2 targets are only "guaranteed to build", which the "verify build" jobs do. We do test a handful of T2 targets, but that's out of convenience rather than requirement.
> 
> That said, of course testing everything is ideal. I'll ping the Fuschia maintainers after history gets split.
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5127/63001363035fc2ed974f82689416c85e3725ebab..48d1e3d7dfad35067b0ab87765da4c693752b319)_

I did come across the source browser. That's were I got confused by the vendored libc implementations. The SDK provides a single set of headers. This makes searches more "coherent."

The testing part was not meant for others. It was a personal reminder. But I understood the testing policy wrong. Just reread "Platform Support" in the rustc book. Thanks.

src/fuchsia/aarch64.rs
dybucc: I looked it up in the `libc` crate. There's a comment in https://github.com/rust-lang/libc/blob/main/src/unix/nto/mod.rs#L799-L810. `f128` has equivalent alignment and size constraints to `__int128` in gcc.

I read the RFC for `f128`. It didn't comment on this. I have assumed the comment correct. But it could be wrong.

tgross35:
Fuchsia maintainers @erickt @Nashenas88 could you clarify which libc headers we should be looking at in the Fuchsia sources?

src/fuchsia/mod.rs
```diff
@@ -759,7 +759,15 @@ s! {
...
+        pub __name: *const c_char,
+        pub __c11: c_int,
+        pub _a_stacksize: crate::size_t,
+        pub _a_guardsize: crate::size_t,
+        pub _a_stackaddr: *mut c_void,
+        pub _a_detach: c_int,
+        pub _a_sched: c_int,
+        pub _a_policy: c_int,
+        pub _a_prio: c_int,
     }
```
tgross35, line 799: Commit 1: we prefer to keep these opaque. At least the fields should be private, but doesn't the libc just define a repr like this?

src/fuchsia/mod.rs
```diff
@@ -1081,12 +1053,14 @@ cfg_if! {
...
 
-pub const SIG_DFL: sighandler_t = 0 as sighandler_t;
-pub const SIG_IGN: sighandler_t = 1 as sighandler_t;
-pub const SIG_ERR: sighandler_t = !0 as sighandler_t;
+pub const INT_MIN: c_int = -1 - 0x7fffffff;
+pub const INT_MAX: c_int = 0x7fffffff;
+
+pub const SIG_ERR: Option<sighandler_t> = None;
+pub const SIG_DFL: Option<sighandler_t> = None;
+pub const SIG_IGN: Option<sighandler_t> = None;
```
tgross35: Commit 1: can you clarify the `sighandler_t` change? If `sighandler_t` is `size_t`, this doesn't seem right.

src/fuchsia/mod.rs
```diff
@@ -1102,14 +1076,14 @@ pub const FD_CLOEXEC: c_int = 0x1;
...
-pub const SIGIOT: c_int = 6;
+pub const SIGIOT: c_int = crate::SIGABRT;
 
 pub const S_ISUID: mode_t = 0o4000;
 pub const S_ISGID: mode_t = 0o2000;
 pub const S_ISVTX: mode_t = 0o1000;
 
 pub const IF_NAMESIZE: size_t = 16;
-pub const IFNAMSIZ: size_t = IF_NAMESIZE;
+pub const IFNAMSIZ: size_t = crate::IF_NAMESIZE;
```
tgross35: Commit 1: don't these work fine as-is?

src/fuchsia/mod.rs
```diff
@@ -354,7 +354,7 @@ s! {
     }
 
     pub struct fd_set {
-        fds_bits: [c_ulong; FD_SETSIZE as usize / ULONG_SIZE],
+        pub fds_bits: [c_ulong; FD_SETSIZE as usize / 8 / size_of::<c_long>()],
```
tgross35: Commit 2: if there hasn't been demand for these fields to be public, make them private.

Commit message has a typo in "It possibley also fixes the definition.", but I'm not sure what that is referring to?

src/fuchsia/mod.rs
```diff
@@ -1009,6 +1000,18 @@ s_no_extra_traits! {
...
+        sa_handler: Option<extern "C" fn(c_int)>,
+        sa_sigaction: Option<extern "C" fn(c_int, *mut siginfo_t, *mut c_void)>,
+    }
+
+    pub struct sigaction {
+        pub __sa_handler: crate::__c_anonymous_sigaction___sa_handler,
+        pub sa_mask: crate::sigset_t,
+        pub sa_flags: c_int,
+        pub sa_restorer: Option<extern "C" fn()>,
+    }
```
tgross35: Commit 3: sigaction is broken on all platforms and we should probably fix them all at once, so this commit should be split out. See also:

* https://github.com/rust-lang/libc/pull/2107
* https://github.com/rust-lang/libc/pull/5076
* https://github.com/rust-lang/libc/issues/4998

src/fuchsia/mod.rs
tgross35, line 1: Commit "fuchsia: mirror type aliases": could you add a bulleted list to the commit message of which of these are wrong (i.e. now fixed) vs. just simplified?

src/fuchsia/mod.rs
```diff
@@ -994,6 +994,11 @@ s_no_extra_traits! {
         pub sival_ptr: *mut c_void,
     }
 
+    pub union fpos_t {
+        __opaque: [c_char; 16],
+        __align: c_double,
+    }
```
tgross35, line 1058: Commit "fuchsia: add `fpos_t`" we have that FIXME in a few places but I've never really understood it. Does this actually improve things for users?

src/fuchsia/mod.rs
```diff
@@ -521,15 +516,10 @@ s! {
         pub f_files: crate::fsfilcnt_t,
         pub f_ffree: crate::fsfilcnt_t,
         pub f_favail: crate::fsfilcnt_t,
-        #[cfg(target_endian = "little")]
-        pub f_fsid: c_ulong,
-        #[cfg(all(target_pointer_width = "32", not(target_arch = "x86_64")))]
-        __f_unused: Padding<c_int>,
-        #[cfg(target_endian = "big")]
         pub f_fsid: c_ulong,
```
tgross35, line 534: Commit "fuchsia: fix record definitions": are these endian/bit-width constants in the Fuchsia source? If so, we may as well just keep it around to make things easier if there is a 32-bit port in the future.

src/fuchsia/x86_64.rs
```diff
@@ -55,7 +82,18 @@ s! {
...
+        __reserved1: Padding<[c_ulonglong; 8]>,
+    }
+
+    pub struct ucontext_t {
+        pub uc_flags: c_ulong,
+        pub uc_link: *mut ucontext_t,
+        pub uc_stack: crate::stack_t,
+        pub uc_mcontext: crate::mcontext_t,
+        pub uc_sigmask: crate::sigset_t,
+        pub __fpregs_mem: [c_ulong; 64],
```
tgross35: Commit "fuchsia: fix `bits/signal.h` types": probably can leave this field private

src/fuchsia/x86_64.rs
```diff
@@ -7,7 +7,34 @@ use crate::prelude::*;
...
+        pub significand: [c_ushort; 4],
+        pub exponent: c_ushort,
+        pub padding: [c_ushort; 3],
+    }
+
+    pub struct __c_anonymous__fpstate__xmm {
+        element: [c_uint; 4],
+    }
+
+    pub struct _fpstate {
```
tgross35: Commit "fuchsia: fix `bits/signal.h` types": nit, put `__c_anonymous_*` types after the struct they're a part of, which is what we do most places.

src/fuchsia/riscv64.rs
```diff
@@ -51,4 +51,16 @@ s! {
...
+        pub regs: [c_ulong; 32]
+    }
+
+    pub struct ucontext_t {
+        uc_flags: c_ulong,
+        uc_link: *mut crate::ucontext_t,
+        uc_stack: crate::stack_t,
+        uc_sigmask: crate::sigset_t,
+        uc_mcontext: crate::mcontext_t,
+    }
```
tgross35, line 67: Commit "fuchsia: fix bits/signal.h types": Should these fields be public?

src/fuchsia/aarch64.rs
tgross35, line 1: Commit "fuchsia: fix bits/signal.h types": the aarch64 and riscv64 definitions look new, please make sure the commit message is accurate

tgross35, line 1: Commit "fuchsia: fix bits/signal.h types": the commit message says "Architecture-specific types were wrongly defined. This is fixed now.", and a few other commits do something similar. That doesn't really say what changed - a message that describes the action helps review so we know what to look for, e.g.

> Replace the x86_64 `mcontext_t` opaque field with `gregs`/`fregs`, matching upstream definitions.

(https://docs.kernel.org/process/submitting-patches.html#describe-your-changes has some good guidelines here. Not all of it is relevant here of course, but at least the "Describe user-visible impact" and "Describe your changes in imperative mood" bits are great guidelines for all commit messages.)

tgross35 (commented):
Thank you for splitting it up, this makes things much easier :) Left some notes from a first pass review.

Once the maintainers respond about which libc we should use, please add source links to the commit messages.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5127/00183a7e838b35441481bbb712046b0b124def19..d1e7799b8d3f3d447a84910fa8d06fac1609ff4f)*

src/fuchsia/mod.rs
dybucc: It doesn't. See https://cs.opensource.google/fuchsia/fuchsia/+/main:zircon/third_party/ulib/musl/include/bits/alltypes.h;l=453-463?q=symbol:%5Cbpthread_attr_t%5Cb&ss=fuchsia.

src/fuchsia/mod.rs
dybucc: `sighandler_t` is not `size_t` anymore.

src/fuchsia/mod.rs
dybucc: Yes. The definitions were equivalent. Now they're mirrored.

src/fuchsia/mod.rs
dybucc: `ULONG_SIZE` was not equivalent to `8 / size_of::<c_long>()`.

src/fuchsia/mod.rs
dybucc: It provides the right size and alignment. The previous definition was behind `extern_ty!`. It seems `fpos_t` is not an opaque type in Fuchsia. The public header contains the full definition.

src/fuchsia/mod.rs
dybucc: I don't understand what you mean. There's no endian-specific code in Fuchsia. All of the targets they ship to are little-endian.

src/fuchsia/mod.rs
dybucc: Should I open a separate PR? Or should I remove the commit altogether? The mentioned PRs solve it slightly differently.

src/fuchsia/mod.rs
dybucc: There's none. The patch mirrors definitions. The previous ones were not wrong. They were fixed-width integers. Now they're platform-dependent C types. But Fuchsia only supports LP64 targets. They're effectively equivalent.

The `sighandler_t` type is an exception. The new definition may be clearer. It has implications with `sigaction`. The old definition could be considered "equivalent." It's gone from `size_t` to `extern "C" fn(c_int)`.

src/fuchsia/x86_64.rs
dybucc: Should they? The definition seems public. See https://cs.opensource.google/fuchsia/fuchsia/+/main:zircon/third_party/ulib/musl/include/bits/signal.h;l=68-75?q=pcre:y%20case:y%20symbol:%5Cbmcontext_t%5Cb&ss=fuchsia.

src/fuchsia/riscv64.rs
dybucc: Yes. I missed them in this module.

src/fuchsia/mod.rs
tgross35: I meant the `crate::` prefix, aren't the constants in scope?

src/fuchsia/mod.rs
tgross35: Ah, right, that's not just "possibly" then. Thanks for clarifying in the commit message.

Patch "fuchsia: mirror fd_set definition" lgtm

src/fuchsia/mod.rs
tgross35: Could you put this patch in a draft PR? It won't merge on its own but it's helpful to see the alternatives, considering this is a bit of a mess.

src/fuchsia/mod.rs
tgross35: That just wasn't done in this patch, so this change would need to move with the `sighandler_t` changes.

Note that for any type alias (`sighandler_t`), `Option<extern "C" fn(c_int)>` is going to be a more accurate representation than `extern "C" fn(c_int);` because it's probably nullable in C. If the definition is only `extern "C" fn(c_int);`, then passing `NULL` to a function that takes `sighandler_t` is immediate UB - wrapping it in an option prevents that.

Now of course neither of these help here because `sighandler_t` needs to be able to accept arbitrary values. https://github.com/rust-lang/libc/issues/4998 is about the only possible improvement here, so sighandler changes should be dropped until the libs team has time to discuss.

src/fuchsia/mod.rs
tgross35: The commit says "Existing type aliases are equivalent or wrong" so please clarify that. It also says `sigaction_t` but I think you meant `sighandler_t`.

That change isn't one we can do though, https://github.com/rust-lang/libc/pull/5127#discussion_r3466005095

src/fuchsia/mod.rs
tgross35: That was my question, whether the source has things like `#if __LONG_MAX > 0x7fffffff && __BYTE_ORDER == __BIG_ENDIAN` that musl does. If it did, I would want to preserve the equivalent here - but if it doesn't then this is a good cleanup.

src/fuchsia/x86_64.rs
tgross35: Everything seems public in C :) if most fields are normal but only one starts with `__`, sometimes that's a hint that it's semi-private. So we can leave it as such, and make it public if there's ever demand.

src/fuchsia/mod.rs
tgross35: Huh, that's an interesting deviation from https://github.com/kraj/musl/blob/8cb84492b0245d70b2cd0edd523e2b55c7ad67a9/include/alltypes.h.in#L85 then. Still preferred to make these private.

I wonder what `__c11` is used for internally, can't seem to find it consumed in the source.

src/fuchsia/mod.rs
tgross35: I was thinking POSIX guided toward making this an incomplete type, which they did https://pubs.opengroup.org/onlinepubs/9699919799.orig/basedefs/stdio.h.html, but with issue 8 they now specify a complete object type https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/stdio.h.html. Interesting.

With that in mind this change lgtm, but the commit message should say something like "fill in the definition for `fpos_t`" rather than "add `fpos_t`" (since it already exists)

src/fuchsia/aarch64.rs
dybucc: I think it's better now.

src/fuchsia/x86_64.rs
dybucc: Noted. Done.

src/fuchsia/mod.rs
dybucc: I think it's safe to do this. Fuchsia does not provide support for 64-bit targets [[1]].

[1]: https://fuchsia.dev/fuchsia-src/contribute/governance/rfcs/0111_fuchsia_hardware_specifications#required_64-bit_cpu_and_platform

src/fuchsia/mod.rs
dybucc: Done.

src/fuchsia/mod.rs
dybucc: Done.

src/fuchsia/mod.rs
dybucc: Just dropped those changes from the commit.

src/fuchsia/mod.rs
dybucc: They are. I made all non-"primitive" C types crate-relative.

src/fuchsia/mod.rs
dybucc: Done.

src/fuchsia/mod.rs
dybucc: The description didn't mention that anymore. Maybe I pushed changes after you started the re-review?

Just dropped the changes with `sighandler_t`.

src/fuchsia/riscv64.rs
```diff
@@ -3,10 +3,6 @@ use crate::prelude::*;
 
 // From psABI Calling Convention for RV64
 pub type __u64 = c_ulonglong;
-pub type wchar_t = i32;
```
tgross35, line 6: Commit "fuchsia: mirror type aliases": it looks like `wchar_t` was `i32` on rv64 and x86, but `u32` on aarch64 so they probably shouldn't be combined

src/fuchsia/mod.rs
```diff
@@ -1105,14 +1105,14 @@ pub const FD_CLOEXEC: c_int = 0x1;
...
-pub const SIGIOT: c_int = 6;
+pub const SIGIOT: c_int = crate::SIGABRT;
 
 pub const S_ISUID: mode_t = 0o4000;
 pub const S_ISGID: mode_t = 0o2000;
 pub const S_ISVTX: mode_t = 0o1000;
 
 pub const IF_NAMESIZE: size_t = 16;
-pub const IFNAMSIZ: size_t = IF_NAMESIZE;
+pub const IFNAMSIZ: size_t = crate::IF_NAMESIZE;
```
tgross35: Commit "fuchsia: fix constant definitions": leave as-is, we tend not to use `crate::` when it's in the same module, and these are right next to each other anyway.

src/fuchsia/mod.rs
```diff
@@ -1029,6 +1029,11 @@ s_no_extra_traits! {
         pub sival_ptr: *mut c_void,
     }
 
+    pub union fpos_t {
+        __opaque: [c_char; 16],
+        __align: c_double,
+    }
```
tgross35, line 1058: Commit "fuchsia: fill in definition of `fpos_t`": LGTM @ d8710c029ca15b12f8a49c4050bd69d149ddc51f.

src/fuchsia/aarch64.rs
tgross35, line 1: Commit "fuchsia: deprecate nonexistent types": LGTM @ 9ace3f960f8f271cebef279569e3f604109806ab

src/fuchsia/mod.rs
```diff
@@ -476,16 +476,11 @@ s! {
...
         pub gl_pathv: *mut *mut c_char,
-        pub gl_offs: size_t,
-        pub gl_flags: c_int,
-
-        __unused1: Padding<*mut c_void>,
-        __unused2: Padding<*mut c_void>,
-        __unused3: Padding<*mut c_void>,
-        __unused4: Padding<*mut c_void>,
-        __unused5: Padding<*mut c_void>,
+        pub gl_offs: crate::size_t,
```
tgross35: Commit "fuchsia: fix record definitions": `size_t` is in the prelude, no need for `crate::`

src/fuchsia/mod.rs
```diff
@@ -476,16 +476,11 @@ s! {
     }
 
     pub struct glob_t {
-        pub gl_pathc: size_t,
+        pub gl_pathc: crate::size_t,
         pub gl_pathv: *mut *mut c_char,
-        pub gl_offs: size_t,
-        pub gl_flags: c_int,
```
tgross35, line 430: Commit "fuchsia: fix record definitions": field removed, seems to match https://cs.opensource.google/fuchsia/fuchsia/+/main:zircon/third_party/ulib/musl/include/glob.h;l=18;drc=575a4c11f7d91a70e82ad5a28a216744148b6e8b but would prefer maintainer ack if possible

src/fuchsia/mod.rs
```diff
@@ -583,14 +573,11 @@ s! {
...
+        pub __val: [c_int; 2],
     }
 
     pub struct cpu_set_t {
-        #[cfg(all(target_pointer_width = "32", not(target_arch = "x86_64")))]
-        bits: [u32; 32],
-        #[cfg(not(all(target_pointer_width = "32", not(target_arch = "x86_64"))))]
-        bits: [u64; 16],
+        pub __bits: [c_ulonglong; 128 / size_of::<c_long>()],
     }
```
tgross35, line 588: Commit "fuchsia: fix record definitions": don't change from private to public unless there is a need

src/fuchsia/mod.rs
```diff
@@ -808,7 +782,7 @@ s! {
...
 
     pub struct sigset_t {
-        __val: [c_ulong; 16],
+        pub __bits: [c_ulong; 128 / size_of::<c_long>()],
     }
```
tgross35, line 803: Ditto

src/fuchsia/mod.rs
```diff
@@ -873,15 +847,16 @@ s! {
...
 
+    #[repr(align(8))]
     pub struct siginfo_t {
         pub si_signo: c_int,
         pub si_errno: c_int,
         pub si_code: c_int,
-        pub _pad: [c_int; 29],
-        _align: [usize; 0],
+        pub __si_fields: [c_int; 28],
     }
```
tgross35, line 877: Ditto, the new fields here can remain private unless somebody shows up with a need for them

src/fuchsia/mod.rs
```diff
@@ -952,13 +927,14 @@ s! {
...
-        __ss_align: size_t,
+        __ss_padding:
+            Padding<[c_char; 128 - size_of::<c_long>() - size_of::<crate::sa_family_t>()]>,
+        __ss_align: c_ulong,
     }
 
     pub struct utsname {
         pub sysname: [c_char; 65],
-        pub nodename: [c_char; 65],
+        pub nodename: [c_char; crate::HOST_NAME_MAX.cast_unsigned() as usize + 1],
```
tgross35: Not in MSRV

src/fuchsia/mod.rs
```diff
@@ -3140,31 +3118,43 @@ f! {
...
+    pub fn CPU_CLR(cpu: usize, cpuset: &mut cpu_set_t) -> c_ulonglong {
+        (cpu / 8 >= size_of::<crate::cpu_set_t>())
+            .then_some(0)
+            .unwrap_or_else(|| {
+                let c = &mut cpuset.__bits[cpu / 8 / size_of::<c_long>()];
+                *c &= !(1 << (cpu % (8 * size_of::<c_long>())));
+
+                *c
+            })
     }
```
tgross35: I don't think these are supposed to return values, or at least that's more of an implementation detail than something we need to show. Per https://man7.org/linux/man-pages/man3/CPU_SET.3.html they're treated as `void` functions.

src/fuchsia/mod.rs
```diff
@@ -3140,31 +3118,43 @@ f! {
...
-        let size_in_bits = 8 * size_of_val(&cpuset.bits[0]);
-        let (idx, offset) = (cpu / size_in_bits, cpu % size_in_bits);
-        0 != (cpuset.bits[idx] & (1 << offset))
+        (cpu / 8 >= size_of::<crate::cpu_set_t>())
+            .then_some(false)
+            .unwrap_or_else(|| {
+                (cpuset.__bits[cpu / 8 / size_of::<c_long>()]
+                    & (1 << (cpu % (8 * size_of::<c_long>()))))
+                    != 0
+            })
```
tgross35: `then`/`then_some` tends to obfuscate control flow, just use `if`.

src/fuchsia/aarch64.rs
tgross35, line 1: Commit "fuchsia: fix `bits/signal.h` types": LGTM @ 33d311a4a82c3a707195fcb5b4e912820e00c730, but should wait on a maintainer to test ideally.

Note: when you link on codesearch, use `l-r` or links->commit to get a permalink in case sources move around.

tgross35 (commented):
For any of the commits where I said LGTM, if you want to split them to a separate PR I'd be happy to apply them now to thin down how much is in this PR.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5127/f08bfce345c586260a5678dc4442b3e9de856382..2b8b9da01dce54c3cff09660be60ea7768c54cba)*

src/fuchsia/riscv64.rs
dybucc: Fair. But there's something to be noted here. `wchar_t` is always defined in terms of `__WCHAR_TYPE__`. This type does not seem to appear anywhere in the Fuchsia tree [[1]]. I haven't had a way of verifying if the current definitions are correct.

[1]: https://cs.opensource.google/search?q=__WCHAR_TYPE__&sq=&ss=fuchsia

tgross35: That one comes from the compiler, search for `__WCHAR_TYPE__` at https://c.godbolt.org/z/efjeK48j5. So I believe the existing definitions are correct.

src/fuchsia/mod.rs
dybucc: Done.

src/fuchsia/riscv64.rs
dybucc: You're right. Thanks.

tgross35:
I haven't done a completely final look-through here but at this point, holding off for a while to give the maintainers a chance to respond.

erickt:
Hi folks! Thanks for this from the fuchsia team! I’ll be a little delayed reviewing this but I asked other folks on the team that hopefully will be able to get to this sooner.

src/fuchsia/mod.rs
erickt: We forked from musl many years ago. As far as I can tell, we don't use `__c11` for anything, so it might just be an artifact from our fork.

src/fuchsia/mod.rs
```diff
@@ -16,36 +16,63 @@ pub type size_t = usize;
 pub type ptrdiff_t = isize;
 pub type intptr_t = isize;
 pub type uintptr_t = usize;
-pub type ssize_t = isize;
+pub type ssize_t = crate::intptr_t;
```
erickt: I think it'd be more correct to use  `c_long` here https://cs.opensource.google/fuchsia/fuchsia/+/main:zircon/third_party/ulib/musl/include/bits/alltypes.h;l=186.

src/fuchsia/mod.rs
erickt: Yeah this change looks correct.

src/fuchsia/mod.rs
```diff
@@ -477,15 +523,10 @@ s! {
         pub f_files: crate::fsfilcnt_t,
         pub f_ffree: crate::fsfilcnt_t,
         pub f_favail: crate::fsfilcnt_t,
-        #[cfg(target_endian = "little")]
```
erickt, line 480: I think this is incorrect based off of https://cs.opensource.google/fuchsia/fuchsia/+/main:zircon/third_party/ulib/musl/include/sys/statvfs.h;l=16. Maybe we'd want this?

```rust
#[cfg(target_endian = "little")]
pub f_fsid: c_ulong,
pub __f_unused: Padding<[c_uchar; 2 * std::mem::size_of::<c_int>() - std::mem::size_of::<c_long>()]>,
#[cfg(not(target_endian = "little"))]
pub f_fsid: c_ulong,
```

src/fuchsia/mod.rs
erickt: I'm guessing this is just an artifact from when we forked from musl a few years ago?

src/fuchsia/mod.rs
```diff
@@ -828,17 +854,22 @@ s! {
...
     }
 
+    #[repr(align(8))]
     pub struct siginfo_t {
         pub si_signo: c_int,
         pub si_errno: c_int,
         pub si_code: c_int,
-        pub _pad: [c_int; 29],
-        _align: [usize; 0],
+        __si_fields: [c_int; 28],
```
erickt: From https://cs.opensource.google/fuchsia/fuchsia/+/main:zircon/third_party/ulib/musl/include/signal.h;l=137, I think this would be more correct:

```rust
__pad<[c_char; 128 - 2 * size_of::<c_int>() - size_of::<c_long>()]>,
```

src/fuchsia/mod.rs
```diff
@@ -970,53 +1003,28 @@ s! {
         pub sigev_value: crate::sigval,
         pub sigev_signo: c_int,
         pub sigev_notify: c_int,
-        pub sigev_notify_function: fn(crate::sigval),
+        pub sigev_notify_function: Option<extern "C" fn(crate::sigval)>,
```
erickt, line 1024: Since we don't have this type I don't think these changes are necessary.

src/fuchsia/mod.rs
```diff
@@ -3006,17 +3098,6 @@ pub const O_NOFOLLOW: c_int = 0x00000080;
 pub const HUGETLB_FLAG_ENCODE_SHIFT: u32 = 26;
 pub const MAP_HUGE_SHIFT: u32 = 26;
 
-// intentionally not public, only used for fd_set
-cfg_if! {
```
erickt, line 3010: Should these be deprecated rather than removed?

erickt (changes requested):
Wow, you did a ton of work, thanks for doing it! Overall looks good to me, just a few things I noted.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5127/076d48ffbddf9408319f72ef7361452f8b0206d1..e3e83d69610abf7db9302a65c95eeb5aab6047bb)*

src/fuchsia/mod.rs
dybucc: I tracked those changes back to a revision where Fuchsia didn't have the same
musl `alltypes.h` header file for all three supported architectures in a single
file [^1]. It seems like this field and the one before it (`__name`) was
intended for detecting C11 threads in pthreads, and possibly for extensions to
C11 threads.

Maybe some other project did that, or maybe that never quite happened. I can't
say for sure.

[^1]:
    https://cs.opensource.google/fuchsia/fuchsia/+/8cd2693ba1f15e7337f154bb5ab1c12faaaead9f

src/fuchsia/mod.rs
dybucc: I don't think so. They are not public and they were only used in the old
(before this commit) definition of `fd_set`. Removing them won't break anybody
and they're not used anywhere else.

src/fuchsia/aarch64.rs
dybucc: Done.

src/fuchsia/mod.rs
dybucc: That's right. The reason why I left it as an `intptr_t` was that the type that
one aliases is `isize`, which we use later on in a C macro implemented as a
Rust free function, which itself uses `<*const>::offset` in Rust's libstd
(which expects an `isize` as its one parameter.)

I assumed that considering the Fuchsia project supports only 64-bit
architectures and of those supported, LP64 is prevalent, `intptr_t` (`isize`)
would be effectively equivalent to `c_long`.

Changed now. The one place where it's expected as an `isize` for
`<*const>::offset` has been made to use a primitive `as` cast. This shouldn't
silently overflow so long as the supported targets all follow LP64.

src/fuchsia/mod.rs
dybucc: There's no supported Rust target for Fuchsia whose endianness is _not_
little-endian, so that's why I skipped the whole conditional compilation aspect
in this instance.

I'm assuming the Fuchsia team doesn't discard adding support for some
big-endian target, then?

src/fuchsia/mod.rs
dybucc: I just added the whole anonymous union. It's not public but that should ensure
the size for that one field is correct.

dybucc:
> Wow, you did a ton of work, thanks for doing it! Overall looks good to me, just a few things I noted.
>
> *[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5127/076d48ffbddf9408319f72ef7361452f8b0206d1..e3e83d69610abf7db9302a65c95eeb5aab6047bb)*

Thanks for getting back. I've reviewed the changes and left some comments as
well.

Also, I always refer to the header files shipped under `arch/` in the Fuchsia
IDK, which actually match the changes I made.

I'm wondering which place should we assume as being the definitive reference
for all things Fuchsia. I went for the IDK because the CodeSearch site just has
a ton of results concerning both the Zircon kernel as well as multiple third
party libc implementations (musl, LLVM-libc.)

src/fuchsia/mod.rs
dybucc: We do, though [^1]?

[^1]:
    https://cs.opensource.google/fuchsia/fuchsia/+/main:zircon/third_party/ulib/musl/include/signal.h;l=209-216;drc=95974268cca265449c090e3c06dfdaa055ec90a8

src/fuchsia/aarch64.rs
tgross35: That link is no longer updated, here's a permalink https://github.com/rust-lang/libc/blob/f4d67773492709f79fcbd4b6f60dd26778291d87/src/unix/nto/mod.rs#L799-L810.

The comment is only about QNX, it doesn't apply to anything else. Other platforms can and do have different `long double`s, see https://github.com/rust-lang/rust/pull/140417 for some examples.

The `f128` RFC isn't relevant here.

src/fuchsia/mod.rs
tgross35: Looks like `cpu_set_t.__bits` is still public

src/fuchsia/mod.rs
```diff
@@ -902,27 +966,33 @@ s! {
...
-        __ss_align: size_t,
+        __ss_padding:
+            Padding<[c_char; 128 - size_of::<c_long>() - size_of::<crate::sa_family_t>()]>,
+        __ss_align: c_ulong,
     }
 
     pub struct utsname {
         pub sysname: [c_char; 65],
-        pub nodename: [c_char; 65],
+        pub nodename: [c_char; HOST_NAME_MAX as usize + 1],
```
tgross35, line 948: We now have a handful of functions for casting types without risking overflow, https://github.com/rust-lang/libc/blob/efe08d750008352469c0546ca98d469d927a68fe/src/types.rs#L69-L115. No need to change it here since there is no `int_cast_usize` yet, just a heads up.

src/fuchsia/mod.rs
tgross35: I am a bit hesitant here: this is correct to the letter, but possibly not to the spirit:

* There has been discussion about whether `rustc` could say that `usize` is compatible with `size_t` and `isize` is compatible with `ssize_t` (considering it seems like POSIX is willing to get rid of the weird "only represent -1" requirement).
* All but one platform currently has `ssize_t`->`isize` or `ssize_t`->`intptr_t`->`isize`, even though they're probably technically defined to a `long`

I don't think it's worth making this change with those couple things in consideration.

src/fuchsia/mod.rs
```diff
@@ -1104,7 +1104,7 @@ pub const FD_CLOEXEC: c_int = 0x1;
 pub const USRQUOTA: c_int = 0;
 pub const GRPQUOTA: c_int = 1;
 
-pub const SIGIOT: c_int = 6;
+pub const SIGIOT: c_int = SIGABRT;
```
tgross35, line 1207: "fuchsia: fix constant definitions": Nit, commit message says fix but this isn't a behavior fix - just resolving a mismatch

src/fuchsia/aarch64.rs
```diff
@@ -27,6 +31,10 @@ s! {
         __unused: Padding<[c_uint; 2]>,
     }
 
+    #[deprecated(
+        since = "0.2.187",
+        note = "This type doesn't exist. It's not part of the Fuchsia SDK."
+    )]
```
tgross35, line 38: "fuchsia: deprecate nonexistent types": there are a couple of slightly different messages in this commit. I'd suggest making them all the same, maybe:

> This type does not exist in Fuchsia and will be removed from `libc`.

With, if applicable, "Use ... instead.".

Also the versions should ideally be updated to 0.2.189, we've since had a release.

(Applies to most of this commit)

Aside from consistency of the messages, this commit LGTM.

src/fuchsia/mod.rs
```diff
@@ -586,10 +576,7 @@ s! {
     }
 
     pub struct cpu_set_t {
-        #[cfg(all(target_pointer_width = "32", not(target_arch = "x86_64")))]
-        bits: [u32; 32],
-        #[cfg(not(all(target_pointer_width = "32", not(target_arch = "x86_64"))))]
-        bits: [u64; 16],
+        pub __bits: [c_ulonglong; 128 / size_of::<c_long>()],
     }
```
tgross35, line 588: Thought I noted this elsewhere but now can't find it, should remain private

src/fuchsia/mod.rs
```diff
@@ -872,15 +846,59 @@ s! {
...
+
+    pub struct __c_anonymous___si_fields___si_common {
+        pub __first: __c_anonymous___si_common___first,
+        pub __second: __c_anonymous___si_common___second,
+    }
+
+    pub struct __c_anonymous___first___piduid {
+        pub si_pid: pid_t,
+        pub si_uid: uid_t,
+    }
```
tgross35: If you wrap `__si_fields` in `Padding` it gives you a default `true` impl, so all the inner structs/unions can live in `s_no_extra_traits` and drop manual impls.

This should all be able to be crate-private.

src/fuchsia/mod.rs
```diff
@@ -921,27 +929,16 @@ s! {
...
     pub struct pthread_rwlockattr_t {
-        size: [u8; crate::__SIZEOF_PTHREAD_RWLOCKATTR_T],
+        pub __attr: [c_uint; 2],
     }
 
-    #[repr(align(4))]
     pub struct pthread_condattr_t {
-        size: [u8; crate::__SIZEOF_PTHREAD_CONDATTR_T],
+        pub __attr: c_uint,
     }
```
tgross35, line 910: Another few unneeded private->public

src/fuchsia/mod.rs
```diff
@@ -1049,53 +1046,28 @@ s! {
...
-            not(any(target_arch = "arm", target_arch = "x86_64"))
-        ),
-        repr(align(8))
-    )]
     pub struct pthread_mutex_t {
-        size: [u8; crate::__SIZEOF_PTHREAD_MUTEX_T],
+        pub _m_attr: c_uint,
+        pub _m_lock: c_int,
+        pub _m_waiters: c_int,
+        pub _m_count: c_int,
```
tgross35: Ditto

src/fuchsia/mod.rs
```diff
@@ -1049,53 +1046,28 @@ s! {
...
-    #[cfg_attr(target_pointer_width = "32", repr(align(4)))]
-    #[cfg_attr(target_pointer_width = "64", repr(align(8)))]
-    #[cfg_attr(target_arch = "x86", repr(align(4)))]
-    #[cfg_attr(not(target_arch = "x86"), repr(align(8)))]
     pub struct pthread_cond_t {
-        size: [u8; crate::__SIZEOF_PTHREAD_COND_T],
+        pub _c_head: *mut c_void,
+        pub _c_clock: c_int,
+        pub _c_tail: *mut c_void,
+        pub _c_lock: c_int,
```
tgross35: Ditto

tgross35 (changes requested):
Noticed a few other things, note it also looks like CI is failing still

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5127/df537bcc2ae6743b4b8cb4dfe471ba6f27006266..b0c08966812f000ab2e7c4b60d30c926fcd97e8c)*

tgross35:
> Also, I always refer to the header files shipped under `arch/` in the Fuchsia IDK, which actually match the changes I made.
> 
> I'm wondering which place should we assume as being the definitive reference for all things Fuchsia. I went for the IDK because the CodeSearch site just has a ton of results concerning both the Zircon kernel as well as multiple third party libc implementations (musl, LLVM-libc.)

@erickt would you be able to clear this up a bit? We're trying to figure out what portions of CodeSearch get pulled together to make the SDK that Rust should be supporting.

src/fuchsia/aarch64.rs
dybucc: Why would you have assumed `f64` on AArch64? You quote it as `f128` in [^1], and
the IEEE 754-2008 quad-precision type (mentioned in the AArch64 reference you
provide in that issue) is 128 bits wide [^3].

As per [^2], the alignment requirement in AArch64 should be the same for both
`i128` and `f128`. Considering this is a non-public padding field, I think it's
alright the way it is in this PR.

[^1]: rust-lang/rust#140417
[^2]:
    https://github.com/ARM-software/abi-aa/blob/2024Q3/aapcs64/aapcs64.rst#51fundamental-data-types
[^3]: https://ieeexplore.ieee.org/document/8766229

src/fuchsia/mod.rs
dybucc: Sorry. Done.

src/fuchsia/mod.rs
dybucc: So I'm guessing we should wait and hear back from the target maintainer?
Considering the change is minor, I will hold off from changing the patch
containing the change for now.

src/fuchsia/mod.rs
dybucc: Noted.

src/fuchsia/mod.rs
tgross35: I'd say just drop it, if there's a strong motivation to change things then we can do multiple platforms at once. I'm not really expecting that to be the case, though, given the first point above.

src/fuchsia/mod.rs
dybucc: Done.

src/fuchsia/mod.rs
dybucc: Done.

src/fuchsia/aarch64.rs
tgross35: My list there is pretty incomplete, don't think I even considered Fuchsia - which is why I asked for a link if you had one. `long double` is usually `f128` on AArch64 but that's not a for sure thing - the OS can pick whatever it wants. Windows and all Apple platforms pick `f64` for (as I understand it) cross-platform consistency, which is why I thought Fuchsia may fall into the same boat.

Anyway I assume this is probably fine, a citation that `long double` is `f128` specifically on aarch64-fuchsia would just be nice to have for a few reasons.

dybucc: I couldn't find any information either. Done.

src/fuchsia/aarch64.rs
dybucc: Done.

src/fuchsia/mod.rs
dybucc: Done.

src/fuchsia/mod.rs
dybucc: Done.

src/fuchsia/mod.rs
dybucc: Done.

src/fuchsia/mod.rs
dybucc: Done.

dybucc:
> Noticed a few other things, note it also looks like CI is failing still
>
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5127/df537bcc2ae6743b4b8cb4dfe471ba6f27006266..b0c08966812f000ab2e7c4b60d30c926fcd97e8c)_

CI is failing because there's a type, `epoll_data`, that is triggering a warning
about being used even though it's marked `#[deprecated]`. This is not really
something I should fix in this PR because the reason why that happens is
unrelated to the submitted patchset.

That one type is declared under `s_no_extra_traits`, which in the case of
unions, will add a manual `Debug` impl after declaring the type. If you mark the
type as deprecated, then the impl shouldn't exist either, as otherwise it would
count as using a deprecated item. I've looked into solving this with a little
hack in the macro subtree that would pick up that one attribute out of the
optional attributes that the type can be annotated with, and would annotate the
impl we append with an `#[allow(deprecated)]`. Alternatively, we can just
unconditionally expand the macro to use this `allow` on the impl; It should be
fine, considering it's not an `expect` and thus won't trigger another warning.

@rustbot ready

src/fuchsia/mod.rs
dybucc: Done.

tgross35 (approved):
LGTM aside from the failure, @erickt if you get the change to rereview soon that would be appreciated, otherwise I'll merge in a few days.

> CI is failing because there's a type, `epoll_data`, that is triggering a warning about being used even though it's marked `#[deprecated]`. This is not really something I should fix in this PR because the reason why that happens is unrelated to the submitted patchset.

It's certainly not unrelated if applying the patchset causes CI to fail. It's fine to include a patch adapting for new behavior before a patch that would otherwise break them (or to do that in a separate PR), or, if it's tiny, just change it in the same patch.

@rustbot author since of course merging is blocked on that

> That one type is declared under `s_no_extra_traits`, which in the case of unions, will add a manual `Debug` impl after declaring the type. If you mark the type as deprecated, then the impl shouldn't exist either, as otherwise it would count as using a deprecated item.

Deprecation is a hint to users of `libc`, what we do internally doesn't really matter. We have to use our own deprecated API all the time.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5127/63b34a986c570232352e2aee1ea35afe31229998..69eecbc157caafc5832e1c5d000ed5eb1ec286fd)*

dybucc:
I think it's done now. I went for the easy fix; Adding an `allow(deprecated)` to
all expansions of the `Debug` impl on `union`s in `s_no_extra_traits`.

@rustbot ready

tgross35 (approved):

SnoozeThis:
Resolved.

@rustbot label +S-waiting-on-review. @rustbot release-assignment.

PR #5128, opened 2026-05-31
build: add `rust-toolchain.toml` file

dybucc:
# Description

At present, one is either required to already have the `nightly` toolchain set as default in their
rustup set up, or otherwise configure a manual override with `rustup override set`. This makes it so
that stuff like autoformatting through `rustfmt` fails, as the `.rustfmt.toml` file is configured
with unstable features.

This commit adds a `rust-toolchain.toml` file that makes it so that the user will be reported as
requiring to add the corresponding component when calling, among others, `cargo fmt`, or otherwise
will silently perform the operation by switching to the `nightly` channel.

Because this has some slight consequences to some CI jobs, this patch includes changes there as
well.

Note it may be preferable to also add a list of components to make a single failed `cargo` command
automtically install all components the `libc` crate requires, but I've left that out for the time
being.

## Sources

None

## Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -22,7 +22,7 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
       - name: Install Rust (rustup)
-        run: rustup update stable --no-self-update && rustup default stable
+        run: rustup update stable --no-self-update && rustup override set stable
       - name: Run release-plz
         uses: release-plz/action@064f4d1e36c843611ddf013be726beaa4ad804db # v0.5.129
         env:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -55,7 +55,7 @@ jobs:
       # Here we use the latest stable Rust toolchain already installed by GitHub
       # Ideally we should run it for every target, but we cannot rely on unstable toolchains
       # due to Clippy not being consistent between them.
-      - run: cargo clippy --workspace --exclude libc-test --exclude ctest-test --all-targets -- -D warnings
+      - run: cargo +stable clippy --workspace --exclude libc-test --exclude ctest-test --all-targets -- -D warnings
 
   # This runs `cargo build --target ...` for all T1 and T2 targets`
   verify_build:
@@ -380,7 +380,7 @@ jobs:
     - name: Install Rust
       run: rustup update "$MSRV" --no-self-update && rustup default "$MSRV"
     - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
-    - run: cargo build -p ctest
+    - run: cargo +"$MSRV" build -p ctest
 
   docs:
     name: Ensure docs build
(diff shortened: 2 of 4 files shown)


JohnTitor (changes requested):
This breaks https://github.com/rust-lang/libc/blob/fe7f7b4936d035152d0063d0539ae48d59b4f471/.github/workflows/ci.yaml#L383 and https://github.com/rust-lang/libc/blob/main/.github/workflows/release.yaml, needs a tweak for them.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5128/dbd7ad9a0ef5205226d8541f2a1a7cbdd9da2989..5a8a8e0f2744d3ba3c1b411c02347e2c78c9a360)*

dybucc:
CI is failing for reasons unrelated to the changes introduced in the PR.

@rustbot ready

.github/workflows/release.yaml
```diff
@@ -7,6 +7,7 @@ permissions:
   contents: write
 
 on:
+  workflow_dispatch:
```
JohnTitor: Why did you add this? We don't use that trigger and I don't want to add it as in general it just adds another attack surface.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5128/022b3249ba8f5f31e52bdecde0eac10f34abe5d7..c38f3962b224f7b5d99df50b751a7779994ec383)*

dybucc: That was only part of some manual testing, and I forgot to remove it before pushing to the branch
that's in the PR.

tgross35 (changes requested):
I'd rather not do this; if somebody checks out this repo for the purpose of building, it should work with whatever their default toolchain is. We can usually assume that developers are using nightly, or at least can use +nightly if they see formatting CI fail.

If you (or other developers) usually use stable but this has been getting in your way, you can run `rustup override set nightly`.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5128/022b3249ba8f5f31e52bdecde0eac10f34abe5d7..b3f8d5658eb453df545cd70f39b26f179da50f94)*

dybucc:
The whole point of the patch was to avoid manual intervention, under the assumption that installing
a toolchain is likely possible, lest the developer has no internet connection. Otherwise, setting it
up and possibly having the user tweak those settings if they happen to not have a nightly toolchain
available seemed reasonble. I don't quite understand what's the logical reason not to want this, but
I'm likely missing something. All in all, it's formatting matters, so closing now.

tgross35:
To clarify a bit: a `rust-toolchain` file usually implies that a version is required to do anything with the project, i.e. build it at all. That's not the case here, it only applies to people who are hacking on `libc` (that's what I meant by "developer") and happen to need to run the formatting script, which isn't everybody. And it's a pretty safe assumption that many contributors around rust-lang have nightly as their default, or have an override ready to go.

There are a number of people who just clone this repo to see whether it still builds for their T3 platform and CI bots that pick up changes and do the same, it's reasonable to keep that status quo working.

dybucc:
Now that makes sense. I completely forgot the fact that the `rust-toolchain` file has larger semantic implications. I get it now.

PR #5129, opened 2026-05-31
vxworks: add `cfg` to definition of `off64_t` and `off_t`

dybucc:
# Description

This PR improves the definition of `off_t`. This applies to targets whose operating system is VxWorks.

The `libc` crate defines `off_t` as a `c_longlong`. This is only correct in programs compiled as RTPs. Programs compiled as VxWorks kernels are different. They expose `off_t` as a `c_long`.

The `off64_` type is not present in RTP processes.

Automatic detection of this does not seem possible in Rust. A specific flag is issued to gcc to expose the feature test macro. This patch adds an equivalent `cfg`. The crate user must manually set it.

Two redundant routines were removed. These are explicitly out of the shipped SDK. It seems like the `libc` crate had polyfills set up. These would always return an error value.

## Sources

- Lines 23-33 of file `sysroot/usr/h/public/base/b_off_t.h`. This was found in the VxWorks downloadable SDK for QEMU x86_64.
- File `usr/3pp/develop/usr/include/python3.9/pyconfig.h`. This was found in the same as SDK as quoted above. This defines certain included features. `pread` and `pwrite` are not defined.

## Checklist

### DIFF
--- a/src/new/vxworks/mod.rs
+++ b/src/new/vxworks/mod.rs
@@ -1,4 +1,7 @@
 //! VxWorks libc.
+//!
+//! VxWorks allows compiling different types of programs. The `libc` crate only
+//! supports RTPs. Refer to the RTP definitions in case of discrepancy.
 // FIXME(vxworks): link to headers needed.
 
 pub(crate) mod unistd;
--- a/src/vxworks/mod.rs
+++ b/src/vxworks/mod.rs
@@ -58,7 +58,12 @@ pub type pthread_key_t = c_ulong;
 
 // From b_off_t.h
 pub type off_t = c_longlong;
-pub type off64_t = off_t;
+
+#[deprecated(
+    since = "0.2.187",
+    note = "Kernel mode definitions are unsupported. Use `off_t` instead."
+)]
+pub type off64_t = c_longlong;
 
 // From b_BOOL.h
 pub type BOOL = c_int;
@@ -2412,14 +2417,6 @@ safe_f! {
     pub const fn WSTOPSIG(status: c_int) -> c_int {
         (status >> 16) & 0xFF
     }
-}
-
-pub unsafe fn pread(_fd: c_int, _buf: *mut c_void, _count: size_t, _offset: off64_t) -> ssize_t {
-    -1
-}
-
-pub unsafe fn pwrite(_fd: c_int, _buf: *const c_void, _count: size_t, _offset: off64_t) -> ssize_t {
-    -1
 }
 
 pub unsafe fn posix_memalign(memptr: *mut *mut c_void, align: size_t, size: size_t) -> c_int {


dybucc:

tgross35:
@rustbot blocked

Holding off until there is a good alternative to off64_t

SnoozeThis:
(https://snoozeth.is/Pf0pTEeZDMs) I will wait until Wed, 19 Aug 2026 08:31:52 UTC and then add label S-waiting-on-review and remove label S-blocked.

@rustbot claim.

dybucc:
@tgross35 The PR has been updated. I don't think I can find a better solution. Both `off_t` and `off64_t` are defined in VxWorks. The best I can do is to gate them.

`off_t` will always be defined. Its bit width may be 64-bits in RTPs. Its bit width may be 32-bits or 64-bits in kernel programs. `off64_t` will only be defined in kernel programs. `off64_t` will always be 64-bits wide.

@rustbot ready review

src/vxworks/mod.rs
```diff
@@ -57,8 +57,14 @@ pub type stat64 = crate::stat;
...
-pub type off_t = c_longlong;
-pub type off64_t = off_t;
+cfg_if! {
+    if #[cfg(vxworks_rtp)] {
+        pub type off_t = c_longlong;
+    } else {
+        pub type off_t = c_long;
+        pub type off64_t = c_longlong;
+    }
+}
```
tgross35: Given the existing definitions seem to match rtp, could there be an assumption that vxworks builds are always targeting RTPs rather than kernel? I don't see it documented here https://doc.rust-lang.org/rustc/platform-support/vxworks.html but that's usually a reasonable assumption for most targets.

Cc maintainers @biabbas @hax0kartik

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5129/60297178ac454f730335325c2fa6d37fe93886e2..0b3f97d8efad4768134110d1a366693e3ce193ca)*

hax0kartik:
hi, thanks for the PR but we do not support rust in VxWorks kernel mode. I’m going to reject this change as the platform maintainer.

tgross35:
Ack, thanks for taking a look!

@dybucc this can probably be repurposed as a docs PR to help others in the future. We don't really have anything like this yet but I think the `new` module would be a reasonable place to keep platform-specific docs, https://github.com/rust-lang/libc/blob/main/src/new/vxworks/mod.rs.

@rustbot author
@SnoozeThis /cancel

SnoozeThis:
Snooze cancelled.

@rustbot release-assignment.

dybucc:
> Ack, thanks for taking a look!
> 
> @dybucc this can probably be repurposed as a docs PR to help others in the future. We don't really have anything like this yet but I think the `new` module would be a reasonable place to keep platform-specific docs, https://github.com/rust-lang/libc/blob/main/src/new/vxworks/mod.rs.
> 
> @rustbot author @SnoozeThis /cancel

That's done now. The PR documents that. Kernel symbols are deprecated. This includes `off64_t`. This is not exposed in RTPs.

@rustbot ready

src/vxworks/mod.rs
```diff
@@ -58,7 +58,13 @@ pub type pthread_key_t = c_ulong;
 
 // From b_off_t.h
 pub type off_t = c_longlong;
-pub type off64_t = off_t;
+
+#[deprecated(
+    since = "0.2.187",
+    note = "Use `off_t` instead. This type does not exist. It's only present in VxWorks kernels. \
+            Those are unsupported in Rust."
```
tgross35: Simplify to something like "Kernel-mode definitions are not supported in VxWorks. Use `off_t` instead.". Otherwise LGTM.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5129/00183a7e838b35441481bbb712046b0b124def19..df963e74334c5b7ba598dcee23459e6b2ddc04ca)*

src/new/vxworks/mod.rs
```diff
@@ -1,4 +1,8 @@
 //! VxWorks libc.
+//!
+//! VxWorks allows compiling different types of programs. Exposed bindings are
+//! with respect to RTPs. VxWorks kernel interfaces are not exposed. `off_t` is
+//! an example of this. It's a `c_long` in kernels. It's a `c_longlong` in RTPs.
```
tgross35:
```rust
//! VxWorks allows compiling different types of programs but `libc` only supports
//! RTPs (real-time processes), not kernel interfaces. Bindings should match RTP
//! definitions when there is a discrepancy.
```

Dropping the concrete example

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5129/00183a7e838b35441481bbb712046b0b124def19..df963e74334c5b7ba598dcee23459e6b2ddc04ca)*

src/vxworks/mod.rs
dybucc: Done.

src/new/vxworks/mod.rs
dybucc: Done.

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5129/f08bfce345c586260a5678dc4442b3e9de856382..d2f6329f587de23a8945e174a606065bd7c4e8d9)*

PR #5131, opened 2026-06-01
refactor: adjust definition of `off_t` in wasi

dybucc:
# Description

This PR addresses a small misconfiguration in the way we handle bit-widths of the `off_t` type in WASI. The definition for the analogous `time_t` type fits exactly that of the upstream repo; Namely, a `c_longlong`. This is not the case with the former, as upstream defines it with the same type as `time_t`, but we expose it in the `libc` crate as a fixed-width signed integer.

This is not so much a problem as it is an inconsistency, because WASI assumes LLP64 by their use of `c_longlong`, and so the effective bit-width of a type defined as a `c_longlong` and a type defined as an `i64` is the same. Such change could be considered irrelevant, so assume the patch in this PR is solely a proposal and not a required change.

## Sources

- WASI libc sources show how `off_t` is defined as a `c_longlong`.

  > <https://github.com/WebAssembly/wasi-libc/blob/04bd527eb37512303f83efc01e4fb6222392849e/libc-bottom-half/headers/public/__typedef_off_t.h#L4>

## Checklist


dybucc:
CI seems to be failing for reasons unrelated to the PR. Triggering a rerun should do it.

dybucc:

tgross35 (approved):
Makes sense, thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5131/f4d67773492709f79fcbd4b6f60dd26778291d87..8cc81a5454dddcbd9ea326663943b4f10ebc6bcf)*

PR #5132, opened 2026-06-01
newlib: fix definition of `time_t` and `off_t`

dybucc:
# Description

This PR addresses bit width mismatches for both `time_t` and file offset types. This applies to the `linux/newlib` module.

Upstream and forks for each supported target provide conditional definitions. `time_t` can be configured in the newlib build script. File offset types require manual intervention in a specific header file. The file containing the default definition will check for a macro that is defined in that specific "override" file. The default definition is a C `long`.

A number of supported targets were using either one of the above overrides. This was not being taken into account in the `libc` crate. This patch fixes that by further expanding conditionally compiled type definitions.

## Sources

- Cygwin's, espressif's and the Vita SDK's newlib forks showing conditional definition of the `time_t` type and the `off_t` type. Note Vita SDK's fork contains definitions relative to multiple target systems, but the one under `include/sys` has taken priority. The definition under `include/vita/sys` `#include`s the former to define its `off_t` type.

  > <https://github.com/cygwin/cygwin/blob/6926523688bab730ae1b56f24ad6fb203b7350af/newlib/libc/include/sys/_types.h#L185>
  > <https://github.com/espressif/newlib-esp32/blob/esp-4.5.0/newlib/libc/include/sys/_types.h#L185>
  > <https://github.com/search?q=repo%3Avitasdk%2Fnewlib+%2F%5Cb_time_t_%5Cb%2F&type=code>
  > <https://github.com/cygwin/cygwin/blob/6926523688bab730ae1b56f24ad6fb203b7350af/newlib/libc/include/sys/_types.h#L100>
  > <https://github.com/espressif/newlib-esp32/blob/67537d91e601c40e2b52b9b865518abb7eecc85b/newlib/libc/include/sys/_types.h#L100>
  > <https://github.com/vitasdk/newlib/blob/fbb8375870d799758e0b4358d2f1708781aa26a6/newlib/libc/include/sys/_types.h#L100>

  Sources on devkitARM could not be found online, but they can be found under the installed toolchain's directory in `devkitARM/` by running `ripgrep` with the following command.

  > ```sh
  > rg -C 5 -S -uu -g "**/*.h" -e "\b_time_t_\b"
  > ```

- Newlib sources documenting the `configure` build script non-default option for using a `c_long` for their `time_t` definition.

  > <https://github.com/cygwin/cygwin/blob/6926523688bab730ae1b56f24ad6fb203b7350af/newlib/README#L351>
- Cygwin upstream codesearch showing the definition of `_off_t`, which defines `__off_t`, and eventually `off_t`. Note how this is defined in `machine/_types.h` (as an `i64`,) and later that is checked in `sys/_types.h` (falling back to a `c_long`.)

  > <https://github.com/search?q=repo%3Acygwin%2Fcygwin%20symbol%3A%2F%5Cb_*off_t%5Cb%2F&type=code>
- espressif's `esp-32` fork of newlib, `vita`'s and RTEMS' showing how they don't have a definition in terms an `i64` under `machine/_types.h`.

  > <https://github.com/search?q=repo%3Aespressif%2Fnewlib-esp32+path%3Anewlib%2Flibc%2Finclude+%2F%5Cb_*off_t%5Cb%2F&type=code>
  > <https://github.com/search?q=repo%3Avitasdk%2Fnewlib+path%3Anewlib%2Flibc%2Finclude+%2F%5Cb_*off_t%5Cb%2F&type=code>
  > <https://gitlab.rtems.org/search?search=path%3Anewlib%2Flibc%2Finclude+extension%3Ah+%22_*off_t%22&nav_source=navbar&project_id=416&group_id=1086&search_code=true&repository_ref=main>

## Checklist


dybucc:
CI seems to be failing for reasons unrelated to the changes introduced in the patch. A rerun should
do it.

dybucc:

tgross35:
I couldn't remember what the possible env+os combinations were for these platforms, here's a quick list.

```text
> $cfg | find 'target_env="newlib"' | get tgt | each {|tgt| $cfg | find $tgt -n | find -nr '(target_env|target_os)' } | flatten
  #               tgt                        cfg
─────────────────────────────────────────────────────────
  0   armv6k-nintendo-3ds            target_env="newlib"
  1   armv6k-nintendo-3ds            target_os="horizon"
  2   armv7-rtems-eabihf             target_env="newlib"
  3   armv7-rtems-eabihf             target_os="rtems"
  4   armv7-sony-vita-newlibeabihf   target_env="newlib"
  5   armv7-sony-vita-newlibeabihf   target_os="vita"
  6   riscv32imac-esp-espidf         target_env="newlib"
  7   riscv32imac-esp-espidf         target_os="espidf"
  8   riscv32imafc-esp-espidf        target_env="newlib"
  9   riscv32imafc-esp-espidf        target_os="espidf"
 10   riscv32imc-esp-espidf          target_env="newlib"
 11   riscv32imc-esp-espidf          target_os="espidf"
 12   xtensa-esp32-espidf            target_env="newlib"
 13   xtensa-esp32-espidf            target_os="espidf"
 14   xtensa-esp32s2-espidf          target_env="newlib"
 15   xtensa-esp32s2-espidf          target_os="espidf"
 16   xtensa-esp32s3-espidf          target_env="newlib"
 17   xtensa-esp32s3-espidf          target_os="espidf"
```

tgross35 (commented):
I'm having some difficulty extracting a simplified goal from the top post. Is this the net delta on targets we currently support?

* On armv7-rtems, `time_t` changes from i32 to i64
* On armv7-rtems, `off_t` changes from i64 to `c_long`, which should be i64->i32
* On armv7-vita, `off_t` changes from `c_int` to `c_long`. I think there is no change here?

If this sounds accurate, ping the target maintainer (see https://doc.rust-lang.org/rustc/platform-support/armv7-rtems-eabihf.html) for confirmation. This is generally a good idea when tricky things like this change on less popular targets, there may be config or other nuance that doesn't show up in the source code.

Also please don't call this a refactor, that term usually means cleaning up code without changing behavior - if the user experience changes somehow then it's kind of misleading. (Applies to some of your other PRs as well).

> Note this option is "documented" as being meant for end users of the libc crate to enable, but no information on it is provided anywhere in all of the libc crate.

It's mostly intended to be set by espidf build tooling, but it would be good to at least mention it. A PR would be welcome in the crate-level docs, after https://github.com/rust-lang/libc/pull/5179 adds some structure.

> It would also be great to know whether the libc crate considers horizon to be the OS for both the Nintendo Switch and the Nintendo 3DS targets, or only for the Nintendo 3DS target. 

`aarch64-nintendo-switch-freestanding` does not appear to set a `target_env`, so I would assume the latter.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5132/1a9bdc4d16edc21dfa66e655d1edf8e4eac16ba8..6abed2e3fb19ef683287f4ae1cbddba634ce0d16)*

dybucc:
The changes here are required to fit the definitions upstream of `time_t`, which was generally mismatched.

The changes for file offset types are concerned with either

1. having a mirrored definition, rather an an equivalent one, and
2. with the same bit width mismatches as with `time_t`.

I'll open a PR soon with the documentation for ESP-IDF.

@rustbot ready

dybucc:
Hey @thesummer!

I was wondering if I could get confirmation on these changes on the RTEMS target you maintain?

dybucc:
Hey @nikarh @pheki @zetanumbers!

I was wondering if I could get confirmation on these changes on the Sony PS Vita target you maintain?

src/unix/newlib/mod.rs
```diff
@@ -6,14 +6,15 @@ pub type blksize_t = i32;
 pub type clockid_t = c_ulong;
 
 cfg_if! {
-    if #[cfg(any(target_os = "espidf"))] {
+    if #[cfg(any(
+        target_os = "espidf",
+        target_os = "vita",
+        target_os = "rtems",
+        target_arch = "aarch64"
+    ))] {
```
pheki: For `vita`, this one seems correct, as by looking at the definitions it seems to be defined as a long. There should also be no change in behavior as both `c_int` and `c_long` are aliases to `i32` in this platform.

Sources in the top level review comment.

src/unix/newlib/mod.rs
```diff
@@ -54,13 +55,10 @@ cfg_if! {
...
-    if #[cfg(any(
-        target_os = "horizon",
-        all(target_os = "espidf", not(espidf_time32))
-    ))] {
-        pub type time_t = c_longlong;
-    } else {
+    if #[cfg(all(target_os = "espidf", espidf_time32))] {
         pub type time_t = i32;
+    } else {
+        pub type time_t = i64;
```
pheki, line 61: This one looks wrong for the vita, which uses 32-bit time, as `--enable-newlib-long-time_t` is set when building newlib in vitasdk (currently the only supported SDK).

The correct definition would be `c_long` (effectively `i32`).

Sources in the top-level comment.

pheki (changes requested):
Thanks for asking!

Just need a change as the Vita uses 32-bit `time_t`.

Keeping it on the top-level comment for easy reference, here are the sources I could find (somewhat redundant with the finds in the root PR comment but in vitasdk's fork):

### `off_t`

- [`_off_t` is defined](https://github.com/vitasdk/newlib/blob/fbb8375870d799758e0b4358d2f1708781aa26a6/newlib/libc/include/sys/_types.h#L45-L47)
- [`__off_t` is defined](https://github.com/vitasdk/newlib/blob/fbb8375870d799758e0b4358d2f1708781aa26a6/newlib/libc/include/sys/_types.h#L99-L103)
- [`off_t` is defined](https://github.com/vitasdk/newlib/blob/fbb8375870d799758e0b4358d2f1708781aa26a6/newlib/libc/sys/vita/sys/types.h#L160-L163)

### `time_t`

- [`--enable-newlib-long-time_t` is set in CMakeLists.txt](https://github.com/vitasdk/buildscripts/blob/366f4140fb89193c55b2e115652da7dc28e3f7b4/CMakeLists.txt#L584)
- In configure:
  - [`newlib_long_time_t` is set](https://github.com/vitasdk/newlib/blob/fbb8375870d799758e0b4358d2f1708781aa26a6/newlib/configure#L2522-L2533)
  - [`_WANT_USE_LONG_TIME_T` is set](https://github.com/vitasdk/newlib/blob/fbb8375870d799758e0b4358d2f1708781aa26a6/newlib/configure#L12554-L12559)
- In configure.in:
  - [`newlib_long_time_t` is set](https://github.com/vitasdk/newlib/blob/fbb8375870d799758e0b4358d2f1708781aa26a6/newlib/configure.in#L251-L260)
  - [`_WANT_USE_LONG_TIME_T` is set](https://github.com/vitasdk/newlib/blob/fbb8375870d799758e0b4358d2f1708781aa26a6/newlib/configure.in#L514-L516)
- [`_TIME_T_` is set](https://github.com/vitasdk/newlib/blob/fbb8375870d799758e0b4358d2f1708781aa26a6/newlib/libc/include/sys/_types.h#L184-L189)
- [`time_t` is set](https://github.com/vitasdk/newlib/blob/fbb8375870d799758e0b4358d2f1708781aa26a6/newlib/libc/sys/vita/sys/types.h#L116-L120)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5132/204104b3e960cdf2490bb521e1d0f52267fe847d..c2b097039427763c72c581e6b1028a7732659ac5)*

tgross35:
> The changes here are required to fit the definitions upstream of `time_t`, which was generally mismatched.
>
> The changes for file offset types are concerned with either
> 
>     1. having a mirrored definition, rather an an equivalent one, and
>     2. with the same bit width mismatches as with `time_t`.

Indeed libc doesn't match every source, but with every change I need to figure out a succinct description of what (1) is changing to look more similar to source without affecting users, vs. (2) is a behavior change of some form. IOW I'm interested in the intent more than the problem list, to make sure I'm on the same page as PR authors.

To help with that, are the summary bullet points at https://github.com/rust-lang/libc/pull/5132#pullrequestreview-4531240926 accurate?

src/unix/newlib/mod.rs
dybucc: Thanks. That's fixed now.

dybucc:
Just updated the PR description. It's hopefully clearer now.

The extent of the changes has now changed.
- RTEMS is unaffected.
- The changes on `vita` are as you outlined.
- AArch64 had changes to its file offset types.
- `horizon` has a gone from 32-bit to 64-bit `time_t`.

pheki:
I may be misunderstanding or missing something here, but I have some questions. First regarding this:

> having a mirrored definition, rather an an equivalent one, and

You kept the vita with `pub type time_t = i32;`. If the idea is to have a mirrored definition, shouldn't it be `c_long`?

> `horizon` has a gone from 32-bit to 64-bit `time_t`.

Wasn't it already 64-bit, as `c_longlong` is an alias to `i64` on all supported platforms?

> RTEMS is unaffected.
> AArch64 had changes to its file offset types.

It seems like `time_t` has changed from `i32` to `i64` for those, no?

thesummer:
> Hey @thesummer!
> 
> I was wondering if I could get confirmation on these changes on the RTEMS target you maintain?

I just checked on the hardware with a compiled binary and both `off_t` and `time_t` are 8 bytes long.
So, to me the current state of the PR looks correct. Both times the RTEMS configuration should fall to the `else` block which sets the type to `i64`, right?

tgross35:
> AArch64 had changes to its file offset types.

Which targets does this hit? I didn't see any from my quick search.

And then, espidf targets should also be unchanged right?

@pheki has a few good questions about the rest.

dybucc:
> I may be misunderstanding or missing something here, but I have some questions. First regarding this:
> 
> > having a mirrored definition, rather an an equivalent one, and
> 
> You kept the vita with `pub type time_t = i32;`. If the idea is to have a mirrored definition, shouldn't it be `c_long`?
> 
> > `horizon` has a gone from 32-bit to 64-bit `time_t`.
> 
> Wasn't it already 64-bit, as `c_longlong` is an alias to `i64` on all supported platforms?
> 
> > RTEMS is unaffected.
> > AArch64 had changes to its file offset types.
> 
> It seems like `time_t` has changed from `i32` to `i64` for those, no?

Each of your queries is an item.

1. You're right. I missed that. I also missed it in ESP-IDF.
2. You're right. Changes were logged wrongly. They referred to the prior state of the branch. They did not refer to `main`.
3. Ibid.

dybucc:
> > AArch64 had changes to its file offset types.
> 
> Which targets does this hit? I didn't see any from my quick search.
> 
> And then, espidf targets should also be unchanged right?
> 
> @pheki has a few good questions about the rest.

AArch64 has changed. The changes are as follows:

- `dev_t`: `u32` -> `c_short`.
- `ino_t`: `u32` -> `c_ushort`.
- `off_t`: `i64` -> `c_long`. Not quite a change.

ESP-IDF is unchanged.

dybucc:
> > Hey @thesummer!
> > I was wondering if I could get confirmation on these changes on the RTEMS target you maintain?
> 
> I just checked on the hardware with a compiled binary and both `off_t` and `time_t` are 8 bytes long. So, to me the current state of the PR looks correct. Both times the RTEMS configuration should fall to the `else` block which sets the type to `i64`, right?

Thanks. That's right.

tgross35:
> AArch64 has changed. The changes are as follows:

Target, not arch - we don't seem to have any newlib aarch64 targets, unless I'm missing something?

dybucc:
You're right. We don't. I confused the devkitA64 SDK. Rust support on the `aarch64-nintendo-switch-freestanding` target is not implemented in terms of that.

No AArch64 targets are affected.

src/unix/newlib/mod.rs
```diff
@@ -6,14 +6,14 @@ pub type blksize_t = i32;
 pub type clockid_t = c_ulong;
 
 cfg_if! {
-    if #[cfg(any(target_os = "espidf"))] {
+    if #[cfg(any(
+        target_os = "espidf",
+        target_os = "vita",
+        target_arch = "aarch64"
```
tgross35: Could you drop aarch64 if not needed?

src/unix/newlib/mod.rs
tgross35, line 19: For rtems, where do these definitions come from? It matches what @thesummer tested, but per https://gitlab.rtems.org/contrib/newlib-cygwin/-/blob/282d57d2a8462c56ceeafcc10095cab35343eb09/newlib/libc/include/sys/_types.h#L46, https://gitlab.rtems.org/contrib/newlib-cygwin/-/blob/282d57d2a8462c56ceeafcc10095cab35343eb09/newlib/libc/include/sys/_types.h#L56, https://gitlab.rtems.org/contrib/newlib-cygwin/-/blob/282d57d2a8462c56ceeafcc10095cab35343eb09/newlib/libc/include/sys/_types.h#L75 it looks `dev_t = c_short` (i16), `ino_t = c_ushort` (u16), `off_t = c_long` (i32)

@thesummer may know more.

tgross35 (commented):
I think these are fine but have a question about the rtems source. This has gone through a few changes, could you post an updated summary of what changed on horizon/vita/rtems/espidf?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5132/69c52f243cf5497e90a59875a4443171c782edfc..d00f11fdc04d147f919d997b724c03224e92eed5)*

src/unix/newlib/mod.rs
dybucc: Notice that the definitions in those files exist only when `__machine_{type}_t_defined` is not defined. This means definitions under `newlib/libc/sys/_types.h` are merely fallback definitions. The actual definitions in RTEMS are in `newlib/libc/sys/rtems/include/machine/_types.h`.

- `off_t`: https://github.com/RTEMS/sourceware-mirror-newlib-cygwin/blob/854594be786023d7e92c9352d8f54e64f9fb36eb/newlib/libc/sys/rtems/include/machine/_types.h#L15
- `dev_t`: https://github.com/RTEMS/sourceware-mirror-newlib-cygwin/blob/854594be786023d7e92c9352d8f54e64f9fb36eb/newlib/libc/sys/rtems/include/machine/_types.h#L12
- `ino_t`: https://github.com/RTEMS/sourceware-mirror-newlib-cygwin/blob/854594be786023d7e92c9352d8f54e64f9fb36eb/newlib/libc/sys/rtems/include/machine/_types.h#L21

src/unix/newlib/mod.rs
dybucc: Sure. I also removed the `aarch64` module.

src/unix/newlib/mod.rs
tgross35: Ack, thanks for clarifying.

Since these specific types are RTEMS-specific, I think it would be good to restrict this to `target_os = "rtems"` and not have a fallback. Which means any new OS using newlib won't build until they add these definitions, but that's a good thing.

dybucc:
An updated summary of the changes follows.

- `vita` went from `i32` for `time_t` to `c_long`.
- `vita` went from `off_t` as a `c_int` to `off_t` as a `c_long`.
- AArch64 bindings got removed. There's no officially supported Rust target with that triple combination.

src/unix/newlib/mod.rs
dybucc: Except they're not. There's other targets that also use those. An example is `horizon`. See `arm-non-eabi/include/machine/_types.h`.

Do you mean to have multiple `cfg_if` blocks with the same definitions but a different predicate?

I can't link to sources. They're part of the downloadable SDK.

tgross35: RTEMS and Horizon in that case - I just mean to exhaustively list the targets we know about, and let everything else error out or even `compile_error!`.

dybucc: I think it's done. I also noticed builds were failing. The `newlib/generic` module items were not being reexported.

src/unix/newlib/mod.rs
```diff
@@ -6,18 +6,21 @@ pub type blksize_t = i32;
...
+        target_os = "rtems",
+        target_os = "horizon",
+        target_os = "arm",
+        target_os = "powerpc"
+    ))] {
         pub type dev_t = u32;
         pub type ino_t = u32;
         pub type off_t = i64;
+    } else {
+        std::compile_error! { "unsupported target" }
```
tgross35:
```suggestion
        core::compile_error!("unsupported target");
```

Needs to work with no-std, and syntax nit

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5132/076d48ffbddf9408319f72ef7361452f8b0206d1..3a97ab32d238194f7a4adb91a6a6d1ce61903193)*

src/unix/newlib/mod.rs
```diff
@@ -6,18 +6,21 @@ pub type blksize_t = i32;
...
-    } else if #[cfg(any(target_os = "vita"))] {
-        pub type dev_t = c_short;
-        pub type ino_t = c_ushort;
-        pub type off_t = c_int;
-    } else {
+    } else if #[cfg(any(
+        target_os = "rtems",
+        target_os = "horizon",
+        target_os = "arm",
+        target_os = "powerpc"
```
tgross35: These have `target_os`, but I don't think we need arch-based config anyway?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5132/076d48ffbddf9408319f72ef7361452f8b0206d1..3a97ab32d238194f7a4adb91a6a6d1ce61903193)*

tgross35:
> An updated summary of the changes follows.
> 
>     * `vita` went from `i32` for `time_t` to `c_long`.
>     * `vita` went from `off_t` as a `c_int` to `off_t` as a `c_long`.
>     * AArch64 bindings got removed. There's no officially supported Rust target with that triple combination.

Isn't there a change to rtems here? And doesn't Horizon also change `time_t` alias but not its size, similar to vita?

I'm admittedly getting lost with the cycles this has gone through. Here's a table of what I understand is happening currently:

OS | ty | old | new | effective change
---|---|---|---|---
risc32/xtensa32-espidf | off_t | long | long | none
arm32-horizon | off_t | i64 | i64 | none
arm32-rtems | off_t | i64 | i64 | none
arm32-vita | off_t | int | long | alias only, no size change
risc32/xtensa32-espidf time32 | time_t | i32 | long | alias only, no size change
risc32/xtensa32-espidf not(time32) | time_t | long long | i64 | alias only, no size change
arm32-horizon | time_t | long long | i64 | alias only, no size change
arm32-rtems | time_t | i32 | i64 | ABI change, 32->64 bits
arm32-vita | time_t | i32 | long | alias only, no size change

I think that's all reasonable but if it's not accurate, please post an updated table.

tgross35 (changes requested):
(for clarification, CI fixes, and squashing the last two commits)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5132/076d48ffbddf9408319f72ef7361452f8b0206d1..3a97ab32d238194f7a4adb91a6a6d1ce61903193)*

src/unix/newlib/mod.rs
dybucc: yup, both rtems and horizon are already satisfied by the arm predicate. done.

src/unix/newlib/mod.rs
dybucc: sorry. done.

dybucc:
that's it. before i removed the powerpc bindings, that one would have also been affected by the time_t change. but from grepping through the target_env cfg of all supported rust powerpc targets, none seem to use newlib.

in terms of changes affecting something other than the bit width of some types, both the aarch64 bindings and the powerpc bindings were removed.

@rusbot ready

src/unix/newlib/mod.rs
```diff
@@ -950,19 +950,10 @@ cfg_if! {
...
-}
-
-cfg_if! {
-    if #[cfg(target_os = "rtems")] {
+    } else if #[cfg(target_os = "rtems")] {
         mod rtems;
         pub use self::rtems::*;
+    } else {
+        core::compile_error!("unsupported target");
     }
```
tgross35, line 982: "newlib: remove powerpc bindings": After this patch, when is the `rtems` arm ever being hit?

Previously `rtems` was created and imported only if `target_os = "rtems"`, which makes sense. Now it needs to not hit `target_arch = "arm"` - but our only rtems platform is arm?

The pattern is preexisting but mixing arch and OS in the same `cfg_if` is kind of weird.

tgross35 (changes requested):
Couple more things. Could you also split the table from https://github.com/rust-lang/libc/pull/5132#issuecomment-4880519018 into two and paste it into the "newlib: change definition of time_t" and "newlib: change definition of file offset types"? That's easier to understand than something text-y.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5132/67c9a34fda91bc26a50cd907b1e8a65ee32bfb65..4c2721caa50a02f7499e6689f93f8e9e43aa20a8)*

physwkim:
Measured RTEMS data below, but first a problem with the module restructure that
I think blocks this as written.

## The `cfg_if` merge makes `mod rtems` unreachable

On `main` today there are **two** `cfg_if` blocks at the end of
`src/unix/newlib/mod.rs`: one selecting the arch module (`mod arm`), and a
separate one that glob-imports `mod rtems` **on top** for
`target_os = "rtems"`. This PR merges them into one chain and places the RTEMS
arm *after* the arch arm:

```rust
} else if #[cfg(target_arch = "arm")] {
    mod arm;
    pub use self::arm::*;
} else if #[cfg(target_os = "rtems")] {   // unreachable for armv7-rtems-eabihf
    mod rtems;
    pub use self::rtems::*;
} else {
    core::compile_error!("unsupported target");
}
```

`armv7-rtems-eabihf` is **both** `target_arch = "arm"` and
`target_os = "rtems"` (`rustc --print cfg --target armv7-rtems-eabihf`), so it
takes the `arm` arm and `mod rtems` is never imported. Everything in
`src/unix/newlib/rtems/mod.rs` — `sockaddr_un`, `AF_UNIX`, `RTLD_DEFAULT`, the
whole signal set, `pthread_create`, `pthread_condattr_setclock`, `getentropy`,
`arc4random_buf`, `setgroups`, the `W*` helpers — silently disappears.

Reproduced against this PR applied to current `main`, with a one-file crate
that only names two of those items:

```
error[E0425]: cannot find value `AF_UNIX` in crate `libc`
error[E0425]: cannot find type `sockaddr_un` in crate `libc`
```

The same crate builds on `main` unmodified. Putting the `target_os = "rtems"`
arm *before* the `target_arch` arms fixes it, but note that RTEMS genuinely
needs *both* modules — that is why the second `cfg_if` glob-imports rather than
selecting — so an `else if` chain may not be the right shape here at all.

## Measured type data

Compiled and **run** on `armv7-rtems-eabihf` (RTEMS 6.0.0, gcc 13.3.0, newlib
`1b3dcfd`, `xilinx_zynq_a9_qemu` under `qemu-system-arm`), from
`<sys/types.h>` alone:

```
TC time_t     size=8 align=8 signed=1      TC clock_t    size=8 align=8 signed=0
TC dev_t      size=8 align=8 signed=0      TC clockid_t  size=4 align=4 signed=1
TC ino_t      size=8 align=8 signed=0      TC rlim_t     size=8 align=8 signed=1
```

- **`time_t = i64` is right for RTEMS** — this PR's change is correct there,
  and `off_t = i64` for arm is already correct (measured 8).
- **`dev_t` and `ino_t` are 8 bytes on RTEMS, not 4.** This PR moves them into
  an `any(target_arch = "arm", target_arch = "powerpc")` arm that keeps `u32`,
  so RTEMS still gets the wrong width.
- `rlim_t` (8, signed), `clock_t` (8, unsigned — `libc` has `c_long` via
  `newlib/arm`) and `clockid_t` (signed — `libc` has `c_ulong`) are also wrong
  for RTEMS and untouched here.

Why `time_t` is worth landing sooner: at `i32`, `libc::timespec` is 8 bytes
where RTEMS's is 16, so `std`'s `SystemTime::now`/`Instant::now` hand
`clock_gettime` an 8-byte `MaybeUninit` slot and the kernel writes 12 bytes into
it — 4 past the end on every clock read, with a canary 8 bytes out clobbered.
`tv_nsec` then lands where the 8-byte view never reads it:

```
std SystemTime[0] secs=567993600 subsec_nanos=0
std SystemTime[4] secs=567993600 subsec_nanos=0     <- several ms later
std Instant elapsed secs=0 subsec_nanos=0
```

Every sub-second duration on RTEMS silently measures zero — timeouts, rate
limits and watchdogs all read 0, with no panic and no diagnostic.

I have opened #5308 for the six scalar types and #5307 for a separate RTEMS
socket-address defect. #5308 overlaps this PR on `time_t`; happy to rebase it
down to the five types this one does not cover if this lands first — flagging
so the two are not merged blind to each other. Whatever lands needs to reach
`libc-0.2` to change anything for `std`, which depends on `0.2.x`.

tgross35:
@physwkim did you see my comment at https://github.com/rust-lang/libc/pull/5132#discussion_r3622267705 about the fallthrough?

If you have a reason to believe this PR is incorrect, please leave a succinct handwritten review comment. _Do not_ post AI-generated blobs to communicate to humans.

physwkim:
> @physwkim did you see my comment at [#5132 (comment)](https://github.com/rust-lang/libc/pull/5132#discussion_r3622267705) about the fallthrough?
> 
> If you have a reason to believe this PR is incorrect, please leave a succinct handwritten review comment. _Do not_ post AI-generated blobs to communicate to humans.

Sorry, I missed your earlier comment about the fallthrough - yes, same issue.

src/unix/newlib/mod.rs
dybucc: Fixed. Back when I finished the patch, I thought RTEMS targets wouldn't even
build because they'd fall back to the branch where the dummy
`target_arch_not_implemented` module is. I failed to realize it goes into the
`arm` branch beforehand, and so I transitevely failed to realize adding RTEMS to
that `cfg_if` invocation would skip over the target OS' module.

tgross35: Can't remember whether I told you already or not, but I've added a script `./etc/libc-util.py check-all-targets libc` that will make sure all targets build. It takes a little while and like 60GB in `~/cache` (to avoid toolchain pinning), though you can restrict with `--only`/`--skip`.

Anyway I run that with some frequency and as far as I'm aware, the only nonbuilding targets are listed at https://github.com/rust-lang/libc/blob/5db5914d7ecf8d3fe32394d315f29012039630ca/etc/libc-util.py#L343-L361.

dybucc:
> Couple more things. Could you also split the table from https://github.com/rust-lang/libc/pull/5132#issuecomment-4880519018 into two and paste it into the "newlib: change definition of time_t" and "newlib: change definition of file offset types"? That's easier to understand than something text-y.

Done.

@rustbot ready

src/unix/newlib/mod.rs
dybucc: Noted. I noticed there were new utilities while rebasing PRs, but I hadn't taken
a look at the scripts. Thanks.

src/unix/newlib/mod.rs
tgross35, line 1: There's a typo in the message for 0bb0085bfbb02359023b912da468bbcceb71394c ("newlib: change definition of file offset types"), ino_t was never `i64`. Not going to block though

tgross35 (approved):
Looks great, thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5132/081721d0d0ce0f4c38dab8acbc65494d3a0ccb8e..8b8665b6927caeaa89f6ced2c1fde61d91241917)*

src/unix/newlib/mod.rs
dybucc: True. Fixed.

PR #5141, opened 2026-06-04
Fix constants accidentally removed for Linux uClibc targets

### LINKED ISSUE #5140: IPV6_TRANSPARENT is not defined for uClibc targets
The `IPV6_TRANSPARENT` constant at `src/unix/linux_like/mod.rs` is defined only for non-uClibc targets. However, this constant has nothing to do with uClibc itself. The constant is available via `linux/in6.h` which comes from Linux headers, so it is available even on uClibc based systems.

I've also checked all other `IP_xxx` and `IPV6_xxx` constants defined in the same file and behind `#[cfg(not(target_env = "uclibc"))]`. They all come from either `linux/in.h` or `linux/in6.h`.

I've double-checked the constants aren't gated behind some `#ifdef` using a short C code compiled using a Buildroot based cross-complier with uClibc 1.0.45. All the constants are available.

This breaks the `socket2` crate builds for uClibc targets starting from `socket2` version `0.6.3` which depends on `IPV6_TRANSPARENT`.

I propose removing the `cfg_if!` block and `#[cfg(not(target_env = "uclibc"))]` for all `IP_xxx` and `IPV6_xxx` constants defined in `src/unix/linux_like/mod.rs`.

operutka:
# Description

This PR fixes constants that were, probably accidentally, removed by commit 2fe1d91 for uClibc in all operating systems. However, based on the commit message, the commit was supposed to fix things for L4Re, not break things for other operating systems.

So this PR limits the uClibc exclusions introduced in 2fe1d91 to apply only if the target OS is L4Re. This also fixes rust-lang/libc#5140.

# Sources

I believe no links to sources are necessary here. The affected constants were previously available for all linux-like operating systems and environments. This PR simply reverts things to the previous state for all operating systems except L4Re.

# Checklist

# Notes

There was nothing to update in `libc-test/semver`. The constants are already present in `linux.txt`.

I'm unable to run the tests locally though. After building custom `rustc` and `rust-std` for `armv7-unknown-linux-uclibceabihf`, I'm able to initiate the build of `libc-test` but it fails because the following header files are missing in my case:

* `linux/can/j1939.h`
* `linux/mount.h`
* `linux/openat2.h`

I guess they require a more up to date Linux headers then I currently have in my toolchain. When I comment out the header files, I get a ton of missing symbol errors. I'll try to re-run the test with a more recent toolchain.

### DIFF
--- a/src/unix/linux_like/mod.rs
+++ b/src/unix/linux_like/mod.rs
@@ -805,11 +805,11 @@ pub const IP_TRANSPARENT: c_int = 19;
 pub const IP_ORIGDSTADDR: c_int = 20;
 pub const IP_RECVORIGDSTADDR: c_int = IP_ORIGDSTADDR;
 pub const IP_MINTTL: c_int = 21;
-#[cfg(not(target_env = "uclibc"))]
+#[cfg(not(all(target_os = "l4re", target_env = "uclibc")))]
 pub const IP_NODEFRAG: c_int = 22;
-#[cfg(not(target_env = "uclibc"))]
+#[cfg(not(all(target_os = "l4re", target_env = "uclibc")))]
 pub const IP_CHECKSUM: c_int = 23;
-#[cfg(not(target_env = "uclibc"))]
+#[cfg(not(all(target_os = "l4re", target_env = "uclibc")))]
 pub const IP_BIND_ADDRESS_NO_PORT: c_int = 24;
 pub const IP_MULTICAST_IF: c_int = 32;
 pub const IP_MULTICAST_TTL: c_int = 33;
@@ -831,9 +831,9 @@ pub const IP_PMTUDISC_DONT: c_int = 0;
 pub const IP_PMTUDISC_WANT: c_int = 1;
 pub const IP_PMTUDISC_DO: c_int = 2;
 pub const IP_PMTUDISC_PROBE: c_int = 3;
-#[cfg(not(target_env = "uclibc"))]
+#[cfg(not(all(target_os = "l4re", target_env = "uclibc")))]
 pub const IP_PMTUDISC_INTERFACE: c_int = 4;
-#[cfg(not(target_env = "uclibc"))]
+#[cfg(not(all(target_os = "l4re", target_env = "uclibc")))]
 pub const IP_PMTUDISC_OMIT: c_int = 5;
 
 // IPPROTO_IP defined in src/unix/mod.rs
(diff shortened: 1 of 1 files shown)


operutka:
I've got an update related to the local tests. I've generated a new toolchain using Buildroot with up to date kernel headers but I'm still unable to build the `libc-test`. There is a ton of errors unrelated to the constants that I've made available again through this PR. The error log is attached.

Given that, I believe the tests never worked on the Linux uClibc target in the first place and, in fact, the `libc-test` does not build even if I revert my commit.

I think fixing the tests for uClibc is a whole different job that's well outside of scope of this PR.

[build-error.log](https://github.com/user-attachments/files/28634319/build-error.log)

JohnTitor (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5141/dbd7ad9a0ef5205226d8541f2a1a7cbdd9da2989..56476a2c6edc6b18aa45e8aafb88569b5caf60d4)*

skrap:
I had already fixed these errors and more in https://github.com/rust-lang/libc/pull/5046, but it has been waiting for a review for 2 months.  Did I mess up the process somehow, or is there something else wrong with my PR?

operutka:
@skrap I'm sorry about the interference. I didn't notice your PR :) I can see you've put a lot of effort there. But I'm not a maintainer, so I can't really answer your question.

skrap:
> @skrap I'm sorry about the interference. I didn't notice your PR :) I can see you've put a lot of effort there. But I'm not a maintainer, so I can't really answer your question.

Ah yes, of course - sorry, I meant the question more for the reviewers who might be watching this issue as I've not been getting responses in my PR.  I am super happy to see other folks working on the uclibc support. Thank you!

I thought I was the only one who particularly cared about uclibc which is why I've never pushed for it to have representation in the tier 2 architectures, but if you are interested in `armv7-unknown-linux-uclibceabihf` and/or its peers and want to team up, I would be happy to discuss!  (Tier 2 now requires 2 maintainers per the [target tier policy](https://doc.rust-lang.org/nightly/rustc/target-tier-policy.html#tier-2-target-policy).)

operutka:
> I thought I was the only one who particularly cared about uclibc which is why I've never pushed for it to have representation in the tier 2 architectures, but if you are interested in armv7-unknown-linux-uclibceabihf and/or its peers and want to team up, I would be happy to discuss! (Tier 2 now requires 2 maintainers per the [target tier policy](https://doc.rust-lang.org/nightly/rustc/target-tier-policy.html#tier-2-target-policy).)

You are definitely not alone. I do care about uClibc support on Linux for all ARM architectures from v5 up to AArch64. We build our embedded project for these targets quite a lot. And I believe there are other people who also build for these targets. I think uClibc is still very much alive in the embedded Linux world.

Unfortunately, I don't have the capacity to become a full-time maintainer for a build target. I'm months behind with my own open source projects and I don't want to commit to something if I already know I won't be able to give it enough of my time. And, frankly, I'm not sure I know all the nitty-gritty details about uClibc :sweat_smile: But I'm perfectly fine with posting bug fixes whenever I bump into something that does not work.

JohnTitor:
Oh sorry about that. I was simply unaware of your PR as I wasn't involved in a review at all.

skrap:
> Oh sorry about that. I was simply unaware of your PR as I wasn't involved in a review at all.

All good @JohnTitor.  This PR is progress, and progress is great.  I'm not sure what system might be in place (if any) to ensure PRs get attention, but maybe mine just fell through a process hole.  Is there anything I should be doing to get it reviewed, or are occasional reminders the best strategy?  Thanks for any advice.

JohnTitor:
I believe you did a good way, addressing reviews and updating the PR status (i.e. label).
I should've reviewed PRs order by oldest (with filtering `S-waiting-on-review`). Sometimes I do this but sometimes I just check newer notifications in my inbox (like I have limited time window before my day job). This time that's the case. :bow:

PR #5142, opened 2026-06-04
emscripten: deprecate file offset types

dybucc:
# Description

This PR deprecates all LFS64 types and routines in emscripten. This target follows quite closely the musl upstream. The changes they make are not concerned with file offset types. The only exception to this is the `blkcnt_t` type. This type is always 32-bits wide no matter the target machine.

All other types have received the same treatment as in rust-lang/libc#5170.

## Sources

- Upstream emscripten sources showing both the final `alltypes.h` header with the `off_t` definition and the input to the `sed` script that generates it.

  > <https://github.com/emscripten-core/emscripten/blob/1637a13961ab1d6f9595db4db26bfc16a68e6ad8/system/lib/libc/musl/arch/emscripten/bits/alltypes.h#L221>
  > <https://github.com/emscripten-core/musl/blob/85e0e3519655220688e757b9d5bfd314923548bd/include/alltypes.h.in#L29>

- Upstream emscripten fork of musl showing what has been assumed to be the "main" definition of `off64_t` and related types when both the `_GNU_SOURCE` and `_LARGEFILE64_SOURCE` feature test
  macros are present.

  > <https://github.com/emscripten-core/musl/blob/85e0e3519655220688e757b9d5bfd314923548bd/include/sys/types.h#L79>

## Checklist

### DIFF
--- a/libc-test/semver/emscripten.txt
+++ b/libc-test/semver/emscripten.txt
@@ -7,4 +7,3 @@ getgrnam_r
 getpwnam_r
 getpwuid_r
 in6_pktinfo
-posix_fallocate64
--- a/libc-test/semver/emscripten.txt
+++ b/libc-test/semver/emscripten.txt
@@ -8,4 +8,3 @@ getgrnam_r
 getpwnam_r
 getpwuid_r
 in6_pktinfo
-posix_fallocate64
--- a/src/unix/linux_like/emscripten/mod.rs
+++ b/src/unix/linux_like/emscripten/mod.rs
@@ -28,19 +28,38 @@ pub type fsfilcnt_t = u32;
 pub type rlim_t = u64;
 pub type nlink_t = u32;
 
-pub type ino64_t = crate::ino_t;
+#[deprecated(since = "0.2.187", note = "Use `ino_t` instead. See #5142.")]
+pub type ino64_t = ino_t;
+
+#[deprecated(since = "0.2.187", note = "Use `off_t` instead. See #5142.")]
 pub type off64_t = off_t;
+
+#[deprecated(since = "0.2.187", note = "Use `blkcnt_t` instead. See #5142.")]
 pub type blkcnt64_t = crate::blkcnt_t;
+
+#[deprecated(since = "0.2.187", note = "Use `rlim_t` instead. See #5142.")]
 pub type rlim64_t = crate::rlim_t;
 
+#[deprecated(since = "0.2.187", note = "Use `rlimit` instead. See #5142.")]
 pub type rlimit64 = crate::rlimit;
+
+#[deprecated(since = "0.2.187", note = "Use `flock` instead. See #5142.")]
 pub type flock64 = crate::flock;
+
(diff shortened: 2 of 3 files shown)


dybucc:
Style check fails for reasons outside the changes introduced in this PR. The formatting checks seem
to run on a post-macro expanded crate output. When annotating the item produced by the `extern_ty`
macro with an attribute, this expands with a wrongly indented function block, even though the actual
source code is correctly indented. The output log of the CI job shows this.

tgross35:
@rustbot blocked

Holding off until there is a good alternative to off64_t

SnoozeThis:
(https://snoozeth.is/FoBvqh9wkR8) I will wait until Wed, 19 Aug 2026 08:32:43 UTC and then add label S-waiting-on-review and remove label S-blocked.

@rustbot claim.

tgross35:
Actually I think I'm missing something - what does this PR fix? I only see it adding deprecation warnings and deleting `ino64_t` (was that intentional?)

dybucc:
The `ino_t` removal was not intentional. PRs are getting frozen because there's yet to be a solution for 32-bit platforms. This should be no issue in emscripten because the unsuffixed types were already fixed-precision 64-bit wide types. Anybody who was using the LFS64 types can switch to the unsuffixed types with no regressions with a nice warning advising them of it.

The PR description got updated with also mentioning that the blkcnt_t type is an exception. Its bit width does not change so that's the same as with the other file offset types. But it is always defined as a 32-bit and not as a 64-bit signed integer.

And then this transitively applies to the affected records that use these types.

@rustbot ready review

tgross35:
> This should be no issue in emscripten because the unsuffixed types were already fixed-precision 64-bit wide types. Anybody who was using the LFS64 types can switch to the unsuffixed types with no regressions with a nice warning advising them of it.

That is technically feasible, but I am taking user confusion and effort into account. I'd rather ship all of these related deprecations at once shortly before 1.0, so users can adjust everything necessary in one go. Deprecating on a few platforms per release means users are likely going to have to update the same `#[cfg(any(target_os = ...))]` blocks multiple times - my concern is this may encourage users to `allow(deprecated)`.

@rustbot blocked

dybucc:
Can I assume work here is done?

tgross35:
I haven't looked in detail but at least the deprecation message should be shortened to just link the issue, and there's no need for a macro.

dybucc:
Done.

tgross35:
As with the others, I'm happy to take this now if the deprecate is swapped for a comment

@rustbot author

dybucc:
Done.

@rustbot ready

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5142/023c0a3c184c2d7fcd528b64e867075660a5884d..576da17bc498ad4834395dbef475d05bcf7653b1)*

SnoozeThis:
Resolved.

@rustbot label +S-waiting-on-review. @rustbot release-assignment.

PR #5143, opened 2026-06-04
Remove unnecessary clones in match statements

FranciscoTGouveia:
Since the variables `tb` and `defaultbits` will no longer be used after, they can be safely moved rather than cloned. Apologies if I overlooked something.

CC @nunoplopes


JohnTitor (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5143/dbd7ad9a0ef5205226d8541f2a1a7cbdd9da2989..e4b7918e530307412f2b719cbc03497486412d48)*

PR #5146, opened 2026-06-07
ctest: probe optional compiler flags

chitao1234:
# Description

Add `TestGenerator::flag_if_supported` so ctest callers can request C compiler flags that should only be passed when the selected compiler supports them.

This also changes ctest’s built-in warning suppressions for `-Wno-address-of-packed-member` and `-Wno-unknown-warning-option` to use `cc::Build::flag_if_supported`. Some older toolchains (e.g. the default one on dragonfly) reject them when `-Werror` is enabled.

The generated test compiler setup now sets the target before probing flags, so support checks use the same target configuration as the actual compile.

# Sources

Not applicable.

# Checklist

### DIFF
--- a/ctest/src/generator.rs
+++ b/ctest/src/generator.rs
@@ -62,6 +62,7 @@ pub struct TestGenerator {
     pub(crate) includes: Vec<PathBuf>,
     out_dir: Option<PathBuf>,
     pub(crate) flags: Vec<String>,
+    pub(crate) flags_if_supported: Vec<String>,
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5146/022b3249ba8f5f31e52bdecde0eac10f34abe5d7..7329c4815ea4ecaf55f08ae99dcb815fe4581d9d)*

PR #5151, opened 2026-06-08
refactor: remove `ELAST` across BSDs

dybucc:
# Description

This PR completely removes the symbols that were deprecated in rust-lang/libc#5118. This follows from the comments by @JohnTitor in [[1]]. In this instance, removal implies both deletion in the souce file where the constant is declared and purging of the corresponding symbols in the SemVer-tracking file.

[1]: https://github.com/rust-lang/libc/pull/5118#issuecomment-4642281193

## Sources

See rust-lang/libc#5118.

## Checklist


dybucc:

tgross35:
Mentioned at https://github.com/rust-lang/libc/pull/5118#issuecomment-4740074755 that I think we probably shouldn't actually remove these, just document that they may be inconsistent.

Note that there isn't a need to send PRs removing deprecated values; it will be pretty trivial to delete them all after we decide the 0.2 branch is frozen. Until then, it's preferable to keep the `libc-0.2` and `main` branches from diverging as much as reasonably possible, since that means backports can continue to happen with fewer conflicts.

PR #5157, opened 2026-06-08
unix: add preadv2 and pwritev2 to android

Elie-Kh:
Add preadv2 and pwrite2 to linux and Android, making majority of linux_like based systems have support, to allow using the new flags introduced in https://github.com/rust-lang/libc/pull/4452. Move definitions to linux_like/mod.rs instead of each separate system alone.

- [X] Relevant tests in `libc-test/semver` have been updated - no tests to modify
- [~] Tested locally (`cd libc-test && cargo test --target mytarget`);

Getting error while testing locally due to different linux version and mismatch of SW_MAX and MS_NOUSER. Modifying those to match local system (0x11 and 0x80000000) shows passing tests using `cargo test`.

Unable to test other systems, need to run CI

Android source for preadv2 and pwritev2 support: https://android.googlesource.com/platform/bionic/+/cf59e19e22fc7c4795fd955eeecd1f457d79eba0/libc/include/sys/uio.h#117

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2403,7 +2403,13 @@ fn test_android(target: &str) {
             "SOF_TIMESTAMPING_OPT_RX_FILTER" => true,
 
             // FIXME(android): Requires >= 6.9 kernel headers.
-            "AT_HWCAP3" | "AT_HWCAP4" => true,
+            "AT_HWCAP3" | "AT_HWCAP4" | "RWF_NOAPPEND" => true,
+
+            // FIXME(android): Requires >= 6.11 kernel headers.
+            "RWF_ATOMIC" => true,
+
+            // FIXME(android): Requires >= 6.14 kernel headers.
+            "RWF_DONTCACHE" => true,
 
             _ => false,
         }
@@ -2436,6 +2442,9 @@ fn test_android(target: &str) {
 
             // Added in API level 30, but tests use level 28.
             "memfd_create" | "mlock2" | "renameat2" | "statx" | "statx_timestamp" => true,
+
+            // Added in API level 33, but tests use level 28.
+            "preadv2" | "pwritev2" => true,
 
             // Added in glibc 2.25.
             "getentropy" => true,
--- a/libc-test/semver/android.txt
+++ b/libc-test/semver/android.txt
@@ -2364,6 +2364,14 @@ RT_TABLE_MAIN
 RT_TABLE_UNSPEC
 RUSAGE_CHILDREN
 RUSAGE_SELF
+RWF_APPEND
+RWF_ATOMIC
+RWF_DONTCACHE
+RWF_DSYNC
+RWF_HIPRI
+RWF_NOAPPEND
+RWF_NOWAIT
+RWF_SYNC
 R_OK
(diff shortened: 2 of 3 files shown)


JohnTitor:
Seems you messed up the commits in the branch, could you fix it to have exact one commit?

Elie-Kh:
Sure thing, it's been a minute since I've worked with github, let me fix that

Elie-Kh:
And that should have it fixed!

JohnTitor:
Seems you need some more ignores: https://github.com/rust-lang/libc/actions/runs/27483200559/job/81234642626?pr=5157

src/unix/linux_like/linux/mod.rs
```diff
@@ -4370,6 +4370,22 @@ extern "C" {
...
+        flags: c_int,
+    ) -> ssize_t;
+    #[cfg_attr(gnu_file_offset_bits64, link_name = "pwritev64v2")]
+    pub fn pwritev2(
+        fd: c_int,
+        iov: *const crate::iovec,
+        iovcnt: c_int,
+        offset: off_t,
+        flags: c_int,
+    ) -> ssize_t;
```
tgross35: I think these aren't available in uclibc so should be left as is (link a source please if this isn't accurate)

tgross35 (changes requested):
Can you please link the Android sources you are referencing so we don't need to dig them up? (This is requested in the PR template)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5157/38c6231f667b698224c5dc4ffa87b7c48e0e2b47..49416c1490faf021e2c504162f8529d3d8d8062f)*

src/unix/linux_like/linux/mod.rs
tgross35: Can the `linux_like/ilnux` module be restored to the way it was before? I don't think there is any reason for this part of the change.

tgross35 (changes requested):

Elie-Kh:
The change introduced that includes use of preadv2 and pwritev2 is going in Chromium for the Crosvm component (contribution going at https://chromium-review.git.corp.google.com/c/crosvm/crosvm/+/8020959) therefore not strictly an Android change

src/unix/linux_like/linux/mod.rs
Elie-Kh: You are correct for uclibc, I added a cfg to skip. I explained in another comment that the change is being introduced in crosvm, we benefit from having Linux support.

tgross35: It was already supported on Linux prior to this PR? You're moving the definitions from the `gnu` and `musl` modules to `linux` then disabling `uclibc`, which is no net change.

tgross35 (changes requested):

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5157/184114c0adbea6090b78bd934d573c140cecb2c5..0c977675b3d62add8b28535a535501a576c80dbd)*

PR #5160, opened 2026-06-10
android: Add getpwent/setpwent/endpwent

### LINKED ISSUE #3014: `get`/`setpwent` is missing on aarch64 android

sylfn:
Closes rust-lang/libc#3014
Supersedes rust-lang/libc#3015

# Description

This pull request adds support for `getpwent`, `setpwent` and `endpwent` for Android targets. I have no idea how to gate them behind API >= 26, and previous PR didn't do this either so I assume it's fine

# Sources

https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/pwd.h;l=89;drc=02ce401d1e2b31586c94c8b6999364bbbce27388

# Checklist

Tried testing locally on `x86_64-unknown-linux-gnu` (NixOS) host with:
- [cross (git)](https://github.com/cross-rs/cross), `cross test --target aarch64-linux-android` (NDK r25b, API 28, Android 9.0.0_r1, from their a bit dated [Dockerfile](https://github.com/cross-rs/cross/blob/29d00c7803f221f1b3f35e561b03792368fb8339/docker/Dockerfile.aarch64-linux-android#L17)): failed to compile, errors on `IFF_NO_CARRIER`, `TUN_F_*`, `PR_*` and several others.
- `cargo zigbuild --tests --target aarch64-linux-android` (zig 0.16.0): failed due to missing include `sys/param.h`

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2232,6 +2232,9 @@ fn test_android(target: &str) {
             "endgrent" => true,
 
             // Added in API level 26, but some tests use level 24.
+            "getpwent" | "setpwent" | "endpwent" => true,
+
+            // Added in API level 26, but some tests use level 24.
             "getdomainname" | "setdomainname" => true,
 
             // FIXME(android): bad function pointers:
--- a/libc-test/semver/android.txt
+++ b/libc-test/semver/android.txt
@@ -3437,6 +3437,7 @@ dup
 dup2
 duplocale
 endgrent
+endpwent
 endservent
 epoll_create
 epoll_create1
@@ -3568,6 +3569,7 @@ getppid
 getpriority
 getprotobyname
 getprotobynumber
+getpwent
 getpwnam
 getpwnam_r
 getpwuid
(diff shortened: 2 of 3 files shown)


JohnTitor:
You need to skip the test on https://github.com/rust-lang/libc/blob/f4d67773492709f79fcbd4b6f60dd26778291d87/libc-test/build.rs#L1815 to make CI pass.
@rustbot author

tgross35 (approved):
Thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5160/f4d67773492709f79fcbd4b6f60dd26778291d87..cb88ae7f7ebbb7653b2ae8f67d45af40b95b7d47)*

PR #5164, opened 2026-06-12
fix: remove conflicting items in L4Re uClibc

dybucc:
# Description

This PR removes duplicate definitions of `langinfo.h` constants in L4Re uClibc that cause builds to
fail. This is unrelated to the broken L4Re support in Rust std (more on that below.)

A few of these constants were special-cased in the `uclibc` module. These definitions actually
mismatched those of upstream uclibc-ng. The `linux_l4re_shared` module also specialized these
definitions when the target environment was determined to be uClibc, but this module got the values
right.

This patch altogether removes the symbols from the `uclibc` child module, instead relying on the
ones in the `linux_l4re_shared` module.

This change has been marked stable because the fix does not affect the availability of symbols. It
doesn't really affect the values of those symbols either. Anyone who had been using the `libc` crate
with the affected target triple (`x86_64-unknown-l4re-uclibc`) should not have been capable of even
building their programs lest they patched the `libc` crate themselves.

Building std still fails in L4Re, and [[1]] seems to have already tracked that. No further fixes
have been provided to L4Re becaues it seems the same new maintainer is the one that made the changes
in the `libc` crate that make std not build anymore (like not exposing the `lstat64` routine.) They
probably know best, and they seem to have a working build of std, as per [[2]]. This has made
testing impossible, beyond attempting to build the `libc` crate.

[1]: https://github.com/rust-lang/rust/issues/149198
[2]: https://github.com/rust-lang/rust/issues/149198#issuecomment-3566991950

## Sources

- Upstream uclibc-ng sources showing the enum where the constants we expose are declared. The entire
  file has been included because there are a bunch of variants affected across the entire
  enumeration, which starts at the top level. Note that there are also quite a few variants upstream
  that are straight up disabled.

  > <https://gogs.waldemar-brodkorb.de/oss/uclibc-ng/src/master/include/langinfo.h>

## Checklist

### DIFF
--- a/src/unix/linux_like/linux/uclibc/mod.rs
+++ b/src/unix/linux_like/linux/uclibc/mod.rs
@@ -297,64 +297,6 @@ pub const ENOTSUP: c_int = EOPNOTSUPP;
 pub const IPV6_JOIN_GROUP: c_int = 20;
 pub const IPV6_LEAVE_GROUP: c_int = 21;
 
-// These are different from GNU
-pub const ABDAY_1: crate::nl_item = 0x300;
-pub const ABDAY_2: crate::nl_item = 0x301;
-pub const ABDAY_3: crate::nl_item = 0x302;
-pub const ABDAY_4: crate::nl_item = 0x303;
-pub const ABDAY_5: crate::nl_item = 0x304;
-pub const ABDAY_6: crate::nl_item = 0x305;
-pub const ABDAY_7: crate::nl_item = 0x306;
-pub const DAY_1: crate::nl_item = 0x307;
-pub const DAY_2: crate::nl_item = 0x308;
-pub const DAY_3: crate::nl_item = 0x309;
-pub const DAY_4: crate::nl_item = 0x30A;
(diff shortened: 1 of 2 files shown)


skrap:
I believe this is also fixed in https://github.com/rust-lang/libc/pull/5046

dybucc:
True, will close as duplicate. But before that, what do you think about the `CODESET` value change
that this PR includes but #5046 does not?

skrap:
> True, will close as duplicate. But before that, what do you think about the `CODESET` value change
> that this PR includes but #5046 does not?

In the canonical toolchain I am using (see https://doc.rust-lang.org/rustc/platform-support/armv7-unknown-linux-uclibceabihf.html) `CODESET` is 10.  Do you have some reason to believe otherwise?

dybucc:
You're right. I missed the `#else` branch where the enum discriminant got reset in line 322 of `include/langinfo.h`. I was not using any specific toolchain, though. Just upstream uclibc-ng.

PR #5166, opened 2026-06-13
unix: add SSIZE_MAX constant for unix targets

### LINKED ISSUE #1504: Expose SSIZE_MAX to enable safe read system calls
The results of `read` [1] and `write` [2] system calls are implementation-defined if the `count` parameter is larger than value of `SSIZE_MAX` constant defined in `limits.h` according to POSIX.1 standard.

So in order to make it possible to use `read` and `write` in a portable manner, `SSIZE_MAX` should be exposed on POSIX.1-conformant systems.

[1] https://pubs.opengroup.org/onlinepubs/9699919799/
[2] https://pubs.opengroup.org/onlinepubs/9699919799/

telcharr:
Closes rust-lang/libc#1504

# Description

Adds the POSIX `SSIZE_MAX` constant at the unix level. Defined as `ssize_t::MAX`, since `ssize_t` is `isize` on all unix targets making it correct on 32-bit, 64-bit, and x32 without any `cfg`. Binding the named constant is still worth it for discoverability and parity with the C API.

This is a fixed POSIX type limit, not a `*MAX` enumeration sentinel of the kind rust-lang/libc#3131 targets, so it's not a placeholder constant. 

# Sources

- POSIX / Open Group `<limits.h>`:
  https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html
  > {SSIZE_MAX} Maximum value for an object of type ssize_t.

- glibc `posix/bits/poisx1_lim.h`:
  https://github.com/bminor/glibc/blob/04e750e75b73957cf1c791535a3f4319534a52fc/posix/bits/posix1_lim.h#L164-L172
  ```c
  #ifndef	SSIZE_MAX
  /* ssize_t is not formally required to be the signed type
     corresponding to size_t, but it is for all configurations supported
     by glibc.  */
  # if __WORDSIZE == 64 || __WORDSIZE32_SIZE_ULONG
  #  define SSIZE_MAX	LONG_MAX
  # else
  #  define SSIZE_MAX	INT_MAX
  # endif
  #endif
  ```

- musl `include/limits.h:53`:
  https://git.musl-libc.org/cgit/musl/tree/include/limits.h#n53
  ```c
  #define SSIZE_MAX LONG_MAX
  ```

# Checklist


tgross35:
We don't have numeric limits constants for other types, e.g. there is no `SIZE_MAX` or `LONG_MAX` on most platforms. Any reason `ssize_t::MAX` doesn't work?

telcharr:
`ssize_t::MAX` does work. This is really just for parity with the C API rather than a functional gap. The only thing that nudged me towards it was that `INT_MIN`/`INT_MAX` are already at the unix level, so there seemed to be some precedent. Though those look like the only type-limit constants there, so if they're more of a holdover than something you want to extend, happy to close this. Your call.

tgross35:
I've thought about this a bit: the 1.0 release of libc is going to remove public aliases like `c_long` because they exist in `core` nowadays, and I think users should be able to figure out that they can use e.g. `c_long::MAX` in the place of `LONG_MAX` without even needing `libc`. So I'm going to go ahead and close this and the issue.

Nonetheless, thank you for the PR! If you are interested in helping out in this area still, I'd take a PR deprecating the `<int type>_{MIN,MAX}` constants that are scattered around.

telcharr:
Just double checking before I start working on this since it's been a while, do we still want to deprecate the `<int type>_{MIN,MAX}` constants?

tgross35:
Yeah, I think that would be good if you'd like to submit a PR 👍

PR #5169, opened 2026-06-15
NetBSD: `_lwp_park`: `ts` pointer is mutable

joboet:
# Description

NetBSD 9.0 changed the pointer type of the `timespec` argument to `_lwp_park` to a mutable pointer, since it now returns the remaining time for relative timeouts through that pointer. Taking a `*const timespec` causes `std` to be very subtly unsound under stacked-borrows, since it passes a `&mut timespec`, but that is implicitly reborrowed with a read-only tag upon coercion to a `*const timespec` (I noticed this while adding NetBSD shims to miri):
https://github.com/rust-lang/rust/blob/b5d46ecb51c3e4134b82570cfe718f093daa6390/library/std/src/sys/pal/unix/thread_parking.rs#L34-L43

# Sources

* Current version: https://github.com/NetBSD/src/blob/3bea51a26c6e426303801042b20f9553d7e26485/include/lwp.h#L57
* Pre-9.0: https://github.com/NetBSD/src/blob/3760f161e6dd25c683ad32f20d1ce281f87658db/include/lwp.h#L57

# Checklist

Even though it is a breaking change, but perhaps we can get away with this? It currently causes `std` to be unsound under stacked borrows.

### DIFF
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -2233,7 +2233,7 @@ extern "C" {
     pub fn _lwp_park(
         clock: crate::clockid_t,
         flags: c_int,
-        ts: *const crate::timespec,
+        ts: *mut crate::timespec,
         unpark: crate::lwpid_t,
         hint: *const c_void,
         unparkhint: *mut c_void,


tgross35 (approved):
Technically may break people but that's better than them finding out the hard way, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5169/2f877eaaf8eb52fef58968f6e1b697db5eb9d481..93ca3164a8755ad338325db7dd193f2391050b23)*

Issue #4805, opened 2025-10-30
1.0: Remove lfs64 API (`open64` and similar)


tgross35:
With 1.0 we will default to 64-bit `time_t` and `off_t`, and users will need to opt in to 32-bit versions of these APIs - if we expose them at all. This means the API like `open64` will be redundant with `open`. There isn't a reason to use both unless a user opts in to 32-bit `off_t` but occasionally needs to use 64-bit operations: this doesn't sound like a usecase worth supporting. They also aren't posix.

To back this up, it sounds like musl would deprecate them if they could.

> -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE have nothing to do with supporting large files properly but with exposing the idiotic legacy interfaces with 64 on the end of their names (like open64, lseek64, etc.). only -D_FILE_OFFSET_BITS=64 should ever be used, anywhere, even on glibc.

tgross35:
Clarifying what exactly I think we should do here: I don't think we should deprecate just yet. `libc` is going to go through stages that users are going to follow:

1. 64-bit off_t/time_t is not universally present on 32-bit platforms
2. 64-bit off_t/time_t is available behind an unstable cfg (we're pretty much here now)
3. 64-bit off_t/time_t is available behind a stable cfg
4. 1.0, only 64-bit off_t/time_t on all platforms by default

If we start deprecating API before users have a stable cfg, that seems extremely likely to cause user confusion because if you're using code like this:

```rust
#[cfg(env = "glibc", bits = "32")]
use libc::open64 as open;
```

and would like to start testing with `libc_unstable_gnu_time_bits = "64"`, then how do you avoid deprecation warnings while getting the correct behavior? If you import `libc::open` then your library may stop working if `libc_unstable_gnu_time_bits` is unset. If you cfg your import on `libc_unstable_gnu_time_bits = "64"`, you're going to get surprised when that cfg gets renamed to something without `_unstable`. The fact of the matter is that seeing a name like `open64` currently means you _know_ that you are 64-bit safe regardless of platform, and I don't want to start making that ambiguous without a clear path forward.

So, what to do:

At this time I think we should identify everything we want to deprecate without actually deprecating it yet. E.g.:

```rust
// FIXME(1.0,deprecate): lfs64 binding to be removed
fn open64(...);
```

That way we can still get things in piecemeal without flipping the switch just yet. Once we're past step 3 we can easily find+replace to actually deprecate them, and right before step 4 we delete everything that's deprecated.

[...]

PR #5170, opened 2026-06-15
linux(musl): deprecate LFS64 bindings

dybucc:
# Description

This PR removes support for the exposed LFS64 bindings under Linux musl. This follows from the `libc` crate's attempts at providing a single, unsuffixed type with which to provide 64-bit functionality.

This follows from rust-lang/libc#5032, where I understood `libc` was meant to provide a single, unsuffixed type and not necessarily support aliases (much less conditional aliases.) Upstream musl also defines the LFS64 types routines in terms of the unsuffixed types and routines.

This patch ensures we deprecate these types and the whole `musl/lfs64` module in favor of using only the unsuffixed types and routines.

## Sources

No specific sources are provided. Running the following command with `rg` in the main worktree of latest musl should yield search results that show the uniform aliasing mentioned above across all header files.

```sh
rg -C 15 -s -uu -g "**/*.h" -e "\\b_LARGEFILE64_SOURCE\\b"
```

## Checklist

### DIFF
--- a/src/unix/linux_like/l4re/uclibc/mod.rs
+++ b/src/unix/linux_like/l4re/uclibc/mod.rs
@@ -284,6 +284,7 @@ s! {
         pub __wc: crate::wchar_t,
     }
 
+    // FIXME(1.0,deprecate): lfs binding to be removed
     pub struct fpos64_t {
         pub __pos: crate::off64_t,
         pub __mbstate: __mbstate_t,
--- a/src/unix/linux_like/linux/mod.rs
+++ b/src/unix/linux_like/linux/mod.rs
@@ -18,6 +18,7 @@ pub type dev_t = u64;
 pub type socklen_t = u32;
 pub type mode_t = u32;
 pub type ino64_t = u64;
+// FIXME(1.0,deprecate): lfs binding to be removed
 pub type off64_t = i64;
 pub type blkcnt64_t = i64;
 pub type rlim64_t = u64;
@@ -1014,6 +1015,7 @@ s! {
 cfg_if! {
     if #[cfg(not(target_env = "gnu"))] {
         extern_ty! {
+            // FIXME(1.0,deprecate): lfs binding to be removed
             pub type fpos64_t; // FIXME(linux): fill this out with a struct
         }
     }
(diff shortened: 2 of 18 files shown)


tgross35:
I haven't yet taken a close look here but I think deprecation of `*64` types should hold off until we provide a stable way to enable 64-bit time/offset. I don't think users have great recourse for now - until they can get a 64-bit `dirent` et al. on 32-bit platforms, this deprecation basically encourages splitting imports into 32- and 64-bit, which doesn't really improve anything on its own.

dybucc:
I've double-checked the changes in this patch with upstream. There really isn't a difference between unsuffixed types and suffixed types.

musl will just provide 64-bit definitions no matter what. I don't quite understand how does this change not solve the issue altogether. All users, no matter their machine word size or data model, are already using 64-bit file offset types.

The point being, there's no special treatment as there is with `time_t`.

xbjfk:
The history of these are a bit convoluted but these should probably be removed in 1.0, for glibc as well.

This is not based on any formal standard, and to my understanding was only added by musl for compatibility as a workaround for nonportable C code that did not use a feature test for these. 

In my opinion, LFS types should not be used at all, irrespective of libc. Musl intends to remove all aliases to them, even if `_LARGEFILE64_SOURCE` is defined [1]. `man 7 feature_test_macros` also says (emphasis mine):

> _LARGEFILE64_SOURCE
>       Expose definitions for the alternative API specified by the LFS (Large File Summit) as a "transitional  extension"  to  the  Single  UNIX  Specification.   (See  ⟨http://opengroup.org/platform/lfs.html⟩.)   The  alternative API consists of a set of new objects (i.e., functions and types) whose names are suffixed with "64" (e.g., off64_t versus off_t, lseek64() versus lseek(), etc.).  **New programs should not employ this macro; instead _FILE_OFFSET_BITS=64 should be employed.**

As @dybucc mentioned, the *64 suffixed types are functionally equivalent to the non-suffixed ones on musl. No matter what you use, you always get large file support, which carried over to this crate - off_t is 64 bit even on 32 bit architectures on musl.

More/misc info below for those interested:

Rich himself has this summary, where he describes the history as well.
https://www.openwall.com/lists/musl/2022/09/26/1

Rich opened an issue on this repo: https://github.com/rust-lang/libc/issues/2934

Musl commits relating to this: https://github.com/kraj/musl/commit/f0b85fd926c5a381416c94be62e7bc20d9fdbdb1, https://github.com/kraj/musl/commit/2dd8d5e1b8ba1118ff1782e96545cb8a2318592c, https://github.com/kraj/musl/commit/3b94daba711090f1936f59fae6f7cbcba963b29e, https://github.com/kraj/musl/commit/25e6fee27f4a293728dd15b659170e7b9c7db9bc

Google removed LFS64 from Bionic as well (where funnily enough they also talk about this crate): https://www.openwall.com/lists/musl/2022/09/30/8

Oracle removed LFS64 usages in Java: https://bugs.openjdk.org/browse/JDK-8318696

tgross35:
> The history of these are a bit convoluted but these should probably be removed in 1.0, for glibc as well.

I do not disagree, removing in 1.0 is the plan for every OS possible.

This has been a source of confusion so I wrote up how I think we should do this here: https://github.com/rust-lang/libc/issues/4805#issuecomment-4925196708

dybucc:
just replaced the deprecations with the stage 2 comment.

tgross35 (approved):
Current version LGTM with a rebase

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5170/651a32c07041c87a6208fa5ccdb8a71e577c815d..114e9e0fb8530d341e4e832503be958300743fba)*

dybucc:
Done.

PR #5175, opened 2026-06-17
ci: Switch to `ubuntu:26.04` images

tgross35:
Where possible, change to the latest Ubuntu release. The 32-bit images still aren't bumped because of the time64 mismatch.

There are a handful of supporting changes needed, see individual commit messages for more details. The first two commits are https://github.com/rust-lang/libc/pull/5188.


tgross35:
🎉 that took some effort

tgross35:
Applied as part of https://github.com/rust-lang/libc/pull/5215's backport in https://github.com/rust-lang/libc/pull/5294

PR #5177, opened 2026-06-18
ci: Switch Linux runners to ubuntu-26.04

tgross35:


tgross35:
Need to address https://github.com/torvalds/linux/commit/e6ce36ccc86f6d447808a6e620f56d440d74aa19 and https://github.com/torvalds/linux/commit/64dd89ae01f2708a508e028c28b7906e4702a9a7. I think they can just be removed since that's what linux did, and a quick search doesn't show any use https://github.com/search?q=lang%3Arust+%2F%28VM_LAPTOP_MODE%7CKERN_REALROOTDEV%29%28%5B%5E%3A%5D%7C%24%29%2F&type=code.

PR #5179, opened 2026-06-18
docs: Better document breaking change policy and recommended usage

### LINKED ISSUE #5018: Align docs for breaking change policy
The current documented policy is old and kinda confusing.
For example, we allow breaking changes on the tier 3 targets or for padding fields.
cc @tgross35

tgross35:
For a while there has been an informal policy about what is and isn't allowed regarding breaking changes, but it hasn't been written down anywhere. This is an attempt to start to give our users some more concrete expectations, documented in the "Stability Expectations" section. I expect this will be updated as we think of more cases.

I additionally added a "Usage Recommendations" section that is meant to give actionable suggestions about how to avoid breakage and other problems.

Closes: https://github.com/rust-lang/libc/issues/5018

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,149 @@
-//! libc - Raw FFI bindings to platforms' system libraries
+//! Raw FFI bindings to platform system libraries.
+//!
+//! # Usage Guidelines
+//!
+//! `libc` exposes non-Rust interfaces in Rust, which makes for some caveats to its use that are
+//! not present in most Rust libraries. Observing the following guidelines are recommended to help
+//! avoid soundness and stability pitfalls.
+//!
+//! 1. *Never* construct a `libc` struct with `MaybeUninit::uninit()`, initialize it, then call
+//!    `assume_init`. Many structures have padding fields or may gain fields in the future, and
+//!    it is far too easy to end up calling `assume_init` on partially initialized data.
+//!
+//!    Instead, use `MaybeUninit::zeroed()` or the `Default` implementations that are slowly being
+//!    added. Alternatively, access fields only via raw pointer without ever using `assume_init`.
+//!
+//! 2. Avoid relying on the exact value of constants, the exact length of arrays, or the exact
+//!    types of type aliases, as they may change across `libc` versions. That is, if `libc`
+//!    contains code like:
+//!
+//!    <!-- relevant for how rustdoc displays these structs:
+//!         https://github.com/rust-lang/rust/issues/102456 -->
+//!    ```ignore
+//!    const IFNAMSIZ: usize = 16;
+//!
+//!    pub struct ifreq {
+//!        pub ifr_name: [c_char; IFNAMSIZ],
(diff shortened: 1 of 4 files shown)


tgross35:
@JohnTitor I still need to re-read through this and make sure it actually makes sense, but I'd love your thoughts here when you get the chance.

src/lib.rs
```diff
@@ -1,4 +1,131 @@
...
+//!   possibly linking more than is needed or available. If you are using `libc` without the `std`
+//!   feature, consider starting to add link directives now for a smoother 1.0 transition.
+//!
+//! - `extra_traits`: all types in `libc` implement `Clone`, `Copy`, and `Debug`. The
+//!   `extra_traits` feature adds `Eq`, `Hash`, and `PartialEq`.
+//!
+//!   This feature is expected to be removed in libc 1.0. Libraries should instead hash or check
+//!   equality of only needed fields.
+//!
+//! - The features `const-extern-function`, `align`, and `use_std` are all deprecated.
```
JohnTitor:
```suggestion
//! - The features `const-extern-fn`, `align`, and `use_std` are all deprecated.
```

src/lib.rs
```diff
@@ -1,4 +1,131 @@
...
+//! - If the platform libc has accepted breakage on the C side (typically in the form of removing
+//!   old API), the `libc` crate will follow suit.
+//! - Where possible, `#[deprecated(...)]` will be used to warn about changes before applying them.
+//!   Alternative mitigations may be considered.
+//! - Potentially breaking changes will be well-identified in release notes.
+//! - Beyond this, public API is not expected to change on Tier 1 targets. Tier 2 targets have
+//!   relaxed API stability requirements, and API stability is not enforced on tier 3 targets.
+//!
+//! While this section seems scary, keep in mind that it is meant to cover worst-case scenarios. In
+//! practice, breakage is rare and following the above-discussed [Best Practices](#best-practices)
```
JohnTitor: I might be missing something but I don't see this section.

src/lib.rs
```diff
@@ -1,4 +1,131 @@
-//! libc - Raw FFI bindings to platforms' system libraries
+//! Raw FFI bindings to platforms' system libraries.
+//!
+//! # Usage Recommendations
+//!
+//! `libc` takes a differe
```
JohnTitor: Seems incomplete.

JohnTitor (approved):
Looks great so far, leaving some style comments. Feel free to merge after addressing these nits. Thanks a lot!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5179/e34c8b6da178f87bfbb4613fa9b7a8318548eff3..0b967045d12d29302fbf5d97684b477af9f081c1)*

src/lib.rs
tgross35: Whoops, meant for this to link to the top section. Thanks!

PR #5123, opened 2026-05-30
linux_l4re: annotate unstable `*NUM` constants

dybucc:
# Description

Constants matching the `*NUM` naming scheme whose purpose is that of denoting a limit among other symbols have been deprecated in this patch. This stems from the deprecation efforts in rust-lang/libc#3131.

This is a follow up PR to rust-lang/libc#5120.

## Sources

- The `WCTYPE_ALNUM` constant in the TEEOS tree has not been deprecated because I could not find the symbol in the OpenHarmony SDK. I tried with all versions of the latest SDK as released in both their GitHub and their Gitee, but I had no luck after running `rg` with the following command.

  ```sh
  rg -C 10 -S -g "*.h" -e "WCTYPE_ALNUM"
  ````

- The `PT_NUM` in the Fuchsia source constant has not been deprecated because it does not seem to be used as a limit of any sort.

  > <https://cs.opensource.google/search?q=pt_num&sq=&ss=fuchsia%2Ffuchsia:zircon%2F>
- The `AT_PHNUM` constant in NetBSD, FreeBSD, Android and Linux/L4RE has not been deprecated as the accompanying comments mention they are supposed to denote some number of entries, which I assume as not implying the type of "limit" value we're looking to deprecate.

  > <https://github.com/search?q=repo%3ANetBSD%2Fsrc%20AT_PHNUM&type=code>
  > <https://github.com/search?q=repo%3Afreebsd%2Ffreebsd-src%20AT_PHNUM&type=code>
  > <https://github.com/search?q=repo%3Atorvalds%2Flinux%20at_phnum&type=code>
  > <https://cs.android.com/search?q=at_phnum&sq=&ss=android%2Fplatform%2Fsuperproject>
- The `VFS_MAXTYPENUM` FreeBSD constant has not been deprecated because it does not seem to be used as a the type of "limit" value we've deprecated in related PRs.

  > <https://github.com/search?q=repo%3Afreebsd%2Ffreebsd-src+VFS_MAXTYPENUM&type=code>
- The `B_DEV_BAD_DRIVE_NUM` constant under the Haiku tree has not been deprecated as it seems to just be another enum variant among a few other constants denoting errors.

  > <https://github.com/search?q=repo%3Ahaiku%2Fhaiku%20B_DEV_BAD_DRIVE_NUM&type=code>
[...]

PR #5180, opened 2026-06-18
feat: add macro to declare unstable constants

dybucc:
# Description

This PR adds a new macro that should align with the new usage guidelines in rust-lang/libc#5179.

This macro is meant to be used with constants that are likely to suffer from SemVer-breaking changes but which the new usage guidelines accept as being part of stable `libc` crate relases.

It has been used in rust-lang/libc#5118, rust-lang/libc#5119, rust-lang/libc#5120, rust-lang/libc#5121, rust-lang/libc#5122 and rust-lang/libc#5123.

While that PR gets merged, this PR will be left as a draft. The reason for that is that the link to the usage guidelines may change until rust-lang/libc#5179 is complete.

## Sources

Does not qualify.

## Checklist

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -402,6 +402,21 @@ macro_rules! offset_of {
     }};
 }
 
+/// Defines constants with an accompanying doc comment pointing out their
+/// instability, and links to the crate documentation for usage guidelines.
+macro_rules! ct {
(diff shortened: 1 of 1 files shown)


dybucc:
Closing in favor of having a simpler doc comment.

PR #5188, opened 2026-06-21
test: Set version-dependent skips based on header defines

tgross35:
When there are additions or updates that are only available with newer
platform versions, we don't have a great way to handle testing. The only
thing we can really do is unconditionally skip anything that doesn't
pass on the oldest versions we have on CI.

Improve this by adding a preprocess-only compiler invocation that
collects versions from the headers we are building against. These are
then used to skip 

This initial patch only uses the versions for skips on Linux and
replacing the NetBSD `uname` invocation.

Also included is an IO number fix that showed up in the new tests.

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -11,10 +11,17 @@ use std::path::{
     Path,
     PathBuf,
 };
+use std::process::{
+    Command,
+    Stdio,
+};
+use std::sync::LazyLock;
 use std::{
     env,
     io,
 };
+
+use regex::Regex;
 
 fn do_cc() {
     let target = env::var("TARGET").unwrap();
@@ -176,6 +183,8 @@ fn process_semver_file<W: Write, P: AsRef<Path>>(output: &mut W, path: &mut Path
 fn main() {
     // Avoid unnecessary re-building.
     println!("cargo:rerun-if-changed=build.rs");
+    // Ensure version checking works, even if we don't use it.
+    LazyLock::force(&VERSIONS);
 
     do_cc();
     do_ctest();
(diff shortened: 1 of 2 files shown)


tgross35:
I *think* this is pretty reasonable, and probably the easiest way to make sure anyone on a different kernel version than CI doesn't get a handful of errors, but could use a second set of eyes.

libc-test/build.rs
```diff
@@ -6183,10 +6169,134 @@ fn test_qurt(target: &str) {
     ctest::generate_test(&mut cfg, "../src/lib.rs", "ctest_output.rs").unwrap();
 }
 
+/// Platform versions for checking expected support. These are extracted from headers so should be
+/// accurate for the target we are building, rather than the host (which `uname` would provide).
+static VERSIONS: LazyLock<Versions> = LazyLock::new(Versions::init_from_cc);
+
+#[derive(Clone, Copy, Debug, Default)]
+struct Versions {
+    linux: Option<(u32, u32)>,
```
Thomasdezeeuw, line 6178: Does it make sense to remove the `Option` here and make `(0, 0)` the current `None`, that would make the comparison in the code easier to deal with.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5188/320e0eec10118c02f95d92d03dbdac8afcf55b4b..19cf615a55831a4fef3d9fe39031fd1ff859f47a)*

tgross35: I did it this way is to make it more obvious when there's a version detection failure so we can figure it out and fix it, rather than silently falling back and skipping more than needed. But I forgot to actually make use of this, updated now to unwrap rather than default for Linux.

JohnTitor (approved):
The concept makes sense to me 👍 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5188/076d48ffbddf9408319f72ef7361452f8b0206d1..515564da70aa697633644a2452c558a4753bab01)*

PR #5191, opened 2026-06-22
test: Include `linux/pidfd.h` on linux

tgross35:


tgross35:
`linux/pidfd.h` pulls in `linux/fnctl.h`, which conflicts with the libc `fnctl.h`. The flock conflict can be hacked around by defining `HAVE_ARCH_STRUCT_FLOCK`, which disables the definitions of `struct flock*` in the kernel, but there doesn't seem to be anything similar for `struct f_owner_ex`. So close...

PR #5201, opened 2026-06-23
FreeBSD: Replace unstable env to set version with an unstable cfg

tgross35:
Environment variables affect builds for both the host and target, while cfg can be more accurately scoped with `RUSTFLAGS`.


tgross35:
Fyi @asomers, I'm planning to migrate env config to RUSTFLAGS. So if you are using `RUST_LIBC_UNSTABLE_FREEBSD_VERSION=...` anywhere outside of this repo's testing, it will need to be updated to `--cfg=libc_unstable_freebsd_version="..."`.

PR #5203, opened 2026-06-23
Add solarish port_alert definitions

psumbera:
Add PORT_ALERT_* constants and the port_alert declaration shared by Solaris and illumos.

https://docs.oracle.com/cd/E88353_01/html/E37843/port-alert-3c.html
https://github.com/illumos/illumos-gate/blob/edd2f3461fcd719ff41d34b395ef3f5b5994fad1/usr/src/uts/common/sys/port.h#L97-L99
https://github.com/illumos/illumos-gate/blob/edd2f3461fcd719ff41d34b395ef3f5b5994fad1/usr/src/head/port.h#L48

### DIFF
--- a/src/unix/solarish/mod.rs
+++ b/src/unix/solarish/mod.rs
@@ -1823,6 +1823,10 @@ pub const PORT_SOURCE_ALERT: c_int = 5;
 pub const PORT_SOURCE_MQ: c_int = 6;
 pub const PORT_SOURCE_FILE: c_int = 7;
 
+pub const PORT_ALERT_SET: c_int = 0x01;
+pub const PORT_ALERT_UPDATE: c_int = 0x02;
+pub const PORT_ALERT_INVALID: c_int = PORT_ALERT_SET | PORT_ALERT_UPDATE;
+
 pub const NONROOT_USR: c_short = 2;
 
 pub const EMPTY: c_short = 0;
@@ -2657,6 +2661,7 @@ extern "C" {
         events: c_int,
         user: *mut c_void,
     ) -> c_int;
+    pub fn port_alert(port: c_int, flags: c_int, events: c_int, user: *mut c_void) -> c_int;
     #[cfg_attr(
         any(target_os = "solaris", target_os = "illumos"),
         link_name = "__posix_getgrgid_r"


tgross35 (commented):
Looks reasonable to me, but do you have a header or at least manpage you could link?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5203/00183a7e838b35441481bbb712046b0b124def19..44dce715dbbee4e5aceb09f4c35207806eb5077e)*

psumbera:
> Looks reasonable to me, but do you have a header or at least manpage you could link?

https://docs.oracle.com/cd/E88353_01/html/E37843/port-alert-3c.html

src/unix/solarish/mod.rs
```diff
@@ -2657,6 +2661,7 @@ extern "C" {
         events: c_int,
         user: *mut c_void,
     ) -> c_int;
+    pub fn port_alert(port: c_int, flags: c_int, events: c_int, user: *mut c_void) -> c_int;
```
tgross35, line 2664: Unless I'm looking at the wrong thing, it seems like this definition should be Solaris-only since Illumos takes a struct pointer https://github.com/illumos/illumos-gate/blob/edd2f3461fcd719ff41d34b395ef3f5b5994fad1/usr/src/uts/common/fs/portfs/port.c#L453.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5203/00183a7e838b35441481bbb712046b0b124def19..44dce715dbbee4e5aceb09f4c35207806eb5077e)*

psumbera: > Unless I'm looking at the wrong thing, it seems like this definition should be Solaris-only since Illumos takes a struct pointer https://github.com/illumos/illumos-gate/blob/edd2f3461fcd719ff41d34b395ef3f5b5994fad1/usr/src/uts/common/fs/portfs/port.c#L453.

No. It's the same as on Solaris:

https://github.com/illumos/illumos-gate/blob/edd2f3461fcd719ff41d34b395ef3f5b5994fad1/usr/src/head/port.h#L48

tgross35: Ack, thanks for the updated link. LGTM!

tgross35 (approved):

Issue #1020, opened 2018-06-08
Use Rust union types


asomers:
Rust gained a union type in [`1.19.0`](https://blog.rust-lang.org/2017/07/20/Rust-1.19.html).  libc wraps many C unions poorly, usually by pretending they're structs and ignoring some fields.  For example, https://github.com/rust-lang/libc/blob/master/src/unix/mod.rs#L156 .  libc should update all of these to use the correct union types before release 1.0.  It should be fairly easy to find them all, because all of them are listed as exclusions in libc-test/build.rs.

faern:
I guess we can replace existing structs with real unions. Just that it would be a breaking change, needing a major version bump?

I hope all types that represent unions are going to be implemented as unions before `libc 1.0`? The current representations feels like hacks that were basically just waiting for Rust 1.19?

alexcrichton:
@faern yes new types can probably use `union` as necessary, and if/when libc 1.0 happens we can switch to unions for existing types.

tgross35:
We've been using unions for a while, and the needed breaking changes have at least made it to `main`.

PR #5205, opened 2026-06-23
ctest: check incompatible pointer types for public fields.

### LINKED ISSUE #5196: ctest: detect improper field types
I thought we had a check for this but I guess not, given that https://github.com/rust-lang/libc/issues/1020 doesn't seem to present any errors. This kind of check would be very helpful there.

E.g. for this input:

```c
struct foo {
    int a;
};
```

Produce something like this test:

```c
void check_foo_a(struct foo test_foo) {
    long test_foo_a_0 = test_foo.a;
    long *test_foo_a_1 = &test_foo.a;
}
```

The `_1` test comparing pointers is the important bit here since it should catch `-Wincompatible-pointer-types`. I don't know if the `_0` test will catch anything that doesn't, but it doesn't hurt.

This test can be created for both structs and unions.

mbyx:
# Description
Attempts to fix rust-lang/libc#5196 by adding a check for improper field types into the C side of the field_type test. It then adds this test for union and struct fields.

Also adds a new error type corresponding to `cc::Error` so that in the future it is easier to create tests that expect failure as the success condition.

# Sources

# Checklist

### DIFF
--- a/ctest/templates/test.c
+++ b/ctest/templates/test.c
@@ -104,6 +104,14 @@ CTEST_EXTERN uint64_t ctest_size_of__{{ item.id }}__{{ item.field.ident() }}(voi
 typedef {{ item.volatile_keyword }}{{ item.field_return_type }};
 CTEST_EXTERN ctest_field_ty__{{ item.id }}__{{ item.field.ident() }}
 ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}({{ item.c_ty }} *b) {
+    {#
+        // In order to trigger -Wincompatible-pointer-types, we need to attempt to create a
+        // pointer to the given struct or field type, but using the type translated from Rust.
+    #}
+    ctest_field_ty__{{ item.id }}__{{ item.field.ident() }} ptr;
+    ptr = &b->{{ item.c_field }};
+    (void)ptr; {# avoid unused warnings #}
+
     return &b->{{ item.c_field }};
 }
 {%- endfor +%}
--- a/ctest/tests/basic.rs
+++ b/ctest/tests/basic.rs
@@ -209,3 +209,35 @@ fn test_raw_identifier_field() {
         assert!(result.is_ok());
     }
 }
+
+#[test]
+fn test_mismatched_union_field_ty() {
+    let include_path = PathBuf::from("tests/input");
+    let crate_path = include_path.join("mismatched_union_field_ty.rs");
+    let library_path = "mismatched_union_field_ty.out.a";
+
+    let (mut gen_, _out_dir) = default_generator(1, Some("mismatched_union_field_ty.h")).unwrap();
+
+    if env::var("TARGET_PLATFORM") == env::var("HOST_PLATFORM") {
+        let result = generate_test(&mut gen_, &crate_path, library_path);
+        assert!(result.is_err());
+        // As cc prints its warning/error messages to stderr, we cannot access them from cc::Error,
+        // and so cannot assert that the error was actually due to -Wincompatible-pointer-types.
+    }
+}
+
+#[test]
+fn test_mismatched_struct_field_ty() {
+    let include_path = PathBuf::from("tests/input");
+    let crate_path = include_path.join("mismatched_struct_field_ty.rs");
+    let library_path = "mismatched_struct_field_ty.out.a";
+
+    let (mut gen_, _out_dir) = default_generator(1, Some("mismatched_struct_field_ty.h")).unwrap();
+
+    if env::var("TARGET_PLATFORM") == env::var("HOST_PLATFORM") {
+        let result = generate_test(&mut gen_, &crate_path, library_path);
+        assert!(result.is_err());
+        // FIXME(ctest): As cc prints its warning/error messages to stderr, we cannot access them from cc::Error,
+        // and so cannot assert that the error was actually due to -Wincompatible-pointer-types.
+    }
+}
--- a/ctest/tests/input/mismatched_union_field_ty.h
+++ b/ctest/tests/input/mismatched_union_field_ty.h
@@ -1,0 +1,8 @@
+#include <stdint.h>
+
+// The union must match the size and alignment of the Rust side,
+// otherwise different warnings are emitted that hide -Wincompatible-pointer-types.
+union Bar {
+    int64_t x;
+    int32_t y;
+};
--- a/ctest/src/generator.rs
+++ b/ctest/src/generator.rs
@@ -103,6 +103,8 @@ pub enum GenerationError {
     OsError(std::io::Error),
     #[error("one of {0} environment variable(s) not set")]
     EnvVarNotFound(String),
+    #[error("unable to compile C tests {0}")]
+    CompileError(cc::Error),
 }
 
 impl TestGenerator {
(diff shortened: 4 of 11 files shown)


mbyx:
Also, I'm not sure, but I think this was already tested in a way? Because while working on it I got the following result:
```
cargo:warning=/tmp/.tmpQcOH17/incompatible_ptr_types.out.c:66:9: error: assignment to 'ctest_field_ty__Foo__a' {aka 'float *'} from incompatible pointer type 'int32_t *' {aka 'int *'} [-Werror=incompatible-pointer-types]
cargo:warning=   66 |     ptr = &b->a;
cargo:warning=      |         ^
cargo:warning=/tmp/.tmpQcOH17/incompatible_ptr_types.out.c:69:12: error: returning 'int32_t *' {aka 'int *'} from a function with incompatible return type 'ctest_field_ty__Foo__a' {aka 'float *'} [-Werror=incompatible-pointer-types]
cargo:warning=   69 |     return &b->a;
cargo:warning=      |            ^~~~~
cargo:warning=At top level:
```

For the specific test case at least, the correct warning would have popped up even if the extra test lines had not been added. I might not be understanding the requirements here though.

Oh and if that is the case, the reason for it not detecting before may be that the field was not public, as non public fields can't be checked. I think, it's been a while since I looked at ctest :)

tgross35:
Huh; I thought I saw a case related to https://github.com/rust-lang/libc/issues/1020 where we had an integer as a field but a union is the correct type, and no skips. Now I can't find that, and you're right that it seems covered.

Maybe I was looking at union fields - do you know if we test those?

mbyx:
Yup, we do test those. The `populate_field_ptr_tests` in `template.rs` creates the `field_ptr` test for all public struct and union fields that haven't been manually skipped. And the `field_ptr` test creates a pointer to that field that is returned by the function, while using both the actual and the Rust translated types, so it should catch all instances of this warning.

tgross35:
You're right, sorry for all the confusion here. It looks like tests don't cover this, would you mind adding them? Could just be `ctest-test/src/mismatched-union-field-ty.{c,rs}` and `ctest-test/src/mismatched-struct-field-ty.{c,rs}` to make sure compilation fails.

mbyx:
Added those. For now I've kept the addition to the `field-ptr` test, but it really isn't needed. Here's the output of `cc` for these tests:
```
cargo:warning=/tmp/.tmpSnBGZX/mismatched_union_field_ty.out.c: In function 'ctest_field_ptr__Foo__a':
cargo:warning=/tmp/.tmpSnBGZX/mismatched_union_field_ty.out.c:66:9: error: assignment to 'ctest_field_ty__Foo__a' {aka 'int *'} from incompatible pointer type 'union Bar *' [-Werror=incompatible-pointer-types]
cargo:warning=   66 |     ptr = &b->a;
cargo:warning=      |         ^
cargo:warning=/tmp/.tmpSnBGZX/mismatched_union_field_ty.out.c:69:12: error: returning 'union Bar *' from a function with incompatible return type 'ctest_field_ty__Foo__a' {aka 'int *'} [-Werror=incompatible-pointer-types]
cargo:warning=   69 |     return &b->a;
cargo:warning=      |            ^~~~~
cargo:warning=/tmp/.tmp5OQS7S/mismatched_struct_field_ty.out.c: In function 'ctest_field_ptr__Foo__a':
cargo:warning=/tmp/.tmp5OQS7S/mismatched_struct_field_ty.out.c:66:9: error: assignment to 'ctest_field_ty__Foo__a' {aka 'int *'} from incompatible pointer type 'struct Bar *' [-Werror=incompatible-pointer-types]
cargo:warning=   66 |     ptr = &b->a;
cargo:warning=      |         ^
cargo:warning=/tmp/.tmp5OQS7S/mismatched_struct_field_ty.out.c:69:12: error: returning 'struct Bar *' from a function with incompatible return type 'ctest_field_ty__Foo__a' {aka 'int *'} [-Werror=incompatible-pointer-types]
cargo:warning=   69 |     return &b->a;
cargo:warning=      |        
```

So it is redundant I think.

ctest/templates/test.c
```diff
@@ -104,6 +104,12 @@ CTEST_EXTERN uint64_t ctest_size_of__{{ item.id }}__{{ item.field.ident() }}(voi
 typedef {{ item.volatile_keyword }}{{ item.field_return_type }};
 CTEST_EXTERN ctest_field_ty__{{ item.id }}__{{ item.field.ident() }}
 ctest_field_ptr__{{ item.id }}__{{ item.field.ident() }}({{ item.c_ty }} *b) {
+    // In order to trigger -Wincompatible-pointer-types, we need to attempt to create a
+    // pointer to the given struct or field type, but using the type translated from Rust.
```
tgross35: Could you make this a `{# ... #}` template comment so it doesn't show up in every expansion?

ctest/templates/test.c
```diff
@@ -104,6 +104,12 @@ CTEST_EXTERN uint64_t ctest_size_of__{{ item.id }}__{{ item.field.ident() }}(voi
...
+    ctest_field_ty__{{ item.id }}__{{ item.field.ident() }} ptr;
+    ptr = &b->{{ item.c_field }};
+    (void)ptr;
```
tgross35:
```c
(void)ptr; {# avoid unused warnings #}
```

ctest/tests/input/mismatched_union_field_ty.h
```diff
@@ -0,0 +1,12 @@
...
+// The union must match the size and alignment of the Rust side,
+// otherwise different warnings are emitted that hide -Wincompatible-pointer-types.
+union Bar {
+    int32_t x;
+    float y;
+};
+
+struct Foo {
+    union Bar a; 
+};
```
tgross35, line 8: For these tests I meant to check fields within the union, i.e. drop `struct Foo` and have the Rust mapping be:

```rust
union Bar {
    pub x: i64,
    pub y: core::ffi::c_float,
}
```

No harm in keeping this around with a different name

ctest/tests/basic.rs
```diff
@@ -209,3 +209,35 @@ fn test_raw_identifier_field() {
...
+    let crate_path = include_path.join("mismatched_union_field_ty.rs");
+    let library_path = "mismatched_union_field_ty.out.a";
+
+    let (mut gen_, _out_dir) = default_generator(1, Some("mismatched_union_field_ty.h")).unwrap();
+
+    if env::var("TARGET_PLATFORM") == env::var("HOST_PLATFORM") {
+        let result = generate_test(&mut gen_, &crate_path, library_path);
+        assert!(result.is_err());
+        // As cc prints its warning/error messages to stderr, we cannot access them from cc::Error,
+        // and so cannot assert that the error was actually due to -Wincompatible-pointer-types.
```
tgross35, line 225: Brought this up as https://github.com/rust-lang/cc-rs/issues/1769

tgross35 (commented):
Looks pretty good to me, just a template nit and a test request

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5205/95959125cea361c1b3676b02adb1e4e1772cae9c..db9409bdaa3a3d57cf28167fa303e71d42326a6e)*

tgross35 (approved):
Looks great, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5205/f08bfce345c586260a5678dc4442b3e9de856382..6658e5fc898d2fcfb3e271378049f9052421ee0c)*

PR #5206, opened 2026-06-23
ci: Don't set `fail-fast` in 0.2 CI

tgross35:
Since most CI for 0.2 is backports that affect all targets, don't stop testing one target if another fails.


tgross35:
Part of https://github.com/rust-lang/libc/pull/5207

Issue #5204, opened 2026-06-23
uclibc targets (and others?) cannot perform libc-test macro expansion due to lack of core lib


skrap:
During the development of #5046 , the [request was made](https://github.com/rust-lang/libc/pull/5046/changes#r3444571259) that the bitrotten CI script for armv7-unknown-linux-uclibceabihf be resuscitated and made to work.  However, I encountered difficulties with the somewhat-new macro expansion functions within the libc testing infrastructure.

* The current [macro expansion code](https://github.com/rust-lang/libc/blob/d73763fe397958f427b360fdc76da56df44f1a89/ctest/src/macro_expansion.rs#L20-L24) wants to call the following:
```
RUSTC_BOOTSTRAP=1 rustc -Zunpretty=expanded --edition 2021 /checkout/src/lib.rs --target armv7-unknown-linux-uclibceabihf
```
but this results in the following error:
```
macro expansion failed with exit status: 1: error[E0463]: can't find crate for `core`
  |
  = note: the `armv7-unknown-linux-uclibceabihf` target may not be installed
  = help: consider building the standard library from source with `cargo build -Zbuild-std`

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0463`
```

@tgross35 suggested using `-Zbuild-std` but this is a cargo-ism, not a switch which `rustc` understands.  I also was asked to try running `cargo rustc` in order to invoke the `-Zbuild-std` switch, but this isn't really the same as invoking rustc directly, as it wants to build the whole crate. (More precisely, it wants to build exactly one target within the crate.)  The macro expansion code wants to build exactly one `.rs` file.  This capability is not available from `cargo rustc`.  If you try to make it work you end up with macro expansion trying to call:
```
RUSTC_BOOTSTRAP=1 cargo rustc --profile check --target armv7-unknown-linux-uclibceabihf -- -Zunpretty=expanded/checkout/src/lib.rs
```
which fails with:
```
macro expansion failed with exit status: 101: error: extra arguments to `rustc` can only be passed to one target, consider filtering the package by passing, e.g., `--lib` or `--bin NAME` to specify a single target
```

To reproduce:
* Dockerfile needs some amendments, use  attached Dockerfile at `ci/docker/armv7-unknown-linux-uclibceabihf/Dockerfile`
[...]

PR #5214, opened 2026-06-24
use `cargo rustc` for macro expansion so -Zbuild-std can be supported

skrap:
# Description

Address https://github.com/rust-lang/libc/issues/5204 by invoking macro expansion through `cargo rustc` instead of rustc directly.  This allows the use of the `-Zbuild-std` flag, enabling targets without a prebuilt std (e.g. tier 3 targets) to build the libc tests.

# Checklist

### DIFF
--- a/ctest/Cargo.toml
+++ b/ctest/Cargo.toml
@@ -14,6 +14,7 @@ cc = "1.2.43"
 proc-macro2 = { version = "1.0.103", features = ["span-locations"] }
 quote = "1.0.41"
 syn = { version = "2.0.108", features = ["full", "visit", "extra-traits"] }
+tempfile = "3.23.0"
 thiserror = "2.0.17"
 
 [lints]
@@ -21,4 +22,3 @@ workspace = true
 
 [dev-dependencies]
 pretty_assertions = "1.4.1"
-tempfile = "3.23.0"
--- a/ctest/src/generator.rs
+++ b/ctest/src/generator.rs
@@ -13,6 +13,7 @@ use syn::visit::Visit;
 use thiserror::Error;
 
 use crate::ffi_items::FfiItems;
+use crate::macro_expansion::expand_with_args;
 use crate::template::{
     CTestTemplate,
     RustTestTemplate,
@@ -32,7 +33,6 @@ use crate::{
     Type,
     Union,
     VolatileItemKind,
-    expand,
     get_build_target,
 };
 
@@ -78,6 +78,10 @@ pub struct TestGenerator {
     cfg: Vec<(String, Option<String>)>,
     /// A list of functions that remaps names used in the tests.
     mapped_names: Vec<MappedName>,
+    /// Extra command line args to pass to cargo when generating macro expansions.
+    macro_expansion_cargo_args: Vec<String>,
+    /// Crate name to use when performing macro expansion.
+    crate_name: Option<String>,
     /// The programming language to generate tests in.
     pub(crate) language: Language,
     /// A list of functions that determine what items to skip all tests for.
@@ -723,6 +727,22 @@ impl TestGenerator {
         self
     }
 
+    /// Configures extra arguments to be passed to cargo during macro expansion.  
+    /// This can be used, for example, to pass `-Zbuild-std` if required.
+    pub fn expansion_cargo_args_mut(&mut self) -> &mut Vec<String> {
+        &mut self.macro_expansion_cargo_args
+    }
+
+    /// Configures the crate name which should be used during macro expansion.
+    ///
+    /// If the tested crate uses `#![crate_name = "..."]`, this must be called with the
+    /// same name. Otherwise, there will be an error about `--crate-name` not
+    /// matching.
+    pub fn crate_name(&mut self, name: String) -> &mut Self {
+        self.crate_name = Some(name);
+        self
+    }
+
     /// Configures whether tests for the type of a field is skipped or not.
     ///
     /// The closure is given a Rust struct as well as a field within that
@@ -1087,7 +1107,14 @@ impl TestGenerator {
         crate_path: impl AsRef<Path>,
         output_file_path: impl AsRef<Path>,
     ) -> Result<PathBuf, GenerationError> {
-        let expanded = expand(&crate_path, &self.cfg, get_build_target(self)?).map_err(|e| {
+        let expanded = expand_with_args(
+            &crate_path,
+            &self.cfg,
+            get_build_target(self)?,
+            self.crate_name.as_deref(),
+            &self.macro_expansion_cargo_args,
+        )
+        .map_err(|e| {
             GenerationError::MacroExpansion(crate_path.as_ref().to_path_buf(), e.to_string())
         })?;
         let ast = syn::parse_file(&expanded)
--- a/ctest/src/macro_expansion.rs
+++ b/ctest/src/macro_expansion.rs
@@ -1,4 +1,3 @@
-use std::env;
 use std::fs::canonicalize;
 use std::path::Path;
 use std::process::Command;
@@ -14,18 +13,62 @@ pub fn expand<P: AsRef<Path>>(
     cfg: &[(String, Option<String>)],
     target: String,
 ) -> Result<String> {
-    let rustc = env::var("RUSTC").unwrap_or_else(|_| String::from("rustc"));
+    expand_with_args(crate_path, cfg, target, None, &[])
+}
 
-    let mut cmd = Command::new(rustc);
+/// Extension of `expand` which allows a crate_name override for
+/// libc and other clients which require crate_name to match the
+/// original crate's name.
+pub(crate) fn expand_with_args<P: AsRef<Path>>(
+    crate_path: P,
+    cfg: &[(String, Option<String>)],
+    target: String,
+    crate_name: Option<&str>,
+    extra_cargo_args: &[String],
+) -> Result<String> {
+    let dir = tempfile::tempdir()?;
+
+    // make a Cargo.toml pointing to the crate path
+    let cargo_toml_path = dir.path().join("Cargo.toml");
+    let crate_name = crate_name.unwrap_or("ctest-expansion-tmp");
+    // FIXME(#5238): allow building using an existing manifest
+    let cargo_toml = format!(
+        r#"
+[package]
+name = '''{crate_name}'''  # Needs to match #![crate_name = "..."] if it exists.
+edition = '''{EDITION}'''
+[lib]
+path = '''{}'''
+[lints.rust]
+unexpected_cfgs = "allow"
+"#,
+        canonicalize(crate_path)?.display()
+    );
+    std::fs::write(cargo_toml_path, cargo_toml)?;
+
+    let mut cmd = Command::new(std::env::var("CARGO").unwrap_or("cargo".into()));
     cmd.env("RUSTC_BOOTSTRAP", "1")
-        .arg("-Zunpretty=expanded")
-        .arg("--edition")
-        .arg(EDITION) // By default, -Zunpretty=expanded uses 2015 edition.
-        .arg(canonicalize(crate_path)?);
+        .current_dir(dir.path())
+        .arg("rustc")
+        .arg("--lib")
+        .arg("--profile")
+        .arg("check")
+        .args(extra_cargo_args);
+
+    // In order to avoid warnings causing failures here when the
+    // environment's RUSTFLAGS contains -Dwarnings (or similar),
+    // we will clear RUSTFLAGS.
+    cmd.env_remove("RUSTFLAGS");
+
+    // set an independent target dir so we don't deadlock on the cargo build lock.
+    cmd.arg("--target-dir").arg(dir.path().join("target"));
 
     if !target.is_empty() {
         cmd.arg("--target").arg(target);
     }
+
+    cmd.arg("--");
+    cmd.arg("-Zunpretty=expanded");
 
     // `libc` uses non standard cfg flags as well, which have to be manually expanded.
     for (k, v) in cfg {
@@ -39,7 +82,11 @@ pub fn expand<P: AsRef<Path>>(
 
     if !output.status.success() {
         let stderr = std::str::from_utf8(&output.stderr)?;
-        return Err(format!("macro expansion failed with {}: {}", output.status, stderr).into());
+        return Err(format!(
+            "macro expansion failed with {}: {}, {:?}",
+            output.status, stderr, cmd
+        )
+        .into());
     }
 
     let expanded = std::str::from_utf8(&output.stdout)?.to_string();
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -107,6 +107,11 @@ fn ctest_cfg() -> ctest::TestGenerator {
     // FIXME(1.0): These aliases will eventually be removed.
     cfg.skip_alias(|ty| ty.ident() == "__uint128");
 
+    if env::var("LIBC_CI_ZBUILD_STD").is_ok() {
+        *cfg.expansion_cargo_args_mut() = vec!["-Zbuild-std=core,std".into()];
+    }
+    cfg.crate_name("libc".into());
+
     cfg
 }
 


ctest/src/macro_expansion.rs
```diff
@@ -14,19 +15,50 @@ pub fn expand<P: AsRef<Path>>(
...
+    let dir = tempfile::tempdir()?;
+
+    // make a Cargo.toml pointing to the crate path
+    let cargo_toml_path = dir.path().join("Cargo.toml");
+    let mut cargo_toml = std::fs::File::create(&cargo_toml_path)?;
+    writeln!(
+        &mut cargo_toml,
+        r#"
+[package]
+name = "libc"
```
tgross35: This will need to be configurable:

1. Move this function body to a new `expand_with_name` that takes a `name: Option<&str>` parameter
2. Use that parameter in the template here, or `ctest-expansion` if `None`
3. `expand` can then call `expand_with_name` with `None` to keep the public API
4. Add a `crate_name(&str)` function to `TestGenerator` that saves it in the config
5. Replace internal calls to `expand` with `expand_with_name`, using the stored `crate_name` if available

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/95959125cea361c1b3676b02adb1e4e1772cae9c..b5da0517d87247ef2fcd744279085c29752b4392)*

tgross35:
Cc @mbyx for ctest

ctest/src/macro_expansion.rs
skrap: Since it works as is, what's the need for this change? I think everything using this is either "libc" or it doesn't care if it is called "libc" for the purposes of macro expansion.

tgross35: It's going to be confusing if `ctest` users unrelated to `libc` get an error compiling `libc`, and also other users could be setting `crate_name`

skrap: Ah, I didn't realize this crate was used outside of the libc context.  Thanks.  Also, slightly terrifying for me, ha.

skrap: This change also drops the ability to use the `RUSTC` env var to use a custom `rustc`.  I don't know if there's any way to keep that when using `cargo rustc` or not.  Given that there's external users, I will have to check on that.

tgross35: Not a _lot_ of places, but at least a few important places 🙂 https://github.com/search?q=path%3ACargo.toml+%2F%5Cbctest%5Cb.*%3D%2F&type=code. Certainly nothing to worry about, though, it's a low pressure crate.

ctest/src/macro_expansion.rs
```diff
@@ -14,19 +15,50 @@ pub fn expand<P: AsRef<Path>>(
...
+edition = "{EDITION}"
+[lib]
+path = '{}'
+[lints.rust]
+unexpected_cfgs = "allow"
```
tgross35, line 43: Add a toml comment about why

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/95959125cea361c1b3676b02adb1e4e1772cae9c..1100563fe126f6b76d1d0008b412bedc9815dfd7)*

ctest/src/macro_expansion.rs
```diff
@@ -14,19 +15,50 @@ pub fn expand<P: AsRef<Path>>(
...
+    let cargo_toml_path = dir.path().join("Cargo.toml");
+    let mut cargo_toml = std::fs::File::create(&cargo_toml_path)?;
+    writeln!(
+        &mut cargo_toml,
+        r#"
+[package]
+name = "libc"
+edition = "{EDITION}"
+[lib]
+path = '{}'
```
tgross35: Use a regular string here and escape it by replacing `\`->`\\` and `"`->`\"`, since this will act weird in the (unlikely) case the path contains a `'`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/95959125cea361c1b3676b02adb1e4e1772cae9c..1100563fe126f6b76d1d0008b412bedc9815dfd7)*

ctest/src/macro_expansion.rs
```diff
@@ -14,19 +15,50 @@ pub fn expand<P: AsRef<Path>>(
...
+
+    let mut cmd = Command::new("cargo");
 
-    let mut cmd = Command::new(rustc);
     cmd.env("RUSTC_BOOTSTRAP", "1")
-        .arg("-Zunpretty=expanded")
-        .arg("--edition")
-        .arg(EDITION) // By default, -Zunpretty=expanded uses 2015 edition.
-        .arg(canonicalize(crate_path)?);
+        .env("RUSTFLAGS", "") // ignore RUSTFLAGS so we don't get -Dwarnings
```
tgross35: Add a `TestGenerator::expansion_rustflags(&str)` and set `RUSTFLAGS` only if that has been called, otherwise just let it inherit. Since there might be some global options needed to build.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/95959125cea361c1b3676b02adb1e4e1772cae9c..1100563fe126f6b76d1d0008b412bedc9815dfd7)*

ctest/src/macro_expansion.rs
```diff
@@ -14,19 +15,50 @@ pub fn expand<P: AsRef<Path>>(
...
+        .arg("--lib")
+        .arg("--profile")
+        .arg("check");
+
+    // set an independent target dir so we don't deadlock on the cargo build lock.
+    cmd.arg("--target-dir").arg(dir.path().join("target"));
+
+    if env::var("LIBC_CI_ZBUILD_STD").is_ok() {
+        cmd.arg("-Zbuild-std=core,std");
+    }
```
tgross35: Add `TestGenerator::build_std(bool)` and `TestGenerator::build_core(bool)` to control this, which should set `-Zbuild-std=std` / `-Zbuild-std=core`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/95959125cea361c1b3676b02adb1e4e1772cae9c..1100563fe126f6b76d1d0008b412bedc9815dfd7)*

ctest/src/macro_expansion.rs
```diff
@@ -14,19 +15,50 @@ pub fn expand<P: AsRef<Path>>(
...
     target: String,
 ) -> Result<String> {
-    let rustc = env::var("RUSTC").unwrap_or_else(|_| String::from("rustc"));
+    let dir = tempfile::tempdir()?;
+
+    // make a Cargo.toml pointing to the crate path
+    let cargo_toml_path = dir.path().join("Cargo.toml");
+    let mut cargo_toml = std::fs::File::create(&cargo_toml_path)?;
+    writeln!(
+        &mut cargo_toml,
```
tgross35: Nit: just construct the string with `format!` then use `fs::write` to avoid bothering with the `File` object.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/95959125cea361c1b3676b02adb1e4e1772cae9c..1100563fe126f6b76d1d0008b412bedc9815dfd7)*

ctest/src/macro_expansion.rs
```diff
@@ -14,19 +15,50 @@ pub fn expand<P: AsRef<Path>>(
     cfg: &[(String, Option<String>)],
     target: String,
 ) -> Result<String> {
-    let rustc = env::var("RUSTC").unwrap_or_else(|_| String::from("rustc"));
```
tgross35, line 17: Keep something similar to this but with `CARGO` for setting something not in-path.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/95959125cea361c1b3676b02adb1e4e1772cae9c..1100563fe126f6b76d1d0008b412bedc9815dfd7)*

tgross35 (changes requested):

ctest/src/macro_expansion.rs
skrap: I'm not sure I agree here.  If we want to be safe against creating illegal TOML files, we should honor 
> Basic strings are surrounded by quotation marks ("). Any Unicode character may be used except those that must be escaped: quotation mark, backslash, and the control characters other than tab (U+0000 to U+0008, U+000A to U+001F, U+007F).
(From https://toml.io/en/v1.1.0#string)

If you want this to be bulletproof, I will import the `toml` crate and write out the file with that.  Seems overkill but I can do it.

tgross35: Fair enough. Maybe just `'''...'''` to prevent against the simple case?

ctest/src/macro_expansion.rs
skrap: We were not building with inherited RUSTFLAGS before this change, as RUSTFLAGS are a cargo-ism, and we were not using cargo before, but rather using `rustc` directly.

I think I will be a bit more tactical and just add a `-Awarnings` at the end of `RUSTFLAGS` so we don't error out over warnings if `-Dwarnings` is present.  No new API required, and should be compatible with existing clients.

ctest/src/macro_expansion.rs
skrap: I felt like this was too API-heavy, but I do appreciate that the users of the crate need a way to customize the invocation of cargo for macro expansion.  So I added a more generic "macro expansion extra cargo args" TestGenerator configuration, which can be used for this purpose, and is flexible to many other situations as well.

ctest/src/macro_expansion.rs
skrap: done.

ctest/src/macro_expansion.rs
skrap: good call.  done.

skrap:
@rustbot  review

I think I have addressed the feedback, though in exploring the changes I sometimes made some different choices.  I've left comments in the conversations where appropriate.

ctest/src/generator.rs
```diff
@@ -723,6 +725,20 @@ impl TestGenerator {
         self
     }
 
+    /// Configures any extra arguments which to be passed to cargo
+    /// during macro expansion.  This can be used, for example, to pass
+    /// -Zbuild-std if required.
+    pub fn set_macro_expansion_cargo_args(&mut self, args: Vec<String>) {
+        self.macro_expansion_cargo_args = args;
+    }
```
tgross35: To keep things a bit more composable so you can set args one at a time, replace this with `fn expansion_cargo_arg(&mut self, &str)` and `fn clear_expansion_cargo_args(&mut self)`.

Also these methods should return `&mut Self`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/483795464d6fbbd5fe07d59013f7c9e2bb46ec82..329498887b3317d812fafbf678acef83d5d3dd95)*

ctest/src/generator.rs
```diff
@@ -723,6 +725,20 @@ impl TestGenerator {
         self
     }
 
+    /// Configures any extra arguments which to be passed to cargo
+    /// during macro expansion.  This can be used, for example, to pass
+    /// -Zbuild-std if required.
```
tgross35:
```rust
/// `-Zbuild-std`
```

(backticks)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/483795464d6fbbd5fe07d59013f7c9e2bb46ec82..329498887b3317d812fafbf678acef83d5d3dd95)*

ctest/src/generator.rs
```diff
@@ -723,6 +725,20 @@ impl TestGenerator {
...
+    pub fn set_macro_expansion_cargo_args(&mut self, args: Vec<String>) {
+        self.macro_expansion_cargo_args = args;
+    }
+
+    /// Configures the crate name which should be used during macro expansion.
+    /// This can be important for crates which have crate_name! macros or
+    /// similar, like libc.
+    pub fn set_macro_expansion_crate_name(&mut self, name: String) {
+        self.macro_expansion_crate_name = Some(name);
+    }
```
tgross35, line 744: Just call this `crate_name` in case we want to reuse this beyond expansion, and take `&str` rather than `String`.

`-> &mut Self` as well

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/483795464d6fbbd5fe07d59013f7c9e2bb46ec82..329498887b3317d812fafbf678acef83d5d3dd95)*

ctest/src/generator.rs
```diff
@@ -723,6 +725,20 @@ impl TestGenerator {
...
+    /// Configures any extra arguments which to be passed to cargo
+    /// during macro expansion.  This can be used, for example, to pass
+    /// -Zbuild-std if required.
+    pub fn set_macro_expansion_cargo_args(&mut self, args: Vec<String>) {
+        self.macro_expansion_cargo_args = args;
+    }
+
+    /// Configures the crate name which should be used during macro expansion.
+    /// This can be important for crates which have crate_name! macros or
+    /// similar, like libc.
```
tgross35:
```suggestion
    ///
    /// If the tested crate uses `#![crate_name = "..."]`, this must be called with the
    /// same name. Otherwise, there will be an error about `--crate-name` not
    /// matching.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/483795464d6fbbd5fe07d59013f7c9e2bb46ec82..329498887b3317d812fafbf678acef83d5d3dd95)*

ctest/src/macro_expansion.rs
tgross35: That's perfect!

ctest/src/macro_expansion.rs
```diff
@@ -14,19 +13,63 @@ pub fn expand<P: AsRef<Path>>(
...
+    let cargo_toml = format!(
+        r#"
+[package]
+name = '''{crate_name}'''  # some crates require crate_name! to be usable
+edition = '''{EDITION}'''
+[lib]
+path = '''{}'''
+[lints.rust]
+unexpected_cfgs = "allow"
+"#,
```
tgross35, line 44: Could you add a `// FIXME(#5238): allow building using an existing manifest`? I think we may as well use the crate's `Cargo.toml`, if available, so the build script gets invoked.

https://github.com/rust-lang/libc/issues/5238

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/483795464d6fbbd5fe07d59013f7c9e2bb46ec82..329498887b3317d812fafbf678acef83d5d3dd95)*

ctest/src/macro_expansion.rs
```diff
@@ -14,19 +13,63 @@ pub fn expand<P: AsRef<Path>>(
...
+) -> Result<String> {
+    let dir = tempfile::tempdir()?;
+
+    // make a Cargo.toml pointing to the crate path
+    let cargo_toml_path = dir.path().join("Cargo.toml");
+    let crate_name = crate_name.unwrap_or("ctest-expansion-tmp");
+    let cargo_toml = format!(
+        r#"
+[package]
+name = '''{crate_name}'''  # some crates require crate_name! to be usable
```
tgross35: `crate_name!` isn't a builtin or a macro in this repo. Probably: "Needs to match `#![crate_name = "..."]` if it exists."

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/483795464d6fbbd5fe07d59013f7c9e2bb46ec82..329498887b3317d812fafbf678acef83d5d3dd95)*

ctest/Cargo.toml
tgross35: Since this moved to a regular dependency, the dev dependency can be dropped.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/483795464d6fbbd5fe07d59013f7c9e2bb46ec82..329498887b3317d812fafbf678acef83d5d3dd95)*

ctest/src/macro_expansion.rs
tgross35: Fair point about compatibility! We probably shouldn't suppress warnings though, there could be some papercuts.

I think it might be best to leave exact env up to the user, since sometimes build scripts pick up other env that you may want to control. So perhaps:

```rust
impl TestGenerator {
    fn expansion_env(&mut self, k: impl AsRef<OsStr>, v: impl AsRef<OsStr>) -> &mut Self;
    fn expansion_env_clear(&mut self) -> &mut Self;
}
```

Then call `cmd.env_clear().env(expansion_env)` on the `Command`.

skrap: But to what end?  This is adding API surface which no crate user has asked for, and it might or might not be the correct thing.  The previous code would emit warnings into `/dev/null` and then move on, and I think we should retain that behavior until some motivated person comes along to ask for it to change.

I would prefer we stick to the goal of this PR which is to enable tier 3 folks like me to build libc, and this API is beyond that goal.

ctest/src/generator.rs
skrap: done.

ctest/src/macro_expansion.rs
skrap: done.

ctest/Cargo.toml
skrap: done.

ctest/src/macro_expansion.rs
skrap: done.

ctest/src/generator.rs
skrap: done.

ctest/src/generator.rs
skrap: Curious why we would need a `clear()`.  Why would a crate user add arguments only to remove them?  Again I find myself concerned about increasing the API surface without a clear use case.

skrap:
@rustbot review

Further review feedback.  I'm a bit shy about adding more API than necessary for the stated purpose of the PR.  If we have a clear use case for the API proposed, I think that should be proposed in a separate PR.

ctest/src/generator.rs
```diff
@@ -723,6 +727,22 @@ impl TestGenerator {
...
+    /// This can be used, for example, to pass `-Zbuild-std`` if required.
+    pub fn macro_expansion_cargo_args(&mut self, args: Vec<String>) -> &mut Self {
+        self.macro_expansion_cargo_args = args;
+        self
+    }
+
+    /// Configures the crate name which should be used during macro expansion.
+    /// If the tested crate uses `#![crate_name = "..."]`, this must be called with the
+    /// same name. Otherwise, there will be an error about `--crate-name` not
+    /// matching.
```
tgross35, line 740:
```suggestion
    /// Configures the crate name which should be used during macro expansion.
    ///
    /// If the tested crate uses `#![crate_name = "..."]`, this must be called with the
    /// same name. Otherwise, there will be an error about `--crate-name` not
    /// matching.
```

so rustdoc doesn't mush into the summary

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/03ff348e8a626069954584b4bfc8f6d09c94841c..f752fb92e763ab0ef1acafc7f3ff33c529c53059)*

ctest/src/generator.rs
tgross35: Any kind of pattern that sets defaults then clears them, so you can just use store the args in the `TestGenerator` rather than tracking them separately. Not likely to be used a lot: it's just the sort of thing that's pretty easy to add now, but more likely to be worked around than added in the future.

Rather than clear, something like `expansion_cargo_args_mut(&mut self) -> &mut Vec<String>` would be fine too for more flexibility.

ctest/src/macro_expansion.rs
tgross35: If just calling `.env_clear()` works then I'm fine with that and no new API. But I don't want to hardcode any `RUSTFLAGS` - if those need to be adjusted then we should go with the suggested API.

We _should_ give users a chance to catch anything warnings that happen during expansion in case it's due to the config they're providing. Macro expansion can be env-dependent as well.

tgross35 (changes requested):
Just updating the status here.

Also, regarding the general concern about API surface area: don't worry about it :) as long as it's somewhat sane, for ctest I tend to err on the side of adding small things we might need while related code is already being modified, rather than leaving a gap that might need ctest-specific changes in the future. There might be somebody who appreciates it even if libc doesn't have an immediate use.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/03ff348e8a626069954584b4bfc8f6d09c94841c..f752fb92e763ab0ef1acafc7f3ff33c529c53059)*

ctest/src/macro_expansion.rs
skrap: done. I used `env_remove("RUSTFLAGS")` so as to not affect other env vars.

ctest/src/generator.rs
skrap: done, as you suggested with the accessor to the mutable member variable.

skrap:
@rustbot review

OK, I have gone with your suggestions on the remaining items.  Please have a look.  I will squash the commits into one before merging.

ctest/src/macro_expansion.rs
```diff
@@ -14,19 +13,63 @@ pub fn expand<P: AsRef<Path>>(
...
+        .arg("rustc")
+        .arg("--lib")
+        .arg("--profile")
+        .arg("check")
+        .args(extra_cargo_args);
+
+    // In order to avoid warnings causing failures here when the
+    // environment's RUSTFLAGS contains -Dwarnings (or similar),
+    // we will clear RUSTFLAGS.
+    cmd.env_remove("RUSTFLAGS");
```
tgross35, line 61: Change to `.env_clear()` so Cargo doesn't pick up more flags that are meant for the host.

ctest/src/generator.rs
```diff
@@ -723,6 +727,22 @@ impl TestGenerator {
         self
     }
 
+    /// Configures extra arguments to be passed to cargo during macro expansion.  
+    /// This can be used, for example, to pass `-Zbuild-std`` if required.
```
tgross35:
```suggestion
    /// This can be used, for example, to pass `-Zbuild-std` if required.
```

Mismatched backtick

ctest/src/generator.rs
```diff
@@ -723,6 +727,22 @@ impl TestGenerator {
...
+    pub fn macro_expansion_cargo_args_mut(&mut self) -> &mut Vec<String> {
```
tgross35:
```suggestion
    pub fn expansion_cargo_args_mut(&mut self) -> &mut Vec<String> {
```

Simpler name, there's only one expansion

tgross35 (changes requested):
Last pass, noticed a few small things then happy to merge

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/f69f0fd6eee6ac5058817c79da2a85b7cbcc07ee..c9e1f8af7a871065087d8b127314c73973ac6da5)*

ctest/src/macro_expansion.rs
skrap: I am guessing this will fail because it'll do away with things like `PATH`, which might be necessary for `cargo` to call the correct helpers, but I will do as you ask in the interest of moving this forward.

tgross35: Fair point, feel free to leave it for now if you haven't already tested and I'll poke around some more afterwards

ctest/src/generator.rs
skrap: done.

ctest/src/generator.rs
skrap: done.

ctest/src/macro_expansion.rs
skrap: thanks, leaving in place.

skrap:
@rustbot review

Applied the small changes requested and squashed the PR to a single commit.

tgross35 (approved):
LGTM, thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5214/df537bcc2ae6743b4b8cb4dfe471ba6f27006266..eadda1eaa48723f6500ea0e15dcb3569e27dbf9c)*

PR #5222, opened 2026-06-26
Bump libc-test to edition 2024

tgross35:
These only needs to build with recent Rust, so we may as well allow newer syntax.


tgross35:
The ctest tests need e2021

PR #5225, opened 2026-06-27
Add types and macros for ICMPv6 filtering

Skyb0rg007:
# Description

This commit adds the icmp6_filter type and the corresponding ICMP6_FILTER macros.
The API is defined in RFC 2292; I split-off this code from rust-lang/libc#4954 because the testing code here is more substantial.

These functions are tested in a similar manner to the CMSG macros, as the implementation differs between Linux and BSDs.

# Sources

https://github.com/bminor/glibc/blob/04e750e75b73957cf1c791535a3f4319534a52fc/inet/netinet/icmp6.h
https://github.com/kraj/musl/blob/kraj/master/include/netinet/icmp6.h
https://github.com/freebsd/freebsd-src/blob/9ae367d11de8abbdf53884836c9ba30908c5c8db/sys/netinet/icmp6.h
https://github.com/apple-oss-distributions/xnu/blob/main/bsd/netinet/icmp6.h

Also in the RFC: https://datatracker.ietf.org/doc/html/rfc2292#section-3.2

# Checklist


Skyb0rg007:
Sorry for spamming the CI; I couldn't get it to work on my fork.

Anyways, I'm mostly confident with this PR. The missing parts are:

1. A way to initialize `icmp6_filter`. Right now the only field is private, so you need to use an unsafe method like `std::mem::zeroed`. I don't think this is a big deal.
2. The FreeBSD's `netinet/icmp6.h` header uses `u_int32_t` types, which I couldn't figure out how to properly include. The test code therefore includes the typedefs, which is somewhat inelegant.

Some notes on the implementation:

- The `ICMP_FILTER` functions are marked as safe. I think this is correct, since the functions are all total (due to marking the `typ` argument as `u8`).
- The `icmp6_filter` struct does not export its field. The meaning of a set bit differs on Linux and BSD, so there shouldn't be a good reason to not use the accessor functions.
- The test cases are somewhat slow. It takes ~1s to run them, but its because I am testing all 255 types for pass/fail. This could possibly be improved by using `memcmp` or similar.

Skyb0rg007:
I'll tag @tgross35 since he reviewed my previous PR that this is split from.

Skyb0rg007:
@madsmtm Does iOS support ICMPv6 filter operations? I'm not sure if I need to mark the definitions or headers MacOS-only for the apple target.

Skyb0rg007:
Pinging @tgross35 again; it’s been ~2 months without a review. Making sure this PR hasn’t fallen through the cracks.

madsmtm:
iOS supports ICMPv6 filter operations.

Skyb0rg007:
Because this PR touches some volatile parts of the code such as the test runner, it would be great if this could be reviewed + merged so I don't need to keep resolving merge conflicts. @tgross35

PR #5224, opened 2026-06-26
linux: add seccomp user notification ioctl constants

fullzer4:
## Description

Add missing `SECCOMP_IOCTL_NOTIF_*` ioctl constants for Linux and Android.

These five ioctl constants are defined in the kernel's `linux/seccomp.h` and are used with the seccomp user notification API. The corresponding structures (`seccomp_notif`, `seccomp_notif_resp`, `seccomp_notif_addfd`) and flags were already added in PR rust-lang/libc#4330, but the ioctl numbers were missing.

All Linux platforms use the same ioctl numbers as they are defined using architecture-independent `_IOWR` and `_IOW` macros with a common magic number.

## Sources

- SECCOMP_IOCTL_NOTIF_* definitions: https://github.com/torvalds/linux/blob/master/include/uapi/linux/seccomp.h#L140-L155
- Related PR rust-lang/libc#4330 (added structures and flags): https://github.com/rust-lang/libc/pull/4330

## Checklist

- [x] Relevant tests in `libc-test/semver` have been updated
- [x] No placeholder or unstable values like `*LAST` or `*MAX` are included
- [x] Constants added to both Linux and Android targets

@rustbot label +stable-nominated

PR #5228, opened 2026-06-28
linux: Add SECCOMP_IOCTL_NOTIF_* ioctl numbers

lu-zero:
The userspace-notification structs and flags are already bound; only [the ioctl numbers][ref] were missing.

[ref]: https://github.com/torvalds/linux/blob/master/include/uapi/linux/seccomp.h#L147-L152


tgross35:
Was wondering why this closed, guess it raced with https://github.com/rust-lang/libc/pull/5224

PR #5237, opened 2026-07-01
Add tier 2 illumos CI 

### LINKED ISSUE #1405: Add CI for Solaris / Illumos
We probably should add CI for Illumos / Sun Solaris, or both. I have no idea about any of these two OSes, but I can help any interested party.

If there is no CI system like Travis, Azure Pipelines, Appveyor, or Cirrus-CI that provides native Solaris and/or Illumos VMs, the best way forward is to test Solaris and Illumos under Linux.

For testing other similar targets under Linux we cross-compile from Linux to the target (AFAIK this is possible for both Solaris and Illumos), and then use qemu or kvm to load a boot image, and run the tests there. See for example the s390x/sparc64 ci scripts, which use qemu-system (the old freebsd scripts used this as well), or this redox PR (https://github.com/rust-lang/libc/pull/1403) which uses the redoxer crate, which internally uses kvm. 

cc ing users that have contributed to these targets: @robertdfrench @bgermann @nbaksalyar @jasonbking @papertigers @vojtechkral @binarycrusader

papertigers:
Fixes: rust-lang/libc#1405

As discussed in rust-lang/libc#1405 we want CI support for illumos. It looks like solaris has already added CI support in a previous
commit so we can safely mark the issue as closed.

This adds illumos ci via the `vmactions/omnios-vm` github action and relies on the latest omnios LTS release (r151054) which is supported until 2028-05-01.

Additionally this PR fixes an epoll test issue that was looking for timerfd that I discovered while working on this.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -302,6 +302,7 @@ jobs:
             target: x86_64-unknown-freebsd
           - target: x86_64-pc-solaris
           - target: x86_64-unknown-netbsd
+          - target: x86_64-unknown-illumos
     timeout-minutes: 25
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
@@ -360,6 +361,27 @@ jobs:
             /usr/sbin/pkg_add curl
             curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- \
               --profile minimal --default-toolchain nightly -y
+          run: |
+            set -ex
+            . "$HOME/.cargo/env"
+            which rustc
+            rustc -Vv
+            ./ci/run.sh ${{ matrix.target }}
+
+      - name: test on illumos
+        uses: vmactions/omnios-vm@7f2be0b927aad1a78498c8aeeac4c4ce1fabd322 # v1.3.3
+        if: contains(matrix.target, 'illumos')
+        with:
+          # Use the latest LTS
+          release: "r151054-build"
+          usesh: true
+          mem: 4096
+          copyback: false
+          prepare: |
+            set -ex
+            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- \
+              --profile minimal --default-toolchain nightly -y
+            uname -a
           run: |
             set -ex
             . "$HOME/.cargo/env"
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -1052,7 +1052,7 @@ fn test_solarish(target: &str) {
     );
 
     if is_illumos {
-        headers!(cfg, "sys/epoll.h", "sys/eventfd.h",);
+        headers!(cfg, "sys/epoll.h", "sys/eventfd.h", "sys/timerfd.h",);
     }
 
     if is_solaris {
@@ -1084,6 +1084,8 @@ fn test_solarish(target: &str) {
                 // expose stat.Xtim.tv_nsec fields
                 Some(field.ident().trim_end_matches("e_nsec").to_string() + ".tv_nsec")
             }
+            // epoll_event.data is a union in C; our `u64` field lives at `data.u64`
+            "epoll_event" if field.ident() == "u64" => Some("data.u64".to_string()),
             _ => None,
         }
     });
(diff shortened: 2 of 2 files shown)


src/unix/solarish/mod.rs
```diff
@@ -1798,6 +1798,11 @@ pub const PTHREAD_RWLOCK_INITIALIZER: pthread_rwlock_t = pthread_rwlock_t {
 pub const PTHREAD_MUTEX_NORMAL: c_int = 0;
 pub const PTHREAD_MUTEX_ERRORCHECK: c_int = 2;
 pub const PTHREAD_MUTEX_RECURSIVE: c_int = 4;
+
+#[cfg(target_os = "illumos")]
+pub const PTHREAD_MUTEX_DEFAULT: c_int = 0x8;
```
papertigers: fwiw I am discussing this with a few people before we land this change.

This comes from: https://www.illumos.org/issues/16200, however this value is not in the current sysroot being used to cross compile rustc. There is a plan to [bump the illumos sysroot](https://github.com/illumos/sysroot/pull/5), but even this version will not contain the header value change. This means any software built with this libc change wouldn't function correctly on a sufficiently old illumos kernel. 

We could potentially keep the previous behavior and exclude the header check in the tests but we will have to remind ourselves to later go back and fix this in libc whenever we do have a sufficiently new sysroot.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5237/03ff348e8a626069954584b4bfc8f6d09c94841c..46a02de390df90b8b136704ada15be60b55aa216)*

iximeow: for "sufficiently old" meaning "before ~jan 2024 when 16200 was fixed", though - I think that points away from just changing libc! it's also "weird" to me that when building a normal release Rust, the sysroot there would have a `pthread.h` with `PTHREAD_MUTEX_DEFAULT` that disagrees with this file.

in some sense, the header check here is checking the _wrong_ `pthread.h` for anyone that uses the sync types via libstd, since whatever happens to be around for CI today is only relevant for anyone building libc today. but it's the _right_ `pthread.h` if your code goes and does pthread things, whereas libstd's pthread stuff could be lightly confusing in that case (though `PTHREAD_MUTEX_NORMAL` _didn't_ change so maybe that's all fine).

I've at least talked myself into thinking that leaving this as-is with a note for cutting a future sysroot to clean this up is probably the expedient thing to do?

tgross35: We don't appear to directly use `PTHREAD_MUTEX_DEFAULT` in rust-lang/rust so I'm not worried about this change.

See https://github.com/rust-lang/rust/issues/150756 for tracking the update.

tgross35 (approved):
LGTM but can you please improve the PR title and commit description?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5237/03ff348e8a626069954584b4bfc8f6d09c94841c..46a02de390df90b8b136704ada15be60b55aa216)*

papertigers:
@tgross35 I can change the title and description, however let's wait until others have a chance to chime in on the constant change included in this PR. Although it's likely I will push a change that implements @iximeow suggestion above which seems like the most sensible thing to do.

src/unix/solarish/mod.rs
papertigers: Adjusted in https://github.com/rust-lang/libc/pull/5237/commits/35dae153e77b17bc7abeb9c73a9e261746a05d80

iximeow (approved):
+1 for my part, but i dunno if @tgross35 would want one of the Known illumos People to also chime in

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5237/03ff348e8a626069954584b4bfc8f6d09c94841c..4cf9e8c1710efa2a24a5fa34b6c605b3b2c5ecab)*

tgross35:
I'm fine to take this, if we hear differently from them it's easy enough to revert.

@papertigers could you squash with the PR title/description as the commit message please? Unfortunately we don't have squash+merge on this repo.

papertigers:
@tgross35 I squashed the commits and amended the commit message. Should be good to go, let me know if we need to take care of anything else.

tgross35 (approved):
LGTM, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5237/a2051f2b28b5d0bda81eac91cbcd3178d9cfbb3f..ffebe47cf5e2ef612d8cf5f0947602b27861c9c3)*

PR #5240, opened 2026-07-02
Fix c_void_returns lint warning

jchlanda:
# Description

Nightly added a lint warning on `c_void` returns this patch removes the return type to implicitly return `()`.

See a sample failing job here: https://github.com/rust-lang/libc/actions/runs/28580626120/job/84739740809?pr=5065#step:9:2195

# Checklist

### DIFF
--- a/src/teeos/mod.rs
+++ b/src/teeos/mod.rs
@@ -49,7 +49,7 @@ pub type clockid_t = c_int;
 
 pub type suseconds_t = c_long;
 
-pub type once_fn = extern "C" fn() -> c_void;
+pub type once_fn = extern "C" fn();
 
 pub type pthread_once_t = c_int;
 
--- a/src/unix/aix/mod.rs
+++ b/src/unix/aix/mod.rs
@@ -2693,12 +2693,12 @@ extern "C" {
 
     pub fn pthread_cancel(thread: crate::pthread_t) -> c_int;
 
-    pub fn pthread_cleanup_pop(execute: c_int) -> c_void;
+    pub fn pthread_cleanup_pop(execute: c_int);
 
     pub fn pthread_cleanup_push(
         routine: Option<unsafe extern "C" fn(*mut c_void)>,
         arg: *mut c_void,
-    ) -> c_void;
+    );
 
     pub fn pthread_condattr_getclock(
         attr: *const pthread_condattr_t,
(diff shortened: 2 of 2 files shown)


jchlanda:
cc: @folkertdev

folkertdev:
there is one more at 

```diff
--- a/src/teeos/mod.rs
+++ b/src/teeos/mod.rs
@@ -49,7 +49,7 @@ pub type clockid_t = c_int;
 
 pub type suseconds_t = c_long;
 
-pub type once_fn = extern "C" fn() -> c_void;
+pub type once_fn = extern "C" fn();
 
 pub type pthread_once_t = c_int;

```

folkertdev:
also this is technically a breaking change? but the old signature was incorrect so idk.

tgross35:
> also this is technically a breaking change? but the old signature was incorrect so idk.

Technically yes, but the relevant platforms are all T3 and the return type is kind of  useless so this will be fine.

jchlanda:
> there is one more at
> 
> ```diff
> --- a/src/teeos/mod.rs
> +++ b/src/teeos/mod.rs
> @@ -49,7 +49,7 @@ pub type clockid_t = c_int;
>  
>  pub type suseconds_t = c_long;
>  
> -pub type once_fn = extern "C" fn() -> c_void;
> +pub type once_fn = extern "C" fn();
>  
>  pub type pthread_once_t = c_int;
> ```

I wasn't sure about that one, as opposed to the other places, this is a function pointer type. The compiler doesn't seem to be complaining about it: https://rust.godbolt.org/z/3xTezqxrY
And the jobs are all 🟢

tgross35:
I assume the lint just doesn't catch function pointer types (yet), it would be good to make that change here as well

folkertdev:
It's odd, based on the tests the lint should fire 

https://github.com/rust-lang/rust/blob/c397dae808f70caebab1fc4e11b3edf7e59f58c7/tests/ui/lint/c-void-returns.rs#L20

So maybe that file does not get run in CI? Anyhow, further evidence that we should make the change there too.

jchlanda:
Done now, thank you.

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5240/03ff348e8a626069954584b4bfc8f6d09c94841c..8b2a53341ad759215b59eeeb39f8625997ac4959)*

PR #5242, opened 2026-07-02
hurd: clean up module

dybucc:
# Description

This PR consists of an overall clean up of the GNU Hurd module. It adds proper
handling of types and routines when `_USE_FILE_OFFSET64` is defined by using
the existing `gnu_file_offset_bits64` `cfg`. It also fixes a type definition
and more closely mirrors those of upstream.

It also adds a number of future deprecation annotations. The annotations on
type aliases apply only to 64-bit targets. Other annotations apply to both
64-bit targets and any target with the afore mentioned `cfg` set.

Finally, there were some routines that were duplicated in both the `unix`
module and the `unix/hurd` module. These have been removed from the latter.
This decision was made because the `unix` module's definitions already included
the right link name redirection for bindings where the actual identifier is the
64-bit suffixed one.

## Sources

- glibc.
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/bits/types.h;h=33b582a8eabf09aadc093658db82353c288590d2;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/mach/hurd/bits/typesizes.h;h=514058071bfcea156584fa014e750b402216476b;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/resource.h;h=6b83374331424da439ad0163e1e992d274847fc2;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=libio/bits/types/__fpos_t.h;h=bb04576651b9097b3027e4299cc30c88f334535f;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=libio/bits/types/__fpos64_t.h;h=06a6891154fff74e1ddb6245f4a0467b09c617c5;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/mach/hurd/bits/fcntl.h;h=faa45a1e77e4fdae52cd002fe36bb5e4cd43eb98;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/bits/cpu-set.h;h=ddb79cefc2e5d93003f8da84eca6302f99153a22;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=misc/sys/uio.h;h=77f62ffcc3ac3c6a3d43e494c0062e823e6f624c;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/unistd.h;h=06a6a88b5272f7cc37d1df27f9e6617cdf58d69c;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=libio/stdio.h;h=3bf6a1f63203b111b67b6f101cc188320d3db78c;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=io/sys/sendfile.h;h=e51d19d15c85817970bc40cb51ebf79e0adb67c8;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=io/sys/stat.h;h=3069e187b079461dec1d5c699d49b5aa570a205d;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=io/sys/statfs.h;h=b4c7dc44e7db80dcdf3dca90457903da7ca4c9ba;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=io/sys/statvfs.h;h=6bc86c8335e106e5c993b5e5a2cd3d53eff12891;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=dirent/dirent.h;h=ed2aac64644b0f374aa1f0d1b0806094c6697d2b;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=string/string.h;h=743395b3e3b2dd52194b1a6b8c34df4f8b56d8a1;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=misc/sys/mman.h;h=e84308b1570bf26a071c52660803f1f2fd440944;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=io/fcntl.h;h=d0ad4d665227f30e7f2b2a49e3e823728d190a31;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=resource/sys/resource.h;h=cb76c2d8570b36127eb3a74dda1c6c8d1bb2afec;hb=HEAD>
  - <https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/glob.h;h=4c824eb0d30ec557f9c5d5e1d85d823e1a1900c6;hb=HEAD>

## Checklist

  (see [#3131](https://github.com/rust-lang/libc/issues/3131))

### DIFF
--- a/src/unix/hurd/b32.rs
+++ b/src/unix/hurd/b32.rs
@@ -15,14 +15,14 @@ pub type __u_quad_t = c_ulonglong;
 pub type __intmax_t = c_longlong;
 pub type __uintmax_t = c_ulonglong;
 
-pub type __squad_type = crate::__int64_t;
-pub type __uquad_type = crate::__uint64_t;
+pub type __squad_type = __int64_t;
+pub type __uquad_type = __uint64_t;
 pub type __sword_type = c_int;
 pub type __uword_type = c_uint;
 pub type __slong32_type = c_long;
 pub type __ulong32_type = c_ulong;
-pub type __s64_type = crate::__int64_t;
-pub type __u64_type = crate::__uint64_t;
+pub type __s64_type = __int64_t;
+pub type __u64_type = __uint64_t;
 
 pub type __ipc_pid_t = c_ushort;
 
--- a/src/unix/hurd/mod.rs
+++ b/src/unix/hurd/mod.rs
@@ -86,6 +86,7 @@ cfg_if! {
     }
 }
 
+// FIXME(1.0,deprecate,64): lfs binding to be removed
 pub type off64_t = __off64_t;
 pub type useconds_t = __useconds_t;
 pub type pid_t = __pid_t;
@@ -118,6 +119,7 @@ cfg_if! {
     }
 }
 
+// FIXME(1.0,deprecate,64): lfs binding to be removed
 pub type ino64_t = __ino64_t;
 pub type dev_t = __dev_t;
 pub type mode_t = __mode_t;
@@ -158,8 +160,11 @@ cfg_if! {
     }
 }
 
+// FIXME(1.0,deprecate,64): lfs binding to be removed
 pub type blkcnt64_t = __blkcnt64_t;
+// FIXME(1.0,deprecate,64): lfs binding to be removed
 pub type fsblkcnt64_t = __fsblkcnt64_t;
+// FIXME(1.0,deprecate,64): lfs binding to be removed
 pub type fsfilcnt64_t = __fsfilcnt64_t;
 
 pub type __pthread_spinlock_t = c_int;
@@ -199,6 +204,7 @@ cfg_if! {
     }
 }
 
+// FIXME(1.0,deprecate,64): lfs binding to be removed
 pub type rlim64_t = __rlim64_t;
 
 pub type __rusage_who = c_int;
@@ -260,6 +266,7 @@ pub type nl_item = c_int;
 pub type iconv_t = *mut c_void;
 
 extern_ty! {
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub type fpos64_t; // FIXME(hurd): fill this out with a struct
     pub type timezone;
 }
@@ -391,6 +398,7 @@ s! {
         pub d_name: [c_char; 1usize],
     }
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub struct dirent64 {
         pub d_ino: __ino64_t,
         pub d_reclen: c_ushort,
@@ -515,6 +523,7 @@ s! {
         >,
     }
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub struct stat64 {
         pub st_fstype: c_int,
         pub st_dev: __fsid_t, /* Actually st_fsid */
@@ -588,6 +597,7 @@ s! {
         f_spare: Padding<[c_uint; 3usize]>,
     }
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub struct statfs64 {
         pub f_type: c_uint,
         pub f_bsize: c_ulong,
@@ -620,6 +630,7 @@ s! {
         f_spare: Padding<[c_uint; 3usize]>,
     }
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub struct statvfs64 {
         pub __f_type: c_uint,
         pub f_bsize: c_ulong,
@@ -921,6 +932,7 @@ s! {
         pub domainname: [c_char; _UTSNAME_LENGTH],
     }
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub struct rlimit64 {
         pub rlim_cur: rlim64_t,
         pub rlim_max: rlim64_t,
@@ -959,6 +971,7 @@ s! {
         pub l_pid: __pid_t,
     }
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub struct flock64 {
         #[cfg(target_pointer_width = "32")]
         pub l_type: c_int,
@@ -1274,6 +1287,7 @@ cfg_if! {
     }
 }
 
+// FIXME(1.0,deprecate,64): lfs binding to be removed
 pub const RLIM64_INFINITY: rlim64_t = 0x7fffffffffffffff;
 pub const RLIM_SAVED_MAX: rlim_t = RLIM_INFINITY;
 pub const RLIM_SAVED_CUR: rlim_t = RLIM_INFINITY;
@@ -3630,7 +3644,9 @@ extern "C" {
 
     pub fn dup3(oldfd: c_int, newfd: c_int, flags: c_int) -> c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn pread64(fd: c_int, buf: *mut c_void, count: size_t, offset: off64_t) -> ssize_t;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn pwrite64(fd: c_int, buf: *const c_void, count: size_t, offset: off64_t) -> ssize_t;
 
     pub fn readv(__fd: c_int, __iovec: *const crate::iovec, __count: c_int) -> ssize_t;
@@ -3651,8 +3667,10 @@ extern "C" {
         __offset: off_t,
     ) -> ssize_t;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn preadv64(fd: c_int, iov: *const crate::iovec, iovcnt: c_int, offset: off64_t)
         -> ssize_t;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn pwritev64(
         fd: c_int,
         iov: *const crate::iovec,
@@ -3721,11 +3739,18 @@ extern "C" {
         oldattr: *mut crate::mq_attr,
     ) -> c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn lseek64(__fd: c_int, __offset: off64_t, __whence: c_int) -> off64_t;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
+    #[allow(clashing_extern_declarations)]
     pub fn fgetpos64(stream: *mut crate::FILE, ptr: *mut fpos64_t) -> c_int;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn fseeko64(stream: *mut crate::FILE, offset: off64_t, whence: c_int) -> c_int;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
+    #[allow(clashing_extern_declarations)]
     pub fn fsetpos64(stream: *mut crate::FILE, ptr: *const fpos64_t) -> c_int;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn ftello64(stream: *mut crate::FILE) -> off64_t;
 
     pub fn bind(__fd: c_int, __addr: *const sockaddr, __len: crate::socklen_t) -> c_int;
@@ -3759,6 +3784,7 @@ extern "C" {
 
     #[cfg_attr(gnu_file_offset_bits64, link_name = "sendfile64")]
     pub fn sendfile(out_fd: c_int, in_fd: c_int, offset: *mut off_t, count: size_t) -> ssize_t;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn sendfile64(out_fd: c_int, in_fd: c_int, offset: *mut off64_t, count: size_t) -> ssize_t;
 
     pub fn shutdown(__fd: c_int, __how: c_int) -> c_int;
@@ -4099,8 +4125,10 @@ extern "C" {
         old_value: *mut crate::itimerspec,
     ) -> c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn fstat64(__fd: c_int, __buf: *mut stat64) -> c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn fstatat64(
         __fd: c_int,
         __file: *const c_char,
@@ -4116,37 +4144,50 @@ extern "C" {
         statxbuf: *mut statx,
     ) -> c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn ftruncate64(__fd: c_int, __length: off64_t) -> c_int;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn truncate64(__file: *const c_char, __length: off64_t) -> c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn lstat64(__file: *const c_char, __buf: *mut stat64) -> c_int;
 
     #[cfg_attr(gnu_file_offset_bits64, link_name = "statfs64")]
     pub fn statfs(path: *const c_char, buf: *mut statfs) -> c_int;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn statfs64(__file: *const c_char, __buf: *mut statfs64) -> c_int;
     #[cfg_attr(gnu_file_offset_bits64, link_name = "fstatfs64")]
     pub fn fstatfs(fd: c_int, buf: *mut statfs) -> c_int;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn fstatfs64(__fildes: c_int, __buf: *mut statfs64) -> c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn statvfs64(__file: *const c_char, __buf: *mut statvfs64) -> c_int;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn fstatvfs64(__fildes: c_int, __buf: *mut statvfs64) -> c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn open64(__file: *const c_char, __oflag: c_int, ...) -> c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn openat64(__fd: c_int, __file: *const c_char, __oflag: c_int, ...) -> c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn fopen64(filename: *const c_char, mode: *const c_char) -> *mut crate::FILE;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn freopen64(
         filename: *const c_char,
         mode: *const c_char,
         file: *mut crate::FILE,
     ) -> *mut crate::FILE;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn creat64(path: *const c_char, mode: mode_t) -> c_int;
 
     pub fn mkostemp(template: *mut c_char, flags: c_int) -> c_int;
     pub fn mkostemps(template: *mut c_char, suffixlen: c_int, flags: c_int) -> c_int;
     pub fn mkstemps(template: *mut c_char, suffixlen: c_int) -> c_int;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn tmpfile64() -> *mut crate::FILE;
 
     pub fn popen(command: *const c_char, mode: *const c_char) -> *mut crate::FILE;
@@ -4299,9 +4340,12 @@ extern "C" {
 
     pub fn faccessat(dirfd: c_int, pathname: *const c_char, mode: c_int, flags: c_int) -> c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn stat64(__file: *const c_char, __buf: *mut stat64) -> c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn readdir64(dirp: *mut crate::DIR) -> *mut dirent64;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn readdir64_r(
         dirp: *mut crate::DIR,
         entry: *mut dirent64,
@@ -4317,6 +4361,7 @@ extern "C" {
 
     pub fn __errno_location() -> *mut c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn mmap64(
         __addr: *mut c_void,
         __len: size_t,
@@ -4345,11 +4390,13 @@ extern "C" {
     pub fn fallocate64(fd: c_int, mode: c_int, offset: off64_t, len: off64_t) -> c_int;
     #[cfg_attr(gnu_file_offset_bits64, link_name = "posix_fallocate64")]
     pub fn posix_fallocate(fd: c_int, offset: off_t, len: off_t) -> c_int;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn posix_fallocate64(fd: c_int, offset: off64_t, len: off64_t) -> c_int;
 
     #[cfg_attr(gnu_file_offset_bits64, link_name = "posix_fadvise64")]
     pub fn posix_fadvise(fd: c_int, offset: off_t, len: off_t, advise: c_int) -> c_int;
 
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn posix_fadvise64(fd: c_int, offset: off64_t, len: off64_t, advise: c_int) -> c_int;
 
     pub fn madvise(__addr: *mut c_void, __len: size_t, __advice: c_int) -> c_int;
@@ -4358,9 +4405,11 @@ extern "C" {
 
     #[cfg_attr(gnu_file_offset_bits64, link_name = "getrlimit64")]
     pub fn getrlimit(resource: __rlimit_resource_t, rlim: *mut crate::rlimit) -> c_int;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn getrlimit64(resource: __rlimit_resource_t, rlim: *mut rlimit64) -> c_int;
     #[cfg_attr(gnu_file_offset_bits64, link_name = "setrlimit64")]
     pub fn setrlimit(resource: __rlimit_resource_t, rlim: *const crate::rlimit) -> c_int;
+    // FIXME(1.0,deprecate,64): lfs binding to be removed
     pub fn setrlimit64(resource: __rlimit_resource_t, rlim: *const rlimit64) -> c_int;
 
     pub fn getpriority(which: crate::__priority_which, who: crate::id_t) -> c_int;
--- a/src/unix/hurd/mod.rs
+++ b/src/unix/hurd/mod.rs
@@ -1068,6 +1068,23 @@ impl siginfo_t {
     }
 }
 
+s_no_extra_traits! {
+    pub struct fpos_t {
+        __pos: off_t,
+        __state: __mbstate_t,
+    }
+
+    struct __mbstate_t {
+        __count: c_int,
+        __value: __c_anonymous___mbstate_t___value,
+    }
+
+    union __c_anonymous___mbstate_t___value {
+        __wch: c_int,
+        __wchb: [c_char; 4],
+    }
+}
+
 // const
 
 // aio.h
--- a/src/unix/hurd/mod.rs
+++ b/src/unix/hurd/mod.rs
@@ -4341,6 +4341,7 @@ extern "C" {
     pub fn syncfs(fd: c_int) -> c_int;
     pub fn fdatasync(fd: c_int) -> c_int;
 
+    #[deprecated(since = "0.2.187", note = "This routine doesn't exist upstream.")]
     pub fn fallocate64(fd: c_int, mode: c_int, offset: off64_t, len: off64_t) -> c_int;
     #[cfg_attr(gnu_file_offset_bits64, link_name = "posix_fallocate64")]
     pub fn posix_fallocate(fd: c_int, offset: off_t, len: off_t) -> c_int;
--- a/src/unix/hurd/mod.rs
+++ b/src/unix/hurd/mod.rs
@@ -3717,9 +3717,7 @@ extern "C" {
         oldattr: *mut crate::mq_attr,
     ) -> c_int;
 
-    pub fn lseek64(__fd: c_int, __offset: __off64_t, __whence: c_int) -> __off64_t;
-
-    pub fn lseek(__fd: c_int, __offset: __off_t, __whence: c_int) -> __off_t;
+    pub fn lseek64(__fd: c_int, __offset: off64_t, __whence: c_int) -> off64_t;
 
     pub fn fgetpos64(stream: *mut crate::FILE, ptr: *mut fpos64_t) -> c_int;
     pub fn fseeko64(stream: *mut crate::FILE, offset: off64_t, whence: c_int) -> c_int;
@@ -4097,10 +4095,8 @@ extern "C" {
         old_value: *mut crate::itimerspec,
     ) -> c_int;
 
-    pub fn fstat(__fd: c_int, __buf: *mut stat) -> c_int;
     pub fn fstat64(__fd: c_int, __buf: *mut stat64) -> c_int;
 
-    pub fn fstatat(__fd: c_int, __file: *const c_char, __buf: *mut stat, __flag: c_int) -> c_int;
     pub fn fstatat64(
         __fd: c_int,
         __file: *const c_char,
@@ -4116,11 +4112,9 @@ extern "C" {
         statxbuf: *mut statx,
     ) -> c_int;
 
-    pub fn ftruncate(__fd: c_int, __length: __off_t) -> c_int;
-    pub fn ftruncate64(__fd: c_int, __length: __off64_t) -> c_int;
-    pub fn truncate64(__file: *const c_char, __length: __off64_t) -> c_int;
+    pub fn ftruncate64(__fd: c_int, __length: off64_t) -> c_int;
+    pub fn truncate64(__file: *const c_char, __length: off64_t) -> c_int;
 
-    pub fn lstat(__file: *const c_char, __buf: *mut stat) -> c_int;
     pub fn lstat64(__file: *const c_char, __buf: *mut stat64) -> c_int;
 
     #[cfg_attr(gnu_file_offset_bits64, link_name = "statfs64")]
@@ -4130,15 +4124,11 @@ extern "C" {
     pub fn fstatfs(fd: c_int, buf: *mut statfs) -> c_int;
     pub fn fstatfs64(__fildes: c_int, __buf: *mut statfs64) -> c_int;
 
-    pub fn statvfs(__file: *const c_char, __buf: *mut statvfs) -> c_int;
     pub fn statvfs64(__file: *const c_char, __buf: *mut statvfs64) -> c_int;
-    pub fn fstatvfs(__fildes: c_int, __buf: *mut statvfs) -> c_int;
     pub fn fstatvfs64(__fildes: c_int, __buf: *mut statvfs64) -> c_int;
 
-    pub fn open(__file: *const c_char, __oflag: c_int, ...) -> c_int;
     pub fn open64(__file: *const c_char, __oflag: c_int, ...) -> c_int;
 
-    pub fn openat(__fd: c_int, __file: *const c_char, __oflag: c_int, ...) -> c_int;
     pub fn openat64(__fd: c_int, __file: *const c_char, __oflag: c_int, ...) -> c_int;
 
     pub fn fopen64(filename: *const c_char, mode: *const c_char) -> *mut crate::FILE;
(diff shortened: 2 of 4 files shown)


src/unix/hurd/mod.rs
tgross35, line 1: "hurd: remove duplicate routines": LGTM @ 62ccc9111b636c1d5b25ed2edf14294d842fe0f7

src/unix/hurd/mod.rs
```diff
@@ -889,29 +950,17 @@ s! {
...
-        #[cfg(target_pointer_width = "32")]
         pub l_whence: c_int,
-        #[cfg(target_pointer_width = "64")]
-        pub l_type: c_short,
-        #[cfg(target_pointer_width = "64")]
-        pub l_whence: c_short,
-        pub l_start: __off_t,
+        pub l_start: __off64_t,
         pub l_len: __off64_t,
         pub l_pid: __pid_t,
```
tgross35, line 988: "hurd: use `gnu_file_offset_bits64` to expose lfs": Is this accurate? Looking at https://github.com/bminor/glibc/blob/04e750e75b73957cf1c791535a3f4319534a52fc/sysdeps/mach/hurd/bits/types/struct_flock.h and https://github.com/bminor/glibc/blob/04e750e75b73957cf1c791535a3f4319534a52fc/sysdeps/mach/hurd/i386/bits/types/struct_flock.h, it looks like `c_short` is the right type on non-x86-32.

tgross35 (commented):
I just saw the one thing then noticed the links, are you sure that's the most recent source? Per https://cgit.git.savannah.gnu.org/cgit/hurd/glibc.git/log/?h=t/sysvshm the branch you're linking hasn't been updated since 2021, and none of the branches at https://cgit.git.savannah.gnu.org/cgit/hurd/glibc.git/refs/heads are any newer.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/d1413164cb40f71bdf88f2ce95e491a3959f322d..f16b383de7662fb2a6ee87b732b8ed28360f8d72)*

tgross35:
Yeah - the author of https://github.com/rust-lang/libc/pull/4127 is the lead Hurd developer as I understand it, and that PR links to the regular glibc sourceware repos. I assume that at some point the deviations were resolved and they just switched to developing upstream.

@rustbot author since there's probably some things to revisit with that in mind.

src/unix/hurd/mod.rs
tgross35, line 1: "hurd: deprecate lfs types and routines": Note that like the others, I am not comfortable deprecating the time bindings while users don't have a clean-cut solution.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/d1413164cb40f71bdf88f2ce95e491a3959f322d..f16b383de7662fb2a6ee87b732b8ed28360f8d72)*

src/unix/hurd/mod.rs
dybucc: just reviewed it with the updated glibc sources.

src/unix/hurd/mod.rs
dybucc: so, to a first approximation, how would a clean-cut solution look like?

src/unix/hurd/b32.rs
```diff
@@ -15,14 +15,14 @@ pub type __u_quad_t = c_ulonglong;
 pub type __intmax_t = c_longlong;
 pub type __uintmax_t = c_ulonglong;
 
-pub type __squad_type = crate::__int64_t;
-pub type __uquad_type = crate::__uint64_t;
+pub type __squad_type = __int64_t;
+pub type __uquad_type = __uint64_t;
```
tgross35, line 19: hurd(b32): simplify item paths: LGTM @ acfbb703c6bdf3c018d90b6fc3c657d4d94f5293

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/5848d319e53719d637781eae72de169d8f7867a0..7159a66d83e71c49f2d4c428af776ee9f93e1c40)*

src/unix/hurd/mod.rs
```diff
@@ -232,6 +232,16 @@ extern_ty! {
...
 s! {
+    pub struct fpos_t {
+        pub __pos: off_t,
+        pub __state: __mbstate_t,
+    }
+
+    pub struct __mbstate_t {
+        pub __count: c_int,
+        pub __value: __c_anonymous___mbstate_t___value,
+    }
```
tgross35: "hurd: fill in definition for fpos_t": let's match linux-gnu and keep these fields private, similarly the union.

https://github.com/rust-lang/libc/blob/328f30939c117b1871303a7fcbb30e247cfb990a/src/unix/linux_like/linux/gnu/mod.rs#L343

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/5848d319e53719d637781eae72de169d8f7867a0..7159a66d83e71c49f2d4c428af776ee9f93e1c40)*

src/unix/hurd/mod.rs
tgross35, line 1: "hurd: add proper handling of lfs": looks pretty reasonable to me but I'd like the maintainer to ack

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/5848d319e53719d637781eae72de169d8f7867a0..7159a66d83e71c49f2d4c428af776ee9f93e1c40)*

src/unix/hurd/mod.rs
```diff
@@ -230,6 +264,26 @@ extern_ty! {
...
+    } else {
+        const STAT__SPARE_SIZE: usize = {
+            if size_of::<__fsid_t>() == size_of::<c_int>() {
+                9
+            } else {
+                8
+            }
+        };
+    }
+}
```
tgross35: "hurd: add proper handling of lfs": nit, you can avoid the `cfg-if` here and instead do `... : usize = { if cfg!(gnu_file_offset_bits64) && size_of::<__fsid_t>() == size_of::<c_int>() { ... } ` to save a nesting.

Also, looks like there's an extra `_` between `STAT` and `SPARE` (unless this is in the source?)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/5848d319e53719d637781eae72de169d8f7867a0..7159a66d83e71c49f2d4c428af776ee9f93e1c40)*

src/unix/hurd/mod.rs
```diff
@@ -488,14 +542,20 @@ s! {
...
         pub st_author: __uid_t,
         pub st_flags: c_uint,
-        pub st_spare: [c_int; 8usize],
+        st_spare: Padding<
+            [c_int; if size_of::<__fsid_t>() == size_of::<c_int>() {
+                9
+            } else {
+                8
+            }],
+        >,
```
tgross35, line 551: "hurd: add proper handling of lfs": just put this in a `const STAT64_SPARE_SIZE` like the other, in case we need to reuse that value

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/5848d319e53719d637781eae72de169d8f7867a0..7159a66d83e71c49f2d4c428af776ee9f93e1c40)*

src/unix/hurd/mod.rs
```diff
@@ -1029,10 +1029,7 @@ s! {
     }
 
     pub struct cpu_set_t {
-        #[cfg(all(target_pointer_width = "32", not(target_arch = "x86_64")))]
-        bits: [u32; 32],
-        #[cfg(not(all(target_pointer_width = "32", not(target_arch = "x86_64"))))]
-        bits: [u64; 16],
+        __bits: [__cpu_mask; 1024 / (8 * size_of::<__cpu_mask>())],
```
tgross35: "hurd: adjust cpu_set_t definition": use `__CPU_SETSIZE` and `__NCPUBITS` intermediates like they do rather than hardcoding, just as nonpublic constants.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/5848d319e53719d637781eae72de169d8f7867a0..7159a66d83e71c49f2d4c428af776ee9f93e1c40)*

src/unix/hurd/mod.rs
tgross35, line 1: "hurd: remove duplicate definitions": lgtm @ 8b36f4b296c0c7e1ba57c6d471cc7fcad4fad741

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/5848d319e53719d637781eae72de169d8f7867a0..7159a66d83e71c49f2d4c428af776ee9f93e1c40)*

tgross35: I think that what's at https://github.com/rust-lang/libc/issues/4805 is the best bet, then we can find+replace these back once everything else is ready.

src/unix/hurd/mod.rs
dybucc: done.

src/unix/hurd/mod.rs
dybucc: done.

the identifier upstream is simply _SPARE_SIZE. i decided to add another underscore because there's already a "canonical" underscore used for separation between STAT and the actual identifier. considering this is not a public symbol, i decided to opt out of making it match upstream.

src/unix/hurd/mod.rs
dybucc: done.

src/unix/hurd/mod.rs
dybucc: decided i would make them public. __CPU_SETSIZE was already public before this pr so i made __NCPUBITS also public.

dybucc:
@sthibaul would you mind reviewing the changes concerning lfs?

sthibaul:
> We assume upstream to be the GNU Hurd's fork of glibc hosted at [1](https://github.com/rust-lang/libc/pull/5242#user-content-fn-1-587999cee7fda3074f6ffbb47a032308).

No, it's not.... The last commit there was 6 years ago...

src/unix/hurd/mod.rs
```diff
@@ -448,25 +572,32 @@ s! {
 
     pub struct stat {
         pub st_fstype: c_int,
-        pub st_dev: __fsid_t, /* Actually st_fsid */
-        pub st_ino: __ino_t,
+        pub st_fsid: __fsid_t,
```
sthibaul: No, we really want `st_dev` here, to match what the rest of the world expects. `st_fsid` is the real name in the structure, but `st_dev` is an alias for it, and that's what people expect. See https://github.com/rust-lang/libc/pull/3785

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/70c67fe2ec83eda97646b49cab2bcbd1603a65ea..15c0ee77769272af35bbee234ebfb7dc10f62bf3)*

src/unix/hurd/mod.rs
sthibaul: yes, only i386 is to keep the non-posix-compatible int type for l_type/whence

src/unix/hurd/mod.rs
```diff
@@ -4219,38 +4596,62 @@ extern "C" {
...
-        result: *mut *mut crate::dirent64,
+        entry: *mut dirent64,
+        result: *mut *mut dirent64,
     ) -> c_int;
     pub fn seekdir(dirp: *mut crate::DIR, loc: c_long);
     pub fn telldir(dirp: *mut crate::DIR) -> c_long;
 
     pub fn dirfd(dirp: *mut crate::DIR) -> c_int;
 
-    #[link_name = "__xpg_strerror_r"]
```
sthibaul, line 4242: Why dropping this?
See https://sourceware.org/git/?p=glibc.git;a=blob;f=string/string.h;h=743395b3e3b2dd52194b1a6b8c34df4f8b56d8a1;hb=HEAD#l464 and linux does the same (except on musl, which is not ported to GNU/Hurd)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/70c67fe2ec83eda97646b49cab2bcbd1603a65ea..15c0ee77769272af35bbee234ebfb7dc10f62bf3)*

src/unix/hurd/mod.rs
```diff
@@ -4377,21 +4815,53 @@ extern "C" {
 
     pub fn regfree(preg: *mut crate::regex_t);
 
+    #[cfg_attr(
+        all(gnu_time_bits64, not(gnu_file_offset_bits64)),
+        link_name = "__glob64_time64"
```
sthibaul, line 4508: ? this doesn't match the conditions of https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/glob.h;h=4c824eb0d30ec557f9c5d5e1d85d823e1a1900c6;hb=HEAD#l158

And note that GNU/Hurd doesn't define `__USE_TIME64_REDIRECTS`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/70c67fe2ec83eda97646b49cab2bcbd1603a65ea..15c0ee77769272af35bbee234ebfb7dc10f62bf3)*

src/unix/hurd/mod.rs
```diff
@@ -4377,21 +4815,53 @@ extern "C" {
...
         flags: c_int,
         errfunc: Option<extern "C" fn(epath: *const c_char, errno: c_int) -> c_int>,
-        pglob: *mut crate::glob_t,
+        pglob: *mut glob_t,
     ) -> c_int;
-    pub fn globfree(pglob: *mut crate::glob_t);
-
+    #[cfg_attr(
+        all(gnu_time_bits64, not(gnu_file_offset_bits64)),
+        link_name = "__globfree64_time64"
```
sthibaul, line 4522: ditto

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/70c67fe2ec83eda97646b49cab2bcbd1603a65ea..15c0ee77769272af35bbee234ebfb7dc10f62bf3)*

src/unix/hurd/mod.rs
sthibaul: I'm a bit surprised to see these deprecation warnings indeed, since the Linux port does not have them on its `*64` variants.

sthibaul: Claiming `The definitions are equivalent` when compiling on a 64bit target is wrong: yes on the 64b target they are equivalent, but then the programmer will migrate to the non-64 version and thusly break the 32bit builds unless `gnu_file_offset_bits64` is used. So at most the warnings should be enabled only when the programmer did request for `gnu_file_offset_bits64`, in which case, yes, using the `*64` variants is spurious.

sthibaul: but again, better do this on all targets (Linux too notably) altogether rather than just Hurd

src/unix/hurd/mod.rs
```diff
@@ -4394,21 +4471,45 @@ extern "C" {
...
+    )]
+    #[cfg_attr(
+        all(gnu_file_offset_bits64, not(gnu_time_bits64)),
+        link_name = "globfree64"
+    )]
+    pub fn globfree(pglob: *mut glob_t);
+
+    #[cfg_attr(
+        all(gnu_time_bits64, not(gnu_file_offset_bits64)),
+        link_name = "__glob64_time64"
```
sthibaul: ditto

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/70c67fe2ec83eda97646b49cab2bcbd1603a65ea..15c0ee77769272af35bbee234ebfb7dc10f62bf3)*

src/unix/hurd/mod.rs
```diff
@@ -4394,21 +4471,45 @@ extern "C" {
...
         pattern: *const c_char,
         flags: c_int,
         errfunc: Option<extern "C" fn(epath: *const c_char, errno: c_int) -> c_int>,
-        pglob: *mut glob64_t,
+        pglob: *mut glob_t,
     ) -> c_int;
-    pub fn globfree64(pglob: *mut glob64_t);
+    #[cfg_attr(
+        all(gnu_time_bits64, not(gnu_file_offset_bits64)),
+        link_name = "__globfree64_time64"
```
sthibaul: ditto

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/70c67fe2ec83eda97646b49cab2bcbd1603a65ea..15c0ee77769272af35bbee234ebfb7dc10f62bf3)*

sthibaul:
Thanks for this work!

I'd rather actually test this natively before letting it merged, unfortunately I won't have time to do this before at least tuesday.

src/unix/hurd/mod.rs
```diff
@@ -226,12 +302,53 @@ pub type nl_item = c_int;
...
+        deprecated(
+            since = "0.2.187",
+            note = "Use `fpos_t` instead. Their definitions are equivalent."
+        )
+    )]
     pub type fpos64_t; // FIXME(hurd): fill this out with a struct
     pub type timezone;
 }
 
+const STAT__SPARE_SIZE: usize = {
```
sthibaul: it's a bit sad to make these far from the structure that uses it. Is it not possible to embed the condition in the structure itself? for instance by emitting several st_spare definitions, each conditioned appropriately. We don't really strictly need to define a `STAT__SPARE_SIZE` condition  just to mimick what glibc uses, `_SPARE_SIZE` is not part of the API.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/70c67fe2ec83eda97646b49cab2bcbd1603a65ea..15c0ee77769272af35bbee234ebfb7dc10f62bf3)*

src/unix/hurd/mod.rs
dybucc: done.

src/unix/hurd/mod.rs
dybucc: did that because that one declaration exists when !__USE_GNU [^1], but throughout the libc crate there's definitions affecting the hurd which assume __USE_GNU.

[^1]: https://sourceware.org/git/?p=glibc.git;a=blob;f=string/string.h;h=743395b3e3b2dd52194b1a6b8c34df4f8b56d8a1;hb=HEAD#l460

src/unix/hurd/mod.rs
dybucc: fixed. there's one thing to note, though: we recently made it so that our cfgs to handle redirects of time64 and fileoffset64 could not be set separately (though at present we only issue a warning at compile time.) this may conflict with the behavior in upstream glibc where you seem to be capable of opting into one or the other.

src/unix/hurd/mod.rs
dybucc: those constants weren't public anyway. i've inlined the logic in the structure now.

dybucc:
> > We assume upstream to be the GNU Hurd's fork of glibc hosted at [1](https://github.com/rust-lang/libc/pull/5242#user-content-fn-1-587999cee7fda3074f6ffbb47a032308).
> 
> No, it's not.... The last commit there was 6 years ago...

i forgot to update the pr description, but i was also told about that little after submitting the patches so the current state should reflect upstream glibc.

src/unix/hurd/mod.rs
sthibaul: assuming gnu functions are available is not the same as assuming that it's the gnu behavior that is desired.

Really, the linux crate does use the xpg redirection: https://github.com/rust-lang/libc/blob/main/src/unix/linux_like/linux_l4re_shared.rs#L1699 , and thus programs expect the xpg behavior, not the gnu behavior.

dybucc: i still don't get it. the linux definitions don't affect the hurd but i'm assuming you mean that as an example. why should we provide bindings to the non-gnu version if we have assumed otherwise in other parts of the codebase? granted, the current signature would have to change to that of the gnu version, but that can be readily done as we should be capable of breaking the api on tier 3 targets.

src/unix/hurd/mod.rs
dybucc: those deprecations have now been replaced with comments. they were made before the plan in #4805 had been formulated. now they're just comments tagging some item for future deprecation.

also,

> Claiming The definitions are equivalent when compiling on a 64bit target is wrong: yes on the 64b target they are equivalent, but then the programmer will migrate to the non-64 version and thusly break the 32bit builds unless gnu_file_offset_bits64 is used. So at most the warnings should be enabled only when the programmer did request for gnu_file_offset_bits64, in which case, yes, using the *64 variants is spurious.

aren't libc crate users just meant to take care of having the unsuffixed types be used on 64-bit targets, and of having gnu_file_offset_bits64 set on 32-bit targets?

if they're building for multiple targets, surely they must take into account platform differences.

src/unix/hurd/mod.rs
sthibaul: > the linux definitions don't affect the hurd

Sure. But the linux definition affects what people believe the correct behavior is supposed to be.

And in glibc, the definition of strerror_r is exactly the same on Linux and on the Hurd, so it makes full sense to expose the same interface in the linux crate and in the hurd crate.

> why should we provide bindings to the non-gnu version if we have assumed otherwise in other parts of the codebase?

Because as I said above, “assuming gnu functions are available is not the same as assuming that it's the gnu behavior that is desired”. And I indeed took the Linux case as an example of this, but it's not only that: it's the very same function, with the very same file inside glibc. So really no reason to diverge from Linux on this.

src/unix/hurd/mod.rs
sthibaul: > if they're building for multiple targets, surely they must take into account platform differences.

That's theory...

In practice, people don't think about everything, only really test on 64b platforms, etc. And notably:

> of having gnu_file_offset_bits64 set on 32-bit targets?

AFAIK that's a new thing, not everybody will be doing this for a long time still, so we have to care for them not getting shot in the foot too much.

src/unix/hurd/mod.rs
```diff
@@ -929,6 +1011,10 @@ s! {
         __unused5: Padding<*mut c_void>,
     }
 
+    #[deprecated(
+        since = "0.2.187",
+        note = "Use `glob_t` instead. The definitions are equivalent."
+    )]
```
sthibaul, line 1007: is this a leftover?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/70c67fe2ec83eda97646b49cab2bcbd1603a65ea..0e14b9bd2904aaeae4a337c5fbd9e9b52bcd3ae0)*

sthibaul: I see that it's an explicit commit fd267188100b754df30f29d46cd4f0fd286cc078

But here again, the same happens in linux. The cleanup should really be done together in a separate PR.

I mean: linux and hurd use the same glibc, only a few functions really differ. We should really avoid divergence as much as possible in rust-libc since it's just the same codebase behind ; the more we diverge, the more difficult it becomes to maintain them.

src/unix/hurd/mod.rs
dybucc: Noted. I believe the plan at #4805 should better align with having library
users get more comfortable over time with these options.

src/unix/hurd/mod.rs
dybucc: So IIUC, we're skipping the actual definition to match the Linux definition we
already expose so that users aren't surprised by the GNU/Hurd definitions
differing? If so, done.

src/unix/hurd/mod.rs
dybucc: We don't have a `glob_t` on Linux [^1]. Open to getting rid of these
deprecations, but I think it's best if I don't add a type nobody has requested.

[^1]:
    https://github.com/search?q=repo%3Arust-lang%2Flibc+path%3A**%2Flinux_like%2F**%2F*.rs+%2F%5Cbglob%5Cd*_t%5Cb%2F&type=code

sthibaul: > We don't have a glob_t on Linux [1](https://github.com/rust-lang/libc/pull/5242#user-content-fn-1-33ee3ea0911f9109a97f640b3184d8d9).

Mmm, then I am lost. Your deprecation warning is telling people to use `glob_t`, but Linux does not provide one? Are people supposed to either use glob_t or glob64_t depending on the OS? (only hurd and linux seem to be providing glob64_t)

src/unix/hurd/mod.rs
sthibaul: > by the GNU/Hurd definitions differing?

The GNU/Hurd definition of `strerror_r` does *not* differ from the GNU/Linux definition. Both Linux and Hurd implementations of `strerror_r` come from the same glibc source code.

> done

Thanks.

dybucc: So we're basically assumming that both GNU/Linux and the GNU/Hurd do not have
`_GNU_SOURCE` defined and would always want the POSIX version of this routine?
I understand that this seems to be the default from [^manpage], but I'm not
sure.

[^manpage]: https://linux.die.net/man/3/strerror_r

sthibaul: > So we're basically assumming that both GNU/Linux and the GNU/Hurd do not have
_GNU_SOURCE defined and would always want the POSIX version of this routine?

I'm basically assuming that what people will want is what people are currently getting: *all* ports are currently returning an int, i.e. the posix behavior. So better just align on that, otherwise we will get unexpected issues.

Really, in general any divergence from what the rest of the world is currently doing only brings troubles.

src/unix/hurd/mod.rs
dybucc: So it seems. That's why I mentioned in my last message that I was hesitant to
add another type (referring to `glob_t` on Linux.)

You think that I should remove the deprecation concerning the GNU/Hurd bindings
until somebody requests `glob_t` on Linux and then I can deprecate them all?

sthibaul: > You think

At this point, I don't think anything. I'm *very* surprised that `glob_t` not be available on Linux but `glob64_t` be.

If anything, I think we should align on what is already there for other OSes, and not try to clean up the Hurd crate more than the rest, and rather clean all creates on this point in the same PR when it becomes clear what is aimed for.

sthibaul: I did a quick search in the debian source code, there does not see to be any rust package there that calls `globfree` or `globfree64`, so it does look like the functions have never been really used, and thus the issue never actually raised. But possibly that is so *just because* it is not available on Linux (which is really a gold reference to people), and if it was available on Linux, people would start using it.

In coherence with the lfs cleanup, we should indeed aim for telling people to use glob_t. But I don't see why doing so in a different way than the other lfs functions?

dybucc: The `glob_t` matter is separate from the LFS matters because these are
concerned with non-immediate changes that require first documenting them with a
small `FIXME` and later on (hopefully a few months from now) deprecating them.
This is because file offset types are way too many and have too much potential
impact on end users to deprecate in a single go.

`glob64_t`, on the other hand, is more easily "deprecatable." It's only on the
GNU/Hurd bindings and it's already got a perfect equivalent. I believe this
type we can already deprecate and eventually remove in favor of `glob_t`.

sthibaul: > glob64_t, on the other hand, is more easily "deprecatable." It's only on the GNU/Hurd bindings

No, it's also in the Linux bindings.

src/unix/hurd/mod.rs
```diff
@@ -4377,21 +4519,47 @@ extern "C" {
...
     ) -> c_int;
-    pub fn globfree(pglob: *mut crate::glob_t);
-
+    #[cfg_attr(
+        all(gnu_file_offset_bits64, gnu_time_bits64),
+        link_name = "__glob64_time64"
+    )]
+    #[cfg_attr(
+        all(gnu_file_offset_bits64, not(gnu_time_bits64)),
+        link_name = "glob64"
```
sthibaul: the link_names must be glob64free, not glob64

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/df537bcc2ae6743b4b8cb4dfe471ba6f27006266..f26c265d8570b7ab03d7144cc6d5c0ca22dac1c6)*

src/unix/hurd/mod.rs
```diff
@@ -4377,21 +4519,47 @@ extern "C" {
...
+    #[deprecated(
+        since = "0.2.187",
+        note = "Use `glob` instead. Their definitions are equivalent."
+    )]
     pub fn glob64(
         pattern: *const c_char,
         flags: c_int,
         errfunc: Option<extern "C" fn(epath: *const c_char, errno: c_int) -> c_int>,
-        pglob: *mut glob64_t,
+        pglob: *mut glob_t,
```
sthibaul: the type must remain glob64_t, otherwise this will be bogus in the non-gnu_file_offset_bits64 case

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/df537bcc2ae6743b4b8cb4dfe471ba6f27006266..f26c265d8570b7ab03d7144cc6d5c0ca22dac1c6)*

src/unix/hurd/mod.rs
```diff
@@ -4377,21 +4519,47 @@ extern "C" {
...
     ) -> c_int;
-    pub fn globfree64(pglob: *mut glob64_t);
+    #[cfg_attr(gnu_time_bits64, link_name = "__globfree64_time64")]
+    #[deprecated(
+        since = "0.2.187",
+        note = "Use `globfree` instead. Their definitions are equivalent."
+    )]
+    pub fn globfree64(pglob: *mut glob_t);
```
sthibaul: the type must remain glob64_t, otherwise this will be bogus in the non-gnu_file_offset_bits64 case

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/df537bcc2ae6743b4b8cb4dfe471ba6f27006266..f26c265d8570b7ab03d7144cc6d5c0ca22dac1c6)*

sthibaul (changes requested):

src/unix/hurd/mod.rs
dybucc: Forgive my wording; I meant to say that the only bindings in which both
`glob_t` and `glob64_t` are exposed are those of the GNU/Hurd. That's why we
can simply go ahead and deprecate the latter; Users have an alternative with
`glob_t`.

src/unix/hurd/mod.rs
dybucc: I believe it's actually none of those; `globfree64` seems to be the right
choice [^1]. Fixed now.

[^1]:
    https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/glob.h;h=4c824eb0d30ec557f9c5d5e1d85d823e1a1900c6;hb=HEAD#l161

src/unix/hurd/mod.rs
dybucc: Will it, now? For all we care on the Rust side of things, the types for
`glob_t` and `glob64_t` are the same across FFI boundaries. The function
pointer fields are the only ones that differ upstream, but those we currently
choose to expose as both `*mut c_void` and private.

src/unix/hurd/mod.rs
dybucc: See [^answer].

[^answer]: https://github.com/rust-lang/libc/pull/5242#discussion_r3608017509

src/unix/hurd/mod.rs
sthibaul: But don't we want to enforce through typing that a glob_t returned by glob is freed by globfree, and a glob64_t returned by glob64 is freed by globfree64?

dybucc: I mean, this is a set of C bindings. Actually providing good typing would
likely involve setting up a higher-level abstraction that enforces relatively
well-defined semantics. But this is not really a point of contention for me, so
I can change it if it seems more coherent to you.

sthibaul: Yes, I believe coherency is important, to avoid people getting surprised by what they read

dybucc: Done.

sthibaul (approved):
I have run the rustc build testsuite, it didn't show any regression with this patch applied.

Thanks for your work!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/ee8c04103d69ab696d8c1bbc26da7b18fbc01d2b..ea4abb751f9ba1b93431b87d5e558fba6c9aa4d9)*

sthibaul:
@tgross35 I didn't find in the documentation what more I should be doing to tell that this looks ready to me to go in, so just pinging  to make sure it doesn't stay stuck :)

dybucc:
> I have run the rustc build testsuite, it didn't show any regression with this patch applied.

I'm not sure if the `rustc` test suite is enough to prove these changes
relatively correct. Have you runt the `libc-test` suite in this repo?

sthibaul:
> > I have run the rustc build testsuite, it didn't show any regression with this patch applied.
> 
> I'm not sure if the rustc test suite is enough to prove these changes relatively correct. Have you runt the `libc-test` suite in this repo?

The `rustc` testsuite does exercise quite a lot, yes.

`cargo test` at the root doesn't seem to be testing much.  In `libc-test` there seems to be more content, but not that much, and it seems it wasn't ported to the Hurd target.

dybucc:
I don't think the tests in `libc-test` are meant to be specifically run on all
systems. They're also a very small fraction of the actual tests. `libc-test`
uses `ctest`, which generates at compile-time (when running `cargo test` under
the `libc-test` directory) tests to match all Rust symbols with all C symbols in
search of, say, possible function pointer mismatches. Those aren't "found"
anywhere; They're generated on the fly from scanning the Rust files under `src`.

Though of course, you may have to use some of the `ctest` API to skip certain
bindings like we do with other systems.

src/unix/hurd/mod.rs
```diff
@@ -1068,6 +1078,31 @@ impl siginfo_t {
     }
 }
 
+s_no_extra_traits! {
+    pub union __c_anonymous___mbstate_t___value {
+        __wch: c_int,
+        __wchb: [c_char; 4],
+    }
+}
```
tgross35: "hurd: fill in definition for fpos_t": This need not be public

src/unix/hurd/mod.rs
```diff
@@ -232,6 +232,16 @@ extern_ty! {
...
 s! {
+    pub struct fpos_t {
+        __pos: off_t,
+        __state: __mbstate_t,
+    }
+
+    pub struct __mbstate_t {
+        __count: c_int,
+        __value: __c_anonymous___mbstate_t___value,
+    }
```
tgross35: "hurd: fill in definition for fpos_t": `__mbstate_t` doesn't need to be public

This is also adding traits that we don't need, put them in `s_no_extra_traits` which will allow dropping the custom impls for `__c_anonymous___mbstate_t___value`.

src/unix/hurd/mod.rs
```diff
@@ -4412,20 +4486,38 @@ extern "C" {
...
+
+    #[cfg_attr(gnu_time_bits64, link_name = "__glob64_time64")]
     pub fn glob64(
         pattern: *const c_char,
         flags: c_int,
         errfunc: Option<extern "C" fn(epath: *const c_char, errno: c_int) -> c_int>,
         pglob: *mut glob64_t,
     ) -> c_int;
+    #[cfg_attr(gnu_time_bits64, link_name = "__globfree64_time64")]
     pub fn globfree64(pglob: *mut glob64_t);
```
tgross35, line 4548: "hurd: add proper handling of lfs": could you clarify the interaction with `gnu_time_bits64` here? time64 bits aren't otherwise touched in this PR, `time_t` remains a `long`, so is there going to be a mismatch somewhere?

At first glance I'd expect dropping the `gnu_time_bits64` portions from here and including them in a followup time64 PR would be more accurate, but I'm not sure what the exact effects are.

tgross35 (commented):
Just a few followup bits. Thank you @sthibaul for a very thorough review!

> @tgross35 I didn't find in the documentation what more I should be doing to tell that this looks ready to me to go in, so just pinging to make sure it doesn't stay stuck :)

Removing the label and pinging is all :) I tend to go through libc PRs in batches so sometimes it takes me a bit.

> `cargo test` at the root doesn't seem to be testing much. In `libc-test` there seems to be more content, but not that much, and it seems it wasn't ported to the Hurd target.

By default `cargo test` only runs the root crate which is pretty much nothing. `libc-test` is really what we rely on since it looks at all the public API and compares the Rust and C versions (using [`ctest`](https://docs.rs/ctest/latest/ctest/)). I wouldn't be surprised if this was never run on Hurd though, unfortunately (indeed there is no `test_hurd` function).

If it did work then running once as-is then, once time64 support lands, once with `RUSTFLAGS="--cfg=libc_unstable_gnu_time_bits=\"64\""` would be the thing to do (that's the relevant part of https://github.com/rust-lang/libc/blob/47a05e94e2448665b27eb715e7f23f13d7c152bd/ci/run.sh).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/ee8c04103d69ab696d8c1bbc26da7b18fbc01d2b..ea4abb751f9ba1b93431b87d5e558fba6c9aa4d9)*

src/unix/hurd/mod.rs
dybucc: Done.

src/unix/hurd/mod.rs
dybucc: Done.

src/unix/hurd/mod.rs
dybucc: It would make a difference if we chose to expose the function pointer fields of
the `glob_t` type. Currently, we have them as non-public `Padding` fields with
the size and alignment requirement of a `void *`. That's also why we can
deprecate `glob64_t` and advise in favor of using `glob_t`. These types differ
upstream only in their function pointer fields.

The actual function pointers upstream take `stat`s, which itself contains one of
`timespec` or `time_t` (depending on some feature test macros.) `timespec`
contains `time_t` fields, so yes, whether `gnu_time_bits64` is set or not would
make a difference if we ever chose to expose those fields of either (if not
both) `glob_t` and `glob64_t`.

That's currently not the case, and I assumed it would continue being so.

tgross35: Makes sense; if it's standalone and doesn't interact with other user-facing time64 bits then SGTM

tgross35 (approved):
Thanks! CI may take a couple of tries

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5242/5db5914d7ecf8d3fe32394d315f29012039630ca..9952e403aa52d1f41a506e9d7ea179025b922b57)*

PR #5243, opened 2026-07-03
netbsd: Fix statvfs struct layout and add symbol versioning for NetBSD 10.x

fraggerfox:
# Description

Fixes the `statvfs` structure layout and adds symbol versioning for NetBSD `10.x` (and above) compatibility.

> **Note**: AI assistance was used to discover the fixes as well as to test the fixes within a NetBSD environment.

NetBSD 10.0 changed the `statvfs` structure:
- `f_spare` changed from `[u32; 4]` to `[u64; 4]`
- Added `f_mntfromlabel` field (1024 bytes)
- Total size increased from 2256 to 3296 bytes

Without explicit `link_name` attributes, Rust bindings resolve to compatibility wrappers (`__compat_statvfs`) that expect the old struct layout, causing data corruption even with the correct struct definition.
- `statvfs()` → `__statvfs190`
- `fstatvfs()` → `__fstatvfs190`
- `getmntinfo()` → `__getmntinfo90`

> **Note:** This is a **breaking change for NetBSD 9.x** and earlier. Those versions use `__statvfs90`/`__fstatvfs90` which return the old struct format without `f_mntfromlabel`. NetBSD 11 is expected to release later this year, making NetBSD 9.x increasingly obsolete.
>
> https://endoflife.date/netbsd

## Previous changes

https://github.com/rust-lang/libc/commit/1816f6061

This PR removed the deprecated symbol, which was previously added by @0-wiz-0 and tries to maintain compatibility from NetBSD `9.x` and above.

The current PR, bumps compatibility for NetBSD `10.x` and above.

## Discovery

The problem with `getmntinfo()` was discovered when debugging why [bottom](https://github.com/ClementTsang/bottom) was not populating the `Disks` panel correctly in NetBSD.
- https://github.com/ClementTsang/bottom/issues/640#issuecomment-3925868116

# Sources

NetBSD `statvfs` structure definition:
- https://github.com/NetBSD/src/blob/8648f7862110109be2c2a5c4b9218819efe3b17f/sys/sys/statvfs.h#L66-L99

NetBSD symbol versioning documentation:
- https://man.netbsd.org/versioningsyscalls.9

# Checklist

  - No updates needed: `statvfs`/`fstatvfs` already in `unix.txt`, `getmntinfo` already in `netbsd.txt`
- [x] Tested locally (`cd libc-test && cargo test --target mytarget`)
  - Tested on NetBSD 10.1 x86_64: `CC=cc cargo test --target x86_64-unknown-netbsd --workspace`
  - All 3 test configurations pass (4822 ctest verifications each)

cc: @0323pin

### DIFF
--- a/src/new/netbsd/sys/statvfs.rs
+++ b/src/new/netbsd/sys/statvfs.rs
@@ -35,13 +35,11 @@ s! {
         pub f_namemax: c_ulong,
         pub f_owner: crate::uid_t,
 
-        // This type is updated in a future version
-        f_spare: [u32; 4],
+        f_spare: [u64; 4],
 
         pub f_fstypename: [c_char; _VFS_NAMELEN],
         pub f_mntonname: [c_char; _VFS_MNAMELEN],
         pub f_mntfromname: [c_char; _VFS_MNAMELEN],
-        // Added in NetBSD10
-        // pub f_mntfromlabel: [c_char; _VFS_MNAMELEN],
+        pub f_mntfromlabel: [c_char; _VFS_MNAMELEN],
     }
 }
--- a/src/new/netbsd/sys/statvfs.rs
+++ b/src/new/netbsd/sys/statvfs.rs
@@ -35,13 +35,11 @@ s! {
         pub f_namemax: c_ulong,
         pub f_owner: crate::uid_t,
 
-        // This type is updated in a future version
-        f_spare: Padding<[u32; 4]>,
+        f_spare: [u64; 4],
 
         pub f_fstypename: [c_char; _VFS_NAMELEN],
         pub f_mntonname: [c_char; _VFS_MNAMELEN],
         pub f_mntfromname: [c_char; _VFS_MNAMELEN],
-        // Added in NetBSD10
-        // pub f_mntfromlabel: [c_char; _VFS_MNAMELEN],
+        pub f_mntfromlabel: [c_char; _VFS_MNAMELEN],
     }
 }
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -2249,7 +2249,7 @@ extern "C" {
         ntargets: size_t,
         hint: *const c_void,
     ) -> c_int;
-    #[link_name = "__getmntinfo13"]
+    #[link_name = "__getmntinfo90"]
     pub fn getmntinfo(mntbufp: *mut *mut crate::statvfs, flags: c_int) -> c_int;
     pub fn getvfsstat(buf: *mut crate::statvfs, bufsize: size_t, flags: c_int) -> c_int;
 
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -2255,7 +2255,7 @@ extern "C" {
         ntargets: size_t,
         hint: *const c_void,
     ) -> c_int;
-    #[link_name = "__getmntinfo13"]
+    #[link_name = "__getmntinfo90"]
     pub fn getmntinfo(mntbufp: *mut *mut crate::statvfs, flags: c_int) -> c_int;
     pub fn getvfsstat(buf: *mut crate::statvfs, bufsize: size_t, flags: c_int) -> c_int;
 
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -1619,8 +1619,10 @@ extern "C" {
     pub fn sem_trywait(sem: *mut sem_t) -> c_int;
     pub fn sem_post(sem: *mut sem_t) -> c_int;
     #[cfg_attr(gnu_file_offset_bits64, link_name = "statvfs64")]
+    #[cfg_attr(target_os = "netbsd", link_name = "__statvfs190")]
     pub fn statvfs(path: *const c_char, buf: *mut crate::statvfs) -> c_int;
     #[cfg_attr(gnu_file_offset_bits64, link_name = "fstatvfs64")]
+    #[cfg_attr(target_os = "netbsd", link_name = "__fstatvfs190")]
     pub fn fstatvfs(fd: c_int, buf: *mut crate::statvfs) -> c_int;
 
     #[cfg_attr(target_os = "netbsd", link_name = "__sigemptyset14")]
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -1667,9 +1667,11 @@ extern "C" {
     pub fn sem_post(sem: *mut sem_t) -> c_int;
     #[cfg(not(all(target_os = "linux", target_env = "gnu")))] // defined in new/glibc
     #[cfg_attr(gnu_file_offset_bits64, link_name = "statvfs64")]
+    #[cfg_attr(target_os = "netbsd", link_name = "__statvfs190")]
     pub fn statvfs(path: *const c_char, buf: *mut crate::statvfs) -> c_int;
     #[cfg(not(all(target_os = "linux", target_env = "gnu")))] // defined in new/glibc
     #[cfg_attr(gnu_file_offset_bits64, link_name = "fstatvfs64")]
+    #[cfg_attr(target_os = "netbsd", link_name = "__fstatvfs190")]
     pub fn fstatvfs(fd: c_int, buf: *mut crate::statvfs) -> c_int;
 
     #[cfg_attr(target_os = "netbsd", link_name = "__sigemptyset14")]


0323pin:
LGTM
NetBSD-9 is living it's last days and I don't think we should stop this because of it. It's more important to support 10 and future versions of the OS.

@fraggerfox can you please add the _stable-nominated_ flag to this PR so it gets added to the stable version of `libc`?
Thank you for working on this and thanks to @0-wiz-0 for reviewing.

src/new/netbsd/sys/statvfs.rs
```diff
@@ -35,13 +35,12 @@ s! {
         pub f_namemax: c_ulong,
         pub f_owner: crate::uid_t,
 
-        // This type is updated in a future version
-        f_spare: [u32; 4],
+        __pad1: Padding<u32>,
+        pub f_spare: [u64; 4],
```
tgross35: No need to make this public. Where does `__pad1` come from?

tgross35 (changes requested):
Please handwrite all PR descriptions and commit messages, there's no need to restate the diff and copy the C header. Do make sure to permalink the headers (currently `statvfs.h` is a regular link).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5243/d1413164cb40f71bdf88f2ce95e491a3959f322d..00b5fdde003457c0829178ab82ddec9119ba6d4e)*

tgross35:
I'm actually not sure we should do this; per https://doc.rust-lang.org/rustc/platform-support/netbsd.html we support down to 8.x in some cases.

0323pin:
> I'm actually not sure we should do this; per https://doc.rust-lang.org/rustc/platform-support/netbsd.html we support down to 8.x in some cases.

At any given moment in time, NetBSD only supports the current latest + latest-1. 8 is officially **not** supported anymore. Besides, Rust doesn't even build on 8.

9 will follow the same path as soon as 11 is released. But, do you @0-wiz-0 feel differently about it?

fraggerfox:
> Please handwrite all PR descriptions and commit messages, there's no need to restate the diff and copy the C header. Do make sure to permalink the headers (currently `statvfs.h` is a regular link).
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5243/d1413164cb40f71bdf88f2ce95e491a3959f322d..00b5fdde003457c0829178ab82ddec9119ba6d4e)_

This was handwritten, I put in stuff there for context, but I can always cut down on the content.

I put in the link to the `trunk` based on the examples given in the PR template. It has been changed  it to the hashed one.

The reason I put in a verbose description is because I spent quite a lot of time tracking down the historical changes done for NetBSD previously for making this PR (tracking down the exact commit was not easy either)
- https://github.com/rust-lang/libc/pull/4782

The aim of the description was so that people in the future can find such a change  done more easily.

src/new/netbsd/sys/statvfs.rs
fraggerfox: I have removed the `pub` from `f_spare`.

Without the padding between `f_owner` and `f_spare` caused the Rust struct to be misaligned by 4 bytes, here is the output of the sample program that I used for testing (generated by AI)

```
=== STRUCT LAYOUT COMPARISON ===

WITH padding (correct):
  Total size:      3296 bytes
  f_owner offset:  152
  f_spare offset:  160
  Gap after owner: 4 bytes

WITHOUT padding (packed, wrong):
  Total size:      3292 bytes
  f_owner offset:  152
  f_spare offset:  156
  Gap after owner: 0 bytes

=== CALLING statvfs() WITH BOTH STRUCTS ===

CORRECT struct (with padding):
  f_fstypename:   'ffs'
  f_mntonname:    '/'
  f_mntfromname:  '/dev/ld0a'
  f_spare[0]:     0x0000000000000000

PACKED struct (without padding - WRONG):
  f_fstypename:   ''
  f_mntonname:    ''
  f_mntfromname:  ''
  f_spare[0]:     0x0000000000000000
```

The size of the struct when taken from a native C program matches `3296` bytes.

fraggerfox: ~Hmm, I will need to debug further to see why the CI is failing without the `pub`. will do that once I get a bit more free time.~

Seems like a transient failure, @tgross35 can you help retrigger the CI?
- https://github.com/rust-lang/libc/actions/runs/28645611965/job/84951914997?pr=5243#step:5:243

```
  + /usr/sbin/pkg_add curl
  pkg_add: Can't process [http://cdn.NetBSD.org:80/pub/pkgsrc/packages/NetBSD/x86_64/10.1/All//curl*:](http://cdn.netbsd.org/pub/pkgsrc/packages/NetBSD/x86_64/10.1/All//curl*:) 
  pkg_add: no pkg found for 'curl', sorry.
  pkg_add: 1 package addition failed
```

tgross35: > Without the padding between `f_owner` and `f_spare` caused the Rust struct to be misaligned by 4 bytes, here is the output of the sample program that I used for testing (generated by AI)

That doesn't really track - why would they be mismatched? There are no alignment attributes on either, so did another field change sizes or something?

> Seems like a transient failure, @tgross35 can you help retrigger the CI?
> 
>     * https://github.com/rust-lang/libc/actions/runs/28645611965/job/84951914997?pr=5243#step:5:243
> 
> 
> ```
>   + /usr/sbin/pkg_add curl
>   pkg_add: Can't process [http://cdn.NetBSD.org:80/pub/pkgsrc/packages/NetBSD/x86_64/10.1/All//curl*:](http://cdn.netbsd.org/pub/pkgsrc/packages/NetBSD/x86_64/10.1/All//curl*:) 
>   pkg_add: no pkg found for 'curl', sorry.
>   pkg_add: 1 package addition failed
> ```

We hit this error all the time, I don't know why. @0323pin do you happen to have any ideas here?

tgross35:
> > I'm actually not sure we should do this; per https://doc.rust-lang.org/rustc/platform-support/netbsd.html we support down to 8.x in some cases.
> 
> At any given moment in time, NetBSD only supports the current latest + latest-1. 8 is officially **not** supported anymore. Besides, Rust doesn't even build on 8.
> 
> 9 will follow the same path as soon as 11 is released. But, do you @0-wiz-0 feel differently about it?

The policy seems reasonable to me, would be good to update the platform docs to clarify N-1 is supported.

In that case though, we should hold off here right? If it was just the field changes I wouldn't be concerned, but the link name changes will cause link failures for anybody building on NetBSD 9.

It's also possible to do something like FreeBSD and add an unstable config to toggle things. We really need https://github.com/rust-lang/rfcs/pull/3905...

src/new/netbsd/sys/statvfs.rs
fraggerfox: > > Without the padding between `f_owner` and `f_spare` caused the Rust struct to be misaligned by 4 bytes, here is the output of the sample program that I used for testing (generated by AI)
> 
> That doesn't really track - why would they be mismatched? There are no alignment attributes on either, so did another field change sizes or something?
> 

Nothing changed, this is definitely an alignment issue of the types in struct. My best guess is `gcc` is probably padding it in someway when generating the struct, hence the size mismatch between C and Rust. I am not too familiar with how `gcc` pads the struct (neither in Rust), but I can investigate and get back.

tgross35: GCC doesn't insert padding out of nowhere, it uses the same layout calculations as rust with `#[repr(C)]`. If you just run `libc-test` without the padding it may give a hint.

fraggerfox: @tgross35 : Alright, let me run these tests in a NetBSD environment and get back to you.

0323pin:
@tgross35 

> We hit this error all the time, I don't know why. @0323pin do you happen to have any ideas here?

Actually, I don't know ... `curl` is there.

Screenshot_20260703_191105_Opera Mini beta

tgross35:
In case that's just how network errors show up, I'm going to try a retry with https://github.com/rust-lang/libc/pull/5246.

src/new/netbsd/sys/statvfs.rs
fraggerfox: @tgross35 : I have removed the explict padding. It seems I had a bug in implementation for `bottom` that was reading the values wrongly (since I was using `bottom` as a test harness to check if `libc` changes were working).

I wrote an independent test directly calling `statvfs` struct and it seems to work, my mistake was the small test program I was working with was explicitly recreating the structs and checking if field was it was correct and there was a mistake in that approach.

TL;DR, I have removed the explicit padding, the CI also looks green, thank you @tgross35  for pushing me to check it correctly.

tgross35: Note that we aren't actually testing this type so CI should be green regardless :) https://github.com/rust-lang/libc/blob/e98332b8c8add6fe9e4484db714c5b4202ffbf3a/libc-test/build.rs#L1427-L1428.

If you're interested, I think you could move that from `skip_struct` to `skip_struct_field` and only exclude the fields that differ among versions so we test the rest. But do that in a separate PR preferably.

tgross35:
This looks technically correct at this point, but it sounds like 9.x is still supported so I think we should hold off, given AIUI there isn't any rush. So for this to be ready I think we need:

1. The 11.0 release, which should also come with the deprecation of 9.x
2. An update to https://doc.rust-lang.org/nightly/rustc/platform-support/netbsd.html changing 9.x to 10.x, preferably also just stating a supported version policy

Seems like 11.0 is in RC so let's revisit in a couple months? Unless @0323pin you feel strongly that there's a reason to ship this now.

@rustbot blocked
@SnoozeThis wait 3 months -> remove label S-blocked, add label S-waiting-on-review

SnoozeThis:
(https://snoozeth.is/wmRMPHBukiA) I will wait until Sun, 04 Oct 2026 10:20:42 UTC and then add label S-waiting-on-review and remove label S-blocked.

@rustbot claim.

fraggerfox:
> This looks technically correct at this point, but it sounds like 9.x is still supported so I think we should hold off, given AIUI there isn't any rush. So for this to be ready I think we need:
> 

Good to know and thank you!

>     1. The 11.0 release, which should also come with the deprecation of 9.x
> 
>     2. An update to https://doc.rust-lang.org/nightly/rustc/platform-support/netbsd.html changing 9.x to 10.x, preferably also just stating a supported version policy
> 
> 
> Seems like 11.0 is in RC so let's revisit in a couple months? Unless @0323pin you feel strongly that there's a reason to ship this now.
> 

I am fine with maintaining this patch for now in the `bottom` package itself (in pkgsrc) until this is merged and a release is made.

tgross35:
What exactly is the issue in `bottom`? I don't see `getmntinfo` or `statvfs` mentioned in the linked thread, does it just need the `f_mntfromlabel` field for some reason?

fraggerfox:
> What exactly is the issue in `bottom`? I don't see `getmntinfo` or `statvfs` mentioned in the linked thread, does it just need the `f_mntfromlabel` field for some reason?

`bottom`  uses `sysinfo` crate which uses the `libc` crate to call `getmntinfo()`
- https://github.com/ClementTsang/bottom/blob/main/src/collection.rs#L363 (the condition would include `netbsd` to enable disk status collection for that)
  - `self.sys.disks.refresh(true);` calls `sysinfo` crate.
- https://github.com/GuillaumeGomez/sysinfo/blob/main/src/unix/bsd/netbsd/disk.rs#L295
  - `let count = unsafe { libc::getmntinfo(&mut mnt_buf, libc::MNT_WAIT) };` calls `libc` crate

The `f_mntfromlabel` is not used, but `f_mntfromname` is needed so that the disk panel is populated correctly

- Disks panel showing the correct values after the fixes.
  - image

- Process panel showing the Reads / Writes per process level
  - image

tgross35:
I'm still not sure I follow - `f_mntfromname` appears present in the initial addition of `statvfs` 23 years ago https://github.com/NetBSD/src/commit/6bd1d6d4db7ef2a6921d8113541e0d2483395e84#diff-190233bad1103409b2b5a94638ddb03aa71a0c6f3b84923668a56e8695a9eb87

0323pin:
@tgross35
> Unless @0323pin you feel strongly that there's a reason to ship this now.

I don't mind waiting but to me 9 is nearly dead and doesn't feel like a valid waiting reason. But, I still would like to have a second opinion on this. So, asking @0-wiz-0 once more 😉

tgross35:
It's fine by me, especially since it looks like 9.0 _is_ pretty old, but would prefer https://github.com/rust-lang/rust/blob/main/src/doc/rustc/src/platform-support/netbsd.md gets updated first. Usually that's where we point people if they have questions about what we try to support if there's ever any version confusion.

fraggerfox:
> I'm still not sure I follow - `f_mntfromname` appears present in the initial addition of `statvfs` 23 years ago [NetBSD/src@6bd1d6d#diff-190233bad1103409b2b5a94638ddb03aa71a0c6f3b84923668a56e8695a9eb87](https://github.com/NetBSD/src/commit/6bd1d6d4db7ef2a6921d8113541e0d2483395e84#diff-190233bad1103409b2b5a94638ddb03aa71a0c6f3b84923668a56e8695a9eb87)

The problem is not `statvfs` struct itself but the `getmntinfo()` call, in Rust the unversioned call results in invoking `__compat_getmntinfo()` which returns a different struct called `statfs12`
- https://github.com/NetBSD/src/blob/trunk/lib/libc/compat/gen/compat_getmntinfo.c
- https://github.com/NetBSD/src/blob/trunk/sys/compat/sys/mount.h#L43

My understanding from @0-wiz-0 's explanation when were discussing this is, the `__strong_alias(getmntinfo, __compat_getmntinfo)` causes Rust to link to the `__compat_getmntinfo` (since Rust does not read C headers). 

I am not sure if the explanation is sufficient and answers your question.

*Note: I do not have much experience dealing with linkers, so most of my understanding is based on reading the code, so there may be mistakes*

tgross35:
> The problem is not `statvfs` struct itself but the `getmntinfo()` call, in Rust the unversioned call results in invoking `__compat_getmntinfo()` which returns a different struct called `statfs12`
> 
>     * https://github.com/NetBSD/src/blob/trunk/lib/libc/compat/gen/compat_getmntinfo.c
>     * https://github.com/NetBSD/src/blob/trunk/sys/compat/sys/mount.h#L43
> 
> My understanding from @0-wiz-0 's explanation when were discussing this is, the `__strong_alias(getmntinfo, __compat_getmntinfo)` causes Rust to link to the `__compat_getmntinfo` (since Rust does not read C headers).

Ah! That makes a lot more sense, thanks. How is anything using `getmntinfo` even working at all? It looks like `f_bsize` lines up but pretty much every other field is mismatched between `statfs12` and `statvfs90`/`statvfs`.

Could you just link `__getmntinfo13` here? That should be present in the 9.0 tree https://github.com/NetBSD/src/blob/bc7bd704f7c4458f83e06fa841ecf05c3b23a547/sys/sys/statvfs.h#L155 and the `statvfs` definition matches our current struct https://github.com/NetBSD/src/blob/bc7bd704f7c4458f83e06fa841ecf05c3b23a547/sys/sys/statvfs.h#L66. I think that would mean the completely broken stuff gets fixed without jumping all the way to `__getmntinfo90` (the only benefit of which is `f_mntfromlabel` IIUC).

fraggerfox:
> 
> Ah! That makes a lot more sense, thanks. How is anything using `getmntinfo` even working at all? It looks like `f_bsize` lines up but pretty much every other field is mismatched between `statfs12` and `statvfs90`/`statvfs`.
> 

I am not sure how many consumers of that call are there in pkgsrc, and more importantly who all are using those packages :sweat_smile: 

> Could you just link `__getmntinfo13` here? That should be present in the 9.0 tree https://github.com/NetBSD/src/blob/bc7bd704f7c4458f83e06fa841ecf05c3b23a547/sys/sys/statvfs.h#L155 and the `statvfs` definition matches our current struct https://github.com/NetBSD/src/blob/bc7bd704f7c4458f83e06fa841ecf05c3b23a547/sys/sys/statvfs.h#L66. I think that would mean the completely broken stuff gets fixed without jumping all the way to `__getmntinfo90` (the only benefit of which is `f_mntfromlabel` IIUC).

This was the change done by @0-wiz-0,
- https://github.com/rust-lang/libc/pull/4265/changes#diff-6f421831a4bfdab70668b43fde78840b144d19fa215a50f275dceff584e2509fL2852

which got removed in this 
- https://github.com/rust-lang/libc/commit/1816f6061 (because `__getmntinfo13` has been marked as deprecated).

And then when discussing with @0323pin  and @0-wiz-0 we came to the conclusion of sticking to the `__getmntinfo90` call. 

I could raise another one line change PR to restore the old linking to restore functionality to any existing consumers of this API in pkgsrc, until this PR is ready to merge. What do you think?

tgross35:
> which got removed in this
> 
>     * [1816f6061](https://github.com/rust-lang/libc/commit/1816f6061) (because `__getmntinfo13` has been marked as deprecated).

That's my fault, I guess I misunderstood something about the comment at the time. Sorry about that, but I'd be happy to accept a change back `__getmntinfo13` assuming that fixes things.

You can do that in a separate PR if you'd like and keep this one around to change to `__getmntinfo90` once 11.0 is released.

fraggerfox:
> > which got removed in this
> > ```
> > * [1816f6061](https://github.com/rust-lang/libc/commit/1816f6061) (because `__getmntinfo13` has been marked as deprecated).
> > ```
> 
> That's my fault, I guess I misunderstood something about the comment at the time. Sorry about that, but I'd be happy to accept a change back `__getmntinfo13` assuming that fixes things.
> 
> You can do that in a separate PR if you'd like and keep this one around to change to `__getmntinfo90` once 11.0 is released.

Here you go https://github.com/rust-lang/libc/pull/5251

0-wiz-0:
Sorry, I'm late to the party :-)

Paraphrased from a mail I wrote to @fraggerfox:

NetBSD cares about backwards compatibility. When statvfs was introduced, the problem was that we had existing binaries that call `getmntinfo`. So when the struct argument was changed the first time, we wanted to keep them working, so the new function (with the changed struct argument) got a different name, `__getmntinfo13` and there is magic in the NetBSD headers so if you include the header, calls to `getmntinfo` are rewritten to call `__getmntinfo13` instead.

In the meantime, this happened again, so now the function is called `__getmntinfo90` (and the header redirects to that function), with a new `struct statvfs`.

Rust libc doesn't read the headers, but calls the functions directly, so it needs to do the renames itself - -which is why the

> +    #[cfg_attr(target_os = "netbsd", link_name = "__statvfs190")]

lines are needed.

The root cause that started @fraggerfox 's investigation is IMHO the removal of the `__getmntinfo13` line.

When this is fixed, I think it's ok to wait for the NetBSD 11 release, which is also the point at which NetBSD 9 will be officially declared unsupported.

0323pin:
@tgross35

> This looks technically correct at this point, but it sounds like 9.x is still supported so I think we should hold off ...

NetBSD-11 was released 3 days ago and I've just asked internally, 9 will be eol by the end of August.

tgross35:
Thanks for the update! You could send a PR updating https://doc.rust-lang.org/beta/rustc/platform-support/netbsd.html at this point, the current `main` will be stable in October

tgross35:
We've hit the EOL for NetBSD 9, the platform page needs to be updated then we can merge.

@rustbot author

tgross35 (changes requested):
(status update, needs the r-l/r PR)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5243/76857b255738965f40a4c35c5b2f90af6f183c43..9790eabfeb6bc6554d6b58b608c6e33b558397a6)*

fraggerfox:
> We've hit the EOL for NetBSD 9, the platform page needs to be updated then we can merge.
> 
> @rustbot author

I have raised a PR for this
- https://github.com/rust-lang/rust/pull/162268

cc: @0323pin

PR #5245, opened 2026-07-03
nuttx: clean up module

dybucc:
# Description

This PR consists of an overall clean up of the NuttX module. It adds missing types, fixes one type definition, replaces equivalent with mirrored definitions (fixed-width types with "variable" C types,) and provides correct support for some build-time configuration options.

The build-time configuration options correspond with `CONFIG_SMALL_MEMORY` and `CONFIG_FS_LARGEFILE`. The former enables types with pointer bit width (`size_t` and `ssize_t`) to be narrower than required in the target. It is enabled by default in 16-bit machine word targets but can also be enabled in other targets. The latter option enables support for LFS bindings. The default under all targets is to expose 32-bit types and **no** suffixed 64-bit types. It is only if the above option is set that (1) suffixed types are exposed and (2) unsuffixed types are made 64-bits wide. I have decided against exposing the suffixed types because it's redundant.

## Sources

- NuttX source repository.
  - <https://github.com/search?q=repo%3Aapache%2Fnuttx+path%3Aarch%2F**%2F*+%2F%5Cb_ssize_t%5Cb%2F&type=code>
  - <https://github.com/apache/nuttx/blob/master/include/sys/types.h>

## Checklist

### DIFF
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -14,6 +14,7 @@ pub type intptr_t = isize;
 pub type uintptr_t = usize;
 pub type ssize_t = isize;
 
+#[cfg(not(target_os = "nuttx"))]
 pub type pid_t = i32;
 pub type in_addr_t = u32;
 pub type in_port_t = u16;
--- a/src/unix/nuttx/mod.rs
+++ b/src/unix/nuttx/mod.rs
@@ -30,7 +30,14 @@ pub type suseconds_t = i32;
 pub type tcflag_t = u32;
 pub type clockid_t = i32;
 pub type time_t = i64;
-pub type wchar_t = i32;
+
+cfg_if! {
+    if #[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))] {
+        pub type wchar_t = c_int;
+    } else if #[cfg(any(target_arch = "arm", target_arch = "aarch64"))] {
+        pub type wchar_t = c_uint;
+    }
+}
 
 s! {
     pub struct stat {
--- a/src/unix/nuttx/mod.rs
+++ b/src/unix/nuttx/mod.rs
@@ -6,6 +6,7 @@ use crate::{
     DIR,
 };
 
+pub type pid_t = c_int;
 pub type nlink_t = u16;
 pub type ino_t = u16;
 pub type blkcnt_t = u64;
@@ -24,11 +25,11 @@ pub type pthread_rwlockattr_t = i32;
 pub type pthread_t = i32;
 pub type rlim_t = i64;
 pub type sa_family_t = u16;
-pub type socklen_t = u32;
+pub type socklen_t = c_uint;
 pub type speed_t = usize;
 pub type suseconds_t = i32;
 pub type tcflag_t = u32;
-pub type clockid_t = i32;
+pub type clockid_t = c_int;
 pub type time_t = i64;
 
 cfg_if! {


src/unix/nuttx/mod.rs
```diff
@@ -6,6 +6,8 @@ use crate::{
     DIR,
 };
 
+pub type id_t = c_int;
+pub type key_t = u32;
```
tgross35: "nuttx: add missing types": Isn't this signed? https://github.com/apache/nuttx/blob/9e467388de0285c1c160681aeafaa9c6b5723ac0/include/sys/types.h#L170

Make sure your links to source are permalinks (available under the triple dots) so we know what things looked like if they change in the future.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5245/a2051f2b28b5d0bda81eac91cbcd3178d9cfbb3f..4a84e5c40fbff7df15fe9894a8e7d65b7f78540e)*

src/unix/nuttx/mod.rs
```diff
@@ -14,6 +16,8 @@ pub type cc_t = u8;
 pub type clock_t = i64;
 pub type dev_t = i32;
 pub type fsblkcnt_t = u64;
+pub type loff_t = off_t;
+pub type cpu_set_t = u32;
```
tgross35: "nuttx: add missing types": Is this definition used? I'd prefer to omit it because of the `volatile`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5245/a2051f2b28b5d0bda81eac91cbcd3178d9cfbb3f..4a84e5c40fbff7df15fe9894a8e7d65b7f78540e)*

src/unix/nuttx/mod.rs
```diff
@@ -27,9 +31,11 @@ pub type sa_family_t = u16;
 pub type socklen_t = u32;
 pub type speed_t = usize;
 pub type suseconds_t = i32;
+pub type useconds_t = i32;
```
tgross35: Isn't this unsigned? https://github.com/apache/nuttx/blob/9e467388de0285c1c160681aeafaa9c6b5723ac0/include/sys/types.h#L259

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5245/a2051f2b28b5d0bda81eac91cbcd3178d9cfbb3f..4a84e5c40fbff7df15fe9894a8e7d65b7f78540e)*

src/unix/nuttx/mod.rs
```diff
@@ -27,9 +31,11 @@ pub type sa_family_t = u16;
...
 pub type tcflag_t = u32;
 pub type clockid_t = i32;
 pub type time_t = i64;
+pub type timer_t = *mut c_void;
```
tgross35: Eek far pointers in the definition. Is this actually needed?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5245/a2051f2b28b5d0bda81eac91cbcd3178d9cfbb3f..4a84e5c40fbff7df15fe9894a8e7d65b7f78540e)*

src/unix/nuttx/mod.rs
```diff
@@ -9,7 +9,7 @@ use crate::{
 pub type id_t = c_int;
 pub type key_t = u32;
 pub type nlink_t = u16;
-pub type ino_t = u16;
+pub type ino_t = u32;
```
tgross35: "nuttx: fix `ino_t` definition" this happened just last week https://github.com/apache/nuttx/commit/50377d090966ad3e02703fc54f9d88666a967eed, I have no clue what the release/support schedule looks like but we should probably hold off a bit. Or at least hear from the maintainers.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5245/a2051f2b28b5d0bda81eac91cbcd3178d9cfbb3f..4a84e5c40fbff7df15fe9894a8e7d65b7f78540e)*

src/unix/nuttx/mod.rs
```diff
@@ -28,15 +29,15 @@ pub type pthread_rwlockattr_t = i32;
...
 pub type speed_t = usize;
 pub type suseconds_t = i32;
 pub type useconds_t = i32;
 pub type tcflag_t = u32;
-pub type clockid_t = i32;
+pub type clockid_t = c_int;
 pub type time_t = i64;
 pub type timer_t = *mut c_void;
-pub type wchar_t = i32;
+pub type wchar_t = c_int;
```
tgross35: "nuttx: mirror type definitions" Seems fine but where does this one come from?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5245/a2051f2b28b5d0bda81eac91cbcd3178d9cfbb3f..4a84e5c40fbff7df15fe9894a8e7d65b7f78540e)*

src/unix/nuttx/mod.rs
```diff
@@ -11,18 +11,34 @@ pub type id_t = c_int;
...
+        pub type fpos_t = i32;
+    }
+}
+
 pub type loff_t = off_t;
 pub type cpu_set_t = u32;
 pub type locale_t = *mut i8;
 pub type mode_t = u32;
 pub type nfds_t = u32;
-pub type off_t = i64;
```
tgross35, line 20: "nuttx: provide correct lfs support": It looks like we were already supporting `CONFIG_FS_LARGEFILE`  - why the change?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5245/a2051f2b28b5d0bda81eac91cbcd3178d9cfbb3f..4a84e5c40fbff7df15fe9894a8e7d65b7f78540e)*

src/unix/nuttx/mod.rs
```diff
@@ -34,6 +34,32 @@ cfg_if! {
     }
 }
 
+cfg_if! {
+    if #[cfg(nuttx_small_mm)] {
```
tgross35: "nuttx: expose `cfg` for small memory model" similarly, I don't think there is any need to add this - it's extra configuration complexity in `libc` and has no known users. This would probably have to be a separate target in rustc, anyway.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5245/a2051f2b28b5d0bda81eac91cbcd3178d9cfbb3f..4a84e5c40fbff7df15fe9894a8e7d65b7f78540e)*

src/unix/nuttx/mod.rs
dybucc: done. noted.

src/unix/nuttx/mod.rs
dybucc: it seems like it is upstream [^1]. we don't provide any bindings to routines that use it, though.

[^1]: <https://github.com/search?q=repo%3Aapache%2Fnuttx+path%3A**%2F*.h+%2F%5Cbcpu_set_t%5Cb%2F&type=code>

src/unix/nuttx/mod.rs
dybucc: you're right. done.

src/unix/nuttx/mod.rs
dybucc: not in our bindings. it is used upstream [^1].

what is it with far pointers? is it inconvenient to test against them because of the segment:offset divide, or is there something else?

[^1]: <https://github.com/search?q=repo%3Aapache%2Fnuttx+path%3A**%2F*.h+%2F%5Cbtimer_t%5Cb%2F&type=code>

src/unix/nuttx/mod.rs
dybucc: pre-defined compiler macros. it's defined as \_\_wchar_type\_\_ [^1], which at least in clang expands to a c int under riscv and to a c unsigned int under arm (both 32- and 64-bit) [^2]. supported rust targets are under either one of these isas.

just made it conditionally compiled when i realized the compiler definitions were different under these two architectures.

[^1]: <https://github.com/search?q=repo%3Aapache%2Fnuttx+%2F%5Cb_wchar_t%5Cb%2F&type=code>
[^2]: <https://c.godbolt.org/z/czMY79vsh>

src/unix/nuttx/mod.rs
dybucc: we were not correctly supporting that configuration. without it enabled, all targets get 32-bit file offset types. with it enabled, they all get 64-bit file offset types.

it would likely come off as a surprise if somebody didn't have CONFIG_FS_LARGEFILE enabled but found that our types are 64-bits either way.

src/unix/nuttx/mod.rs
dybucc: removed.

dybucc:
@no1wudi, you think we can already have an ino_t definition that matches upstream? there was a recent change there and this pr includes it, but maybe that's a bit too early to make changes.

@rustbot ready

src/unix/nuttx/mod.rs
dybucc: just pinged the maintainers.

src/unix/nuttx/mod.rs
tgross35: In general, avoid adding things without a specific usecase. If somebody needs it and adds it that's fine, but if not then it's kind of opting us into maintenance effort for something that might not even be used.

This also isn't a correct definition anyway, it's `volatile`.

src/unix/nuttx/mod.rs
tgross35: > what is it with far pointers? is it inconvenient to test against them because of the segment:offset divide, or is there something else?

Rust just doesn't support them at all because LLVM support is still in the works, as far as I know.

It looks like this is only relevant on AVR https://github.com/apache/nuttx/blob/cce73da215469a7709cd5ffec5d0ad8d8c98309e/include/nuttx/compiler.h#L471-L545 (though I didn't scan end-to-end) so the definition is technically ok, though same as https://github.com/rust-lang/libc/pull/5245#discussion_r3539200611 we should avoid adding API with no consumer.

src/unix/nuttx/mod.rs
tgross35: I mean we were _only_ supporting `CONFIG_FS_LARGEFILE`, right? That's completely fine, every platform has dozens of ways you could configure it that make something break for Rust.

src/unix/nuttx/mod.rs
```diff
@@ -6,31 +6,61 @@ use crate::{
...
 pub type cc_t = u8;
 pub type clock_t = i64;
 pub type dev_t = i32;
-pub type fsblkcnt_t = u64;
+
+// LFS64 types with "64"-suffixed names are only exposed when the unsuffixed
+// types are themselves 64-bits wide. We choose to expose ony the latter.
+cfg_if! {
+    if #[cfg(gnu_file_offset_bits64)] {
+        pub type fsblkcnt_t = u64;
```
no1wudi: There are too many configuration options inside Nuttx; it would be better to only support specific projects. For example, FS_LARGEFS must be enabled; otherwise, compilation becomes overly difficult.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5245/67c9a34fda91bc26a50cd907b1e8a65ee32bfb65..6b62a277315e5ef149c6165c56b206e64d99dd0f)*

src/unix/nuttx/mod.rs
no1wudi: Agreed. For now, it's best to add things on an as‑needed basis, since the definitions on the NuttX side may change (e.g., ino_t). Don't introduce unused definitions too early.

no1wudi (changes requested):

src/unix/nuttx/mod.rs
no1wudi: > I mean we were _only_ supporting `CONFIG_FS_LARGEFILE`, right? That's completely fine, every platform has dozens of ways you could configure it that make something break for Rust.

Yes, it is unlikely to support arbitrary configurations of the NuttX ABI on the Rust side, so we must restrict the configuration options that affect fundamental data type definitions and only support the most commonly used combinations.

src/unix/nuttx/mod.rs
dybucc: done. noted.

src/unix/nuttx/mod.rs
dybucc: done.

src/unix/nuttx/mod.rs
dybucc: done. thanks for getting back.

src/unix/mod.rs
```diff
@@ -14,6 +14,7 @@ pub type intptr_t = isize;
 pub type uintptr_t = usize;
 pub type ssize_t = isize;
 
+#[cfg(not(target_os = "nuttx"))]
```
no1wudi, line 17: Revert? https://elixir.bootlin.com/linux/v7.1.2/source/include/uapi/asm-generic/posix_types.h#L28

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5245/5848d319e53719d637781eae72de169d8f7867a0..0e9c84b43d96debb81f536e370651faf2f699939)*

dybucc: not really, the latest commit replicates the nuttx definition [^1] more closely. if anything, the link you provide matches the current definition (a c int and not necessarily a fixed-width integer.)

[^1]: https://github.com/apache/nuttx/blob/master/include/sys/types.h#L157

tgross35 (approved):
The current form LGTM but I'll wait for maintainer ack

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5245/5848d319e53719d637781eae72de169d8f7867a0..0e9c84b43d96debb81f536e370651faf2f699939)*

no1wudi (approved):

no1wudi:
> The current form LGTM but I'll wait for maintainer ack
> 

Thanks, LGTM now

tgross35 (approved):
Thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5245/70c67fe2ec83eda97646b49cab2bcbd1603a65ea..74a5acf9e205f2f20170f864ca969fde33201111)*

PR #5257, opened 2026-07-07
add macros for generating `Default` impls

telcharr:
# Description

Adds `impl_default!`, `s_with_default!`, and `s_no_extra_traits_with_default!` as a start for rust-lang/libc#4975. Unions get no implicit `Default` so a union field would need to supply one via `#[custom_default(unsafe { mem::zeroed::<U>() })]`. As discussed, `custom_default` needs to be the first attribute on a field since the macro matches it literally. 

# Checklist

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -250,6 +250,194 @@ macro_rules! s_no_extra_traits {
     );
 }
 
+/// Like [`s`], but also generates a `Default` impl for every struct in the block.
+macro_rules! s_with_default {
+    ($(
+        $(#[$attr:meta])*
+        $pub:vis $t:ident $i:ident { $($field:tt)* }
+    )*) => ($(
+        s_with_default!(it: $(#[$attr])* $pub $t $i { $($field)* });
+    )*);
+
+    (it: $(#[$attr:meta])* $pub:vis union $i:ident { $($field:tt)* }) => (
+        compile_error!(
+            "unions cannot derive extra traits, use s_no_extra_traits_with_default instead"
+        );
+    );
+
+    (it: $(#[$attr:meta])* $pub:vis struct $i:ident { $($field:tt)* }) => (
+        struct_with_default! {
+            attrs: {
+                #[repr(C)]
+                #[::core::prelude::v1::derive(
+                    ::core::clone::Clone,
+                    ::core::marker::Copy,
+                    ::core::fmt::Debug,
+                )]
+                #[cfg_attr(
+                    feature = "extra_traits",
+                    ::core::prelude::v1::derive(PartialEq, Eq, Hash)
+                )]
+                #[allow(deprecated)]
+            }
+            $(#[$attr])* $pub struct $i { $($field)* }
+        }
+    );
+}
+
+/// Like [`s_no_extra_traits`], but also generates a `Default` impl for every struct in the block.
+///
+/// Unions are emitted just like `s_no_extra_traits!` does, with no `Default`. A struct field of
+/// union type supplies its own default via `#[custom_default(...)]`.
+macro_rules! s_no_extra_traits_with_default {
+    ($(
+        $(#[$attr:meta])*
+        $pub:vis $t:ident $i:ident { $($field:tt)* }
+    )*) => ($(
+        s_no_extra_traits_with_default!(it: $(#[$attr])* $pub $t $i { $($field)* });
+    )*);
+
+    (it: $(#[$attr:meta])* $pub:vis union $i:ident { $($field:tt)* }) => (
+        #[repr(C)]
+        #[::core::prelude::v1::derive(
+            ::core::clone::Clone,
+            ::core::marker::Copy,
+        )]
+        $(#[$attr])*
+        $pub union $i { $($field)* }
+
+        impl ::core::fmt::Debug for $i {
+            fn fmt(&self, f: &mut ::core::fmt::Formatter<'_>) -> ::core::fmt::Result {
+                f.debug_struct(::core::stringify!($i)).finish_non_exhaustive()
+            }
+        }
+    );
+
+    (it: $(#[$attr:meta])* $pub:vis struct $i:ident { $($field:tt)* }) => (
+        struct_with_default! {
+            attrs: {
+                #[repr(C)]
+                #[::core::prelude::v1::derive(
+                    ::core::clone::Clone,
+                    ::core::marker::Copy,
+                    ::core::fmt::Debug,
+                )]
+            }
+            $(#[$attr])* $pub struct $i { $($field)* }
+        }
+    );
+}
+
+/// Emit a struct with the given derive attributes plus a generated `Default` impl.
+///
+/// Fields default to `Default::default()`. A field whose default can't be derived must carry
+/// `#[custom_default(EXPR)]` as its *first* attribute, and `EXPR` is used instead.
+///
+/// This works by scanning each field for `#[custom_default]` attributes. If one exists, the
+/// attribute's contents are added to `processed_field_defaults` and will be used in the expansion
+/// for `Default`. If it does not exist, `Default::default()` is used instead. In either case, the
+/// field is added to `processed_fields` with `#[custom_default]` stripped if necessary, and
+/// `struct_with_default` is invoked again with the remaining fields.
+macro_rules! struct_with_default {
+    // entry; `attrs` is the attribute block the caller wants on the struct (repr, derives, etc.),
+    // which is merged with the struct's own attributes.
+    (
+        attrs: { $($attrs:tt)* }
+        $(#[$attr:meta])*
+        $vis:vis struct $name:ident { $($body:tt)* }
+    ) => {
+        struct_with_default! {
+            @struct
+            attrs: { $($attrs)* $(#[$attr])* }
+            vis: { $vis }
+            name: { $name }
+            processed_fields: { }
+            processed_field_defaults: { }
+            remaining_fields: { $($body)* }
+        }
+    };
+
+    // field led by #[custom_default(...)]
+    (
+        @struct
+        attrs: { $($attrs:tt)* }
+        vis: { $vis:vis }
+        name: { $name:ident }
+        processed_fields: { $($processed_fields:tt)* }
+        processed_field_defaults: { $($processed_field_defaults:tt)* }
+        remaining_fields: {
+            #[custom_default($default:expr)]
+            $(#[$fattr:meta])*
+            $fvis:vis $fname:ident: $fty:ty,
+            $($tail:tt)*
+        }
+    ) => {
+        struct_with_default! {
+            @struct
+            attrs: { $($attrs)* }
+            vis: { $vis }
+            name: { $name }
+            processed_fields: { $($processed_fields)* $(#[$fattr])* $fvis $fname: $fty, }
+            processed_field_defaults: {
+                $($processed_field_defaults)*
+                $(#[$fattr])* $fname: $default,
+            }
+            remaining_fields: { $($tail)* }
+        }
+    };
+
+    // plain field
+    (
+        @struct
+        attrs: { $($attrs:tt)* }
+        vis: { $vis:vis }
+        name: { $name:ident }
+        processed_fields: { $($processed_fields:tt)* }
+        processed_field_defaults: { $($processed_field_defaults:tt)* }
+        remaining_fields: {
+            $(#[$fattr:meta])*
+            $fvis:vis $fname:ident: $fty:ty,
+            $($tail:tt)*
+        }
+    ) => {
+        struct_with_default! {
+            @struct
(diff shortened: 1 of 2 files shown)


src/macros.rs
```diff
@@ -428,6 +428,176 @@ macro_rules! offset_of {
...
+        impl ::core::default::Default for $name {
+            fn default() -> Self {
+                Self { $($inits)* }
+            }
+        }
+    };
+}
+
+/// Like [`s`], but also generates a `Default` impl for every struct in the block.
+macro_rules! s_with_default {
```
tgross35: Nit: move this and `s_no_extra_traits_with_default` to directly below `s_no_extra_traits!` to keep them together (this file kind of flows from most important to less important / helpers)

src/macros.rs
```diff
@@ -428,6 +428,176 @@ macro_rules! offset_of {
...
+                #[cfg_attr(
+                    feature = "extra_traits",
+                    ::core::prelude::v1::derive(PartialEq, Eq, Hash)
+                )]
+                #[allow(deprecated)]
+            }
+            $(#[$attr])* $vis struct $name { $($body)* }
+        }
+        s_with_default! { $($rest)* }
+    };
```
tgross35: Is it possible to keep the same pattern as https://github.com/telcharr/libc/blob/06e7c117b1b867e975c3184dd0daf172624132a6/src/macros.rs#L153-L158? I.e. use repetition rather than the tt muncher - I believe that tends to be more efficient for large macro expansions. (Can't avoid it for the inner expansion, of course)

Ditto for `s_no_extra_traits_with_default`

src/macros.rs
```diff
@@ -428,6 +428,176 @@ macro_rules! offset_of {
...
+}
+
+/// Like [`s_no_extra_traits`], but also generates a `Default` impl for every struct in the block.
+macro_rules! s_no_extra_traits_with_default {
```
tgross35: Add to the docs that this does nothing for unions

src/macros.rs
```diff
@@ -428,6 +428,176 @@ macro_rules! offset_of {
...
+        attrs: { $($attr:tt)* }
+        vis: { $vis:vis }
+        name: { $name:ident }
+        stripped: { $($stripped:tt)* }
+        inits: { $($inits:tt)* }
+        rest: { }
+    ) => {
+        $($derives)*
+        $($attr)*
+        $vis struct $name { $($stripped)* }
```
tgross35: Do `derives` and `attrs` need to be split? It looks like they're always handled together.

If they do need to be split then maybe there are more accurate names, considering `#[repr(C)]` is currently included in `derives`.

src/macros.rs
```diff
@@ -626,4 +796,32 @@ mod macro_checks {
         type Foo;
         pub type Bar;
     }
+
+    s_with_default! {
+        pub struct S3 {
+            pub a: u32,
+            #[custom_default([0; 64])]
```
tgross35: Could you make this something other than the default like `[1; 64]`? To make sure the field attribute is working rather than the `derive`.

src/macros.rs
```diff
@@ -428,6 +428,176 @@ macro_rules! offset_of {
...
+                f.debug_struct(::core::stringify!($name)).finish_non_exhaustive()
+            }
+        }
+
+        s_no_extra_traits_with_default! { $($rest)* }
+    };
+}
+
 #[cfg(test)]
 mod tests {
```
tgross35, line 647: It would be good to add a test that field attributes aren't getting stripped, since that can be easy to miss. Since we test on a variety of arches, maybe something like:

```rust
s_with_default! {
    struct S4 {
        #[cfg(target_arch = "x86_64")]
        a: u8,
        #[cfg(not(target_arch = "x86_64"))]
        a: u64,
    }
}

let s = S4::default();

cfg_if! {
    if #[cfg(target_arch = "x86_64")] {
        assert_eq!(s.a, 0u8);
    } else {
        assert_eq!(s.a, 0u64);
    }
}
```

src/macros.rs
```diff
@@ -428,6 +428,176 @@ macro_rules! offset_of {
...
+        vis: { $vis:vis }
+        name: { $name:ident }
+        stripped: { $($stripped:tt)* }
+        inits: { $($inits:tt)* }
+        rest: {
+            #[custom_default($default:expr)]
+            $(#[$fattr:meta])*
+            $fvis:vis $fname:ident: $fty:ty,
+            $($tail:tt)*
+        }
```
tgross35: Suggested names that I think would make this a bit easier to follow:

- stripped -> processed_fields
- inits -> processed_field_defaults
- rest -> remaining_fields

Mostly to make it clear that these are working with fields and that `stripped` kind of goes with `inits`.

src/macros.rs
```diff
@@ -428,6 +428,176 @@ macro_rules! offset_of {
     }};
 }
 
+/// Emit a struct with the given derive attributes plus a generated `Default` impl.
+///
+/// Fields default to `Default::default()`. A field whose default can't be derived must carry
+/// `#[custom_default(EXPR)]` as its *first* attribute, and `EXPR` is used instead.
+macro_rules! impl_default {
```
tgross35: Small theory of operation (using names from https://github.com/rust-lang/libc/pull/5257/changes#r3548882196) to add, since these things unfortunately get tough to remember a few years down the line:

> This works by scanning each field for `#[custom_default]` attributes. If one exists, the attribute's contents are added to `processed_field_defaults` and will be used in the expansion for `Default`. If it does not exist, `Default::default()` is used instead. In either case, the field is added to `processed_fields` with `#[custom_default]` stripped if necessary, and `impl_default` is invoked again with the remaining fields.

tgross35 (changes requested):
Thank you for putting this together! It looks very workable. I have a handful of requests but all style/maintainability.

Would you be able to add a commit that makes use of this (replacing `s!`) in a few small places so we make sure it works in situ? Fine to do that in a separate PR, if you prefer, as long as it's coming soon.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5257/67c9a34fda91bc26a50cd907b1e8a65ee32bfb65..06e7c117b1b867e975c3184dd0daf172624132a6)*

src/macros.rs
telcharr: Moved.

src/macros.rs
telcharr: Done, both use repetition now. Kept the muncher for inner field expansion. `s_with_default!` also picks up the `compile_error!` for unions that `s!` has which was missing before.

src/macros.rs
telcharr: Added.

src/macros.rs
telcharr: Good catch, no they don't. This was a sloppy leftover of a design I tried initially. They've been merged into a single `attrs` list.

src/macros.rs
telcharr: Added and fixed a bug it caught. I was forwarding field attributes to the struct definition but not to the initializer in the `Default` impl. Field attrs now go to both. Thanks for calling this one out!

Doc comments get forwarded too as a side effect, which seems harmless but trips `unused_doc_comments`, so I added `#[allow(unused_doc_comments)]` on the generated `default()`.

src/macros.rs
telcharr: Changed.

src/macros.rs
telcharr: Renamed.

src/macros.rs
telcharr: Added, thank you for writing this!

telcharr:
Added a commit converting `new/linux_uapi/linux/can.rs` since it hits most cases. The only things it doesn't cover is `#[cfg]` and doc comments. If you'd like me to add a new commit implementing the macros that covers those cases let me know.

@rustbot ready

src/macros.rs
tgross35: I meant to order the macros:

* `s`
* `s_paren`
* `s_no_extra_traits`
* `s_with_default`
* `s_no_extra_traits_with_default`
* `impl_default`

So all the `s`s are together - currently `impl_default` is in the middle

src/macros.rs
tgross35: I missed that there isn't currently a test for this, the change is kind of pointless without it 🙂 could you assert the field values in `S3` and the other structs/unions in that area?

src/macros.rs
```diff
@@ -244,6 +244,194 @@ macro_rules! s_no_extra_traits {
...
+///
+/// Fields default to `Default::default()`. A field whose default can't be derived must carry
+/// `#[custom_default(EXPR)]` as its *first* attribute, and `EXPR` is used instead.
+///
+/// This works by scanning each field for `#[custom_default]` attributes. If one exists, the
+/// attribute's contents are added to `processed_field_defaults` and will be used in the expansion
+/// for `Default`. If it does not exist, `Default::default()` is used instead. In either case, the
+/// field is added to `processed_fields` with `#[custom_default]` stripped if necessary, and
+/// `impl_default` is invoked again with the remaining fields.
+macro_rules! impl_default {
```
tgross35: Maybe rename this to `struct_with_default` or similar since it also emits the struct itself

src/macros.rs
tgross35: Actually I guess `S3`'s pattern pattern is already checked with `CustomDefault`

src/macros.rs
```diff
@@ -244,6 +244,194 @@ macro_rules! s_no_extra_traits {
...
+        $vis struct $name { $($processed_fields)* }
+
+        impl ::core::default::Default for $name {
+            // Field doc comments get forwarded to the initializer alongside `#[cfg]`
+            // they're harmless there but the lint fires, so silence it.
+            #[allow(unused_doc_comments)]
+            fn default() -> Self {
+                Self { $($processed_field_defaults)* }
+            }
+        }
```
tgross35, line 437: I'm realizing that something like `#[cfg(target_arch = "x86_64")]` on the whole struct may not work correctly since `impl Default` won't get the same config. I think we might need to filter the struct attributes and pass any `cfg(...)` to `Default`.

That can be done in a followup though, this gets us most of the way.

(`Debug` for unions has the same issue, guess we just haven't hit it.)

src/macros.rs
```diff
@@ -583,6 +771,42 @@ mod tests {
...
+        // If `custom_default` stripping ate other field attributes, the two `a` fields would
+        // collide. Checked across arches since CI runs many.
+        s_with_default! {
+            struct FieldAttrs {
+                #[cfg(target_arch = "x86_64")]
+                a: u8,
+                #[cfg(not(target_arch = "x86_64"))]
+                a: u64,
+            }
+        }
```
tgross35, line 864: Could you add a `#[cfg(target_arch = "x86_64")]` field that doesn't have a `cfg(not(...))` to go with it? Just thinking about how `a: Default::default()` technically works on all platforms here, but it might not always.

tgross35 (changes requested):
A few small requests here and in the comments above, then I think this should be good to go. Could you squash please? (The change updating `can.rs` can stay a separate commit)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5257/67c9a34fda91bc26a50cd907b1e8a65ee32bfb65..3eab27d6490a133c96c6dc204c60dc08f6ee6539)*

tgross35:
@telcharr gentle nudge, this is pretty close to over the line and would be great to have!

src/macros.rs
telcharr: Reordered, my bad on that one.

src/macros.rs
telcharr: Related catch, I added a value test for the union case that asserts it zeroes through `custom_default`

src/macros.rs
telcharr: Renamed.

src/macros.rs
telcharr: Happy to pick this one up once this lands, if you want it.

src/macros.rs
telcharr: Added a test with a single `#[cfg(target_arch = "x86_64")]` field and no complementary arm.

telcharr:
Sorry this took a bit, it's been a crazy past couple of weeks. Should have everything addressed now, but let me know if I've missed anything.

@rustbot ready

tgross35 (approved):
Looks great, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5257/471be3d9247ddfa17e5c2139397db10b181f73a7..5626dac90a31692c3aedb2f3290a3b4fd10997d0)*

tgross35:
And yes, a follow up with https://github.com/rust-lang/libc/pull/5257#discussion_r3708605415 would be welcome

PR #5259, opened 2026-07-07
aix: clean up module

dybucc:
## Description

This PR is concerned with a general clean up of the AIX module. See the patch
messages for details.

## Checklist

  doc comment

### DIFF
--- a/src/unix/aix/mod.rs
+++ b/src/unix/aix/mod.rs
@@ -3344,5 +3344,7 @@ cfg_if! {
     if #[cfg(target_arch = "powerpc64")] {
         mod powerpc64;
         pub use self::powerpc64::*;
+    } else {
+        core::compile_error!("unsupported target");
     }
 }
--- a/src/unix/aix/mod.rs
+++ b/src/unix/aix/mod.rs
@@ -1,8 +1,4 @@
 use crate::prelude::*;
-use crate::{
-    in_addr_t,
-    in_port_t,
-};
 
 pub type caddr_t = *mut c_char;
 pub type clockid_t = c_longlong;
@@ -85,7 +81,7 @@ s! {
     }
 
     pub struct fsid64_t {
-        pub val: [crate::uint64_t; 2],
+        pub val: [u64; 2],
     }
 
     pub struct timezone {
@@ -100,18 +96,18 @@ s! {
 
     pub struct dirent {
         pub d_offset: c_ulong,
-        pub d_ino: crate::ino_t,
+        pub d_ino: ino_t,
         pub d_reclen: c_ushort,
         pub d_namlen: c_ushort,
         pub d_name: [c_char; 256],
     }
 
     pub struct termios {
-        pub c_iflag: crate::tcflag_t,
-        pub c_oflag: crate::tcflag_t,
-        pub c_cflag: crate::tcflag_t,
-        pub c_lflag: crate::tcflag_t,
-        pub c_cc: [crate::cc_t; crate::NCCS],
+        pub c_iflag: tcflag_t,
+        pub c_oflag: tcflag_t,
+        pub c_cflag: tcflag_t,
+        pub c_lflag: tcflag_t,
+        pub c_cc: [crate::cc_t; NCCS],
     }
 
     pub struct flock64 {
@@ -126,7 +122,7 @@ s! {
 
     pub struct msghdr {
         pub msg_name: *mut c_void,
-        pub msg_namelen: crate::socklen_t,
+        pub msg_namelen: socklen_t,
         pub msg_iov: *mut crate::iovec,
         pub msg_iovlen: c_int,
         pub msg_control: *mut c_void,
@@ -135,14 +131,14 @@ s! {
     }
 
     pub struct statvfs64 {
-        pub f_bsize: crate::blksize64_t,
-        pub f_frsize: crate::blksize64_t,
-        pub f_blocks: crate::blkcnt64_t,
-        pub f_bfree: crate::blkcnt64_t,
-        pub f_bavail: crate::blkcnt64_t,
-        pub f_files: crate::blkcnt64_t,
-        pub f_ffree: crate::blkcnt64_t,
-        pub f_favail: crate::blkcnt64_t,
+        pub f_bsize: blksize64_t,
+        pub f_frsize: blksize64_t,
+        pub f_blocks: blkcnt64_t,
+        pub f_bfree: blkcnt64_t,
+        pub f_bavail: blkcnt64_t,
+        pub f_files: blkcnt64_t,
+        pub f_ffree: blkcnt64_t,
+        pub f_favail: blkcnt64_t,
         pub f_fsid: fsid64_t,
         pub f_basetype: [c_char; 16],
         pub f_flag: c_ulong,
@@ -199,13 +195,13 @@ s! {
         pub ai_protocol: c_int,
         pub ai_addrlen: c_ulong,
         pub ai_canonname: *mut c_char,
-        pub ai_addr: *mut crate::sockaddr,
+        pub ai_addr: *mut sockaddr,
         pub ai_next: *mut addrinfo,
         pub ai_eflags: c_int,
     }
 
     pub struct in_addr {
-        pub s_addr: in_addr_t,
+        pub s_addr: crate::in_addr_t,
     }
 
     pub struct ip_mreq_source {
@@ -234,7 +230,7 @@ s! {
     pub struct sockaddr_in {
         pub sin_len: c_uchar,
         pub sin_family: sa_family_t,
-        pub sin_port: in_port_t,
+        pub sin_port: crate::in_port_t,
         pub sin_addr: in_addr,
         pub sin_zero: [c_uchar; 8],
     }
@@ -242,17 +238,17 @@ s! {
     pub struct sockaddr_in6 {
         pub sin6_len: c_uchar,
         pub sin6_family: c_uchar,
-        pub sin6_port: crate::uint16_t,
-        pub sin6_flowinfo: crate::uint32_t,
+        pub sin6_port: u16,
+        pub sin6_flowinfo: u32,
         pub sin6_addr: crate::in6_addr,
-        pub sin6_scope_id: crate::uint32_t,
+        pub sin6_scope_id: u32,
     }
 
     pub struct sockaddr_storage {
         pub __ss_len: c_uchar,
         pub ss_family: sa_family_t,
         __ss_pad1: Padding<[c_char; 6]>,
-        __ss_align: crate::int64_t,
+        __ss_align: i64,
         __ss_pad2: Padding<[c_char; 1265]>,
     }
 
@@ -263,7 +259,7 @@ s! {
     }
 
     pub struct st_timespec {
-        pub tv_sec: crate::time_t,
+        pub tv_sec: time_t,
         pub tv_nsec: c_int,
     }
 
@@ -274,8 +270,8 @@ s! {
         pub f_blocks: blkcnt64_t,
         pub f_bfree: blkcnt64_t,
         pub f_bavail: blkcnt64_t,
-        pub f_files: crate::uint64_t,
-        pub f_ffree: crate::uint64_t,
+        pub f_files: u64,
+        pub f_ffree: u64,
         pub f_fsid: fsid64_t,
         pub f_vfstype: c_int,
         pub f_fsize: blksize64_t,
@@ -313,7 +309,7 @@ s! {
     }
 
     pub struct cmsghdr {
-        pub cmsg_len: crate::socklen_t,
+        pub cmsg_len: socklen_t,
         pub cmsg_level: c_int,
         pub cmsg_type: c_int,
     }
@@ -343,16 +339,16 @@ s! {
         pub header_data: *mut c_void,
         pub header_length: c_uint,
         pub file_descriptor: c_int,
-        pub file_size: crate::uint64_t,
-        pub file_offset: crate::uint64_t,
-        pub file_bytes: crate::int64_t,
+        pub file_size: u64,
+        pub file_offset: u64,
+        pub file_bytes: i64,
         pub trailer_data: *mut c_void,
         pub trailer_length: c_uint,
-        pub bytes_sent: crate::uint64_t,
+        pub bytes_sent: u64,
     }
 
     pub struct mmsghdr {
-        pub msg_hdr: crate::msghdr,
+        pub msg_hdr: msghdr,
         pub msg_len: c_uint,
     }
 
@@ -366,14 +362,14 @@ s! {
         pub ss_sp: *mut c_void,
         pub ss_size: size_t,
         pub ss_flags: c_int,
-        pub __pad: [c_int; 4],
+        __pad: Padding<[c_int; 4]>,
     }
 
     pub struct posix_spawnattr_t {
         pub posix_attr_flags: c_short,
         pub posix_attr_pgroup: crate::pid_t,
-        pub posix_attr_sigmask: crate::sigset_t,
-        pub posix_attr_sigdefault: crate::sigset_t,
+        pub posix_attr_sigmask: sigset_t,
+        pub posix_attr_sigdefault: sigset_t,
         pub posix_attr_schedpolicy: c_int,
         pub posix_attr_schedparam: sched_param,
     }
@@ -413,9 +409,9 @@ s! {
         pub ut_time: time64_t,
         pub ut_exit: exit_status,
         pub ut_host: [c_char; 256],
-        pub __dbl_word_pad: c_int,
-        pub __reservedA: [c_int; 2],
-        pub __reservedV: [c_int; 6],
+        __dbl_word_pad: Padding<c_int>,
+        __reservedA: Padding<[c_int; 2]>,
+        __reservedV: Padding<[c_int; 6]>,
     }
 
     pub struct regmatch_t {
@@ -429,7 +425,7 @@ s! {
         pub re_cflags: c_int,
         pub re_erroff: size_t,
         pub re_len: size_t,
-        pub re_ucoll: [crate::wchar_t; 2],
+        pub re_ucoll: [wchar_t; 2],
         pub re_lsub: [*mut c_void; 24],
         pub re_esub: [*mut c_void; 24],
         pub re_map: *mut c_uchar,
@@ -452,12 +448,12 @@ s! {
         pub shm_atime: time_t,
         pub shm_dtime: time_t,
         pub shm_ctime: time_t,
-        pub shm_handle: crate::uint32_t,
+        pub shm_handle: u32,
         pub shm_extshm: c_int,
-        pub shm_pagesize: crate::int64_t,
-        pub shm_lba: crate::uint64_t,
-        shm_reserved0: Padding<crate::int64_t>,
-        shm_reserved1: Padding<crate::int64_t>,
+        pub shm_pagesize: i64,
+        pub shm_lba: u64,
+        shm_reserved0: Padding<i64>,
+        shm_reserved1: Padding<i64>,
     }
 
     pub struct stat64 {
@@ -657,63 +653,63 @@ pub const GLOB_NOMATCH: c_int = 0x4000;
 pub const GLOB_NOSYS: c_int = 0x8000;
 
 // langinfo.h
-pub const DAY_1: crate::nl_item = 13;
-pub const DAY_2: crate::nl_item = 14;
-pub const DAY_3: crate::nl_item = 15;
-pub const DAY_4: crate::nl_item = 16;
-pub const DAY_5: crate::nl_item = 17;
-pub const DAY_6: crate::nl_item = 18;
-pub const DAY_7: crate::nl_item = 19;
-pub const ABDAY_1: crate::nl_item = 6;
-pub const ABDAY_2: crate::nl_item = 7;
-pub const ABDAY_3: crate::nl_item = 8;
-pub const ABDAY_4: crate::nl_item = 9;
-pub const ABDAY_5: crate::nl_item = 10;
-pub const ABDAY_6: crate::nl_item = 11;
-pub const ABDAY_7: crate::nl_item = 12;
-pub const MON_1: crate::nl_item = 32;
-pub const MON_2: crate::nl_item = 33;
-pub const MON_3: crate::nl_item = 34;
-pub const MON_4: crate::nl_item = 35;
-pub const MON_5: crate::nl_item = 36;
-pub const MON_6: crate::nl_item = 37;
-pub const MON_7: crate::nl_item = 38;
-pub const MON_8: crate::nl_item = 39;
-pub const MON_9: crate::nl_item = 40;
-pub const MON_10: crate::nl_item = 41;
-pub const MON_11: crate::nl_item = 42;
-pub const MON_12: crate::nl_item = 43;
-pub const ABMON_1: crate::nl_item = 20;
-pub const ABMON_2: crate::nl_item = 21;
-pub const ABMON_3: crate::nl_item = 22;
-pub const ABMON_4: crate::nl_item = 23;
-pub const ABMON_5: crate::nl_item = 24;
-pub const ABMON_6: crate::nl_item = 25;
-pub const ABMON_7: crate::nl_item = 26;
-pub const ABMON_8: crate::nl_item = 27;
-pub const ABMON_9: crate::nl_item = 28;
-pub const ABMON_10: crate::nl_item = 29;
-pub const ABMON_11: crate::nl_item = 30;
-pub const ABMON_12: crate::nl_item = 31;
-pub const RADIXCHAR: crate::nl_item = 44;
-pub const THOUSEP: crate::nl_item = 45;
-pub const YESSTR: crate::nl_item = 46;
-pub const NOSTR: crate::nl_item = 47;
-pub const CRNCYSTR: crate::nl_item = 48;
-pub const D_T_FMT: crate::nl_item = 1;
-pub const D_FMT: crate::nl_item = 2;
-pub const T_FMT: crate::nl_item = 3;
-pub const AM_STR: crate::nl_item = 4;
-pub const PM_STR: crate::nl_item = 5;
-pub const CODESET: crate::nl_item = 49;
-pub const T_FMT_AMPM: crate::nl_item = 55;
-pub const ERA: crate::nl_item = 56;
-pub const ERA_D_FMT: crate::nl_item = 57;
-pub const ERA_D_T_FMT: crate::nl_item = 58;
-pub const ERA_T_FMT: crate::nl_item = 59;
-pub const ALT_DIGITS: crate::nl_item = 60;
-pub const YESEXPR: crate::nl_item = 61;
-pub const NOEXPR: crate::nl_item = 62;
+pub const DAY_1: nl_item = 13;
+pub const DAY_2: nl_item = 14;
+pub const DAY_3: nl_item = 15;
+pub const DAY_4: nl_item = 16;
+pub const DAY_5: nl_item = 17;
+pub const DAY_6: nl_item = 18;
+pub const DAY_7: nl_item = 19;
+pub const ABDAY_1: nl_item = 6;
+pub const ABDAY_2: nl_item = 7;
+pub const ABDAY_3: nl_item = 8;
+pub const ABDAY_4: nl_item = 9;
+pub const ABDAY_5: nl_item = 10;
+pub const ABDAY_6: nl_item = 11;
+pub const ABDAY_7: nl_item = 12;
+pub const MON_1: nl_item = 32;
+pub const MON_2: nl_item = 33;
+pub const MON_3: nl_item = 34;
+pub const MON_4: nl_item = 35;
+pub const MON_5: nl_item = 36;
+pub const MON_6: nl_item = 37;
+pub const MON_7: nl_item = 38;
+pub const MON_8: nl_item = 39;
+pub const MON_9: nl_item = 40;
+pub const MON_10: nl_item = 41;
+pub const MON_11: nl_item = 42;
+pub const MON_12: nl_item = 43;
+pub const ABMON_1: nl_item = 20;
+pub const ABMON_2: nl_item = 21;
+pub const ABMON_3: nl_item = 22;
+pub const ABMON_4: nl_item = 23;
+pub const ABMON_5: nl_item = 24;
+pub const ABMON_6: nl_item = 25;
+pub const ABMON_7: nl_item = 26;
+pub const ABMON_8: nl_item = 27;
+pub const ABMON_9: nl_item = 28;
+pub const ABMON_10: nl_item = 29;
+pub const ABMON_11: nl_item = 30;
+pub const ABMON_12: nl_item = 31;
+pub const RADIXCHAR: nl_item = 44;
+pub const THOUSEP: nl_item = 45;
+pub const YESSTR: nl_item = 46;
+pub const NOSTR: nl_item = 47;
+pub const CRNCYSTR: nl_item = 48;
+pub const D_T_FMT: nl_item = 1;
+pub const D_FMT: nl_item = 2;
+pub const T_FMT: nl_item = 3;
+pub const AM_STR: nl_item = 4;
+pub const PM_STR: nl_item = 5;
+pub const CODESET: nl_item = 49;
+pub const T_FMT_AMPM: nl_item = 55;
+pub const ERA: nl_item = 56;
+pub const ERA_D_FMT: nl_item = 57;
+pub const ERA_D_T_FMT: nl_item = 58;
+pub const ERA_T_FMT: nl_item = 59;
+pub const ALT_DIGITS: nl_item = 60;
+pub const YESEXPR: nl_item = 61;
+pub const NOEXPR: nl_item = 62;
 
 // locale.h
 pub const LC_GLOBAL_LOCALE: crate::locale_t = -1isize as crate::locale_t;
@@ -738,14 +734,14 @@ pub const LC_ALL_MASK: c_int = LC_COLLATE_MASK
     | LC_TIME_MASK;
 
 // netdb.h
-pub const NI_MAXHOST: crate::socklen_t = 1025;
-pub const NI_MAXSERV: crate::socklen_t = 32;
-pub const NI_NOFQDN: crate::socklen_t = 0x1;
-pub const NI_NUMERICHOST: crate::socklen_t = 0x2;
-pub const NI_NAMEREQD: crate::socklen_t = 0x4;
-pub const NI_NUMERICSERV: crate::socklen_t = 0x8;
-pub const NI_DGRAM: crate::socklen_t = 0x10;
-pub const NI_NUMERICSCOPE: crate::socklen_t = 0x40;
+pub const NI_MAXHOST: socklen_t = 1025;
+pub const NI_MAXSERV: socklen_t = 32;
+pub const NI_NOFQDN: socklen_t = 0x1;
+pub const NI_NUMERICHOST: socklen_t = 0x2;
+pub const NI_NAMEREQD: socklen_t = 0x4;
+pub const NI_NUMERICSERV: socklen_t = 0x8;
+pub const NI_DGRAM: socklen_t = 0x10;
+pub const NI_NUMERICSCOPE: socklen_t = 0x40;
 pub const EAI_AGAIN: c_int = 2;
 pub const EAI_BADFLAGS: c_int = 3;
 pub const EAI_FAIL: c_int = 4;
@@ -1149,7 +1145,7 @@ pub const NFSMNT_ACDIRMAX: c_int = 0x0800;
 pub const CPUSTATES: c_int = 4;
 
 // semaphore.h
-pub const SEM_FAILED: *mut sem_t = -1isize as *mut crate::sem_t;
+pub const SEM_FAILED: *mut sem_t = -1isize as *mut sem_t;
 
 // spawn.h
 pub const POSIX_SPAWN_SETPGROUP: c_short = 0x1;
@@ -1434,7 +1430,7 @@ pub const IPC_W: c_int = 0o0200;
 pub const IPC_O: c_int = 0o1000;
 pub const IPC_NOERROR: c_int = 0o10000;
 pub const IPC_STAT: c_int = 102;
-pub const IPC_PRIVATE: crate::key_t = -1;
+pub const IPC_PRIVATE: key_t = -1;
 pub const SHM_LOCK: c_int = 201;
 pub const SHM_UNLOCK: c_int = 202;
 
@@ -2078,8 +2074,8 @@ pub const POWER_9: c_int = 0x20000;
 // sys/time.h
 pub const FD_SETSIZE: c_int = 65534;
 pub const TIMEOFDAY: c_int = 9;
-pub const CLOCK_REALTIME: crate::clockid_t = TIMEOFDAY as clockid_t;
-pub const CLOCK_MONOTONIC: crate::clockid_t = 10;
+pub const CLOCK_REALTIME: clockid_t = TIMEOFDAY as clockid_t;
+pub const CLOCK_MONOTONIC: clockid_t = 10;
 pub const TIMER_ABSTIME: c_int = 999;
 pub const ITIMER_REAL: c_int = 0;
 pub const ITIMER_VIRTUAL: c_int = 1;
@@ -2096,8 +2092,8 @@ pub const DST_USA: c_int = 1;
 pub const DST_WET: c_int = 3;
 
 // sys/termio.h
-pub const CSTART: crate::tcflag_t = 0o21;
-pub const CSTOP: crate::tcflag_t = 0o23;
+pub const CSTART: tcflag_t = 0o21;
+pub const CSTOP: tcflag_t = 0o23;
 pub const TCGETA: c_int = TIOC | 5;
 pub const TCSETA: c_int = TIOC | 6;
 pub const TCSETAW: c_int = TIOC | 7;
@@ -2173,44 +2169,44 @@ pub const _W_STRC: c_int = 0x0000007f;
 
 // termios.h
 pub const NCCS: usize = 16;
-pub const OLCUC: crate::tcflag_t = 2;
-pub const CSIZE: crate::tcflag_t = 0x00000030;
-pub const CS5: crate::tcflag_t = 0x00000000;
-pub const CS6: crate::tcflag_t = 0x00000010;
-pub const CS7: crate::tcflag_t = 0x00000020;
-pub const CS8: crate::tcflag_t = 0x00000030;
-pub const CSTOPB: crate::tcflag_t = 0x00000040;
-pub const ECHO: crate::tcflag_t = 0x00000008;
-pub const ECHOE: crate::tcflag_t = 0x00000010;
-pub const ECHOK: crate::tcflag_t = 0x00000020;
-pub const ECHONL: crate::tcflag_t = 0x00000040;
-pub const ECHOCTL: crate::tcflag_t = 0x00020000;
-pub const ECHOPRT: crate::tcflag_t = 0x00040000;
-pub const ECHOKE: crate::tcflag_t = 0x00080000;
-pub const IGNBRK: crate::tcflag_t = 0x00000001;
-pub const BRKINT: crate::tcflag_t = 0x00000002;
-pub const IGNPAR: crate::tcflag_t = 0x00000004;
-pub const PARMRK: crate::tcflag_t = 0x00000008;
-pub const INPCK: crate::tcflag_t = 0x00000010;
-pub const ISTRIP: crate::tcflag_t = 0x00000020;
-pub const INLCR: crate::tcflag_t = 0x00000040;
-pub const IGNCR: crate::tcflag_t = 0x00000080;
-pub const ICRNL: crate::tcflag_t = 0x00000100;
-pub const IXON: crate::tcflag_t = 0x00000200;
-pub const IXOFF: crate::tcflag_t = 0x00000400;
-pub const IXANY: crate::tcflag_t = 0x00001000;
-pub const IMAXBEL: crate::tcflag_t = 0x00010000;
-pub const OPOST: crate::tcflag_t = 0x00000001;
-pub const ONLCR: crate::tcflag_t = 0x00000004;
-pub const OCRNL: crate::tcflag_t = 0x00000008;
-pub const ONOCR: crate::tcflag_t = 0x00000010;
-pub const ONLRET: crate::tcflag_t = 0x00000020;
-pub const CREAD: crate::tcflag_t = 0x00000080;
-pub const IEXTEN: crate::tcflag_t = 0x00200000;
-pub const TOSTOP: crate::tcflag_t = 0x00010000;
-pub const FLUSHO: crate::tcflag_t = 0x00100000;
-pub const PENDIN: crate::tcflag_t = 0x20000000;
-pub const NOFLSH: crate::tcflag_t = 0x00000080;
+pub const OLCUC: tcflag_t = 2;
+pub const CSIZE: tcflag_t = 0x00000030;
+pub const CS5: tcflag_t = 0x00000000;
+pub const CS6: tcflag_t = 0x00000010;
+pub const CS7: tcflag_t = 0x00000020;
+pub const CS8: tcflag_t = 0x00000030;
+pub const CSTOPB: tcflag_t = 0x00000040;
+pub const ECHO: tcflag_t = 0x00000008;
+pub const ECHOE: tcflag_t = 0x00000010;
+pub const ECHOK: tcflag_t = 0x00000020;
+pub const ECHONL: tcflag_t = 0x00000040;
+pub const ECHOCTL: tcflag_t = 0x00020000;
+pub const ECHOPRT: tcflag_t = 0x00040000;
+pub const ECHOKE: tcflag_t = 0x00080000;
+pub const IGNBRK: tcflag_t = 0x00000001;
+pub const BRKINT: tcflag_t = 0x00000002;
+pub const IGNPAR: tcflag_t = 0x00000004;
+pub const PARMRK: tcflag_t = 0x00000008;
+pub const INPCK: tcflag_t = 0x00000010;
+pub const ISTRIP: tcflag_t = 0x00000020;
+pub const INLCR: tcflag_t = 0x00000040;
+pub const IGNCR: tcflag_t = 0x00000080;
+pub const ICRNL: tcflag_t = 0x00000100;
+pub const IXON: tcflag_t = 0x00000200;
+pub const IXOFF: tcflag_t = 0x00000400;
+pub const IXANY: tcflag_t = 0x00001000;
+pub const IMAXBEL: tcflag_t = 0x00010000;
+pub const OPOST: tcflag_t = 0x00000001;
+pub const ONLCR: tcflag_t = 0x00000004;
+pub const OCRNL: tcflag_t = 0x00000008;
+pub const ONOCR: tcflag_t = 0x00000010;
+pub const ONLRET: tcflag_t = 0x00000020;
+pub const CREAD: tcflag_t = 0x00000080;
+pub const IEXTEN: tcflag_t = 0x00200000;
+pub const TOSTOP: tcflag_t = 0x00010000;
+pub const FLUSHO: tcflag_t = 0x00100000;
+pub const PENDIN: tcflag_t = 0x20000000;
+pub const NOFLSH: tcflag_t = 0x00000080;
 pub const VINTR: usize = 0;
 pub const VQUIT: usize = 1;
 pub const VERASE: usize = 2;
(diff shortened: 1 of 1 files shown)


src/unix/aix/mod.rs
tgross35, line 1: "aix: add new type aliases": as discussed elsewhere, avoid adding API with no consumer

src/unix/aix/mod.rs
```diff
@@ -3427,29 +3377,25 @@ extern "C" {
...
         flags: c_int,
-        new_value: *const crate::itimerspec,
-        old_value: *mut crate::itimerspec,
+        new_value: *const itimerspec,
+        old_value: *mut itimerspec,
     ) -> c_int;
-    pub fn truncate64(path: *const c_char, length: off64_t) -> c_int;
-    pub fn uname(buf: *mut crate::utsname) -> c_int;
+    pub fn truncate64(path: *const c_char, length: off_t) -> c_int;
+    pub fn uname(buf: *mut utsname) -> c_int;
```
tgross35, line 3353: "aix: clean up item paths": Even if equivalent, `*64` functions should always use `*64` types, it's confusing otherwise.

src/unix/aix/mod.rs
```diff
@@ -3233,6 +3233,8 @@ extern "C" {
     ) -> c_int;
     pub fn pread64(fd: c_int, buf: *mut c_void, count: size_t, offset: off_t) -> ssize_t;
     pub fn preadv(fd: c_int, iov: *const crate::iovec, iovcnt: c_int, offset: offset_t) -> ssize_t;
+    pub fn ptrace(request: c_int, id: c_int, address: *mut c_int, data: c_int, buffer: *mut c_int);
+    pub fn ptracex(request: c_int, id: c_int, address: c_longlong, data: c_int, buffer: *mut c_int);
```
tgross35: "aix: add new ptrace routines": is there a usecase for these?

build.rs
```diff
@@ -157,6 +158,11 @@ fn main() {
         set_cfg("linux_time_bits64");
     }
 
+    let aix_large_file_api = env_flag("CARGO_CFG_LIBC_UNSTABLE_LARGE_FILE_API");
+    if aix_large_file_api {
+        set_cfg("aix_large_file_api");
+    }
```
tgross35: "aix: add cfg for certain lfs bindings": This is effectively deleting API that people may be using?

src/unix/aix/mod.rs
```diff
@@ -156,6 +156,32 @@ c_enum! {
...
+    pub struct complex_lock_data {
+        pub status: complex_lock_status,
+        pub flags: c_short,
+        pub recursion_depth: c_short,
+        pub reserved: c_uint,
+    }
+
+    pub struct drw_lock_data {
+        pub status: complex_lock_status,
+    }
```
tgross35: "aix: add new auxiliary lock records": is there a usecase for these?

src/unix/aix/mod.rs
```diff
@@ -182,6 +182,56 @@ s! {
...
+
+    pub struct fsid {
+        pub val: [c_uint; 2],
+    }
+
+    pub struct sigset32_t {
+        pub losigs: c_uint,
+        pub hisigs: c_uint,
+    }
+
```
tgross35: "aix: add new records": same question

src/unix/aix/mod.rs
```diff
@@ -2322,6 +2322,14 @@ pub const TIOCMBIC: c_int = u32_cast_int(0x8004746b);
 pub const TIOCMGET: c_int = 0x4004746a;
 pub const TIOCREMOTE: c_int = u32_cast_int(0x80047469);
 
+// sys/types.h
+pub const NULL: c_int = 0;
+pub const TRUE: c_int = 1;
+pub const FALSE: c_int = 0;
+pub const NBBY: c_int = 8;
+pub const FHSIZE: c_int = 32;
+pub const MAXFIDSZ: c_int = 64;
```
tgross35: "aix: add new constants": In general, we'd prefer for any API associated with a new header (i.e. one where we don't already have a `// ... .h` comment) to go in `src/new`, since we're trying to nudge the structure there anyway.

But the same thing applies, I'm not sure if there is a consumer.

src/unix/aix/powerpc64.rs
tgross35, line 1: "aix: move and fix powerpc64 definitions into aix": I have mixed feelings on this change. On one hand there are likely no other platforms we will ever need to support with AIX, according to https://en.wikipedia.org/wiki/IBM_AIX. On the other hand, having types like `mcontext_t` in a non-arch-specific module feels very wrong.

I think I'd have a mild preference of moving this to `src/new` and matching system headers, if moved at all, that's one fewer history jump in the blame once everything makes it to `new` eventually anyway. But I'll leave it up to the target maintainers.

tgross35 (changes requested):
@xingxue-ibm would you be able to take a look?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5259/67c9a34fda91bc26a50cd907b1e8a65ee32bfb65..5536ea81c9d0cfb8e1c26521ef07a056b7c0777a)*

src/unix/aix/mod.rs
dybucc: removed.

src/unix/aix/mod.rs
dybucc: done.

src/unix/aix/mod.rs
dybucc: none that i know of. i've removed them now.

build.rs
dybucc: that's right.

src/unix/aix/mod.rs
dybucc: yup, they're used in the lock_data_instrumented type that was previously in the aix/powerpc64 module. we were exposing it as an opaque type even though upstream has a full definition. i've merged those commits to make it more cohesive.

not sure if that counts as an api addition.

src/unix/aix/mod.rs
dybucc: removed.

src/unix/aix/mod.rs
dybucc: removed.

dybucc:
i'll change the pr status once the target maintainer answers.

src/unix/aix/mod.rs
tgross35: I'm sure somebody wouldn't mind having them, but that can come separately from any kind of cleanup

src/unix/aix/mod.rs
tgross35: As far as I can tell, these are kernel-internal structures so we're probably safer leaving them opaque. That tends to be the rule of thumb for anything not publicly documented, and I can't find any manpages on these.

build.rs
tgross35: Please always call out breakage in commit messages, I almost overlooked this aspect.

What was the thought process here? Outright removing these is likely to break a lot of people.

dybucc: done. noted.

what do you mean by the thought process? much like other cfgs, i added them to fit an upstream feature test macro that is checked before making the type definitions available. i didn't mention it in the commit because i thought not marking the pr as stable-nominated would do.

src/unix/aix/mod.rs
dybucc: except they're not. i explicitly skipped those and, in fact, added a deprecation notice to a type that _is_ only exposed when programming against the kernel [^1].

[^1]: https://github.com/rust-lang/libc/pull/5259/changes#diff-cdca763d7494a8b247c48e3d317bd4dfbd7c2792f4363fa5070805958650b44aR993-R1017

build.rs
tgross35: We won't have any use for this config in 1.0, we can just delete the *64 versions on platforms where they're identical.

Most PRs should be intended for backport - I hadn't realized you weren't on the same page. I'll post more about this at https://github.com/rust-lang/libc/pull/5170 very soon.

src/unix/aix/mod.rs
tgross35: I still had no luck couldn't find something in their manpages, could you share a link?

build.rs
dybucc: i think that much i already understand (meaning the 64-bit types, not the backporting stuff.) the thing with the types i specifically gated behind the cfg is that they are *not* equivalent to the unsuffixed types unless the upstream macro is defined. in fact, there's a bunch more 64-suffixed types that i did not annotate but rather deprecated because they are, indeed, equivalent if we assume the macro is defined. only those that i annotated in that commit are not; namely, statfs, statvfs, and routines that use them. the ones that i deprecated have now been annotated with the comment you mention in #4805.

i can't provide links because the sources are not public. there's neither any page in the ibm docs that comments on it.

src/unix/aix/mod.rs
dybucc: sources aren't public and i'm not sure i can legally disclose the header file contents. still, i can say it's quite odd. the header file in which those types are declared in is not gated behind the kernel macros i've seen aix use in other header files. and yet here [^1], there's mention of one of the routines in the header file being a so called kernel service, when that same routine is not gated behind kernel macros in said header file.

[^1]: https://www.ibm.com/docs/en/aix/7.3.0?topic=s-simple-lock-init-kernel-service

xingxue-ibm:
Thanks for the effort, @dybucc! Your contribution is greatly appreciated. As discussed with the Rust product team, we believe it is not entirely out of the question that we may add AIX 32-bit support in the future. As such, we kindly suggest holding off on the cleanup for now.

src/unix/aix/mod.rs
tgross35: Right, I'm not asking to look at the headers (I do have access). This isn't about what's gated behind macros, headers invariably have a ton of internal and public API mixed together because that's just how C is. Having documentation in the manpages is kind of supposed to be the way to tell whether something is user-facing or not.

That's not super reliable on all platforms because some are just poorly documented, but we can usually rely on it for AIX (at [this site](https://www.ibm.com/docs/en/aix/7.3.0?topic=o-open-openat-openx-openxat-open64-open64at-open64x-open64xat-creat-creat64-subroutine)) and glibc at least.

Anyway I'm fine to take the structural change here, but the fields and new structures should be made private.

src/unix/aix/mod.rs
```diff
@@ -80,12 +84,254 @@ c_enum! {
...
+        #[cfg(debug_assertions)]
+        pub unlock_caller: tid_t,
+        #[cfg(debug_assertions)]
+        pub lock_cpuid: c_int,
+        #[cfg(debug_assertions)]
+        pub dbg_zero: c_int,
+        #[cfg(debug_assertions)]
+        pub unlock_cpuid: c_int,
+        #[cfg(debug_assertions)]
+        pub dbg_flags: c_int,
```
tgross35: This appears to give the struct a different size based on how you build `libc`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5259/c2ef687b124c334f25d4ccbddff4eeb726ec0a9c..faccc1c44f0a411028418ecb6bcdde4e1820aed1)*

build.rs
tgross35: Are our current definitions (before this PR) correct if we assume the macro is defined?

dybucc: i just looked through the header files again. i'm not sure anymore why is it that i gated those records behind a cfg. there's no uses of the macro in their header files and they are always exposed.

with that, there's only two things that remain true:

- statfs is equivalent to statfs64 if targetting a 64-bit target.
  - this is because statfs has a field that is only present when the target has 64-bit machine word size. otherwise they differ.
- statvfs is equivalent to statvfs64 if targetting a 32-bit target.
  - statvfs has a field of type fsid_t. statvfs64 has in its stead a field of type fsid64_t. fsid_t is only equivalent to fsid64_t under 32-bit targets.

other than that, the definitions are correct as is.

considering [^1], we can't really remove these nor deprecate them altogether. in the multistage plan proposed in #4805, i believe this would translate to:

1. annotate with a fixme comment but also add some other "parameter" to the fixme that indicates it's not meant to be found and replaced with a deprecation notice for all targets. something like
   ```rust
   // FIXME(1.0,deprecate,64): lfs binding to be removed
   ```
2. once we reach the deprecation stage, ensure these alternatively parameterized fixme comments are replaced with conditional deprecations.
3. once we reach the removal stage, don't remove these but rather gate them behind a cfg that makes statfs64 available only under 32-bit targets (where they differ) and statvfs64 only available under 64-bit targets (where they differ.)

i've already went ahead and proceeded with step 1 above. though the parameterization may have to be tweaked a bit if we ever need to depend on some other cfg option (possibly by not just appending a 64 but by using the full predicate, like `// FIXME(1.0,deprecate,target_pointer_width=64)`.)

[^1]: https://github.com/rust-lang/libc/pull/5259#issuecomment-4929421515

dybucc:
> Thanks for the effort, @dybucc! Your contribution is greatly appreciated. As discussed with the Rust product team, we believe it is not entirely out of the question that we may add AIX 32-bit support in the future. As such, we kindly suggest holding off on the cleanup for now.

i'm guessing that means i should hold off on moving the powerpc64 module bindings into the top-level aix module. can i keep then the modifications i made to the types i brought over from the powerpc64 module (of course, moving them back to the powerpc64 submodule)?

src/unix/aix/mod.rs
dybucc: that's right. the header files conditionally include those fields if some DEBUG macro is defined. i thought the nearest equivalent would just be rust's debug_assertions.

src/unix/aix/mod.rs
dybucc: done.

xingxue-ibm:
> > Thanks for the effort, @dybucc! Your contribution is greatly appreciated. As discussed with the Rust product team, we believe it is not entirely out of the question that we may add AIX 32-bit support in the future. As such, we kindly suggest holding off on the cleanup for now.
> 
> i'm guessing that means i should hold off on moving the powerpc64 module bindings into the top-level aix module. can i keep then the modifications i made to the types i brought over from the powerpc64 module (of course, moving them back to the powerpc64 submodule)?

Yes, please hold off on moving the PowerPC64 module for now. Otherwise, feel free to make any changes that improve the implementation. Could you also hold off on backporting any changes to the stable branch for the time being? Again, we greatly appreciate your contributions and thank you for your help!

By the way, how are you testing these changes, and on which AIX OS level(s) have you observed or validated them?

dybucc:
done.

i've been checking the header files on an aix 7.3 machine that the cfarm project provides access to. thus far, testing is pending. i've had no luck cross-compiling a rust toolchain with support for the powerpc64-ibm-aix target, and building straight on the aix machine (to which i'm sshing into) is not quite feasible.

xingxue-ibm:
> done.
> 
> i've been checking the header files on an aix 7.3 machine that the cfarm project provides access to. thus far, testing is pending. i've had no luck cross-compiling a rust toolchain with support for the powerpc64-ibm-aix target, and building straight on the aix machine (to which i'm sshing into) is not quite feasible.

The CI for your branch stopped at ctest with 37 errors.
<details><summary>ctest errors</summary><p>

```
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10951:59: error: invalid application of 'sizeof' to an incomplete type 'struct _file'
  cargo:warning= 10951 | CTEST_EXTERN uint64_t ctest_size_of___file(void) { return sizeof(struct _file); }
  cargo:warning=       |                                                           ^     ~~~~~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10951:73: note: forward declaration of 'struct _file'
  cargo:warning= 10951 | CTEST_EXTERN uint64_t ctest_size_of___file(void) { return sizeof(struct _file); }
  cargo:warning=       |                                                                         ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10952:60: error: invalid application of 'alignof' to an incomplete type 'struct _file'
  cargo:warning= 10952 | CTEST_EXTERN uint64_t ctest_align_of___file(void) { return CTEST_ALIGNOF(struct _file); }
  cargo:warning=       |                                                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:96:30: note: expanded from macro 'CTEST_ALIGNOF'
  cargo:warning=   96 |     #define CTEST_ALIGNOF(T) _Alignof(T)
  cargo:warning=      |                              ^       ~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10952:81: note: forward declaration of 'struct _file'
  cargo:warning= 10952 | CTEST_EXTERN uint64_t ctest_align_of___file(void) { return CTEST_ALIGNOF(struct _file); }
  cargo:warning=       |                                                                                 ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10957:73: error: use of '_simple_lock' with tag type that does not match previous declaration
  cargo:warning= 10957 | CTEST_EXTERN uint64_t ctest_size_of___simple_lock(void) { return sizeof(struct _simple_lock); }
  cargo:warning=       |                                                                         ^~~~~~
  cargo:warning=       |                                                                         union
  cargo:warning=/usr/include/sys/lock_def.h:159:7: note: previous use is here
  cargo:warning=  159 | union _simple_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10958:81: error: use of '_simple_lock' with tag type that does not match previous declaration
  cargo:warning= 10958 | CTEST_EXTERN uint64_t ctest_align_of___simple_lock(void) { return CTEST_ALIGNOF(struct _simple_lock); }
  cargo:warning=       |                                                                                 ^~~~~~
  cargo:warning=       |                                                                                 union
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:96:39: note: expanded from macro 'CTEST_ALIGNOF'
  cargo:warning=   96 |     #define CTEST_ALIGNOF(T) _Alignof(T)
  cargo:warning=      |                                       ^
  cargo:warning=/usr/include/sys/lock_def.h:159:7: note: previous use is here
  cargo:warning=  159 | union _simple_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10960:74: error: use of '_complex_lock' with tag type that does not match previous declaration
  cargo:warning= 10960 | CTEST_EXTERN uint64_t ctest_size_of___complex_lock(void) { return sizeof(struct _complex_lock); }
  cargo:warning=       |                                                                          ^~~~~~
  cargo:warning=       |                                                                          union
  cargo:warning=/usr/include/sys/lock_def.h:165:7: note: previous use is here
  cargo:warning=  165 | union _complex_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10961:82: error: use of '_complex_lock' with tag type that does not match previous declaration
  cargo:warning= 10961 | CTEST_EXTERN uint64_t ctest_align_of___complex_lock(void) { return CTEST_ALIGNOF(struct _complex_lock); }
  cargo:warning=       |                                                                                  ^~~~~~
  cargo:warning=       |                                                                                  union
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:96:39: note: expanded from macro 'CTEST_ALIGNOF'
  cargo:warning=   96 |     #define CTEST_ALIGNOF(T) _Alignof(T)
  cargo:warning=      |                                       ^
  cargo:warning=/usr/include/sys/lock_def.h:165:7: note: previous use is here
  cargo:warning=  165 | union _complex_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10963:70: error: use of '_drw_lock' with tag type that does not match previous declaration
  cargo:warning= 10963 | CTEST_EXTERN uint64_t ctest_size_of___drw_lock(void) { return sizeof(struct _drw_lock); }
  cargo:warning=       |                                                                      ^~~~~~
  cargo:warning=       |                                                                      union
  cargo:warning=/usr/include/sys/lock_def.h:171:7: note: previous use is here
  cargo:warning=  171 | union _drw_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10964:78: error: use of '_drw_lock' with tag type that does not match previous declaration
  cargo:warning= 10964 | CTEST_EXTERN uint64_t ctest_align_of___drw_lock(void) { return CTEST_ALIGNOF(struct _drw_lock); }
  cargo:warning=       |                                                                              ^~~~~~
  cargo:warning=       |                                                                              union
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:96:39: note: expanded from macro 'CTEST_ALIGNOF'
  cargo:warning=   96 |     #define CTEST_ALIGNOF(T) _Alignof(T)
  cargo:warning=      |                                       ^
  cargo:warning=/usr/include/sys/lock_def.h:171:7: note: previous use is here
  cargo:warning=  171 | union _drw_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:16515:9: error: incompatible pointer types assigning to 'ctest_field_ty__flock64__l_start' (aka 'long *') from 'off64_t *' (aka 'long long *') [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 16515 |     ptr = &b->l_start;
  cargo:warning=       |         ^ ~~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:16518:12: error: incompatible pointer types returning 'off64_t *' (aka 'long long *') from a function with result type 'ctest_field_ty__flock64__l_start' (aka 'long *') [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 16518 |     return &b->l_start;
  cargo:warning=       |            ^~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:16527:9: error: incompatible pointer types assigning to 'ctest_field_ty__flock64__l_len' (aka 'long *') from 'off64_t *' (aka 'long long *') [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 16527 |     ptr = &b->l_len;
  cargo:warning=       |         ^ ~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:16530:12: error: incompatible pointer types returning 'off64_t *' (aka 'long long *') from a function with result type 'ctest_field_ty__flock64__l_len' (aka 'long *') [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 16530 |     return &b->l_len;
  cargo:warning=       |            ^~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:19359:9: error: incompatible pointer types assigning to 'ctest_field_ty__stat64__st_size' (aka 'long *') from 'off64_t *' (aka 'long long *') [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 19359 |     ptr = &b->st_size;
  cargo:warning=       |         ^ ~~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:19362:12: error: incompatible pointer types returning 'off64_t *' (aka 'long long *') from a function with result type 'ctest_field_ty__stat64__st_size' (aka 'long *') [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 19362 |     return &b->st_size;
  cargo:warning=       |            ^~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:19755:9: error: incompatible pointer types assigning to 'ctest_field_ty__poll_ctl_ext_t__u' (aka 'union __c_anonymous_poll_ctl_ext_t_u *') from 'union (unnamed union at /usr/include/sys/pollset.h:101:2) *' [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 19755 |     ptr = &b->u;
  cargo:warning=       |         ^ ~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:19758:12: error: incompatible pointer types returning 'union (unnamed union at /usr/include/sys/pollset.h:101:2) *' from a function with result type 'ctest_field_ty__poll_ctl_ext_t__u' (aka 'union __c_anonymous_poll_ctl_ext_t_u *') [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 19758 |     return &b->u;
  cargo:warning=       |            ^~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:21627:9: error: incompatible pointer types assigning to 'ctest_field_ty__pollfd_ext_t__u' (aka 'union __c_anonymous_pollfd_ext_t_u *') from 'union (unnamed union at /usr/include/sys/poll.h:70:2) *' [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 21627 |     ptr = &b->u;
  cargo:warning=       |         ^ ~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:21630:12: error: incompatible pointer types returning 'union (unnamed union at /usr/include/sys/poll.h:70:2) *' from a function with result type 'ctest_field_ty__pollfd_ext_t__u' (aka 'union __c_anonymous_pollfd_ext_t_u *') [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 21630 |     return &b->u;
  cargo:warning=       |            ^~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24892:35: error: invalid application of 'sizeof' to an incomplete type 'struct _file'
  cargo:warning= 24892 |     const uint8_t is_padding_byte[sizeof(struct _file)],
  cargo:warning=       |                                   ^     ~~~~~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24890:21: note: forward declaration of 'struct _file'
  cargo:warning= 24890 | CTEST_EXTERN struct _file ctest_roundtrip___file(
  cargo:warning=       |                     ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24893:25: error: invalid application of 'sizeof' to an incomplete type 'struct _file'
  cargo:warning= 24893 |     uint8_t value_bytes[sizeof(struct _file)]
  cargo:warning=       |                         ^     ~~~~~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24890:21: note: forward declaration of 'struct _file'
  cargo:warning= 24890 | CTEST_EXTERN struct _file ctest_roundtrip___file(
  cargo:warning=       |                     ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24891:18: error: variable has incomplete type 'struct _file'
  cargo:warning= 24891 |     struct _file value,
  cargo:warning=       |                  ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24890:21: note: forward declaration of 'struct _file'
  cargo:warning= 24890 | CTEST_EXTERN struct _file ctest_roundtrip___file(
  cargo:warning=       |                     ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24895:21: error: invalid application of 'sizeof' to an incomplete type 'struct _file'
  cargo:warning= 24895 |     int size = (int)sizeof(struct _file);
  cargo:warning=       |                     ^     ~~~~~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24890:21: note: forward declaration of 'struct _file'
  cargo:warning= 24890 | CTEST_EXTERN struct _file ctest_roundtrip___file(
  cargo:warning=       |                     ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24932:14: error: use of '_simple_lock' with tag type that does not match previous declaration
  cargo:warning= 24932 | CTEST_EXTERN struct _simple_lock ctest_roundtrip___simple_lock(
  cargo:warning=       |              ^~~~~~
  cargo:warning=       |              union
  cargo:warning=/usr/include/sys/lock_def.h:159:7: note: previous use is here
  cargo:warning=  159 | union _simple_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24933:5: error: use of '_simple_lock' with tag type that does not match previous declaration
  cargo:warning= 24933 |     struct _simple_lock value,
  cargo:warning=       |     ^~~~~~
  cargo:warning=       |     union
  cargo:warning=/usr/include/sys/lock_def.h:159:7: note: previous use is here
  cargo:warning=  159 | union _simple_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24934:42: error: use of '_simple_lock' with tag type that does not match previous declaration
  cargo:warning= 24934 |     const uint8_t is_padding_byte[sizeof(struct _simple_lock)],
  cargo:warning=       |                                          ^~~~~~
  cargo:warning=       |                                          union
  cargo:warning=/usr/include/sys/lock_def.h:159:7: note: previous use is here
  cargo:warning=  159 | union _simple_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24935:32: error: use of '_simple_lock' with tag type that does not match previous declaration
  cargo:warning= 24935 |     uint8_t value_bytes[sizeof(struct _simple_lock)]
  cargo:warning=       |                                ^~~~~~
  cargo:warning=       |                                union
  cargo:warning=/usr/include/sys/lock_def.h:159:7: note: previous use is here
  cargo:warning=  159 | union _simple_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24937:28: error: use of '_simple_lock' with tag type that does not match previous declaration
  cargo:warning= 24937 |     int size = (int)sizeof(struct _simple_lock);
  cargo:warning=       |                            ^~~~~~
  cargo:warning=       |                            union
  cargo:warning=/usr/include/sys/lock_def.h:159:7: note: previous use is here
  cargo:warning=  159 | union _simple_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24953:14: error: use of '_complex_lock' with tag type that does not match previous declaration
  cargo:warning= 24953 | CTEST_EXTERN struct _complex_lock ctest_roundtrip___complex_lock(
  cargo:warning=       |              ^~~~~~
  cargo:warning=       |              union
  cargo:warning=/usr/include/sys/lock_def.h:165:7: note: previous use is here
  cargo:warning=  165 | union _complex_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24954:5: error: use of '_complex_lock' with tag type that does not match previous declaration
  cargo:warning= 24954 |     struct _complex_lock value,
  cargo:warning=       |     ^~~~~~
  cargo:warning=       |     union
  cargo:warning=/usr/include/sys/lock_def.h:165:7: note: previous use is here
  cargo:warning=  165 | union _complex_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24955:42: error: use of '_complex_lock' with tag type that does not match previous declaration
  cargo:warning= 24955 |     const uint8_t is_padding_byte[sizeof(struct _complex_lock)],
  cargo:warning=       |                                          ^~~~~~
  cargo:warning=       |                                          union
  cargo:warning=/usr/include/sys/lock_def.h:165:7: note: previous use is here
  cargo:warning=  165 | union _complex_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24956:32: error: use of '_complex_lock' with tag type that does not match previous declaration
  cargo:warning= 24956 |     uint8_t value_bytes[sizeof(struct _complex_lock)]
  cargo:warning=       |                                ^~~~~~
  cargo:warning=       |                                union
  cargo:warning=/usr/include/sys/lock_def.h:165:7: note: previous use is here
  cargo:warning=  165 | union _complex_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24958:28: error: use of '_complex_lock' with tag type that does not match previous declaration
  cargo:warning= 24958 |     int size = (int)sizeof(struct _complex_lock);
  cargo:warning=       |                            ^~~~~~
  cargo:warning=       |                            union
  cargo:warning=/usr/include/sys/lock_def.h:165:7: note: previous use is here
  cargo:warning=  165 | union _complex_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24974:14: error: use of '_drw_lock' with tag type that does not match previous declaration
  cargo:warning= 24974 | CTEST_EXTERN struct _drw_lock ctest_roundtrip___drw_lock(
  cargo:warning=       |              ^~~~~~
  cargo:warning=       |              union
  cargo:warning=/usr/include/sys/lock_def.h:171:7: note: previous use is here
  cargo:warning=  171 | union _drw_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24975:5: error: use of '_drw_lock' with tag type that does not match previous declaration
  cargo:warning= 24975 |     struct _drw_lock value,
  cargo:warning=       |     ^~~~~~
  cargo:warning=       |     union
  cargo:warning=/usr/include/sys/lock_def.h:171:7: note: previous use is here
  cargo:warning=  171 | union _drw_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24976:42: error: use of '_drw_lock' with tag type that does not match previous declaration
  cargo:warning= 24976 |     const uint8_t is_padding_byte[sizeof(struct _drw_lock)],
  cargo:warning=       |                                          ^~~~~~
  cargo:warning=       |                                          union
  cargo:warning=/usr/include/sys/lock_def.h:171:7: note: previous use is here
  cargo:warning=  171 | union _drw_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24977:32: error: use of '_drw_lock' with tag type that does not match previous declaration
  cargo:warning= 24977 |     uint8_t value_bytes[sizeof(struct _drw_lock)]
  cargo:warning=       |                                ^~~~~~
  cargo:warning=       |                                union
  cargo:warning=/usr/include/sys/lock_def.h:171:7: note: previous use is here
  cargo:warning=  171 | union _drw_lock{
  cargo:warning=      |       ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24979:28: error: use of '_drw_lock' with tag type that does not match previous declaration
  cargo:warning= 24979 |     int size = (int)sizeof(struct _drw_lock);
  cargo:warning=       |                            ^~~~~~
  cargo:warning=       |                            union
  cargo:warning=/usr/include/sys/lock_def.h:171:7: note: previous use is here
  cargo:warning=  171 | union _drw_lock{
  cargo:warning=      |       ^
  cargo:warning=37 errors generated.
```
</p></details>

dybucc:
Thanks! I'll start looking into this.

dybucc:
Concerning the `_file` type. That one should be skipped on CI for two reasons:

- It's not an actual type. It's meant to replace `file`, which is an opaque
  type but for some reason was exposed with a full definition before this PR.
- It's not spelled right. That's because `file` is currently deprecated and
  will eventually be removed in favor of `_file`. Then `_file` will be renamed
  to `file`.

`libc-test` doesn't have skips set up for systems we don't test in CI, so I
can't do anything about that one.

Concerning the lock types: I provided bindings for `struct`s instead of
`union`s. I checked the headers and realized I had completely missed that.

Concerning the pointer field type errors, those I forgot to change back when
restoring the `powerpc64` module, so they used the non-LFS types instead of the
`64`-suffixed types.

The rest of the errors are with two `struct`s, `pollfd_ext_t` and
`poll_ctl_ext_t`, which conditionally have either one of an anonymous field or
a named field depending on the value of some macros. c.f. header files
`sys/pollset.h` and `sys/poll.h` under `/usr/include`.

Except for the first and last errors commented above, I just pushed a patch
that should fix them. Mind rerunning tests and seeing whether the pass now?

xingxue-ibm:
> Except for the first and last errors commented above, I just pushed a patch that should fix them. Mind rerunning tests and seeing whether the pass now?

Thanks, reduced to 10 errors.
<details><summary>Errors</summary><p>

```
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10951:59: error: invalid application of 'sizeof' to an incomplete type 'struct _file'
  cargo:warning= 10951 | CTEST_EXTERN uint64_t ctest_size_of___file(void) { return sizeof(struct _file); }
  cargo:warning=       |                                                           ^     ~~~~~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10951:73: note: forward declaration of 'struct _file'
  cargo:warning= 10951 | CTEST_EXTERN uint64_t ctest_size_of___file(void) { return sizeof(struct _file); }
  cargo:warning=       |                                                                         ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10952:60: error: invalid application of 'alignof' to an incomplete type 'struct _file'
  cargo:warning= 10952 | CTEST_EXTERN uint64_t ctest_align_of___file(void) { return CTEST_ALIGNOF(struct _file); }
  cargo:warning=       |                                                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:96:30: note: expanded from macro 'CTEST_ALIGNOF'
  cargo:warning=   96 |     #define CTEST_ALIGNOF(T) _Alignof(T)
  cargo:warning=      |                              ^       ~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:10952:81: note: forward declaration of 'struct _file'
  cargo:warning= 10952 | CTEST_EXTERN uint64_t ctest_align_of___file(void) { return CTEST_ALIGNOF(struct _file); }
  cargo:warning=       |                                                                                 ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:19755:9: error: incompatible pointer types assigning to 'ctest_field_ty__poll_ctl_ext_t__u' (aka 'union __c_anonymous_poll_ctl_ext_t_u *') from 'union (unnamed union at /usr/include/sys/pollset.h:101:2) *' [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 19755 |     ptr = &b->u;
  cargo:warning=       |         ^ ~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:19758:12: error: incompatible pointer types returning 'union (unnamed union at /usr/include/sys/pollset.h:101:2) *' from a function with result type 'ctest_field_ty__poll_ctl_ext_t__u' (aka 'union __c_anonymous_poll_ctl_ext_t_u *') [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 19758 |     return &b->u;
  cargo:warning=       |            ^~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:21627:9: error: incompatible pointer types assigning to 'ctest_field_ty__pollfd_ext_t__u' (aka 'union __c_anonymous_pollfd_ext_t_u *') from 'union (unnamed union at /usr/include/sys/poll.h:70:2) *' [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 21627 |     ptr = &b->u;
  cargo:warning=       |         ^ ~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:21630:12: error: incompatible pointer types returning 'union (unnamed union at /usr/include/sys/poll.h:70:2) *' from a function with result type 'ctest_field_ty__pollfd_ext_t__u' (aka 'union __c_anonymous_pollfd_ext_t_u *') [-Werror,-Wincompatible-pointer-types]
  cargo:warning= 21630 |     return &b->u;
  cargo:warning=       |            ^~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24892:35: error: invalid application of 'sizeof' to an incomplete type 'struct _file'
  cargo:warning= 24892 |     const uint8_t is_padding_byte[sizeof(struct _file)],
  cargo:warning=       |                                   ^     ~~~~~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24890:21: note: forward declaration of 'struct _file'
  cargo:warning= 24890 | CTEST_EXTERN struct _file ctest_roundtrip___file(
  cargo:warning=       |                     ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24893:25: error: invalid application of 'sizeof' to an incomplete type 'struct _file'
  cargo:warning= 24893 |     uint8_t value_bytes[sizeof(struct _file)]
  cargo:warning=       |                         ^     ~~~~~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24890:21: note: forward declaration of 'struct _file'
  cargo:warning= 24890 | CTEST_EXTERN struct _file ctest_roundtrip___file(
  cargo:warning=       |                     ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24891:18: error: variable has incomplete type 'struct _file'
  cargo:warning= 24891 |     struct _file value,
  cargo:warning=       |                  ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24890:21: note: forward declaration of 'struct _file'
  cargo:warning= 24890 | CTEST_EXTERN struct _file ctest_roundtrip___file(
  cargo:warning=       |                     ^
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24895:21: error: invalid application of 'sizeof' to an incomplete type 'struct _file'
  cargo:warning= 24895 |     int size = (int)sizeof(struct _file);
  cargo:warning=       |                     ^     ~~~~~~~~~~~~~~
  cargo:warning=/home/xingxue/libc/target/powerpc64-ibm-aix/debug/build/libc-test-116a5da96aec4195/out/ctest_output.c:24890:21: note: forward declaration of 'struct _file'
  cargo:warning= 24890 | CTEST_EXTERN struct _file ctest_roundtrip___file(
  cargo:warning=       |                     ^
  cargo:warning=10 errors generated.
```
</p></details>

dybucc:
Those I don't think can't do anything about. The `_file` type is temporary
until we can remove `file` and rename `_file` to `file`. The errors concerning
the `union`s can't be fixed either because we're already stretching the
language with anonymous `union`s. Stretching it even more to somehow get
unnamed `struct` fields is likely not something we would want to do. I can make
those fields private, though. It just so happens that they _are_ public when
certain macros are defined. Would like to ask @tgross35 about this.

tgross35:
I haven't yet taken a thorough look at this but will note that this is turning into a pretty big PR. It may be good to split uncontroversial changes into a separate PR so we can actually merge some of it and reduce what's left to focus on.

> The errors concerning the `union`s can't be fixed either because we're already stretching the language with anonymous `union`s. Stretching it even more to somehow get unnamed `struct` fields is likely not something we would want to do. I can make those fields private, though. It just so happens that they _are_ public when certain macros are defined. Would like to ask @tgross35 about this.

`lock_data_instrumented`? Just leave it extern, there's no benefit to filling it in.

dybucc:
> I haven't yet taken a thorough look at this but will note that this is turning into a pretty big PR. It may be good to split uncontroversial changes into a separate PR so we can actually merge some of it and reduce what's left to focus on.

I'm not sure whether there is a controversial change here, though. All conflicts
seem to be resolved, and the above test errors are expected.

> > The errors concerning the `union`s can't be fixed either because we're already stretching the language with anonymous `union`s. Stretching it even more to somehow get unnamed `struct` fields is likely not something we would want to do. I can make those fields private, though. It just so happens that they _are_ public when certain macros are defined. Would like to ask @tgross35 about this.
>
> `lock_data_instrumented`? Just leave it extern, there's no benefit to filling it in.

I was actually referring to `pollfd_ext_t` and `poll_ctl_ext_t`. They contain
fields that are only unnamed in the presence of certain macros. The C tests fail
when trying to compare named unions for those fields (i.e. `__c_anonymous_*`)
with unnamed unions in those fields.

src/unix/aix/mod.rs
tgross35: Hm, that's weird. `#[cfg(debug_assertions)]` wouldn't be accurate here because whatever API we're calling has no clue about Rust's debug assertions so toggling those means you will swap the struct between a correct definition and an incorrect one. Without knowing what it's actually doing with the internals, it's safer to keep this an `extern_ty!`.

src/unix/aix/powerpc64.rs
```diff
@@ -364,6 +364,11 @@ s_no_extra_traits! {
         pub _slockp: *mut lock_data_instrumented,
     }
 
+    #[deprecated(
+        since = "0.2.187",
+        note = "This type is only available when programming against the kernel."
+    )]
+    #[allow(deprecated)]
```
tgross35: "aix(powerpc64): deprecate fileops_t type": is this `allow` still needed? I thought the macro may have been updated recently.

src/unix/aix/powerpc64.rs
```diff
@@ -384,7 +389,13 @@ s_no_extra_traits! {
...
         >,
         pub fo_select: Option<
-            extern "C" fn(file: *mut file, a: c_int, b: *mut c_ushort, c: extern "C" fn()) -> c_int,
+            extern "C" fn(
+                file: *mut file,
+                a: c_int,
+                b: c_ushort,
+                c: *mut c_ushort,
+                c: extern "C" fn(),
+            ) -> c_int,
```
tgross35, line 328: "aix(powerpc64): deprecate fileops_t type": mark this function `unsafe` while you're changing it. Also please update the commit summary to indicate there's a fix here and not just a deprecation.

tgross35:
> > I haven't yet taken a thorough look at this but will note that this is turning into a pretty big PR. It may be good to split uncontroversial changes into a separate PR so we can actually merge some of it and reduce what's left to focus on.
> 
> I'm not sure whether there is a controversial change here, though. All conflicts seem to be resolved, and the above test errors are expected.

It currently passes without errors, right? That needs to be resolved and, of course, this needs the maintainer approval before merge.

By uncontroversial, I mean changes like:

* aix: annotate lfs types as slated for deprecation
* aix: clean up item paths
* aix: annotate certain lfs bindings for deprecation
* aix: annotate most lfs routines for deprecation
* aix: error on unsupported target
* aix(powerpc64): correct padding and reorganize (technically breaking but seems fine)
* aix(powerpc64): deprecate _kernel_simple_lock
* aix(powerpc64): deprecate fileops_t type (though please update the summary to say it's fixing the field)
* aix(powerpc64): add system header comments

Which have more or less been reviewed without issue. "aix: tweak poll_ctl_ext ident", "aix(powerpc64): define full lock_data_instrumented" and "aix(powerpc64): deprecate file type and add _file" OTOH are more opinionated and seem to introduce test failures, and there's a "chore: wip" commit.

It's just less for the reviewers to keep in our mental cache if we can cross some things off the list and look at a few dozen line diff for what's remaining, rather than ~1k total.

> > > The errors concerning the `union`s can't be fixed either because we're already stretching the language with anonymous `union`s. Stretching it even more to somehow get unnamed `struct` fields is likely not something we would want to do. I can make those fields private, though. It just so happens that they _are_ public when certain macros are defined. Would like to ask @tgross35 about this.
> > 
> > 
> > `lock_data_instrumented`? Just leave it extern, there's no benefit to filling it in.
> 
> I was actually referring to `pollfd_ext_t` and `poll_ctl_ext_t`. They contain fields that are only unnamed in the presence of certain macros. The C tests fail when trying to compare named unions for those fields (i.e. `__c_anonymous_*`) with unnamed unions in those fields.

Unfortunately we need to skip fields containing a `__c_anonymous_` type. If it's passing currently though, I'd leave as-is.

xingxue-ibm:
I agree with @tgross35 100%. It’s a good idea to have smaller PRs, each with a clear description of the issue being addressed and the rationale behind the change. This makes reviews more manageable and allows reviewers to focus on a well-defined scope. Smaller, self-contained PRs also make it easier to understand code history in the future, investigate regressions, and revert changes when necessary.

src/unix/aix/mod.rs
dybucc: Noted. I'm remaking the PR now so the changes won't be public for some time.

dybucc:
I'm remaking the PR now into three separate patchsets that should target the
following topics:

- [x] A general clean-up of the `aix` module. This won't be concerned with LFS.
- [x] A general clean-up of the `aix/powerpc64` module. This won't have the
    `lock_data_instrumented` changes but will keep the possibly controversial
    changes to the `struct`s with unnamed `union`s.
- [x] A specific change for LFS stuff in the `aix` top-level module.

Edit: the PR is now separated into three PRs. This one is now concerned with the
general AIX cleanup (not concerning neither one of LFS nor `powerpc64`-specific
bindings.)

@rustbot ready

tgross35:
Seems fine to me, @xingxue-ibm would you be able to run tests?

tgross35:
@xingxue-ibm gentle ping, would you be able to test this?

src/unix/aix/mod.rs
```diff
@@ -546,18 +542,18 @@ s! {
         pub sa_flags: c_int,
     }
 
-    pub struct poll_ctl_ext {
+    pub struct poll_ctl_ext_t {
```
xingxue-ibm: The definition in the AIX header `sys/pollset.h` is as follows. The `_t` suffix in `poll_ctl_ext_t` is purely a C `typedef` alias for the `struct` tag `poll_ctl_ext`. So I think `poll_ctl_ext` is correct.
```
typedef struct poll_ctl_ext {
        uint8_t version;        /* Control descriptor version */
        uint8_t command;        /* PS_ADD, PS_MOD, etc.       */
        short   events;         /* Events to monitor, POLLIN, POLLOUT, etc. */
        int     fd;             /* File descriptor to add/modify/remove     */

        union {
                void    *addr;  /* User-defined pointer */
                uint32_t data32;/* User-defined data    */
                uint64_t data;  /* User-defined data    */
#if __C99_PRAGMA_OPERATOR && !defined(__C99_REQUIRE_FUNC_DECL) && !__cplusplus
        };
#else
        } u;
#endif

        uint64_t reserved64[6]; /* Reserved for future expansion */
} poll_ctl_ext_t;
```

src/unix/aix/mod.rs
```diff
@@ -546,18 +542,18 @@ s! {
...
         pub version: u8,
         pub command: u8,
         pub events: c_short,
         pub fd: c_int,
-        pub u: __poll_ctl_ext_u,
+        pub u: __c_anonymous_poll_ctl_ext_t_u,
```
xingxue-ibm:
```suggestion
        pub u: __c_anonymous_poll_ctl_ext_u,

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5259/023c0a3c184c2d7fcd528b64e867075660a5884d..e6f4800c398f28e3be6791405f857fe6101973f1)*

File: src/unix/aix/mod.rs [hunk 12]
```diff
@@ -566,17 +562,17 @@ s_no_extra_traits! {
 
 cfg_if! {
     if #[cfg(feature = "extra_traits")] {
-        impl PartialEq for __poll_ctl_ext_u {
-            fn eq(&self, other: &__poll_ctl_ext_u) -> bool {
+        impl PartialEq for __c_anonymous_poll_ctl_ext_t_u {
+            fn eq(&self, other: &__c_anonymous_poll_ctl_ext_t_u) -> bool {
```
xingxue-ibm:
```suggestion
        impl PartialEq for __c_anonymous_poll_ctl_ext_u {
            fn eq(&self, other: &__c_anonymous_poll_ctl_ext_u) -> bool {

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5259/023c0a3c184c2d7fcd528b64e867075660a5884d..e6f4800c398f28e3be6791405f857fe6101973f1)*

File: src/unix/aix/mod.rs [hunk 13]
```diff
@@ -566,17 +562,17 @@ s_no_extra_traits! {
...
                     self.addr == other.addr
                         && self.data32 == other.data32
                         && self.data == other.data
                 }
             }
         }
-        impl Eq for __poll_ctl_ext_u {}
-        impl hash::Hash for __poll_ctl_ext_u {
+        impl Eq for __c_anonymous_poll_ctl_ext_t_u {}
+        impl hash::Hash for __c_anonymous_poll_ctl_ext_t_u {
```
xingxue-ibm:
```suggestion
        impl Eq for __c_anonymous_poll_ctl_ext_u {}
        impl hash::Hash for __c_anonymous_poll_ctl_ext_u {

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5259/023c0a3c184c2d7fcd528b64e867075660a5884d..e6f4800c398f28e3be6791405f857fe6101973f1)*

File: src/unix/aix/mod.rs (hunk 10 above)
Comment (xingxue-ibm) [inline src/unix/aix/mod.rs]: The definition in the AIX header `sys/pollset.h` is as follows. The `_t` suffix in `poll_ctl_ext_t` is a C `typedef` alias for the `struct` tag `poll_ctl_ext`. So I think `poll_ctl_ext` is correct.
```
typedef struct poll_ctl_ext {
...
#endif

        uint64_t reserved64[6]; /* Reserved for future expansion */
} poll_ctl_ext_t;
```
```suggestion
    pub struct poll_ctl_ext {

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5259/023c0a3c184c2d7fcd528b64e867075660a5884d..e6f4800c398f28e3be6791405f857fe6101973f1)*

File: src/unix/aix/mod.rs [hunk 14]
```diff
@@ -546,18 +542,18 @@ s! {
...
         pub fd: c_int,
-        pub u: __poll_ctl_ext_u,
+        pub u: __c_anonymous_poll_ctl_ext_t_u,
         reserved64: Padding<[u64; 6]>,
     }
 }
 
 s_no_extra_traits! {
-    pub union __poll_ctl_ext_u {
+    pub union __c_anonymous_poll_ctl_ext_t_u {
```
xingxue-ibm:
```suggestion
    pub union __c_anonymous_poll_ctl_ext_u {

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5259/023c0a3c184c2d7fcd528b64e867075660a5884d..e6f4800c398f28e3be6791405f857fe6101973f1)*

xingxue-ibm:
> @xingxue-ibm gentle ping, would you be able to test this?

Sorry, my bad. I forgot about this one and incorrectly thought we were done. The CI test was failing because of the change from `struct poll_ctl_ext` to `struct poll_ctl_ext_t`. I suggest reverting to `struct poll_ctl_ext`, as that is how it is defined in the AIX header. `poll_ctl_ext_t` is a C `typedef` alias.

src/unix/aix/mod.rs
dybucc: Done.

src/unix/aix/mod.rs
dybucc: Done.

src/unix/aix/mod.rs
dybucc: Done.

src/unix/aix/mod.rs
dybucc: Done.

dybucc:
>> @xingxue-ibm gentle ping, would you be able to test this?
>
>Sorry, my bad. I forgot about this one and incorrectly thought we were
>done. The CI test was failing because of the change from
>`struct poll_ctl_ext` to `struct poll_ctl_ext_t`. I suggest reverting
>to `struct poll_ctl_ext`, as that is how it is defined in the AIX
>header. `poll_ctl_ext_t` is a C `typedef` alias.

Thanks. That's fixed now.

src/unix/aix/mod.rs
dybucc: Done.

xingxue-ibm (approved):
The CI now passes on AIX. LGTM, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5259/008f905256d27eb5dd416d0976678ee6c2855554..a9e7af5279362c828ad0fbca1c092f02b907f2f6)*

tgross35 (approved):
Thanks for authoring and reviewing!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5259/008f905256d27eb5dd416d0976678ee6c2855554..a9e7af5279362c828ad0fbca1c092f02b907f2f6)*

src/unix/aix/powerpc64.rs
tgross35: Outdated comment

src/unix/aix/powerpc64.rs
tgross35: Outdated comment

PR #5263, opened 2026-07-08
apple: add missing SIOC ioctl constants

### LINKED ISSUE #3626: Missing SIOC ioctl constants on MacOS
Target Triple: aarch64-apple-darwin

Hi, I was able to use the SIOCGIFFLAGS ioctl (with the libc::ifreq struct) just fine on Linux, but on Mac, I hit an issue due to the SIOCGIFFLAGS constant not being defined.  It looks like all of the ioctl constants from [sys/sockio.h](https://opensource.apple.com/source/xnu/xnu-201/bsd/sys/sockio.h.auto.html) are missing.

For now I was able to use this workaround:
```rust
#[cfg(any(target_os = "ios", target_os = "macos"))]
const SIOCGIFFLAGS: libc::c_ulong = 0xc0206911; // extracted from macos headers via printing it in a C program

#[cfg(not(any(target_os = "ios", target_os = "macos")))]
const SIOCGIFFLAGS: libc::c_ulong = libc::SIOCGIFFLAGS;
```

yabanana:
# Description

Add the socket and network interface ioctl constants from `<sys/sockio.h>` for Apple targets: everything public in the header except `SIOCGIFADDR`, which is already defined at the BSD level. `SIOCSETVLAN` and `SIOCGETVLAN` are aliases, as in the header.

`sys/sockio.h` is added to the libc-test headers so ctest checks the values.

Closes rust-lang/libc#3626

# Sources

https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/sockio.h

# Checklist

- [x] Tested locally (`cd libc-test && cargo test --target aarch64-apple-darwin`)


src/unix/bsd/apple/mod.rs
```diff
@@ -2316,6 +2316,78 @@ pub const BIOCGDLTLIST: c_ulong = 0xc00c4279;
...
+pub const SIOCSIFASYNCMAP: c_ulong = 0x8020697d;
+pub const SIOCGIFMAC: c_ulong = 0xc0206982;
+pub const SIOCSIFMAC: c_ulong = 0x80206983;
+pub const SIOCSIFKPI: c_ulong = 0x80206986;
+pub const SIOCGIFKPI: c_ulong = 0xc0206987;
+pub const SIOCGIFWAKEFLAGS: c_ulong = 0xc0206988;
+pub const SIOCGIFFUNCTIONALTYPE: c_ulong = 0xc02069ad;
+pub const SIOCSIF6LOWPAN: c_ulong = 0x802069c4;
+pub const SIOCGIF6LOWPAN: c_ulong = 0xc02069c5;
+pub const SIOCGIFDIRECTLINK: c_ulong = 0xc02069de;
```
tgross35: We're trying to nudge anything completely new to the more organized `new` module. Would you be able to put this in a new file under `src/new/apple/xnu/sys`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5263/5848d319e53719d637781eae72de169d8f7867a0..7f4185a94dc2cbfdc28f0bb963ffefd55315a92d)*

tgross35 (changes requested):

src/unix/bsd/apple/mod.rs
JohnTitor: (Maybe it'd be helpful that there's CI gate or something else to enforce it 🤔)

yabanana: Done, moved to `src/new/apple/xnu/sys/sockio.rs`.

src/new/apple/xnu/sys/sockio.rs
```diff
@@ -0,0 +1,77 @@
...
+pub const SIOCSIFASYNCMAP: c_ulong = 0x8020697d;
+pub const SIOCGIFMAC: c_ulong = 0xc0206982;
+pub const SIOCSIFMAC: c_ulong = 0x80206983;
+pub const SIOCSIFKPI: c_ulong = 0x80206986;
+pub const SIOCGIFKPI: c_ulong = 0xc0206987;
+pub const SIOCGIFWAKEFLAGS: c_ulong = 0xc0206988;
+pub const SIOCGIFFUNCTIONALTYPE: c_ulong = 0xc02069ad;
+pub const SIOCSIF6LOWPAN: c_ulong = 0x802069c4;
+pub const SIOCGIF6LOWPAN: c_ulong = 0xc02069c5;
+pub const SIOCGIFDIRECTLINK: c_ulong = 0xc02069de;
```
tgross35: Sorry but just realized these are ioctl constants, we should avoid hardcoding them since they differ across platforms. I added the `_IO*` functions just now to https://github.com/rust-lang/libc/blob/f24764dc460a26c90b2d967fcffb7b341f8393f3/src/new/apple/xnu/sys/ioccom.rs, would you be able to use those?

Please squash when done then I think this should be all set.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5263/5848d319e53719d637781eae72de169d8f7867a0..459d0a9023987f3231a492b9119a8b96e9eb239e)*

tgross35 (changes requested):

src/new/apple/xnu/sys/sockio.rs
yabanana: Rebased on main and switched to the `_IO*` functions. For the structs that aren't in libc yet (`ifaliasreq`, `ifmediareq`, ...) I went with raw `_IOC` and the size, same as `TIOCPTYGNAME` in ttycom.rs. Squashed into a single commit.

tgross35 (approved):
Looks great, thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5263/f24764dc460a26c90b2d967fcffb7b341f8393f3..d186129f76638410c041d1dcac80a1c45227697b)*

Issue #5262, opened 2026-07-08
CI: Replace deprecated Android emulator with Cuttlefish


10ne1:
Hello,

### Problem description

As PR https://github.com/rust-lang/libc/pull/4459 reveals, the Android CI is using infrastructure and images which have been frozen and deprecated for about a decade, from the Android 7-9 eras (2016-2018), while the crate is used against current Android releases, so regressions against modern bionic are invisible to CI.

Furthermore, the deprecated emulator is slow and flaky (e.g. crashes on simple changes like https://github.com/rust-lang/libc/pull/4459 ), pinned around [a bug](https://github.com/rust-lang/libc/blob/main/ci/android-install-sdk.sh#L60) and in general not worth fixing.

### Solution

The modern mechanism for headless Android testing is [Cuttlefish](https://source.android.com/docs/devices/cuttlefish/get-started), which is what AOSP's own CI also runs on and has been adopted by other important FOSS projects like Mesa CI (see their [cuttlefish-runner script](https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/.gitlab-ci/cuttlefish-runner.sh)).

So this issue is to track the migration of the CI tests from the deprecated emulator to Cuttlefish.

It's worth noting that Cuttlefish boots a complete current Android system (kernel + userspace), unlike for example the current x86_64 job which just extracts bionic from a 2016 image and runs the test binaries directly on the Ubuntu host kernel.

I already have two prototypes which pass the x86_64 and aarch64 (fully green jobs with 6700/6480 tests in about 7 minutes runtime), heavily inspired from Mesa CI, but I'm opening an issue first for discussion and to track the wider effort, because this will likely require 4 or more PRs.

#### Coverage shift

If we just replace the deprecated emulator with cuttlefish, then the tests move from the Android 7/9-era bionic to the latest release, which means the oldest-supported-API end is no longer covered by CI.

Thankfully cuttlefish images exist for older branches as old as Android 9 / API 28, so maybe we could have runners for both the old and new.

#### Cuttlefish requires KVM

On **x86_64**, GitHub runners already expose `/dev/kvm`, so we can boot and run a full Android virtual device on the CI host, very similar to how Mesa CI does it.

[...]

PR #4459, opened 2025-05-23
Add `mallopt` and related constants for Android

chriswailes:
# Description

This change adds constants that can be passed to BIONIC's implementation of `mallopt`

# Sources

Constant definitions in BIONIC source: https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/malloc.h;l=215

# Checklist

- [X] Relevant tests in `libc-test/semver` have been updated
- [X] No placeholder or unstable values like `*LAST` or `*MAX` are
  included
- [X] Tested locally (`cd libc-test && cargo test --target mytarget`);
  especially relevant for platforms that may not be checked in CI

PR #5269, opened 2026-07-09
ci: Run x86_64-linux-android tests on Cuttlefish virtual devices

10ne1:
# Description

Part of https://github.com/rust-lang/libc/issues/5262 (please see it for background and discussion).

This migrates `x86_64-linux-android` testing off the deprecated 2016-era infrastructure and onto the modern Cuttlefish-based images and contains just one commit for the migration itself.

Further PRs will be posted for the Android toolchain version ask by Trevor as well as to migrate the aarch64 CI.

#### Some observations from my testing and validation:

- I pinned the jobs to the released Android 17 official image so new releases don't break CI. It's getting tested with the existing API 28 NDK.
- Whole job takes ~7.5 minutes (the guest boots in ~1 minute under KVM).
- The full tier2 matrix still passes (the remaining Android targets stay on the legacy SDK emulator for now).
- I also validated that rust-lang/libc#4459, which crashes the old emulator, passes. We will also need the aarch64 migration to fully unblock it (I'm working on it).
- SELinux is left enforcing in the guest, unlike Mesa CI, which needs to swap GPU drivers inside the guest, so we don't need `-guest_enforce_security=false`. This means we keep the env as close to real devices as possible.

#### One implementation decision is worth highlighting (or changing):

I had to pass `-enable_sandbox=false`, which disables crosvm's minijail device sandbox, because Ubuntu >= 24.04 uses AppArmor to block the unprivileged user namespaces minijail requires. I opted to disable it because the sandbox value in this environment is basically null, it's a defense-in-depth against a hostile guest, but CI runners are ephemeral, there are no untrusted Android apps and the test code already runs unsandboxed on the host in every other non-Cuttlefish/crosvm job.

MesaCI doesn't have this problem because it's running Debian images which are not as restrictive. AFAIK only Ubuntu is available on the GitHub runners, not Debian. In effect, I mirrored the Debian runtime behavior.

If reviewers/maintainers wish, I can try two alternatives (feel free to suggest others):
1. Set `kernel.apparmor_restrict_unprivileged_userns=0`. This allows unprivileged userns host-wide so minijail works, however it weakens the whole host runner for a single process.
2. Grant crosvm a userns exception via an AppArmor profile. I haven't actually tried this yet, but it should be the proper Ubuntu way to fix it, though it adds another moving part and an additional maintenance burden to keep the profile working on newer Ubuntus.

# Sources

No library/API changes; CI-only. Infrastructure references:

- Cuttlefish: https://source.android.com/docs/devices/cuttlefish/get-started
- Host packages: https://github.com/google/android-cuttlefish
- Guest builds: https://ci.android.com/builds/branches/aosp-android-latest-release/grid
- Mesa CI's runner (inspiration): https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/.gitlab-ci/cuttlefish-runner.sh

# Checklist

# AI disclosure


tgross35:
Thank you for getting this going! I know extremely little about android CI so pinging the maintainers for help: @chriswailes @jfgoog @maurer @pirama-arumuga-nainar

.github/workflows/ci.yaml
```diff
@@ -268,13 +271,27 @@ jobs:
             jq -r 'to_entries | map("\(.key)=\(.value|tostring)") | .[]' >>$GITHUB_ENV
         shell: bash
 
+      - name: Set up Cuttlefish
+        if: runner.os == 'Linux' && matrix.env.TEST_CUTTLEFISH
+        run: ./ci/cuttlefish-setup.sh
+
```
tgross35: Can cuttlefish be run from inside docker? That would be rather nice for running locally

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5269/6a08ed3383ef6b1d2c88dc6e17f70514cdec1eef..9c7959ccd86e244cce09e95a5c286351b17d38b9)*

.github/workflows/ci.yaml
```diff
@@ -268,13 +271,27 @@ jobs:
...
 
+      - name: Collect Cuttlefish logs
+        if: always() && matrix.env.TEST_CUTTLEFISH
+        run: ./ci/cuttlefish-logs.sh cuttlefish-logs
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        if: always() && matrix.env.TEST_CUTTLEFISH
+        with:
+          name: cuttlefish-logs-${{ matrix.target }}
+          path: cuttlefish-logs
+          retention-days: 5
```
tgross35: Would it be possible to only stop cuttlefish here, then add the relevant globs to `ci/create-artifacts.py`? So we don't need another script+job

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5269/6a08ed3383ef6b1d2c88dc6e17f70514cdec1eef..9c7959ccd86e244cce09e95a5c286351b17d38b9)*

ci/cuttlefish-setup.sh
```diff
@@ -0,0 +1,67 @@
...
+    # reboot the upstream README asks for.
+    sudo usermod -aG kvm,cvdnetwork,render "$USER"
+    sudo udevadm control --reload-rules
+    sudo udevadm trigger
+    sudo modprobe vhost_net vhost_vsock
+    sudo systemctl restart cuttlefish-host-resources
+
+    # cvd refuses to run unless the current process is in the groups added
+    # above, so re-exec through sudo to pick them up without a relogin.
+    exec sudo -E -u "$USER" "$0" launch
```
tgross35: Can the `sudo` be dropped from here and instead invoke `cuttlefish-setup.sh` with `sudo` in the CI job? There's a lot to go wrong here, I don't want somebody to accidentally run this locally if their `sudo` doesn't ask for a password.

Of course, if it's possible to run in docker then this becomes a nonissue.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5269/6a08ed3383ef6b1d2c88dc6e17f70514cdec1eef..9c7959ccd86e244cce09e95a5c286351b17d38b9)*

.github/workflows/ci.yaml
```diff
@@ -187,7 +187,9 @@ jobs:
           retention-days: 5
 
   test_tier2:
-    name: Test tier2
+    # Explicit name (rather than letting GitHub join all matrix values, which
+    # includes every `env` value) so entries with long env blocks stay short.
+    name: Test tier2 (${{ matrix.target }}${{ matrix.artifact-tag && format(', {0}', matrix.artifact-tag) }})
```
tgross35: "ci: Name tier2 jobs explicitly from target and artifact-tag" you can drop this commit, the longish names are fine for pretty much everything else. You can also just set `env: { ANDROID_VERSION: android9 }` then pick that up in scripts to set the relevant cuttlefish env, which is probably better anyway because that's easier to run locally.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5269/6a08ed3383ef6b1d2c88dc6e17f70514cdec1eef..9c7959ccd86e244cce09e95a5c286351b17d38b9)*

.github/workflows/ci.yaml
```diff
@@ -238,6 +238,12 @@ jobs:
             env: { TEST_CUTTLEFISH: 1 }
             # Keep in sync with the Android build pinned in ci/cuttlefish-setup.sh
             artifact-tag: android17
+          - target: x86_64-linux-android
+            # Android 9 (API 28) image: the oldest Android release providing
+            # Cuttlefish images, and the API floor the NDK toolchain targets
+            # (x86_64-linux-android28-clang)
+            env: { TEST_CUTTLEFISH: 1, CUTTLEFISH_BUILD: 7799594, CUTTLEFISH_TARGET: aosp_cf_x86_64_phone-userdebug }
+            artifact-tag: android9
```
tgross35: How is it that we aren't hitting any CI failures from this old version with API that didn't exist at the time? That's kind of surprising

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5269/6a08ed3383ef6b1d2c88dc6e17f70514cdec1eef..9c7959ccd86e244cce09e95a5c286351b17d38b9)*

.github/workflows/ci.yaml
10ne1: I think so, if we passthrough /dev/kvm, /dev/vhost-net & co, though I haven't tried it yet.

I'll work on this, all the other requests are pretty easy / small changes in comparison. Thanks!

.github/workflows/ci.yaml
10ne1: Yes, this should be pretty easy to do. Thanks!

.github/workflows/ci.yaml
tgross35: If it's hacky like that then I don't think it's worth it, I was just hoping it might be possible to do something self-contained

.github/workflows/ci.yaml
10ne1: The NDK toolchain the CI is already using is capped at API 28, so only what's present in API 28 gets tested on both Android 9 and 17. Nothing newer than API 28 gets built into the test binaries, so symbol resolution works (nothing missing).

We could, of course, start testing APIs which are newer than 28 (latest is API 37 IIRC in Android 17), however that is a test coverage extension or shift beyond the Cuttlefish migration.

I'm also trying to keep this migration as small as possible without affecting test coverage too much. :smile: 

(We could even drop the android 9 image entirely and just test the latest 17 with multiple toolchain APIs... TBH I have no strong opinions here). 

I am willing to follow guidance for all these variants / options / possibilities.

tgross35: I actually have no idea how this all works. Is API version independent of Android version (seems like it) and the API version defines which symbols are available regardless of the Android version? How does that work with things like kernel headers?

10ne1: > Is API version independent of Android version (seems like it)

Yes and no :upside_down_face: Hope I manage to explain this clearly:

Each new Android release brings with it a new API level, for example Android 9 is API 28, Android 10 is API 29 and so on. [Here is a table with them all](https://apilevels.com/).

When you build with the NDK, you target a "floor" API at build time. Targeting API 28 makes it work with Android 9 and later. That is why our CI passes with both Android 9 and 17.

> and the API version defines which symbols are available regardless of the Android version?

Yes, that's exactly how it works at build time. At run-time, the device's bionic decides and it works because bionic is append-only (symbols don't get removed), so a binary built against API 28 finds everything it references on any device from Android 9 and up.

> How does that work with things like kernel headers?

The NDK ships a single snapshot of "sanatized" Linux UAPI headers (from the [bionic kernel uapi](https://android.googlesource.com/platform/bionic/+/main/libc/kernel/README.md)) and the kernel UAPI is also append only which guarantees compatibility. The UAPI headers are not gated by the NDK API level, the same headers are used for all levels within an NDK.

Hope this all makes sense, I try to not over-complicate too much.

(think of the API levels as a frozen and explicit version of what glibc does implicit via its symbol versioning with the difference that glibc can also ship new versions of an existing name).

tgross35: So kernel API is tied to the NDK version, which is independent of the API level / version you're actually targeting, but `libc` API is tied to the API level? I think that makes more sense.

Is there any point in testing with the old API level? As I understand it, that probably just means we'll have to conditionally skip more things in tests if people want to add API that's only present since after API 28.

(If there's a way to detect targeted Android version from headers, it would be great if you could add that to https://github.com/rust-lang/libc/blob/70c67fe2ec83eda97646b49cab2bcbd1603a65ea/libc-test/build.rs#L6370-L6413 to make it easier to do those conditional skips (separate from all this of course))

.github/workflows/ci.yaml
10ne1: This was actually simpler than I expected :tada: 

Google's apt repo provides `cuttlefish-base` and `cuttlefish-user` packages which provide services like `cuttlefish-host-resources` which helps automate much of the boilerplate and we can pass flags like `--device /dev/kvm --device /dev/vhost-vsock --cap-add NET_ADMIN` exactly like `google/android-cuttlefish`'s GitHub actions to our own `ci/run-docker.sh` script.

While testing this (the containerized flow passes the env explicitely via flags), I also uncovered a very embarrassing bug: :sweat: 

Ubuntu ships `sudo-rs` which doesn't preserve the environment with the `-E` flag (it issues just a warning when it drops the env! :angry: ), so on on the android 9 job it silently defaulted to android 17.

Maybe this is a sign that we should only use an android 17 image to test multiple API levels (eg oldest API 28 and latest 37) for each architecture and call it a day, don't bother with multiple image releases.

ci/cuttlefish-setup.sh
10ne1: Yes, it's trivial to move the sudo up, however the docker route seems the way to go, after digging deeper into it (see my above comment on your containerization suggestion).

.github/workflows/ci.yaml
10ne1: > So kernel API is tied to the NDK version, which is independent of the API level / version you're actually targeting, but libc API is tied to the API level? I think that makes more sense.

Yes, that's exactly right, with just a very minor tangential observation (sorry can't help myself from going into these details, been working with toolchains for a long time, backporting kernel headers and such for a lot of systems :sweat_smile: ):

The kernel UAPI has nothing to gate behind the NDK version because there's no linking at the kernel level (it's just syscalls), so when a new constant gets passed to an older kernel, it just issues an EINVAL or ENOSYS if the syscall doesn't exist, then the libc falls back to another syscall or maybe seccomp intervenes causing a SIGSYS, depending how the system is configured.

So I wouldn't say that the kernel API is **tied** to the NDK version it's more like each NDK ships whatever happened to be "current" at the time, with nothing gating or versioning it.

> Is there any point in testing with the old API level? As I understand it, that probably just means we'll have to conditionally skip more things in tests if people want to add API that's only present since after API 28.

Excellent point and yes, I think there's value in testing at least an "old" and "new" API (similar to how the musl CI also does it btw), because most users typically build with a minimum SDK version which is typically quite far below the latest release.

You're also correct on the cost, staying on the current API 28 does require conditional skips in libc-test on every post-28 addition. They [already exist](https://github.com/rust-lang/libc/blob/main/libc-test/build.rs#L2432). :smile:

Quoting from the [Rust Android platform doc](https://doc.rust-lang.org/rustc/platform-support/android.html):

> Rust will support the most recent Long Term Support (LTS) Android Native Development Kit (NDK). By default Rust will support all API levels supported by the NDK, but a higher minimum API level may be required if deemed necessary.

Acording to the [NDK compatibility page](https://github.com/android/ndk/wiki/Compatibility), API levels starting from 20-21 onwards are supported, which are even older than our current 28 :sweat_smile: 

To cut the story short, I would mirror what we do with musl new/old: keep 28 (since we have it...) and just add a new / latest in a separate PR, after the cuttlefish. 

Also I'd keep both tested against the latest image which should support all APIs, to simplify things.

> If there's a way to detect targeted Android version from headers...

Yes, I'm pretty certain there is, will do this as well. Thanks!

10ne1:
@tgross35 I addressed all your requests, so I kindly ask for a re-review (it's only 1 commit now).

I'll address the detecting targeted Android version from headers in a separate PR, since by doing that I uncovered another, pre-existing, bug.

.github/workflows/ci.yaml
maurer: Everything seems fine, but a couple extra details to make things clear in the future:

In C/C++, a built library technically has two API levels:
* Minimum SDK level - this is what has been treated with above, and is how you select which toolchain entry points you're using (which really just sets a `--target` flag under the hood, but used to do more).
* Target/Max SDK level - this doesn't directly appear in C/C++ builds, but kind of exists implicitly. In C/C++, you can dynamically query the SDK level, and based on it, choose whether to dynamically access a symbol. Rust does not currently support this. When a native library is packaged as part of an App (the usual deployment if you aren't part of the OS), the target SDK level also determines what policies the app must follow - if the OS gets stricter about things, your target SDK level indicates what policies you must have adapted to.

Re: Kernel versions - the UAPI headers are shipped with the NDK with our latest kernel as of the time the NDK is shipped. However, Android devices may be running *substantially* older kernels. The [current supported kernels](https://source.android.com/docs/core/architecture/kernel/android-common#compatibility-matrix) probably give you an idea, but you'd need to join this with the mapping between OS release numbers and SDK version numbers to know the minimum kernel version available.

Since we're overall targeting 21, it's so old it's not on that table - that has a minimum kernel version of 3.4

Obviously apps are free to try to call things and see what's there, but they may get ENOSYS or a similar error code.

maurer:
Overall LGTM from Android PoV. We may want to port this logic to the main Rust repo at some point to test newer API calls in libstd.

maurer (approved):

.github/workflows/ci.yaml
```diff
@@ -236,6 +236,9 @@ jobs:
           - target: x86_64-apple-darwin
             os: macos-26-intel
           - target: x86_64-linux-android
+            env: { TEST_CUTTLEFISH: 1 }
```
tgross35, line 239: In one of the followups, could you drop this and instead match on the target in `run-docker.sh`? Keeps our CI setup cleaner and means there's less magic env to set if running locally.

tgross35 (approved):
LGTM as well, let's get this in

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5269/70c67fe2ec83eda97646b49cab2bcbd1603a65ea..a621872e3e7798eb450c155451d2c6a32e5afb00)*

.github/workflows/ci.yaml
10ne1: Yes, I can do that.

10ne1: Done: https://github.com/rust-lang/libc/pull/5316

PR #5270, opened 2026-07-10
emscripten: add pthread_sigmask, sigwait, faccessat, pthread_kill, se…

guybedford:
Adds remaining gated `wasm32-unknown-emscripten` APIs which are implemented by Emscripten - `pthread_sigmask`, `sigwait`, `sigwaitinfo`, `faccessat`, `pthread_kill`.

Link: https://github.com/emscripten-core/emscripten/blob/6a93f7db16cd8e5214b4c3c08a361ccab9ffc67e/system/lib/libc/musl/include/signal.h#L225-L227
Link: https://github.com/emscripten-core/emscripten/blob/6a93f7db16cd8e5214b4c3c08a361ccab9ffc67e/system/lib/libc/musl/include/signal.h#L230-L231
Link: https://github.com/emscripten-core/emscripten/blob/6a93f7db16cd8e5214b4c3c08a361ccab9ffc67e/system/lib/libc/musl/include/unistd.h#L88

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -3227,7 +3227,6 @@ fn test_freebsd(target: &str) {
 
 fn test_emscripten(target: &str) {
     assert!(target.contains("emscripten"));
-    #[expect(unused_variables)] // remove once we need a version check
     let emscripten = VERSIONS.emscripten.unwrap();
 
     let mut cfg = ctest_cfg();
@@ -3394,6 +3393,9 @@ fn test_emscripten(target: &str) {
             // Emscripten does not support fork/exec/wait or any kind of multi-process support
             // https://github.com/emscripten-core/emscripten/blob/3.1.68/tools/system_libs.py#L1100
             "execv" | "execve" | "execvp" | "execvpe" | "fexecve" | "wait4" => true,
+
+            // Emscripten's `pthread_kill` used to only be linkable when building with `-pthread`
+            "pthread_kill" if emscripten < (6, 0) => true,
 
             _ => false,
         }
--- a/src/unix/linux_like/emscripten/mod.rs
+++ b/src/unix/linux_like/emscripten/mod.rs
@@ -1458,6 +1458,14 @@ extern "C" {
         buflen: size_t,
         result: *mut *mut crate::group,
     ) -> c_int;
+    pub fn sigwait(set: *const crate::sigset_t, sig: *mut c_int) -> c_int;
+    pub fn sigwaitinfo(set: *const crate::sigset_t, info: *mut crate::siginfo_t) -> c_int;
+    pub fn sigtimedwait(
+        set: *const crate::sigset_t,
+        info: *mut crate::siginfo_t,
+        timeout: *const crate::timespec,
+    ) -> c_int;
+    pub fn faccessat(dirfd: c_int, pathname: *const c_char, mode: c_int, flags: c_int) -> c_int;
 }
 
 // Alias <foo> to <foo>64 to mimic glibc's LFS64 support
--- a/src/new/common/posix/pthread.rs
+++ b/src/new/common/posix/pthread.rs
@@ -187,7 +187,12 @@ extern "C" {
 
     // FIXME(reorg): In recent POSIX versions, this is a signal.h function and not required
     // in pthread.
-    #[cfg(any(target_os = "android", target_os = "l4re", target_os = "linux"))]
+    #[cfg(any(
+        target_os = "android",
+        target_os = "emscripten",
+        target_os = "l4re",
+        target_os = "linux"
+    ))]
     pub fn pthread_kill(thread: crate::pthread_t, sig: c_int) -> c_int;
 
     #[cfg(all(target_os = "linux", not(target_env = "ohos")))]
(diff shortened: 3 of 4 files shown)


src/unix/linux_like/emscripten/mod.rs
```diff
@@ -1458,6 +1458,20 @@ extern "C" {
...
+    ) -> c_int;
+    pub fn sigwait(set: *const crate::sigset_t, sig: *mut c_int) -> c_int;
+    pub fn sigwaitinfo(set: *const crate::sigset_t, info: *mut crate::siginfo_t) -> c_int;
+    pub fn sigtimedwait(
+        set: *const crate::sigset_t,
+        info: *mut crate::siginfo_t,
+        timeout: *const crate::timespec,
+    ) -> c_int;
+    pub fn faccessat(dirfd: c_int, pathname: *const c_char, mode: c_int, flags: c_int) -> c_int;
+    pub fn pthread_kill(thread: crate::pthread_t, sig: c_int) -> c_int;
```
tgross35: For the two pthread functions, could you reexport in `src/new/emscripten/pthread.rs` instead?

Please add permalinks to headers in the commit message

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5270/6a08ed3383ef6b1d2c88dc6e17f70514cdec1eef..ce12efbc2aa00e47cd097190f09076bafef55b3c)*

guybedford: Moved, and added the links.

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5270/6a08ed3383ef6b1d2c88dc6e17f70514cdec1eef..bf4b9fe6e6d8d06320db7d71245ccc5c8b12207b)*

libc-test/build.rs
```diff
@@ -3390,6 +3390,10 @@ fn test_emscripten(target: &str) {
             // https://github.com/emscripten-core/emscripten/blob/3.1.68/tools/system_libs.py#L1100
             "execv" | "execve" | "execvp" | "execvpe" | "fexecve" | "wait4" => true,
 
+            // `pthread_kill` is only linkable when building with `-pthread`; it lives in
+            // emscripten's pthread library rather than the base libc.
+            "pthread_kill" => true,
+
```
tgross35, line 3399: Could we just pass `cfg.flag("-lpthread")` for testing purposes?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5270/6a08ed3383ef6b1d2c88dc6e17f70514cdec1eef..9bbac393c442f8e15981deb9378ea1208c7c8370)*

guybedford: I think it would need to be something more like `RUSTFLAGS="-Ctarget-feature=+atomics,+bulk-memory -Clink-arg=-pthread" + -Zbuild-std`.

tgross35: So `pthread_create` is fine but `pthread_kill` requires atomics? That seems weird. I'm fine to just keep the skip if that's accurate, but I'd like to understand why this is the case first.

guybedford: Yes, that's correct, I can confirm I've verified all the other `pthread_*` are fine as they are either included in the base musl libc or populated by `library_pthread_stub.c`. In the latest version of Emscripten the skip isn't needed either - this CI is locked to 3.1.68 though, so we're working against that.

tgross35: In that case could you change this to `"pthread_kill" if emscripten < (6, 0) => true` so we start testing it if run with a newer version? I landed https://github.com/rust-lang/libc/commit/70c67fe2ec83eda97646b49cab2bcbd1603a65ea just recently for this.

The version comes from https://doc.rust-lang.org/rustc/platform-support/wasm32-unknown-emscripten.html#requirements for reference, I'm not sure if there might be an advantage to bumping in the near future (cc @hoodmane)

guybedford: Nice, sure I've added that. Bumping the target Emscripten version at some point to a more recent sensible baseline sounds like a good idea.

hoodmane: Yeah Emscripten 3.1.68 is quite old.

libc-test/build.rs
```diff
@@ -3395,6 +3393,9 @@ fn test_emscripten(target: &str) {
             // https://github.com/emscripten-core/emscripten/blob/3.1.68/tools/system_libs.py#L1100
             "execv" | "execve" | "execvp" | "execvpe" | "fexecve" | "wait4" => true,
 
+            // Emscripten's `pthread_kill` used to only be linkable when building with `-pthread`
+            "pthread_kill" if matches!(VERSIONS.emscripten, Some(v) if v < (6, 0)) => true,
+
```
tgross35, line 3399: Sorry, just getting back to reviews. Why doesn't this just use `emscripten` as it existed? The `matches!` is more complicated.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5270/70c67fe2ec83eda97646b49cab2bcbd1603a65ea..1265e1e46241e8f81066bc9e73338b4f4b7c8d3c)*

guybedford: Ah, this predates the version-check commit landing on main - now that it's in the base, switched back to the plain `emscripten < (6, 0)` form.

tgross35 (approved):
Looks great, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5270/bbd92536455546ae4cdb7959647c02d08c8fac1c..c03f0f35b5860349fbcc3a67cb01b84e1a079b98)*

PR #5274, opened 2026-07-11
solarish: add CLOCK_PROCESS_CPUTIME_ID and CLOCK_THREAD_CPUTIME_ID

Toasterson:
# Description
This does two things described below for reference if somebody needs to verify those. 

illumos [usr/src/uts/common/sys/time_impl.h](https://github.com/illumos/illumos-gate/blob/master/usr/src/uts/common/sys/time_impl.h#L119-L128) defines CLOCK_THREAD_CPUTIME_ID=2 (alias CLOCK_PROF) and
CLOCK_PROCESS_CPUTIME_ID=5, but libc never bound them for solarish, so any crate that references these clock ids (e.g. the cpu-time crate) fails to build on illumos/solaris. And since I needed that I implemented them :)

Also drops the stale comment claiming clock_gettime(3c) "doesn't seem to accept anything other than CLOCK_REALTIME". This got fixed with illumos-gate commit [dea9f5e6a49](https://github.com/illumos/illumos-gate/commit/dea9f5e6a49) (2021-10-16, ["14126 clock_gettime() could work with thread/proc clocks"](https://www.illumos.org/issues/14126)), which taught clock_gettime(3c) to honor the thread/process CPU-time clocks; both Oracle Solaris 11.4 and current illumos document and accept them.

# Sources
- https://github.com/illumos/illumos-gate/blob/master/usr/src/uts/common/sys/time_impl.h#L119-L128
- https://github.com/illumos/illumos-gate/commit/dea9f5e6a49
- https://www.illumos.org/issues/14126

# Checklist


tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5274/70c67fe2ec83eda97646b49cab2bcbd1603a65ea..ce23171bfae282c75a47443b76b804369aa09f22)*

PR #5275, opened 2026-07-11
l4re: update definitions and test infra

farao:
# Description

The definitions in l4re have changed, mostly due to uclibc being updated and uclibc config having changed. Also, this makes the test infrastructure work for l4re again.

# Sources

* L4Re: https://github.com/l4re/

# Checklist


JohnTitor (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5275/70c67fe2ec83eda97646b49cab2bcbd1603a65ea..c53e5fec966d69592b5975bf5767063903de36d8)*

farao:
Nice, thanks for the quick review and merge! Are there by any chance plans to make a new release soon?

tgross35:
I'm planning on a new release in the next couple of days

libc-test/build.rs
```diff
@@ -6381,10 +6385,12 @@ struct Versions {
 impl Versions {
     fn init_from_cc() -> Self {
         let src = r#"
-            #ifdef __linux__
+            #if defined(__linux__) && defined(__has_include)
+            #if __has_include(<linux/version.h>)
```
tgross35, line 6389: Does l4re define `__linux__`? Or what's the reason for this change?

farao: It's a bit more complicated. We use the linux gcc for compiling for l4re (which means we configure/parametrize it to link to our own libs, to use the l4re loader etc.). Therefore, `__linux__` is set in this case because it's a linux-gcc but we don't use it as such.

tgross35: Thanks for the clarification. Is there a definition like `__l4re__` that winds up set? And/or a way to check the l4re version?

PR #5277, opened 2026-07-12
chore: remove some clippy lint `allow`s

ada4a:
# Description

The FIXME says these lints should be fixed, but it looks like they already are, as no warnings appeared when I removed the `allow`s. So I removed them for good.

# Checklist

This change doesn't cause any warnings on the `libc-0.2` branch either, so:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -193,10 +193,7 @@ ptr_as_ptr = "warn"
 unnecessary_semicolon = "warn"
 
 # FIXME(clippy): these should be fixed if possible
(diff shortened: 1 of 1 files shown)


tgross35 (approved):
Huh, I wonder if these were target-specific or something. Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5277/6094feef0ba041cdfa35c81a14c2b319eb197771..0d9f9c50c7a532e50ce2f5be6c5dd426685ed6a0)*

PR #5280, opened 2026-07-13
CI: Detect Android API from the toolchain

10ne1:
# Description

This is a follow-up to PR https://github.com/rust-lang/libc/pull/5269 to address @tgross35 request to detect the Android API levels from the toolchain instead of hardcoding it in the tests.

The first 2 commits address @tgross35 request, while the last two fix 2 separate pre-existing uncovered issues (details in the commit messages).

# Sources

No library/API changes; this is test-harness and CI-only.

Useful references:
  - Android API levels and the `__ANDROID_API__` / `__ANDROID_MIN_SDK_VERSION__` / `__ANDROID_API_FUTURE__` macros:
    https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/include/android/api-level.h
  - NDK per-API-level Clang wrapper naming convention (`<triple><api>-clang`):
    https://developer.android.com/ndk/guides/other_build_systems
  - cc crate's own handling of NDK wrapper names (`get_compiler()` skips adding its own --target when it recognizes this naming convention, pinned at the version this repo's `Cargo.lock` locks to):
    https://github.com/rust-lang/cc-rs/blob/cc-v1.2.66/src/lib.rs#L4401

# Checklist

# AI disclosure

### DIFF
--- a/ci/docker/aarch64-linux-android/Dockerfile
+++ b/ci/docker/aarch64-linux-android/Dockerfile
@@ -18,18 +18,22 @@ WORKDIR /android/
 COPY android* /android/
 
 ENV ANDROID_ARCH=aarch64
+# Single source of truth for the API level: it selects the emulator system
+# image (passed to android-install-sdk.sh) and names the CC wrapper below, so
+# the toolchain can never drift from the image the tests run on.
+ARG ANDROID_API=24
 ENV PATH=$PATH:/android/linux-x86_64/bin:/android/sdk/cmdline-tools/tools:/android/sdk/platform-tools
 
 RUN /android/android-install-ndk.sh
-RUN /android/android-install-sdk.sh $ANDROID_ARCH
+RUN /android/android-install-sdk.sh $ANDROID_ARCH $ANDROID_API
 RUN mv /root/.android /tmp
 RUN chmod 777 -R /tmp/.android
 RUN chmod 755 /android/sdk/cmdline-tools/tools/* /android/sdk/emulator/qemu/linux-x86_64/*
 
 ENV PATH=$PATH:/rust/bin \
-    CARGO_TARGET_AARCH64_LINUX_ANDROID_LINKER=aarch64-linux-android28-clang \
+    CARGO_TARGET_AARCH64_LINUX_ANDROID_LINKER=aarch64-linux-android${ANDROID_API}-clang \
     CARGO_TARGET_AARCH64_LINUX_ANDROID_RUNNER=/tmp/runtest \
-    CC_aarch64_linux_android=aarch64-linux-android28-clang \
+    CC_aarch64_linux_android=aarch64-linux-android${ANDROID_API}-clang \
     AR_aarch64_linux_android=llvm-ar \
     HOME=/tmp
 
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -2060,6 +2060,10 @@ fn test_android(target: &str) {
     };
     let x86 = target.contains("i686") || target.contains("x86_64");
     let aarch64 = target.contains("aarch64");
+    // The API level the NDK toolchain targets, which caps the available libc
+    // API surface (see `Versions`). Detection is required, so unwrap and let a
+    // toolchain we can't read fail loudly rather than silently skip every test.
+    let android = VERSIONS.android.unwrap();
 
     let mut cfg = ctest_cfg();
     cfg.define("_GNU_SOURCE", None);
@@ -2251,8 +2255,7 @@ fn test_android(target: &str) {
             "posix_spawn_file_actions_t" => true,
             "posix_spawnattr_t" => true,
 
-            // Added in API level 24
-            "if_nameindex" => true,
+            "if_nameindex" if android < 24 => true,
 
             _ => false,
         }
@@ -2450,8 +2453,11 @@ fn test_android(target: &str) {
             "reallocarray" => true,
             "__system_property_wait" => true,
 
-            // Added in API level 30, but tests use level 28.
-            "memfd_create" | "mlock2" | "renameat2" | "statx" | "statx_timestamp" => true,
+            "memfd_create" | "mlock2" | "renameat2" | "statx" | "statx_timestamp"
+                if android < 30 =>
+            {
+                true
+            }
 
             // Added in API level 33, but tests use level 28.
             "preadv2" | "pwritev2" => true,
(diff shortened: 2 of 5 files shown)


libc-test/build.rs
```diff
@@ -6380,17 +6376,19 @@ struct Versions {
     netbsd: Option<(u32, u32)>,
     macos: Option<(u32, u32)>,
     emscripten: Option<(u32, u32)>,
+    /// Android API level. Unlike the (major, minor) platform versions
+    /// above, Android versions its libc ABI with a single increasing
+    /// integer (e.g. 28). There is no minor component.
+    android: Option<u32>,
```
tgross35:
```suggestion
    /// Android API level (no minor version).
    android: Option<u32>,
```

Simplification

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5280/6094feef0ba041cdfa35c81a14c2b319eb197771..cc8308135961ef9d7394abf666654f5458047cd2)*

libc-test/build.rs
```diff
@@ -2045,6 +2045,10 @@ fn test_android(target: &str) {
     };
     let x86 = target.contains("i686") || target.contains("x86_64");
     let aarch64 = target.contains("aarch64");
+    // The API level the NDK toolchain targets, which caps the available libc
+    // API surface (see `Versions`). Items introduced later must be skipped when
+    // building below their level and also when no level could be detected.
+    let skip_below_api = |level: u32| VERSIONS.android.map_or(true, |api| api < level);
```
tgross35: Match the pattern in other places:

```rust
let android = VERSIONS.android.unwrap();
```

Or `.unwrap_or(0)` if the version is unreliable.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5280/6094feef0ba041cdfa35c81a14c2b319eb197771..cc8308135961ef9d7394abf666654f5458047cd2)*

libc-test/build.rs
```diff
@@ -2237,7 +2241,7 @@ fn test_android(target: &str) {
             "posix_spawnattr_t" => true,
 
             // Added in API level 24
-            "if_nameindex" => true,
+            "if_nameindex" => skip_below_api(24),
```
tgross35: To follow that, this would be `"if_nameindex" if android < 24 -> true`. (Applies a few places below)

You can also optionally delete all the "Added in API level" comments since the check makes it obvious.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5280/6094feef0ba041cdfa35c81a14c2b319eb197771..cc8308135961ef9d7394abf666654f5458047cd2)*

ci/docker/aarch64-linux-android/Dockerfile
tgross35, line 1: "ci: Point each Android toolchain at the API level bionic actually runs" is there any way to do this more automatically? Seems easy to get out of sync

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5280/6094feef0ba041cdfa35c81a14c2b319eb197771..cc8308135961ef9d7394abf666654f5458047cd2)*

libc-test/build.rs
```diff
@@ -6464,6 +6469,15 @@ impl Versions {
...
                 "__GLIBC_MINOR__" => ret.glibc.get_or_insert_default().1 = value.parse().unwrap(),
+                // The API level is in __ANDROID_API__ (old clang) or __ANDROID_MIN_SDK_VERSION__
+                // (modern clang), so take a number from either. When the API is undetected (e.g.
+                // the triple carries no API level), its consumers fall back to skipping every
+                // API-gated test as if the toolchain targeted the oldest level possible.
+                "__ANDROID_API__" | "__ANDROID_MIN_SDK_VERSION__" => {
+                    if let Ok(level) = value.parse() {
+                        ret.android = Some(level);
+                    }
+                }
```
tgross35: I'd prefer to make version detection failures noisy, to make sure it always works in our CI. Looks like `__ANDROID_MIN_SDK_VERSION__` appeared in Clang 12 (2021, https://clang.godbolt.org/z/Gva1e75TE) so it's probably okay to just check for that and `unwrap()`. Unless we need to account for that `__ANDROID_API_FUTURE__` definition?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5280/6094feef0ba041cdfa35c81a14c2b319eb197771..cc8308135961ef9d7394abf666654f5458047cd2)*

tgross35 (changes requested):
(status update)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5280/6094feef0ba041cdfa35c81a14c2b319eb197771..cc8308135961ef9d7394abf666654f5458047cd2)*

libc-test/build.rs
10ne1: I'll use .unwrap_or(0) just to be safe. Thanks!

libc-test/build.rs
10ne1: I don't think we need to account for it, will update.

libc-test/build.rs
tgross35: Can we reliably get the version? I actually have a preference for just `.unwrap()` so it's noisy if version detection fails, otherwise we could accidentally start skipping more than needed and have no idea.

tgross35: If `__ANDROID_MIN_SDK_VERSION__` isn't reliable, maybe `Versions::init_from_cc` can grab the env that shows up for the cuttlefish setup?

ci/docker/aarch64-linux-android/Dockerfile
10ne1: I think we can remove some of the duplication and "get out of sync" risk by having a single source of truth ANDROID_API Dockerfile env var, but I don't think there's a way to fully automate the API selection/detection.

I'll mark this resolved once I update the commits (still need to do some testing).

libc-test/build.rs
10ne1: I think it's reliable, I just have an aversion to plain `unwrap()`, as I view it as an anti-pattern in general, though in this case its use is defensible. Will finish my testing then I'll update the PR.

10ne1:
I've rebased to fix a trivial conflict with the `wasi_sdk` addition.

tgross35 (approved):
Thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5280/0f50a2e49ffb7d660127228e63292f8d67715777..069f0333c65df5e88b1f87f591738b3b059b6fec)*

PR #5281, opened 2026-07-14
ci: Run aarch64-linux-android tests with qemu-user

10ne1:
# Description

This is a draft because it depends on https://github.com/rust-lang/libc/pull/5269 and https://github.com/rust-lang/libc/pull/5280 (only the last commit is new here, the others belong to those PRs and will drop out on rebase once they merge).

Part of https://github.com/rust-lang/libc/issues/5262 (please see it for background and discussion).

This migrates `aarch64-linux-android` testing off the deprecated 2016-era emulator and contains just one commit for the migration itself. 

Unlike the x86_64 migration it does not boot a Cuttlefish device: GitHub's arm64 runners expose no `/dev/kvm` and Cuttlefish cannot run without KVM. Instead, the test binaries run directly under qemu-user against the bionic linker and libraries extracted from a current official Android build, so this is a stopgap until GitHub exposes KVM on arm64 runners, at which point aarch64 can move to full Cuttlefish too.

Further PRs will be posted to raise the targeted Android API level (which widens test coverage; see below) and to look into 32-bit support.

#### Some observations from my testing and validation:

- The job is pinned to the same released Android 17 build as the x86_64 Cuttlefish job, so new releases don't break CI (the pins reference each other and get bumped together).
- The toolchain deliberately stays at API 24, matching the retired emulator's runtime, so the API-gated test skips from https://github.com/rust-lang/libc/pull/5280 resolve exactly as before: this PR changes the test infrastructure, not the coverage.
- It runs on the standard x86_64 Ubuntu runners — no arm64 runners, KVM, or emulator boot involved. The job is green with 6528 libc-test checks and takes ~4 minutes total, versus ~25 for the old emulator job (when its boot didn't hang).
- The full tier2 matrix still passes (32-bit arm stays on the legacy SDK emulator for now).
- Together with https://github.com/rust-lang/libc/pull/5269 this fully unblocks rust-lang/libc#4459, which crashes the old emulator; I validated earlier that its tests pass under this setup on both arches.

# Sources

No library/API changes; CI-only. Infrastructure references:

- Guest builds:
  https://ci.android.com/builds/branches/aosp-android-latest-release/grid
- Runtime APEX (bionic's home since Android 10): 
  https://source.android.com/docs/core/ota/apex
- qemu-user mode:
  https://www.qemu.org/docs/master/user/main.html
- Cuttlefish (the eventual destination once arm64 runners get KVM):
  https://source.android.com/docs/devices/cuttlefish/get-started

# Checklist

# AI disclosure


maurer:
This approach might work for `libc` depending on what tests we want to run, but unfortunately will be insufficient for upgrading the more general Rust test infrastructure (e.g. libstd), because selinux policies, filesystem mount patterns, etc. will not be in the right place.

For `libc` usage, the main things I can see being a potential issue:
* Dynamic symbol resolution - [linker namespaces](https://source.android.com/docs/core/architecture/vndk/linker-namespace) have config files they expect to read. This could result in untestable behavior around `dlsym` & friends.
* 16k pages are unlikely to be enabled in the kernel configuration of the system running the docker image. This means that 16k page related incompatibilities will remain untestable. This is not a regression, as the previous system runs such an old kernel it does not have these either.
* If you test anything related to `ptrace` & friends, the fact that `debuggerd` is not running may cause issues.

tgross35:
We could run on the ubuntu-arm runners. Unless, is there a better way to handle this? I assume `qemu-system` works but that's so slow, I just recently removed the last bits of it from CI.

Cc also @CrystalSplitter since you commented on the issue

tgross35:
Is it accurate that the old emulator did cross-arch emulation fine but the new one doesn't? That's kind of an interesting effect.

maurer:
The old emulator did cross-arch emulation, but doesn't do it fine - the reason we can't update any further on the main rustc infra is that there are bugs in old emulator's cross-arch, and they're not ever going to be fixed.

I know that cuttlefish can do cross-arch and its support is based on qemu (so it is very likely similar to qemu-system), but haven't looked into exactly how the invocation works and it's not the main supported flow. The reason I know it works is that you can check out an Android tree, lunch a `aosp_cf_arm64_only_phone` target, build, and then run `launch_cvd` and you'll boot it in cuttlefish, just more slowly.

10ne1:
@maurer @tgross35

So I think the consensus is to wait for arm nested virtualization / /dev/kvm support on the github runners, at which point we can match x64 testing?

I'm fine with it, just want confirmation. :smile:

tgross35:
Do the `ubuntu-26.04-arm` runners not work in this case? (Not sure whether that virtualization is needed specifically for the cross testing, or for any emulators running on those `ubuntu-26.04-arm` machines)

maurer:
@10ne1 
The two options are:
* Wait for nested virt on aarch64 (unsure how long that will take?)
* Run the aarch64 tests on an x86_64 machine through cuttlefish. While I know it's possible, I don't know the exact invocation for this.

@tgross35 
The virtualization is needed for the default way to launch cuttlefish, which assumes host arch = target arch. You might be able to run system emulators on arm and run arm code unaccelerated, but I haven't checked.

tgross35:
Are you certain that `ubuntu-26.04-arm` doesn't support the needed virtualization? https://github.com/actions/runner-images/issues/12933#issuecomment-3281653063 and https://docs.github.com/en/actions/reference/runners/larger-runners#limitations-for-macos-larger-runners (linked from that comment) both only mention Apple platforms as not supporting nested virtualization. Doesn't mean that the `-arm` platforms aren't another undocumented exception of course.

Also I came across this, any chance it could be a solution? https://github.com/reactivecircus/android-emulator-runner

10ne1:
I was not aware of the [android-emulator-runner]( https://github.com/reactivecircus/android-emulator-runner) project, thank you @tgross35 for pointing it out. I will definitely check it out and report back.

I do plan to research more into this topic overall, to come up with alternative solutions, this qemu-user is just a first draft I attempted.

Many thanks to both of you!

10ne1:
So I looked into the [android-emulator-runner](https://github.com/reactivecircus/android-emulator-runner) project and unfortunately it works only for x86, since it requires KVM.

A colleague of mine also spoke with ARM and they discontinued their efforts in this area and there is no agreement with GitHub to enable more features such as KVM or nested virtualization.

So it might take a very long time until cuttlefish gets the ability to properly run on ARM GitHub runners.

tgross35:
> So I looked into the [android-emulator-runner](https://github.com/reactivecircus/android-emulator-runner) project and unfortunately it works only for x86, since it requires KVM.

x86 hosts but can't you emulate arm? See `arm64-v8a` at https://github.com/reactivecircus/android-emulator-runner#configurations

10ne1:
@tgross35 good news, I got it to work :tada: 

The trade-off is that it's much slower than the current PR, about 21 minutes to boot Android (I guess it's comparable to the old emulator), then the tests run in about another 5 min 16 sec, versus a total of 4 m 10 s for this PR:

  - qemu-user (this PR): 4m10s, same 6528 checks, no booted system.
  - Cuttlefish qemu_cli: 33m43s, same 6528 checks, complete Android 17 with real init, selinux, linker namespaces, debuggerd. Answers every fidelity caveat @maurer raised.
  
 I'll leave this PR as is, for historical reference, and I'll open a new PR with the emulation approach.

Thank you everyone so much for your patience!

PR #5289, opened 2026-07-20
ci: Don't set `needs` when in the merge queue

tgross35:
That is just a way to save on CI for pull requests. Once something is added to the merge queue we should be reasonably confident it will complete, so we may as well start all jobs at once.


tgross35:
Looks like conditional `needs` isn't supported :(

PR #5304, opened 2026-07-21
ci: Re-enable powerpc-unknown-linux-gnu

tgross35:
This was disabled in eb7045bd7e97 ("Temporarily disable powerpc-unknown-linux-gnu tests") because of an unclear SIGILL. It seems this has been resolved, so add the platform back.


tgross35:
Looks like not yet, probably stuck until we can bump the 32bit targets' Ubuntu versions.

PR #5312, opened 2026-07-22
meta: Explicitly include items in packaged crate

bushrat011899:
## Description

Exclude `/etc` folder from `cargo package` by explicitly listing items to include. Currently published version of `libc` includes [`/etc/libc-util.py`](https://crates.io/crates/libc/0.2.187/code/etc/libc-util.py) which `cargo-deny` picks up as an executable. Added in https://github.com/rust-lang/libc/commit/27a38e3bda16c72f1bff3af9432e2282427843df 

## Checklist

- [X] ~~Relevant tests in `libc-test/semver` have been updated~~ N/A
- [X] ~~Commit messages permalink to headers for added or changed API~~ N/A
- [X] ~~Placeholder or unstable values like `*LAST` or `*MAX` have the standard doc comment~~ N/A
- [X] Tested locally (`cargo test -p libc-test --target mytarget`);
  - Tested using `cargo package` and inspecting that the Python script is excluded.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,15 @@ name = "libc"
 version = "1.0.0-alpha.4"
 keywords = ["libc", "ffi", "bindings", "operating", "system"]
 categories = ["external-ffi-bindings", "no-std", "os"]
-exclude = ["/ci/*", "/.github/*", "/triagebot.toml"]
+include = [
+    "/src",
+    "/tests",
+    "/examples",
+    "/build.rs",
+    "/CHANGELOG*",
+    "/LICENSE*",
+    "/README*",
+]
 description = "Raw FFI bindings to platform libraries like libc."
 edition = "2021"
 license = "MIT OR Apache-2.0"


tgross35:
Does it maybe make sense to switch from `exclude` to `include`? I feel like this has surprised us in the past.

tgross35 (changes requested):
I think that's the better path. Would you be able to change the `exclude` to `include`? And cover at least:

* src
* tests
* examples (doesn't exist now but could in the future. Not sure if it will reject it)
* build.rs
* CHANGELOG*
* LICENSE*
* README*

(if you could post a diff of `cargo package --list`, that would be helpful too. Otherwise I can get it)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5312/a1fcda25057b609984288ba2a280011367a493c1..a85e8f85c16b30494aeb95334264bf487aa71f2e)*

bushrat011899:
> Does it maybe make sense to switch from `exclude` to `include`? I feel like this has surprised us in the past.

Yeah I think that makes more sense. I would assume it's easier to notice (and yank) if a `libc` version is missing required files than if it inadvertently includes additional ones. 

> I think that's the better path. Would you be able to change the `exclude` to `include`? And cover at least:
> 
>     * src
>     * tests
>     * examples (doesn't exist now but could in the future. Not sure if it will reject it)
>     * build.rs
>     * CHANGELOG*
>     * LICENSE*
>     * README*
> 
> (if you could post a diff of `cargo package --list`, that would be helpful too. Otherwise I can get it)

The non-existent `examples` folder isn't an issue it seems so no problems there. Comparing before and after this commit using `diff` I get:

```diff
2,6d1
< .editorconfig
< .git-blame-ignore-revs
< .gitignore
< .release-plz.toml
< .rustfmt.toml
8d2
< CONTRIBUTING.md
16,17d9
< etc/libc-util.py
< rust-bors.toml
```

I think all of those make sense to exclude, but happy to amend!

tgross35 (approved):
Looks great, thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5312/5db5914d7ecf8d3fe32394d315f29012039630ca..972f69634e590ab86b869b032eba86ca7519e318)*

PR #5315, opened 2026-07-22
linux(uclibc): add `cfg` for LFS bindings

dybucc:
# Description

This PR adds a new `cfg` for file offset types in uClibc to ensure target
triples with a 32-bit machine word size can also have 64-bit unsuffixed file
offset types. None of these changes affect 64-bit platforms.

The LFS types continue being exposed by default in all platforms, but when the
`cfg` option is issued, they will have equivalent definitions to the unsuffixed
types.

Would like to ask @operutka, @lancethepants, @skrap and @japaric whether you
think it's worth it to add this `cfg`, you folks being or having been
maintainers for targets using uClibc. It was originally part of rust-lang/libc#5165, but got
split because our need to customize rust-lang/libc to fit upstream build options
wasn't entirely clear.

## Sources

- Upstream uclibc-ng sources showing `ino_t` defined in terms of its 64-bit
  variant, and the definition for the 64-bit variant. There's a "chain" of types
  that leads to the final type with which `ino64_t` is defined. This chain is
  made most obvious by the final type alias, which appends a `_TYPE` and
  prepends two underscores to the type identifier in question. Notice how the
  `__uquad_t` type is always defined as a 64-bit type, even on systems where
  that requires using a record type with a two-element array (32-bit per
  element).

  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/typesizes.h#L33>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/types.h#L55-L70>
- Upstream uclibc-ng sources showing `off_t` defined in terms of its 64-bit
  variant when LFS64 support and `__USE_FILE_OFFSET64` are enabled. Much like
  the above, there's a "chain" of types leading to the type whose bitwidth is
  made clear. The chain is shortened to the last few types.

  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/typesizes.h#L37>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/types.h#L109-L121>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/types.h#L55-L70>
- Upstream uclibc-ng sources showing `rlim_t` defined in terms of its 64-bit
  variant when LFS64 support and `__USE_FILE_OFFSET64` are enabled. This follows
  the same reasoning as the above.

  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/resource.h#L131-L138>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/types.h#L149-L150>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/typesizes.h#L39-L40>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/types.h#L107>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/types.h#L110-L122>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/types.h#L55-L70>
- Upstream uclibc-ng sources showing `blkcnt_t` defined in terms of its 64-bit
  variant. This follows the same reasoning as the above.

  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/include/sys/stat.h#L89-L96>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/types.h#L172-L173>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/typesizes.h#L41-L42>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/types.h#L109-L121>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/types.h#L55-L70>
- Upstream uclibc-ng sources showing `fsblckcnt_t` and `fsfilcnt_t` defined in
  terms of their 64-bit variants. This first of the below sources also includes
  references to the above types.

  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/include/sys/types.h#L233-L265>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/typesizes.h#L43-L44>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/typesizes.h#L45-L46>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/types.h#L110-L122>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/types.h#L55-L70>
- Upstream uclibc-ng sources showing how `flock` and `flock64` are independent,
  but when defined with `__USE_FILE_OFFSET64`, will end up having the same
  record layout. The second source shows how this definition can actually be
  unified across all child modules to the `uclibc` module because the fields
  that are conditionally compiled under MIPS will never be any different if
  `__USE_FILE_OFFSET64` is defined.

  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/x86_64/bits/fcntl.h#L166-L189>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/mips/bits/fcntl.h#L164-L198>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/arm/bits/fcntl.h#L152-L175>
- Upstream uclibc-ng sources showing how `statvfs` and `statvfs64` are
  independent, but like the above, have equivalent definitions when
  `__USE_FILE_OFFSET64` is defined. This also shows how there does not seem to
  be a need for the fields to be reordered depending on the system's endianness.

  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/statvfs.h#L28-L75>
- Upstream uclibc-ng sources showing how `statfs` and `statfs64` are
  independent, but like the above, have equivalent definitions when
  `__USE_FILE_OFFSET64` is defined. A bunch of field types were changed to both
  accurately reflect the upstream types used, and to add missing fields. The
  former has not been documented in full in the below sources because the types
  involved are already part of the sources cited in other list items. Some
  required special-casing in MIPS platforms has been necessary, as there were
  fields that had altogether different types. These were only ever so slightly
  tweaked for the same reasons as above. `statfs` in MIPS got unified as the
  definitions for both the o32 and N64/N32 ABIs are equivalent under
  `__USE_FILE_OFFSET64`.

  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/statfs.h#L24-L64>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/mips/bits/statfs.h#L24-L70>
- Upstream uclibc-ng sources showing the same as the above but for `stat` and
  `stat64`. There was also some special casing required in MIPS platforms,
  though the definitions for both x86_64 and arm converged. These have been kept
  in their corresponding modules because there's need for MIPS to have their own
  and it was decided against a top-level, field-specific `cfg`.

  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/common/bits/stat.h#L35-L130>
  > <https://github.com/wbx-github/uclibc-ng/blob/50c470ef4e688e6eea6fd1eff13083e4fd7b7c95/libc/sysdeps/linux/mips/bits/stat.h#L38-L214>

## Checklist

  doc comment

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -4041,6 +4041,13 @@ fn test_linux(target: &str) {
         // glibc versions older than 2.29.
         .define("__GLIBC_USE_DEPRECATED_SCANF", None);
 
+    if uclibc
+        && pointer_width == 32
+        && env::var("CARGO_CFG_LIBC_UNSTABLE_UCLIBC_FILE_OFFSET_BITS").is_ok()
+    {
+        cfg.cfg("uclibc_file_offset_bits64", None);
+    }
+
     config_gnu_bits(target, &mut cfg);
     // The L4Re libc headers contain some L4Re helper functions which are not needed for the libc
     // interface and must not be added to the libc crate
--- a/build.rs
+++ b/build.rs
@@ -39,6 +39,8 @@ const ALLOWED_CFGS: &[&str] = &[
     "musl_redir_time64",
     "vxworks_lt_25_09",
     "libc_pauthtest",
+    // Corresponds with `__USE_FILE_OFFSET64` in uClibc.
+    "uclibc_file_offset_bits64",
 ];
 
 // Extra values to allow for check-cfg.
@@ -217,6 +219,13 @@ fn main() {
             set_cfg("gnu_file_offset_bits64");
             set_cfg("gnu_time_bits64");
         }
+    }
+
+    match env::var("CARGO_CFG_LIBC_UNSTABLE_UCLIBC_FILE_OFFSET_BITS") {
+        Ok(val) if val == "64" && target_env == "uclibc" && target_ptr_width == "32" => {
+            set_cfg("uclibc_file_offset_bits64")
+        }
+        _ => (),
     }
 
     // On CI: deny all warnings
--- a/build.rs
+++ b/build.rs
@@ -54,6 +54,8 @@ enum Cfg {
     /// transition to 64-bit `time_t` and thus need `__*_time64` redirects. Implies `musl_v1_2`
     /// and 32-bit arch.
     MuslRedirTime64,
+    /// Corresonds to `__USE_FILE_OFFSET64` in uClibc-ng.
+    UclibcFileOffsetBits64,
 }
 
 const ALLOWED_CFGS: &[Cfg] = &[
@@ -76,6 +78,7 @@ const ALLOWED_CFGS: &[Cfg] = &[
     Cfg::MuslV1_2,
     Cfg::Musl32Time64,
     Cfg::MuslRedirTime64,
+    Cfg::UclibcFileOffsetBits64,
 ];
 
 impl fmt::Display for Cfg {
@@ -100,6 +103,7 @@ impl fmt::Display for Cfg {
             Cfg::MuslV1_2 => "musl_v1_2",
             Cfg::Musl32Time64 => "musl32_time64",
             Cfg::MuslRedirTime64 => "musl_redir_time64",
+            Cfg::UclibcFileOffsetBits64 => "uclibc_file_offset_bits64",
         })
     }
 }
@@ -263,9 +267,16 @@ fn main() {
     }
 
     let uclibc_time64_env = env_flag("CARGO_CFG_LIBC_UNSTABLE_UCLIBC_TIME64");
+    let uclibc_off64_env = env_flag("CARGO_CFG_LIBC_UNSTABLE_UCLIBC_OFF64");
+
     let uclibc_time64 = target_env == "uclibc" && uclibc_time64_env;
+    let uclibc_off64 = target_env == "uclibc" && target_ptr_width == "32" && uclibc_off64_env;
+
     if uclibc_time64 {
         cfgs.push(Cfg::LinuxTimeBits64);
+    }
+    if uclibc_off64 {
+        cfgs.push(Cfg::UclibcFileOffsetBits64);
     }
 
     let mut gnu_tb_env = env::var("CARGO_CFG_LIBC_UNSTABLE_GNU_TIME_BITS");
@@ -491,6 +502,10 @@ fn validate_cfg(list: &[Cfg], target_env: &str, target_os: &str, target_ptr_widt
                     list.contains(&Cfg::MuslV1_2),
                     "{cfg:?} set on non-musl1.2 platform"
                 )
+            }
+            Cfg::UclibcFileOffsetBits64 => {
+                assert_eq!(target_env, "uclibc", "{cfg:?} set with env {target_env}");
+                assert_eq!(target_ptr_width, "32", "{cfg:?} set on non-32-bit platform");
             }
         }
     }
--- a/libc-test/build/main.rs
+++ b/libc-test/build/main.rs
@@ -3995,6 +3995,10 @@ fn test_linux(t: &Target) {
         // glibc versions older than 2.29.
         .define("__GLIBC_USE_DEPRECATED_SCANF", None);
 
+    if uclibc && p32 && env::var("CARGO_CFG_LIBC_UNSTABLE_UCLIBC_OFF64").is_ok() {
+        cfg.cfg("uclibc_file_offset_bits64", None);
+    }
+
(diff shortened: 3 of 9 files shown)


tgross35:
Seems reasonable to me from a quick skim, but we need the maintainers to weigh whether this is needed.

tgross35:
@skrap any thoughts here? Pretty much up to you for whether or not this is worth it.

libc-test/build.rs
```diff
@@ -4041,6 +4041,13 @@ fn test_linux(target: &str) {
         // glibc versions older than 2.29.
         .define("__GLIBC_USE_DEPRECATED_SCANF", None);
 
+    if uclibc
+        && pointer_width == 32
+        && env::var("CARGO_CFG_LIBC_UNSTABLE_UCLIBC_FILE_OFFSET_BITS").is_ok()
```
tgross35: One note without looking at the rest is this should check the value, `CARGO_CFG_LIBC_UNSTABLE_UCLIBC_FILE_OFFSET_BITS=1` shouldn't work. But a binary flag like `libc_unstable_uclib_off64` is likely better, I kind of regret accepting multiple values for the glibc flag.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5315/5db5914d7ecf8d3fe32394d315f29012039630ca..729cebcaadb9996e850970d5217a6863d1a1d4ef)*

dybucc: Done.

skrap:
> @skrap any thoughts here? Pretty much up to you for whether or not this is worth it.

Sorry, I was AFK for a bit.  I'll try to have a look at this by early/mid next week.

PR #5061, opened 2026-04-16
fix(newlib): update stat struct to match official headers

chengr4:
# Description

This PR fixes rust-lang/libc#2795

> Note: If any platforms that meet the above criteria are added, they would need to create their own definition and use that (maybe a `cfg_if!`) - by ian-h-chamberlain

> ps. thank @zetanumbers for investigating targets' compatibility

# Sources

- https://sourceware.org/git/?p=newlib-cygwin.git;a=blob;f=newlib/libc/include/sys/stat.h;h=6525272dd40123d6ea4e7355a8e6f3a3f3d0dbb8;hb=HEAD

# Checklist

- [x] Relevant tests in `libc-test/semver` have been updated
- [x] No placeholder or unstable values like `*LAST` or `*MAX` are
  included (see [#3131](https://github.com/rust-lang/libc/issues/3131))
- [x] Tested locally (`cd libc-test && cargo test --target mytarget`);
  especially relevant for platforms that may not be checked in CI

Issue #2795, opened 2022-05-22
Incorrect newlib `stat` struct definition


ZetaNumbers:
https://github.com/rust-lang/libc/blob/1a311288a8f7523bfe5f88f57c6e96724eb16842/src/unix/newlib/generic.rs#L8-L26

[newlib/libc/include/sys/stat.h](https://sourceware.org/git/?p=newlib-cygwin.git;a=blob;f=newlib/libc/include/sys/stat.h;h=6525272dd40123d6ea4e7355a8e6f3a3f3d0dbb8;hb=HEAD)

```c
struct  stat 
{
  dev_t         st_dev;
  ino_t         st_ino;
  mode_t        st_mode;
  nlink_t       st_nlink;
  uid_t         st_uid;
  gid_t         st_gid;
  dev_t         st_rdev;
  off_t         st_size;
#if defined(__svr4__) && !defined(__PPC__) && !defined(__sun__)
  time_t        st_atime;
  time_t        st_mtime;
  time_t        st_ctime;
#else
  struct timespec st_atim;
  struct timespec st_mtim;
  struct timespec st_ctim;
  blksize_t     st_blksize;
  blkcnt_t      st_blocks;
#if !defined(__rtems__)
  long          st_spare4[2];
#endif
#endif
};

#if !(defined(__svr4__) && !defined(__PPC__) && !defined(__sun__))
#define st_atime st_atim.tv_sec
#define st_ctime st_ctim.tv_sec
#define st_mtime st_mtim.tv_sec
#endif
```

I would like to see `defined(__svr4__) && !defined(__PPC__) && !defined(__sun__)` case being handled for targets without nanoseconds.

JohnTitor:
@AzureMarker Could you take a look at this? A breaking change (hiding some fields on some env) would be required to fix it, but I guess you're using this struct.

ZetaNumbers:
Oh and I believe after some digging some months ago, i found no rust target that would satisfy `defined(__svr4__) && !defined(__PPC__) && !defined(__sun__)`.

AzureMarker:
We're only using the struct for the "horizon" OS (Nintendo 3DS):
https://github.com/rust-lang/libc/blob/1a311288a8f7523bfe5f88f57c6e96724eb16842/src/unix/newlib/horizon/mod.rs#L59-L74

PR #5322, opened 2026-07-23
Re-add `stat` definition for Vita and 3DS (to fix compilation on main)

pheki:
## Description

Re-add `stat` definition for Vita and 3DS to fix compilation on main

[This comment](https://github.com/rust-lang/libc/pull/5132#discussion_r3631645349) made me realize that these targets are failing on main (and I'm one of the target maintainers for `armv7-sony-vita-newlibeabihf`).

It seems that both were broken by rust-lang/libc#5061 becasuse of a missing re-export, but at the same time it made me realize that the new definition seems incorrect for the Vita.

VitaSDK uses its own newlib fork, with its `stat` defined in https://github.com/vitasdk/newlib/blob/fbb8375870d799758e0b4358d2f1708781aa26a6/newlib/libc/sys/vita/sys/stat.h#L27-L55

For horizon, that definition is the same as the old one (and libc-0.2 one), and the discussion in rust-lang/libc#2795 implies it's correct, but I'm not sure where can we find the canonical headers for it.

## Checklist

- [X] Commit messages permalink to headers for added or changed API
- [X] Placeholder or unstable values like `*LAST` or `*MAX` have the standard
  doc comment

Checked locally using both `./etc/libc-util.py check-all-targets --only {target} libc` and `cargo +nightly check --target {target} -Z build-std=core` for:

- `armv7-sony-vita-newlibeabihf`
- `armv6k-nintendo-3ds`

And TYSM for maintaining `libc`!

### DIFF
--- a/etc/libc-util.py
+++ b/etc/libc-util.py
@@ -344,8 +344,6 @@ class CheckAllTargets:
         # libc problems
         ("aarch64-unknown-nto-qnx800", "libc error, unsupported arch"),
         ("aarch64.*-gnu_ilp32.*time_bits=64", "libc error, time64 mismatches"),
-        ("armv6k-nintendo-3ds", "libc error, stat missing"),
-        ("armv7-sony-vita-newlibeabihf", "libc error, stat missing"),
         ("csky.*gnu.*time_bits=64", "libc error, time64 mismatches"),
         ("i686-pc-nto-qnx700", "libc error, unsupported arch"),
         ("managarm-mlibc", "libc error, unresolved import"),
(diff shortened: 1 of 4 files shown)


tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5322/0f50a2e49ffb7d660127228e63292f8d67715777..e60b38942045e82885a0016e6a4e4e0c984c1da0)*

tgross35:
@rustbot label -stable-nominated +stable-declined

I guess we don't need this backported because https://github.com/rust-lang/libc/pull/5061 wasn't either. If there is some or all of that you think makes sense to backport, let me know.

PR #5325, opened 2026-07-24
freebsd: move `net/if_mib.h` to `src/new` module

dybucc:
## Description

This PR updates rust-lang/libc#3367 by rebasing onto latest `main` and adjusting the solution
to fit the plan outlined at [^1].

The patch itself moves the `net/if_mib.h` definitions that overlap with those of
`netlink/netlink.h` into a module of their own under the `new` module. This is
to avoid blocking future additions of `netlink.h` API, which seems more
prevalent according to the author of rust-lang/libc#3367.

This should unblock rust-lang/libc#3201.

## Checklist

  doc comment

### DIFF
--- a/src/new/freebsd/net/if_mib.rs
+++ b/src/new/freebsd/net/if_mib.rs
@@ -1,0 +1,48 @@
+//! Header: `net/if_mib.h`
+//!
+//! https://github.com/freebsd/freebsd-src/blob/86691d52a6d3796ad36ba474cf0a9493f6d99202/sys/net/if_mib.h
+
+use crate::prelude::*;
+
+/// non-interface-specific
+pub const IFMIB_SYSTEM: c_int = 1;
+/// per-interface data table
+pub const IFMIB_IFDATA: c_int = 2;
+
+/// generic stats for all kinds of ifaces
+pub const IFDATA_GENERAL: c_int = 1;
+/// specific to the type of interface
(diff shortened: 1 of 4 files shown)


JohnTitor (approved):
Thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5325/0f50a2e49ffb7d660127228e63292f8d67715777..4bb032174ed3e4289f0dc65ba3c635da33d6520d)*

dybucc:
I've not looked too deep into this, but from the VMActions docs [^1], maybe the
issue is that the `prepare` step is cached and that just skips altogether the
Rust install?

Maybe the fix goes through explicitly disabling the `cache-after-prepare`
configuration option? I could very well be wrong, but the docs don't really
mention anything about the side effects of that step being cached, so maybe
that's why the Rust install is skipped in runs where the cache is still hot.

__Edit__: forget I said anything; Just noticed the logs report the run was not
cached. The issue seems to be in the curl download of the rustup script, which
seems to have one of the two endpoints unexpectedly disconnect from the TCP
connection.

[^1]: <https://github.com/vmactions/freebsd-vm/tree/77ed28d336d03fe19a3f4f7266c1d2c4714dd79d/#10-cache-the-vm-image-after-prepare>

PR #5326, opened 2026-07-24
freebsd(13.2): add `netlink/netlink.h` support

### LINKED ISSUE #3194: Netlink support on FreeBSD 13.2+

dybucc:
## Description

This PR updates rust-lang/libc#3201 with merge conflicts resolved and follows the new plan at
[^1].

The patch adds support for `netlink.h` interfaces in OpenBSD, where there's an
item resolution conflict if we expose the Rust bindings alongside those of
`if_mib.h`. This set of APIs is "scoped" in C because they live on separate
headers. In rust-lang/libc, we reexport all items at the root crate level, which
makes item resolution fail.

Note this depends on rust-lang/libc#5325. It won't pass tests but it will build. This is
because the test templates will gather all items in a single file, so item
resolution fails. We can't really skip these items altogether from the tests, so
it may just be necessary to extend `ctest` to allow skipping module-specific
Rust items.

Closes: https://github.com/rust-lang/libc/pull/3201
Closes: https://github.com/rust-lang/libc/issues/3194

## Checklist

  doc comment


tgross35:
The API looks fine from a quick skim, but since there is no hurry, I think it may be worth trying to add support to ctest first so the tricky test setup isn't needed. (It's useful otherwise too.)

Sketched some of that up at https://github.com/rust-lang/libc/issues/5344

dybucc:
Noted. Since you already pinged some contributor on that issue, I'll wait and
see whether they want to implement it. Otherwise, I'll do it.

tgross35:
Could you try adding a separate `TestGenerator` instance like @mbyx recommended in that issue? Since it's split off, that should avoid the need to reexport `libc::netlink` at test root too.

tgross35:
Either author or blocked, depending on whether that works.

@rustbot author

libc-test/build/main.rs
```diff
@@ -2518,6 +2522,103 @@ fn test_freebsd(target: &str) {
...
+                !matches!(
+                    c.ident(),
+                    "CTRL_CMD_UNSPEC"
+                        | "CTRL_CMD_NEWFAMILY"
+                        | "CTRL_CMD_DELFAMILY"
+                        | "CTRL_CMD_GETFAMILY"
+                        | "CTRL_CMD_NEWOPS"
+                        | "CTRL_CMD_DELOPS"
+                        | "CTRL_CMD_GETOPS"
+                        | "CTRL_CMD_NEWMCAST_GRP"
```
tgross35: These could be put into functions like `is_netlink_const(c: &Const) -> bool` so the list can be shared the two places it's needed.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5326/023c0a3c184c2d7fcd528b64e867075660a5884d..266acdfdf9bf20b6bbd141f7f4e3a2ba226e895c)*

libc-test/build/main.rs
```diff
@@ -2518,6 +2522,103 @@ fn test_freebsd(target: &str) {
     let freebsd14 = matches!(freebsd_ver, Some(n) if n >= 14);
     let freebsd15 = matches!(freebsd_ver, Some(n) if n >= 15);
 
+    if let Some(net_header) = net_header {
```
tgross35: I think you can drop the `net_header` arg, one invocation of `test_freebsd` can create all 2-3 `ctest` objects. I assume you're doing this to share the setup above, so I applied 023c0a3c184c2d7fcd528b64e867075660a5884d to make that cfg a bit easier to share.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5326/023c0a3c184c2d7fcd528b64e867075660a5884d..266acdfdf9bf20b6bbd141f7f4e3a2ba226e895c)*

src/new/freebsd/netlink/mod.rs
```diff
@@ -0,0 +1,6 @@
+//! Directory: `netlink/`
+//!
+//! <https://github.com/freebsd/freebsd-src/tree/df9d6403caa6426e92f5e100602f4d2be474bbae/sys/netlink>
```
tgross35: Fyi, no need to permalink these kind of files since we do want to see updates, and the file locations are usually stable enough. As long as there's a permalink in the commit message to show us what it actually looked like at that point in time.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5326/023c0a3c184c2d7fcd528b64e867075660a5884d..266acdfdf9bf20b6bbd141f7f4e3a2ba226e895c)*

libc-test/build/main.rs
dybucc: Done.

libc-test/build/main.rs
dybucc: Done. I didn't actually mean that. That was just cruft from prior attempts with
a different approach.

src/new/freebsd/netlink/mod.rs
dybucc: You mean the files for the upstream directories only, or also the files
referring to the upstream header files?

dybucc:
It seems having separate invocations gets the job done just fine. Still, that
was the same thing I tried in the initial patch. The added complexity comes when
we need to get the SemVer tests to pass.

Those seem to me like they're going to need extending (though it's notably less
work than with ctest as it's all in libc-test's build script.)

tgross35:
I haven't tried this but do things just work™️ if the semver tests have lines like `netlink::NETLINK_ADD_MEMBERSHIP`?

dybucc:
>I haven't tried this but do things just work™️ if the semver tests have lines
>like `netlink::NETLINK_ADD_MEMBERSHIP`?

@tgross35 They do. Just cleaned up commit history as all tests seem to pass now.

src/new/mod.rs
```diff
@@ -65,7 +65,14 @@ cfg_if! {
...
         mod freebsd;
-        pub(crate) use freebsd::*;
+        // FIXME(1.0,remove): glob reexport should be the default.
+        pub use freebsd::netlink;
+        pub(crate) use freebsd::{
+            net,
+            netinet6,
+            sys,
+            unistd,
+        };
```
tgross35: Does this work?

```rust
pub(crate) use freebsd::*;
pub use netlink;
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5326/5b25cab671fe373fab2b672e47fe6baaa4410951..30133e03605b4f8d6adb67e87d23c560a455fe26)*

src/new/freebsd/netlink/netlink_generic.rs
```diff
@@ -0,0 +1,33 @@
...
+pub const CTRL_ATTR_MAXATTR: c_int = 5;
+pub const CTRL_ATTR_OPS: c_int = 6;
+pub const CTRL_ATTR_MCAST_GROUPS: c_int = 7;
+pub const CTRL_ATTR_POLICY: c_int = 8;
+pub const CTRL_ATTR_OP_POLICY: c_int = 9;
+pub const CTRL_ATTR_OP: c_int = 10;
+
+pub const CTRL_ATTR_MCAST_GRP_UNSPEC: c_int = 0;
+pub const CTRL_ATTR_MCAST_GRP_NAME: c_int = 1;
+pub const CTRL_ATTR_MCAST_GRP_ID: c_int = 2;
```
tgross35: I think these aren't in the semver file

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5326/5b25cab671fe373fab2b672e47fe6baaa4410951..30133e03605b4f8d6adb67e87d23c560a455fe26)*

libc-test/semver/freebsd.txt
```diff
@@ -2172,6 +2172,59 @@ msgrcv
...
+netlink::netlink::NLM_F_ECHO
+netlink::netlink::NLM_F_EXCL
+netlink::netlink::NLM_F_MATCH
+netlink::netlink::NLM_F_MULTI
+netlink::netlink::NLM_F_NONREC
+netlink::netlink::NLM_F_REPLACE
+netlink::netlink::NLM_F_REQUEST
+netlink::netlink::NLM_F_ROOT
+netlink::netlink::NL_ITEM_ALIGN_SIZE
+netlink::netlink::SOL_NETLINK
```
tgross35: It would be better if we can keep all this directly within a `netlink` module. In the exports, perhaps an in-file `mod netlink { pub use ... }`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5326/5b25cab671fe373fab2b672e47fe6baaa4410951..30133e03605b4f8d6adb67e87d23c560a455fe26)*

libc-test/build/main.rs
```diff
@@ -2518,6 +2606,28 @@ fn test_freebsd(target: &str) {
...
+
+    // We must restore the above sure-skips because `TestGenerator` shares skips
+    // between cloned instances (here `cfg` and `netlink_cfg`.)
+    cfg.skip_struct(|_| false)
+        .skip_const(|_| false)
+        .skip_union(|_| false)
+        .skip_alias(|_| false)
+        .skip_static(|_| false)
+        .skip_fn(|_| false)
+        .skip_c_enum(|_| false);
```
tgross35: Sharing skips isn't intentional, why does this happen?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5326/5b25cab671fe373fab2b672e47fe6baaa4410951..30133e03605b4f8d6adb67e87d23c560a455fe26)*

libc-test/build/main.rs
```diff
@@ -2986,6 +3100,10 @@ fn test_freebsd(target: &str) {
 
     cfg.skip_struct(move |struct_| {
         match struct_.ident() {
+            // This is tested in a separate ctest invocation because some symbols
+            // `netlink/netlink.h` conflict with some other symbols from `net/if_mib.h`.
+            "sockaddr_nl" => true,
```
tgross35: Is this one needed still? I think we should be checking a `sockaddr_nl` in each `ctest` invocation

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5326/5b25cab671fe373fab2b672e47fe6baaa4410951..30133e03605b4f8d6adb67e87d23c560a455fe26)*

dybucc: Yes, it's still needed. `sockaddr_nl` exists only under `netlink/` in the
FreeBSD tree [^1]. Not skipping it would cause failure in the non-`netlink` test
run because of undefined symbols on the C side of things.

[^1]: <https://github.com/search?q=repo%3Afreebsd%2Ffreebsd-src+path%3Asys%2F**%2F*.h+symbol%3A%2F%5Cbsockaddr_nl%5Cb%2F&type=code>

libc-test/build/main.rs
dybucc: A prior review comment [^1] mentioned a recent patch that allows sharing skips
between instances of `TestGenerator`. I assumed you wanted me to use the now
shallow copy semantics of that type to avoid calling anew `ctest_cfg()` for
`netlink_cfg`.

This then needs resetting to the "defaults," because otherwise some skipping
functions that we do call with `netlink_cfg` but not with `cfg` end up affecting
`cfg`.

[^1]: <https://github.com/rust-lang/libc/pull/5326#discussion_r3747219428>

libc-test/semver/freebsd.txt
dybucc: I believe that's now fixed. The `freebsd::netlink` module exposed both `netlink`
and `netlink_generic` submodules, so I instead changed the reexport in
`src/new/mod.rs` to have the inner two modules exported.

Not everything is within a single `netlink` module, but at least we have a
shortened path.

src/new/freebsd/netlink/netlink_generic.rs
dybucc: Done.

src/new/mod.rs
dybucc: Not quite, but something similar does work. The current reexports have been
changed to instead have first the "regular" modules, followed by a reexport of
the `freebsd::netlink` module.

The latter itself contains two submodules, `netlink` and `netlink_generic`,
which are now also used in the SemVer tests, as well as being the only ones
exposed to downstream crates.

libc-test/build/main.rs
tgross35: Sort of, but there's no shallow copy. I guess the `Rc<dyn Fn()>` might look that way but it's just letting closures be used from multiple places like function pointers - it's not like cloning to a `TestGenerator` then adding skips to one means that the other `TestGenerator` gets those same skips.

(The pattern for something like that would be `field: Rc<RefCell<Box<dyn Fn()>>>` or `field: Rc<RefCell<Vec<Box<dyn Fn()>>>`.)

So you can delete the `.skip_union(|_| false) ...`

libc-test/build/main.rs
```diff
@@ -2502,6 +2590,28 @@ fn test_freebsd(t: &Target) {
     let freebsd14 = matches!(freebsd_ver, Some(n) if n >= 14);
     let freebsd15 = matches!(freebsd_ver, Some(n) if n >= 15);
 
+    let mut netlink_cfg = cfg.clone();
+    headers!(netlink_cfg, "netlink/netlink.h",);
+    netlink_cfg
+        .skip_struct(|ty| !matches!(ty.ident(), "sockaddr_nl"))
```
tgross35: Nit: put this in a `is_netlink_struct` function. Not super useful on its own but it's consistent with `is_netlink_const` and easier to expand if needed, and will the `"sockaddr_nl" => true,` more obvious.

src/new/freebsd/netlink/mod.rs
tgross35: For any of the files in `new`, regular links rather than permalinks are fine. Just makes it easier to see the most recent version without them going out of sync.

src/new/freebsd/netlink/netlink_generic.rs
tgross35: Looks like these could all be `c_enum!` to match source

src/new/mod.rs
tgross35: I think it would be preferable to have a single module since that's how most of the crate is. If there's no easy then it would be fine with just:

```rust
pub mod netlink {
    use super::freebsd;
    pub use freebsd::netlink::*;
    pub use freebsd::netlink_generic::*;
}
```

tgross35 (changes requested):
Cc @asomers, could you take a look when you get the chance?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5326/8b4ebdc6514d20ddbdac718aade170bf400e2515..be6cc74a64878d2d0fbc36bf9f9f55e8670ff2d0)*

src/new/freebsd/netlink/netlink.rs
```diff
@@ -0,0 +1,77 @@
...
+pub const NETLINK_PKTINFO: c_int = 3;
+pub const NETLINK_BROADCAST_ERROR: c_int = 4;
+pub const NETLINK_NO_ENOBUFS: c_int = 5;
+pub const NETLINK_RX_RING: c_int = 6;
+pub const NETLINK_TX_RING: c_int = 7;
+pub const NETLINK_LISTEN_ALL_NSID: c_int = 8;
+pub const NETLINK_LIST_MEMBERSHIPS: c_int = 9;
+pub const NETLINK_CAP_ACK: c_int = 10;
+pub const NETLINK_EXT_ACK: c_int = 11;
+pub const NETLINK_GET_STRICT_CHK: c_int = 12;
```
asomers: What about NETLINK_MSG_INFO ?

src/new/freebsd/netlink/netlink.rs
```diff
@@ -0,0 +1,77 @@
...
+
+pub const NLM_F_NONREC: c_int = 0x100;
+
+pub const NLM_F_CAPPED: c_int = 0x100;
+pub const NLM_F_ACK_TLVS: c_int = 0x200;
+
+pub const NLMSG_NOOP: c_int = 0x1;
+pub const NLMSG_ERROR: c_int = 0x2;
+pub const NLMSG_DONE: c_int = 0x3;
+pub const NLMSG_OVERRUN: c_int = 0x4;
```
asomers: What about NLMSG_MIN_TYPE?  It's not clear to me from the header if that constant is expected to be constant for all time, or if it can change with newer netlink versions.  Or if it's supposed to be kernel-only.  In the latter two cases, it would make sense to leave it out.

src/new/freebsd/netlink/mod.rs
dybucc: Done.

src/new/mod.rs
dybucc: Done. For the sake of keeping a close matching between our module tree and
upstream's directory tree under `sys`, I've gone with the above solution.

libc-test/build/main.rs
dybucc: Done, but I've still got questions.

How is there not? `TestGenerator` automatically derives `Clone` and that would
make the inner `Vec<Skip>` use its `Clone` implementation, which itself depends
on the `Clone` impl of `Skip`. `Skip` is just an alias to a `Rc<...>`, so it
would call into the `Clone` impl for `Rc<T: ?Sized>`, which will yield a shallow
copy of `T`.

Granted, you'd still need some form of interior mutability to change those
skips, but I don't think skips are made to be "modified" once "registered."

What am I missing?

libc-test/build/main.rs
dybucc: Done.

src/new/freebsd/netlink/netlink_generic.rs
dybucc: Done.

libc-test/build/main.rs
```diff
@@ -2470,6 +2470,98 @@ fn test_android(t: &Target) {
 }
 
 fn test_freebsd(t: &Target) {
+    // Ensures we have only one place where the `netlink/netlink.h` constants
+    // are checked in. This is used both when generating the full test suite,
+    // and when generating the `netlink`-specific test suite.
+    #[inline]
+    fn is_netlink_const(it: &ctest::Const) -> bool {
```
tgross35: Drop `#[inline]`, no need to really worried about that kind of performance optimizations in tests and this isn't a bottleneck.

src/new/freebsd/mod.rs
```diff
@@ -5,5 +5,6 @@
 
 pub(crate) mod net;
 pub(crate) mod netinet6;
+pub mod netlink;
```
tgross35: This can now be `pub(crate)` right?

src/new/mod.rs
```diff
@@ -65,7 +65,14 @@ cfg_if! {
...
         mod freebsd;
-        pub(crate) use freebsd::*;
+        pub(crate) use freebsd::{net, netinet6, sys, unistd};
+        // FIXME(1.0,remove): should be part of the above reexport
+        pub mod netlink {
+            use super::freebsd;
+
+            pub use freebsd::netlink::netlink::*;
+            pub use freebsd::netlink::netlink_generic::*;
+        }
```
tgross35, line 74: Maybe just rename `src/new/freebsd/netlink/netlink.rs` to `netlink_.rs` or something, so we can keep the glob export? Think this might be done elsewhere as well.

Just add a comment by the `mod netlink_` explaining why it has that name.

libc-test/build/main.rs
```diff
@@ -2517,6 +2609,18 @@ fn test_freebsd(t: &Target) {
...
+    let mut netlink_cfg = cfg.clone();
+    headers!(netlink_cfg, "netlink/netlink.h",);
+    netlink_cfg
+        .skip_struct(|ty| !is_netlink_struct(ty))
+        .skip_const(|c| !is_netlink_const(c))
+        .skip_union(|_| true)
+        .skip_alias(|_| true)
+        .skip_static(|_| true)
+        .skip_fn(|_| true)
+        .skip_c_enum(|_| true);
```
tgross35, line 2623: These shouldn't be needed anymore right?

libc-test/build/main.rs
tgross35: IME the term "shallow copy" typically means (1) there exists a real or theoretical corresponding deep copy, and (2) the difference between a shallow and deep copy is observable. With this code:

```rust
let cfg1 = TestGenerator::new();
let cfg2 = cfg1.clone();
cfg1.skip_struct(|ty| ty.ident() == "foo");
// ...
```

I'd definitely consider it a shallow copy if both `cfg1` and `cfg2` skip `foo` as a result (would be pretty surprising API). I thought that's what you were expecting at first based on the comments and is why I mentioned the `RefCell` patterns, but perhaps this wasn't correct?

That shouldn't be what happens, `cfg1` and `cfg2` are decoupled. Existing skips do stick around when cloning, but it's more like cloning a `Vec` to get the same base items then being able to add on independently.

There _is_ a "shallow copy" with the `Rc<dyn Fn()>` and it wouldn't be technically incorrect to refer to it as such, but I'd find it an unintuitive use of the term because you need to go out of your way and use interior mutability to observe it. It's not too far off from duplicating a `fn()` or `&u32`—technically shallow and observable, but the distinction isn't especially meaningful.

This is mostly just opinion, though, I'm sure others may look at it differently.

tgross35 (changes requested):
Also it looks like CI isn't yet passing, currently stuck on style but make sure the FreeBSD tests pass too.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5326/2eda1e52fe2efdf93c1e5c65e346343a530d8bc3..d22b942f3709ec49990b7dd2bcf2c1c9175f8c4e)*

libc-test/build/main.rs
dybucc: Thanks, I understand now. Were we to clone later on in the body of some
test function, the skips would, indeed, be carried over to the “new”
instance of `TestGenerator`. But we clone before any skips are set, so
those that are set on the clone don’t affect the original instance that
we will be using later on.

Cloning causes branching, which means cloning at the start of the “tree”
will create a separate ramification, which only carries as shared state
that which was already present at the point of cloning, but nothing
more.

src/new/mod.rs
dybucc: I don't think this is a good idea. The `new` module's directory layout
is meant to provide a fairly accurate mapping between upstream and our
Rust module tree. Starting to allow this type of "exceptions" seems to
me like we're just muddying things up (to eventually be back at having
something as hard to grok as we have now outside `new`.)

I don't know of prior precedent to this, so we could very well be
already allowing this. Still, considering this is simply to replace an
inline module with the containing reexports, it doesn't seem worth it.

libc-test/build/main.rs
dybucc: I don't think so. IIUC, `TestGenerator` will parse the Rust files, which
we presently can't restrict to a subset of the codebase. If these are
removed, we'd be setting up tests for all the types that are not part of
the `netlink` module, even if we've only provided the `#include` for the
`netlink` interfaces (a few lines above this comment.)

src/new/freebsd/mod.rs
dybucc: Yup. Done.

libc-test/build/main.rs
dybucc: Done.

dybucc:
~~CI fails now because of a recent `unused`-group lint that was added in
rust-lang/rust#157036. We happen to have macros that add this by
default, but that don't merge them when munching the `repr` attributes.
Maybe that is the solution, but I have other PR reviews to go through
first, so I can't work on that just yet.~~

~~**Edit**: The CI failures are already being addressed at #5446.~~

@rustbot ready

libc-test/build/main.rs
tgross35: Ah, right, that's inconvenient. Mind adding a comment?

Would be nice if we had a way to restrict to certain modules, feel free to open an issue if you're up to it.

src/new/mod.rs
tgross35: We do have a few exceptions when things don't work well, e.g. `if.h` is a common header that needs to be named `if_.rs` because of the keyword. To me it's worth a minor exception to avoid duplicate the API list in `new/mod.rs`, those should be pretty few and far between. But I won't block on it.

tgross35:
Overall looks pretty good to me, @asomers or original PR author @ydirson would you mind taking a look?

src/new/mod.rs
dybucc: I am guessing you mean to rename the top-level `freebsd::netlink`
module, as that is the one that is reexported in `new`'s root module. If
so, done.

libc-test/build/main.rs
dybucc: Flawless nerd sniping.

**Edit**: I'll get to it once I'm done reviewing reviews.

PR #5339, opened 2026-07-29
add environ to unix

### LINKED ISSUE #2520: Add "environ"
Please, reexport UNIX's variable `environ`

https://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html

LusterSourav:
see commit messages.

- [x] `libc-test/semver/unix.txt` updated
- [x] commit messages link to POSIX exec spec
- [x] not applicable (no placeholder values)
- [x] `cargo test -p libc-test` — 15 passed

Closes rust-lang/libc#2520

### DIFF
--- a/libc-test/semver/linux-gnu.txt
+++ b/libc-test/semver/linux-gnu.txt
@@ -597,6 +597,7 @@ dlmopen
 dlvsym
 eaccess
 endutxent
+environ
 epoll_pwait2
 ethhdr
 euidaccess
--- a/src/unix/linux_like/linux/gnu/mod.rs
+++ b/src/unix/linux_like/linux/gnu/mod.rs
@@ -1304,6 +1304,9 @@ extern "C" {
     pub fn mempcpy(dest: *mut c_void, src: *const c_void, n: size_t) -> *mut c_void;
 
     pub fn tgkill(tgid: crate::pid_t, tid: crate::pid_t, sig: c_int) -> c_int;
+
+    // glibc provides this in <unistd.h> with _GNU_SOURCE
+    pub static mut environ: *mut *mut c_char;
 }
 
 cfg_if! {


libc-test/build.rs
LusterSourav, line 1: Need to add FreeBSD too...

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5339/f51fe550b737a2a6303c55955f2c3e95da127fae..4ddcfba3251535cfce41e3890a8717718512c405)*

libc-test/build.rs
```diff
@@ -407,6 +407,9 @@ fn test_apple(target: &str) {
         }
     });
 
+    // not in apple headers, declare it yourself
+    cfg.skip_static(|s| s.ident() == "environ");
```
tgross35: If it's not even available in headers, I don't think we should be declaring it either.

The gnu manpage https://man7.org/linux/man-pages/man7/environ.7.html says:

> Historically and by standard, environ must be declared in the user
> program.  However, as a (nonstandard) programmer convenience,
> environ is declared in the header file <unistd.h> if the
> _GNU_SOURCE feature test macro is defined (see
> [feature_test_macros(7)](https://man7.org/linux/man-pages/man7/feature_test_macros.7.html)).

We could define it on glibc but I'm not sure that's super useful if the users need to define it themselves in anything else. This kind of makes me iffy about adding `environ` at all. Any thoughts @JohnTitor?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5339/f51fe550b737a2a6303c55955f2c3e95da127fae..4ddcfba3251535cfce41e3890a8717718512c405)*

LusterSourav: The man page  quote says it itself   "environ must be declared in the user program." That is what libc is doing on Linux  glibc already provides it. On macOS/freebsd t he symbol is at link time, headers  just do not say so. Rust stdlib  already declares it directly...

LusterSourav:
image

LusterSourav:
image

LusterSourav:
> image

https://man7.org/linux/man-pages/man7/environ.7.html

LusterSourav:
> image

https://man7.org/linux/man-pages/man7/feature_test_macros.7.html

libc-test/build.rs
tgross35: `libc` isn't the user program, it's bindings to the platform `libc` - which may be glibc or may be something else.

I'm hesitantly okay to allow this on gnu targets or other platforms that specify it in headers (deferring to Yuki here), but I don't want to provide a definition in places the platform library doesn't. IOW, let's not add something just to need to skip its tests.

Please make sure your commit message links to headers as well. The POSIX spec is also helpful to have, but it doesn't tell us anything about which platforms actually implement it.

tgross35 (changes requested):
(updating status)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5339/f51fe550b737a2a6303c55955f2c3e95da127fae..abfd0d2197053c6ac911bc819a1fd5aeafa4b34b)*

LusterSourav:
environ now linux-gnu only. glibc declares it in <unistd.h> behind _GNU_SOURCE. no ctest skips elsewhere. commit messages link to headers.

libc-test/build.rs
LusterSourav: > `libc` isn't the user program, it's bindings to the platform `libc` - which may be glibc or may be something else.
> 
> I'm hesitantly okay to allow this on gnu targets or other platforms that specify it in headers (deferring to Yuki here), but I don't want to provide a definition in places the platform library doesn't. IOW, let's not add something just to need to skip its tests.
> 
> Please make sure your commit message links to headers as well. The POSIX spec is also helpful to have, but it doesn't tell us anything about which platforms actually implement it.

okeee
but 2 + linux libcs do declare in  environ right hm???
musl & bionic 
https://git.musl-libc.org/cgit/musl/plain/include/unistd.h

image


https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/include/unistd.h

image


same as glibc link, and bionic unconditionally linkso as per your advice they should be covered too so could i  add both hm? no skips needed .. since libc test already defines gnu sorucs for the linux house,apple/bsb alos out as header don’t declered will it work !!

LusterSourav: *bsd

LusterSourav: @tgross35

tgross35 (approved):
May as well take this but please clean up the history - 15 commits is a lot for a four line diff :)

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5339/9658498e6757664f9bb70c2f3ab41deaac5f9247..2675498a9293d1ca9227f1e726e1a101a8b98347)*

tgross35:
Please always include a link to headers in the commit message, the manpage is fine too but it's the headers we check against. I've already seen https://github.com/sailfishos-mirror/glibc/blob/4a07bb292f921c10e71fbf48c4a7f44391feb06c/posix/unistd.h#L566 though so this is okay.

PR #5342, opened 2026-07-30
linux: select powerpc64 speed symbols by ELF ABI

amaanq:
## Description

Starting with Rust 1.98, `powerpc64-unknown-linux-gnuelfv2` will be a
supported triple. During my attempts to try to build a bunch of projects
for this new target, though, libc failed to compile. This is because the
`cf*speed` link names assumed all big-endian powerpc64 targets use ELFv1,
which was reasonable in the past, but not anymore. Now, this uses the
target ABI to select the correct symbol versions instead.

## Checklist

- [ ] Commit messages permalink to headers for added or changed API
- [ ] Placeholder or unstable values like `*LAST` or `*MAX` have the standard
  doc comment

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -135,6 +135,7 @@ targets = [
     "powerpc64-ibm-aix",
     "powerpc64-unknown-freebsd",
     "powerpc64-unknown-linux-gnu",
+    "powerpc64-unknown-linux-gnuelfv2",
     "powerpc64-wrs-vxworks",
     "riscv32-wrs-vxworks",
     "riscv32gc-unknown-linux-musl",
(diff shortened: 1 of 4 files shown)


tgross35:
Maintainer @Gelbpunkt could you review?

Gelbpunkt:
Sorry for taking so long to reply, I was quite busy these last days. This change LGTM in principle, libc-test passes afterwards and fails to link previously and it's definitely correct.

The change to `verify-build.py` should also be replicated in the `Cargo.toml`.

amaanq:
> The change to verify-build.py should also be replicated in the Cargo.toml.

Oh yeah, thanks for catching that. I've updated `Cargo.toml` now

tgross35 (approved):
LGTM, thanks! And thanks @Gelbpunkt for reviewing.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5342/471be3d9247ddfa17e5c2139397db10b181f73a7..45226ffd3b74dddc5ac27fee12a1e429663541a5)*

PR #5345, opened 2026-07-31
linux: complete `siginfo_t` definition

### LINKED ISSUE #716: Problem in implementing siginfo_t for Linux
The definition of `siginfo_t` in Linux (https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/siginfo.h) uses a inlined `union` type as its last field. Rust did not have support for union types, I guess that was why those fields were left as a padding array in current `libc` implementation. 

I made some attempts to use `union` (introduced in Rust 1.19) to represent `siginfo_t`. However the resulting struct is 8 bytes larger than the original one (the one currently in libc, which is the same size as the C struct type). See the code here: https://play.rust-lang.org/?gist=99630206fcc8fd44b452d22552b37793&version=stable

My guess is the 8 more bytes come from two places: 
1. the union type is 8 bytes aligned, but the first 3 fields of `siginfo_t` are 12 bytes (3x 4-bytes integer), so the padding is needed after the first 3 fields, which makes the offset of the union type at 16 bytes (instead of 12 bytes)
2. because the union type is 8 bytes aligned, so the size of the union type cannot be 4*29 (which is not a multiple of the 8 bytes alignment), thus padding is needed at the end to make it a muliple of 8. 

I am not sure why the C definition does not have issues like this (probably because it uses inlined `union` in the struct definition) and how to properly implement `siginfo_t` in Rust. Please correct me if I am wrong. For my project, I have to extract data from the padding array (which looks awful).

dybucc:
## Description

Completes the `siginfo_t` definition under Linux targets. Closes rust-lang/libc#716.

See the patch messages for details on each of the GNU targets, uClibc
targets and musl targets.

## Checklist

      standard doc comment


xtqqczze:
[CI / Test tier1 (x86_64-unknown-linux-gnu) (pull_request)](https://github.com/rust-lang/libc/actions/runs/30643224956/job/91198110029?pr=5345) failure tracked by https://github.com/rust-lang/libc/issues/5350.

tgross35:
Would you mind using this as an opportunity to move `signifo_t` to `src/new`? At least on glibc, that should make things a lot easier with less duplication. No need to be perfect with `siginfo-arch.h` since that's tricky to replicate in Rust, `cfg` is fine.

Also fyi, @xtqqczze found https://github.com/sailfishos-mirror/glibc as a mirror for glibc that's much easier to search and less flaky, that's fine to link (as is sourceware, of course).

tgross35:
There will be quite a bit less to review after that, so will hold off taking a closer look. Let me know if you need help figuring out the module structure.

@rustbot author

dybucc:
I'm not sure what, but I had something else to do in this patchset. Hopefully
the review will surface it.

@rustbot ready

src/new/glibc/sysdeps/unix/linux/bits/types/siginfo_t.rs
```diff
@@ -0,0 +1,214 @@
...
+        si_uid: crate::uid_t,
+        si_sigval: crate::sigval,
+    }
+
+    struct __c_anonymous__si_fields__sigchld {
+        si_pid: crate::pid_t,
+        si_uid: crate::uid_t,
+        si_status: c_int,
+        si_utime: crate::clock_t,
+        si_stime: crate::clock_t,
```
tgross35: On x32, this needs some kind of alignment adjustment https://github.com/sailfishos-mirror/glibc/blob/4a07bb292f921c10e71fbf48c4a7f44391feb06c/sysdeps/unix/sysv/linux/x86/bits/siginfo-arch.h#L12-L14, maybe a `struct __SI_CLOCK_T(crate::clock_t)` that gets an alignment attribute on specific platforms.

src/new/glibc/sysdeps/unix/linux/bits/types/siginfo_t.rs
```diff
@@ -0,0 +1,214 @@
...
+
+const __SI_MAX_SIZE: usize = 128;
+const __SI_PAD_SIZE: usize = if cfg!(target_pointer_width = "64") {
+    __SI_MAX_SIZE / size_of::<c_int>() - 4
+} else {
+    __SI_MAX_SIZE / size_of::<c_int>() - 3
+};
+
+s_no_extra_traits! {
+    pub struct siginfo_t {
```
tgross35, line 46: Similarly it needs an alignment attribute on x32 for `__SI_ALIGNMENT`. I guess these were preexisting but may as well be fixed here.

src/new/glibc/sysdeps/unix/linux/sparc/bits/siginfo_arch.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    #[inline]
+    pub unsafe fn si_pkey(&self) -> u32 {
+        unsafe { self._sifields._sigfault._bounds._pkey }
+    }
+
+    #[cfg(target_pointer_width = "64")]
+    #[inline]
+    pub unsafe fn si_band(&self) -> c_int {
```
tgross35: This can be a `type __SI_BAND_TYPE = ...` to match the header

src/new/glibc/sysdeps/unix/linux/bits/types/siginfo_t.rs
tgross35, line 1: Could you merge the `siginfo_arch` files into `siginfo_t`? Since there are just a few changed fields we may as well reduce the duplication, and it's easier to see the differences (like the original source).

You _could_ keep the files separate for `__SI_BAND_TYPE` and `__SI_CLOCK_T` but I don't think that's worth it for 1-2 `cfg_if`s. Just make a note that this one file represents multiple headers.

src/new/musl/signal.rs
```diff
@@ -0,0 +1,209 @@
...
+
+s_no_extra_traits! {
+    // `mips*` targets swap the `s_errno` and `s_code` fields otherwise this struct is
+    // target-agnostic (see https://www.openwall.com/lists/musl/2016/01/27/1/2)
+    pub struct siginfo_t {
+        pub si_signo: c_int,
+        #[cfg(not(any(target_arch = "mips", target_arch = "mips64")))]
+        pub si_errno: c_int,
+        pub si_code: c_int,
+        #[cfg(any(target_arch = "mips", target_arch = "mips64"))]
```
tgross35: You'll also need `mips32r6` and `mips64r6` here, and everywhere mips is configured. Somewhat annoyingly [#t-compiler > mipsr6 targets &#96;target_arch&#96; confusion](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/mipsr6.20targets.20.60target_arch.60.20confusion/with/620228225).

src/new/glibc/mod.rs
```diff
@@ -40,6 +40,13 @@ mod sysdeps {
...
+pub(crate) use sysdeps::nptl::{
+    bits as nptl_bits,
+    pthread,
+};
 #[cfg(target_os = "linux")]
-pub(crate) use sysdeps::unix::linux::*;
+pub(crate) use sysdeps::unix::linux::{
+    bits as linux_bits,
+    net,
+};
```
tgross35: If bits isn't needed, can only `pthread` be reexported?

It might even be possible to just drop the `pub(crate)` from the `bits` modules, since they're not really needed at this level.

src/new/mod.rs
```diff
@@ -197,10 +197,15 @@ cfg_if! {
...
+
         #[cfg(target_env = "gnu")]
-        pub use net::route::*;
-        #[cfg(target_env = "gnu")]
-        pub use signal::*;
+        pub use self::{
+            linux_bits::types::siginfo_t::*,
+            net::route::*,
+            signal::*,
+        };
```
tgross35, line 208: The loose idea is that the modules reexported here should correspond to headers that are used. That is, `#include <linux/tls.h>` maps to `pub use linux::tls::*;`, and `#include <signal.h>` maps to `pub use signal::*`. So these probably need a `mod signal` somewhere that reexports from `bits`, since users don't import `sysdeps/linux/common/bits/signinfo.h` directly.

tgross35 (changes requested):
Sorry this has been on my todo list for a while, finally got through it

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5345/bd415df57bc97f3cba6299d2d392bbb65566f923..796c4ae07c104399595acb113cac2aa72f32db74)*

src/new/glibc/sysdeps/unix/linux/bits/types/siginfo_t.rs
dybucc: Done.

src/new/glibc/sysdeps/unix/linux/bits/types/siginfo_t.rs
dybucc: Done.

src/new/musl/signal.rs
dybucc: Done.

src/new/glibc/sysdeps/unix/linux/bits/types/siginfo_t.rs
dybucc: Done.

src/new/glibc/sysdeps/unix/linux/sparc/bits/siginfo_arch.rs
dybucc: Done.

src/new/mod.rs
dybucc: I think it's done now. I removed the reexports and instead used the
existing `signal` reexport by reexporting the `sigaction_t` module items
within the `signal` module.

src/new/glibc/mod.rs
dybucc: Sure. Done.

tgross35 (approved):
Looks great, thank you!

Happy to merge this now since it's an improvement, but would you be able to do a followup removing `siginfo_t` skips in `libc-test/build/main.rs`? I don't know that all can be removed but looks like possibly a couple.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5345/96b78673b6e396ee68c5161798faf0df11623290..627a3e498388a8457f9d9b32c036d9b8cc2fd763)*

dybucc:
>Happy to merge this now since it's an improvement, but would you be
>able to do a followup removing `siginfo_t` skips in
>`libc-test/build/main.rs`? I don't know that all can be removed but
>looks like possibly a couple.

I'll get to it once I'm done reviewing reviews.

PR #5348, opened 2026-07-31
apple: change `BPF_ALIGNMENT` to `usize`

sbogomolov:
# Description

The header defines `BPF_ALIGNMENT` as `sizeof(int32_t)`, which is `usize`, not `c_int`.

Breaking change for anyone matching on the current type, so no stable-nomination.

# Sources

Apple `BPF_ALIGNMENT` (`net/bpf.h`):
- https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/net/bpf.h#L126

# Checklist

- [x] Relevant tests in `libc-test/semver` have been updated (no new symbols)
- [x] Tested locally (`cd libc-test && cargo test --target mytarget`)

### DIFF
--- a/src/new/apple/xnu/net/bpf.rs
+++ b/src/new/apple/xnu/net/bpf.rs
@@ -47,7 +47,7 @@ pub const DLT_ATM_RFC1483: c_uint = 11; // LLC/SNAP encapsulated atm
 pub const DLT_RAW: c_uint = 12; // raw IP
 pub const DLT_LOOP: c_uint = 108;
 
-pub const BPF_ALIGNMENT: c_int = size_of::<i32>() as c_int;
+pub const BPF_ALIGNMENT: size_t = size_of::<i32>();
 
 pub const BIOCGRSIG: c_ulong = _IOR::<c_uint>('B' as c_ulong, 114);
 pub const BIOCSRSIG: c_ulong = _IOW::<c_uint>('B' as c_ulong, 115);


xtqqczze:
[CI / Test tier1 (x86_64-unknown-linux-gnu) (pull_request)](https://github.com/rust-lang/libc/actions/runs/30661200576/job/91257589127?pr=5348) failure tracked by https://github.com/rust-lang/libc/issues/5350.

xtqqczze:
I think stable nomination is appropriate here since this is a fix for an incorrect type definition rather than an intentional API change.

src/new/apple/xnu/net/bpf.rs
```diff
@@ -47,7 +47,7 @@ pub const DLT_ATM_RFC1483: c_uint = 11; // LLC/SNAP encapsulated atm
 pub const DLT_RAW: c_uint = 12; // raw IP
 pub const DLT_LOOP: c_uint = 108;
 
-pub const BPF_ALIGNMENT: c_int = size_of::<i32>() as c_int;
+pub const BPF_ALIGNMENT: usize = size_of::<i32>();
```
xtqqczze:
```suggestion
pub const BPF_ALIGNMENT: size_t = size_of::<i32>();
```

`size_t` is a type included in the prelude.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5348/56b4d3a1c8a9d9a5da7514d8c35cff9fd55f326c..5d73fadfd7a7fa09b98354333b4c336814295eaa)*

sbogomolov:
> I think stable nomination is appropriate here since this is a fix for an incorrect type definition rather than an intentional API change.

src/new/apple/xnu/net/bpf.rs
sbogomolov: Changed type to `size_t`.

xtqqczze (commented):
LGTM

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5348/56b4d3a1c8a9d9a5da7514d8c35cff9fd55f326c..8895eea9ce3015f57398b0cd252ba481a4c36be8)*

xtqqczze:
@sbogomolov you need to rebase to rerun CI

tgross35 (approved):
LGTM. I don't love backporting here - I'd call the fact that they use `sizeof` in the source more of an implementation detail than confirmation that the type should be `size_t`. But this will harmonize with FreeBSD and I doubt there are many users not already casting to `usize` to do something with alignment, so let's do it.

Cc @xingxue-ibm, AIX may want to change `BPF_ALIGNMENT` to `size_t` as well.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5348/1d33a6e674fa591f28dce52db6a6901b48b67ceb..dbf19604b7e049cb7467e240c668acb471cc03be)*

xtqqczze:
>  I'd call the fact that they use `sizeof` in the source more of an implementation detail than confirmation that the type should be `size_t`.

`bindgen` actually generates this as a `u32`:

```console
$ bindgen $(xcrun --show-sdk-path)/usr/include/net/bpf.h --clang-macro-fallback | grep BPF_ALIGNMENT
pub const BPF_ALIGNMENT: u32 = 4;
```

sbogomolov:
> `bindgen` actually generates this as a `u32`:
> 
> ```
> $ bindgen $(xcrun --show-sdk-path)/usr/include/net/bpf.h --clang-macro-fallback | grep BPF_ALIGNMENT
> pub const BPF_ALIGNMENT: u32 = 4;
> ```

As far as I understand, `u32` here is a value heuristic, not type information. For integer macro constants `bindgen` evaluates the value (which is `4`) and then picks a Rust type by magnitude using its default rule, which is the smallest of `u32`/`u64` that fits. In C the expression `sizeof(...)` will have type `size_t`.

xingxue-ibm:
> Cc @xingxue-ibm, AIX may want to change BPF_ALIGNMENT to size_t as well.

Thanks, @tgross35! Posted PR https://github.com/rust-lang/libc/pull/5373 for AIX.

Issue #5350, opened 2026-08-01
ctest: failures due to nightly-2026-07-31 cargo layout change


xtqqczze:
The cause appears to be the new build directory layout introduced by https://github.com/rust-lang/cargo/pull/17258.

[general > libc CI red after the 07/30 nightly cargo layout change @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/122651-general/topic/libc.20CI.20red.20after.20the.2007.2F30.20nightly.20cargo.20layout.20change/near/613934956)

cc: @weihanglo

tgross35:
Resolved by https://github.com/rust-lang/libc/pull/5347

PR #5351, opened 2026-08-01
crate: clean up leftovers from pre-rustc 1.19 without `union`s

dybucc:
## Description

This PR tries to address issue rust-lang/libc#1020. It cleans up a bunch of old ctest skips
over fields and types that were previously (more than six years ago) not proper
`union`s.

Note as well that `_channel_connect_attr` in QNX I couldn't check even though it
seems like it may be some left over `struct`-as-`union` type. I just coulnd't
get access to the SDK in the QNX site.

Then there's an unnamed field that is itself a `union` in Haiku targets under
`sigaction`. I've not changed that one because I don't know of a way of handling
a combination of both unnamed `union` and unnamed field.

The same applies to `cpu_topology_node_info` in Haiku targets.

For details on the rest of the changes, see the patch messages.

## Checklist

  doc comment

### DIFF
--- a/src/unix/haiku/mod.rs
+++ b/src/unix/haiku/mod.rs
@@ -368,12 +368,6 @@ s! {
         sa_userdata: *mut c_void,
     }
 
-    pub struct sem_t {
-        pub type_: i32,
-        pub named_sem_id: i32, // actually a union with unnamed_sem (i32)
-        padding: Padding<[i32; 2]>,
-    }
-
     pub struct ucred {
         pub pid: crate::pid_t,
         pub uid: crate::uid_t,
@@ -496,6 +490,19 @@ s! {
         pub ut_line: [c_char; 16],
         pub ut_host: [c_char; 128],
         __ut_reserved: Padding<[c_char; 64]>,
+    }
+}
+
+s_no_extra_traits! {
+    pub struct sem_t {
+        pub type_: i32,
+        pub named_sem_id: __c_anonymous_sem_t_u,
+        padding: Padding<[i32; 2]>,
+    }
+
+    pub union __c_anonymous_sem_t_u {
+        pub named_sem_id: i32,
+        pub unnamed_sem: i32,
     }
 }
 
--- a/src/unix/solarish/illumos.rs
+++ b/src/unix/solarish/illumos.rs
@@ -48,12 +48,6 @@ s! {
         pub fi_name: [c_char; crate::FILNAME_MAX as usize],
     }
 
-    #[cfg_attr(any(target_arch = "x86", target_arch = "x86_64"), repr(packed(4)))]
-    pub struct epoll_event {
-        pub events: u32,
-        pub u64: u64,
-    }
-
     pub struct utmpx {
         pub ut_user: [c_char; _UTX_USERSIZE],
         pub ut_id: [c_char; _UTX_IDSIZE],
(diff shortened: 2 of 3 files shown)


xtqqczze:
[CI / Test tier1 (x86_64-unknown-linux-gnu) (pull_request)](https://github.com/rust-lang/libc/actions/runs/30710449140/job/91396864310?pr=5351) failure tracked by https://github.com/rust-lang/libc/issues/5350.

src/unix/haiku/mod.rs
```diff
@@ -499,6 +493,37 @@ s! {
...
+
+        impl Eq for sem_t {}
+
+        impl core::hash::Hash for sem_t {
+            fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+                self.type_.hash(state);
+            }
+        }
+    }
+}
```
tgross35: Just drop the impls unless they're needed for building successfully

src/unix/solarish/illumos.rs
```diff
@@ -69,6 +63,39 @@ s! {
...
+
+        impl Eq for epoll_event {}
+
+        impl core::hash::Hash for epoll_event {
+            fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+                self.events.hash(state);
+            }
+        }
+    }
+}
```
tgross35: I would just drop these as well if unneeded

src/unix/solarish/illumos.rs
```diff
@@ -69,6 +63,39 @@ s! {
...
+        pub events: u32,
+        pub data: epoll_data,
+    }
+
+    pub union epoll_data {
+        pub ptr: *mut c_void,
+        pub fd: c_int,
+        pub u32: u32,
+        pub u64: u64,
+    }
```
tgross35, line 78: Illumos maintainers @jclulow @pfmooney this is a useful change but technically breaking, do you want it in 0.2?

tgross35 (changes requested):
Looks like a great cleanup, couple trait requests.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5351/56b4d3a1c8a9d9a5da7514d8c35cff9fd55f326c..078195b4f0e99eb89b760ef5989aa5d780cd7442)*

src/unix/haiku/mod.rs
dybucc: They are needed to build successfully when the `extra_traits` feature is
enabled.

Edit: Just realized what you mean; Sorry, fixed.

src/unix/solarish/illumos.rs
dybucc: Done.

tgross35 (approved):
LGTM, thanks!

I actually don't think we should have the epoll_event changes in 0.2 since we opted out of backporting the fix on Linux (1d10a3f). I'll exclude the relevant commits when picking.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5351/4ef1831337bab5d3663fdb328b03c6b247e897ec..0978aac4456e450ce005268490a465751c1fe211)*

src/unix/solarish/illumos.rs
tgross35: Deciding against this, see https://github.com/rust-lang/libc/pull/5351#pullrequestreview-4841980434

tgross35:
Excluding b0204f23b0ed52f8e778d62d898e4d4357de1f15 from the backport since that would need d25537932eac271, which isn't backported.

tgross35:
Similarly skipping 08fa94fc2bccbcaa01bcb5ab3ea2da58d787564 which needs 1d10a3f40d6f119f1fb44dda032f4248936f6a8c

tgross35:
3b624acda5807d3ebabe46359086066fe08f0ed5 as well, breaking change

dybucc:
Not sure if we can close #1020 already, or if we should wait until we
really near 1.0 because of the above breaking changes.

tgross35:
Good call, I think we can close that. As long as there aren't other union-related mismatches.

PR #5353, opened 2026-08-02
docs: wrap bare URLs in doc comments so `cargo doc` builds off Linux

samuraixwandering:
## Description

`RUSTDOCFLAGS: -Dwarnings` is set for the whole CI workflow, but the `docs` job runs on `ubuntu-26.04`, so it only ever documents the Linux modules. Fifteen doc comments under `src/new` carry bare URLs, which rustdoc reports as `this URL is not a hyperlink` — and every one of them is in an Apple or BSD tree the Linux job never compiles. The result is that `cargo doc --workspace --no-deps` fails outright on those hosts while CI stays green.

This wraps each URL in angle brackets, the form already used elsewhere in the same directories (for example `src/new/apple/xnu/net/mod.rs` and `src/new/freebsd/netinet6/mod.rs`). It also drops a stray trailing apostrophe from the URL in `src/new/freebsd/sys/mod.rs`.

Verified with `RUSTDOCFLAGS=-Dwarnings cargo doc --no-deps --target ...`, which fails on `main` and passes on this branch:

| target | `main` | this branch |
|---|---|---|
| `aarch64-apple-darwin` | 3 errors | ok |
| `x86_64-apple-darwin` | 3 errors | ok |
| `x86_64-unknown-freebsd` | 9 errors | ok |
| `x86_64-unknown-netbsd` | 1 error | ok |

The two remaining lines are in an OpenBSD module and an `i386`-gated one. Neither target has a std available locally so I could not verify them directly, but the edit is identical in form to the thirteen that are verified.

No API change, so there is nothing to add to `libc-test/semver`.

## Checklist

- [x] Relevant tests in `libc-test/semver` have been updated — n/a, no API change
- [x] Commit messages permalink to headers for added or changed API — n/a, no API change
- [x] Placeholder or unstable values like `*LAST` or `*MAX` have the standard doc comment — n/a
- [x] Tested locally (`cargo test -p libc-test --target aarch64-apple-darwin`) — passes, 7080 tests

### DIFF
--- a/src/new/apple/xnu/mod.rs
+++ b/src/new/apple/xnu/mod.rs
@@ -4,7 +4,7 @@
 
 /// Directory: `arm/`
 ///
-/// https://github.com/apple-oss-distributions/xnu/tree/main/bsd/arm
+/// <https://github.com/apple-oss-distributions/xnu/tree/main/bsd/arm>
 #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
 pub(crate) mod arm {
     pub(crate) mod _mcontext;
@@ -12,7 +12,7 @@ pub(crate) mod arm {
 
 /// Directory: `i386/`
 ///
-/// https://github.com/apple-oss-distributions/xnu/tree/main/bsd/i386
+/// <https://github.com/apple-oss-distributions/xnu/tree/main/bsd/i386>
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 pub(crate) mod i386 {
     pub(crate) mod _mcontext;
@@ -22,7 +22,7 @@ pub(crate) mod mach;
 
 /// Directory: `machine/`
 ///
-/// https://github.com/apple-oss-distributions/xnu/tree/main/bsd/machine
+/// <https://github.com/apple-oss-distributions/xnu/tree/main/bsd/machine>
 pub(crate) mod machine {
     pub(crate) mod _mcontext;
 }
--- a/src/new/apple/xnu/sys/mod.rs
+++ b/src/new/apple/xnu/sys/mod.rs
@@ -9,7 +9,7 @@ pub(crate) mod ttycom;
 
 /// Directory: `sys/_types`
 ///
-/// https://github.com/apple-oss-distributions/xnu/tree/main/bsd/sys/_types
+/// <https://github.com/apple-oss-distributions/xnu/tree/main/bsd/sys/_types>
 pub(crate) mod _types {
     pub(crate) mod _ucontext;
 }
--- a/src/new/freebsd/net/dlt.rs
+++ b/src/new/freebsd/net/dlt.rs
@@ -1,6 +1,6 @@
 //! Header: `net/dlt.h`
 //!
-//! https://github.com/freebsd/freebsd-src/blob/main/sys/net/dlt.h
+//! <https://github.com/freebsd/freebsd-src/blob/main/sys/net/dlt.h>
 
 use crate::prelude::*;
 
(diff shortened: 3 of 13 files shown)


xtqqczze:
[CI / Test tier1 (x86_64-unknown-linux-gnu) (pull_request)](https://github.com/rust-lang/libc/actions/runs/30731849277/job/91473405425?pr=5353) failure tracked by https://github.com/rust-lang/libc/issues/5350.

xtqqczze:
This does appear to fix every instance of the `rustdoc::bare_urls` lint. For example:

```log
warning: this URL is not a hyperlink
 --> src/new/apple/xnu/mod.rs:7:5
  |
7 | /// https://github.com/apple-oss-distributions/xnu/tree/main/bsd/arm
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: bare URLs are not automatically turned into clickable links
  = note: `#[warn(rustdoc::bare_urls)]` on by default
help: use an automatic link instead
  |
7 | /// <https://github.com/apple-oss-distributions/xnu/tree/main/bsd/arm>
  |     +                                                                +
```

That said, the `docs` job should ideally cover all of the documented targets, e.g.:

```sh
cargo metadata --no-deps --format-version 1 \
  | jq -r '.packages[].metadata.docs.rs?.targets[]?'
```

xtqqczze:
> The two remaining lines are in an OpenBSD module and an `i386`-gated one. Neither target has a std available locally so I could not verify them directly, but the edit is identical in form to the thirteen that are verified.

For reference, I verified the openbsd changes with:

```sh
cargo +nightly hack doc -q -Z build-std --ignore-private --workspace --no-deps --target x86_64-unknown-openbsd
```

So I suppose we could use something like this in CI:

```sh
cargo +nightly hack doc -Z build-std --ignore-private --workspace --no-deps \
    $(cargo metadata --no-deps --format-version 1 \
        | jq -r '.packages[].metadata.docs.rs?.targets[]?' \
        | xargs -I{} echo --target {})
```

although that may be impracticable for all 131 targets.

samuraixwandering:
Thanks for verifying the OpenBSD one, and for the pointer to #5350.

I should correct something in my description. I said two lines were unverified, but only OpenBSD actually was. The `i386` doc comment sits under `#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]`, so `x86_64-apple-darwin` does compile it, and that target was in the set I checked. On `main`:

```
RUSTDOCFLAGS="-Dwarnings" cargo doc --no-deps --target x86_64-apple-darwin
  src/new/apple/xnu/mod.rs:15    <- the i386 line
  src/new/apple/xnu/mod.rs:25
  src/new/apple/xnu/sys/mod.rs:12
```

and it passes on this branch. So it was 14 of 15 verified locally rather than 13, and with your OpenBSD run that covers all 15.

On widening the `docs` job, I think you are right that it should cover more than one target, otherwise this will drift back. I would rather do it as a separate PR though, so this one stays a lint fix with no CI changes. It also needs a decision that is worth its own discussion, since `-Z build-std` pins the job to nightly and 131 targets is likely too slow, so the docs.rs target list or some subset of it is probably where it lands. Happy to open that if there is interest.

xtqqczze:
One small clarification: `-Z build-std` is only necessary for certain targets, such as `aarch64-wrs-vxworks` (whose standard library isn't shipped with stable toolchains) and newer targets like `aarch64-unknown-qnx`, which are not yet available in the stable toolchain.

xtqqczze:
The docs.rs build logs are available here: https://docs.rs/crate/libc/latest/builds/3916860/aarch64-apple-darwin.txt

tgross35 (approved):
These are internal so it doesn't really matter, but also doesn't hurt. Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5353/56b4d3a1c8a9d9a5da7514d8c35cff9fd55f326c..07492247ffa6274d983d2193d32dd0725838ef23)*

PR #5358, opened 2026-08-03
apple: add `netinet/tcp.h` flag and option constants

samuraixwandering:
# Description

Socket options already exist for Apple (TCP_NODELAY, TCP_MAXSEG from the BSD module, TCP_KEEPALIVE from Apple) but the header-and-options side does not. This adds the TH_* flags (c_int because TH_AE is 0x100 and the composed masks exceed a byte), the TCPOPT_*/TCPOLEN_* pairs and the SACK limits. TCPOPT_MULTIPATH is left out as it is behind #if MPTCP and invisible to userspace. MAX_SACK_BLKS got the usage-guidelines comment; the other two limits are protocol-fixed.

# Sources

https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/netinet/tcp.h#L107-L181

# Checklist

- [x] Placeholder or unstable values like `*LAST` or `*MAX` have the standard doc comment
- [x] Tested locally (`cargo test -p libc-test --target aarch64-apple-darwin`)


src/new/apple/xnu/netinet/tcp.rs
```diff
@@ -0,0 +1,83 @@
...
+pub const TCPOPT_SACK: c_int = 5;
+pub const TCPOLEN_SACKHDR: c_int = 2;
+/// Length of one SACK block.
+pub const TCPOLEN_SACK: c_int = 8;
+pub const TCPOPT_TIMESTAMP: c_int = 8;
+pub const TCPOLEN_TIMESTAMP: c_int = 10;
+/// RFC 1323 appendix A.
+pub const TCPOLEN_TSTAMP_APPA: c_int = TCPOLEN_TIMESTAMP + 2;
+pub const TCPOPT_TSTAMP_HDR: c_int =
+    TCPOPT_NOP << 24 | TCPOPT_NOP << 16 | TCPOPT_TIMESTAMP << 8 | TCPOLEN_TIMESTAMP;
```
tgross35: Add parens around each `(x << y)`, makes it easier to read

src/new/apple/xnu/netinet/tcp.rs
```diff
@@ -0,0 +1,83 @@
...
+/// AccECN order 1.
+pub const TCPOPT_ACCECN1: c_int = 174;
+/// Empty option contains kind and length.
+pub const TCPOLEN_ACCECN_EMPTY: c_int = 2;
+/// Length of each AccECN counter.
+pub const TCPOLEN_ACCECN_COUNTER: c_int = 3;
+
+pub const TCPOPT_SACK_PERMIT_HDR: c_int =
+    TCPOPT_NOP << 24 | TCPOPT_NOP << 16 | TCPOPT_SACK_PERMITTED << 8 | TCPOLEN_SACK_PERMITTED;
+pub const TCPOPT_SACK_HDR: c_int = TCPOPT_NOP << 24 | TCPOPT_NOP << 16 | TCPOPT_SACK << 8;
```
tgross35: Ditto here

tgross35 (changes requested):
Diff looks pretty fine with some small tweaks, but please handwrite all PR descriptions and commit messages.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5358/4ef1831337bab5d3663fdb328b03c6b247e897ec..b70e76f5e84161f892e7d88f770aedefacab4b01)*

samuraixwandering:
Understood. Parens added, and the commit message and PR description are handwritten now.

@rustbot ready

src/new/apple/xnu/netinet/tcp.rs
```diff
@@ -0,0 +1,83 @@
...
+// `th_flags`, so the family is wider than a single octet.
+pub const TH_FIN: c_int = 0x01;
+pub const TH_SYN: c_int = 0x02;
+pub const TH_RST: c_int = 0x04;
+pub const TH_PUSH: c_int = 0x08;
+pub const TH_ACK: c_int = 0x10;
+pub const TH_URG: c_int = 0x20;
+pub const TH_ECE: c_int = 0x40;
+pub const TH_CWR: c_int = 0x80;
+pub const TH_AE: c_int = 0x100;
```
tgross35: Not sure I understand the comment; is it saying that you'd do something like `tcp.th_x2 = TH_AE` rather than `tcp.th_flags = TH_AE`? If so, I don't think that's worth noting.

tgross35: Or is it something weird, like that constant is intended to apply across fields somehow?

tgross35 (commented):
Thanks for updating, one last clarification

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5358/4ef1831337bab5d3663fdb328b03c6b247e897ec..cf8b68497dd06d17f2abe2f2f071255fdecff27d)*

src/new/apple/xnu/netinet/tcp.rs
samuraixwandering: > Or is it something weird, like that constant is intended to apply across fields somehow?

The second one. They are numbered as if `th_x2` and `th_flags` were a single field with `th_flags` as the low octet, so `TH_AE` is bit 8 and `TH_ACE` (`TH_AE|TH_CWR|TH_ECE`, 0x1c0) spans both. No field takes the constant as-is: `th_flags` is a byte, and the AE bit inside `th_x2` would be `TH_AE >> 8`. `TH_ACE` is the AccECN codepoint from RFC 9768.

That is also why the family is `c_int`. I will reword the comment to say that rather than repeat the header's "maps into th_x2". And will rebase onto main as well.

tgross35: Ah okay, so you're always expected to mask/shift `TH_ACE`? SGTM then

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5358/6b4c11aecf2eb45ce51b673e632d31b11c37bf81..dafd62c8087fcdf173d089b670f723490a059ca6)*

PR #5359, opened 2026-08-03
Add `pthread_gettid_np` for Android and glibc

valentynkit:
Link: https://github.com/bminor/glibc/blob/d2097651cc57834dbfcaa102ddfacae0d86cfb66/sysdeps/nptl/pthread.h#L1320-L1322
Link: https://android.googlesource.com/platform/bionic/+/731631f300090436d7f5df80d50b6275c8c60a93/libc/include/pthread.h#173

## Description

Returns the Linux TID for a given thread, unlike `gettid`, which only returns the id of the calling thread.

This is useful for std in the thread spawn path: the parent can set the TID of the child it just created, whereas today the TID can only be read once the child itself starts running.

This came up while working on rust-lang/rust#160219, which fills the TID when the child starts running and explains what's missing to fill it from the parent right after creation.

libc already carries a by-handle TID query for [Apple](https://github.com/rust-lang/libc/blob/4ef1831337bab5d3663fdb328b03c6b247e897ec/src/unix/bsd/apple/mod.rs#L4348)

pthread_gettid_np was added in glibc [2.42+](https://sourceware.org/bugzilla/show_bug.cgi?id=27880); on Android it has been available since API level 21.

## Checklist

  doc comment - not applicable

### DIFF
--- a/libc-test/build.rs
+++ b/libc-test/build.rs
@@ -5090,6 +5090,8 @@ fn test_linux(target: &str) {
             "posix_spawn_file_actions_addclosefrom_np" if gnu && sparc64 => true,
             // Needs glibc 2.35 or later.
             "posix_spawn_file_actions_addtcsetpgrp_np" if gnu && sparc64 => true,
+            // Needs glibc 2.42 or later.
+            "pthread_gettid_np" if gnu && versions.glibc.unwrap() < (2, 42) => true,
 
             // FIXME(linux): Deprecated since glibc 2.30. Remove fn once upstream does.
             "sysctl" if gnu => true,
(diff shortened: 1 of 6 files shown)


valentynkit:
cc: @tgross35

tgross35 (approved):
LGTM, please squash. Will just need a rebase to fix the docs job.

Also no need to ping, maintainers are subscribed already

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5359/4ef1831337bab5d3663fdb328b03c6b247e897ec..0fd66817fb4896ef6951a2bd809f365469ca3116)*

tgross35:
Thanks!

PR #5370, opened 2026-08-04
Clean up the `libc-test` build script

tgross35:
This file is quite large and doesn't have a lot of room for organization. Move to a directory where at least we can split up modules not related to `ctest` config.


tgross35:
Created lots of conflicts but automatic rebase should be able to figure out the move

PR #5372, opened 2026-08-05
macros: forward `cfg` attributes to generated impls

telcharr:
## Description

Follow-up to rust-lang/libc#5257 for the `cfg` issue that was spotted, plus the same on `Debug` for unions.

Attributes are now split into `cfg_attrs` and `other_attrs`, both go on the type, only the `cfg`s repeat on the impl. Needed them as `tt` rather than `meta` so `cfg` is still matchable.

The generated impls also warn when the type or its fields are deprecated. `s_no_extra_traits!` got an `allow(deprecated)` for that but the newer macros never picked it up.

The splitting arms are duplicated between `union_with_debug!` and `struct_with_default!`. Sharing them needs a callback macro which felt worse to read, but let me know if you'd rather have that. 

## Checklist

  doc comment

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -215,37 +215,26 @@ macro_rules! s_paren {
 /// Most structs will prefer to use [`s`].
 macro_rules! s_no_extra_traits {
     ($(
-        $(#[$attr:meta])*
+        $(#$attr:tt)*
         $pub:vis $t:ident $i:ident { $($field:tt)* }
     )*) => ($(
-        s_no_extra_traits!(it: $(#[$attr])* $pub $t $i { $($field)* });
+        s_no_extra_traits!(it: $(#$attr)* $pub $t $i { $($field)* });
     )*);
 
-    (it: $(#[$attr:meta])* $pub:vis union $i:ident { $($field:tt)* }) => (
-        #[repr(C)]
-        #[::core::prelude::v1::derive(
-            ::core::clone::Clone,
-            ::core::marker::Copy,
-        )]
-        $(#[$attr])*
-        $pub union $i { $($field)* }
-
-        #[allow(deprecated)]
-        impl ::core::fmt::Debug for $i {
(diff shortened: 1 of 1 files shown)


src/macros.rs
```diff
@@ -908,6 +1067,31 @@ mod tests {
         assert_eq!(s.x, 0);
         assert_eq!(unsafe { s.u.a }, 0);
     }
+
+    #[test]
+    fn s_with_default_keeps_struct_cfg() {
+        // The opposite of the configured-out types in `macro_checks`. With the `cfg` true the
+        // type and its `Default` both exist, and the other attributes still apply.
+        s_with_default! {
+            #[cfg(not(any()))]
```
tgross35: Can this and other tests use `cfg(false)`, or does the MSRV check yell at that? If so, it's nicer to read than `any()`/`not(any)`

tgross35 (commented):
Sorry this took so long, one change then looks great

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5372/e27c7f1e2b281d7c551e6b394b897b4ff0a230e6..98e9c48eabc34e9611f3becad1f6c6b637eec276)*

src/macros.rs
telcharr: Done, I don't see any errors related to this but it does look like CI is angry about something unrelated.

telcharr:
~~I'm fairly certain these CI issues are unrelated and I _think_ this should be good. But maybe worth waiting until CI is green?~~ 

Whatever that was seems to have been resolved!

And no problems at all on timing. Happy to be helping out!

@rustbot ready

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5372/e54d28e4415cd400d75c5c1d7e3455970533a9f0..f4fefb39fa851b236006268cb2531af6721f1682)*

PR #5376, opened 2026-08-06
musl: Rename `musl_v1_2_3` to `musl_v1_2`

### LINKED ISSUE #5200: Rename `musl_v1_2_3` and `RUST_LIBC_UNSTABLE_MUSL_V1_2_3` to `musl_v1_2` and `RUST_LIBC_UNSTABLE_MUSL_V1_2`
There has been a bit of confusion with these names because there aren't actually any breaks in the 1.2.3 vs. 1.2.0 that we need to be aware of. I think it would be good to replace any `v1_2_3` references with `v1_2` before we start to tell people these options are stable.

If this is done, we should still detect `RUST_LIBC_UNSTABLE_MUSL_V1_2_3` for a little while and emit a warning because I know some people are using it.

Cc @xbjfk since you're the original author here

tgross35:
1.2.0 was the release that introduced 64-bit `time_t` everywhere [1],
which is the main thing we are interested in with this config. There is
nothing special about 1.2.3, and most gated API will work down to 1.2.0.
This has been a source of mild confusion.

To make things more clear, drop the `_3`. We continue to support
`libc_unstable_musl_v1_2_3` for now but emit a warning if set.

Closes: https://github.com/rust-lang/libc/issues/5200

[1]: https://musl.libc.org/releases.html

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -198,34 +198,34 @@ jobs:
           - target: aarch64-linux-android
           - target: aarch64-unknown-linux-musl
           - target: aarch64-unknown-linux-musl
-            env: { TEST_MUSL_V1_2_3: 1 }
+            env: { TEST_MUSL_V1_2: 1 }
             artifact-tag: new-musl
           - target: arm-linux-androideabi
           - target: arm-unknown-linux-gnueabihf
           - target: arm-unknown-linux-musleabihf
           - target: arm-unknown-linux-musleabihf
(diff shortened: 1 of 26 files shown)


tgross35:
@xbjfk mind taking a look at this?

xbjfk:
Hey, sorry was not able to take a look earlier but LGTM

tgross35:
Thanks for taking a look when you could :)

PR #5378, opened 2026-08-06
wasi: add locale category constants

### LINKED ISSUE #5352: wasi: `setlocale` is declared but the `LC_*` category constants (`LC_CTYPE`, `LC_ALL`, etc.) are missing
`src/wasi/mod.rs` declares the `setlocale` extern function:

```rust
pub fn setlocale(category: c_int, locale: *const c_char) -> *mut c_char;
```

but does not define any of the `LC_CTYPE` / `LC_ALL` / `LC_COLLATE` / `LC_MONETARY` / `LC_NUMERIC` / `LC_TIME` / `LC_MESSAGES` constants that every other libc target (unix, windows, etc.) defines alongside `setlocale`. This makes the function nearly unusable — callers have no portable way to pass a valid `category` argument for the wasi target without hardcoding the raw integer themselves.

wasi-libc (the actual C library `rust-lang/libc`'s wasi target binds against) already defines these constants and they work correctly at runtime — this is a `rust-lang/libc` binding gap, not a wasi-libc or WASI capability gap. From wasi-libc's [`libc-top-half/musl/include/locale.h`](https://github.com/WebAssembly/wasi-libc/blob/main/libc-top-half/musl/include/locale.h):

```c
#define LC_CTYPE    0
#define LC_NUMERIC  1
#define LC_TIME     2
#define LC_COLLATE  3
#define LC_MONETARY 4
#define LC_MESSAGES 5
#define LC_ALL      6
```

I confirmed these values work as expected when manually declared and passed to `libc::setlocale`/`libc::tolower` under `wasm32-wasip1` via `wasmtime` 47.0.3 — `setlocale(0 /* LC_CTYPE */, "")` succeeds and returns a non-null pointer, and subsequent `tolower`/`toupper` calls behave normally.

This is user-visible: [uutils/coreutils](https://github.com/uutils/coreutils)'s `dd` implementation wants to call `libc::setlocale(libc::LC_CTYPE, c"".as_ptr())` before building its ASCII case-conversion tables (to match GNU `dd`'s locale-aware `conv=lcase`/`ucase` behavior), and has to `#[cfg(not(target_os = "wasi"))]`-gate that call out entirely because `libc::LC_CTYPE` doesn't exist for the wasi target — see [`conversion_tables.rs`](https://github.com/uutils/coreutils/blob/main/src/uu/dd/src/conversion_tables.rs). The workaround silently falls back to non-locale-aware ASCII conversion on WASI specifically, with no way to opt back into real locale support even though the underlying C library supports it fine.

Suggested fix: add the standard `LC_*` constant block (matching wasi-libc's musl-derived values above) to `src/wasi/mod.rs`, the same way it's done for e.g. `src/unix/linux_like/mod.rs`.

### Meta

`rustc --version --verbose`:
```
rustc 1.97.1 (8bab26f4f 2026-07-14)
```

subotac:
## Description

Adds the locale category constants defined by wasi-libc to the WASI bindings and semver list. This allows callers to use `setlocale` without hardcoding category values.

Fixes rust-lang/libc#5352.

Source: https://github.com/WebAssembly/wasi-libc/blob/6d8745c8cec1aaa82c24f2e3d7a544f7ea9c2089/libc-top-half/musl/include/locale.h#L23-L29

## Checklist

- [x] No placeholder or unstable values are added
- [x] Tested locally with `cargo test -p libc-test --target wasm32-wasip1`


JohnTitor (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5378/3a6ec284fcb676598ecf097243d56dafe0937393..3eddb865ea9f75e08b1d8dc0c99cb658a2f5fade)*

PR #5380, opened 2026-08-06
aix(powerpc64): clean up module

dybucc:
## Description

Stems from rust-lang/libc#5259. Is concerned only with the `aix/powerpc64` module. For details
on the changes, see the relevant patch messages.

## Checklist

  doc comment

### DIFF
--- a/src/unix/aix/powerpc64.rs
+++ b/src/unix/aix/powerpc64.rs
@@ -5,6 +5,8 @@ extern_ty! {
     pub type lock_data_instrumented;
 }
 
+pub type simple_lock_data = c_int;
+
 s! {
     pub struct sigset_t {
         pub ss_set: [c_ulong; 4],
@@ -289,6 +291,15 @@ impl siginfo_t {
 }
 
 s_no_extra_traits! {
+    pub union _simple_lock {
+        _slock: simple_lock_data,
+        _slockp: *mut lock_data_instrumented,
+    }
+
+    #[deprecated(
+        since = "0.2.187",
+        note = "Use `_simple_lock` instead. This type doesn't exist upstream."
+    )]
     pub union _kernel_simple_lock {
         pub _slock: c_long,
         pub _slockp: *mut lock_data_instrumented,
@@ -331,8 +342,8 @@ s_no_extra_traits! {
         pub f_dir_off: crate::off_t,
         // Should be pointer to 'cred'
         pub f_cred: *mut c_void,
-        pub f_lock: _kernel_simple_lock,
-        pub f_offset_lock: _kernel_simple_lock,
+        pub f_lock: _simple_lock,
+        pub f_offset_lock: _simple_lock,
         pub f_vinfo: crate::caddr_t,
         pub f_ops: *mut fileops_t,
         pub f_parentp: crate::caddr_t,
--- a/src/unix/aix/powerpc64.rs
+++ b/src/unix/aix/powerpc64.rs
@@ -331,6 +331,7 @@ s_no_extra_traits! {
         pub fo_fstat: Option<extern "C" fn(file: *mut file, sstat: *mut crate::stat) -> c_int>,
     }
 
+    #[repr(align(256))]
     pub struct file {
         pub f_flag: c_long,
         pub f_count: c_int,
--- a/src/unix/aix/powerpc64.rs
+++ b/src/unix/aix/powerpc64.rs
@@ -1,4 +1,3 @@
-use crate::off_t;
 use crate::prelude::*;
 
 // Define lock_data_instrumented as an empty enum
@@ -21,8 +20,8 @@ s! {
         pub l_sysid: c_uint,
         pub l_pid: crate::pid_t,
         pub l_vfs: c_int,
-        pub l_start: off_t,
-        pub l_len: off_t,
+        pub l_start: crate::off_t,
+        pub l_len: crate::off_t,
     }
 
     pub struct statvfs {
@@ -34,7 +33,7 @@ s! {
         pub f_files: crate::fsfilcnt_t,
         pub f_ffree: crate::fsfilcnt_t,
         pub f_favail: crate::fsfilcnt_t,
-        pub f_fsid: c_ulong,
+        pub f_fsid: crate::fsid_t,
         pub f_basetype: [c_char; 16],
         pub f_flag: c_ulong,
         pub f_namemax: c_ulong,
@@ -78,8 +77,8 @@ s! {
         pub st_type: c_uint,
         pub st_gen: c_uint,
         st_reserved: Padding<[c_uint; 9]>,
-        pub st_padto_ll: c_uint,
-        pub st_size: off_t,
+        st_padto_ll: Padding<c_uint>,
+        pub st_size: crate::off_t,
     }
 
     pub struct statfs {
@@ -107,7 +106,7 @@ s! {
         pub aio_lio_opcode: c_int,
         pub aio_fildes: c_int,
         pub aio_word1: c_int,
-        pub aio_offset: off_t,
+        pub aio_offset: crate::off_t,
         pub aio_buf: *mut c_void,
         pub aio_return: ssize_t,
         pub aio_errno: c_int,
@@ -126,11 +125,11 @@ s! {
     }
 
     pub struct __vmx_context_t {
-        pub __vr: [crate::__vmxreg_t; 32],
-        pub __pad1: [c_uint; 3],
+        pub __vr: [__vmxreg_t; 32],
+        __pad1: Padding<[c_uint; 3]>,
         pub __vscr: c_uint,
         pub __vrsave: c_uint,
-        pub __pad2: [c_uint; 3],
+        __pad2: Padding<[c_uint; 3]>,
     }
 
     pub struct __vsx_context_t {
@@ -138,8 +137,8 @@ s! {
     }
 
     pub struct __tm_context_t {
-        pub vmx: crate::__vmx_context_t,
-        pub vsx: crate::__vsx_context_t,
+        pub vmx: __vmx_context_t,
+        pub vsx: __vsx_context_t,
         pub gpr: [c_ulonglong; 32],
         pub lr: c_ulonglong,
         pub ctr: c_ulonglong,
@@ -158,7 +157,7 @@ s! {
         pub tmcontext: c_char,
         pub tmstate: c_char,
         pub prevowner: c_char,
-        pub pad: [c_char; 5],
+        pad: Padding<[c_char; 5]>,
     }
 
     pub struct __context64 {
@@ -176,7 +175,7 @@ s! {
         pub fpeu: c_char,
         pub fpinfo: c_char,
         pub fpscr24_31: c_char,
-        pub pad: [c_char; 1],
+        pad: Padding<[c_char; 1]>,
         pub excp_type: c_int,
     }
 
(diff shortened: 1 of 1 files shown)


src/unix/aix/powerpc64.rs
```diff
@@ -325,7 +330,13 @@ s_no_extra_traits! {
...
         >,
         pub fo_select: Option<
-            extern "C" fn(file: *mut file, a: c_int, b: *mut c_ushort, c: extern "C" fn()) -> c_int,
+            extern "C" fn(
+                file: *mut file,
+                a: c_int,
+                b: c_ushort,
+                c: *mut c_ushort,
+                c: extern "C" fn(),
+            ) -> c_int,
```
tgross35, line 328: "aix(powerpc64): deprecate `fileops_t` type": If you're updating the fields, it would be best to also mark the function pointers `unsafe`. Also the commit summary should mention there's more than just a deprecation.

src/unix/aix/powerpc64.rs
```diff
@@ -7,6 +7,13 @@ extern_ty! {
 
 pub type simple_lock_data = c_int;
 
+extern_ty! {
+    /// This is meant to be the `file` type upstream under `sys/ldr.h`. We
+    /// currently expose the kernel definition but that is slated for removal.
+    /// This opaque type will then be renamed to `file`.
+    pub type _file;
+}
```
tgross35: This is introduced in "aix(powerpc64): deprecate `file` type for `_file`" but used in a previous patch.

I think I'd almost say just replace `file` with an extern type now, rather than trying to migrate via deprecation. Somebody would have to be doing something really weird to wind up affected by the change, and we have T3 leeway.

tgross35 (commented):
Couple notes but of course we'll need to hear from @xingxue-ibm.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5380/74e58cc7294fa594b205a33f8235a2b264455b13..c6f59e4cd1d367b4341e6055b9ad3f0b7bc9f592)*

src/unix/aix/powerpc64.rs
```diff
@@ -255,24 +263,60 @@ s! {
...
+    #[deprecated(
+        since = "0.2.187",
+        note = "Use `_simple_lock` instead. This type doesn't exist upstream."
+    )]
     pub union _kernel_simple_lock {
         pub _slock: c_long,
         pub _slockp: *mut lock_data_instrumented,
     }
 
+    #[deprecated(
```
xingxue-ibm: The same as `struct file`. `_KERNEL` is intended for users implementing kernel extensions rather than the kernel itself.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5380/74e58cc7294fa594b205a33f8235a2b264455b13..c6f59e4cd1d367b4341e6055b9ad3f0b7bc9f592)*

src/unix/aix/powerpc64.rs
```diff
@@ -1,11 +1,19 @@
...
 use crate::prelude::*;
 
 // Define lock_data_instrumented as an empty enum
 extern_ty! {
     pub type lock_data_instrumented;
 }
 
+pub type simple_lock_data = c_int;
+
+extern_ty! {
```
xingxue-ibm: The definition of struct file is gated by the _KERNEL macro, which is intended for users implementing kernel extensions rather than the kernel itself. The libc crate also includes other items, such as functions gated by _KERNEL, that are used by popular crates. Since we do not rule out the possibility of users writing kernel extensions, and struct file is published in the sys/file.h header, I think it is reasonable to continue using it as is.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5380/74e58cc7294fa594b205a33f8235a2b264455b13..c6f59e4cd1d367b4341e6055b9ad3f0b7bc9f592)*

src/unix/aix/powerpc64.rs
```diff
@@ -293,50 +337,63 @@ s_no_extra_traits! {
...
+                b: c_ushort,
+                c: *mut c_ushort,
+                c: extern "C" fn(),
+            ) -> c_int,
         >,
         pub fo_close: Option<extern "C" fn(file: *mut file) -> c_int>,
         pub fo_fstat: Option<extern "C" fn(file: *mut file, sstat: *mut crate::stat) -> c_int>,
     }
 
+    #[deprecated(
```
xingxue-ibm: ditto

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5380/74e58cc7294fa594b205a33f8235a2b264455b13..c6f59e4cd1d367b4341e6055b9ad3f0b7bc9f592)*

src/unix/aix/powerpc64.rs
```diff
@@ -293,50 +337,63 @@ s_no_extra_traits! {
             ) -> c_int,
         >,
         pub fo_select: Option<
-            extern "C" fn(file: *mut file, a: c_int, b: *mut c_ushort, c: extern "C" fn()) -> c_int,
+            extern "C" fn(
+                file: *mut file,
+                a: c_int,
+                b: c_ushort,
+                c: *mut c_ushort,
+                c: extern "C" fn(),
```
xingxue-ibm:
```suggestion
                d: Option<unsafe extern "C" fn()>,

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5380/74e58cc7294fa594b205a33f8235a2b264455b13..c6f59e4cd1d367b4341e6055b9ad3f0b7bc9f592)*

File: src/unix/aix/powerpc64.rs (hunk 1 above)
Comment (dybucc) [reply src/unix/aix/powerpc64.rs:328]: Done. I also split that commit into more isolated patches.

File: src/unix/aix/powerpc64.rs (hunk 2 above)
Comment (dybucc) [reply src/unix/aix/powerpc64.rs]: Done.

File: src/unix/aix/powerpc64.rs (hunk 6 above)
Comment (dybucc) [reply src/unix/aix/powerpc64.rs]: Done.

File: src/unix/aix/powerpc64.rs (hunk 4 above)
Comment (dybucc) [reply src/unix/aix/powerpc64.rs]: So I guess `file` continues being exposed as a full type and not an opaque type?
Still, I may be missing something but it seems to me like the declaration in
`sys/file.h` does not declare a full type.

It's still opaque on the C side of things even when `_KERNEL` is `#define`d.

I did remove the deprecations, though. The only exception is
`_kernel_simple_lock`, for which I have left the `_simple_lock` type in its
stead.

Comment (xingxue-ibm) [reply src/unix/aix/powerpc64.rs]: > So I guess `file` continues being exposed as a full type and not an opaque type? 
>
I think so.
> Still, I may be missing something but it seems to me like the declaration in `sys/file.h` does not declare a full type.
> It's still opaque on the C side of things even when `_KERNEL` is `#define`d.
> 
Which AIX level is yours? I checked AIX 7.2 and 7.3, the definition in `sys/file.h` is as follows.
```
 40 #ifdef _KERNEL
 ...
 76 __align(256) struct file {
 77         long    f_flag;         /* see fcntl.h for definitions               */
 78         int     f_count;        /* reference count                           */
 79         short   f_options;      /* file flags not passed through vnode layer */
 80         short   f_type;         /* descriptor type                           */
 81
 82         /* The following fields are overloaded when the file struct is freed,
 83          * except for the locks which must remain static. See struct freefile.
 84          */
 85
 86         struct vnode *f_data;           /* pointer to vnode struct           */
 87 #define f_vnode       f_data
 88
 89         offset_t        f_offset;       /* read/write character pointer      */
 90         off_t           f_dir_off;      /* BSD style dir offsets             */
 91         struct ucred    *f_cred;        /* process cred at open              */
 92
 93         Simple_lock     f_lock;         /* file structure fields lock        */
 94         Simple_lock     f_offset_lock;  /* file structure offset field lock  */
 95
 96         caddr_t         f_vinfo;        /* any info vfs needs                */
 97         fileops_t       *f_ops;         /* operations vector                 */
 98
 99         caddr_t         f_parentp;      /* the file's parent vnode or fid    */
100         caddr_t         f_fnamep;       /* fname pointer, may be embedded    */
101         char            f_fdata[F_FDATASZ];/* embedded file data, pad to 256 */
102 };
...
```
> I did remove the deprecations, though. The only exception is `_kernel_simple_lock`, for which I have left the `_simple_lock` type in its stead.
>
That is fine.

tgross35: > The definition of struct file is gated by the _KERNEL macro, which is intended for users implementing kernel extensions rather than the kernel itself. The libc crate also includes other items, such as functions gated by _KERNEL, that are used by popular crates. Since we do not rule out the possibility of users writing kernel extensions, and struct file is published in the sys/file.h header, I think it is reasonable to continue using it as is.

To clarify, you mean that people are using `libc` to write kernel extensions? We don't really support that on other arches and have phased it out where anything accidentally got added, I think internal kernel API is out of `libc`'s scope.

xingxue-ibm: An AIX kernel extension is a loadable module that runs within the kernel address space and extends operating system functionality, such as device drivers and communication or I/O subsystems. It can consist of any routine added to the kernel and can be dynamically loaded or unloaded by a privileged user. System calls such as `sethostid`, `sethostname`, and `gethostname` are guarded by `_KERNEL` in header `net/proto_uipc.h`.

dybucc: >> So I guess `file` continues being exposed as a full type and not an opaque type? 
>>
>I think so.
>> Still, I may be missing something but it seems to me like the declaration in `sys/file.h` does not declare a full type.
>> It's still opaque on the C side of things even when `_KERNEL` is `#define`d.
>> 
>Which AIX level is yours? I checked AIX 7.2 and 7.3, the definition in `sys/file.h` is as follows.
>```
> 40 #ifdef _KERNEL
> ...
> 76 __align(256) struct file {
> 77         long    f_flag;         /* see fcntl.h for definitions               */
> 78         int     f_count;        /* reference count                           */
> 79         short   f_options;      /* file flags not passed through vnode layer */
> 80         short   f_type;         /* descriptor type                           */
> 81
> 82         /* The following fields are overloaded when the file struct is freed,
> 83          * except for the locks which must remain static. See struct freefile.
> 84          */
> 85
> 86         struct vnode *f_data;           /* pointer to vnode struct           */
> 87 #define f_vnode       f_data
> 88
> 89         offset_t        f_offset;       /* read/write character pointer      */
> 90         off_t           f_dir_off;      /* BSD style dir offsets             */
> 91         struct ucred    *f_cred;        /* process cred at open              */
> 92
> 93         Simple_lock     f_lock;         /* file structure fields lock        */
> 94         Simple_lock     f_offset_lock;  /* file structure offset field lock  */
> 95
> 96         caddr_t         f_vinfo;        /* any info vfs needs                */
> 97         fileops_t       *f_ops;         /* operations vector                 */
> 98
> 99         caddr_t         f_parentp;      /* the file's parent vnode or fid    */
>100         caddr_t         f_fnamep;       /* fname pointer, may be embedded    */
>101         char            f_fdata[F_FDATASZ];/* embedded file data, pad to 256 */
>102 };
>...
>```
>> I did remove the deprecations, though. The only exception is `_kernel_simple_lock`, for which I have left the `_simple_lock` type in its stead.
>>
>That is fine.

You're right. I double-checked and noticed I had been looking at the
`extern` declaration further down below. I cross-reference with AIX
7.3.

The patch making `file` opaque now simply forces the right alignment
requirement.

xingxue-ibm (approved):
LGTM, thank you so much for cleaning this up!

Changes to the CI build and `semver` tests are needed for the CI checks to pass. I will submit a follow-up PR after this patch lands.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5380/72f013506e8e58e4e4f8d3c506c2028037db6543..00d56b213d801b25d3a47378c5343fcc9318f7a6)*

tgross35 (approved):
Thanks for the changes and @xingxue-ibm for reviewing!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5380/72f013506e8e58e4e4f8d3c506c2028037db6543..00d56b213d801b25d3a47378c5343fcc9318f7a6)*

xingxue-ibm:
Thank you, @tgross35!

PR #5384, opened 2026-08-07
android: stop skipping `ioctl` in tests

dybucc:
## Description

Removes a skip in ctest for `ioctl` in Android targets. This was previously not
possible because there were issues with an overload exposed in Bionic libc. This
overload, though, is only conditionally exposed.

The macro used to skip the definition upstream has been added to the `#define`s
that we have set up in `test_android` in libc-test's build script.

## Checklist

  doc comment

### DIFF
--- a/libc-test/build/main.rs
+++ b/libc-test/build/main.rs
@@ -1983,6 +1983,10 @@ fn test_android(target: &str) {
     let mut cfg = ctest_cfg();
     cfg.define("_GNU_SOURCE", None);
 
+    // This ensure we avoid the `ioctl` overload for request parameters that use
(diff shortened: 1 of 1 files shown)


tgross35 (approved):
Cc @maurer but I don't expect this to be problematic, the linked header documents this define as a way to get rid of the override. So LGTM

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5384/b37a8dec5c988a0ef09595240f873cd2bd2b9ac7..8dc4d35ddc178c8f3c534ad6f31ca2f359166373)*

PR #5385, opened 2026-08-07
ctest: allow variadics to be tested.

### LINKED ISSUE #4367: ctest: support varidics
ctest should be able to generate tests for functions that make use of varidics. See https://github.com/JohnTitor/ctest2/issues/58 for context.

mbyx:
## Description
Removes the check for variadics and adds support for translating variadic types between C and Rust.

While not used currently, might be useful down the line, especially if we ever decide to stop using void pointers for foreign function tests.

Closes rust-lang/libc#4367
## Checklist

- [X] Commit messages permalink to headers for added or changed API
- [X] Placeholder or unstable values like `*LAST` or `*MAX` have the standard
  doc comment
- [X] Tested locally (`cargo test -p libc-test --target mytarget`);

### DIFF
--- a/ctest/src/cdecl.rs
+++ b/ctest/src/cdecl.rs
@@ -254,6 +254,11 @@ pub(crate) fn func_ptr(args: Vec<CTy>, ret: CTy) -> CTy {
     }
 }
 
+/// Create a variadic function argument.
+pub(crate) fn variadic() -> CTy {
+    named("...", Constness::Mut)
+}
+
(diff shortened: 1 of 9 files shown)


tgross35 (approved):
Awesome! Thank you

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5385/bcc2a055f981b8684d2b7c07d537897ff83b399c..2d467a25a3a54105e603c0568803107aff5f9169)*

PR #5390, opened 2026-08-08
macros: add support for `exhaustive` attribute

dybucc:
## Description

Attempts to take the first steps in closing rust-lang/libc#4080. This should
allow us to add a private field for each of the records declared within
`s_with_default` or `s_no_extra_traits_with_default`.

This should in turn enforce those records to be initialized field-by-field in
downstream crates, while avoiding the lint that currently pops up when using a
`struct` marked `non_exhaustive` in FFI contexts.

## Checklist

      doc comment
- [x] Tested locally (`cargo test -p libc-test --target mytarget`); especially

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1167,6 +1167,56 @@ mod tests {
     }
 
     #[test]
+    fn s_with_default_is_non_exhaustive() {
+        // Without `#[exhaustive]`, the record should have an additional field
+        // added at the end. If this test compiles, it has it.
+        s_with_default! {
+            struct Something {
+                a: u32,
+            }
+        }
+
+        let s = Something::default();
+        assert_eq!(s.__non_exhaustive, ());
+    }
+
+    #[test]
+    fn s_with_default_uses_exhaustive() {
+        // With `#[exhaustive]`, the record should be regurgitated as-is. If
+        // this test compiles, then it works.
+        s_with_default! {
+            #[exhaustive]
+            struct Something {
+                a: u32,
+            }
+        }
+
+        #[allow(unused)]
+        let s = Something {
+            a: Default::default(),
+        };
+    }
+
+    #[test]
+    fn s_with_default_uses_mixed_exhaustive() {
+        // `#[exhaustive]` should work when sandwiched between attributes. If
+        // the test compiles, then it works.
+        s_with_default! {
+            #[repr(align(8))]
+            #[exhaustive]
+            #[repr(align(2))]
+            struct Something {
+                a: u32,
+            }
+        }
+
+        #[allow(unused)]
+        let s = Something {
+            a: Default::default(),
+        };
+    }
+
+    #[test]
     fn s_with_default_uses_custom_default() {
         // A non-default value proves `custom_default` is used rather than a derived default.
         s_with_default! {
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1,11 +1,10 @@
 /// A macro for defining #[cfg] if-else statements.
 ///
-/// This is similar to the `if/elif` C preprocessor macro by allowing definition
-/// of a cascade of `#[cfg]` cases, emitting the implementation which matches
-/// first.
+/// This is similar to the `if/elif` C preprocessor macro by allowing definition of a cascade of
+/// `#[cfg]` cases, emitting the implementation which matches first.
 ///
-/// This allows you to conveniently provide a long list #[cfg]'d blocks of code
-/// without having to rewrite each clause multiple times.
+/// This allows you to conveniently provide a long list #[cfg]'d blocks of code without having to
+/// rewrite each clause multiple times.
 macro_rules! cfg_if {
     // match if/else chains with a final `else`
     ($(
@@ -162,8 +161,8 @@ macro_rules! prelude {
 ///
 /// Also mark the type with `repr(C)`.
 ///
-/// Use [`s_no_extra_traits`] for structs where the `extra_traits` feature does not
-/// make sense, and for unions.
+/// Use [`s_no_extra_traits`] for structs where the `extra_traits` feature does not make sense, and
+/// for unions.
 macro_rules! s {
     ($(
         $(#[$attr:meta])*
@@ -197,8 +196,8 @@ macro_rules! s {
     );
 }
 
-/// Implement `Clone`, `Copy`, and `Debug` for a tuple struct, as well as `PartialEq`, `Eq`,
-/// and `Hash` if the `extra_traits` feature is enabled.
+/// Implement `Clone`, `Copy`, and `Debug` for a tuple struct, as well as `PartialEq`, `Eq`, and
+/// `Hash` if the `extra_traits` feature is enabled.
 ///
 /// Unlike `s!`, this does *not* mark the type with `repr(C)`. Users should provide their own
 /// `repr` attribute via `$attr` as necessary.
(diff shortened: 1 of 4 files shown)


tgross35:
Could you base this on top of https://github.com/rust-lang/libc/pull/5372? It looks like there may be some overlap. Feel free to leave a review there as well, if you have any suggestions.

src/macros.rs
```diff
@@ -424,18 +424,112 @@ macro_rules! struct_with_default {
...
+
+    // found `exhaustive`; parameter `found` is not matched against `false` in
+    // case somebody annotates a `struct` with more than one `exhaustive`. If we
+    // didn't match a wildcard, that attribute would go down to the general case
+    // of the muncher and add it to the packed matcher of `processed_attrs`.
+    (
+        @attr
+        attrs: { #[exhaustive] $($attrs:tt)* }
+        processed_attrs: { $($prev_attrs:tt)* }
+        found: { $_:tt }
```
tgross35: `found_exhaustive_attr` or so, to make it more clear in other blocks without the comment

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5390/bcc2a055f981b8684d2b7c07d537897ff83b399c..f11b35278539933fd8b842388526a63e84492532)*

src/macros.rs
```diff
@@ -424,18 +424,112 @@ macro_rules! struct_with_default {
...
+    };
+
+    (false decl: $(#[$attr:meta])* $vis:vis $name:ident { $($field:tt)* }) => {
+        $(#[$attr])*
+        $vis struct $name { $($field)* __non_exhaustive: () }
+    };
+    (false default: { $($field_default:tt)* }) => {
+        Self { $($field_default)* __non_exhaustive: () }
+    };
 }
```
tgross35: Give this a more descriptive name, and prefer the `field: $value` pattern for matchers like the other macros have. That's easier to follow than needing to figure out what different token sequences indicate.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5390/bcc2a055f981b8684d2b7c07d537897ff83b399c..f11b35278539933fd8b842388526a63e84492532)*

src/lib.rs
```diff
@@ -152,6 +152,9 @@
 // Not all macros and all patterns are used on all targets.
 #![allow(unused_macros)]
 #![allow(unused_macro_rules)]
+// FIXME(rust-lang/rust#132699): get rid of this once we stop using the private
+// field.
+#![allow(clippy::manual_non_exhaustive)]
```
tgross35: Can this be scoped to only where needed?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5390/bcc2a055f981b8684d2b7c07d537897ff83b399c..f11b35278539933fd8b842388526a63e84492532)*

src/macros.rs
dybucc: Done.

src/lib.rs
dybucc: Sure it can. Done.

src/macros.rs
dybucc: Done.

dybucc:
I think it's done. I've created a new patchset based off of #5372 [^1],
but I haven't yet dropped the commits in this patch. Once that PR
merges, I'll drop the ones in this patch and cherry pick the ones from
the above patch.

[^1]: <https://github.com/dybucc/libc/commits/feature/default-macro-cfg/>

@rustbot ready

tgross35 (changes requested):
@rustbot author

Needs the updates after https://github.com/rust-lang/libc/issues/4080. Would you mind also renaming `s_with_default!` to `s2!` and similar for `s_no_extra_traits_with_default`? Since with this PR they'll be doing more than just adding `Default`.

Please also include this in their doc comments.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5390/bd415df57bc97f3cba6299d2d392bbb65566f923..f596ba741e492b895e456b02d8a3969ce1b5f0d7)*

src/macros.rs
```diff
@@ -405,24 +405,34 @@ macro_rules! union_with_debug {
...
+/// Attributes are split into `cfg_attrs` and `other_attrs` before the fields
+/// are scanned. Both go on the struct, but only the `cfg`s are repeated on the
+/// `Default` impl. A `cfg` decides whether the type exists at all, so without
+/// it a configured-out struct leaves an impl behind referring to a type that
+/// isn't there.
+///
+/// Out of `other_attrs`, we scan for `#[exhaustive]`. If found, we remove it
+/// but take into account that the record should be expanded _without_ an
+/// additional private field. The scan for both `cfg` attributes and the
+/// (non-existent) `exhaustive` attribute is done in one linear pass.
```
tgross35: Looks like some autoformatting got you, seems like this was rewrapped from 100 to 80.

src/new/helenos/time.rs
tgross35, line 15: Could you mark this one `#[exhaustive]`? `timespec` shouldn't ever get new fields.

src/new/linux_uapi/linux/can.rs
tgross35, line 1: Same thing for the structs in this file that don't have private fields. We can probably remove them at some point, feel free to add a FIXME, but we should keep the breakage smaller.

Do this in the first commit (adding non-exhaustive/exhaustive) to avoid changing this in one patch then restoring behavior after.

src/macros.rs
```diff
@@ -433,6 +443,7 @@ macro_rules! struct_with_default {
             cfg_attrs: { }
             other_attrs: { }
             remaining_attrs: { $($attrs)* $(#$attr)* }
+            found_exhaustive: { false }
```
tgross35: Nit: either call this `found_exhaustive_attr`, or `needs_non_exhaustive_field` and flip the logic

src/macros.rs
```diff
@@ -575,30 +623,80 @@ macro_rules! struct_with_default {
...
+        $(#[$attr])*
+        $vis struct $name { $($field)* }
+    };
+    (
+        found_exhaustive: { true },
+        expansion: { default_impl },
+        body: { $($field_default:tt)* }
+    ) => {
+        Self { $($field_default)* }
+    };
```
tgross35, line 745: Nit: match simple statements without the braces so it doesn't look like something is getting grouped

```rust
found_exhaustive: true,
expansion: default_impl,
```

src/macros.rs
```diff
@@ -575,30 +623,80 @@ macro_rules! struct_with_default {
...
+            // initializer too. Docs are harmless there but trip the lint, so
+            // silence it.
             #[allow(unused_doc_comments)]
             fn default() -> Self {
-                Self { $($processed_field_defaults)* }
+                finalize_exhaustiveness! {
+                    found_exhaustive: { $found_exhaustive },
+                    expansion: { default_impl },
+                    body: { $($processed_field_defaults)* }
+                }
```
tgross35: Rather than use the same `finalize_exhaustiveness` macro for the expansion and default impl, split into to macros like `emit_struct_definition` and `emit_struct_default_body` to make things a bit easier to follow.

tgross35 (changes requested):
Looks pretty great to me! Nothing major here

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5390/96b78673b6e396ee68c5161798faf0df11623290..d492384f31e961eead4c921e190083e54ed3e45d)*

src/macros.rs
dybucc: Nay, this one was fully intentional. I decided to go for this one
because I saw some other areas of libc that don't fully comply with the
80 columns for comments, and 100 columns for doc comments style guide.

tgross35: What style guide is that? If it suggests doc comments are 100 then this moves away from that.

In any case, I'd prefer to leave as-is.

dybucc: Done.

src/new/helenos/time.rs
dybucc: Done.

src/new/linux_uapi/linux/can.rs
dybucc: Done.

src/macros.rs
dybucc: Done.

src/macros.rs
dybucc: Done.

src/macros.rs
dybucc: Done.

tgross35 (approved):
Looks great, thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5390/df6561676506cd9c36ebe84b3b2f000661b64767..9754699fd92fa38acec757269a1346a4a7c99c5f)*

PR #5393, opened 2026-08-10
ctest: Implement `Clone` for `TestGenerator`

tgross35:
We can do this pretty easily by replacing `Box` with `Rc` for the `dyn Fn` items.


tgross35:
Added to make it easier to run multiple `ctest` instances with the same setup.

Cc @mbyx

PR #5404, opened 2026-08-11
deprecate the `<int type>_{MIN,MAX}` constants

telcharr:
Follow-up to rust-lang/libc#5166. Deprecates the `<int type>_{MIN,MAX}` constants across the platform modules, with notes pointing at the type's own `MIN`/`MAX`.

### DIFF
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -1184,9 +1184,9 @@ cfg_if! {
 pub const HOST_NAME_MAX: c_int = 255;
 
 #[deprecated(since = "0.2.190", note = "Use `c_int::MIN` instead.")]
-pub const INT_MIN: c_int = -1 - 0x7fffffff;
+pub const INT_MIN: c_int = c_int::MIN;
 #[deprecated(since = "0.2.190", note = "Use `c_int::MAX` instead.")]
-pub const INT_MAX: c_int = 0x7fffffff;
+pub const INT_MAX: c_int = c_int::MAX;
 
 pub const SIG_DFL: sighandler_t = 0 as sighandler_t;
 pub const SIG_IGN: sighandler_t = 1 as sighandler_t;
--- a/src/fuchsia/mod.rs
+++ b/src/fuchsia/mod.rs
@@ -1183,7 +1183,9 @@ cfg_if! {
 
 pub const HOST_NAME_MAX: c_int = 255;
 
+#[deprecated(since = "0.2.190", note = "Use `c_int::MIN` instead.")]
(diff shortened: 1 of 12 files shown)


src/fuchsia/mod.rs
```diff
@@ -1183,7 +1183,9 @@ cfg_if! {
 
 pub const HOST_NAME_MAX: c_int = 255;
 
+#[deprecated(since = "0.2.190", note = "Use `c_int::MIN` instead.")]
 pub const INT_MIN: c_int = -1 - 0x7fffffff;
+#[deprecated(since = "0.2.190", note = "Use `c_int::MAX` instead.")]
 pub const INT_MAX: c_int = 0x7fffffff;
```
tgross35: As part of this, would you mind changing the values to what we're suggesting to replace? E.g. `pub const INT_MIN: c_int = c_int::MAX;`

That way tests should pick up any cases where they don't match up.

tgross35 (changes requested):
One requested update for the constants, assuming things check out after that then LGTM

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5404/0f136e82b9de3a35a430439d598bf6b53b88b059..3fba53b3bfdab2e4d83d83dc965c802d8d01455d)*

src/fuchsia/mod.rs
telcharr: Done. I did move some values on 64-bit hurd, which I think were just wrong before. The `*_FAST16/32` ones are `c_long/ulong` on that target and the pointer-width ones are `isize`/`usize` but all were written with 32-bit literals. One thing I'm a bit less sure on is that `INTPTR_MIN/MAX` and `PTRDIFF_MIN/MAX` needed their declared type moved to `intptr_t`/`ptrdiff_t` for this to compile since `__intptr_t` is `c_long` and won't unify with `isize`. If you'd rather I left those two alone let me know.

telcharr:
Couldn't find qurt, solid, switch, xous, or 64-bit hurd in the CI anywhere, so I built them locally with `-Zbuild-std=core`. All came back clean, apart from qurt at first, where `core` itself wouldn't build for hexagon on 1.97.1. I was able to get it to build on 1.99.0-nightly though. Same error as https://github.com/rust-lang/stdarch/issues/2134.

@rustbot ready

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5404/0f136e82b9de3a35a430439d598bf6b53b88b059..57412b6f3251e47de5f7f837a7c3ab4c76c2d177)*

PR #5400, opened 2026-08-11
feat: add missing lch* syscalls to *BSD platforms

ngie-eign:
This change adds the following system calls to various 4.4BSD derivatives:
 
* FreeBSD(-like)
    * lchmod ([DragonFlyBSD][1], [FreeBSD][2])
* macOS
    * [lchflags][3]
    * [lchmod][4]
* NetBSD
    * [lchmod][5]
 
These system calls are available on all ancient versions of *BSD (4.4BSD, FreeBSD 3.0, etc). More specific historical information about when the functions were added is available in the manpages and headers referenced.
 
1: https://leaf.dragonflybsd.org/cgi/web-man?command=lchmod
2: https://man.freebsd.org/cgi/man.cgi?query=lchmod
3: https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/stat.h#L597
4: https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/stat.h#L598
5: https://man.netbsd.org/lchmod.2

PR #5405, opened 2026-08-12
backport: feat: add missing lch* syscalls to *BSD platforms

ngie-eign:
_(This is a back port test for rust-lang/libc#5400)_

This change adds the following system calls to various 4.4BSD derivatives:

* FreeBSD(-like)
    * lchmod ([DragonFlyBSD][1], [FreeBSD][2])
* macOS
    * [lchflags][3]
    * [lchmod][4]
* NetBSD
    * [lchmod][5]

These system calls are available on all ancient versions of *BSD (4.4BSD, FreeBSD 3.0, etc). More specific historical information about when the functions were added is available in the manpages and headers referenced.

This change also adds *chflags and *chmod to macos' semver file (additional syscalls are missing from that file).

1: https://leaf.dragonflybsd.org/cgi/web-man?command=lchmod
2: https://man.freebsd.org/cgi/man.cgi?query=lchmod
3: https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/stat.h#L597
4: https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/stat.h#L598
5: https://man.netbsd.org/lchmod.2

(cherry picked from commit 5eac8d9a34aa0f3ca0990c2abd80fbe4a3f0b388)

### DIFF
--- a/libc-test/semver/apple.txt
+++ b/libc-test/semver/apple.txt
@@ -2148,6 +2148,8 @@ key_t
 killpg
 kqueue
 labs
+lchflags
+lchmod
 lio_listio
 listxattr
 load_command
--- a/libc-test/semver/freebsd.txt
+++ b/libc-test/semver/freebsd.txt
@@ -2117,6 +2117,7 @@ kld_load
 kqueue
 labs
 lchflags
+lchmod
 lcong48
 lio_listio
 lockf
--- a/libc-test/semver/netbsd.txt
+++ b/libc-test/semver/netbsd.txt
@@ -1410,6 +1410,7 @@ labs
 lastlog
 lastlogx
 lchflags
+lchmod
 lcong48
 lgetxattr
 lio_listio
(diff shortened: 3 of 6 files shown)


tgross35:
Thanks for putting this together but there's no need, I do backports in a batch prior to a release

ngie-eign:
> Thanks for putting this together but there's no need, I do backports in a batch prior to a release

I was just confirming that it worked before adding the comment -- but good to know :)!

tgross35:
Ah sorry, didn't mean to delete your tests! It may have failed at some point anyway since this branch doesn't get the little needed CI fixes that come up (until it's release time).

Also, I have no idea whether you'd be interested but if you're involved in FreeBSD and working on Rust things, it wouldn't hurt to list yourself as a target maintainer. No pressure or commitment, just means you might get pinged for FreeBSD-specific questions. (The current maintainers have things under control and Alan is very involved, but more people never hurts.)

Just a PR to https://github.com/rust-lang/rust/blob/6660e286ea36304054d6ca5264f5392513622b66/src/doc/rustc/src/platform-support/freebsd.md if so.

PR #5411, opened 2026-08-14
crate: expose `cfg` for general 64-bit `time_t` functionality

dybucc:
## Description

This patch adds support for a new `cfg` for easily toggling support for
64-bit `time_t` in supported platforms. This really only sets up the any
one of `linux_time_bits64`, `gnu_time_bits64` or `uclibc_time64`.

The name of the `cfg` is `time64`.

Note support for the equivalent flag in musl has not been included
because we already have set-up automatic detection and toggling of the
corresponding `cfg` under supported targets [^1].

[^1]:
    <https://github.com/rust-lang/libc/blob/1a8e71f33b1d6ea1e072210e7fc994417bbb2e34/build.rs#L181-L190>

## Checklist

  standard doc comment


build.rs
```diff
@@ -190,7 +196,7 @@ fn main() {
     }
```
tgross35: This should also set the `musl_v1_2` flag, since that's pretty much all it's gating

build.rs
```diff
@@ -40,6 +40,9 @@ const ALLOWED_CFGS: &[&str] = &[
     "musl_redir_time64",
     "vxworks_lt_25_09",
     "libc_pauthtest",
+    // Global flag to enable one of `linux_time_bits64` or `gnu_time_bits64`.
+    // The musl flags are enabled by default on supported platforms.
+    "time64",
```
tgross35: I think this isn't actually needed, this list is only for what gets sent to `check-cfg` and we won't use `time64` directly (at least for now)

tgross35 (changes requested):
Could you also update CI to use this cfg rather than the current glibc/musl cfg? Since this is what we're most likely to ship, we should make sure it works.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5411/1a8e71f33b1d6ea1e072210e7fc994417bbb2e34..98e48b9655207c40e511c394f9827a6be1f6e688)*

build.rs
dybucc: I don't think I understand. You mention the `musl_v1_2` flag, but that
is already set in that particular code block. Further, you also mention
in another review comment that I shouldn't need the `time64` flag, so I
don't think there's anything to change in this particular area.

build.rs
dybucc: Done.

dybucc:
I think it's done now. While checking through the CI workflow file, I
noticed that we even though we set the "updated but deprecated"
`RUST_LIBC_UNSTABLE_MUSL_V1_2` environment variable for certain targets
[^1], we still check for `RUSTC_LIBC_UNSTABLE_MUSL_V1_2_3` in `build.rs`
[^2].

I was wondering what's our stance on that change for stable releases,
and whether we should even set the deprecated environment variable in
CI. It seems wrong.

[^1]: <https://github.com/rust-lang/libc/blob/75b2850150d0c50fa012ce965b7466d46da8e4e3/.github/workflows/ci.yaml#L223>
[^2]: <https://github.com/rust-lang/libc/blob/75b2850150d0c50fa012ce965b7466d46da8e4e3/build.rs#L162-L164>

@rustbot ready

.github/workflows/ci.yaml
```diff
@@ -386,34 +386,34 @@ jobs:
...
+          msrv="$(
+            cargo metadata --format-version 1 |
+            jq -r --arg CRATE_NAME ctest '.packages | map(select((.name == $CRATE_NAME) and (.id | startswith("path+file")))) | first | .rust_version'
+          )"
+          echo "MSRV: $msrv"
+          echo "MSRV=$msrv" >> "$GITHUB_ENV"
+      - name: Install Rust
+        run: rustup update "$MSRV" --no-self-update && rustup default "$MSRV"
+      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
+      - run: cargo build -p ctest
```
tgross35, line 402: It's fine to reformat this if it makes things more consistent, but please put it in a separate commit

ci/verify-build.py
```diff
@@ -409,13 +409,10 @@ def test_target(cfg: Cfg, target: Target) -> TargetResult:
...
     run([*cmd, "--features=extra_traits"], rustflags=rustflags)
 
-    if "gnu" in target_env and target_bits == "32":
-        # Equivalent of _TIME_BITS=64
-        run(cmd, rustflags=f'{rustflags} --cfg=libc_unstable_gnu_time_bits="64"')
-
-    if "musl" in target_env:
-        # Check with breaking changes from musl, including 64-bit time_t on 32-bit
-        run(cmd, rustflags=f"{rustflags} --cfg=libc_unstable_musl_v1_2")
+    if "gnu" in target_env and target_bits == "32" or "musl" in target_env:
```
tgross35: Use parens to make the order of operations clear

tgross35 (changes requested):
Two small things then LGTM

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5411/bd415df57bc97f3cba6299d2d392bbb65566f923..36137bcbf879745072c07e1fbed303ff2c5ec917)*

ci/verify-build.py
dybucc: Done.

.github/workflows/ci.yaml
dybucc: Done. I didn't even notice this. Must have been format on save while
editing the CI file.

dybucc:
Wondering if you saw [^1], @tgross35?

[^1]: <https://github.com/rust-lang/libc/pull/5411#issuecomment-5461679894>

@rustbot ready

build.rs
tgross35: Sorry missed this - I meant that the variables should imply each other e.g. `musl_v1_2 = musl && (musl_v1_2 || time64)`, `uclibc_use_time64 |= time64`. But don't worry about this, I'm noticing some other things about this file that I'll clean up after merge.

tgross35:
> I think it's done now. While checking through the CI workflow file, I noticed that we even though we set the "updated but deprecated" `RUST_LIBC_UNSTABLE_MUSL_V1_2` environment variable for certain targets [1](#user-content-fn-1-14d5e709cfcf504bc31ee68509252d12), we still check for `RUSTC_LIBC_UNSTABLE_MUSL_V1_2_3` in `build.rs` [2](#user-content-fn-2-14d5e709cfcf504bc31ee68509252d12).
> 
> I was wondering what's our stance on that change for stable releases, and whether we should even set the deprecated environment variable in CI. It seems wrong.

Oh, I did mean to update `main.yaml` here. Could you do that?

I'm not sure I follow the rest, why do you consider `RUST_LIBC_UNSTABLE_MUSL_V1_2` "updated but deprecated"? And what do we need to take a stance on regarding stable releases?

tgross35:
Could you rebase this once https://github.com/rust-lang/libc/pull/5508 lands? `let musl_v1_2 = ...`, `let uclibc_time64 = ...`, and `let gnu32_time64`.

tgross35:
Also please update the commit summary and PR title to mention the name of the config, it doesn't actually show up anywhere except the env form :)

ci/verify-build.py
```diff
@@ -409,13 +409,10 @@ def test_target(cfg: Cfg, target: Target) -> TargetResult:
...
-        # Equivalent of _TIME_BITS=64
-        run(cmd, rustflags=f'{rustflags} --cfg=libc_unstable_gnu_time_bits="64"')
-
-    if "musl" in target_env:
-        # Check with breaking changes from musl, including 64-bit time_t on 32-bit
-        run(cmd, rustflags=f"{rustflags} --cfg=libc_unstable_musl_v1_2")
+    if ("gnu" in target_env and target_bits == "32") or "musl" in target_env:
+        # Global toggle for _TIME_BITS=64 on glibc, and musl breaking changes
+        # concerning 64-bit time_t.
+        run(cmd, rustflags=f"{rustflags} --cfg=libc_unstable_time64")
```
tgross35, line 419: I added uclibc to these checks, so this will need to get updated

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5411/96b78673b6e396ee68c5161798faf0df11623290..3198d469daac8266e1585789bb4823ca3a61d359)*

tgross35 (changes requested):
Updating status. I think that this merging will be the cutoff for the next release.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5411/96b78673b6e396ee68c5161798faf0df11623290..3198d469daac8266e1585789bb4823ca3a61d359)*

build.rs
dybucc: I think that's done now.

ci/verify-build.py
dybucc: Done.

dybucc:
>> I think it's done now. While checking through the CI workflow file, I
>> noticed that we even though we set the "updated but deprecated"
>> `RUST_LIBC_UNSTABLE_MUSL_V1_2` environment variable for certain
>> targets [1](#user-content-fn-1-14d5e709cfcf504bc31ee68509252d12), we
>> still check for `RUSTC_LIBC_UNSTABLE_MUSL_V1_2_3` in `build.rs`
>> [2](#user-content-fn-2-14d5e709cfcf504bc31ee68509252d12).
>>
>> I was wondering what's our stance on that change for stable releases,
>> and whether we should even set the deprecated environment variable in
>> CI. It seems wrong.
>
>Oh, I did mean to update `main.yaml` here. Could you do that?
>
>I'm not sure I follow the rest, why do you consider
>`RUST_LIBC_UNSTABLE_MUSL_V1_2` "updated but deprecated"? And what do we
>need to take a stance on regarding stable releases?

Just updated the workflow file.

I consider `RUST_LIBC_UNSTABLE_MUSL_V1_2` deprecated because we phased
out support for those environment variables a while ago in favor of
unstable `cfg`s. CI sets the `cfg` when it detects `TEST_MUSL_V1_2`.

I also consider it updated because #5200 brought with it changes to the
specific musl version for which the `cfg` gates functionality. Though
`RUST_LIBC_UNSTABLE_MUSL_V1_2` has never been supported.

The build script continues to check for `RUST_LIBC_UNSTABLE_MUSL_V1_2_3`
(note the `_3`.) I guess this didn't get changed because it only really
continues being supported for backwards-compatibility.

@rustbot ready

libc-test/build/main.rs
```diff
@@ -748,7 +749,7 @@ fn test_windows(t: &Target) {
     // Needed for the Windows `time_t` test.
     println!("cargo::rustc-check-cfg=cfg(gnu_time_bits64)");
 
-    if x86_32 && gnu && win_gnu_x86_time64 {
+    if x86_32 && gnu && (win_gnu_x86_time64 || time64) {
```
tgross35: Make `time64` set `win_gnu_x86_time64` so that variable is accurate. I.e. `win_gnu_x86_time64 |= time64`

libc-test/build/main.rs
tgross35, line 3986: Similarly, update the `||`s to include `time64` since that option should force `musl_v1_2`

libc-test/build/main.rs
```diff
@@ -3986,7 +3990,7 @@ fn test_linux(t: &Target) {
     }
 
     let uclibc_use_time64 = env_flag("CARGO_CFG_LIBC_UNSTABLE_UCLIBC_TIME64");
-    if uclibc && uclibc_use_time64 {
+    if uclibc && (uclibc_use_time64 || time64) {
         cfg.cfg("linux_time_bits64", None);
     }
```
tgross35: Ditto here for uclibc

libc-test/build/main.rs
tgross35, line 1: "crate: expose `cfg` for general time64" please mention the cfg name in the commit message summary, it doesn't show up anywhere in this commit

tgross35 (changes requested):
Biggest thing here is the commit message update, the rest are just nits that I could do in a followup. LGTM overall though, if you're not around I can just update the commit message.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5411/df6561676506cd9c36ebe84b3b2f000661b64767..21e32e883207464310a45d1aa988c043d735a441)*

tgross35:
> I consider `RUST_LIBC_UNSTABLE_MUSL_V1_2` deprecated because we phased out support for those environment variables a while ago in favor of unstable `cfg`s. CI sets the `cfg` when it detects `TEST_MUSL_V1_2`.
> 
> I also consider it updated because #5200 brought with it changes to the specific musl version for which the `cfg` gates functionality. Though `RUST_LIBC_UNSTABLE_MUSL_V1_2` has never been supported.
> 
> The build script continues to check for `RUST_LIBC_UNSTABLE_MUSL_V1_2_3` (note the `_3`.) I guess this didn't get changed because it only really continues being supported for backwards-compatibility.

Yup that all makes sense. After this merges and before the release, I'll delete any time-specific env that hasn't already existed in a release.

libc-test/build/main.rs
dybucc: It does, though (before this review, I mean.)

libc-test/build/main.rs
dybucc: Done.

libc-test/build/main.rs
dybucc: Done.

libc-test/build/main.rs
dybucc: Done.

dybucc:
>Biggest thing here is the commit message update, the rest are just nits that I
>could do in a followup. LGTM overall though, if you're not around I can just
>update the commit message.
>
>*[View changes since this
>review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5411/df6561676506cd9c36ebe84b3b2f000661b64767..21e32e883207464310a45d1aa988c043d735a441)*

I am around, only on the `ctest` extension PR.

@rustbot ready

ci/verify-build.py
tgross35: ... I guess I didn't actually send the commit I had ready. Oh well, thanks for adding it here.

libc-test/build/main.rs
tgross35: What am I missing? 390156328eeb3968fd0de12776fc2d1cd22e7f22 has the text:

> crate: expose cfg for general time64
>
> Add `cfg` enabling `time64` functionality across all supported targets.
> This ensures users have a simple entry point to the crate functionality
> gated behind one of `linux_time_bits64`, `uclibc_time64` and
> `gnu_time_bits64`. The new `cfg` goes by the name `time64`.

`` `libc_unstable_time64` `` is not in the summary, and I don't see it mentioned at all

tgross35: This one needs to be `musl && (loongarch64 || hexagon || pauthtest/* || ohos */ || time64)`, then `|| time64` dropped from `musl_v1_2` (since that one should imply musl)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5411/0c0db5cd9f3c658d6b511328a6dd823a270d82cb..6aaa737d1887c6eee56040e0627f91304fb1b69e)*

dybucc: It does; At the end. I mention `time64` but not `libc_unstable_time64`
because all our public `cfg`s require `libc_unstable_` to be prepended
to them. It seems to me like the commit history is something internal.

The PR summary also mentions it, though I get it is confusing to say
that the `cfg` itself is named `time64`.

>The name of the `cfg` is `time64`.

dybucc: Done.

tgross35: It won't always be the case that they have a common prefix, this one is going to be stabilized as `libc_time64`. Just refer to them by the exact name to avoid confusion, when somebody greps in commit logs they're most likely going to be looking for the exact name rather than a substring of it.

tgross35:
@dybucc There's still a test failure here, and could you please add the full cfg name to the commit summary? https://github.com/rust-lang/libc/pull/5411#discussion_r4034754260

tgross35 (changes requested):
(status update)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5411/00f5e5c126a6e9224f07fbe1f41253bb3149b9ad..c3b6f88b02036c1d49e09bda7535a659a1b588ea)*

dybucc:
>@dybucc There's still a test failure here, and could you please add the
>full cfg name to the commit summary?
>https://github.com/rust-lang/libc/pull/5411#discussion_r4034754260

Done, except the test. Sorry, I checked off this PR from my list of
pending stuff without properly finishing work on it. The test failure
seems odd; This PR introduced no changes that should make it fail.

The assertion about the `linux_time_bits64` `cfg` being only valid for
Linux may be too restrictive. This is currently set in `main` also for
targets using uClibc [^1].

The reason why this might not be popping up in CI workflow runs across
other PRs is because there is already another (likely unrelated) CI
failure concerning an AIX target. L4Re tests haven't run at that point.

[^1]: <https://github.com/rust-lang/libc/blob/35c8b7d847f64265cb578343dc5d1be467fe63db/build.rs#L265-L269>

tgross35:
AIX is probably https://github.com/rust-lang/rust/pull/162979, feel free to add a commit disabling that with a FIXME for now.

> The assertion about the `linux_time_bits64` `cfg` being only valid for Linux may be too restrictive. This is currently set in `main` also for targets using uClibc [1](#user-content-fn-1-3fa8fda50b6a8a5d0d37b19055636c3d).
> 
> The reason why this might not be popping up in CI workflow runs across other PRs is because there is already another (likely unrelated) CI failure concerning an AIX target. L4Re tests haven't run at that point.

Is that one failing? I didn't notice it hitting any error last time I ran check-all-targets, but that wasn't super recent.

I guess uclibc should have its own cfg independent of Linux, like glibc does. Or we could merge more cfgs together in the future after this single-cfg option is around for a bit.

dybucc:
>AIX is probably https://github.com/rust-lang/rust/pull/162979, feel
>free to add a commit disabling that with a FIXME for now.

Just commented out the AIX target from being checked in CI.

>I guess uclibc should have its own cfg independent of Linux, like glibc
>does. Or we could merge more cfgs together in the future after this
>single-cfg option is around for a bit.

I've made available a new `cfg` for uClibc targets. This may need
further discussion, so for now it continues to use the same Linux `cfg`
in code, but the build script differentiates from uClibc.

Back when this was introduced, there were no issues with the uClibc
targets setting up a seemingly Linux-only flag [^1].

[^1]: <https://github.com/rust-lang/libc/pull/5046/changes#diff-aedce4e8d381f53dae77653609e3174da7ee2a184308251ad33f7a47bebabe14R3755-R3760>

.github/workflows/ci.yaml
```diff
@@ -392,34 +392,34 @@ jobs:
...
-    - name: Install Rust
-      # FIXME(rust-lang/rust#160439): unpin once the issue is fixed
-      run: rustup update nightly-2026-08-02 --no-self-update && rustup default nightly-2026-08-02
-    - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
-    - run: cargo doc --workspace --no-deps
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Install Rust
+        # FIXME(rust-lang/rust#160439): unpin once the issue is fixed
```
xtqqczze, line 419: nit: can we include full URL: `https://github.com/rust-lang/rust/pull/162979`, makes navigation a bit easier.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5411/35c8b7d847f64265cb578343dc5d1be467fe63db..eaabb4a439e4a01368dbcb31647db893634154ff)*

xtqqczze: Image

dybucc: I wouldn't do that simply because there's other `FIXME` comments
throughout the codebase already using this style. I think only when the
links are outside GitHub is it that full links are included.

If we were to include the full link, though, I would think it best to
not include it in between parenthesis, but rather as part of the comment
"body" (e.g. `FIXME: see <link-here>`.)

(If that's the GitHub PR extension for VSCode, I think there's a setting
for it to attempt to link to the right issue when detecting these
shortforms in any file.)

xtqqczze: I think using a format like `rust-lang/rust#160439` is generally over-indexing on GitHub. I don't think readers should need to configure an extension just to make the reference navigable.

dybucc: I digress but, of course, this is open source, so be my guest.

xtqqczze: By the way, since https://github.com/rust-lang/rust/pull/162979 was recently merged, is this pin expected to be removable once `nightly-2026-09-20` is released, or is there another issue that still needs to be addressed?

tgross35:
Cherry picked the AIX skip as https://github.com/rust-lang/libc/pull/5541 to unblock other things

.github/workflows/ci.yaml
dybucc: I don't think the issue the comment [^1] mentions is related to
rust-lang/rust#162979.

[^1]: rust-lang/rust#160439

tgross35 (approved):
Sorry for the conflict but I think that should resolve the uclibc issue. LGTM assuming CI passes

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5411/0a29e47c39172f430c70bd2158ef2b97370ace4b..050b9e51c91c9b0f9002cc2f4ebb8983c9e8522a)*

PR #4621, opened 2025-08-07
Add additional fields to statx

Thomasdezeeuw:
# Description

Updates statx to include all fields as of Linux v6.16.

# Sources

* Linux: https://github.com/torvalds/linux/blob/038d61fd642278bab63ee8ef722c50d10ab01e8f/include/uapi/linux/stat.h#L99-L193
* glibc: https://sourceware.org/git/?p=glibc.git;a=blob;f=io/bits/types/struct_statx.h;h=f8ecc53768e0980d9ef3c737c2141ae769f80e4f;hb=f762ccf84f122d1354f103a151cba8bde797d521#l55-l62

# Checklist

- [x] Relevant tests in `libc-test/semver` have been updated
- [x] No placeholder or unstable values like `*LAST` or `*MAX` are
  included (see [#3131](https://github.com/rust-lang/libc/issues/3131))
- [x] Tested locally (`cd libc-test && cargo test --target mytarget`);
  especially relevant for platforms that may not be checked in CI

@rustbot label +stable-nominated

PR #5412, opened 2026-08-16
linux_like: add missing `STATX_*` constants

valentynkit:
libc already has `struct statx` fields through, which were added in rust-lang/libc#4621 but there are missing masks, so several fields are exposed with no way to request them. 
Adding the five constants the header defines and libc does not.

The Android NDK  doesn't have `STATX_DIO_READ_ALIGN` yet, so left out for Android. Three others landed in r28 and CI still runs r27, so they're skipped in `test_android`, without it, the fails will probably fail.

Earlier attempt adding these constants: rust-lang/libc#4401, closed unfinished.

Link: https://github.com/torvalds/linux/blob/8cd9520d35a6c38db6567e97dd93b1f11f185dc6/include/uapi/linux/stat.h#L218-L221
Link: https://github.com/torvalds/linux/blob/8cd9520d35a6c38db6567e97dd93b1f11f185dc6/include/uapi/linux/stat.h#L257
Link: https://android.googlesource.com/platform/bionic/+/a63b21091ada240d92f8ceadb1cabbdedaaab81b/libc/kernel/uapi/linux/stat.h#95
Link: https://man7.org/linux/man-pages/man2/statx.2.html

## Checklist

  doc comment


JohnTitor (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5412/1a8e71f33b1d6ea1e072210e7fc994417bbb2e34..0d0e5ac8f48257d1bc0b2c8577d4c247266e6858)*

PR #5413, opened 2026-08-16
deps: Bump cc to 1.4.3

xtqqczze:


JohnTitor:
Any motivation for this? If v1.4.3 is required for something, it'd be better to update Cargo.toml as well.

xtqqczze:
It's not required, just a bump to align with the recent `cc` MSRV increase.

tgross35:
We get the periodic updates via dependabot, though it would be nicer to configure rennovate like some other repos have

JohnTitor:
Yeah, without a strong need (like a fix for CI), I usually prefer updating deps via automation not to introduce unwanted/unrelated changes and to enforce the cooldown config (libc should configure it though!).

xtqqczze:
@tgross35 I've not seen dependabot bumping Rust crates in this repo?

tgross35:
Indeed, looks like it's just been the ones I've manually done. Opened https://github.com/rust-lang/libc/pull/5415

tgross35:
This got picked up by the renovatebot tooling in d9195db775bd89b00530e8a1a6cf5f341100bce5 so should be all set. Not sure why GH/triagebot didn't mark it as conflicted (guess it's just empty?)

PR #5434, opened 2026-08-21
glibc: Add `f_type` to `statvfs` and `statvfs64`

### LINKED ISSUE #4121: `statvfs` and `statvfs64` got a new field `f_type`
This came from the reserved field, we just need to adjust our type https://github.com/bminor/glibc/commit/92861d93cdad13834f4d8f39504b550a80ad8200.

valentynkit:
glibc 2.39 allocated one of the six reserved `__f_spare` ints to `f_type` for `statvfs` and `statvfs64`, which mirrors `statfs`.
https://github.com/sailfishos-mirror/glibc/commit/92861d93cdad13834f4d8f39504b550a80ad8200

1. Move `statvfs` and `statvfs` defininitions for glibc to `src/new/glibc`, this follows the ongoing migration into `src/new`, which reduces duplication and models glibc's source tree.
The `__f_unused` padding fields is only declared on some targets, so it's gated behind a `cfg`.

	`riscv32` and `riscv64` declared `__f_spare as pub, which was inconsistent with the other declarations and hand no grounded reason to stay, now it's private.

2. Add `f_type` to `statvsfs` and `statvfs64` for glibc in shared declarations, using `c_uint` type.

## Open Questions
1. `riscv32` `statvfs64` doesn't have `__f_unused` in its declaration, but its neighboring `statvfs` has this padding, so I think one of the two must be wrong, for me it seems like accidental bug or inconsistency, however I may be missing some reason behind this. To avoid making changes to the layout of this struct, I didn't touch it here: this field is excluded on `statvfs64` for riscv32`, so no layout changes. Should it be addressed separately?
2. Shared `statvfs64` uses `u64` for the counters, some architectures used `fsblkcnt64_t` / `fsfilcnt64_t`, which should be the same as u64, but wasn't defined on all architectures, so `u64` was used instead.
Are there any specific difference in using `fsblkcnt64_t` / `fsfilcnt64_t` instead of `u64` that I may be missing?

Closes rust-lang/libc#4121

## Checklist

- [x] Relevant tests in `libc-test/semver` have been updated, no change needed.
- [x] No placeholder or unstable values like *LAST or *MAX.
- [x] Tested locally (`cargo test -p libc-test --target mytarget`); especially relevant for platforms that may not be checked in CI

### DIFF
--- a/libc-test/build/main.rs
+++ b/libc-test/build/main.rs
@@ -5200,9 +5200,12 @@ fn test_linux(t: &Target) {
             }
             // invalid application of 'sizeof' to incomplete type 'long unsigned int[]'
             ("mcontext_t", "__extcontext") if musl && loongarch64 => true,
-            // FIXME(#4121): a new field was added from `f_spare`
-            ("statvfs", "__f_spare") => true,
-            ("statvfs64", "__f_spare") => true,
+            // glibc 2.39 took one of the six spares for `f_type`
+            ("statvfs" | "statvfs64", "f_type" | "__f_spare")
+                if gnu && versions.glibc.unwrap() < (2, 39) =>
+            {
+                true
+            }
             // the `xsk_tx_metadata_union` field is an anonymous union
             ("xsk_tx_metadata", "xsk_tx_metadata_union") => true,
             // After musl 1.2.0, the type becomes `int` instead of `long`.
--- a/src/new/glibc/mod.rs
+++ b/src/new/glibc/mod.rs
@@ -24,6 +24,14 @@ mod posix {
 #[cfg(target_os = "linux")]
 pub(crate) mod signal;
 
+/// Source directory: `io/sys/`
+///
+/// <https://github.com/sailfishos-mirror/glibc/tree/master/io/sys>
+pub(crate) mod sys {
+    #[cfg(target_os = "linux")]
+    pub(crate) mod statvfs;
+}
+
 /// Source directory: `sysdeps/`
 ///
 /// <https://github.com/sailfishos-mirror/glibc/tree/master/sysdeps>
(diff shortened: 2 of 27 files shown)


valentynkit:
Is it okay to have it as one PR, or should dedup in gnu/b32 and gnu/b64 be separate PR?
When I started adding `f_type` noticed that there were a lot of duplication of identical `statvfs` and `statvfs64`, so addressed it in separate commits before adding `f_type`.

tgross35:
Any chance you'd be willing to move `statvfs*` to `src/new/glibc/sysdeps/unix/linux` first? That would get rid of a ton of duplication. It's a bit tricky but it models the glibc structure and reexports on arches that have an override, see `sigaction`/`signum_arch` in https://github.com/rust-lang/libc/blob/2eda1e52fe2efdf93c1e5c65e346343a530d8bc3/src/new/glibc/sysdeps/unix/linux/mod.rs.

Doing this in the same PR is fine as long as it's a separate commit, we use rebase merging.

tgross35:
I missed a couple of things:

> There is also [musl](https://github.com/kraj/musl/blob/0784374d561435f7c787a555aeab8ede699ed298/include/sys/statvfs.h#L26-L27) that is missing `f_type`. It wasn't addressed in this PR, should I address musl also, or any other not just glibc?

Mind doing musl in a followup?

> Also `riscv32` and `riscv64` declare their `__f_spare` field as pub, it feels inconsistent with the other declarations, and for `riscv64` it is the reason it keeps its own declaration (`riscv32` differs in also lacking `__f_unused`). Is there a specific reason to have it `pub __f_spare`, or should this be addressed and fixed?

It's fine to make it non-`pub` in the deduplication PR, just make sure to note it in the commit message.

Few things here so
@rustbot author

valentynkit:
> Mind doing musl in a followup?

Yeah, I would like to handle it as follow up PR, if that's okay.

tgross35 (approved):
Looks great, thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5434/6ae84e49192725e0606c12c63a9e129c96f12c3d..385a2089cbd75e40f7489d7beedb077b7823df69)*

PR #5435, opened 2026-08-23
Change private and deprecated padding fields to use `Padding`

### LINKED ISSUE #1453: Wrap padding fields in `Padding(MaybeUninit<T>)`

dybucc:
## Description

Attempts to change whichever padding fields didn’t already use our
custom `Padding` type. Closes rust-lang/libc#1453.

Originally, this description didn't contain the command used for
searching the files. Other than the search results found by the
following shell command, each of those files was manually inspected.

```text
rg -g "src/**/*.rs" -s -n -e "pad|res(erve(d)?)?|unuse(d)?|fill(er)?|spare"
```

A split PR with changes to the public field lives now at
rust-lang/libc#5444.

## Checklist

      standard doc comment


tgross35 (changes requested):
Could you split this into separate commits?

1. Changes to fields that are already private
2. Changes to fields that are deprecated
3. Changes to fields that are public (preferably a separate PR for labels)

That way I can backport the first two but hold off on the last one. I know our policy technically does allow us to change them, but some of these are pretty popular structs so I'd like to be a bit cautious anyway.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5435/2eda1e52fe2efdf93c1e5c65e346343a530d8bc3..7b8c9ca7909d65b1d69653444f14bdb1bcea16b0)*

tgross35 (approved):
LGTM, thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5435/bd415df57bc97f3cba6299d2d392bbb65566f923..d448eacdcee31bb1930e74a3c4bc3ce0aeb77bcb)*

tgross35 (changes requested):
Ah sorry, this will need a rebase due to https://github.com/rust-lang/libc/pull/5434

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5435/bd415df57bc97f3cba6299d2d392bbb65566f923..d448eacdcee31bb1930e74a3c4bc3ce0aeb77bcb)*

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5435/7c1cc59f2fec89c87c1a5908b5435effc8fbeb77..706af3d6d1c3c4c9a3e3d3918cd07dd798d040b2)*

PR #5441, opened 2026-08-26
Exclude the helper python script from the published package

weiznich:
It seems like this is not really helpful for consumers as the script itself states that it is for developing libc itself. Therefore it might just be excluded from the published crates.io package.


tgross35:
Thanks for putting this together, but this is already covered by https://github.com/rust-lang/libc/pull/5312. That will be included in the next release.

weiznich:
Perfect, that's no problem for me

PR #5442, opened 2026-08-27
bsd/apple: add missing NET_RT_DUMP2 definition

npry:
The missing constant is defined here:

https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/socket.h#L582

It's defined next to `NET_RT_IFLIST2`, which was already included here right next to the new definition.


JohnTitor (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5442/6ae84e49192725e0606c12c63a9e129c96f12c3d..7e8d70776585fcb74746b3b078444902b32470bb)*

PR #5443, opened 2026-08-28
linux_like(android,linux): remove `af_alg_iv` unsound implementations in `extra_traits` feature

dybucc:
## Description

Closes rust-lang/libc#1501. See that issue for details on the
unsoundness report. That was fixed a while ago on `main` but the patch
would have broken stable, so a deprecation warning was applied instead.
Follows a brief discussion at
<https://github.com/rust-lang/libc/issues/1501#issuecomment-5405396053>.

Note the automatically-derived implementations for `Debug` and `Clone`
for this type seem odd now. The last member is a FAM upstream, so I
think `Debug` can just live with that, but `Clone` wouldn't feel right
to provide without some word of warning to users about the "extended"
length beyond the trailing `struct` member _not_ being copied over to
the new instance of the type. I guess downstream users can then use a
newtype to implement `Clone` as they see fit; The fields on `af_alg_iv`
are public, after all.

## Checklist

      standard doc comment

### DIFF
--- a/src/unix/linux_like/android/mod.rs
+++ b/src/unix/linux_like/android/mod.rs
@@ -610,13 +610,6 @@ s! {
 }
 
 s_no_extra_traits! {
-    /// WARNING: The `PartialEq`, `Eq` and `Hash` implementations of this
-    /// type are unsound and will be removed in the future.
-    #[deprecated(
-        note = "this struct has unsafe trait implementations that will be \
-                removed in the future",
-        since = "0.2.80"
-    )]
     pub struct af_alg_iv {
         pub ivlen: u32,
         pub iv: [c_uchar; 0],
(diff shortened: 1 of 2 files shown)


tgross35 (approved):
LGTM, thanks! Will merge as part of the next release's backports

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5443/ef0906e20828777175f65caa7e681a0ce33c559a..c76399fffb75e1e7cc39ebda45a83495796f9465)*

tgross35:
No need to update this but for future reference, if you mention a commit sha in a commit message it's good to include a `abcd12345678 ("Title of the commit")` to hint what it's about, and to help the message survive any unlikely but possible force pushes. (Mentioned at https://docs.kernel.org/process/submitting-patches.html#describe-your-changes, I usually do the 12 characters.)

dybucc:
Mind closing #1501, @tgross35? I think this PR didn't get linked with
that issue because I set the "Closes" message after opening it.

tgross35:
Done, thanks!

PR #5445, opened 2026-08-29
Unix: scope sighandler_t to GNU/Linux platforms

chengr4:
This PR fixes part of rust-lang/libc#1273 [^1] and rust-lang/libc#1359.

Question: 
With `sighandler_t` removed on non-GNU targets, should I also clean up the existing skip rules such as`skip_alias("sighandler_t")` and `rename_type` etc. in `libc-test/build/main.rs` in this PR? (I am worried that I might break the tests. 🫣)

---
[EDIT 2026-09-03]

`sighandler_t` is a GNU extension provided by glibc on Linux and Hurd.
Non-GNU Unix platforms [^2] do not define `sighandler_t` in their native C headers (BSDs define `sig_t` instead).

Previously, `sighandler_t` was declared in `src/unix/mod.rs` for all Unix platforms, 
causing type mismatches and requiring workarounds in test configurations.

## Changes

1. Scope `pub type sighandler_t = size_t;` only to GNU/Linux-like platforms (`linux_like` and `hurd`).
2. Remove `sighandler_t` from non-GNU platforms where it does not exist natively in C headers.
3. (Fix CI error) Update FreeBSD and Apple test configurations to skip `SIG_DFL`, `SIG_IGN`, and `SIG_ERR` in `ctest`.

[^1]: https://github.com/rust-lang/libc/issues/1273#issuecomment-5436574721
[^2]: BSDs, Solaris, AIX, and Cygwin


chengr4:
Where does the PR message template go? 🤔

src/unix/mod.rs
```diff
@@ -1341,23 +1354,40 @@ extern "C" {
...
         target_os = "dragonfly",
         target_os = "netbsd",
         target_os = "openbsd"
-    ))]
-    pub fn signal(signum: c_int, handler: sig_t) -> sig_t;
+    )))]
+    pub fn signal(
+        signum: c_int,
+        handler: Option<unsafe extern "C" fn(c_int)>,
+    ) -> Option<unsafe extern "C" fn(c_int)>;
```
chengr4: no confidence

tgross35: This cfg is complex. Delete `signal` from this file and add it back to each of the child modules, with the correct definition.

tgross35 (changes requested):
There are still some tests failing. Could you also update the commit message and PR description to mention this is only on platforms that don't exist? Sounds a bit like unconditional removal as-is.

> > With `sighandler_t` removed on non-GNU targets, should I also clean up the existing skip rules such as`skip_alias("sighandler_t")` and `rename_type` etc. in `libc-test/build/main.rs` in this PR? (I am worried that I might break the tests. 🫣)

Yes please! The point is to clean up those tests :)

Basically, ideally, all mentions of `sighandler_t` should be deleted and then `libc` is updated as-needed to make things work. We might need a few skips still not nearly as many as are currently there.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5445/96b78673b6e396ee68c5161798faf0df11623290..dfcb972ed417594fd30b5ebf515792dc2bde495e)*

chengr4:
@rustbot ready

**Note:** Honestly, I didn't go through `<signal.h>` on all Unix platforms to carefully check my fix; I only randomly and roughly picked a few platforms to read.

src/unix/mod.rs
```diff
@@ -250,9 +249,23 @@ pub const INT_MIN: c_int = c_int::MIN;
...
+    ))] {
+        pub const SIG_DFL: sighandler_t = 0 as sighandler_t;
+        pub const SIG_IGN: sighandler_t = 1 as sighandler_t;
+        pub const SIG_ERR: sighandler_t = !0 as sighandler_t;
+    } else {
+        pub const SIG_DFL: size_t = 0;
+        pub const SIG_IGN: size_t = 1;
+        pub const SIG_ERR: size_t = !0;
+    }
+}
```
tgross35, line 268: These should be split across platfoms too

src/unix/linux_like/mod.rs
```diff
@@ -8,6 +8,7 @@ pub type timer_t = *mut c_void;
 pub type useconds_t = u32;
 pub type key_t = c_int;
 pub type id_t = c_uint;
+pub type sighandler_t = size_t;
```
tgross35, line 11: Is this actually available on all linux-like platforms? I.e. emscripten, linux, android, l4re

tgross35 (changes requested):
Skimming this it looks plausable, but could you scope only to removing `sighandler_t` on unsupported platforms? Currently it does that but also changes the signatures to take `Option<unsafe extern "C" fn(c_int)>` while the actual 1:1 substitute is `size_t`. That is something we want and we can't remove all test skips until that happens, but it would be better in a separate PR.

@dybucc could you review this as well since you've been working in the area?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5445/60c933f0eb80a5324025abb9e72d30a2da4c6629..039ddbe4d3e197c022fc6d6cd59151d49ab55ef2)*

tgross35:
Actually could you make this PR just split the definitions up and deprecate `sighandler_t` on unsupported platforms, then actually remove them in a separate PR? That way we can backport the deprecations.

Also, no AI when communicating with humans please. That includes commit messages and PR descriptions. (This needs a squash anyway)

Sorry for the conflict, I cleaned up the `crate::c_*` in a separate commit since it's unrelated to this and I noticed a few other cases 4e4a96a0d461420a6d61e430dbd5a810d1645c0b.

src/unix/linux_like/mod.rs
chengr4: I will check

chengr4:
> Also, no AI when communicating with humans please. That includes commit messages and PR descriptions. (This needs a squash anyway)

I use AI as a drafting tool for my commit messages and PR templates because English is not my first language. However, every single word is reviewed, judged, and intentionally chosen by me. I never copy-paste AI output without double-checking. Every sentence in PRs or commit messages is I believe it can help with code review and git history tracing. 

Could you point out specifically where you feel it is over-expressed? E.g., which paragraphs you think are unnecessary

I am open to any suggestions for improvement.

tgross35:
Well I've never had any issue communicating with you so your English is quite alright :)

There's nothing specifically bad but this is a policy we have. When AI writes PRs or commit descriptions it tends to focus on "what changed" while human authors tend to focus on the "why did it change", which aids review. (E.g. the mechanical change of skipping SIG_* consts in tests is mentioned multiple times in the PR and commit messages, but the very notable change to the function signature is not mentioned at all.)

chengr4:
> There's nothing specifically bad but this is a policy we have. When AI writes PRs or commit descriptions it tends to focus on "what changed" while human authors tend to focus on the "why did it change", which aids review. (E.g. the mechanical change of skipping SIG_* consts in tests is mentioned multiple times in the PR and commit messages, but the very notable change to the function signature is not mentioned at all.)

Got it, I will work toward improving this in that direction. **Again, please do not hesitate to point out anything you feel could be improved.**

One more quick thing: since I know how important `libc` is for the Rust ecosystem, I am honestly super nervous about breaking things every time I touch it. That is a big reason why I wanted to write down every single change.

---

> but could you scope only to removing sighandler_t on unsupported platforms?  

I will think how to do it.

dybucc:
>@dybucc could you review this as well since you've been working in the
>area?

Two things; If we're moving `sighandler_t` and `signal`, maybe it's
worth it to just move them to `new`'s module tree. Then maybe it's worth
it to look at MinGW's `sighandler_t` under Windows as well.

MinGW is not strictly tied to GNU environments, but it sure borrows from
those, so it may be worth at least revisiting the use of `usize` in the
Windows module for `sighandler_t`.

I think MSVC doesn't have `sighandler_t`, so gating that in the Windows
module may also be in order.

chengr4:
> MinGW is not strictly tied to GNU environments, but it sure borrows from
those, so it may be worth at least revisiting the use of usize in the
Windows module for sighandler_t.
> I think MSVC doesn't have sighandler_t, so gating that in the Windows
module may also be in order.

Note: This PR only focuses on Unix platforms; We need another PR for Windows.

src/unix/linux_like/mod.rs
chengr4: I checked all. I think this is correct.

- (glibc) https://github.com/sailfishos-mirror/glibc/blob/d5bd0f22305277fa0d431bb9d733aa9456706573/signal/signal.h#L188
- (musl) https://github.com/kraj/musl/blob/1c82569f4cc53f5bb25c820051c2a02463f46cab/include/signal.h#L273
- (uClibc-ng) https://github.com/wbx-github/uclibc-ng/blob/3f2ef085d8455a13fcdc4d4d473c0385dd65d55e/include/signal.h#L226
- (Android) https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/bits/signal_types.h;l=73;drc=61197364367c9e404c7da6900658f1b16c42d0da
- (Emscripten) https://github.com/emscripten-core/emscripten/blob/d6c521a7f05449857c76bd99e396895583cf2083/system/lib/libc/musl/include/signal.h#L273
- (L4Re) https://github.com/kernkonzept/l4re-core/blob/941f09bbf01ee153c5b66caff87c868deab3c91c/libc/musl/contrib/musl/include/signal.h#L278

PR #5446, opened 2026-08-30
fix: fix some CI issues

JohnTitor:
- fix: move i686-pc-windows-gnu to without host tools target
- fix: remove repeated reprs
- fix: use Debian to sync with Cuttlefish


tgross35:
> fix: use Debian to sync with Cuttlefish

What was the issue with that one?

JohnTitor:
@tgross35 

> What was the issue with that one?

Ubuntu and Debian use different package names and Cuttlefish switched to Debian, thus CI fails to install required packages, see https://github.com/rust-lang/libc/actions/runs/33300799228/job/99241781853#step:8:814

(I guess renaming in some way could be possible but Using the same image with Cuttlefish seems a clearer way to me 🤔 )

tgross35:
Oh 🙃 that's rather interesting, makes sense. Cc @10ne1 since this affects some things you've been working on

PR #5447, opened 2026-08-30
musl: Make rustc-dep-of-std imply musl_v1_2

Gelbpunkt:
The musl baseline for Rust is 1.2.5 since 1.93.0. Therefore, it is safe to make rustc-dep-of-std imply musl_v1_2, and this lets us start using statx from libc in std.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -151,7 +151,8 @@ fn main() {
         _ => (),
     }
 
-    let mut musl_v1_2 = env_flag("CARGO_CFG_LIBC_UNSTABLE_MUSL_V1_2");
+    let mut musl_v1_2 =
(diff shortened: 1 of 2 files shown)


tgross35 (changes requested):
The change here looks fine to me, though a comment in `Cargo.toml` would be helpeful. However, I'd rather not merge this until there's a rust-lang/rust PR setting this cfg via bootstrap and doing try jobs on our musl targets. That won't be the permanent of course, but I would be better to find anything broken and fix it without needing to ping pong between libc releases and rustc PRs.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5447/bd415df57bc97f3cba6299d2d392bbb65566f923..eeb2e8798e4d6a4383a3f66cad249f10f9ad76df)*

PR #5448, opened 2026-08-30
linux_like: Expose statx on all musl versions

Gelbpunkt:
Exposing the constants and the method in itself is not a breaking change, users are responsible for knowing whether they can call the method without running into linker errors, just like on glibc.

### DIFF
--- a/libc-test/build/main.rs
+++ b/libc-test/build/main.rs
@@ -4413,8 +4413,8 @@ fn test_linux(t: &Target) {
             // FIXME(musl): Struct has changed for new musl versions
             "tcp_info" if musl => true,
 
-            // FIXME(musl): Supported in new musl but we don't have a new enough version in CI.
-            "statx" | "statx_timestamp" if musl => true,
+            // Added in musl 1.2.5
+            "statx" | "statx_timestamp" if old_musl => true,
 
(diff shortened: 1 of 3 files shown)


tgross35:
I think you'll need to update the test build.rs to skip these `if old_musl`, but the changes here LGTM assuming CI passes

Gelbpunkt:
I'm slightly confused why this passes tests in CI...

Gelbpunkt:
Ah: we unconditionally skip `statx` no matter the version...

Gelbpunkt:
Found a few issues with statx on musl in the process of this, will enable the checks and fix them in this PR

Gelbpunkt:
For `renameat2` testing, we'd have to disable linking the self contained musl, which in theory should work but it's a bit more involved on powerpc64. I'll leave that out of here and maybe look into it later

tgross35 (approved):
LGTM, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5448/e54d28e4415cd400d75c5c1d7e3455970533a9f0..71c0057400c13a00452fb8db3c6a6579675c8a6d)*

Gelbpunkt:
oh nevermind, you already took care of that :)

Issue #5209, opened 2026-06-24
Downstream CI for `libc`


tgross35:
`libc` is tested a handful of places downstream for targets that aren't Tier 2+, or have less common configurations. This issue is intended to aggregate a list of any such public CI.

| Targets | Maintainer | Link | Status Badge |
|---------|------------|------|--------------|

If you are interested in testing a target that can be run in Docker or natively on GitHub Actions (e.g. with a self-hosted runner), see https://github.com/tgross35/libc-downstream-ci for a simple template to get a daily job going.

If you are already running CI downstream anywhere with public results, comment a link and I'll update the table.

tgross35:
Cc @Gelbpunkt @asomers @semarie @maurer @skrap @neuschaefer @sunshowers @erickt @Nashenas88 @ivmarkov @MabezDev @SergioGasquez because I believe all of you have either expressed interest in running downstream CI, or are already doing so.

maurer:
Android's downstream CI is currently only operating on released copies of `libc` (when we update the compiler or the version of the crate available to platform developers). The team recently shrunk, so it's unlikely that a CI integrating ToT `libc` with Android is going to happen. (Compiler would be first on the list, and that's also not likely to happen.)

PR #5455, opened 2026-08-31
linux: fix unusable riscv32 GNU syscall numbers

### LINKED ISSUE #5379: Unusable syscall numbers defined for riscv32gc-unknown-linux-gnu
riscv32gc-unknown-linux-gnu currently defines this:

https://github.com/rust-lang/libc/blob/74e58cc7294fa594b205a33f8235a2b264455b13/src/unix/linux_like/linux/gnu/b32/riscv32/mod.rs#L667

But riscv32 doesn't have time32 syscalls:
https://github.com/torvalds/linux/blob/v7.1/arch/riscv/kernel/Makefile.syscalls
https://github.com/torvalds/linux/blob/v7.1/include/uapi/asm-generic/unistd.h#L269
https://github.com/bminor/glibc/blob/glibc-2.42/sysdeps/unix/sysv/linux/riscv/rv32/arch-syscall.h#L64-L68

These must either be left undefined or aliased to time64 syscalls, as musl does:

https://github.com/rust-lang/libc/blob/74e58cc7294fa594b205a33f8235a2b264455b13/src/unix/linux_like/linux/musl/b32/riscv32/mod.rs#L647

I only checked the SYS_futex, but other time32 system calls may have the same issue.

xjtu-ctgg:
## Description

Fix the syscall definitions for `riscv32gc-unknown-linux-gnu`.

RISC-V 32-bit does not provide the old time32 syscall implementations. This updates the GNU riscv32 module to use the corresponding time64 syscall numbers and removes the generic names that refer to unavailable time32 syscalls.

Closes rust-lang/libc#5379

## Sources

- https://github.com/torvalds/linux/blob/v7.1/arch/riscv/kernel/Makefile.syscalls
- https://github.com/torvalds/linux/blob/v7.1/include/uapi/asm-generic/unistd.h#L269
- https://github.com/bminor/glibc/blob/glibc-2.42/sysdeps/unix/sysv/linux/riscv/rv32/arch-syscall.h#L64-L68

## Tests

- `cargo +nightly-2026-08-23 test --workspace`
- `python3 ci/verify-build.py --toolchain nightly-2026-08-23 --only riscv32gc-unknown-linux-gnu`
- `cargo +nightly-2026-08-23 fmt --check`
- `git diff --check`


src/unix/linux_like/linux/gnu/b32/riscv32/mod.rs
```diff
@@ -760,3 +755,23 @@ pub const SYS_memfd_secret: c_long = 447;
...
+pub const SYS_timerfd_settime: c_long = SYS_timerfd_settime64;
+pub const SYS_utimensat: c_long = SYS_utimensat_time64;
+pub const SYS_pselect6: c_long = SYS_pselect6_time64;
+pub const SYS_ppoll: c_long = SYS_ppoll_time64;
+pub const SYS_recvmmsg: c_long = SYS_recvmmsg_time64;
+pub const SYS_mq_timedsend: c_long = SYS_mq_timedsend_time64;
+pub const SYS_mq_timedreceive: c_long = SYS_mq_timedreceive_time64;
+pub const SYS_rt_sigtimedwait: c_long = SYS_rt_sigtimedwait_time64;
+pub const SYS_futex: c_long = SYS_futex_time64;
+pub const SYS_sched_rr_get_interval: c_long = SYS_sched_rr_get_interval_time64;
```
tgross35: Could you preserve the existing ordering?

tgross35 (changes requested):
Are the values like `SYS_futex` not usable at all in C? `riscv32gc-unknown-linux-*` are T3 so they can be removed or deprecated first rather than aliased.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5455/86216b9d968c92a78edf2161c84a70b40347326a..6d8d553fd3ff9c1e8821ee3078823f4c154e0a80)*

src/unix/linux_like/linux/gnu/b32/riscv32/mod.rs
xjtu-ctgg: Thanks, updated. The generic names are now removed rather than aliased, and the `*_time64` constants stay in their original positions. I also rebased onto the latest main.  (^^;

tgross35 (approved):
LGTM: in the kernel's unistd.h all removed constants are gated by `#if defined(__ARCH_WANT_TIME32_SYSCALLS) || __BITS_PER_LONG != 32`, and this matches the glibc file. Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5455/96b78673b6e396ee68c5161798faf0df11623290..4fe2850504c8c325658b1550002142a04f8dafce)*

tgross35:
Are you developing for this platform? I would love to have some downstream CI, rv32-linux is very breakage prone because it's the one target that is 64-bit time_t by default at a time when we're trying to get our ducks in a row with that.

See https://github.com/rust-lang/libc/issues/5209 for instructions if you might be able to set something up

xjtu-ctgg:
Thanks! I'm not currently developing on rv32 Linux, so I don't have a downstream environment I can reliably maintain.

tgross35:
For what it's worth, if you're able to run the target in docker+qemu that's enough. We just can't do that in this repo because of the target tier policy

PR #5458, opened 2026-09-01
Fix WIFSTOPPED on NetBSD

krobelus:
If we expand the definitions from
https://github.com/NetBSD/src/blob/d04b0c735abc997743bb3faa74464524cbe7becd/sys/sys/wait.h#L61
we get

	#define WIFCONTINUED(x)	(x == 0xffff)
	#define WIFSTOPPED(x)	((x & 0177) == 0177 && !WIFCONTINUED(x))

Our definition of WIFSTOPPED forgot the !WIFCONTINUED(x) condition.
Siince the wait(3p) status for "continued" is 0xffff, this causes
WIFSTOPPED to return true for continued child processes.

Fix this.

Originally reported at https://github.com/fish-shell/fish-shell/issues/12929

### DIFF
--- a/src/unix/bsd/netbsdlike/netbsd/mod.rs
+++ b/src/unix/bsd/netbsdlike/netbsd/mod.rs
@@ -1875,7 +1875,7 @@ f! {
     }
 
     pub const safe fn WIFSTOPPED(status: c_int) -> bool {
-        (status & 0o177) == 0o177
+        (status & 0o177) == 0o177 && !WIFCONTINUED(status)
     }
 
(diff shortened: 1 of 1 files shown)


tgross35 (approved):
Thanks!

It would be nice to add non-public `_WSTATUS`, `_WSTOPPED`, and `_WCONTINUED` to make this match the source more, but no need to wait on that.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5458/96b78673b6e396ee68c5161798faf0df11623290..9371f4e77942b0e8f1990a9417d50d5dd4d501fe)*

PR #5464, opened 2026-09-04
apple: Disable the `mach.h` header in tests

tgross35:
This was added in https://github.com/rust-lang/libc/commit/f3b2c50d6a974e1e22a542a11356652083b91dec ("Fix iOS, tvOS, watchOS and visionOS tests") but the mach API has been deprecated, so we shouldn't need it.

### DIFF
--- a/libc-test/build/main.rs
+++ b/libc-test/build/main.rs
@@ -176,7 +176,6 @@ fn test_apple(t: &Target) {
         (macos, "libproc.h"),
         "limits.h",
         "locale.h",
-        "mach/mach.h",
         "malloc/malloc.h",
         (macos, "net/bpf.h"),
         (macos, "net/dlil.h"),


tgross35:
@madsmtm any thoughts about what to do here? I don't know whether these should be deprecated as part of mach, or if that header has more.

Failure list:

- thread_policy_set
- thread_info
- mach_error_string
- mach_error_t
- vm_allocate
- vm_deallocate
- host_statistics64
- host_processor_info
- task_for_pid
- task_info
- task_create
- task_terminate
- task_threads
- host_statistics

madsmtm:
All of the APIs you've listed there are indeed Mach APIs.

I haven't followed the history of why `libc` decided to deprecate Mach APIs, but if that is indeed the plan, then these should probably live in `mach2`.

Did that answer the question?

PR #5466, opened 2026-09-04
[0.2] OpenBSD: Deprecate `syscall`, `mincore`, and `KERN_NSELCOLL`

tgross35:
Commit 790e5c6d3900 ("Remove removed items in OpenBSD") removed these items on `main`, but there was no backport or deprecation to 0.2. Mark these items deprecated here.

Relevant relnotes, from [1]:

> Antiquated mincore(2) will not be needed and was removed, eliminating
> an interface that exposed physical machine information unnecessarily. 

From [2]:

> Removed the obsolete kern.nselcoll sysctl(2). 

From [3]:

> Removed support for syscall(2), the "indirection system call," a
> dangerous alternative entry point for all system calls and incompatible
> with the precision system call entry point scheme we are heading
> towards.

[1]: https://www.openbsd.org/plus65.html
[2]: https://www.openbsd.org/plus72.html
[3]: https://www.openbsd.org/plus75.html

### DIFF
--- a/src/unix/bsd/mod.rs
+++ b/src/unix/bsd/mod.rs
@@ -865,6 +865,10 @@ cfg_if! {
     // on iOS and macOS).
     if #[cfg(not(any(target_os = "tvos", target_os = "watchos")))] {
         extern "C" {
+            #[cfg_attr(
+                target_os = "openbsd",
+                deprecated(since = "0.2.190", note = "Removed in OpenBSD 7.5")
+            )]
             pub fn syscall(num: c_int, ...) -> c_int;
         }
     }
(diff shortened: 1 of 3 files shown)


tgross35:
@semarie mind reviewing this?

semarie (approved):
LGTM

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5466/444db4456f3866519b7eacd35a8bb59484bb7bbf..4a94d2fcede7068e03ceee8cac3a44cc31d10ec6)*

PR #5469, opened 2026-09-04
qnx: Add missing definitions for io-sock

lambinoo:
I'm currently working on getting socket2 to compile for QNX8 networking stack. This PR adds some missing network related structs and constants.

### DIFF
--- a/src/unix/nto/io_sock/mod.rs
+++ b/src/unix/nto/io_sock/mod.rs
@@ -50,6 +50,18 @@ s! {
         pub sdl_slen: c_uchar,
         pub sdl_data: [c_char; 46],
     }
+
+    pub struct ip_mreqn {
+        pub imr_multiaddr: crate::in_addr,
+        pub imr_address: crate::in_addr,
+        pub imr_ifindex: c_int,
+    }
+
+    pub struct ip_mreq_source {
+        pub imr_multiaddr: crate::in_addr,
+        pub imr_sourceaddr: crate::in_addr,
+        pub imr_interface: crate::in_addr,
+    }
 }
 
 pub const SCM_CREDS: c_int = 0x03;
@@ -89,6 +101,12 @@ pub const PF_NATM: c_int = AF_NATM;
 pub const pseudo_AF_HDRCMPLT: c_int = 31;
 pub const SIOCGIFADDR: c_int = u32_cast_int(0xc0206921);
 pub const SO_SETFIB: c_int = 0x1014;
+pub const TCP_KEEPIDLE: c_int = 256;
+pub const TCP_KEEPINTVL: c_int = 512;
+pub const TCP_KEEPCNT: c_int = 1024;
+pub const IP_RECVTOS: c_int = 68;
+pub const IP_ADD_SOURCE_MEMBERSHIP: c_int = 70;
+pub const IP_DROP_SOURCE_MEMBERSHIP: c_int = 71;
 
 extern "C" {
     pub fn sendmmsg(


src/unix/nto/io_sock/mod.rs
```diff
@@ -50,6 +54,20 @@ s! {
         pub sdl_slen: c_uchar,
         pub sdl_data: [c_char; 46],
     }
+
+    // https://github.com/freebsd/freebsd-src/blob/f0cf0b8266eef39b13917f7bed808daf6d6a2d3e/sys/netinet/in.h#L549
+    pub struct ip_mreqn {
+        pub imr_multiaddr: in_addr,
+        pub imr_address: in_addr,
+        pub imr_ifindex: c_int,
+    }
```
tgross35: Please put these links into the commit message, rather than in comments

tgross35 (commented):
Looks reasonable enough to me but the target maintainers should review/test. Cc @flba-eb @gh-tr @jonathanpallant @japaric

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5469/9a7a2844e13ee4a38775b86f1e46f10ae977b83b..543e03ac9fc40c8629f0f1af77c0f0136f5e5f1a)*

src/unix/nto/io_sock/mod.rs
lambinoo: Updated! Also fixed the the `in_addr` fields to `crate::in_addr` since that's what the rest of the file is using

src/unix/nto/io_sock/mod.rs
```diff
@@ -6,6 +6,7 @@
 //! * `x86_64-pc-nto-qnx710`
 //! * `aarch64-unknown-qnx`
 //! * `x86_64-pc-qnx`
+//!
```
tgross35, line 9: Looks spurious

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5469/7e30c329edc6f96fb2d5190d71cbbeb715432c75..53c3ebee5652dea69887be51c2ea6e6518a78caa)*

tgross35:
Also this is labled "fix" but this is just an API addition. For future reference we don't use conventional commits but instead use scoping, makes it much easier to get a changelog (no need to change it's just an FYI)

tgross35 (approved):
LGTM, just waiting on a target maintainer to confirm

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5469/7e9f74959ca7523465236e116bfda9714b3eaeab..5702a8e9b0af3f7692dbc837d68dfe4ce2599f7c)*

jonathanpallant:
I would strongly advise against linking to the FreeBSD headers in the commit message. The *only* place to check these constants is in QNX SDP 8.0 and/or its online documentation, because whilst it might be *based* on FreeBSD, QNX could have changed any of these constants to suit their needs.

Using `target/qnx/usr/include/netinet/in.h` in my copy of QNX SDP 8, I have verified:

* [x] `ip_mreqn`
* [x] `ip_mreq_source `
* [x] `TCP_KEEPIDLE`
* [x] `TCP_KEEPINTVL`
* [x] `TCP_KEEPCNT`
* [x] `IP_RECVTOS`
* [x] `IP_ADD_SOURCE_MEMBERSHIP`
* [x] `IP_DROP_SOURCE_MEMBERSHIP`

lambinoo:
Should I remove the references to freeBSD then from the commit message? Mostly wanted to link to publicly available headers, but it's true that the SDP is the only source of truth here.

jonathanpallant:
~~Also, the `ip_mreqn` and `ip_mreq_source` structures don't exist in QNX SDP 7.1 from what I can see, so these should probably be excluded from those targets.~~

Uh, scratch that. I didn't have `io-sock` installed. They look the same in QNX SDP 7.1.

lambinoo:
> ~Also, the `ip_mreqn` and `ip_mreq_source` structures don't exist in QNX SDP 7.1 from what I can see, so these should probably be excluded from those targets.~
> 
> Uh, scratch that. I didn't have `io-sock` installed. They look the same in QNX SDP 7.1.

Yeah, they should be in `usr/include/io-sock/netinet/in.h`. From my understand, QNX is backporting anything on the io-sock side from 8 to 7.1. So they should be the same most if not all the time.

lambinoo:
Removed any references to FreeBSD from the commit message

jonathanpallant (approved):

tgross35:
Thanks Jonathan for reviewing

PR #5491, opened 2026-09-06
build: Assert that cfg makes sense

tgross35:
Adjust configuration so that we are not e.g. setting `freebsd12` on
non-FreeBSD targets, and add assertions that similar configuration
setups make sense.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -36,12 +36,12 @@ enum Cfg {
 
     VxworksLt25_09,
 
-    /// Corresponds to `__USE_TIME_BITS64` in UAPI
+    /// Corresponds to `__USE_TIME_BITS64` in UAPI. Implies 32-bit Linux target.
     LinuxTimeBits64,
-    /// Corresponds to `_FILE_OFFSET_BITS=64` in glibc
+    /// Corresponds to `_FILE_OFFSET_BITS=64` in glibc. Implies 32-bit GNU target.
     GnuFileOffsetBits64,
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -36,12 +36,12 @@ enum Cfg {
...
-    /// Corresponds to `__USE_TIME_BITS64` in UAPI
+    /// Corresponds to `__USE_TIME_BITS64` in UAPI. Implies 32-bit Linux target.
     LinuxTimeBits64,
-    /// Corresponds to `_FILE_OFFSET_BITS=64` in glibc
+    /// Corresponds to `_FILE_OFFSET_BITS=64` in glibc. Implies 32-bit GNU target.
     GnuFileOffsetBits64,
-    /// Corresponds to `_TIME_BITS=64` in glibc. Also used in x86 Windows with
-    /// GNU to expose a 64-bit `time_t`.
+    /// Corresponds to `_TIME_BITS=64` in glibc. Also used in x86 Windows with GNU
+    /// to expose a 64-bit `time_t`. Implies 32-bit GNU target and 64-bit `off_t`.
```
dybucc, line 44: I don't think it's right to say that it implies 64-bit `off_t`. Windows
has no issues with `off_t`; It is `time_t` that is made 64-bits when
this `cfg` is issued.

tgross35: Good point, updating as part of https://github.com/rust-lang/libc/pull/5508

PR #5492, opened 2026-09-06
Make `rustc-dep-of-std` work with `extra-traits`

### LINKED ISSUE #2064: Features 'rustc-dep-of-std' and 'extra_traits' don't work together

tgross35:
All that is needed is some import adjustments.

Closes: https://github.com/rust-lang/libc/issues/2064

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -74,7 +74,14 @@ macro_rules! prelude {
             #[allow(unused_imports)]
             pub(crate) use core::clone::Clone;
             #[allow(unused_imports)]
+            pub(crate) use core::cmp::{
+                Eq,
+                PartialEq,
+            };
+            #[allow(unused_imports)]
             pub(crate) use core::default::Default;
+            #[allow(unused_imports)]
+            pub(crate) use core::iter::Iterator;
             #[allow(unused_imports)]
             pub(crate) use core::marker::{
                 Copy,
@@ -89,12 +96,14 @@ macro_rules! prelude {
             pub(crate) use core::{
                 assert,
                 cfg,
+                compile_error,
                 debug_assert,
                 fmt,
                 hash,
                 iter,
                 mem,
                 ptr,
+                unimplemented,
             };
 
             #[allow(unused_imports)]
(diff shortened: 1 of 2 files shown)


xtqqczze:
Contributes to https://github.com/rust-lang/libc/issues/5496.

PR #5495, opened 2026-09-06
libc: Make `rustc-dep-of-std` work in tests

xtqqczze:
Contributes to https://github.com/rust-lang/libc/issues/5496.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -833,6 +833,7 @@ macro_rules! offset_of {
 #[cfg(test)]
 mod tests {
     use core::any::TypeId;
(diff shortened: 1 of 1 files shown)


tgross35 (approved):
I don't think this cfg really needs to work, but also doesn't hurt. So thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5495/7e30c329edc6f96fb2d5190d71cbbeb715432c75..606b21c9eb6c7ef6982f6b941ea798a982b890d1)*

PR #5515, opened 2026-09-08
Redisable constants for L4Re

farao:
These constants were disabled for uclibc (and therefore implicitly disabled for L4Re). Now that they are not disabled for uclibc anymore, explicitly disable them for L4Re.


tgross35 (approved):
Thanks!

By the way if you happen to have a test setup for l4re that tests libc automatically (or could do so), consider linking at https://github.com/rust-lang/libc/issues/5209.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5515/eb975fffca9e8127ad9d61e0437ef7f03cfb8afa..0e696f351540790f0972737b45ad04fb47e8e879)*

PR #5517, opened 2026-09-08
build: Don't warn when enabling opting into musl 1.2 via multiple ways

Gelbpunkt:
In Alpine, we have to set both the old and new `cfg` *and* the environment variable since various versions of libc are being used by packaged software.

This will then emit the warnings about the options being deprecated, even though we fully intentionally set them. I believe the primary purpose of the warnings is to tell people to migrate to the new `cfg`, so we can just not print a warning if musl 1.2 is being opted into via a different way (and the `cfg` or env var has no effect) already.


tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5517/7470c73a1f49ccb290c0ae407cdf3e17dc9af10d..54e23715497c61da9c3be943c9c7400eb23847c0)*

PR #5518, opened 2026-09-08
glibc: add `__f_unused` to `statvfs64` on riscv32

### LINKED ISSUE #5512: glibc: statvfs64 is missing `__f_unused` on riscv32-gnu
Currently `statvfs64` on riscv32 layout seems inconsistent with it neighboring `statvfs`, and the source in glibc. It creates different layout that may be expected, so may cause UB because of shifted bytes on the struct.
https://github.com/rust-lang/libc/blob/008f905256d27eb5dd416d0976678ee6c2855554/src/new/glibc/sysdeps/unix/linux/bits/statvfs.rs#L16-L23

https://github.com/rust-lang/libc/blob/008f905256d27eb5dd416d0976678ee6c2855554/src/new/glibc/sysdeps/unix/linux/bits/statvfs.rs#L40-L50

In glibc, both of them have __f_unused for 32-bit targets.
https://github.com/bminor/glibc/blob/04e750e75b73957cf1c791535a3f4319534a52fc/sysdeps/unix/sysv/linux/bits/statvfs.h#L49-L51
https://github.com/bminor/glibc/blob/04e750e75b73957cf1c791535a3f4319534a52fc/sysdeps/unix/sysv/linux/bits/statvfs.h#L70-L72
https://github.com/bminor/glibc/blob/04e750e75b73957cf1c791535a3f4319534a52fc/sysdeps/unix/sysv/linux/bits/statvfs.h#L24-L27

It was found when working on #5434, during the migration and deduplication to `new`, the missing field of `statvfs64` on riscv32 was deliberately left out as it was before to not change layout of it , but based on the source it should have that field, or are there any specific reason it didn't get that field?

Yasser-Ameur:
`statvfs64` is missing `__f_unused` on riscv32, so every field after `f_fsid`
sits four bytes off.

glibc gates both structs on the same macro and riscv32 satisfies it: `__WORDSIZE`
is 32 there and nothing defines `__SYSCALL_WORDSIZE`. `statvfs` already carries
the field. This was the FIXME left in rust-lang/libc#5434.

There is no riscv32 ctest job, so I checked with a probe instead:
`offset_of!(statvfs64, f_flag)` moves 60 to 64 on riscv32 and stays 72 on
x86_64.

Closes rust-lang/libc#5512

### DIFF
--- a/src/new/glibc/sysdeps/unix/linux/bits/statvfs.rs
+++ b/src/new/glibc/sysdeps/unix/linux/bits/statvfs.rs
@@ -37,15 +37,12 @@ s! {
         pub f_ffree: u64,
         pub f_favail: u64,
         pub f_fsid: c_ulong,
-        // FIXME(riscv32): glibc declares this field on riscv32 too, but we have
-        // never declared it here.
+        // Mirrors `_STATVFSBUF_F_UNUSED` in the header. x32 is excluded because
+        // its `__SYSCALL_WORDSIZE` is 64, aarch64 because glibc always sets
+        // `__WORDSIZE` to 64.
         #[cfg(all(
             target_pointer_width = "32",
-            not(any(
-                target_arch = "x86_64",
-                target_arch = "aarch64",
-                target_arch = "riscv32"
-            ))
+            not(any(target_arch = "x86_64", target_arch = "aarch64"))
         ))]
         __f_unused: Padding<c_int>,
         pub f_flag: c_ulong,


tgross35 (changes requested):
That's a rather long description for a simple change. Please handwrite all communication with users, including PR descriptions and commit messages.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5518/df6561676506cd9c36ebe84b3b2f000661b64767..30b029529e89c69b60a8a2900e52726b4a5573ea)*

tgross35 (approved):
Thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5518/df6561676506cd9c36ebe84b3b2f000661b64767..ff2c5ec66c216c239ecc49b650752fb962abbdeb)*

Issue #4650, opened 2025-08-13
ctest: Figure out a way to test anonymous structs/unions


tgross35:
We should provide a `cfg` way to identify a union or struct field as anonymous. This can probably be a `is_field_anonymous` function that would make `ctest` "inline" the relevant tests.

This discussion was started at https://github.com/rust-lang/libc/pull/4610#discussion_r2272226633 but I don't think we actually need that tricky regex pattern.

Issue #4649, opened 2025-08-13
ctest: Figure out a way to test private items


tgross35:
We have some definitions that aren't crate-public but come from C, so we should probably test them. One idea about how to do this is to just paste private items into the output file, since we have the AST version available.

Mentioned at https://github.com/rust-lang/libc/pull/4610#discussion_r2272258882

mbyx:
Would it be possible to just walk the entire crate and make everything public by modifying the AST, then just testing the new version? Kind of like what `src-hotfix` does right now by replacing the `::`.

xtqqczze:
The quick-and-dirty approach would be to make all of the private fields `pub(crate)`. This is what `rustix` does, for example.

tgross35:
I'd be fine with that visibility change but does that help us? libc-test is a separate crate.

We could probably make it `include!` or `#[path = "..."]` but I assume that would come with other issues.

PR #5521, opened 2026-09-11
musl: fix `siginfo_t` definition

dybucc:
## Description

rust-lang/libc#5345 messed up one thing about the musl bindings; Fields `__first` and
`__second` from `__si_common` in the untagged union were made records
instead of untagged unions themselves.

## Checklist

  standard doc comment


tgross35 (approved):
Huh, why didn't this show up in tests? I guess they're private and the structs here were still smaller than any other union variant?

In any case thanks for the fix, LGTM.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5521/df6561676506cd9c36ebe84b3b2f000661b64767..fbe44d5eba8374f1e08fe336df674bc8aaa7ac6c)*

xtqqczze:
> Huh, why didn't this show up in tests?

It showed up in https://github.com/nix-rust/nix/pull/2821#issuecomment-5625978442. Going forward, it might be worth testing `libc` against a few downstream crates as well, since I only happened to catch this because it surfaced there when compiling against the `libc-0.2` branch of `libc`.

tgross35:
I would love to have that kind of feedback, though I don't think we can commit to maintaining CI for downstream projects here. Maybe interested projects (nix, rustix, etc) could add a [`repository_dispatch`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#repository_dispatch) trigger to their workflow that runs on pushes to `libc-0.2` (use `patch.crates-io` to link to the git), then we link it here https://github.com/rust-lang/libc/issues/5209.

xtqqczze:
@tgross35 Regarding https://github.com/nix-rust/nix/pull/2821 and https://github.com/bytecodealliance/rustix/pull/1676, is there anyone you can reach out to who might be able to help get these merged soon? Otherwise, these crates will break with the next `libc` release, and users will need to pin `libc`.

xtqqczze:
@tgross35 It would also help if you could backport these changes soon, so I can test against the `libc-0.2` branch rather than my own cherry-picked [`libc-0.2-fix-siginfo_t`](https://github.com/xtqqczze/rust-lang-libc/tree/libc-0.2-fix-siginfo_t) branch.

tgross35:
Pinged on the PR for one and on Zulip for the other. Thank you for doing the work to get that together! I'm loosely holding off the release waiting on the rustix fix, but that issue has been open for a long time so I'm not sure it's worth waiting much longer.

dybucc:
>Huh, why didn't this show up in tests? I guess they're private and the structs
>here were still smaller than any other union variant?
>
>In any case thanks for the fix, LGTM.
>
>*[View changes since this
>review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5521/df6561676506cd9c36ebe84b3b2f000661b64767..fbe44d5eba8374f1e08fe336df674bc8aaa7ac6c)*

We don't test against C anonymous `union`s. The affected fields are anonymous
`union`s within anonymous `union`s (not directly the descendants of the `union`s
in the `siginfo_t`, though.)

Though this type seems to be popular enough to merit a test under
`libc-test/tests`.

tgross35:
I was thinking it would affect the shape of `siginfo_t`, but I guess there were larger variants. Would love to test anonymous unions, I think it should be possible https://github.com/rust-lang/libc/issues/4650. But I'm not sure that would help here because the fields are private https://github.com/rust-lang/libc/issues/4649.

In any case, a dedicated test sounds great to me.

PR #5522, opened 2026-09-11
qnx: add missing definitions for io-sock

lambinoo:
This is a backport of PR rust-lang/libc#5469 to enable socket2 and other dependencies to use this


tgross35:
Thanks for the interest, but backports are done by maintainers before release

PR #5524, opened 2026-09-11
CI: migrate aarch64 Android to Cuttlefish

10ne1:
Hello, this moves the aarch64 Android CI from the decade old legacy emulator (Android 7) to the modern Android  17 matching the existing x86_64 test infra.

 **Deliberately not included**, migrating these is left for future PRs:
 1. `arm-linux-androideabi` needs more investigation.
 2. `aosp_cf_arm64_only_phone` is using a 64-bit-only image, so moving it needs a different image variant and its own investigation.

Once the above 2 are also migrated, the legacy emulator logic can be removed entirely.

Running aarch64 tests is still pretty slow because it runs via emulation (no KVM on github arm runners), but it is booting a full modern Android and especially unblocks PRs like https://github.com/rust-lang/libc/pull/4459.

The approaches compared, time-wise:
  - qemu-user (PR https://github.com/rust-lang/libc/pull/5281): **4m2s**. Fastest by far, but it is a userspace ABI shim against the bionic libs, not a running Android, so there is no kernel and no real device behind the syscalls.
  - legacy SDK emulator (current `main`): **11m40s - 14m28s**. A full system, but an  Android 7 / API 24 image from 2016 which is buggy and blocks PRs like rust-lang/libc#4459.
  - Cuttlefish through qemu (this PR), full Android 17, ABI `arm64-v8a`: **~36m**, of which 14--21m is the boot depending on load and only ~6m is actually running tests.
  - Cuttlefish on x86_64 with crosvm + KVM (current `main`, for reference):
    **~6m10s**. This is roughly what aarch64 would cost if github's arm64 runners  exposed `/dev/kvm`. The gap is the price of TCG.

Credit goes to @maurer for pointing out that Cuttlefish can boot an arm64 guest on an x86_64 host: it drops `-accel` and swaps `-cpu host` for `-cpu max` when the guest arch does not match the host, so `-vm_manager=qemu_cli` runs the device under TCG.

This worked nicely so we don't need an extra layer like [android-emulator-runner](https://github.com/reactivecircus/android-emulator-runner) which unfortunately only works with images up to 2022.

Please see more details in the commit messages.

cc @tgross35


tgross35:
> * qemu-user (PR [ci: Run aarch64-linux-android tests with qemu-user #5281](https://github.com/rust-lang/libc/pull/5281)): **4m2s**. Fastest by far, but it is a userspace ABI shim against the bionic libs, not a running Android, so there is no kernel and no real device behind the syscalls.

What's the downside of not having a kernel? Most of our tests look at ABI shape, with the exception of function pointer checks which I presume would still work with that version.

I don't think we can accept a job that takes longer than ~10 minutes, currently this more than doubles total CI time. That's a big tax on every PR and not worth it for a single tier 2 target.

There are two ways to get this covered that could be easier in some ways than the hoops you've had to jump through:

* If you or somebody else interested in Android have the contacts to set up self-hosted runners, we could enable them here similar to https://forge.rust-lang.org/infra/docs/external-ci-runners.html. These could be e.g. aarch64 runners with kvm, or something dedicated for Android that keeps the emulator booted.
* Just test downstream https://github.com/rust-lang/libc/issues/5209, which does mean post-merge but can do more complex things and run for longer than we'd like to have here.

PR #5527, opened 2026-09-13
ci: Test riscv64gc-unknown-linux-musl

Gelbpunkt:
This is a tier 2 with host tools target since 1.99 but has been tier 2 for a while longer.


src/unix/linux_like/linux/musl/b64/riscv64/mod.rs
```diff
@@ -47,17 +47,15 @@ s! {
     }
 
     pub struct ipc_perm {
-        pub __key: crate::key_t,
+        pub __ipc_perm_key: crate::key_t,
```
tgross35: Huh; is https://github.com/rust-lang/libc/commit/3f81aadb0f1d1381c78f9b49da5267b3c466b138 not correct then?

I'd like to have this clarified but the rest LGTM

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5527/ca0144d6c9e0e8f9722f928ff4f69398e93e5213..70b7b9b3a819d20009a736dcf74a6688aab4e770)*

Gelbpunkt: Hmmm, in the headers this is `__ipc_perm_key`, but looking at https://git.musl-libc.org/cgit/musl/commit/?id=d1b6fc6ecc59441a37269d6512c5d34d1a2412cd, it looks like it's `__key` if included via `sys/ipc.h`

Gelbpunkt: It seems like `__key` is the intended name, I'll see if I can get ctest to pass without this change

Gelbpunkt: Works!

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5527/0c0db5cd9f3c658d6b511328a6dd823a270d82cb..f645f199ecc2ea4368448fde45ee34b194ccfd95)*

Issue #1401, opened 2019-06-14
Panic at FD_SET


killme2008:
* Platform:  Linux dennis-thinkpad 4.15.0-39-generic #42-Ubuntu SMP Tue Oct 23 15:48:01 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
* Info:   When try to create more connections to a mysql server, panic at `FD_SET`.
* Error:

```
panicked 'index out of bounds: the len is 16 but the index is 53' at "/root/.cargo/registry/src/mirrors.ustc.edu.cn-12df342d903acd47/libc-0.2.58/src/unix/notbsd/mod.rs:1177"
```

* code line:  https://github.com/rust-lang/libc/blob/0.2.58/src/unix/notbsd/mod.rs#L1177

killme2008:
We don't use libc derectly, but use [nix-0.11.0](https://github.com/nix-rust/nix) such as

```rust
    let mut fd_set = select::FdSet::new();
    let socket_fd = fd;
    fd_set.insert(socket_fd);
```

Panic at ` fd_set.insert(socket_fd)`.

gnzlbg:
@killme2008 could you compile with `--features=extra_traits`, and dump `fd_set` and `socket_fd` before the insert call

```rust
    let mut fd_set = select::FdSet::new();
    let socket_fd = fd;
    dbg!(fd_set);
    dbg!(socket_fd);
    fd_set.insert(socket_fd);
```

?

cc @asomers

asomers:
This is probably a bug in your application.  `select` can only be used with file descriptors up to `FD_SETSIZE`.

gnzlbg:
@asomers does nix expose a "safe" wrapper over these APIs ?

asomers:
No need.  libc's exported `FD_SET` is already safe.  That's why the application panic()ed rather than segfaulted.

gnzlbg:
Ok, then I'm going to leave this open here. While I agree with you that the bug is in user code, I think we should be providing a better error message in this situation, e.g., by asserting at the beginning of the function that the indices must be in bounds, and explaining what those bounds are, at least in debug mode via a `debug_assert!`.

tgross35:
[...]

PR #5528, opened 2026-09-13
linux: Improve panic message for out-of-range fds in FD_* functions

danielcanencia:
Part of rust-lang/libc#1401

Fixes the out-of-bounds FD_SET/FD_CLR/FD_ISSET panic for Linux
(linux_like). The issue also calls for the same clear panic message in
other platforms; those remain a follow-up (cygwin, haiku, solaris, etc).

Note:
- FD_ZERO needs no change — it takes no fd and fills the set, so there
  is nothing to range-check.


src/unix/linux_like/mod.rs
```diff
@@ -1847,21 +1847,26 @@ f! {
     pub unsafe fn FD_CLR(fd: c_int, set: *mut fd_set) -> () {
         let fd = fd as usize;
         let size = size_of_val(&(*set).fds_bits[0]) * 8;
-        (*set).fds_bits[fd / size] &= !(1 << (fd % size));
-        return;
+        *(*set).fds_bits.get_mut(fd / size).unwrap_or_else(|| {
+            core::panic!("fd {fd} out of range: valid fds are 0..FD_SETSIZE (0..{FD_SETSIZE})")
+        }) &= !(1 << (fd % size));
```
tgross35, line 1852: This is tough to read, use `let Some(...) = ... else { ... }` instead

src/unix/linux_like/mod.rs
```diff
@@ -1847,21 +1847,26 @@ f! {
     pub unsafe fn FD_CLR(fd: c_int, set: *mut fd_set) -> () {
         let fd = fd as usize;
         let size = size_of_val(&(*set).fds_bits[0]) * 8;
-        (*set).fds_bits[fd / size] &= !(1 << (fd % size));
-        return;
+        *(*set).fds_bits.get_mut(fd / size).unwrap_or_else(|| {
+            core::panic!("fd {fd} out of range: valid fds are 0..FD_SETSIZE (0..{FD_SETSIZE})")
```
tgross35, line 1851: Add `panic!` to the prelude (update the `macro_rules! prelude`)

src/unix/linux_like/mod.rs
```diff
@@ -1847,21 +1847,26 @@ f! {
     pub unsafe fn FD_CLR(fd: c_int, set: *mut fd_set) -> () {
```
tgross35, line 1847: Could you please update these functions everywhere they exist? They exist in a lot of modules, not just Linux.

tgross35 (changes requested):
Thanks for putting this together!

> The issue also calls for the same clear panic message in other platforms; those remain a follow-up (cygwin, haiku, solaris, etc).

I'd prefer to get them all done at the same time.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5528/ca0144d6c9e0e8f9722f928ff4f69398e93e5213..444c960cdfb46cda789d115ab5d56d01a133b128)*

src/unix/linux_like/mod.rs
danielcanencia: Sure. I'll take a look at it in the next few days.

PR #5529, opened 2026-09-15
musl: Move `statfs.h` and `statvfs.h` definitions to `src/new`

valentynkit:
This follows the ongoing migration into `src/new`, which reduces
duplication and models musl's source tree.

Arch specific definitions are kept for `mips`, `mips64` and `s390x`,
mirroring musl's `arch/*/bits/statfs.h`.

`statfs64` is now a type alias for `statfs`, as it already was on
`s390x`. musl has no second struct: `sys/statfs.h` `#define`s
`statfs64` to `statfs`, and the lfs types `fsblkcnt64_t` and
`fsfilcnt64_t` to `fsblkcnt_t` and `fsfilcnt_t`. This crate declares
both spellings as `c_ulonglong` on every musl target, so the duplicated
structs had identical layout. 

`statvfs64` is now also a type alias for `statvfs` following the same reasons as before.

Let me know if I should revert the changes for using type alias, and just do the migration without it, but from my findings it looks safe to do.
https://github.com/kraj/musl/blob/1c82569f4cc53f5bb25c820051c2a02463f46cab/include/sys/statfs.h#L21-L26
https://github.com/rust-lang/libc/blob/b99bf9dd59aae009ef10cea14785be50d6fb5aa3/src/unix/linux_like/linux/musl/mod.rs#L39-L46

Link: https://github.com/kraj/musl/blob/1c82569f4cc53f5bb25c820051c2a02463f46cab/include/sys/statfs.h
Link: https://github.com/kraj/musl/blob/1c82569f4cc53f5bb25c820051c2a02463f46cab/include/sys/statvfs.h
Link: https://github.com/kraj/musl/blob/1c82569f4cc53f5bb25c820051c2a02463f46cab/arch/s390x/bits/statfs.h
Link: https://github.com/kraj/musl/blob/1c82569f4cc53f5bb25c820051c2a02463f46cab/arch/mips/bits/statfs.h
Link: https://github.com/kraj/musl/blob/1c82569f4cc53f5bb25c820051c2a02463f46cab/arch/mips64/bits/statfs.h


valentynkit:
These changes follow from previous work in this repo, the ongoing migration to src/new suggests contributions like this are useful, but let me know if this isn't wanted

 Two things worth attention: 
 - making `statfs64`/ `statvfs64` type aliases instead of duplicate structs (but some fields use lfs bindings like  `fsblkcnt64_t`)
 - musl `statvfs` has `f_type`, so it was not added in this PR (glibc side #5434), to keep the move PR separate as it already contains the move of 2 definitions.

PR #5537, opened 2026-09-20
Unix: deprecate sighandler_t on non-GNU platforms

chengr4:
This PR is a split-out part and the first step of rust-lang/libc#5445 to make the review easier.
The target is to fix part of rust-lang/libc#1273 and rust-lang/libc#1359 and **backport to 0.2**.

Checked, available linux-like platforms:

- (glibc) https://github.com/sailfishos-mirror/glibc/blob/d5bd0f22305277fa0d431bb9d733aa9456706573/signal/signal.h#L188
- (musl) https://github.com/kraj/musl/blob/1c82569f4cc53f5bb25c820051c2a02463f46cab/include/signal.h#L273
- (uClibc-ng) https://github.com/wbx-github/uclibc-ng/blob/3f2ef085d8455a13fcdc4d4d473c0385dd65d55e/include/signal.h#L226
- (Android) https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/bits/signal_types.h;l=73;drc=61197364367c9e404c7da6900658f1b16c42d0da
- (Emscripten) https://github.com/emscripten-core/emscripten/blob/d6c521a7f05449857c76bd99e396895583cf2083/system/lib/libc/musl/include/signal.h#L273
- (L4Re) https://github.com/kernkonzept/l4re-core/blob/941f09bbf01ee153c5b66caff87c868deab3c91c/libc/musl/contrib/musl/include/signal.h#L278


src/unix/bsd/mod.rs
```diff
@@ -30,6 +30,11 @@ pub type regoff_t = c_int;
 #[cfg(not(target_os = "dragonfly"))]
 pub type regoff_t = off_t;
 pub type sig_t = Option<unsafe extern "C" fn(c_int)>;
+#[deprecated(
+    since = "0.2.191",
+    note = "sighandler_t is a GNU extension and not supported on this platform; use sig_t instead."
+)]
+pub type sighandler_t = size_t;
```
chengr4, line 37: I am not quite sure whether should we add this back to BSD family.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/libc/5537/35c8b7d847f64265cb578343dc5d1be467fe63db..1ae3ead6bf1b08d13e491abe666dc6105b487276)*

PR #5540, opened 2026-09-20
uclibc: Replace `linux_time_bits64` with a new  `uclibc32_time64` cfg

tgross35:
`linux_time_bits64` should correlate to the kernel's
`__USE_TIME_BITS64`, which is set on both 32- and 64-bit platforms where
`time_t` is 64 bits. The uclibc version of this config is
`UCLIBC_USE_TIME64` which is only set on 32-bit platforms.

Split this up by introducing `uclibc32_time64` to closer match
`UCLIBC_USE_TIME64`. As an effect, setting `libc_unstable_uclibc_time64`
no longer crashes builds for 64-bit uclibc targets. The setup also
closer matches what is done for GNU targets.

Note that `linux_time_bits64` is not actually the "64-bit time_t on any
platform" yet that the kernel's version is, so it still isn't set on
64-bit platforms after this change.

* `UCLIBC_USE_TIME64` definition:
  https://github.com/wbx-github/uclibc-ng/blob/3f2ef085d8455a13fcdc4d4d473c0385dd65d55e/extra/Configs/Config.in#L1036-L1055
* Implication of `__USE_TIME_BITS64`:
  https://github.com/wbx-github/uclibc-ng/blob/3f2ef085d8455a13fcdc4d4d473c0385dd65d55e/include/features.h#L458-L460
* Entrypoint for 64-bit `time_t`:
  https://github.com/wbx-github/uclibc-ng/blob/3f2ef085d8455a13fcdc4d4d473c0385dd65d55e/libc/sysdeps/linux/common/bits/typesizes.h#L50-L54
* Arm `struct shimd_ds`:
  https://github.com/wbx-github/uclibc-ng/blob/3f2ef085d8455a13fcdc4d4d473c0385dd65d55e/libc/sysdeps/linux/arm/bits/shm.h#L48-L69


tgross35:
Cc @skrap, though this should be a nop change

build.rs
```diff
@@ -54,6 +54,9 @@ enum Cfg {
     /// transition to 64-bit `time_t` and thus need `__*_time64` redirects. Implies `musl_v1_2`
     /// and 32-bit arch.
     MuslRedirTime64,
+    /// Corresponds to the `UCLIBC_USE_TIME64` config / `__UCLIBC_USE_TIME64__` define. Indicates
+    /// that `*time64` syscalls should be used. Implies 32-bit.
```
dybucc, line 58: Not sure if we should mention here that it also applies to 64-bit MIPS
with the 32-bits ABI.

